PR #2, opened 2018-03-04
Option to remove formatting before diff

70ray:
new file unformat.inc with function to remove formatting
and either completely unwrap or keep line structure but
remove leading and trailing spaces

diff.php: add option to remove formatting (not unwrap)
before compare.
If at least one round is F then show link option to remove
formatting or not.
Default is remove for P to F, keep for F to F

new file page_compare.php to compare pages with format
removed and unwrapped between selected rounds for a
project, either all pages or for the users pages in either
the first or second round. Makes a list of page image names
which link to diff.php

project.php: link added to page_compare.php if project has
reached formatting stage

### DIFF
--- a/pinc/Project.inc
+++ b/pinc/Project.inc
@@ -583,6 +583,29 @@ XML;
         }
         return FALSE;
     }
+    
+    public function check_pages_table_exists()
+    {
+        if (!$this->pages_table_exists)
+        {
+            echo "<p class='warning'>";
+            if ($this->archived != 0)
+            {
+                echo _("The project has been archived, so page details are not available.");
+            }
+            elseif ($this->state == PROJ_DELETE)
+            {
+                echo _("The project has been deleted, so page details are not available.");
+            }
+            else
+            {
+                echo _("Page details are not available for this project.");
+            }
+            echo "</p>";
+            return false;
+        }
+        return true;
+    }
 }
 
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
--- a/pinc/Project.inc
+++ b/pinc/Project.inc
@@ -584,26 +584,25 @@ XML;
         return FALSE;
     }
     
-    public function check_pages_table_exists()
+    public function check_pages_table_exists(&$message)
     {
         if (!$this->pages_table_exists)
         {
-            echo "<p class='warning'>";
             if ($this->archived != 0)
             {
-                echo _("The project has been archived, so page details are not available.");
+                $message = _("The project has been archived, so page details are not available.");
             }
             elseif ($this->state == PROJ_DELETE)
             {
-                echo _("The project has been deleted, so page details are not available.");
+                $message = _("The project has been deleted, so page details are not available.");
             }
             else
             {
-                echo _("Page details are not available for this project.");
+                $message = _("Page details are not available for this project.");
             }
-            echo "</p>";
             return false;
         }
+        $message = "";
         return true;
     }
 }
--- a/pinc/Project.inc
+++ b/pinc/Project.inc
@@ -6,6 +6,7 @@ include_once($relPath.'forum_interface.inc'); // topic_create
 include_once($relPath.'SettingsClass.inc');
 include_once($relPath.'User.inc');
 include_once($relPath.'MARCRecord.inc');
+include_once($relPath.'stages.inc'); // is_formatting_round()
 
 class NonexistentProjectException extends Exception { }
 
@@ -565,6 +566,22 @@ XML;
         $updated_record->load_yaz_array(unserialize(base64_decode($row["updated_array"])));
 
         return $updated_record;
+    }
+
+    // check if project has entered a formatting round
+    public function has_entered_formatting_round()
+    {
+        global $PROJECT_STATES_IN_ORDER;
+        $state_index = array_search($this->state, $PROJECT_STATES_IN_ORDER);
+        for($i=$state_index; $i>=0; $i--)
+        {
+            $round = get_Round_for_project_state($PROJECT_STATES_IN_ORDER[$i]);
+            if($round && is_formatting_round($round))
+            {
+                return TRUE;
+            }
+        }
+        return FALSE;
     }
 }
 
--- a/project.php
+++ b/project.php
@@ -688,7 +688,7 @@ function do_project_info_table()
         }
         else
         {
-            if ($project->pages_table_exists)
+            if($project->check_pages_table_exists($detail))
             {
                 $url = "$code_url/tools/project_manager/page_detail.php?project=$projectid&amp;show_image_size=0";
                 $blurb = _("Images, Pages Proofread, & Differences");
@@ -700,16 +700,6 @@ function do_project_info_table()
                     $url3 = "$code_url/tools/project_manager/page_compare.php?project=$projectid";
                     $blurb3 = _("Compare without formatting");
                     $detail .= " &middot; <a href='$url3'>$blurb3</a>";
-                }
-            }
-            else
-            {
-                if ($project->archived != 0) {
-                    $detail = _("The project has been archived, so page details are not available.");
-                } elseif ($project->state == PROJ_DELETE) {
-                    $detail = _("The project has been deleted, so page details are not available.");
-                } else {
-                    $detail = _("Page details are not available for this project.");
                 }
             }
             echo_row_a( _("Page Detail"), $detail);
--- a/project.php
+++ b/project.php
@@ -694,12 +694,12 @@ function do_project_info_table()
                 $blurb = _("Images, Pages Proofread, & Differences");
                 $url2 = "$url&amp;select_by_user";
                 $blurb2 = _("Just my pages");
-                $detail = "<a href='$url'>$blurb</a> | <a href='$url2'>$blurb2</a>";
+                $detail = "<a href='$url'>$blurb</a> &middot; <a href='$url2'><b>$blurb2</b></a>";
                 if($project->has_entered_formatting_round())
                 {
                     $url3 = "$code_url/tools/project_manager/page_compare.php?project=$projectid";
                     $blurb3 = _("Compare without formatting");
-                    $detail .= " | <a href='$url3'>$blurb3</a>";
+                    $detail .= " &middot; <a href='$url3'>$blurb3</a>";
                 }
             }
             else
--- a/project.php
+++ b/project.php
@@ -694,7 +694,13 @@ function do_project_info_table()
                 $blurb = _("Images, Pages Proofread, & Differences");
                 $url2 = "$url&amp;select_by_user";
                 $blurb2 = _("Just my pages");
-                $detail = "<a href='$url'>$blurb</a> &gt;&gt;<a href='$url2'>$blurb2</a>&lt;&lt;";
+                $detail = "<a href='$url'>$blurb</a> | <a href='$url2'>$blurb2</a>";
+                if($project->has_entered_formatting_round())
+                {
+                    $url3 = "$code_url/tools/project_manager/page_compare.php?project=$projectid";
+                    $blurb3 = _("Compare without formatting");
+                    $detail .= " | <a href='$url3'>$blurb3</a>";
+                }
             }
             else
             {
@@ -2160,4 +2166,3 @@ function do_page_table()
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 
 // vim: sw=4 ts=4 expandtab
-?>
--- a/styles/layout.less
+++ b/styles/layout.less
@@ -881,5 +881,15 @@ table.search-column {
     }
 }
 
+/* ------------------------------------------------------------------------ */
+/* Page compare grid */
+
+.grid-wrapper {
+    display: grid;
+    grid-template-columns: max-content max-content;
+    grid-column-gap: 2em;
+    grid-row-gap: 2px;
+}
+
 /* vim: sw=4 ts=4 expandtab
 */
--- a/tools/project_manager/diff.php
+++ b/tools/project_manager/diff.php
@@ -67,19 +67,10 @@ else
     $R_format = is_formatting_round($R_round);
 }
 
-
-switch($format)
+if(!$format) // no parameter passed
 {
-    case "keep":
-        $remove_format = false;
-        break;
-    case "remove":
-        $remove_format = true;
-        break;
-    default:
-        // remove format if only one of the rounds is formatting
-        $remove_format = ($L_format xor $R_format);
-        break;
+    // default to remove format if only one of the rounds is formatting
+    $format = ($L_format xor $R_format) ? "remove" : "keep";
 }
 
 $query = "
@@ -96,7 +87,7 @@ if ( $can_see_names_for_this_page) {
     $R_label .= " ($R_user)";
 }
 
-if($remove_format)
+if($format == "remove")
 {
     // also remove blank lines and leading and trailing spaces
     $L_text = remove_formatting($L_text, false);
@@ -124,7 +115,7 @@ echo "<h1>$project_title</h1>\n";
 echo "<h2>$image_link</h2>\n";
 
 do_navigation($projectid, $image, $L_round_num, $R_round_num, 
-              $L_user_column_name, $L_user);
+              $L_user_column_name, $L_user, $format);
 echo $navigation_text;
 
 $url = "$code_url/project.php?id=$projectid&amp;expected_state=$state";
@@ -135,7 +126,7 @@ echo "\n<p><a href='$url'>$label</a>\n";
 if($L_format || $R_format)
 {
     // show option to change compare method
-    if($remove_format)
+    if($format == "remove")
     {
         $format_label = _("Compare with formatting");
         $_GET["format"] = "keep";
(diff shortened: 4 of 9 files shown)


pinc/unformat.inc
```diff
@@ -0,0 +1,42 @@
+<?php
+function is_format_round($label)
```
cpeel: We do this "is this round a formatting/proofreading round" in several places (eg: round_project_listing_presort() in `pinc/stages.inc`). We should move this into the Round class in `pinc/RoundDescriptor.inc`, something like:
```
    function is_proofreading_round()
    {
        return startswith($this->id, 'P');
    }

    function is_formatting_round()
    {
        return startswith($this->id, 'F');
    }
```

pinc/unformat.inc
```diff
@@ -0,0 +1,42 @@
...
+{
+    return "F" == $label[0];
+}
+
+function remove_formatting($text, $unwrap)
```
cpeel: This probably best belongs in `pinc/DPage.inc` since it manipulates the data for a page.

project.php
```diff
@@ -694,7 +695,14 @@ function do_project_info_table()
                 $blurb = _("Images, Pages Proofread, & Differences");
                 $url2 = "$url&amp;select_by_user";
                 $blurb2 = _("Just my pages");
-                $detail = "<a href='$url'>$blurb</a> &gt;&gt;<a href='$url2'>$blurb2</a>&lt;&lt;";
+                $detail = "<a href='$url'>$blurb</a> | <a href='$url2'>$blurb2</a>";
+                $state_index = array_flip($PROJECT_STATES_IN_ORDER);
+                if($state_index[$project->state] >= $state_index[PROJ_F1_AVAILABLE])
```
cpeel: Using PROJ_F1_AVAILABLE is non-ideal as in theory different sites can customize the different rounds. It would be more round-about, but probably overall better, if you did something like this (untested) within the Project class:
```
    public function has_been_through_formatting_round()
    {
        global $PROJECT_STATES_IN_ORDER;
        $state_index = array_search($this->state, $PROJECT_STATES_IN_ORDER);
        for($i=$state_index; $i>=0; $i--)
        {
            $round = get_Round_for_project_state($PROJECT_STATES_IN_ORDER[$i]);
            if($round)
            {
                if($round->is_formatting_round())
                    return TRUE;
            }
        }
        return FALSE;
    }
```

tools/project_manager/diff.php
```diff
@@ -61,6 +62,19 @@
     $R_label = $R_round->id;
 }
 
+$L_format = is_format_round($L_label);
+$R_format = is_format_round($R_label);
+$format = array_get($_GET, "format", "default");
```
cpeel: Move this to the top of the file with the other $_GET validation. It should also be using `get_enumerated_param()` with a default parameter instead of a blanket array_get().

tools/project_manager/diff.php
```diff
@@ -61,6 +62,19 @@
...
+if($format == "keep")
```
cpeel: Please use brackets in all if/then/else statements if you're going to use them in any of them.

tools/project_manager/page_compare.php
```diff
@@ -0,0 +1,182 @@
...
+include_once($relPath.'misc.inc');
+include_once($relPath.'theme.inc');
+include_once($relPath.'Project.inc');
+include_once($relPath."unformat.inc"); // remove_formatting(), unspace()
+
+require_login();
+
+$projectid = validate_projectID('project', @$_GET['project']);
+
+$project = new Project($projectid );
```
cpeel: remove space before closing paren.

tools/project_manager/page_compare.php
```diff
@@ -0,0 +1,182 @@
...
+    } elseif ($project->state == PROJ_DELETE) {
+        echo _("The project has been deleted, so page details are not available.");
+    } else {
+        echo _("Page details are not available for this project.");
+    }
+    echo "</p>";
+    exit();
+}
+
+$L_round_id = array_get($_GET, "L_round_id", "P3");
```
cpeel: Move all $_GET accesses to the top of the file rather than have them scattered throughout. This allows one place to do user-input validation.

tools/project_manager/page_compare.php
```diff
@@ -0,0 +1,182 @@
...
+
+// if this is first entry don't do anything else
+if(!isset($_GET['compare']))
+    exit();
+
+$state_index = array_flip($PROJECT_STATES_IN_ORDER);
+
+if($state_index[$project->state] < $state_index["{$R_round_id}.proj_avail"])
+{
+    echo "$R_round_id has not started";
```
cpeel: This needs to use `_()` (and sprintf()).

tools/project_manager/page_compare.php
```diff
@@ -0,0 +1,182 @@
...
+    default: // all
+        $condition = "1";
+        break;
+}
+
+if(!$right_complete)
+    $condition .= " AND state='$R_round->page_save_state'";
+
+$query = "
+SELECT image, $L_text_column_name, $R_text_column_name
```
cpeel: Please indent these 4 spaces so it stands out as SQL. Also, move the closing quote to a line by itself. See Project:get_hold_status() in `pinc/Project.inc` for an example.

tools/project_manager/page_compare.php
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    function render($projectid, $L_round_id, $R_round_id)
+    {
+        global $Round_for_round_id_;
+
+        echo "<form action='page_compare.php' method='GET'>
+            <input type='hidden' name='project' value='$projectid'>\n
+            <input type='hidden' name='compare'>\n";
+        echo "<p>Compare ";
```
cpeel: This can' be translated as-is.

tools/project_manager/page_compare.php
```diff
@@ -0,0 +1,182 @@
...
+        $this->echo_round_select($L_round_id, "L_round_id", array_keys($Round_for_round_id_));
+        echo " to ";
+        $this->echo_round_select($R_round_id, "R_round_id", array('F1', 'F2'));
+        $this->echo_page_select('all', _("All pages"));
+        $this->echo_page_select('left', _("My first"));
+        $this->echo_page_select('right', _("My second"));
+        echo "<input type='submit' value='Go'></p></form>\n";
```
cpeel: "Go" is not translateable.

pinc/unformat.inc
jmdyck: `RoundDescriptor.inc` is generic, but the P/F distinction is pgdp-specific. Putting the code in `stages.inc` would be more appropriate, because that's completely pgdp-specific.

70ray:
Thanks for the comments. I've made some changes as suggested and a few other changes. I made a separate commit, re-ordered to beginning, for the functions to find if it is a proofing or formatting round since it is self-contained and rebased everything else. Not sure if I should edit the original comment to reflect current reality.

70ray:
I seem to have closed this request? Re-opening it with new comment:
Option to remove formatting before diff:
New function unformat() in DPage.inc to remove formatting and either completely unwrap or keep line structure but remove leading and trailing spaces.
New function has_entered_formatting_round() in Project class in project.inc.
Diff.php: add option to remove formatting (not unwrap)  before compare. If at least one round is F then show link option to remove  formatting or not. Default is remove for P to F, keep for F to F.
New file page_compare.php to compare pages with format removed and unwrapped between selected rounds for a project, either all pages or for the users pages in either the first or second round. Makes a list of page image names which link to diff.php.
Project.php: link added to page_compare.php if project has reached formatting stage.

tools/project_manager/page_compare.php
```diff
@@ -0,0 +1,203 @@
...
+echo "<p>";
+foreach($diff_pages as $imagename)
+{
+    echo "<a href='$code_url/tools/project_manager/diff.php?project=$projectid&amp;image=$imagename&amp;L_round_num=$L_round_num&amp;R_round_num=$R_round_num&amp;format=remove' target='_blank'>$imagename</a>\n";
+}
+echo "</p>";
+
+function not_started($round_id)
+{
+    global $state_index, $project;
```
cpeel: It would be better to pass these into the function instead of making them global.

tools/project_manager/page_compare.php
```diff
@@ -0,0 +1,203 @@
...
+// Draw the results
+echo "<p>", _("Clicking on a link will show the differences in a new window or tab."), "</p>\n";
+echo "<p>";
+foreach($diff_pages as $imagename)
+{
+    echo "<a href='$code_url/tools/project_manager/diff.php?project=$projectid&amp;image=$imagename&amp;L_round_num=$L_round_num&amp;R_round_num=$R_round_num&amp;format=remove' target='_blank'>$imagename</a>\n";
+}
+echo "</p>";
+
+function not_started($round_id)
```
cpeel: This function could use a better name -- particularly since it's doing two things (returning true/false if it hasn't started and also printing something out if it hasn't). Also better to make it a positive-named function instead of a negative. Perhaps `has_project_started_round()`.

tools/project_manager/page_compare.php
```diff
@@ -0,0 +1,203 @@
...
+            $sel_str .= ">$round</option>";
+        }
+        $sel_str .= "</select>\n";
+        return $sel_str;
+    }
+
+    function radio_string($value, $label)
+    {
+        $checked = ($this->page_set === $value) ? " checked" : "";
+        return "<input type='radio' name='page_set' value=$value$checked>$label\n";
```
cpeel: $value needs single-quotes around it and $label needs `html_safe()`

pinc/Project.inc
```diff
@@ -566,6 +566,23 @@ XML;
...
+
+    // check if project has entered a formatting round
+    public function has_entered_formatting_round()
+    {
+        global $PROJECT_STATES_IN_ORDER;
+        $state_index = array_search($this->state, $PROJECT_STATES_IN_ORDER);
+        for($i=$state_index; $i>=0; $i--)
+        {
+            $round = get_Round_for_project_state($PROJECT_STATES_IN_ORDER[$i]);
+            if($round)
```
cpeel: You could simplify this somewhat by:
```
if($round && is_formatting_round($round))
```
Totally optional.

cpeel: Although you need to include_once `stages.inc` at the top of this file if you're using a function from it.

tools/project_manager/page_compare.php
```diff
@@ -0,0 +1,203 @@
...
+
+$project = new Project($projectid);
+
+$state = $project->state;
+$title = $project->nameofwork;
+$sub_title = _('Compare pages with formatting removed');
+
+output_header("$sub_title: $title", NO_STATSBAR);
+
+echo "<h1>$title</h1>\n";
```
cpeel: I know other pages put the name of the work in the `<h1>` but it's really not helpful. This page isn't focused on the project, it's focused on comparing pages. I would swap these and put "Compare pages with formatting removed" as the `<h1>` and the nameofwork as the `<h2>`.

tools/project_manager/page_compare.php
```diff
@@ -0,0 +1,203 @@
...
+echo "<h2>$sub_title</h2>\n";
+
+$project_url = "$code_url/project.php?id=$projectid&amp;expected_state=$state";
+$label = _("Return to Project Page");
```
cpeel: Remove the newline between this and the echo below, which is the only thing in this file that uses this label, and add one above this line.

tools/project_manager/page_compare.php
```diff
@@ -0,0 +1,203 @@
...
+
+echo "<p><a href='$project_url'>$label</a></p>\n";
+
+if (!$project->pages_table_exists)
+{
+    echo "<p>";
```
cpeel: Should these be styled with `class='warning'` or `error`? I could go either way on that.

tools/project_manager/page_compare.php
```diff
@@ -0,0 +1,203 @@
...
+    exit();
+}
+else
+{
+    echo "<p>", sprintf(_("Comparing %d pages"), $num_rows), "</p>\n";
+}
+
+// make an array of imagenames of pages with diffs
+$diff_pages = array();
+for($rownum=0; $rownum < $num_rows; $rownum++)
```
cpeel: Since you don't use $rownum anywhere, this would be better as:
```php
while($page_res = mysqli_fetch_assoc($res))
```

tools/project_manager/page_compare.php
70ray: Done this, I did wonder originally whether to do that but since they are declared at the top level they are not any more global than they would be anyway. I suppose the best way to do it would be to make them members of a class but that would involve changing some of the original code rather than just adding stuff.

tools/project_manager/page_compare.php
70ray: I copied this block of code from "tools/project_manager/page_detail.php" so should we change it there also or perhaps better to make it a function somewhere?

70ray:
Thanks Casey, I've made the changes. Sorry about rebasing. If you like I'll do something about the warning for pages_table_exists in page_detail.php.

tools/project_manager/page_compare.php
70ray: Oops. It is my new page! I'll look at improving this.

70ray:
I just realised I can change page_compare without worrying about breaking any existing code so I'll look at improving the parameters.

70ray:
I've re-structured page_compare to use a class. Unfortunately this will probably make re-reviewing difficult but I think it is an improvement. Also improved error handling for project errors which could output messages before the header and give uncaught exceptions.

tools/project_manager/page_compare.php
```diff
@@ -0,0 +1,213 @@
...
+        global $PROJECT_STATES_IN_ORDER;
+
+        $this->L_round_options = array('P1', 'P2', 'P3', 'F1');
+        $this->R_round_options = array('F1', 'F2');
+        $this->state_index = array_flip($PROJECT_STATES_IN_ORDER);
+    }
+
+    function get_data()
+    {
+        $this->projectid = array_get($_GET, 'project', null);
```
cpeel: I prefer you still use `validate_projectID('project', @$_GET['project']);` so it matches the pattern elsewhere in the code. Yes this has the problem that we output an error message before we start the page, but I'd prefer we be consistent as much as possible. In the future perhaps we'll change how we do this, but this should never happen unless someone is mucking with URLs in which case it's their fault they are getting invalid HTML returned from the server.

tools/project_manager/page_compare.php
```diff
@@ -0,0 +1,213 @@
...
+        else
+        {
+            $error_message = _("No project is specified");
+        }
+
+        if($this->projectid)
+        {
+            try
+            {
+                $this->project = new Project($this->projectid);
```
cpeel: You don't need the try/catch. On the TEST server when we throw exceptions we print the full exception to help debugging. On PROD we catch the exception in a global exception handler and just print the error. See the top of `pinc/base.inc` for the code that does this.

project.php
```diff
@@ -694,7 +694,13 @@ function do_project_info_table()
                 $blurb = _("Images, Pages Proofread, & Differences");
                 $url2 = "$url&amp;select_by_user";
                 $blurb2 = _("Just my pages");
-                $detail = "<a href='$url'>$blurb</a> &gt;&gt;<a href='$url2'>$blurb2</a>&lt;&lt;";
+                $detail = "<a href='$url'>$blurb</a> | <a href='$url2'>$blurb2</a>";
```
cpeel: We may want to rethink how we're outputting these strings. $blurb1 and $blurb2 point to the same page with different parameters. Presumably we used `>> <<` to draw attention to what most(?) people will want to use -- which is just their pages. Removing that and placing them all in `|`s makes them appear as though they are all equivalent. Perhaps something like: `$blurb1 &middot; <b>$blurb2</b> | $blurb3`? That somewhat mirrors what we do in the navbar.

70ray:
Reverted the last commit which changes error handling.
Modified the links in project.php.
Revised the check_pages_table_exists() function and re-used it also in project.php

tools/project_manager/page_compare.php
```diff
@@ -0,0 +1,190 @@
...
+            echo "<p class='warning'>$warn_message</p>\n";
+            exit();
+        }
+
+        // draw the round selectors
+        echo "<form action='page_compare.php' method='GET'>
+            <input type='hidden' name='project' value='$this->projectid'>\n
+            <input type='hidden' name='compare'>\n";
+        // TRANSLATORS: %1$s and %2$s are selectors e.g. P3, %3$s etc. are radio button options
+        echo '<p>', sprintf(_('Compare %1$s to %2$s for %3$s %4$s %5$s'),
```
cpeel: I suspect this is going to be challenging to translate well. Moreover "My first" and "My second" are really confusing labels since they really mean "use all of the pages that I proofread in the [first|second] round" (and I had to read the code to figure that out). This would probably be better as a small form with labels rather than a sentence. Perhaps something like:
```
Compare rounds:
Round 1   [P3]
Round 2   [F1]
Show      o All pages
          o Pages I worked on in Round 1
          o Pages I worked on in Round 2
```

70ray: It would probably be possible to add some javascript to change the text but seems like adding more complexity.
So it would be like
Compare rounds:   [P3]   [F1]
Show      o All pages
          o Pages I worked on in P3
          o Pages I worked on in F1

cpeel: I don't think you need to do any fancy javascript, I just think you need a form/table with labels you can use to refer to things. Much like we do with the review_work.php page (ie: "work round" and "compare round").

70ray:
Made changes as suggested.

styles/layout.less
```diff
@@ -881,5 +881,15 @@ table.search-column {
     }
 }
 
+/* ------------------------------------------------------------------------ */
+/* Page compare grid */
+
+.grid-wrapper {
+  display: grid;
```
cpeel: Indents in this file are 4 spaces, not two.

70ray:
Indent in layout.less changed and some rebasing for minor changes and reverted changes.

cpeel:
This looks great Ray, thank you for your patience as we worked through my comments (and learned more about github PRs along the way)!

70ray:
Thanks for all your help Casey.

PR #4, opened 2018-03-09
Allow large file uploads

cpeel:
We are limited in how large files can be for uploading in PHP by the size of the post request and maximum memory. On pgdp.net this restricts us to 48MB files requiring PMs to split larger .zips into smaller pieces and upload them.

Now that we are supporting HTML 5 we can use 3rd party tools like resumable.js that allow uploading the file in pieces and combining them on the server.

This pulls in jQuery, in addition to resumable.js, as a JS foundation library we can use to help us write better javascript code. I've created a `pinc/3rdparty` directory to house these 3rd party scripts we're pulling into our codebase and moved the mediawiki code into that as well.

A demo of this is available at https://www.pgdp.org/~cpeel/c.large_uploads/tools/project_manager/remote_file_manager.php


70ray:
I tested this. When choosing a file I get an error message "2016-05-04-Llanberis.zip has type not allowed, please upload files of type application/zip". It is a compressed (in windows 10) folder with pictures.
I looked at the code but found it hard to figure out what the "Choose file" button does so could not investigate.

70ray:
Looking at the ctrl-U source listing there is a stray </p> because of  a nested <p class='warning'>Maximum file size is 25.00 MB.</p>. This doesn't seem to be in the github file. Perhaps I am looking at the wrong thing.

cpeel:
Turns out the version in `~cpeel/c.large_uploads` had changes that weren't in my development sandbox. That's what I get for having code in two places. I've updated the code in the branch to not do the nested `<p>`s although I don't think that is related to your problem.

cpeel:
I'm hoping this last commit solves your problem, @70ray:
```
-        fileType: ['application/zip'],
+        fileType: ['zip'], // use extension not mime type since zips have many
```
Should be live at https://www.pgdp.org/~cpeel/c.large_uploads/tools/project_manager/remote_file_manager.php

70ray:
It's working now. Is it possible to cancel the upload once it has started? Does it stop if you navigate away from the page?

cpeel:
> Is it possible to cancel the upload once it has started? Does it stop if you navigate away from the page?

You can navigate away from the page and it will stop uploading. If upload isn't finished and you attempt to upload the same file again (in that session, another browser session, or even another browser) it will pick up where it left off -- or at least that's what the resumable.js docs say. If the file is never finished being uploaded, the partial upload will be cleaned out on the server eventually (they live in `/tmp`).

70ray:
Is it possible to add some text like this to the page?

cpeel:
I've clarified the verbiage above the file upload input fields. I don't think most users need all the gory details but knowing they can navigate away from the page to cancel the upload is useful.

srjfoo:
I successfully started uploading a 436.13 MB zip, navigated away from the page, started the upload again, and successfully completed it. It also correctly detects and refuses an attempt to upload a zip > 1GB. Now for another question. I went ahead and tested two zips I knew should have illegal filenames. The uploads were done before the zips were rejected, whereas if you try to upload a zip that's too large, it fails immediately. Is it possible to detect and reject illegal filenames before beginning the upload?

cpeel:
> Is it possible to detect and reject illegal filenames before beginning the upload?

Yes, but not easily. We get the file-size (and file-type) check "free" with the javascript library we are using. The filename validation occurs in PHP code and is shared between both the old and new upload method. If we wanted to have the filename checks in javascript prior to upload we'd need to have two copies of them -- one for each method -- which I'd prefer not to do.

srjfoo:
That's reasonable, thanks; it just seemed to me that it was a waste of
bandwidth to upload prior to checking the filename for validity.

PR #7, opened 2018-03-24
fix for problem with % in code_name

70ray:
variables used for escaped strings instead of using sprintf

### DIFF
--- a/tools/project_manager/manage_image_sources.php
+++ b/tools/project_manager/manage_image_sources.php
@@ -448,17 +448,16 @@ class ImageSource
             die;
         }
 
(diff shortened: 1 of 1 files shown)


jmdyck:
The problem isn't having a '%' in the value of code_name, it's having a '%' in the value of $std_fields_sql.

More generally, the problem is having a call to sprintf() where the format string contains a variable-interpolation. This can be done safely (either by constraining or escaping the interpolated value), but we're probably better off avoiding it.

Ray's PR eliminates the call to sprintf, which looks like a reasonable fix to me.

PR #8, opened 2018-03-24
Ubuntu 16.04 fixes

cpeel:
This fixes a variety of small issues required to work with Ubuntu 16.04. Surprisingly, very very few of these are for PHP 7.x -- the bulk of the work is to support newer versions of MySQL that are less forgiving with lax SQL.

### DIFF
--- a/SETUP/db_schema.sql
+++ b/SETUP/db_schema.sql
@@ -248,8 +248,8 @@ CREATE TABLE `past_tallies` (
 #
 
 CREATE TABLE `pg_books` (
-  `etext_number` smallint(5) unsigned NOT NULL default '0',
-  `formats` tinytext NOT NULL,
+  `etext_number` int(10) unsigned NOT NULL,
+  `formats` varchar(255) NOT NULL DEFAULT '',
   PRIMARY KEY  (`etext_number`)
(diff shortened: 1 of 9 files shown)


cpeel:
I did it this way so that we play nice with other database users (MediaWiki and phpBB in our case, could be other users in other cases). Setting stuff in my.conf applies those settings to everything that uses it, which might not want sql_mode changes.

wf49670:
Hadn't thought about other users of the DB server, Casey. Good reason. Might be nice to mention that in the code :)

cpeel:
Good idea. I'll document that and why the margin changes (which were to accommodate the slightly different legend layout changes that came in with jpgraph 4.x).

70ray:
Can't see any problems.
If the code should work with PHP 7.2, there are same 'each' statements around which are deprecated. Some in tasks.php.

cpeel:
>If the code should work with PHP 7.2, there are same 'each' statements around which are deprecated. Some in tasks.php.

The 7.2 assert fixing came out of investigating the prior commit. There's no hard requirement that this works with 7.2 ... at least not yet. I'll add those to my to-fix list.

PR #9, opened 2018-03-27
Move manager and sitemanager flags into usersettings

cpeel:
This consolidates the manager and sitemanager flags from the `users` table to the `usersettings` table. This simplifies several different parts of the code.

This also logs when access to these states is granted and revoked and in doing so centralizes that logic into the User class.


cpeel:
Note that there are noncvs script changes that need to come along with this. Those changes are done, just need to be rolled out.

SETUP/upgrade/12/20180216_alter_access_log.php
```diff
@@ -0,0 +1,23 @@
...
+include_once($relPath.'base.inc');
+
+header('Content-type: text/plain');
+
+// ------------------------------------------------------------
+
+echo "Increasing size of access_log.activity column..\n";
+$sql = "
+    ALTER TABLE access_log
+        CHANGE COLUMN activity activity VARCHAR(32) NOT NULL DEFAULT '';
```
jmdyck: If you replace "CHANGE" with "MODIFY", you only need to give the column name once.

SETUP/db_schema.sql
```diff
@@ -11,7 +11,7 @@ CREATE TABLE `access_log` (
   `subject_username` varchar(25) NOT NULL default '',
   `modifier_username` varchar(25) NOT NULL default '',
   `action` varchar(16) NOT NULL default '',
-  `activity` varchar(10) NOT NULL default '',
+  `activity` varchar(32) NOT NULL default '',
```
jmdyck: The commit message should say why the column needs to be widened.

tools/site_admin/manage_site_access_privileges.php
```diff
@@ -105,20 +106,29 @@ function show_toggles_form($username, $user_settings)
...
     {
         $user_current_value = $user_settings->get_boolean($setting_name);
-        $checked_attr = ($user_current_value == 'yes' ? 'checked' : '');
+        $checked_attr = ($user_current_value ? 'checked' : '');
+
+        $access_log_entry = get_latest_access_change_entry($username, $setting_name);
+        if($access_log_entry)
+            $changed = strftime('%Y-%m-%d %H:%M', $access_log_entry["timestamp"]);
+        else
+            $changed = "<i>" . _("unknown") . "</i>";
```
wf49670: Looking at:
https://www.pgdp.org/~cpeel/c/tools/site_admin/manage_site_access_privileges.php?username=wfarrell 
I'm not sure I like that column of italic "unknown". Could we consider leaving it blank unless an access log entry was found?

SETUP/db_schema.sql
cpeel: <nod> I'll do so after at the end of this CR since that will rebase all of the commits.

tools/site_admin/manage_site_access_privileges.php
cpeel: Sure. I used _unknown_ to be explicit that we really don't know when that was set rather than leave it blank and make the user guess, but I leave that decision to you, @srjfoo, and @lhamilton1.

SETUP/upgrade/12/20180216_alter_access_log.php
cpeel: Neat -- why haven't I noticed that before?

tools/site_admin/manage_site_access_privileges.php
srjfoo: I would prefer to have something there, rather than having it be blank, and <i>unknown</i> makes as much sense as anything to me. It differentiates it from the section of the script that doesn't have a date component associated with it.

One other note -- since enabling a sitemanager is not included in the script (for valid reasons), the sawiki instructions should be updated once the code goes live.

tools/modify_access.php
```diff
@@ -145,12 +142,11 @@
...
-    $yesno = ( $action_type == 'grant' ? 'yes' : 'no' );
 
-    $userSettings =& Settings::get_Settings($subject_username);
-    $userSettings->set_value("$activity_id.access", $yesno);
-
-    log_access_change( $subject_username, $pguser, $activity_id, $action_type );
+    if($action_type == 'grant')
+        $user->grant_access("$activity_id.access", $pguser);
+    else
+        $user->revoke_access("$activity_id.access", $pguser);
```
jmdyck: {grant/revoke/request}_access should be invoked with $activity_id, not "$activity_id.access".

$activity_id is for the access_log table; "$activity_id.access" is for the usersettings table.

pinc/access_log.inc
```diff
@@ -38,3 +38,23 @@ function get_first_granted_date($username, $stage)
 
     return $row["timestamp"];
 }
+
```
jmdyck: [Commenting here because github won't let me comment directly on the commit message.]

> Make an access_log record when a user privilege is granted or revoked.

That already happens. You need to add something like "via manage_site_access_privileges.php".

> This slightly abuses the access_log record, which in the past has only taken a stage name for an activity.

It's not an abuse, the access_log table was designed from the start to handle more than just stage-access. That's why the column is labelled 'activity', not 'stage'. Moreover, this isn't the first non-stage use: the table has entries with activity='DU' and activity='P2_mentor', neither of which is a stage id.

> Now it also takes other access privileges that are more verbose.

(Change "takes" to "records", I think.) So that's why you need to widen the 'activity' column. I think it would make sense to squash the two commits.

tools/modify_access.php
cpeel: <sigh> You're right, it is for these but not for the things in `manage_site_access_privileges.php` because They're Different. I'll have to noodle this over.

cpeel: Ok, I've fixed this in 4aefbc1.

pinc/access_log.inc
cpeel: I'll squash this and the database update together and update the commit message before merging.

cpeel:
I've addressed all open issues (I think) and have rebased the branch and fixed the commit messages. I think this is ready to go.

PR #10, opened 2018-03-28
Allow call-out to local customization script

cpeel:
The pgdp.net production server had customized the `update_from_sf` script to do local filesystem changes (copy over files, apply patches). This abstracts that functionality such that customization is still possible without changing code in source control that we have to then manually patch.

### DIFF
--- a/SETUP/update_from_github
+++ b/SETUP/update_from_github
@@ -8,7 +8,8 @@ fi
 this_script_file=$0
 site_config_file=$1
 
-script_dir=`dirname $this_script_file`
+export script_dir=`dirname $this_script_file`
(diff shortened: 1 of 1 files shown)


jmdyck:
For comparison, I want to look at the way we used to do it. Does that still exist?

cpeel:
> For comparison, I want to look at the way we used to do it. Does that still exist?

Yes, on prod8 compare:
* before: `/home/dpadmin/moved_SETUP/update_from_sf`
* after: this change + `/home/dpadmin/bin/customize_deployment`

jmdyck:
Thanks. Looks okay to me.

PR #11, opened 2018-03-28
Don't use sprintf() as $std_fields_sql can contain %

cpeel:
Your eyes do not deceive you, this isn't the same thing as https://github.com/DistributedProofreaders/dproofreaders/pull/7 but it sure does look like it! That fix was for `manage_image_sources.php` and this one is for `manage_special_days.php`.


70ray:
Looks alright to me.

PR #12, opened 2018-04-03
'each' (deprecated in PHP7.2) replaced in 3 files.

70ray:
Also adjustment to table in authors/manage.php to make it clearer which multiple biographies belong to each author.


cpeel:
This looks great - thanks Ray!

PR #13, opened 2018-04-04
Add a check for good/bad word lists to project quick check

cpeel:
This add a simple automatic test to project quick check to see if there is at least one word on both the good and bad word lists for a project. If there isn't, it returns a Warning as this might be intentional.


wf49670:
Looks good to me, Casey.

PR #14, opened 2018-04-05
Update PHP and phpBB version support

cpeel:
After ea40ee5d merged we know the code works with PHP 7.0.

When we merged ba74ae04, we broke support for phpBB 3.0. This
is because accounts/login.php pulls in metarefresh.inc which
pulls in slim_header.inc. ba74ae04 changed slim_header.inc to
bring in theme.inc which eventually brings in user.inc which
breaks phpBB 3.0 because it also defines a User class. phpBB 3.0
is out of support and phpBB 3.1 was released in Oct 2014 so
rather than trying to hack a fix in, lets just say 3.1 is the
minimum supported version.

### DIFF
--- a/SETUP/installation.txt
+++ b/SETUP/installation.txt
@@ -20,7 +20,8 @@ PHP
     PHP version 5.3.2 is the minimum version supported, although any
     version >= 5.3.2 should work.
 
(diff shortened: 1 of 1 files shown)


jmdyck:
Looks okay to me.

PR #15, opened 2018-04-11
Translation fixes

cpeel:
This incorporates lots of excellent feedback from @srjfoo during her work shepherding the French translation through to production. This makes strings like "none" and "all" easier to translate and consolidates some strings.


srjfoo:
The changes look good to me.

PR #16, opened 2018-04-14
Remove writeBIGtable support

cpeel:
From the commit message:

    writeBIGtable was added in the far distant past with the intent
    of migrating to one large project_pages table to hold all projects
    rather than each project in its own table. This functionality was
    never fully realized and writeBIGtable never enabled on pgdp.net.
    Moreover many spots in the code make updates to individual project
    tables without touching project_pages making this feature's
    accuracy and utility highly suspect.


SETUP/CHANGELOG.md
```diff
@@ -3,20 +3,21 @@
...
-
 
 ## R??????
 Scripts supporting this upgrade are in `SETUP/upgrade/12`
 
 * Support Ubuntu 16.04 middleware (cpeel)
 * Support jpgraph version 4.2.0 (cpeel)
 * Move manager and sitemanager columns to usersettings (cpeel)
+* `$writeBIGtable` support, including `project_pages` table, has been
+  removed (cpeel)
```
jmdyck: The 2 lines above belong to this PR, but not the 6 below.

jmdyck:
There's also 2 lines to be deleted from SETUP/check_db_schema.template.

jmdyck:
Other than those, looks good to me.

SETUP/CHANGELOG.md
cpeel: Yup, but this got you to look at them, didn't it :)

cpeel: Note that they are in a separate commit.

PR #19, opened 2018-04-28
Improve project creation error messages

cpeel:
During project creation, if no record is selected after the
external catalog search and 'Create Project' is selected, we throw
a very useless error to the user. Instead, lets provide information
they can actually use.

### DIFF
--- a/tools/project_manager/editproject.php
+++ b/tools/project_manager/editproject.php
@@ -140,7 +140,7 @@ else
 
     if ($fatal_error != '')
     {
-        $fatal_error = _('site error') . ': ' . $fatal_error;
+        $fatal_error = _('Error') . ': ' . $fatal_error;
         echo "<p class='error'>$fatal_error</p>";
         exit;
     }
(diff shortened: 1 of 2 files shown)


wf49670:
Looks good to me, Casey.

tools/project_manager/editproject.php
```diff
@@ -184,31 +184,20 @@ function set_from_marc_record()
     {
         global $pguser;
 
-        if (!isset($_POST['rec']))
+        $encoded_marc_array = array_get($_POST, "rec", "");
+        if(!$encoded_marc_array)
         {
-            return sprintf(_("parameter '%s' is unset"), 'rec');
+            return sprintf(_("No record selected. If no results are suitable, select '%s' to create the project manually."), "No Matches");
```
jmdyck: "No Matches" should part of the translated string (because the referenced button's text is itself translated).

cpeel: I did it this way on purpose so that the "No Match" string pulls from the same translation to prevent the two strings from accidentally being translated differently, except that I forgot to wrap the string in `_()` -- I'll fix that.

jmdyck: Okay. In that case, external_catalog_search.php (at about line 141) has another string referring to that button that you might want to treat similarly.

PR #20, opened 2018-06-03
Update proofreading interface help

cpeel:
These are updates to the proofreading interface help to reflect
the changes made in the DPOD wiki. Updates courtesy of lhamilton.

### DIFF
--- a/faq/prooffacehelp.php
+++ b/faq/prooffacehelp.php
@@ -19,21 +19,17 @@ if ( $i_type == 0 )
 {
     output_header('Standard Proofreading Interface Help', NO_STATSBAR);
     echo "
-    <center>
-    <h2>Standard Proofreading Interface Help</h2>
-    <br>
+    <h1>Standard Proofreading Interface Help</h1>
     <table>
     <tr><td>Updated</td><td>05/24/2018</td><td>lhamilton</td></tr>
     </table>
-    </center>
     ";
 }
 else
 {
     output_header('Enhanced Proofreading Interface Help', NO_STATSBAR);
     echo "
-    <center>
-    <h2>Enhanced Proofreading Interface Help</h2>
+    <h1>Enhanced Proofreading Interface Help</h1>
     Version 1.3<br>
     <br>
     <table>
@@ -44,7 +40,6 @@ else
     <tr><td>Updated</td><td>05/24/2005</td><td>pourlean</td></tr>
     <tr><td>Updated</td><td>05/24/2018</td><td>lhamilton</td></tr>
     </table>
-    </center>
     ";
 }
 
@@ -52,7 +47,7 @@ else
 
 ?>
 
-<h3>Introduction</h3>
+<h2>Introduction</h2>
 
 <p>
 To understand the interface,
@@ -81,7 +76,6 @@ They are:
     or else this partially-proofread page may move to the next round
     when the rest of the project does!
     </li>
-    <br>
 
     <li>
     <b><a name=progr>"IN PROGRESS"</a></b>:
@@ -95,7 +89,6 @@ They are:
     If you do not complete them,
     eventually the system will reclaim them for someone else to proofread.
     </li>
-    <br>
 
     <li>
     <b><a name=return>"RETURN TO ROUND"</a></b>
@@ -107,7 +100,6 @@ They are:
     Whoever next clicks on the "Start Proofreading" link for this project
     will get the page to proofread.
     </li>
-    <br>
 
     <li>
     <b><a name=bad>"BAD"</a></b>
@@ -557,7 +549,7 @@ else
 {
 ?>
 
-<CENTER><DIV ALIGN="CENTER"><TABLE
+<DIV ALIGN="CENTER"><TABLE
 BORDER="1" WIDTH="630" CELLPADDING="6">
 
 <TR><TD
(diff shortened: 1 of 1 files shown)


lhamilton1:
Approved

faq/prooffacehelp.php
```diff
@@ -21,6 +21,10 @@
     echo "
     <center>
     <h2>Standard Proofreading Interface Help</h2>
+    <br>
+    <table>
+    <tr><td>Updated</td><td>05/24/2018</td><td>lhamilton</td></tr>
```
wf49670: Suggestion: Use ISO standard date format, 2018-05-24

(Though, I see we use US format below, so maybe consistency is best.)

cpeel: Yes, I'd prefer ISO-8601 as well but I think we should be consistent with the other dates in this file.

faq/prooffacehelp.php
```diff
@@ -596,12 +637,13 @@ function echo_row( $name, $tooltip, $button_image_base, $accelerator )
 <P><A NAME="ikeys"> </A> <P><FONT SIZE="+1">Accelerator Keys (accesskeys)</FONT><P></TD></TR>
 <TR>
 <TD VALIGN="TOP" COLSPAN="2">
-Several of the commonly used interface buttons have been assigned an accesskey value.
+<p>These accelerator keys may not work for everyone.</p>
```
wf49670: I would say: ... everyone, and the modifier key (shown as Alt below) may be different depending on your Operating System and browser.

faq/prooffacehelp.php
```diff
@@ -697,6 +739,16 @@ function echo_row( $name, $tooltip, $button_image_base, $accelerator )
 <h4>Common Tags</h4>
 
 <p>
+You may sometimes find formatting already present in the text.
+Do not add or correct this formatting information; the formatters
+will do that later in the process. However, you can remove it if
+it interferes with your proofreading. The &lt;x&gt; button in the
```
wf49670: It's not &lt;x&gt;
Rather, it is &lt;x&gt; but struckthrough, and should be shown that way.

faq/prooffacehelp.php
```diff
@@ -752,20 +804,23 @@ function echo_row( $name, $tooltip, $button_image_base, $accelerator )
 </p>
 
 <h5>Shortcuts to insert common tags</h5>
+<p>
+    These shortcuts may not work for everyone.
```
wf49670: ... for everyone, and the Alt may be a different key combination depending on your Operating System and browser.

faq/prooffacehelp.php
```diff
@@ -752,20 +804,23 @@ function echo_row( $name, $tooltip, $button_image_base, $accelerator )
...
 <table border='1'>
   <tr><td>
-    Bold
+    Bold (Proofreaders do not add bolding codes)
   </td><td>
     Alt-b
   </td></tr>
   <tr><td>
-    Italics
+    Italics (Proofreaders do not add bolding codes)
```
wf49670: bolding -> italic

faq/prooffacehelp.php
```diff
@@ -752,20 +804,23 @@ function echo_row( $name, $tooltip, $button_image_base, $accelerator )
...
   </td><td>
     Alt-i
   </td></tr>
   <tr><td>
-    Small caps
+    Small caps (Proofreaders do not add bolding codes)
```
wf49670: bolding -> Small caps

faq/prooffacehelp.php
cpeel: Please coordinate with @lhamilton1 on the wording of this, whatever you both decide is fine with me and needs to be updated on the wiki as well.

faq/prooffacehelp.php
cpeel: Ditto, please coordinate with @lhamilton1.

faq/prooffacehelp.php
cpeel: Fixed, thanks.

faq/prooffacehelp.php
cpeel: Fixed, thanks.

faq/prooffacehelp.php
cpeel: Good catch, fixed.

lhamilton1:
I'm OK with the changes, Walt suggest. Once it's in Test I'll take a close look to double-check but liked the added phrasing.

    On Monday, June 4, 2018, 11:30:42 AM EDT, Casey Peel <notifications@github.com> wrote:  
 
 
@cpeel commented on this pull request.

In faq/prooffacehelp.php:
> @@ -596,12 +637,13 @@ function echo_row( $name, $tooltip, $button_image_base, $accelerator )
 <P><A NAME="ikeys"> </A> <P><FONT SIZE="+1">Accelerator Keys (accesskeys)</FONT><P></TD></TR>
 <TR>
 <TD VALIGN="TOP" COLSPAN="2">
-Several of the commonly used interface buttons have been assigned an accesskey value.
+<p>These accelerator keys may not work for everyone.</p>

Please coordinate with @lhamilton1 on the wording of this, whatever you both decide is fine with me and needs to be updated on the wiki as well.

faq/prooffacehelp.php
cpeel: Updated.

faq/prooffacehelp.php
cpeel: Updated.

faq/prooffacehelp.php
```diff
@@ -569,6 +604,7 @@ function echo_row( $name, $tooltip, $button_image_base, $accelerator )
 echo_row( 'Change Interface Layout', 'Change Interface Layout', 'bt4+bt5', '6' );
 // echo_row( 'Check for Common Errors', 'Check for Common Errors', 'bt6', '' );
 echo_row( 'Run WordCheck', 'Run WordCheck', 'bt16', '' );
+echo_row( 'Preview', 'Format preview', 'bt20', '' );
```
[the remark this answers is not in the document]
cpeel: No. The left one is what will be displayed on the Standard Interface button, and selects the text, and the right is the tooltip.

faq/prooffacehelp.php
```diff
@@ -569,6 +604,7 @@ function echo_row( $name, $tooltip, $button_image_base, $accelerator )
 echo_row( 'Change Interface Layout', 'Change Interface Layout', 'bt4+bt5', '6' );
```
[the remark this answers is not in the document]
cpeel: There was some invalid HTML (not escaping tooltips in HTML attributes) but that wasn't the problem. The problem is `lang_dir()` cruft that needs to be removed. I'll follow up with a separate PR for that.

PR #21, opened 2018-06-03
Allow denying user-requested access

cpeel:
In addition to grant/revoke, admins and evaluators have the ability to deny a request for access too. This fixes a regression from the merge of cleanup_users_table in commit a99d656f.

### DIFF
--- a/pinc/User.inc
+++ b/pinc/User.inc
@@ -148,6 +148,11 @@ class User
         $this->_change_access("request", $activity, 'n/a', $use_access_suffix);
     }
 
+    public function deny_access($activity, $requester, $use_access_suffix=TRUE)
+    {
+        $this->_change_access("deny_request_for", $activity, $requester, $use_access_suffix);
+    }
+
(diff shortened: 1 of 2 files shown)


srjfoo:
It looks good to me, but I don't trust my code-reviewing abilities well enough not to want someone else to look at it. However, when the code was available in Casey's sandbox, I did test it, and it worked as it was supposed to.

pinc/User.inc
```diff
@@ -158,6 +163,8 @@ class User
         $userSettings =& Settings::get_Settings($this->username);
         if($action_type == "request")
             $userSettings->set_value("$activity$suffix", "requested");
+        elseif($action_type == "deny_request_for")
```
wf49670: We also seem to be missing handling for "revoke" (see public function revoke_access on line 141 of this module).

cpeel: "revoke" is handled as part of the else statement below.

wf49670: Ah. Right. I see that now.

PR #24, opened 2018-06-05
Change 'Me Too!' link after it has been selected - Task 1679

cpeel:
After a user has 'Me Too'd a task, the button goes away. This can lead to confusion why it shows up for users on some pages but not others. It would be better to indicate that they have already "Me Too!'d the task.

While I was in here I did some massaging on the CSS to bring it better inline with HTML5, however it isn't fully compliant.

### DIFF
--- a/tasks.php
+++ b/tasks.php
@@ -936,6 +936,8 @@ select.taskselect  { font-size: small; color:#03008F; background-color:#EEF7FF; 
 input[type="text"] { font-size: small; border:1px solid #000000; margin:2px; padding:0px; background-color:#EEF7FF; }
 input[type="button"],
 input[type="submit"] { font-size: small; color:#FFFFFF; font-weight:bold; border:1px ridge #000000; margin:2px; padding: 0px 5px; background-color:#838AB5; }
+input[type="button"]:disabled,
+input[type="submit"]:disabled { background-color: #AAAAAA; }
 legend.task        { font-weight:bold; }
 fieldset.task      { width:35em; border:#2266AA solid 1px; }
 small.task         { font-family:Verdana; font-size: small; }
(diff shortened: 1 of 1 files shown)


tasks.php
```diff
@@ -1081,12 +1088,12 @@ function select_and_list_tasks($sql_condition)
...
     else {
-        echo "<tr bgcolor='#ffffff'><td colspan='7'><center>No tasks found!</center></td></tr>";
+        echo "<tr><td colspan='7'><center>No tasks found!</center></td></tr>";
     }
     echo "</table><br>\n";
     // if 2 tasks or more found, display the number of reported tasks
     if (@mysqli_num_rows($sql_result) > 1) {
-        echo "<p>" . @mysqli_num_rows($sql_result) . " tasks listed.</p>";
+        echo "<small class='task'>" . @mysqli_num_rows($sql_result) . " tasks listed.</small>";
     }
```
srjfoo: I'm probably missing something here -- but I'm seeing a switch from a block element to an inline element. Is this already dealt with elsewhere?

70ray:
Seems ok to me.

tasks.php
wf49670: I agree it's a change, Sharon. But the text string is inside a div element and that's all that should be needed, technically, I believe. The W3C HTML validator doesn't object to it, at least.

wf49670:
It looks OK to me, if we're not interested (at the moment) in "fixing" the handful of td elements that specify width= directly rather than using CSS or style=.

tasks.php
srjfoo: Thanks, Walt, that's what I was missing. Yes, a div element is plenty.

srjfoo:
And to me.

PR #26, opened 2018-06-06
remove_formatting function improved and put in a class

70ray:
The end of footnotes and other notes is now detected correctly
by passing over paired square brackets. Asterisks for continuation
notes are also removed. A new UnFormatter class is used to avoid
more global functions.
New test file page_compare_test.php added in new directory with
.htaccess file to allow it to be run


cpeel:
Rather than just add manual_web tests, create unit tests like `SETUP/tests/MARCRecordTest.php`.

Unit tests allow code in the PageUnformatter class to be validated as part of check-in. Move your .txt files into `SETUP/tests/data` and rename them so that they are more parallel (footnotes-formatted.txt, footnotes-unformatted.txt, poem-formatted.txt, etc). Create a new `SETUP/tests/PageUnformatterTest.php` file that contains a class to do unit testing. Each test is a class method and would load the *-formatted.txt file, unformat it, and compare it to the -unformatted.txt version. Then if you make logic changes in the PageUnformatter() we have tests to ensure they are correct. As we find more pages that we need to adjust the unformatter, we can add more tests.

See the `README.md` file in that directory for how to run the tests.

70ray:
New SETUP/tests/PageUnformatterTest.php file created and some tests added in new subdirectories so can use a dataProvider to run them. README file revised for manual testing.

cpeel:
Great use of dataProvider in the unit tests!

PR #27, opened 2018-06-10
Refactor pgdp.net FAQ redirects

cpeel:
What started out as a simple fix to remove the unused `lang_dir()` turned into a refactor on how we're doing redirects to pgdp.net FAQs. This centralizes the logic into `pinc/faq.inc` rather than being distributed into all of the FAQs themselves.

A very large bonus is that links to externalized docs now point to the final destination *for that language* instead of the version in the code and then redirected to the version in English.

Note that this renamed/moved every proofreading and formatting guideline file except for `faq/proofreading_guidelines.php`. This could result in some dead links. For pgdp.net I suggest we add some local fixes to redirect the users to the new location they should bookmark rather than having those redirects in the code.


faq/FAQ_Formatting_Guide.txt
```diff
@@ -249,7 +249,7 @@ Good for browsers and text editors, bad for CVS
...
 
-<p>Post-processors require more experience than ordinary proofreaders.  Because they are preparing the text for uploading to Project Gutenberg, they are the final editors of the text.  Because of this, post-processing is only available for proofreaders who have completed at least <b>400</b> pages in the first and/or second proofreading rounds.  Also, post-processors should be very familiar with the <a href="document.php">Proofreading Guidelines</a> before attempting to post-process.</p>
+<p>Post-processors require more experience than ordinary proofreaders.  Because they are preparing the text for uploading to Project Gutenberg, they are the final editors of the text.  Because of this, post-processing is only available for proofreaders who have completed at least <b>400</b> pages in the first and/or second proofreading rounds.  Also, post-processors should be very familiar with the <a href="proofreading_guidelines.php">Proofreading Guidelines</a> before attempting to post-process.</p>
 <br>
```
srjfoo: document.php morphed from the proofing and formatting guidelines before the Great Change into the formatting guidelines, didn't they? In fact, that whole para needs to be rewritten, because access to PP is restricted to those who've done at least 400 F1 pages.

faq/FAQ_Formatting_Guide.txt
```diff
@@ -399,7 +399,7 @@ they are the final editors of the text.
 Because of this, post-processing is only available for proofreaders who have completed at least
 <b>400</b> pages in the first and/or second proofreading rounds. 
 Also, post-processors should be very familiar with the
-<a href="document.php">Proofreading Guidelines</a> before attempting to post-process.
+<a href="proofreading_guidelines.php">Proofreading Guidelines</a> before attempting to post-process.
 </p>
```
srjfoo: This may need to be rewritten, too. document.php was mapped to formatting)guidelines.php, not proofreading_guidelines.php.

faq/FAQ_Formatting_Guide.txt
cpeel: Nothing in the source code references this file. We could consider deleting it actually. The last real change in this file was in 2003.

srjfoo: Is there a good reason for taking a good look at the file and doing a real update on it? If not, deleting it sounds like a reasonable idea.

cpeel: I don't see a good reason to update it, but I defer to @lhamilton1.

lhamilton1: I'm fine with deletion. Thanks.

srjfoo:
Ah, I didn't think about that. Thanks. What threw me was the outdated pieces of information (I was looking at the trees, not the forest).

PR #28, opened 2018-06-30
Restrict nameofwork and author inputs to 255 chars

cpeel:
The size of the db columns for nameofwork and author is 255. Prior versions of MySQL would silently truncate longer strings to the size of the column, but modern versions throw an error. Prevent users from inputting strings longer than 255 for these columns.

Without a doubt there are other instances where we should be doing this type of enforcement, but this addresses this instance.

### DIFF
--- a/tools/project_manager/edit_common.inc
+++ b/tools/project_manager/edit_common.inc
@@ -15,10 +15,13 @@ function just_echo( $field_value )
 
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 
-function text_field( $field_value, $field_name )
+function text_field( $field_value, $field_name, $args=array())
 {
+    $maxlength = array_get($args, "maxlength", NULL);
+    $maxlength_attr = $maxlength ? "maxlength='$maxlength'" : '';
+
     $enc_field_value = attr_safe($field_value);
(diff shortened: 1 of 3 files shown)


tools/project_manager/edit_common.inc
```diff
@@ -15,10 +15,13 @@ function just_echo( $field_value )
 
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 
-function text_field( $field_value, $field_name )
+function text_field( $field_value, $field_name, $args=array())
 {
+    $maxlength = array_get($args, "maxlength", NULL);
```
jmdyck: Since the default is NULL, you could just say
`$maxlength = @$args["maxlength"];`

cpeel: I've come to dislike using `@` for this purpose and would prefer we stop using it in most all cases. It's like blindly catching and ignoring exceptions in python. I see this as a bad programming practice and would rather explicitly handle cases rather than swallowing errors.

jmdyck: Yeah, I can agree to that.

Add it to https://www.pgdp.net/wiki/DP_Code_Best_Practices#Coding_conventions ?

cpeel: Added as https://www.pgdp.net/wiki/DP_Code_Best_Practices#Squashing_errors.2Fwarnings adjust wording as you see fit.

PR #29, opened 2018-07-16
Added ability to set zoom percent using numpad (task #1531)

mlazaric:
[Here](https://www.cambiaresearch.com/articles/15/javascript-char-codes-key-codes) is a simple reference for key codes where I got the numpad codes. [Here](https://www.pgdp.net/c/tasks.php?action=notify_me&task_id=1531) is task #1531 which this solves. :)

On a slightly related note, it might be beneficial to allow enter (13) because it would allow users to set the zoom without having to click the button.


cpeel:
> On a slightly related note, it might be beneficial to allow enter (13) because it would allow users to set the zoom without having to click the button.

I like this idea if you want to add an additional commit to your branch to enable that.

mlazaric:
> I like this idea if you want to add an additional commit to your branch to enable that.

Done in 7389135. :)

PR #30, opened 2018-07-16
Only alert if both word lists are empty

cpeel:
Rather than issuing a warning if one of the word lists is empty,
only alert if both of them are empty. This was both the original
ask and what was announced, but not what was coded.


srjfoo:
Looks good to me.

PR #34, opened 2018-07-24
Remote file manager updates

cpeel:
This started with a request from @srjfoo to allow all users to upload files into the Commons folder.

While I was here I standardized some of the page styles, streamlined some common operations to remove interstitials, and addressed a race condition in the resumable upload reassembly.

### DIFF
--- a/tools/project_manager/remote_file_manager.php
+++ b/tools/project_manager/remote_file_manager.php
@@ -538,14 +538,35 @@ function do_resumable_chunk()
     // attempt to assemble any completed files; this needs to run both for
     // testChunks and uploads to handle edge failure cases where all of the
     // parts have been uploaded but not reassembled
+
+    // Ensure we have every chunk we're looking for and the entire file's
+    // worth of data. Chunks are uploaded sequentially, so start at the top
+    // and work our way down to fail early.
     $size_on_server = 0;
-    foreach(scandir($staging_dir) as $filename)
+    for($i=$total_chunks; $i>=1; $i--)
     {
-        $size_on_server = $size_on_server + filesize("$staging_dir/$filename");
+        $chunk_name = "$staging_dir/$hashed_filename.part.$i";
+
+        // if the chunk doesn't exist, return
+        if(!is_file($chunk_name))
+            return;
+
+        $size_on_server = $size_on_server + filesize($chunk_name);
     }
 
+    // We have all the chunks, but we need to confirm all of them have finished
+    // uploading. This can happen if they are being uploaded concurrently.
     if($size_on_server >= $total_size)
     {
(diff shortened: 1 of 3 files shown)


tools/project_manager/remote_file_manager.php
```diff
@@ -952,7 +964,8 @@ function get_current_dir_relative_path($home_dirname)
 
     // Only SAs are allowed access to other home folders.
     if (!user_may_access_all_upload_dirs()) {
-        if (!startswith("$abspath/", "$abs_uploads_dir/$home_dirname/")) {
+        if (!startswith("$abspath/", "$abs_uploads_dir/$home_dirname/") &&
+            !startswith("$abspath/", "$abs_uploads_dir/$commons_rel_dir/")) {
             fatal_error( _("You are restricted to your home folder and its descendants") );
```
srjfoo: Should this error message be changed to "You are restricted to your home folder or the Commons folder and their descendants"?

tools/project_manager/remote_file_manager.php
```diff
@@ -1242,6 +1255,14 @@ function show_home_link()
     echo "<p><a href='?action=showdir'>$text</a></p>\n";
 }
 
+function show_commons_link()
+{
+    global $commons_rel_dir;
+    $text = sprintf(_("Return to Commons folder"));
+    $url = "?cdrp=" . urlencode($commons_rel_dir);
```
srjfoo: I thought this was going to be changed to "Go to Commons folder", since if their home directory isn't the Commons, "Return" doesn't make sense.

tools/project_manager/remote_file_manager.php
```diff
@@ -683,11 +726,14 @@ function do_showmove()
     // Remove first element (which is $home_path itself)
     unset($user_subdirs[0]);
     $valid_target_dirs = array_merge($valid_target_dirs, $user_subdirs);
-    natcasesort($valid_target_dirs);
 
-    // Allow users to tranfer files to Commons too
-    if(!in_array($commons_dir, $valid_target_dirs))
-        array_unshift($valid_target_dirs, $commons_dir);
+    // And all subdirectories in the Commons directory
+    $common_subdirs = searchdir($commons_dir, 2, "DIRS");
```
srjfoo: What does the "And" in the comment refer to? To me it makes most sense following the "Allow users to transfer files to Commons too", but it looks like that was deleted.

tools/project_manager/remote_file_manager.php
cpeel: This was changed, I believe you were looking at an older commit.

cpeel:
I've addressed the review comments in fixup commits that will be rebased before merging.

PR #35, opened 2018-07-25
Add XML and RSS feeds for smooth reading

mlazaric:
Based on this task: https://www.pgdp.net/c/tasks.php?action=show&task_id=1108

Also fixes that error with RSS and XML feeds which occurred when a book title contained letters with diacritical marks. (@cpeel found the solution to that)

### DIFF
--- a/feeds/backend.php
+++ b/feeds/backend.php
@@ -3,7 +3,7 @@ $relPath="./../pinc/";
 include_once($relPath.'base.inc');
 include_once($relPath.'misc.inc'); // xmlencode()
 
-$content = get_enumerated_param($_GET, 'content', 'posted', array('posted', 'postprocessing', 'proofing', 'news')); // Which feed the user wants
+$content = get_enumerated_param($_GET, 'content', 'posted', array('posted', 'postprocessing', 'proofing', 'smoothreading', 'news')); // Which feed the user wants
 $refreshDelay = 30 * 60; // Time in seconds for how often the feeds get refreshed
 $refreshAge = time()-$refreshDelay; // How long ago $refreshDelay was in UNIX time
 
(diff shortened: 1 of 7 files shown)


cpeel:
What you've implemented looks great!

I feel that we should be exposing the XML/RSS feeds somewhere in the UI for people to find them. Likely somewhere on the [SR page](https://www.pgdp.net/c/tools/post_proofers/smooth_reading.php). @srjfoo probably has some ideas. The pgdpdev slack channel may be a better place to hash that out.

srjfoo:
Off the top of my head, I can start with adding a news item about it to the SR page, and perhaps a general one mentioning all of the feeds on the AH (perhaps fixed for a week or two, and then random)? I'd like some help with wording, though.

Oh, and the blog!

@cpeel, Those are the places I can think of that wouldn't take extra coding. If you want the links displayed permanently on the round and pool pages, maybe in the statsbar? The XML/RSS feeds aren't stats, but not everything in the statsbar is. If not there, then we definitely need a discussion about where.

Sharon

cpeel:
I was thinking something permanent on the page. Perhaps something as simple as adding the RSS/XML images used for the other feeds onto the SR page under the "Projects Available for Smooth Reading" (ie, at https://github.com/DistributedProofreaders/dproofreaders/blob/master/pinc/showavailablebooks.inc#L119).

cpeel:
I also like the idea of news items, etc too though.

srjfoo:
That sounds like a good idea. My initial thought wouldn't work for the SR pool page, anyway, as it doesn't have (and doesn't really need) a statsbar.

mlazaric:
Inline with the heading or on the next line? 

Inline:

image

Next line:

image

srjfoo:
I'd say next line, also, but on looking at the SR pool page, does it really make sense to have a heading "Projects Available for Smooth Reading" followed immediately by "Projects Currently Available"

Perhaps a two-stage change -- combine the two headings (if possible, and there's not a reason for having both) into "Projects Currently Available for Smooth Reading", and then adding an explanation of what they get by subscribing to a feed at the end of the para following the heading. Something like "If you use RSS/XML feeds, you will (get notification?)(see books added to the list?) as soon as a new book is added to the pool. (I've never used RSS enough to really understand how they work, so I don't know how to describe what they get.)

Do you think that the SR team should be asked before that double heading is combined into one?

Sharon

cpeel:
Compared to the other round pages, SR is a special snowflake because it has two project listings - the "Projects you have volunteered to Smooth Read" and the "Projects Available for Smooth Reading". "Projects Currently Available" is generated from shared code between all the rounds and SR, so changing that is going to be non-trivial. We could just remove the "Projects Available for Smooth Reading" but that might introduce some ambiguity on what the second projects listing is.

Either way this is scope creep and I suggest we defer it from this PR.

srjfoo:
If it's going to cause scope creep, then yes, definitely defer it.

So, back to the original question. Will just adding the links be enough, without any explanation? If so, I also vote "next line", if not, I vote to add an explanation with the links. Perhaps where it is (though if so, back to creep again, wouldn't "Projects Newly Available for Smooth Reading" make more sense?), perhaps at the end of or as another para below. Whichever is easier.

cpeel:
Actually, only printing one of the headers for SR is easier after I stopped thinking about it and seems worthwhile to include here:
```
diff --git a/pinc/showavailablebooks.inc b/pinc/showavailablebooks.inc
index 4ca4b20..b4b9e33 100644
--- a/pinc/showavailablebooks.inc
+++ b/pinc/showavailablebooks.inc
@@ -172,8 +172,12 @@ function show_projects_for_stage(
 {
     global $pguser;

-    $title = _('Projects Currently Available');
-    echo "\n<h2>$title</h2>";
+    // SR is a special snowflake that prints its own header
+    if($stage != get_Stage_for_id("SR"))
+    {
+        $title = _('Projects Currently Available');
+        echo "\n<h2>$title</h2>";
+    }
     echo "\n<p>{$stage->description}</p>";

     if ($show_filter_block)
```

Perhaps instead of just the links on the new line, having the following paragraph:
> Feeds of all projects currently available for smooth reading: [XML] [RSS]

And don't forget to wrap it in `_()` for translation:
```
// TRANSLATORS: %s are image links to the XML and RSS feeds
echo "<p>";
echo sprintf(_("Feeds of all projects currently available for smooth reading: %s"), $xml_rss_images);
echo "</p>";
```

mlazaric:
It looks like this after the last commit: 

image

srjfoo:
I'd capitalize Smooth Reading, but other than that, looks good to me.

cpeel:
Indeed - we do tend to capitalize it everywhere else.

@mlazaric, if you would make that change and rebase it into this last commit I'll get this merged.

mlazaric:
@cpeel amended to the last commit.  :) 

image

cpeel:
This is great - thank you for working on this!

PR #36, opened 2018-07-26
functionality for notification of new tasks

70ray:
implements task #1694.
a new value 'notify_new' is added to the UserSetting 'taskctr_notice'
for this purpose.
(the existing code uses also the value 'all'. This seems to be
only partially implemented since there is no way of setting it).

### DIFF
--- a/tasks.php
+++ b/tasks.php
@@ -456,6 +456,7 @@ function create_task_from_form_submission($formsub)
     global $versions_array;
     global $now_sse;
     global $requester_u_id;
+    global $site_abbreviation;
 
     if (empty($formsub['task_summary']) || empty($formsub['task_details'])) {
         return "You must supply a Task Summary and Task Details.";
@@ -521,7 +522,7 @@ function create_task_from_form_submission($formsub)
         global $tasks_url, $code_url;
         maybe_mail(
             $task_assignee_user->email,
-            "DP Task Center: Task #$task_id has been assigned to you",
+            "$site_abbreviation Task Center: Task #$task_id has been assigned to you",
             $task_assignee_user->username . ", you have been assigned task #$task_id.  Please visit this task at $tasks_url?action=show&task_id=$task_id.\n\nIf you do not want to accept this task please edit the task and change the assignee to 'Unassigned'.\n\n--\nDistributed Proofreaders\n$code_url\n\nThis is an automated message that you had requested please do not respond directly to this e-mail.\r\n"
         );
     }
@@ -567,7 +568,7 @@ if (!isset($_REQUEST['task_id'])) {
         case 'list_open':
             // The user just entered the Task Center
             // (e.g., by clicking the "Report a Bug" link).
-            TaskHeader("All Open Tasks");
+            TaskHeader("All Open Tasks", true);
             list_all_open_tasks();
             break;
 
@@ -579,7 +580,7 @@ if (!isset($_REQUEST['task_id'])) {
                 ShowNotification($errmsg, true);
                 break;
             }
-            TaskHeader("All Open Tasks");
+            TaskHeader("All Open Tasks", true);
             list_all_open_tasks();
             break;
 
@@ -922,7 +923,7 @@ function dropdown_select($field_name, $current_value, $array)
     echo "</select>\n";
 }
 
-function TaskHeader($header)
+function TaskHeader($header, $show_new_alert = false)
 {
     global $tasks_url, $pguser;
     $js_data = <<<EOS
@@ -978,15 +979,18 @@ EOS;
     echo "<tr><td width='50%'>";
     echo "<a href='$tasks_url'>Task Center Home</a> | <a href='$tasks_url?action=show_creation_form'>New Task</a>";
     echo "<form method='get' style='display: inline;'>";
-    if($notified_for_new)
+    if($show_new_alert)
     {
-        echo "<input type='hidden' name='action' value='unnotify_new'>";
-        echo " | <input type='submit' value='Stop New Task Alerts'>";
-    }
-    else
-    {
-        echo "<input type='hidden' name='action' value='notify_new'>";
-        echo " | <input type='submit' value='Receive New Task Alerts'>";
+        if($notified_for_new)
+        {
+            echo "<input type='hidden' name='action' value='unnotify_new'>";
+            echo " | <input type='submit' value='Stop New Task Alerts'>";
+        }
+        else
+        {
+            echo "<input type='hidden' name='action' value='notify_new'>";
+            echo " | <input type='submit' value='Receive New Task Alerts'>";
+        }
     }
     echo "</form>";
     echo "</td>\n";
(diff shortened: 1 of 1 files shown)


tasks.php
```diff
@@ -1563,17 +1588,28 @@ function TaskComments($tid)
...
 }
 
-function NotificationMail($tid, $message)
+function NotificationMail($tid, $message, $new_task = false)
 {
     global $code_url, $tasks_url, $pguser;
-    $notify_setting_all = Settings::get_users_with_setting('taskctr_notice', 'all');
-    $notify_setting_this = Settings::get_users_with_setting('taskctr_notice', $tid);
-    $users_to_notify = array_merge($notify_setting_all, $notify_setting_this);
+    if($new_task)
```
wf49670: With this change, it appears that users with taskcenter_notice = all will no longer receive notifications for new tasks. They are supposed to receive notifications for all new or updated tasks.

(Your earlier comment that there's no way to set "all" is not quite correct. Yes, there is no user interface to set it, but it can be set directly in the database by a Squirrel, and the "all" functionality is being used, and must continue to work as it currently does.)

70ray: Thanks Walt, I'l do that. Do you think it would it be useful to add a user interface to set "all"?

wf49670:
I will defer to Casey and others about adding an interface to set (or 
unset) "all", Ray.

I suspect there is one user (FallenArchangel) who would make use of it. 
I don't know if there are others. But mostly I don't know whether 
there's any philosophical reason to allow or disallow "normal" users 
from having that capability. Today I think that only a few Squirrels 
have that value set, though we'd probably be willing to set it for any 
of our Developers if they asked.

Regards,
Walt

srjfoo:
I think that Casey and Michael probably have the best understanding of any potential impact on the server, so I think one of them should weigh in. As far as I know, the main impact would be in outgoing email traffic.

Sharon

cpeel:
At a high level, I'm fine with adding this functionality.

I'm not a fan of using a link to enact this change. The rest of the page uses links to access a new page of some kind and form elements (or at least buttons) to do something like search or add a comment or 'me too' a task.

The other aspect is that the links looks awfully out of place over there. We should consider extracting the `Task Center Home` and `New Task` links from the search box and putting them all over in that corner, something like:
```
Task Center Home | New Task | [Receive New Task Alerts]
```
Where the latter is a button similar to the Search button. 

I pulled your branch and coded up what I'm thinking in my sandbox (at least it's there now) if you want to look: https://www.pgdp.org/~cpeel/c/tasks.php

70ray:
I agree with Casey about buttons and links. I was following the example of the "Signup for task notifications" link on the task form so we should really also change that to a button.

tasks.php
```diff
@@ -580,7 +582,21 @@ function create_task_from_form_submission($formsub)
             TaskHeader("All Open Tasks");
             list_all_open_tasks();
             break;
-        }
+
+        case 'notify_new':
+            $userSettings =& Settings::get_Settings($pguser);
+            $userSettings->add_value('taskctr_notice', 'notify_new');
+            metarefresh(0, "$tasks_url");
```
cpeel: You don't need the quotes around the variable here.

tasks.php
```diff
@@ -580,7 +582,21 @@ function create_task_from_form_submission($formsub)
...
+//            TaskHeader("All Open Tasks");
```
cpeel: Please remove commented-out code.

tasks.php
```diff
@@ -580,7 +582,21 @@ function create_task_from_form_submission($formsub)
...
+  //          list_all_open_tasks();
+            break;
+
+        case 'unnotify_new':
+            $userSettings =& Settings::get_Settings($pguser);
+            $userSettings->remove_value('taskctr_notice', 'notify_new');
+            TaskHeader("All Open Tasks");
```
cpeel: We should probably be consistent if we're redirecting or not for both notify_new/unnotify_new.

tasks.php
```diff
@@ -788,12 +804,12 @@ function handle_action_on_a_specified_task()
     elseif ($action == 'notify_me') {
         $userSettings =& Settings::get_Settings($pguser);
         $userSettings->add_value('taskctr_notice', $task_id);
-        TaskDetails($task_id);
+        metarefresh(0, "$tasks_url?action=show&task_id=$task_id");
```
cpeel: I'm uncertain this will work because things have already been output so a header redirect will fail. We either need to continue with the old method or move this up and out of this function and do the evaluation earlier before TaskHeader().

tasks.php
70ray: doh! ... forgot to save file from editor.

tasks.php
```diff
@@ -1602,32 +1602,36 @@ function TaskComments($tid)
...
-    global $code_url, $tasks_url, $pguser;
+    global $code_url, $tasks_url, $pguser, $date_str, $time_of_day_str;
+
+    $result = mysqli_query(DPDatabase::get_connection(), "SELECT task_summary, task_details FROM tasks WHERE task_id = $tid LIMIT 1");
+    if(!$result)
+        return;
+    $row = mysqli_fetch_assoc($result);
+    $task_summary = $row['task_summary'];
+
+    $message = $message . "\nYou can see task #$tid by visiting $tasks_url?action=show&task_id=$tid\n\n--\n
```
cpeel: `action=show` is the default so it isn't needed in the url, eg: https://www.pgdp.org/c/tasks.php?task_id=12

tasks.php
```diff
@@ -1602,32 +1602,36 @@ function TaskComments($tid)
...
         if ($username != $pguser) {
             $user = new User($username);
-            maybe_mail($user->email, $message_header,
-                $message_intro
-              . "$message" . "\n"
-              . "You can see task #$tid by visiting $tasks_url?action=show&task_id=$tid" . "\n\n"
-              . "--" . "\n"
-              . "Distributed Proofreaders" . "\n" . "$code_url" . "\n\n"
-              . "This is an automated message that you had requested, please do not respond directly to this e-mail.");
+            maybe_mail($user->email, "DP Task #$tid: $task_summary", $message);
```
cpeel: We should use `$site_abbreviation` instead of `DP` here. And yes, this was poorly done in the prior code.

tasks.php
cpeel: I don't think it's useful for most members to watch everything that happens in the task center, so I vote no on a UI to set the 'all' attribute.

tasks.php
cpeel: Also, while the former method down near line 164 was more "wordy" with each message line being on its own line in the PHP code, it made it much easier to read.

tasks.php
```diff
@@ -586,16 +586,13 @@ function create_task_from_form_submission($formsub)
         case 'notify_new':
             $userSettings =& Settings::get_Settings($pguser);
             $userSettings->add_value('taskctr_notice', 'notify_new');
-            metarefresh(0, "$tasks_url");
-//            TaskHeader("All Open Tasks");
-  //          list_all_open_tasks();
+            metarefresh(0, $tasks_url);
```
cpeel: This is minor, but worth pointing out. If you are on a task page (eg: https://www.pgdp.org/~cpeel/c/tasks.php?action=show&task_id=2229) and use the Receive New Task Alerts button you get shunted to the main tasks page instead of back to the task you were viewing. Ideally this would go back to the task page. One probably-easy solution would be to only show the button if the user was looking at the full task list rather than just a single task.

Either way this isn't super-important so feel free to push back on this request.

tasks.php
70ray: Being an old and low-level programmer it's hard to get out of the habit of making the code more efficient (like making two strings into one instead of letting the computer do it). I'll put it back like it was.

tasks.php
70ray: I've made the button only appear if the user is looking at all tasks. There are some cases where the button does not appear when it could: e.g. after editing a task. It would need more work to fix that ...

cpeel:
Thanks for working on this Ray!

PR #37, opened 2018-07-27
Fix double escaped task title and duplication of tasks and comments

mlazaric:
## Fix task_summary being escaped twice

This was specifically affecting the title of some task related pages such as the show task page and the search task page.

### First part: the show details of task page 

For example https://www.pgdp.net/c/tasks.php?action=show&task_id=1707

Before:  
image

After:
image

### Second part: the search task page

Before:
image

After:
image

### Related task

- https://www.pgdp.net/c/tasks.php?action=show&task_id=1711

## Fix task duplication caused by reloading the page 

This occurred when submitting a task and then reloading the page, meaning the same request would be sent to the server and it would create another task with the same data.

### Related tasks

- https://www.pgdp.net/c/tasks.php?action=show&task_id=1690  

- https://www.pgdp.net/c/tasks.php?action=show&task_id=1168

- https://www.pgdp.net/c/tasks.php?action=show&task_id=470

##  Fix comment duplication caused by reloading the page

Similar to the previous issue, it occurred when submitting a comment and then reloading the page.

### Related tasks

Can't seem to find any related tasks, but I've seen it happen in a couple places, for example:

- https://www.pgdp.net/c/tasks.php?action=show&task_id=1707 (duplicated comment)

- https://www.pgdp.net/c/tasks.php?action=show&task_id=580 (duplicated comment)

- https://www.pgdp.net/c/tasks.php?action=show&task_id=1388 (duplicated comment)


tasks.php
```diff
@@ -577,9 +577,14 @@ function create_task_from_form_submission($formsub)
...
             }
-            TaskHeader("All Open Tasks");
-            list_all_open_tasks();
-            break;
+            else
+            {
+                // If we successfully create the task, we should reload
+                //   the page to clear the POST data and make sure that
+                //   reloading does not lead to duplicated tasks.
+                header("Location: $code_url/tasks.php");
```
cpeel: The "DP" way to do this is `metarefresh()` which handles some test conditions too.

tasks.php
```diff
@@ -728,6 +739,8 @@ function handle_action_on_a_specified_task()
     }
     elseif ($action == 'add_comment') {
         if (!empty($_POST['task_comment'])) {
+            global $code_url;
```
cpeel: Put global declarations at the top of the function.

tasks.php
```diff
@@ -582,7 +583,7 @@ function create_task_from_form_submission($formsub)
                 // If we successfully create the task, we should reload
                 //   the page to clear the POST data and make sure that
                 //   reloading does not lead to duplicated tasks.
-                header("Location: $code_url/tasks.php");
+                metarefresh(0, "$code_url/tasks.php");
```
70ray: There already exists a variable $tasks_url which you could use instead of $code_url/tasks.php.

mlazaric: Thanks! Fixed in dfa3273172502edbe898e5c978c01025f7501666 and f49ee4e079d3571be2a841eee4cb9061b4c3a8be :)

tasks.php
mlazaric: Fixed in 8a53710d6c0fc4d8f798834b13a96575251b00b8, switched `$code_url` to `$tasks_url` thanks to @70ray in f49ee4e079d3571be2a841eee4cb9061b4c3a8be.

tasks.php
mlazaric: Switched to `metarefresh` in 44ec740ad8c7fdf99dee6a4f0f5bbe066c6194c5 and 8a53710d6c0fc4d8f798834b13a96575251b00b8.

cpeel:
Nicely, done. If you will squash those commits we can get this merged.

mlazaric:
Squashed :)

PR #39, opened 2018-07-30
In proofing rounds character check only

70ray:
I did this some time ago; I think there were several commits but they seem to have got lost, possibly as a result of rebasing to master and forced push.
This branch makes a different button or icon in formatting and proofing rounds. Some functions have an additional parameter to distinguish the proofing/formatting rounds. In the proofing rounds the button is called 'character check'. There is a new icon for enhanced interface with a 'C'.  The only check it does in proofing rounds is for tab characters. The preview shows characters with diacritical marks as html entities and subscripts and superscripts. In formatting rounds these functions are also done as well as the previous formatting checks.


wf49670:
Also, if would be useful if it detected Windows-1252 characters such as the oe ligature, and UTF-8 characters that have become numeric HTML entities and flagged them as errors.

70ray:
Thanks Walt, I'll look at flagging characters that could be done with Latin-1 etc.

70ray:
Made a new branch proof_check_1a

PR #40, opened 2018-07-30
Add link tags for automatic detection of RSS feeds

mlazaric:
Inspired by [this post](https://developer.mozilla.org/en-US/docs/Web/RSS/Getting_Started/Syndicating#Adding_the_.3Clink.3E), I have added the relevant `link` tags which allow RSS feed readers to determine which RSS feed is connected to that page. 

### [RSS Feed Reader extension for Chrome](https://chrome.google.com/webstore/detail/rss-feed-reader/pnjaodmkngahhkoihejjehlcdlnohgmp)

If you open `/c/list_etexts.php?x=g` the extension detects the RSS feed and allows you to add it easily, which looks like this.

image

### Firefox

If you open the same page, you are able to choose the option "Subscribe to This Page..." which allows you to create a Live Bookmark.

image

### Completed Gold E-Texts

- URL: `/c/list_etexts.php?x=g`
- Title of RSS Feed: `Completed Gold E-Texts`
- RSS Feed: `/c/feeds/backend.php?content=posted&type=rss`

### In Progress Silver E-Texts

- URL: `/c/list_etexts.php?x=s`
- Title of RSS Feed: `In Progress Silver E-Texts`
- RSS Feed: `/c/feeds/backend.php?content=postprocessing&type=rss`

### Now Proofreading Bronze E-Texts

- URL: `/c/list_etexts.php?x=b`
- Title of RSS Feed: `Now Proofreading Bronze E-Texts`
- RSS Feed: `/c/feeds/backend.php?content=proofing&type=rss`

### Smooth Reading Pool Preview

- URL: `/c/tools/post_proofers/smooth_reading.php`
- Title of RSS Feed: `Smooth Reading E-Texts`
- RSS Feed: `/c/feeds/backend.php?content=smoothreading&type=rss`

### DIFF
--- a/list_etexts.php
+++ b/list_etexts.php
@@ -18,21 +18,30 @@ if($x == "g") {
     $title = _("Completed Gold E-Texts");
     $state = SQL_CONDITION_GOLD;
     $info = _("Below is the list of Gold e-texts that have been produced on this site. Gold e-texts are books that have passed through all phases of proofreading, formatting, and post-processing. They have been submitted to Project Gutenberg and are now available for your enjoyment and download.");
+    $rss_content = "posted";
 } elseif ($x == "s") {
     $type = "Silver";
     $title = _("In Progress Silver E-Texts");
     $state = SQL_CONDITION_SILVER;
     $info = _("Below is the list of Silver e-texts that have almost completed processing on our site. Silver e-texts are books that have passed through all phases of proofreading and formatting and are now in the post-processing phase. Post-processing is the final assembly stage in which one volunteer performs a series of checks for consistency and correctness before the e-book is submitted to Project Gutenberg for your enjoyment and download.");
+    $rss_content = "postprocessing";
 } elseif ($x == "b") {
     $type = "Bronze";
     $title = _("Now Proofreading Bronze E-Texts");
     $state = SQL_CONDITION_BRONZE;
     $info = _("Below is the list of Bronze e-texts that are currently available for proofreading on this site. Bronze e-texts are what our newest volunteers see and what you can work on now by logging in. These e-texts are in the initial stages of proofreading where everyone has a chance to correct any OCR errors which may be found. After going through a number of other phases, the e-text then goes to an experienced volunteer for final assembly (post-processing), after which the e-text is submitted to Project Gutenberg for your enjoyment and download.");
+    $rss_content = "proofing";
 } else {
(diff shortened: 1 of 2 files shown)


list_etexts.php
```diff
@@ -18,21 +18,26 @@
...
     $state = SQL_CONDITION_BRONZE;
     $info = _("Below is the list of Bronze e-texts that are currently available for proofreading on this site. Bronze e-texts are what our newest volunteers see and what you can work on now by logging in. These e-texts are in the initial stages of proofreading where everyone has a chance to correct any OCR errors which may be found. After going through a number of other phases, the e-text then goes to an experienced volunteer for final assembly (post-processing), after which the e-text is submitted to Project Gutenberg for your enjoyment and download.");
+    $rss_content = "proofing";
 } else {
     die("x parameter must be 'g', 's', or 'b'. ('$x')");
 }
 
-output_header($title);
+output_header($title, SHOW_STATSBAR, array(
+    "head_data" => "<link rel='alternate' type='application/rss+xml' href='$code_url/feeds/backend.php?content=$rss_content&type=rss' title='$title' />",
```
cpeel: $title needs to be escaped with `attr_safe()` in case the translated version has non-attribute-safe characters. Also, consider pulling the array into its own variable and passing that rather than breaking the open_header() call onto multiple lines (totally valid, but most of the code uses the assign-to-variable-and-pass-it approach for additional header attributes).

tools/post_proofers/smooth_reading.php
```diff
@@ -22,7 +22,9 @@
 }
 
 // we show more columns when user is logged in, so we don't have room for the stats bar
-output_header( $header_text, $logged_in ? NO_STATSBAR : SHOW_STATSBAR);
+output_header($header_text, $logged_in ? NO_STATSBAR : SHOW_STATSBAR, array(
+    "head_data" => sprintf("<link rel='alternate' type='application/rss+xml' href='$code_url/feeds/backend.php?content=smoothreading&type=rss' title='%s' />", _("Smooth Reading E-Texts")),
```
cpeel: Ditto need for `attr_safe()`'ing the title attribute and use of a variable.

list_etexts.php
mlazaric: Good catch! Fixed both in 1d8ed3d3601b81fadf3c699c6ef650e870e27495 :)

cpeel:
Looks good - squash those and we'll land this.

PR #41, opened 2018-07-30
Various RSS feed optimizations and configurations

mlazaric:
- Added @srjfoo's and @jmdyck's optimized query for SR
- Changed number of items in RSS feed to 20
- Disabled RSS refresh delay on TEST server to make testing easier

### DIFF
--- a/feeds/backend.php
+++ b/feeds/backend.php
@@ -30,17 +30,21 @@ if(!file_exists($xmlfile) || filemtime($xmlfile) < $refreshAge) {
         switch($content) {
             case "posted":
                 $condition = sprintf("state='%s'", PROJ_SUBMIT_PG_POSTED);
+                $desc = sprintf(_("The latest releases posted to Project Gutenberg from %1\$s."), $site_name);
                 break;
             case "postprocessing":
                 $condition = sprintf("state='%s'", PROJ_POST_FIRST_AVAILABLE);
+                $desc = sprintf(_("The latest releases available at %1\$s for post-processing."), $site_name);
                 break;
             case "proofing":
                 $condition = sprintf("state='%s'", PROJ_P1_AVAILABLE);
+                $desc = sprintf(_("The latest releases available at %1\$s for proofreading."), $site_name);
                 break;
             case "smoothreading":
                 $condition = "
                     state = 'proj_post_first_checked_out' AND
                     smoothread_deadline > UNIX_TIMESTAMP()";
+                $desc = sprintf(_("The latest releases available at %1\$s for Smooth Reading."), $site_name);
                 break;
         }
         $data = '';
@@ -79,7 +83,7 @@ if(!file_exists($xmlfile) || filemtime($xmlfile) < $refreshAge) {
                 <atom:link href=\"$encoded_url\" rel=\"self\" type=\"application/rss+xml\" />
                 <title>".xmlencode($site_name)." - " . _("Latest Releases") . "</title>
                 <link>".xmlencode($code_url)."</link>
-                <description>".xmlencode(sprintf( _("The latest releases posted to Project Gutenberg from %1\$s."), $site_name))."</description>
+                <description>".xmlencode($desc)."</description>
                 <webMaster>".xmlencode($site_manager_email_addr)." (" . xmlencode(_("Site Manager")) . ")</webMaster>
                 <pubDate>".xmlencode($lastupdated)."</pubDate>
                 <lastBuildDate>".xmlencode($lastupdated)."</lastBuildDate>
--- a/feeds/backend.php
+++ b/feeds/backend.php
@@ -4,7 +4,8 @@ include_once($relPath.'base.inc');
 include_once($relPath.'misc.inc'); // xmlencode()
 
 $content = get_enumerated_param($_GET, 'content', 'posted', array('posted', 'postprocessing', 'proofing', 'smoothreading', 'news')); // Which feed the user wants
-$refreshDelay = 30 * 60; // Time in seconds for how often the feeds get refreshed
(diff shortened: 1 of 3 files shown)


feeds/backend.php
```diff
@@ -97,7 +124,7 @@
 
     if ($content == "news") {
         $data = '';
-        $result = mysqli_query(DPDatabase::get_connection(), "SELECT * FROM news_items ORDER BY date_posted DESC LIMIT 10");
```
mlazaric: Where is the button for this feed? I forgot to add a `link` tag for it in my previous PR.

feeds/backend.php
```diff
@@ -30,21 +33,45 @@
         switch($content) {
             case "posted":
                 $condition = sprintf("state='%s'", PROJ_SUBMIT_PG_POSTED);
```
mlazaric: We could go back to `$state` instead of `$condition` since the special case of SR is being handled in a different way. Not sure if we _want_ to though.

feeds/backend.php
cpeel: Yeah, it's for all of the round pages. That query was never updated when the news stuff got updated over the years so it's showing things that are hidden/disabled and for all languages. The code that is commented out in `site_news.inc` isn't exactly useful because the feed (as coded) is showing news for the entire site, whereas the per-round page should show just those for the round (well, `news_page_id` specifically).

I'm partial to removing the news feeds entirely since getting them working correctly with the user's language, etc is probably not worth the effort for a feature that has been disabled for probably a decade or more.

feeds/backend.php
```diff
@@ -30,21 +33,45 @@
...
+                WHERE $condition 
+                ORDER BY modifieddate DESC 
+                LIMIT $limit";
+        }
+
         $data = '';
-        $result = mysqli_query(DPDatabase::get_connection(), "SELECT * FROM projects WHERE $condition ORDER BY modifieddate DESC LIMIT 10");
+        $result = mysqli_query(DPDatabase::get_connection(), $query);
         while ($row = mysqli_fetch_array($result)) {
             $posteddate = date("r",($row['modifieddate']));
```
cpeel: `modifieddate` isn't being returned in the new SR query so this should be generating a NOTICE if `$testing = 1`. For SR, this should probably be the e.timestamp value. You could do this via:
```
SELECT 
    projectid, nameofwork, genre, language, FROM_UNIXTIME(e.timestamp) as modifieddate
```

feeds/backend.php
```diff
@@ -30,21 +33,45 @@
...
-                    smoothread_deadline > UNIX_TIMESTAMP()";
+                $query = "
+                    SELECT 
+                        projectid, nameofwork, genre, language, FROM_UNIXTIME(e.timestamp)
+                    FROM projects
+                    JOIN project_events e USING (projectid)
+                    WHERE
+                        e.event_type = 'smooth-reading' AND
+                        e.details1 = 'text available' AND
+                        e.timestamp > UNIX_TIMESTAMP() - (30*24*60*60)
```
cpeel: Might add a comment above this query that we're only looking at SR projects that have moved into SR in the last 30 days -- hence this bit.

feeds/backend.php
cpeel: Nah, I don't see a need to change it back.

feeds/backend.php
mlazaric: Added in 9ec53768b5bc3af587162cebb76e68cc2d1b508c.

feeds/backend.php
mlazaric: I modified the query to add `modifieddate` in 9ec53768b5bc3af587162cebb76e68cc2d1b508c, but I'm not sure what the "generating a NOTICE" part refers to.

srjfoo: If you assign 'modifieddate' as an alias for FROM_UNIXTIME(e.timestamp), doesn't that have the potential of being confusing down the line, since modifieddate is an actual field name in another table, and the two are not necessarily the same? (Or am I being too picky?)

feeds/backend.php
mlazaric: Should I remove the news stuff in `backend.php` or should we discuss it further and leave it to a future PR?

feeds/backend.php
cpeel: I think you're being too picky here. :)

feeds/backend.php
cpeel: The referencing code in site_news.inc has been commented out since before 2005 which makes it 13-year-old dead code and we should just remove it.

feeds/backend.php
```diff
@@ -2,10 +2,13 @@
 $relPath="./../pinc/";
 include_once($relPath.'base.inc');
 include_once($relPath.'misc.inc'); // xmlencode()
+include_once($relPath.'site_vars.php'); // $testing
```
cpeel: `site_vars.php` is pulled in with `base.inc` - you should never need to pull this into a script.

mlazaric: Good to know, removed in ee6e168b6c62f5c2edbf44e1bb7e4e8d2890f4cf :)

feeds/backend.php
```diff
@@ -79,7 +106,7 @@
                 <atom:link href=\"$encoded_url\" rel=\"self\" type=\"application/rss+xml\" />
                 <title>".xmlencode($site_name)." - " . _("Latest Releases") . "</title>
                 <link>".xmlencode($code_url)."</link>
```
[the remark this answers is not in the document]
cpeel: That would be best, yes.

cpeel:
Autosquash this and lets get this merged!

PR #43, opened 2018-08-02
Disable some project transitions during Smooth Reading (#701)

mlazaric:
Based on [this task](https://www.pgdp.net/c/tasks.php?action=show&task_id=701), I have disabled some project transitions while the project is available for smooth reading. 

## site_managers

The following actions have been disabled for a `site_manager` during SR:

 - Return to Available
 - Upload for Verification
 - Change State to Unavailable for Post-Processing
 - Change State to Completed Post-Processing
 - Change State to Completed and Posted to Project Gutenberg

### Screenshots

image

image

## holder

The following actions have been disabled for a `holder` during SR:

 - Return to Available
 - Upload for Verification
 - Change State to Unavailable for Post-Processing

### Screenshots

image

image


pinc/ProjectTransition.inc
```diff
@@ -163,6 +164,14 @@ class ProjectTransition
         die("transition has bad 'restriction' value: '$simple_who_restriction'");
     }
 
+    function is_disabled($project)
```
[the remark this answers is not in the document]
mlazaric: I was considering future proofing it, but I was not sure if that would be over complicating things.

In cacb48f2f51826f7bcb5c24f83183801f21b0704, I ended up going for the approach of having `why_disabled` which returns the reason or an empty string if there is no reason and `is_disabled` which simply checks the `why_disabled` function. `is_disabled` checks `why_disabled` to avoid code duplication and makes it easier to maintain since we don't have to maintain conditions in two places.

Arguably we don't really need `is_disabled` any more since we can just check `why_disabled` manually but since they're both pretty simple and fast functions, we could leave them both as a convenience.

tools/changestate.php
```diff
@@ -39,6 +39,13 @@
     fatal_error( _("You are not permitted to perform this action.") );
 }
 
+if ($transition->is_disabled($project))
+{
+    $body = _("This function is disabled while the project is in the Smooth Reading Pool.<br/><br/>If you believe this is an error, please contact db-req for assistance.");
```
cpeel: This text should take `why_disabled()` into account.

tools/upload_text.php
```diff
@@ -41,6 +41,15 @@
     and the upload does not succeed, upload a small placeholder zip file 
     instead and email %s for assistance."), $db_requests_email_addr);
 
+// Deny post_1 and return_1 if the project is currently in SR
+if (($stage == 'post_1' || $stage == 'return_1') &&
+    $project->is_available_for_smoothreading())
+{
+    $title = _("Disabled during Smooth Reading");
+    $body = _("This function is disabled while the project is in the Smooth Reading Pool.<br/><br/>If you believe this is an error, please contact db-req for assistance.");
```
cpeel: For translation purposes and HTML rendering, I would do:
```
$body = '<p>' . _("This function is disabled...") . '</p>' .
        '<p>' . _("If you believe this...") . '</p>';
```
Also note that we use HTML5, not XHTML, so it would be `<br>` not `<br/>` if you were to use them.

tools/changestate.php
cpeel: For translation purposes and HTML rendering, I would do:
```
$body = '<p>' . _("This function is disabled...") . '</p>' .
        '<p>' . _("If you believe this...") . '</p>';
```
Also note that we use HTML5, not XHTML, so it would be `<br>` not `<br/>` if you were to use them.

cpeel:
Great - autosquash and I'll get this merged!

PR #45, opened 2018-08-11
Redirect Me Too request to the show task page

mlazaric:
This is done for two reasons:

 1. We want to clear the POST data so that by refreshing the user does not get the awkward Confirm Form Resubmission prompt.

 2. After the Me Too request is finished, the user should be redirected to `https://www.pgdp.net/c/tasks.php?action=show&task_id=TASK_ID` and not to `https://www.pgdp.net/c/tasks.php`.

Related task:
 * https://www.pgdp.net/c/tasks.php?action=show&task_id=1715

### DIFF
--- a/tasks.php
+++ b/tasks.php
@@ -641,10 +641,11 @@ function handle_action_on_a_specified_task()
         $userSettings->remove_value('taskctr_notice', $task_id);
         metarefresh(0, "$tasks_url?task_id=$task_id");
     }
-    // We don't want a TaskHeader for add_comment
+    // We don't want a TaskHeader for add_comment or add_metoo
     //   because then we wouldn't be able to redirect
     //   the user.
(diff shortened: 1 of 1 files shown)


tasks.php
```diff
@@ -820,9 +821,8 @@ function handle_action_on_a_specified_task()
         }
         mysqli_free_result($meTooCheck);
 
-        // No need to display a different error message if the user was refreshing
-        ShowNotification("Thank you for your report!  It has been recorded below.", false, "info");
-        TaskDetails($task_id);
+        // Redirect back to show task page to clear POST data
+        metarefresh(0,"$tasks_url?action=show&task_id=$task_id");
```
cpeel: Please add a space after the comma. Go ahead and squash the fixup and we'll get his merged.

mlazaric: Done :)

PR #46, opened 2018-08-11
Only use search_text if it has been passed

cpeel:
There are edge cases (mostly through URL manipulation) where search_text may not be set. This handles those cases more gracefully.

This was found from the daily php_errors report from pgdp.net.

### DIFF
--- a/tasks.php
+++ b/tasks.php
@@ -359,19 +359,21 @@ function SearchParams_get_sql_condition($request_params)
 {
     global $testing, $SearchParams_choices;
 
-    if ($testing) echo_html_comment("\$request_params['search_text'] = {$request_params['search_text']}");
+    $condition = "1";
+    if(isset($request_params['search_text']))
+    {
+        $search_text = $request_params['search_text'];
+        if ($testing)
+            echo_html_comment("\$request_params['search_text'] = $search_text");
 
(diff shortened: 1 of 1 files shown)


tasks.php
```diff
@@ -359,19 +359,21 @@ function SearchParams_get_sql_condition($request_params)
...
-    $search_text_details = $request_params['search_text'];
+    $condition = "1";
+    if(isset($request_params['search_text']))
+    {
+        $search_text = $request_params['search_text'];
+        if ($testing)
+            echo_html_comment("\$request_params['search_text'] = $search_text");
 
-    $condition = sprintf("
+        $condition .= sprintf(" AND
```
mlazaric: Why are we concatenating `$condition` and not simply setting it without the `AND`?

cpeel: Either works. I went with this because it's similar to how it is done elsewhere in the codebase. The idea being that we start with "WHERE 1" and then keep adding conditionals (as appropriate), eg:
* tools/project_manager/page_compare.php
* tools/project_manager/page_table.inc
* pinc/ProjectSearchForm.inc

mlazaric: Okie dokie, everything else looks great!

PR #48, opened 2018-08-29
Add ability to subscribe to SR reports

mlazaric:
Based on this task: https://www.pgdp.net/c/tasks.php?action=show&task_id=1615

It adds a new event called `sr_posted` which is fired every time someone uploads a new smooth reading report. The event is subscribable and off by default. Only the post proofer of that specific project can subscribe to the event.

### Event subscriptions for a non post proofer

image

### Event subscriptions for a post proofer

image

### Current email message

image

### DIFF
--- a/SETUP/db_schema.sql
+++ b/SETUP/db_schema.sql
@@ -701,6 +701,7 @@ CREATE TABLE `user_project_info` (
   `iste_sr_available` tinyint(1) NOT NULL default '0',
   `iste_ppv_enter` tinyint(1) NOT NULL default '0',
   `iste_posted` tinyint(1) NOT NULL default '0',
+  `iste_sr_reported` tinyint(1) NOT NULL default '0',
   PRIMARY KEY  (`username`,`projectid`),
   KEY `projectid` (`projectid`)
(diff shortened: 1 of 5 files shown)


tools/upload_text.php
```diff
@@ -345,6 +343,10 @@
         }
     }
 
+    if ($stage == 'smooth_done') {
```
cpeel: Lets put opening `{` on a newline so the format matches the rest (or most) of this file.

SETUP/db_schema.sql
```diff
@@ -701,6 +701,7 @@ CREATE TABLE `user_project_info` (
   `iste_sr_available` tinyint(1) NOT NULL default '0',
   `iste_ppv_enter` tinyint(1) NOT NULL default '0',
   `iste_posted` tinyint(1) NOT NULL default '0',
+  `iste_sr_posted` tinyint(1) NOT NULL default '0',
```
cpeel: Echoing @srjfoo's comment in slack `sr_reported` or even `sr_uploaded` would probably be clearer.

cpeel:
@srjfoo - is there anything else you want to test with this? If not, we can just squash and merge it.

PR #52, opened 2018-09-26
Validate uploaded zip files are actually zip files

cpeel:
Right now users can uploaded non-zip files by just renaming the file with a .zip extension. This validates that the files uploaded are valid zip files.


srjfoo:
Looks good to me.

PR #54, opened 2018-10-02
Only output postednum field for non-SR feeds

cpeel:
I discovered this bug in the daily php_errors report. postednum wasn't defined for SR feeds (which makes sense because if it's in SR it hasn't been posted). We should stop outputting the postednum for SR feeds.


srjfoo:
Looks good to me.

PR #56, opened 2018-10-12
Don't hard-code DPCustomMono2 anywhere

cpeel:
Now that all (modern) web browsers can use DPCustomMono2 because it's a web font, it's being used in a few places that many users don't expect although they aren't new. This changes those places to use the user's preferred proofreading font instead, giving them control over what is used and making it consistent in several places.


srjfoo:
Tested and looks good to me.

PR #59, opened 2018-11-09
Allow selection of image sources with numeric codes

cpeel:
Image sources can be created with arbitrary alpha-numeric codes, including all-numeric codes like `0`. All-numeric codes mess up the image source selection drop-down due to how php associative arrays work. This fixes the drop-down problem. Ultimately we should fix the image sources design to hide the code from the user since it's an internal-only field.

This is Task 1657.


srjfoo:
Tested in cpeel's sandbox, and behaves as it should -- i.e. "DPT Internal", as the default image source, is not overridden by an image source whose ID is a numeric zero.

PR #61, opened 2018-11-22
Pm new get

70ray:
Changed back to GET instead of POST for project search so that searches can be bookmarked.
Add an invert filter checkbox for the language selector so that it is possible to search for all languages except a few. Otherwise selecting most of the language options would make the url too long. The other multi-selects do not have this problem because there are fewer options. This also simplifies the code slightly.

### DIFF
--- a/pinc/ProjectSearchForm.inc
+++ b/pinc/ProjectSearchForm.inc
@@ -14,7 +14,7 @@ class ProjectSearchWidget
         }
     }
 
-    function get_html_control($search_data)
+    function get_html_control()
     {
         # make all widgets 100% width
         $size_attr = "style='width: 100%;'";
@@ -32,9 +32,9 @@ class ProjectSearchWidget
 
         if ( $this->type == 'text' )
         {
-            if ( isset($search_data[$this->id]) )
+            if ( isset($_GET[$this->id]) )
             {
-                $value_attr = "value='" . attr_safe($search_data[$this->id]) . "'";
+                $value_attr = "value='" . attr_safe($_GET[$this->id]) . "'";
             }
             else // initial_value is set for base class widget
             {
@@ -55,12 +55,12 @@ class ProjectSearchWidget
             }
             foreach ( $this->options as $option_value => $option_label )
             {
-                if(isset($search_data[$this->id]))
+                if(isset($_GET[$this->id]))
                 {
-                    if(($this->can_be_multiple) && is_array($search_data[$this->id]))
-                        $selected = in_array($option_value, $search_data[$this->id]);
+                    if(($this->can_be_multiple) && is_array($_GET[$this->id]))
+                        $selected = in_array($option_value, $_GET[$this->id]);
                     else
-                        $selected = ($option_value == $search_data[$this->id]);
+                        $selected = ($option_value == $_GET[$this->id]);
                 }
                 else
                     $selected = ($option_value == $this->initial_value);
@@ -72,21 +72,19 @@ class ProjectSearchWidget
         }
     }
 
-    function echo_search_item($search_data)
+    function echo_search_item()
     {
         // could fix width of heading cells so they line up when 2nd table is below
         echo "
             <tr>
                 <th class='right-align top-align'>$this->label</th>
-                <td class='search-input top-align left-align'>".$this->get_html_control($search_data)."</td>
+                <td class='search-input top-align left-align'>".$this->get_html_control()."</td>
             </tr>";
     }
 
-    function get_sql_contribution($search_data)
+    function get_sql_contribution()
     {
-        if (!isset($search_data[$this->id]))
-            return NULL;
-        $value = $search_data[$this->id];
+        $value = array_get($_GET, $this->id, '');
         if ( $value == '' )
             return NULL;
         list($column_name,$comparator) = $this->q_contrib;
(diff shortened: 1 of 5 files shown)


pinc/ProjectSearchResults.inc
```diff
@@ -107,9 +107,13 @@ class Column
                 if($sec_sort == $this->id)
                     $sec_sorter = "&nbsp;^";
```
cpeel: If you're going to use brackets for the `else`, use them for the `if` part too.

pinc/ProjectSearchResults.inc
```diff
@@ -650,9 +641,21 @@ class ProjectSearchResults
...
 
+function get_search_configure_link()
+{
+    $origin = urlencode($_SERVER['REQUEST_URI']);
+    return "<a href='{$_SERVER['PHP_SELF']}?show=config&amp;origin=$origin'>" . _("Configure Result") . "</a>";
+}
+
+function make_link($new_array, $text, $anchor = '')
+{
+    $url = "{$_SERVER['PHP_SELF']}?" . http_build_query(array_replace($_GET, $new_array)) . "$anchor";
```
cpeel: I realize this was in the code before your change, but we should avoid using `$_SERVER['PHP_SELF']` for security reasons. Odds are good you can just remove it completely and make this `"?" . http_build_query(...`

tools/project_manager/projectmgr.php
```diff
@@ -87,61 +76,50 @@
...
+    // keep "user_all" or we won't know it is user all
+    $_GET = array_merge($_GET, array('project_manager' => $pguser));
     // TRANSLATORS: Abbreviation for Project Manager
     $sub_title = sprintf(_("All PM projects: %s"), $pguser);
 }
 elseif ($show_view == "user_active")
 {
     $condition = "$PROJECT_IS_ACTIVE_sql AND username = '$pguser'";
-    set_session_user_active();
+    $_GET = array_merge($_GET, array('project_manager' => $pguser, 'state' => array_diff($PROJECT_STATES_IN_ORDER, array(PROJ_SUBMIT_PG_POSTED, PROJ_DELETE))));
```
cpeel: Break this across multiple lines, eg:
```
    $_GET = array_merge($_GET, array(
        'project_manager' => $pguser,
        'state' => array_diff($PROJECT_STATES_IN_ORDER, array(PROJ_SUBMIT_PG_POSTED, PROJ_DELETE))
    ));
```

tools/project_manager/projectmgr.php
```diff
@@ -87,61 +76,50 @@
...
-    );
-}
-
-function create_shortcut_link($text, $url, $show_view="")
+function create_shortcut_link($text, $show_val, $show_view="")
 {
-    if($show_view != $url)
-        return "<a href='{$_SERVER['PHP_SELF']}?show=$url'>$text</a>";
+    if($show_view != $show_val)
+        return "<a href='{$_SERVER['PHP_SELF']}?show=$show_val'>$text</a>";
```
cpeel: Again, we should try to remove `$_SERVER['PHP_SELF']` here. I don't think it's needed.

tools/search.php
```diff
@@ -61,15 +55,13 @@
...
+// show must be search
+$condition = $search_form->get_condition();
 
-// use p_search so it will use saved results if needed
-$search_results = new ProjectSearchResults('p_search', "PS");
-echo "<p><a href='{$_SERVER['PHP_SELF']}?show=blank_search_form'>" . _("Start New Search") . "</a> | " . get_refine_search_link() . " | " . $search_results->get_search_configure_link() . "</p>";
+echo "<h1>", _("Search Results"), "</h1>\n";
 
+$search_results = new ProjectSearchResults("PS");
+echo "<p><a href='{$_SERVER['PHP_SELF']}?show=search_form'>" . _("Start New Search") . "</a> | " . get_refine_search_link() . " | " . get_search_configure_link() . "</p>";
```
cpeel: Ditto removing `$_SERVER['PHP_SELF']`

pinc/ProjectSearchForm.inc
```diff
@@ -32,9 +33,9 @@ class ProjectSearchWidget
 
         if ( $this->type == 'text' )
         {
-            if ( isset($search_data[$this->id]) )
+            if ( isset($_GET[$this->id]) )
             {
-                $value_attr = "value='" . attr_safe($search_data[$this->id]) . "'";
+                $value_attr = "value='" . attr_safe($_GET[$this->id]) . "'";
```
cpeel: This file should `include_once($relPath."misc.inc");` for `attr_safe()` and now `array_get()` and not rely on something else bringing it in.

pinc/ProjectSearchForm.inc
```diff
@@ -334,32 +341,29 @@ class ProjectSearchForm
 
     public function render()
     {
-        if(isset($_SESSION['search_data']))
-            $search_data = $_SESSION['search_data'];
-        else
-            $search_data = array(); // empty array to avoid problems
         echo "<p>" . _("Search for projects matching the following criteria:")."</p>\n";
-        // query data can be too big for GET so use POST
+        // query data can be too big for GET
```
cpeel: We should either add more information to this comment or remove it.

pinc/ProjectSearchForm.inc
```diff
@@ -334,32 +341,29 @@ class ProjectSearchForm
...
         echo "</table>$table_header";
         while($widget_index < $widget_count)
-            $this->widgets[$widget_index++]->echo_search_item($search_data);
+            $this->widgets[$widget_index++]->echo_search_item();
         echo "
             </table>
             <div class='center-align' style='clear: both;'></div>
             <input type='submit' value='", attr_safe(_("Search")), "'>
-            <input type='button' onclick=\"window.location.assign('{$_SERVER['PHP_SELF']}?show=blank_search_form');\" value='", attr_safe(_("Clear form")), "'>
+            <input type='button' onclick=\"window.location.assign('{$_SERVER['PHP_SELF']}?show=search_form');\" value='", attr_safe(_("Clear form")), "'>
```
cpeel: Ditto comments below about not using `$_SERVER['PHP_SELF']`

70ray:
I've tested the cases where $_SERVER['PHP_SELF'] was removed to confirm they all work correctly.

PR #62, opened 2018-11-26
Make User->load() private

cpeel:
To reduce the possibility of bugs, make User->load() private. This ensures that an object can only be loaded by the constructor (or the two new static functions), and prevent the case where a developer
has created a loaded object and then loaded different data into part, but not all, of it.

This mirrors what the new NonactivatedUser class does.

### DIFF
--- a/pinc/User.inc
+++ b/pinc/User.inc
@@ -63,7 +63,7 @@ class User
         return isset($this->table_row[$name]);
     }
 
-    public function load($field, $value, $strict=TRUE)
+    private function load($field, $value, $strict=TRUE)
     {
         if(in_array($field, $this->string_fields))
         {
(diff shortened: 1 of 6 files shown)


pinc/User.inc
```diff
@@ -170,6 +170,26 @@ class User
         log_access_change($this->username, $requester, $activity, $action_type);
     }
 
+    // static functions
+
+    // Load a User record by ID
+    // e.g. $user = User::load_from_id($id);
+    public static function load_from_id($id)
```
jmdyck: The code should (and mostly does) clearly distinguish between (uses of) `users.id` and `users.u_id`. So I think this method should be called `load_from_u_id` or `load_from_uid`.

cpeel: I was on the fence about this. I really hate that `users.id` is named as it is because it's only used during activation. If that column were to be renamed `users.activation_id` would that satisfy your concern or are you more concerned that the name doesn't match the database column? Ideally `users.u_id` would be renamed `users.id` too but that would be a very intrusive change that I'm just not ready for.

jmdyck: > If [`users.id`] were to be renamed `users.activation_id` would that satisfy your concern

Hm. It occurs to me that the name `activation_id` is somewhat ambiguous: you're thinking of it as the id that's *supplied* with an activation request, but it could also be taken as the id that's *generated* by an activation request (i.e. `u_id`). Maybe `registration_id` instead?

Or maybe don't call it an "id" at all: `registration_token`?

Or I suppose `activation_token` would be safe from the above ambiguity.

Or `activation_key` in the sense that it unlocks the activation? (Although it would be a little odd to have a `users` column named "key" that wasn't a table key.)

> or are you more concerned that the name doesn't match the database column?

Yeah.

> Ideally users.u_id would be renamed users.id too but that would be a very intrusive change that I'm just not ready for.

If you replace `load_from_id` with `load_from_uid` (say) in this PR, then changing it *back* if/when `u_id` is renamed to `id` would be only a small increment to that work.

cpeel: I like both `registration_token` and `load_from_uid` as a transition to (hopefully, eventually) renaming both columns.

cpeel: I just pushed up a fixup commit with these changes. If it looks good I'll rebase it before merging.

PR #63, opened 2018-11-26
Use number input type for days/sample_limit

cpeel:
For a better UX, let the browser enforce a non-negative numeric value for the days and sample_limit variables by using the HTML5 <input type="number">.

[Task 1725](https://www.pgdp.net/c/tasks.php?task_id=1725)


mlazaric:
I like this solution, but it fails on: 

- 100.0 - https://www.pgdp.org/~cpeel/c/tools/proofers/review_work.php?work_round_id=P1&review_round_id=P2&days=100.0&sample_limit=0
- 100e2 - https://www.pgdp.org/~cpeel/c/tools/proofers/review_work.php?work_round_id=P1&review_round_id=P2&days=100e2&sample_limit=0

tools/proofers/review_work.php
```diff
@@ -82,11 +82,11 @@
 echo  "</tr>";
 echo  "<tr>";
 echo   "<td>" . _("Max days since last save") . "</td>";
-echo   "<td><input name='days' type='text' size='4' value='$days'></td>";
+echo   "<td><input name='days' type='number' min='0' value='$days'></td>";
```
mlazaric: Is there any reason someone might put 0 as `days`? Maybe `min='1'` would be more appropriate for `days`? 

Should we also add a maximum value constraint?

cpeel:
> I like this solution, but it fails on:
> 
> * 100.0 - https://www.pgdp.org/~cpeel/c/tools/proofers/review_work.php?work_round_id=P1&review_round_id=P2&days=100.0&sample_limit=0
> * 100e2 - https://www.pgdp.org/~cpeel/c/tools/proofers/review_work.php?work_round_id=P1&review_round_id=P2&days=100e2&sample_limit=0

That's a fair point. I'm more interested in making sure the first one works gracefully, if the second one throw the exception I'm not so worried about it. I could change the validation to be `get_float_param()` and then cast them both to an integer -- that would ensure that the most common use-cases are handled gracefully (a user inputs a valid non-exponential number).

tools/proofers/review_work.php
cpeel: I'm not sure if there's a great use-case for zero as a valid option, but it works (and with my luck someone uses it that way :)

Ditto maximum value - both could be very large and still work so I'm not sure there's a good reason to limit them.

mlazaric:
I think you've covered all the _reasonably_ valid options. :) 

Another thing worth discussing: in the very unlikely case someone accidentally or intentionally inputs an invalid value, should we show them a user friendly message like this:

image

Or just output the exception message as we do currently?

image

cpeel:
> I'm just curious -- why disallow fractional days? There are many areas in the code where we do allow them.

We could allow fractional days (step=0.1, for instance) but the downside is that for browsers that offer -/+ buttons for the input would only increment in 0.1 steps, which would be unintuitive for most users. I'm not certain most users really grok tenths of a day (on top of the fact that it uses the `current_time - $days`, not `midnight - $days`). I could be convinced that fractional days are useful here however.

cpeel:
> I think you've covered all the _reasonably_ valid options. :)
> 
> Another thing worth discussing: in the very unlikely case someone accidentally or intentionally inputs an invalid value, should we show them a user friendly message like this:
> 
> image
> 
> Or just output the exception message as we do currently?
> 
> image

I mean, we *could* but I have zero sympathy from people updating URLs directly. And I'm lazy.

PR #64, opened 2018-12-01
Fix bug related to subscribing to `sr_reported`

mlazaric:
#### 1. Always prints `sr_reported` as an unsubscribed event
image

#### 2. Allows non PP users to subscribe to `sr_reported` by changing the `name` parameter of an input to `sr_reported`
Before changing `name`:
image
After changing `name`:
image

Submitting results in this:
image

### Description of the solution
`set_project_event_subs.php` should not allow non PP users to subscribe to `sr_reported`, nor should it print a message saying the user has been unsubscribed.

### Related task:
  -  https://www.pgdp.net/c/tasks.php?action=show&task_id=1726

### Available for testing: 
 - https://www.pgdp.org/~mlazaric/c/


tools/set_project_event_subs.php
```diff
@@ -7,12 +7,19 @@
...
 $projectid = validate_projectID('projectid', @$_POST['projectid']);
+$project = new Project($projectid);
 $return_uri = urldecode($_POST['return_uri']);
 
 $subs = array();
 $unsubs = array();
 foreach ( $subscribable_project_events as $event => $label )
 {
+    // Disallow subscribing to SR report posted event for anyone who is not a PP
+    if ($event == 'sr_reported' && !$project->PPer_is_user($pguser))
```
mlazaric: It's worth noting that this is duplicated code, the same code is used in `project.php` ([here](https://github.com/DistributedProofreaders/dproofreaders/blob/master/project.php#L1142-L1145)), so it might be worthwhile to turn that condition into a function `can_user_subscribe_to_event($username, $projectid, $event)` to future proof it.

srjfoo: For the comment, it's more correct to say "for anyone who is not the PP for the project"

cpeel: @mlazaric True, although I'm not going to block this merging because of that. I will merge this unless you want to do another commit to abstract that logic out into a function.

tools/set_project_event_subs.php
```diff
@@ -7,12 +7,18 @@
...
 
 $projectid = validate_projectID('projectid', @$_POST['projectid']);
+$project = new Project($projectid);
 $return_uri = urldecode($_POST['return_uri']);
 
 $subs = array();
 $unsubs = array();
 foreach ( $subscribable_project_events as $event => $label )
 {
+    if (!can_user_subscribe_to_project_event( $pguser, $projectid, $event ))
```
mlazaric: @srjfoo raised a valid concern about this. If someone is a PP for a project, subscribes to this event and then returns the project to the pool (stops being the PP), they cannot unsubscribe themselves from the event. 

We could _silently_ unsubscribe them from the event (add `unsubscribe_user_from_project_event` before the `continue`), however I am not sure whether that is the best way to handle it.

cpeel: I think silently unsubscribing them from the event makes sense, but we want that to be in the code that returns the project to the PP pool.

srjfoo: I agree -- I just hadn't mentioned it yet. The best time to do it is when the project is returned to the pool.

pinc/ProjectTransition.inc
```diff
@@ -315,6 +315,16 @@ class ProjectTransition
...
+        //     In case of the Post Proofer returning the project to Post Proofing pool,
+        // we want to unsubscribe them from the 'sr_reported' event because they are
+        // no longer the Post Proofer for that project and the event is meant for
+        // Post Proofers.
+        if ($this->from_state == PROJ_POST_FIRST_CHECKED_OUT &&
+            $this->to_state == PROJ_POST_FIRST_AVAILABLE)
+        {
+            unsubscribe_user_from_project_event( $project->checkedoutby, $projectid, 'sr_reported' );
+        }
+
```
srjfoo: Looks good to me.

PR #65, opened 2018-12-11
Names of anonymous users now visible on mentors page

70ray:
This enables mentors to identify different users, see task  #1729
This change is in accordance with the Privacy Statement


70ray:
In the tables listing proofreaders the anonymous users do not link to the statistics page (which would not show any information about them).

PR #66, opened 2018-12-12
change save_and_preview to preview without save in edit project

70ray:


cpeel:
The downside of removing "Save and Preview" is that we no longer have the ability to save the project and remain on the edit page as both "Save" buttons now jump us to another page instead. The "save and go to another page" is not a standard user experience and I think a better overall design would be to have `[Preview]` `[Save]` `[Quit]` buttons -- all of which do what is on the label -- and make the project ID in the form link to the project page (the PM page is already one-click accessible from the navbar).

I suspect any of these changes, including what you have already, will require discussion with the PMs because of how they change the saving-a-project experience.

70ray:
Seems like a good plan

70ray:
Seems like a good plan but looking at it again this page should perhaps not have a header with a nav bar since this makes it easy to accidentaly lose your work by clicking on a link, but be more like a 'modal dialog' where the only way out is to explicitly save or cancel.
This page is also used for creating a new project when you get here from the calalogue search page so having butons to save and go to project or PM page is not unreasonable in this case.
We could have yet another button to 'save and continue editing'. In the case of a new project this button could be named 'Create project' which would reload the editing page with the new project id (which didn't exist before).

cpeel:
We have tons of forms that are shown with the navbar and navigating away from a page without saving changes is a "normal" part of the web, so I'm not worried about that. We don't use modal dialogs anywhere else, so that would be an outlier and not something I'd like to see started here without an intent to implement that site-wide.

I'm loath to add yet another button to this form -- it already has a few too many IMHO. However, without doing a full UX exercise I think the simplest thing is going to be to split out the `Save` and `Preview` functionality into separate buttons, making the full set:
```
[Preview] [Save] [Save and Go To PM Page] [Save and Go To Project] [Quit Without Saving]
```

70ray:
That's similar to what I was thinking (without the change to navbar). I've made the buttons just like what you propose.

cpeel:
Thank you for fixing this!

PR #67, opened 2018-12-13
Properly escape values in dc.xml

cpeel:
We should properly escape values in the dc.xml file so that it is a valid XML file.  Task 1730

### DIFF
--- a/pinc/Project.inc
+++ b/pinc/Project.inc
@@ -1,7 +1,7 @@
 <?php
 include_once($relPath.'user_is.inc');
 include_once($relPath.'project_states.inc');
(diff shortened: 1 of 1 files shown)


70ray:
It all looks fine to me.

70ray:
Looking at dc.xml files for projects on the test server, there are some odd texts in some fields such as description and date but this is a different issue.

cpeel:
> Looking at dc.xml files for projects on the test server, there are some odd texts in some fields such as description and date but this is a different issue.

Yes, I believe those are due to bad MARC records for those items on the test server.

PR #68, opened 2018-12-15
Remove 'active' column from users table

cpeel:
The `users.active` column is not maintained by the code and thus is never used. Many places in the code query against `users.t_last_activity` to determine when they were last seen. This is fast and doesn't require maintaining another field in the database.

While the `crontab/inactivate_users.php` crontab is being run on PROD and setting `active = 'no'`, there is nothing to change that back to `active = 'yes'` (or the original empty value). This means that `crontab/notify_users.php` is never actually sending emails. Better to just remove these entirely and reimplement based on `users.t_last_activity` if that notification functionality is desired.


jmdyck:
> While the `crontab/inactivate_users.php` crontab is being run on PROD and setting `active = 'no'`, there is nothing to change that back to `active = 'yes'` (or the original empty value). This means that `crontab/notify_users.php` is never actually sending emails.

The first sentence is correct, but the second isn't quite. From the database, it looks as if every account was initialized with `active='yes'` until about 2003-05-08. (Mind you, I don't see this happening in the code, so there was probably some hacking of the db involved.) This is about 10k users, so `notify_users.php` has presumably sent lots of emails over the years, but fewer and fewer lately, as more and more members of that original cohort get their `active` set to `'no'` by `inactivate_users.php`. (In 2018, there were 2 such, and they were presumably sent emails a month prior by `notify_users.php`.) Currently, there are still 26 users with `active='yes'`.

But that's just clarification and historical background, not arguing against the PR at all.

jmdyck:
Note that there are also 3 files in noncvs on pgdp.net with (commented out) references to the `active` column:
- pm_away.php
- pp_old.php
- user_pp.php

When this PR is merged and deployed, it would make sense to remove those comments.

PR #69, opened 2018-12-19
Exclude birthdays for 'anonymous' users

cpeel:
We should only birthdays for users who have their privacy preferences set to public.


srjfoo:
Looks good to me.

PR #71, opened 2018-12-23
Fix missed break in switch syntax for sr_reported

cpeel:
This missed break caused the notification message body for sr_reported events to include sr_complete text as well.


srjfoo:
Looks good to me.

PR #72, opened 2018-12-24
Use type=number inputs for numeric form fields

cpeel:
In addition to the Task Center (which was the request from Task 1733) I scoured the code for places where we could be using type=numeric input fields. These are the ones that I found, fixed, and tested.

### DIFF
--- a/tasks.php
+++ b/tasks.php
@@ -788,16 +788,20 @@ function handle_action_on_a_specified_task()
         }
     }
     elseif ($action == 'add_related_task') {
-        process_related_task($pre_task, 'add', @$_POST['related_task']);
+        $related_task_id = (int)get_float_param($_POST, 'related_task', null, 1, null);
+        process_related_task($pre_task, 'add', $related_task_id);
     }
     elseif ($action == 'remove_related_task') {
-        process_related_task($pre_task, 'remove', @$_POST['related_task']);
+        $related_task_id = (int)get_float_param($_POST, 'related_task', null, 1, null);
+        process_related_task($pre_task, 'remove', $related_task_id);
     }
     elseif ($action == 'add_related_topic') {
-        process_related_topic($pre_task, 'add', @$_POST['related_posting']);
+        $related_posting_topic = (int)get_float_param($_POST, 'related_posting', null, 1, null);
+        process_related_topic($pre_task, 'add', $related_posting_topic);
     }
     elseif ($action == 'remove_related_topic') {
-        process_related_topic($pre_task, 'remove', @$_POST['related_posting']);
+        $related_posting_topic = (int)get_float_param($_POST, 'related_posting', null, 1, null);
+        process_related_topic($pre_task, 'remove', $related_posting_topic);
     }
     elseif ($action == 'add_metoo') {
         global $os_array, $browser_array;
@@ -838,7 +842,7 @@ function process_related_task($pre_task, $action, $related_task_id)
 
     // Validate task_id. It must be an integer >= 1
     $related_task_id = trim($related_task_id);
-    if (empty($related_task_id) || !is_numeric($related_task_id) || $related_task_id < 1) {
+    if (!is_numeric($related_task_id) || $related_task_id < 1) {
         ShowNotification("You must supply a related task ID.", true);
         return;
     }
@@ -884,7 +888,7 @@ function process_related_topic($pre_task, $action, $related_topic_id)
 
     // Validate related_topic_id. It must be an integer >= 1
     $related_topic_id = trim($related_topic_id);
-    if (empty($related_topic_id) || !is_numeric($related_topic_id) || $related_topic_id < 1) {
+    if (!is_numeric($related_topic_id) || $related_topic_id < 1) {
         ShowNotification("You must supply a related topic ID.", true);
         return;
     }
--- a/tasks.php
+++ b/tasks.php
@@ -618,7 +618,7 @@ function handle_action_on_a_specified_task()
     // Default 'action' when a task is specified:
     if (is_null($action)) $action = 'show';
 
-    $task_id = get_integer_param($_REQUEST, 'task_id', null, 1, null);
+    $task_id = (int)get_float_param($_REQUEST, 'task_id', null, 1, null);
 
     // Fetch the state of the specified task
     // before any requested changes.
@@ -968,7 +968,8 @@ td.taskproperty    { width: 40%; font-weight: bold; }
 td.taskvalue       { width: 60%; border-bottom: #CCCCCC 1px solid; }
 
 select.taskselect  { font-size: small; color:#03008F; background-color:#EEF7FF; }
-input[type="text"] { font-size: small; border:1px solid #000000; margin:2px; padding:0px; background-color:#EEF7FF; }
+input[type="number"],
+input[type="text"] { font-size: small; border:1px solid #000000; margin:2px; padding:0px; background-color:#EEF7FF; width: 4em; }
 input[type="button"],
 input[type="submit"] { font-size: small; color:#FFFFFF; font-weight:bold; border:1px ridge #000000; margin:2px; padding: 0px 5px; background-color:#838AB5; }
 input[type="button"]:disabled,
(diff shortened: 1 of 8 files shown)


tools/mentors/view_page_text_image.php
```diff
@@ -206,7 +206,7 @@
 <form method="get" action="view_page_text_image.php">
 <input type="hidden" name="projectid" value="<?php echo $projectid; ?>">
 <input type="hidden" name="page" value="<?php echo $page; ?>">
-<input type="text" maxlength="3" name="percent" size="3" value="<?php echo $percent; ?>">%
+<input type="number" name="percent" value="<?php echo $percent; ?>" min="1" max="100">%
```
mlazaric: Might be a good idea to add `required`.

Does not handle `5.0` and `5e0` well but I do not think it is much a concern here.

mlazaric: https://github.com/DistributedProofreaders/dproofreaders/blob/a4f9aa136899dfd0b9ce52519e203be257c1c326/tools/mentors/view_page_text_image.php#L127

`page` could be a `number` as well.

tools/project_manager/displayimage.php
```diff
@@ -57,7 +57,7 @@
 <input type="hidden" name="preload" value="<?php echo $preload; ?>">
 
 <?php echo _("Resize"); ?>:
-<input type="text" maxlength="3" name="percent" size="3" value="<?php echo $percent; ?>">%
+<input type="number" name="percent" min="1" max="100" value="<?php echo $percent; ?>">%
```
mlazaric: Same as the previous one, might be a good idea to add `required`.  Does not handle `5e0` and `5.0` well but I guess that is not too important.

tools/proofers/mktable.php
```diff
@@ -43,8 +43,8 @@
 <body>
 <form method="POST">
 <input type="submit" value="Create">
-<input type="text" name="row" value="<?php echo $row; ?>" size="2"> rows and
-<input type="text" name="col" value="<?php echo $col; ?>" size="2"> columns table
+<input type="number" name="row" value="<?php echo $row; ?>" style='width: 3em;'> rows and
+<input type="number" name="col" value="<?php echo $col; ?>" style='width: 3em;'> columns table
```
mlazaric: `min=1` and `required` for both would make sense. 

Handles `5.0` and `5e0` well.

tools/authors/add.php
```diff
@@ -346,7 +346,7 @@ function echo_date_fields($bd) {
 <tr><td><?php echo _('Year'); ?>:</td><td>
 <input type="radio" name="<?php echo $bd; ?>yearRadio" value="0"<?php echo (_var($bd, 'yearRadio')=='0'?' CHECKED':''); ?>><?php echo _('Unknown'); ?>
 <br><input type="radio" name="<?php echo $bd; ?>yearRadio" value="1"<?php echo (_var($bd, 'yearRadio')=='1'?' CHECKED':''); ?> onClick="this.form.<?php echo $bd; ?>year.focus();" ><?php echo _('As entered'); ?>:
-<input type="text" name="<?php echo $bd; ?>year" size="4" maxlength="4"<?php echo (_var($bd, 'yearRadio')=='1'?' VALUE="'.abs(_var($bd, 'year')).'"':''); ?> onFocus="this.form.<?php echo $bd; ?>yearRadio[1].checked=true;">
+<input type="number" name="<?php echo $bd; ?>year" style='width: 4em;'<?php echo (_var($bd, 'yearRadio')=='1'?' VALUE="'.abs(_var($bd, 'year')).'"':''); ?> onFocus="this.form.<?php echo $bd; ?>yearRadio[1].checked=true;">
```
mlazaric: Can't access this page, but it looks fine, maybe a `required` and `min` would not hurt.

tools/site_admin/manage_special_days.php
```diff
@@ -244,27 +244,27 @@ function show_edit_form()
...
-        $this->_show_edit_row('open_month',_('Open Month'),false,2);
-        $this->_show_edit_row('open_day',_('Open Day'),false,2);
-        $this->_show_edit_row('close_month',_('Close Month'),false,2);
-        $this->_show_edit_row('close_day',_('Close Day'),false,2);
-        $this->_show_edit_row('date_changes',_('Date Changes'),false);
-        $this->_show_edit_row('info_url',_('Info URL'),false);
-        $this->_show_edit_row('image_url',_('Image URL'),false);
+        $this->_show_edit_row('comment', _('Comment'), 'textarea');
+        $this->_show_edit_row('color', _('Color'), 'text', 8);
+        $this->_show_edit_row('open_month', _('Open Month'), 'number', null, 1, 12);
```
mlazaric: Can't access this page, but it looks fine, might be a good idea to add `required` on them.

tools/site_admin/show_common_words_from_project_word_lists.php
```diff
@@ -89,7 +89,7 @@
 
     echo "<tr>";
     echo "<td>" . _("Percentage cutoff:") . "</td>";
-    echo "<td><input type='text' name='cutoff' value='$cutoff' size='4'>% - ";
+    echo "<td><input type='number' style='width: 4em;' name='cutoff' value='$cutoff'>% - ";
```
mlazaric: Can't access this page, might be a good idea to add `min=0` and `required`.

userprefs.php
```diff
@@ -924,7 +924,14 @@ function _show_textfield( $field_name, $current_value, $extras )
 {
     list($size, $rest) = $extras;
     $current_value_esc = attr_safe($current_value);
-    echo "<input type='text' name='$field_name' value='$current_value_esc' size='$size'>$rest";
+    echo "<input type='text' style='width: ${size}em' name='$field_name' value='$current_value_esc'>$rest";
+}
+
+function _show_numberfield( $field_name, $current_value, $extras )
```
mlazaric: https://github.com/DistributedProofreaders/dproofreaders/blob/a4f9aa136899dfd0b9ce52519e203be257c1c326/userprefs.php#L552-L567

These two could be `numberfield` too.

Might be a good idea to add `required` to all of them since it throws `Fatal error: Uncaught InvalidArgumentException` if one of them was not given. 

Might be a good idea to add `min=1` to most (if not all of them) since negative values make no sense in this context. 

Handles `5.0` and `5e0` well.

tasks.php
```diff
@@ -1012,7 +1013,7 @@ function hideSpan(id) { document.getElementById(id).style.display="none"; }
     echo "<form action='$tasks_url' method='get'><input type='hidden' name='action' value='show'>";
     echo "<b><small class='task'>Show Task #</small></b>";
     echo "&nbsp;\n";
-    echo "<input type='text' name='task_id' size='12'>&nbsp;\n";
+    echo "<input type='number' name='task_id' size='12'>&nbsp;\n";
```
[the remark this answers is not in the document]
cpeel: Good call on adding the minimum, done.

I've changed `get_integer_param()` to `(int)get_float_param()` which will handle the decimal ones better. If someone really does want to put in exponential-formatted numbers, they're welcome to the exception :)

I wasn't aware of the `required` attribute! It's new in HTML5. Separate from this PR I'll go through and add that to various forms in the DP code.

tools/mentors/view_page_text_image.php
cpeel: Actually, I don't think it can. Page "numbers" are really strings because they can contain more than just numbers (and also may have leading zeros).

tools/proofers/mktable.php
cpeel: Added `min`.

tools/site_admin/show_common_words_from_project_word_lists.php
cpeel: Added `min=0`

userprefs.php
cpeel: Good call on those two fields that I missed. I've fixed those and added `min=1` to all of them.

tasks.php
```diff
@@ -1672,7 +1673,7 @@ function RelatedTasks($tid)
     echo "<form action='$tasks_url' method='post'>";
     echo "<input type='hidden' name='action' value='add_related_task'>";
     echo "<input type='hidden' name='task_id' value='$tid'>";
-    echo "<input type='text' name='related_task' size='30'>&nbsp;&nbsp;";
+    echo "<input type='number' name='related_task' size='30'>&nbsp;&nbsp;";
```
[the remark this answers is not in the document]
cpeel: Added `min=1` to this. Will address `required` in a separate PR.

I fixed the validation for the related tasks, good catch!

tools/authors/add.php
cpeel: Missed this one in my first round of CR updates. Got it now.

PR #73, opened 2018-12-26
Escape key now quits preview

70ray:
escape is deactivated when leving preview and also when entering
the configure pane and reactivated going back to preview from
configure


cpeel:
This is great Ray, thank you for coding this and sorry it took so long to get it merged.

PR #76, opened 2018-12-29
Pull delete_pages tool from noncvs and standardize tool styles

cpeel:
This pulls `noncvs/delete_pages.php` into the code proper. While here I standardized the output & styles some across the copy/rename/delete tools so they look less odd. They still make liberal use of `<pre/>` tags, but such is the life of an admin.


srjfoo:
Updated scripts tested and work as expected.

PR #77, opened 2018-12-31
Fix two small visual bugs on tasks page

mlazaric:
### Before:

image

 * https://www.pgdp.net/c/tasks.php?search_text=abcdefg

### After:

image

 * https://www.pgdp.org/~mlazaric/c/tasks.php?search_text=abcdefg

### Before:

image

 * https://www.pgdp.org/c/tasks.php

### After:

image

 * https://www.pgdp.org/~mlazaric/c/tasks.php

### DIFF
--- a/tasks.php
+++ b/tasks.php
@@ -1140,7 +1140,7 @@ function select_and_list_tasks($sql_condition)
         }
     }
     else {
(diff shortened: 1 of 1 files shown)


cpeel:
Good catch and fix!

PR #79, opened 2019-01-28
Show sever time on pm page

70ray:
Update every minute from server

### DIFF
--- a/scripts/time.js
+++ b/scripts/time.js
@@ -1,16 +1,42 @@
-/*global $ apiUrl*/
+/*global $ apiUrl setInterval*/
 $(function () {
     "use strict";
     var timeSpan = $("#server-time");
+    var day = "";
+    var msec0 = Date.now();
 
-    function showTime(data) {
-        timeSpan.text(data.daytime);
+    function pad2(n) {
+        return ("0" + n).slice(-2);
     }
 
-    function updateClock() {
-        $.getJSON(apiUrl, {'q': 'v1/serverdaytime'}, showTime);
+    function showTime(hour, minute) {
+        timeSpan.text(day + " " + pad2(hour) + ":" + pad2(minute));
     }
 
-    setInterval(updateClock, 60000);
-    updateClock();
+    function resetTime(data) {
+        var hour = data.hour;
+        var minute = data.minute;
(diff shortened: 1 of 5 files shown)


70ray:
This implements task 1719

cpeel:
This seems like overkill. I would think it simpler for the page to load with the server's timezone pre-populated in the JS code somewhere and then have JS calculate the time given the difference in timezones and render it on the page without calls back to the server.

```js
<script src="https://cdnjs.cloudflare.com/ajax/libs/moment.js/2.24.0/moment.js"></script>
<script src="https://cdnjs.cloudflare.com/ajax/libs/moment-timezone/0.5.23/moment-timezone-with-data.js"></script>
<script>
// page is loaded with this populated from PHP function date_default_timezone_get();
let serverTimezone = "UTC";

// then in the timer, calculate and output the current time
let now = moment();
let timeString = moment.tz(serverTimezone).format("ddd HH:mm");
console.log(timeString);
</script>
```

70ray:
Yes, it does seem unnecessary to get data from the server that you could easily calculate. I'll make a new branch to do it differently.

scripts/time.js
```diff
@@ -1,16 +1,42 @@
...
-    setInterval(updateClock, 60000);
-    updateClock();
+    function resetTime(data) {
+        var hour = data.hour;
+        var minute = data.minute;
+        day = data.day;
+        // milliseconds from midnight
+        var msecs = 60000 * (parseInt(minute) + 60 * parseInt(hour));
+        // milliseconds from epoch to previous midnight
+        msec0 = Date.now() - msecs;
```
cpeel: Please don't try to do date math manually. Date math is hard and we should be using libraries that handle it and all of the edge cases for us. This won't account for daylight saving time changes.

The moment-based coding snippit from my prior comment should be all that you need to implement this without ever having to make an API call. It does rely on the client's local time & timezone being correct, but that is very reasonable. If that's not the level of precision that you are looking for, the server could load the page with the current time and the client could use moment to calculate a difference between it and the client time on page lod and use that, via moment, to calculate the time differences.

70ray:
Thanks Casey, I've made a new branch.

PR #81, opened 2019-01-31
Show updating day and time on PM page

70ray:
implements task 1719


70ray:
Although the localisation works for currently implemented DP locales we should have a more generic procedure which cannot break in the future, i.e. try to load a file that does not exist.

70ray:
Do we want to link to external sites or put the js files on our site?
In either case perhaps there should be some documented procedure for updating to the latest versions.

scripts/time.js
```diff
@@ -0,0 +1,13 @@
+/*global $ setInterval moment serverTimezone*/
```
cpeel: As-is this code is intimately tied to the PM page so I would prefer it to be next to the page that uses it (`tools/project_manager/projectmgr.js`).

tools/project_manager/projectmgr.php
```diff
@@ -50,8 +50,25 @@
...
+$server_timezone = date_default_timezone_get();
+$lang_code = short_lang_code();
+$js_files = [
+    "$code_url/tools/dropdown.js",
+    "$code_url/scripts/time.js",
+    "https://cdnjs.cloudflare.com/ajax/libs/moment.js/2.24.0/moment.js",
+    "https://cdnjs.cloudflare.com/ajax/libs/moment-timezone/0.5.23/moment-timezone-with-data.js"
+];
+
+if($lang_code != "en")
```
cpeel: I assume(?) that if there isn't a locale file available that moment will fall back to English, which is great. That at least helps buffer the case where a DP language will be added that moment doesn't support.

scripts/time.js
70ray: Done this.

PR #82, opened 2019-02-05
Add 'required' attribute to HTML form elements

cpeel:
HTML5 supports the new 'required' attribute on HTML form elements that are required to submit the form. The browser won't allow the page to be submitted until the element is filled in. This doesn't negate the necessity for server-side validation, but it does provide a nicer user experience.

### DIFF
--- a/tasks.php
+++ b/tasks.php
@@ -1018,7 +1018,7 @@ EOS;
     echo "<form action='$tasks_url' method='get'><input type='hidden' name='action' value='show'>";
     echo "<b><small class='task'>Show Task #</small></b>";
     echo "&nbsp;\n";
-    echo "<input type='number' name='task_id' min='1'>&nbsp;\n";
+    echo "<input type='number' name='task_id' min='1' required>&nbsp;\n";
     echo "<input type='submit' value='Go!'>\n";
     echo "</form>";
     echo "</td></tr></table><br>\n";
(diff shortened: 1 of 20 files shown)


tasks.php
```diff
@@ -1177,7 +1177,7 @@ function TaskForm($task)
     echo "<td colspan='2'>";
     echo "<b>" . property_get_label('task_summary', FALSE) . "&nbsp;</b>";
     echo "&nbsp;&nbsp;";
-    echo "<input type='text' name='task_summary' value=\"$task_summary_enc\" size='60' maxlength='80'>";
+    echo "<input type='text' name='task_summary' value=\"$task_summary_enc\" size='60' maxlength='80' required>";
```
mlazaric: Might be a good idea to make `task_details` required as well, since not providing it shows this:
 
image

tools/mentors/view_page_text_image.php
```diff
@@ -115,7 +115,7 @@
     echo "<form method='get' action='view_page_text_image.php' target='_top'>\n";
     if(!$project) {
         echo _("Project ID") . ":&nbsp;";
-        echo "<input type='text' maxlength='25' name='projectid' size='25' value='" . attr_safe($projectid) . "'> \n";
+        echo "<input type='text' maxlength='25' name='projectid' size='25' value='" . attr_safe($projectid) . "' required> \n";
```
mlazaric: `percent` could also be required, since passing an empty value shows this:

image

https://github.com/DistributedProofreaders/dproofreaders/blob/7b774cfbe8997606fa9efba979bfd3156844a5da/tools/mentors/view_page_text_image.php#L209

cpeel:
Yes, I should have done `userprefs.php` but I was lazy. 😄 I've fixed that now.

srjfoo:
Looks good to me, too (sorry to be slow reacting).

userprefs.php
```diff
@@ -268,7 +268,7 @@ function echo_general_tab() {
         _('E-mail'), 'email', 'email',
```
mlazaric: I have just noticed that we do not check if `email` is actually a valid email address, for example this is considered valid:

image

Maybe we should at least use [`<input type="email" ...>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/email)?

cpeel: Oh nice! I didn't know about that either. I've added that for userprefs and for account registration.

PR #83, opened 2019-02-06
Remove code for opening the project page in an existing window

70ray:


cpeel:
Your change does more than remove the code for opening the project in an existing, named window. It also removes the code that opens a new proofing window of a specific size. Personally, I think that's a better solution in today's tab + window world, but this is a big enough change that it probably needs a wider community discussion.

I'd love to rip out `pinc/resolution.inc` and the user prefs that use it, so I'm 100% behind this change.

70ray:
Perhaps we could just make it work like it does now except for re-using an old window which shouldn't be too controversial.

cpeel:
> Perhaps we could just make it work like it does now except for re-using an old window which shouldn't be too controversial.

Agreed. I think if you just rip out the `$window_name_base` nonsense and just pass `_blank` to `window.open` that might work.

70ray:
New revised pull request made

PR #84, opened 2019-02-09
Always open link to project page in new window

70ray:
rather than reusing one opened earlier (if this option is enabled)
(It would be possible to remove the prep... function and
incorporate it into the get_onclick... function but means more
changes)

### DIFF
--- a/pinc/js_newwin.inc
+++ b/pinc/js_newwin.inc
@@ -10,6 +10,12 @@ function prep_for_links_to_project_pages()
 {
     global $userP, $i_resolutions;
 
+    // Ensure that the guts of this function are only done once,
+    // because don't want to define newProofWin more than once.
+    static $this_func_has_been_called = FALSE;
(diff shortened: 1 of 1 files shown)


pinc/js_newwin.inc
```diff
@@ -8,32 +8,20 @@ include_once($relPath.'resolution.inc');
 
 function prep_for_links_to_project_pages()
 {
-    global $userP, $i_resolutions, $code_url;
-
-    // Ensure that the guts of this function are only done once,
```
cpeel: We still need this logic in here to prevent the function from being called twice and two instances of this javascript function being included in the page.

Alternatively we could make this function return the javascript and put that in the js_data part of the `output_header()` call, but that would be a more intrusive change and I'm fine if we just put the don't-duplicate-output back in.

70ray: Done this.

PR #85, opened 2019-02-11
Multi anchor

70ray:
Multiple anchors for the same footnote is not itself an issue but could indicate a misread anchor.


70ray:
This has now been tested by srjfoo and cmspence.

srjfoo:
Worked fine for me, and confirmed by cmspence.

PR #87, opened 2019-02-13
Search results PG etext number now links to PG book

70ray:


cpeel:
This is [Task 1738](https://www.pgdp.net/c/tasks.php?action=show&task_id=1738)

PR #89, opened 2019-02-18
Allow users to zoom in more than 100%

cpeel:
f43d4da63 enforced maximum image zoom levels to 100%, but users need to be able to zoom in more than 100%.


jmdyck:
Yup, those are the changes I hacked into the live code this morning, and those are the only occurrences of `max=` in the code.

PR #90, opened 2019-02-19
Proof check 1a

70ray:
Rebased version of proof_check_1 with check for all html entities


70ray:
Perhaps better to do something like this at server.

PR #91, opened 2019-02-20
Val proj

70ray:
The motivation for this is to avoid duplicating the projectID regex, and also to move any text output associated with handling errors up to the top level in order to facilitate eventual implementation of an api.

### DIFF
--- a/pinc/Project.inc
+++ b/pinc/Project.inc
@@ -25,7 +25,7 @@ class Project
             $row = mysqli_fetch_assoc($res);
             if (!$row)
             {
-                throw new NonexistentProjectException("no project with projectid='$arg'");
+                throw new NonexistentProjectException(sprintf(_("There is no project with projectid '%s'"), html_safe($arg)));
             }
             $row['t_retrieved'] = time();
         }
--- a/pinc/Project.inc
+++ b/pinc/Project.inc
@@ -26,7 +26,7 @@ class Project
             $row = mysqli_fetch_assoc($res);
             if (!$row)
             {
-                throw new NonexistentProjectException(sprintf(_("There is no project with projectid '%s'"), html_safe($arg)));
+                throw new NonexistentProjectException(sprintf(_("There is no project with projectid '%s'"), $arg));
             }
             $row['t_retrieved'] = time();
         }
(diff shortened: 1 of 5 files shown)


pinc/Project.inc
```diff
@@ -768,14 +769,30 @@ function does_project_page_table_exist($projectid)
 // long form (projectID####) or just the project number (####).
 // If it does, it returns it, otherwise dies with a warning.
 function validate_projectID($param_name, $value, $allownull = false)
+{
+    try
```
cpeel: You don't need the try/catch block at all, just make the call to the file that will throw the exception. We have an exception handler that will output the error message. See `pinc/base.inc`

pinc/Project.inc
```diff
@@ -768,14 +769,30 @@ function does_project_page_table_exist($projectid)
...
+    {
+        die($exception->getMessage());
+    }
+}
+
+// Checks if $value has the form of a valid project ID, either in its
+// long form (projectID####) or just the project number (####).
+// If it does, it returns it, otherwise throws an exception.
+// this is intended to replace validate_projectID()
+function ex_validate_projectID($param_name, $value, $allownull = false)
```
cpeel: You don't need a new function, just change the die() to throw an exception and let the exception handler handle printing it out.

project.php
```diff
@@ -46,13 +46,18 @@
 $DEFAULT_DETAIL_LEVEL = 3;
 
 // Validate all the input
-$projectid      = validate_projectID('id', @$_GET['id']);
-$expected_state = get_enumerated_param($_GET, 'expected_state', null, $PROJECT_STATES_IN_ORDER, true);
-$detail_level   = get_integer_param($_GET, 'detail_level', $DEFAULT_DETAIL_LEVEL, $MIN_DETAIL_LEVEL, $MAX_DETAIL_LEVEL);
-
-// -----------------------------------------------------------------------------
+try
```
cpeel: You don't need the try/catch block, just let the exception kill the process.

70ray:
Thanks Casey, I've made these changes. This has simplified the code.

cpeel:
Code-wise this looks really great - thank you for doing it!

@srjfoo - can you please do some validation testing on the code that has been changed?

srjfoo:
Tested, and seems to be working as expected.

PR #95, opened 2019-03-10
Pm hold

70ray:
Enable current hold to be removed on PM page


pinc/ProjectSearchResults.inc
```diff
@@ -384,8 +384,25 @@ class HoldColumn extends Column
...
-            echo "&#x2714;"; // heavy check mark
+        {
+            if($project->can_be_managed_by_current_user)
+            {
+                echo "
+                    <form method='post' action='$code_url/tools/remove_project_hold.php'>
+                    <input type='hidden' name='projectid' value='$project->projectid'>
+                    <input type='hidden' name='curr_state' value='$project->state'>
+                    <input type='hidden' name='return_uri' value='", urlencode($_SERVER['REQUEST_URI']), "'>
+                    <button type='submit' title='", _("Click to remove hold"), "'>", _("Remove hold"), "</button>
```
cpeel: `_("Click to remove hold")` needs to be wrapped in `attr_safe()`

pinc/project_holds.inc
```diff
@@ -0,0 +1,46 @@
...
+    }
+    $sql = "
+        INSERT INTO project_holds
+        VALUES $values
+    ";
+    try_query($sql);
+    log_project_event( $projectid, $pguser, 'add_holds', join($states, ' ') );
+}
+
+function try_query($sql)
```
cpeel: Please don't make yet another function to wrap mysqli() calls. If you want to have mysqli throw an exception directly, consider using `mysqli_report(MYSQLI_REPORT_ERROR | MYSQLI_REPORT_STRICT)` before the call in the function instead.

pinc/project_holds.inc
```diff
@@ -0,0 +1,46 @@
+<?php
+
+include_once($relPath.'project_events.inc'); // log_project_event
+include_once($relPath.'misc.inc'); // surround_and_join
+
+function remove_holds($projectid, $states)
```
cpeel: Consider putting these functions in the Project class instead.

70ray:
I've made these changes. Shall I delete this branch and push a rebased version?

pinc/Project.inc
```diff
@@ -465,6 +466,38 @@ EOS;
         return $hold_states;
     }
 
+    public function add_holds($states)
+    {
+        global $pguser;
+
+        $values = '';
```
cpeel: This is a bit cumbersome, consider using an array:
```
$values = [];
foreach ($states as $state)
{
    $values[] = "('$this->projectid', '$state')";
}
$values = join(', ', $values);
```

tools/remove_project_hold.php
```diff
@@ -0,0 +1,28 @@
...
+include_once($relPath.'metarefresh.inc'); // metarefresh()
+
+require_login();
+
+$projectid = validate_projectID('projectid', @$_POST['projectid']);
+
+$project = new Project($projectid);
+if (!$project->can_be_managed_by_current_user)
+{
+    echo "<p>", _('You are not authorized to manage this project.'), "</p>\n";
```
cpeel: I think in most other places in the code we just use die() instead:
```
die(_('You are not authorized to manage this project.'));
```
We don't actually open an HTML page so I don't think we need to bother with the paragraph tags.

70ray:
I've made these changes. Shall I delete this branch and push a rebased version?

70ray:
I've made these changes. Shall I delete this branch and push a rebased version?

cpeel:
You can rebase into this branch if you want, or I can just squash the commits into one with the merge -- either works for me.

@srjfoo, can you please validate these changes?

70ray:
@srjfoo Latest code is now in pm sandbox https://www.pgdp.org/~rp31/pm/tools/project_manager/projectmgr.php

@cpeel If its ok for you I'll leave it like it is.

cpeel:
> @cpeel If its ok for you I'll leave it like it is.
Totally fine to leave as-is. Thanks!

srjfoo:
Apologies for being so slow -- works fine for me.

tools/remove_project_hold.php
```diff
@@ -0,0 +1,27 @@
...
+{
+    die(_('You are not authorized to manage this project.'));
+}
+
+$return_uri = urldecode($_POST['return_uri']);
+$states = array($_POST['curr_state']);
+
+$project->remove_holds($states);
+$title = _("Removing hold");
+metarefresh(1, $return_uri, $title, $title);
```
cpeel: One last thought - we should immediately redirect the user and not have a 1 second delay. This will be a better user experience. So lets change the 1 second to 0 seconds.

pinc/Project.inc
```diff
@@ -465,6 +466,39 @@ EOS;
...
+        global $pguser;
+
+        $values = [];
+        foreach ($states as $state)
+        {
+            $values[] = "('$this->projectid', '$state')";
+        }
+        $values = join(', ', $values);
+
+        mysqli_report(MYSQLI_REPORT_ERROR | MYSQLI_REPORT_STRICT);
```
cpeel: Based on our conversation in the other MR, until we vet the code and understand if subsequent `mysqli` calls can handle exceptions being thrown on failures, we should stick with the current convention of `mysqli_query(...) or die("error")` instead of using this to make errors turn into raised exceptions.

I think what you have is the direction we want to go, but I want it to be in `base.inc` for all calls everywhere after we make sure it doesn't break anything.

PR #97, opened 2019-03-17
Genre translated and sorted

70ray:
in PM, project search, round, pool and SR pages
using temporay genre translation table
It was necessary to make the genre table so the sql query could sort correctly. Perhaps it would be better to make permanent tables.


pinc/showavailablebooks.inc
```diff
@@ -384,14 +390,15 @@ function show_project_listing(
...
-        FROM projects
+        FROM projects NATURAL JOIN genres
         $optional_join_clause
         WHERE
             $filtered_project_selector
         ORDER BY
             $presort
             $curr_sql_orderclause
     ";
+    mysqli_report(MYSQLI_REPORT_ERROR | MYSQLI_REPORT_STRICT);
```
cpeel: We should think if there is a compelling reason to not just make this a global setting in `base.inc`.

cpeel:
I vacillated between "OMGosh, this is hideous -- there has to be a better way" to "this is freaking brilliant!" for about 30 minutes. I've talked myself into the latter 😄

pinc/genre_table.inc
```diff
@@ -0,0 +1,28 @@
+<?php
+include_once($relPath.'genres.inc');
+
+function create_genre_table()
```
cpeel: Rename this `create_temporary_genre_translation_table()` and move this into `pinc/genres.inc` and call it right before we need it (in `_search()` and `_load_project_filter_field_values()`, etc). The function can check if the table exists and if not, it creates it with the logic you have here. This way we try to keep the table creation closer to the code that is going to use it.

pinc/genre_table.inc
```diff
@@ -0,0 +1,28 @@
...
+{
+    $sql = "
+    CREATE TEMPORARY TABLE genres
```
cpeel: Consider renaming the table `genre_translations`.

pinc/showavailablebooks.inc
70ray: The manual says PHP versions since 5.3.4 and 5.2.15 "Changing the reporting mode is now be per-request, rather than per-process." so we mght not be able to do that.

pinc/genre_table.inc
70ray: I've also increased genre size to VARCHAR(50) because that is what it is in the projects table.
We could make a Genre class instead of using the global variable $GENRES. But $GENRES is used in a noncvs file so I won't do it now. (How would we synchronize such a change? presumably move the file out of noncvs first?)

70ray: I couldn't check if the table exists elegantly:
It looks like CREATE TABLE IF NOT EXISTS still creates a table but doesn't issue a warning,
SHOW TABLES and the INFORMATION_SCHEMA TABLES Table does not list temporary tables,
INFORMATION_SCHEMA INNODB_TEMP_TABLE_INFO Table needs PROCESS privilege.
In the end just tried to access the table and caught exception if doesn't exist.

pinc/showavailablebooks.inc
cpeel: From a per-request vs per-process perspective either we can do this here and in base.inc or we can't do it at all. The question is really do we want all of our `mysqli()` calls to have this behavior or just the ones that happen after this line in the code?

70ray: I think I completely misunderstood what "per-request, rather than per-process" means. I still don't know what it means but if putting the mysqli_report line in base.inc would affect all subsequent mysqli_querys I'm all for it.

70ray: I've put it in base.inc. It works.

cpeel: Right, I know it will work there for what you are doing, the question is: will this inadvertently mess up anything else. In other words: are there places in the code that we want to do something different than raise/die on a failed mysqli call. I think the answer is probably yes and we need to wrap those in a try/except before we put this in base.inc.

70ray: Agreed.

70ray:
I've pushed a new commit with suggested changes. Travis CI is failing but seems to be outside my control.

70ray:
genres.inc contains a mysqli_report(MYSQLI_REPORT_ERROR | MYSQLI_REPORT_STRICT);
(I couldn't find a better way to check if the temp. table exists as above)
It is difficult to avoid this. If we used the false result of query to detect table doesn't exist it would break if we put mysqli_report() in base.inc.
I've checked that none of the mysqli_query() calls which can happen on the same visit to the server can break. (The only way I can see they could break is if they depend on a false result to do something).
Perhaps you can find a better solution.

cpeel:
We can have the code support both with and without exceptions:
```php
function maybe_create_temporary_genre_translation_table()
{
    $db_link = DPDatabase::get_connection();

    $sql = "
    CREATE TEMPORARY TABLE genre_translations
    (
        genre VARCHAR(50) NOT NULL,
        trans_genre VARCHAR(50) NOT NULL,
        PRIMARY KEY (genre)
    )
    ";

    try {
        $result = mysqli_query($db_link, $sql);
    } catch(mysqli_sql_exception $e) {
        // in case we throw exceptions in the future
        $result = false;
    }

    // if the create table succeeded, populate the table
    // if it didn't, we assume the table is already created
    if($result)
    {
        foreach($GENRES as $key => $value)
        {   
            $key = mysqli_real_escape_string($db_link, $key);
            $value = mysqli_real_escape_string($db_link, $value);
            $sql = "INSERT INTO genre_translations SET genre = '$key', trans_genre = '$value'";
            mysqli_query($db_link, $sql);
        }
    }
}
```

70ray:
Thanks Casey, I've used this code.

cpeel:
Code-wise this looks great, thanks @70ray!

@srjfoo, would you mind doing a final validation check on this and we'll get it merged in?

srjfoo:
Tested, and looks good to me.

PR #98, opened 2019-03-23
Move genre translations into a function

cpeel:
Rather than putting genres into a global variable upon file load, lets move them into a function that returns the list.

### DIFF
--- a/pinc/genres.inc
+++ b/pinc/genres.inc
@@ -1,92 +1,102 @@
 <?php
 
-$GENRES = array(
-    'Other'              => _('Other'),
-    'Adventure'          => _('Adventure'),
-    'Agriculture'        => _('Agriculture'),
-    'Animals'            => _('Animals'),
-    'Anthropology'       => _('Anthropology'),
-    'Archaeology'        => _('Archaeology'),
-    'Architecture'       => _('Architecture'),
-    'Art'                => _('Art'),
-    'Astronomy'          => _('Astronomy'),
-    'Autobiography'      => _('Autobiography'),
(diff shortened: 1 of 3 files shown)


pinc/genres.inc
```diff
@@ -1,92 +1,95 @@
 <?php
 
-$GENRES = array(
```
70ray: $GENRES is also used in a noncvs file projects_with_odd_values.php. Should we update it?

tools/project_manager/edit_common.inc
```diff
@@ -77,9 +77,9 @@ function language_list($language)
 
 function genre_list($genre)
 {
-    global $GENRES;
+    $genres = load_genre_translation_array();
```
70ray: Perhaps this could also be translated and sorted.

pinc/genres.inc
cpeel: Good eye. Yes, it will need to get updated too. Generally I make a note for myself to do those edits on TEST when the change is merged into master and then into PROD when it gets published.

tools/project_manager/edit_common.inc
cpeel: We could move the sorting-by-value into this function, sure. I don't understand your question about being translated however, can you elaborate please?

70ray: Sorry, forget translate, wasn't thinking clearly.

cpeel: Updated the function to return the list sorted by the translated name.

PR #99, opened 2019-03-23
Subscribe user to task after commenting

cpeel:
When a user comments on a task, subscribe them to it for future comments or changes.

Task 1743


srjfoo:
Looks good to me.

PR #101, opened 2019-03-23
Set minimum PHP version to 7.0

cpeel:
The new minimum PHP version going forward is 7.0. This obviates the need for `undo_all_magic_quotes()` since magic quotes do not function after PHP 5.3.

Tagging all active developers to make sure everyone is aware of this change.


70ray:
I came too late to the php party to know what magic quotes were about. Seems fine to me.

PR #102, opened 2019-03-23
Update tests to work with phpunit >=5.7 and <8

cpeel:
Our phpunit tests were created against phpunit4. phpunit6 introduced an updated, and backwards-incompatiable, namespace which was backported to phpunit5.7 (the last phpunit release to support PHP 5.x). This commit updates our tests to work with phpunit 5.7, 6.x, and 7.x.

phpunit6 also started flagging tests that did not do useful asserts as "risky", which they are since tests should assert that something useful was validated.

As part of this, I've updated the version of phpunit on the TEST server in `/usr/bin/phpunit` to be phpunit6 (was phpunit4). phpunit6 is the last to support PHP 7.0 which is what the TEST server is running. And if you are getting a headache trying to keep all of this in your head, you're not alone.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,11 @@
 language: php
+php:
+    # Test on a wide array of PHP versions. 7.3 and later can't be added until
+    # the phpunit tests are updated and a more recent version of phpunit is
+    # specified below.
+    - 7.0
+    - 7.1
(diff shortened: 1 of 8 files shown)


srjfoo:
Sounds good.

PR #103, opened 2019-03-23
Update MediaWiki diff engine to 1.31.1

cpeel:
For page diffs, the DP code uses code from MediaWiki. The MediaWiki code currently in DP is from MediaWiki 1.26.2 -- the last LTS release to support PHP 5.3. Now that we are upgrading our minimum PHP version to 7.0, we can update this to the version in MediaWiki 1.31.1 -- the latest LTS release to support PHP 7.0.

The two files in this MR that aren't straight lifts from MediaWiki are:
* `pinc/3rdparty/mediawiki/README.md`
* `pinc/DifferenceEngineWrapper.inc`

Functionally I can't find any changes between what we have now and what this gives us. The benefit to staying recent is that it's less work to integrate new versions when we do want/need to pick up changes.

This merge commit is gated on the one that increases our minimum PHP version (https://github.com/DistributedProofreaders/dproofreaders/pull/101). This code is available for futzing with in https://www.pgdp.org/~cpeel/c.diff-sandbox


srjfoo:
I've tested, and compared the diff sandbox with what's in the main test site, and I have not found any visible differences.

PR #104, opened 2019-03-26
fixes for two bugs and diagnostic die added

70ray:
Specify projects.genre in ProjectSearchForm.inc
to avoid ambiguous genre.
Move maybe_create_temporary_genre_translation_table() in
filter_project_list so never avoided.


70ray:
Better way found

PR #107, opened 2019-04-03
Add server time page was built to footer

cpeel:
Add server time at the bottom of every page. This gives users an indication of when the page was loaded if it's been sitting up on their computer for a while.


srjfoo:
Looks good to me.

PR #108, opened 2019-04-08
Re-write filter_project_list.inc using classes

70ray:
This implements task 1748.
Save serialised data instead of sql string and construct
sql from the data.
change language selector to select languages from all languages
with another selector to control selection mode.
Make summary string using values instead of keys. This means
having translated genres and actual special day names.
Add javascript to hide the language selection mode when
'any' selected when it would not be meaningful.
Also automate the difficulty checkboxes so selecting 'All' deselects the other options and vice versa.

### DIFF
--- a/pinc/filter_project_list.inc
+++ b/pinc/filter_project_list.inc
@@ -1,5 +1,6 @@
 <?php
 include_once($relPath.'misc.inc'); // attr_safe()
+include_once($relPath.'iso_lang_list.inc');
 
 // This file includes functions to handle the creation, update, and display of
 // the project filters used on the Round, Project, and Smooth Reading pages.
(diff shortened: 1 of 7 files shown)


70ray:
pinc/filter_project_list.inc has been largely re-written, I think the conflicts are because the genre translation changes happened in parallel with this. I don't know how to go about resolving them but the new code should completely replace the former.
A different point: In order to avoid having to link the javascript file into all three top level php files which include (indirectly) pinc/filter_project_list.inc I made a new function in pinc/html_page_common.inc to add js file names to a global array which can be used in the header. In order for this to work I changed the order of the include files in round.php so theme.inc is before showavailablebooks.inc. The order was ok in the other php files (pool and smooth_reading). Not sure if this is a good approach.

pinc/filter_project_list.inc
```diff
@@ -11,485 +12,420 @@ include_once($relPath.'misc.inc'); // attr_safe()
...
-    if($filter_changed || $username=="")
-        $form_data_source=$data;
-    else
-        $form_data_source=get_current_project_filter_data($username, $filter_type);
-
-    // display the form
-    display_project_filter_form($filter_type, $filter_label, $form_data_source, $state_sql, $custom_display_fields);
-    display_current_project_filter_settings($username, $filter_type);
-}
+add_script_file("$code_url/scripts/filter_project.js");
```
cpeel: We should not do this. Includes should only add functions and classes to namespaces, not define global variables or have secondary effects. A lot of old DP code doesn't follow this and we're slowly working to fix it.

Pages should be explicit about what javascript and other resources they are using.

70ray: I was uncertain about whether this was a good idea as I said above. I'll change to doing it the normal way.
And write a description.

70ray: In fact my function would cause the js file to be included in activity_hub where it is not needed so I see the point of the php files including js files explicitly.

PR #109, opened 2019-04-09
Re-write filter_project_list.inc using classes

70ray:
The main purpose of this change is to select the languages from all languages rather than those currently available in the round or pool.
This is task 1748.
The file displays the filters and makes the sql string and filter summary used in showavailablebooks.inc and activity_hub.php. The sql and filter summary were stored in the database.
In order to show the selected options in the filter selectors the sql was parsed to recover the arrays of selected options.
Since the sql used to select the languages will be considerably different the existing functions would be unlikely to work without modification.
It seemed like a better approach would be to store the data, which is actually the POST data but comes to this file in a function parameter, in the database in a serialised form. We can then use this data to draw the selectors and also to construct the sql when we need it and avoid the parsing altogether.
This would need considerable changes so it seemed like a good idea to re-write the file using class elements which draw the selectors and make sql components in the same way as in the project search form and results files.
After some experimentation and feedback the language selection now incorpoates an additional selector to choose whether the selected languages should be filtered as primary-only, primary-with-any or primary-or-secondary. With the 'any language' option in the main selector no filtering is done on language but the primary-only option could be taken to imply that languages-with would not appear in this case. To avoid this situation the option selector is hidden in this case using javascript.
Some other changes have also been made:
It was apparent that the new stored data could be used to make the summary text thus avoiding storing this additional data. This would make exactly the same text as previously. However this is unsatisfactory: the genres and difficulties are not translated and the special days are shown as the key acronyms rather than the full text. Instead we now construct the text components at the same time as drawing the filters using better text. These components can be assemmbled into the multiline text used on the round/pool pages and a single line version stored to use in the activity hub. The code previously used in the activity hub to convert the multiline text to a single line can then be removed.
The difficulty checkboxes were somewhat unsatisfactory since is was not clear what should happen if the 'All' option was selected as well as some other options. The javascript also automates this so when 'All' is checked the other options are unchecked and vice versa.

### DIFF
--- a/pinc/filter_project_list.inc
+++ b/pinc/filter_project_list.inc
@@ -180,7 +180,7 @@ class DifficultyElement extends ProjectFilterElement
     }
 }
 
-function process_and_display_project_filter_form($username, $filter_type, $filter_label, $data, $state_sql, $custom_display_fields=NULL)
+function process_and_display_project_filter_form($username, $filter_type, $filter_label, $data, $state_sql, $custom_display_fields = NULL)
 {
     $display_fields = [
         "language" => TRUE,
@@ -193,13 +193,14 @@ function process_and_display_project_filter_form($username, $filter_type, $filte
     // overwrite the default display state of the form fields with any
     // custom display requests
     if($custom_display_fields && is_array($custom_display_fields))
-        $display_fields=array_merge($display_fields, $custom_display_fields);
-
+    {
+        $display_fields = array_merge($display_fields, $custom_display_fields);
+    }
     // if there is data save it, else get saved data
     // if they want to reset the filter, zero out the data passed in
     if(array_key_exists("reset", $data))
     {
-        $data=[];
+        $data = [];
         save_data($username, $filter_type, $data);
     }
     else if(array_key_exists("apply", $data))
@@ -413,6 +414,7 @@ function get_project_filter_display($pguser, $filter_type)
 
 function get_raw_saved_data($pguser, $data_name)
 {
+    $data_name = mysqli_real_escape_string(DPDatabase::get_connection(), $data_name);
     $query = "SELECT value FROM user_filters WHERE username = '$pguser' AND filtertype = '$data_name'";
     $result = mysqli_query(DPDatabase::get_connection(), $query) or die(mysqli_error(DPDatabase::get_connection()));
     $row = mysqli_fetch_assoc($result);
--- a/pinc/filter_project_list.inc
+++ b/pinc/filter_project_list.inc
@@ -1,5 +1,6 @@
 <?php
 include_once($relPath.'misc.inc'); // attr_safe()
+include_once($relPath.'iso_lang_list.inc');
 
 // This file includes functions to handle the creation, update, and display of
 // the project filters used on the Round, Project, and Smooth Reading pages.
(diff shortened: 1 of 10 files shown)


pinc/filter_project_list.inc
```diff
@@ -11,485 +12,418 @@ include_once($relPath.'misc.inc'); // attr_safe()
...
+        "genre" => TRUE,
+        "username" => TRUE,
+        "difficulty" => TRUE,
+        "special_code" => TRUE,
+        "checkedoutby" => FALSE,
+    ];
+    // overwrite the default display state of the form fields with any
+    // custom display requests
+    if($custom_display_fields && is_array($custom_display_fields))
+        $display_fields=array_merge($display_fields, $custom_display_fields);
```
cpeel: Add spaces around assignments ` = `.

pinc/filter_project_list.inc
```diff
@@ -11,485 +12,418 @@ include_once($relPath.'misc.inc'); // attr_safe()
...
-
-            $filter_segments[]=sprintf("$field='%s'", mysqli_real_escape_string(DPDatabase::get_connection(), $field_datum));
-            $display_segments[]=$field_datum;
-        }
-
-        // if we have any filter segments, build the filter
-        // note the the filter needs to be parseable by
-        // _populate_project_filter_data_from_filter()
-        if(count($filter_segments))
+        $data=[];
```
cpeel: Add spaces around assignments ` = `.

pinc/filter_project_list.inc
cpeel: Also, please wrap this in `{}`s

pinc/filter_project_list.inc
```diff
@@ -11,485 +12,418 @@ include_once($relPath.'misc.inc'); // attr_safe()
...
+function get_project_filter_display($pguser, $filter_type)
 {
-    $load_associative = FALSE;
+    return get_raw_saved_data($pguser, "${filter_type}_display");
+}
 
-    switch($field)
+function get_raw_saved_data($pguser, $data_name)
+{
+    $query = "SELECT value FROM user_filters WHERE username = '$pguser' AND filtertype = '$data_name'";
```
cpeel: We should escape `$data_name` to double-ensure that this can't result in an SQL injection sometime in the future.

pinc/filter_project_list.inc
```diff
@@ -11,485 +12,418 @@ include_once($relPath.'misc.inc'); // attr_safe()
...
+    echo "</tr>\n";
+    echo "</table>\n";
 
-function get_current_project_filter_data($username, $filter_type)
-// Return a data array of the current project filter
-{
-    $filter=get_project_filter_sql($username, $filter_type);
-    return _populate_project_filter_data_from_filter($filter);
-}
+    $difficulty_element->echo_html_control();
```
cpeel: I have a slight preference for keeping the difficulty and the buttons inside the table, but only a slight.

pinc/filter_project_list.inc
```diff
@@ -11,485 +12,418 @@ include_once($relPath.'misc.inc'); // attr_safe()
...
 
-function display_current_project_filter_settings($username, $filter_type)
-// Print out the current project filter settings for a given user and filter type
-{
-    // if the username is blank (ie: the user isn't logged in) just return
-    // as there would be nothing to show them
-    if(empty($username))
-        return;
+    echo "<br><input type='submit' class='margin-half-em' name='reset' value='", attr_safe(_("Remove Filter")), "'>";
```
cpeel: We should either indent the difficulty a half-em too, or not indent these two a half-em to be consistent.

70ray:
Thanks Casey, I'm going away for a few days but I'll look at it when Iget back.

pinc/filter_project_list.inc
70ray: removed class='margin-half-em'. Added spacing between the buttons

pinc/filter_project_list.inc
70ray: If the difficulty and buttons are in the table we would need to specify colspan to make it validate which means we would have to count the columns (colspan='0' doesn't validate).
We could use flexbox instead of table.

cpeel: What you have now looks great.

cpeel:
Thanks Ray! If you can address the two other small things I identified I will get this merged in.

pinc/filter_project_list.inc
70ray: Done this now

pinc/filter_project_list.inc
70ray: Done this if I've understood correctly.

PR #110, opened 2019-04-28
JS function to mark line endings in line-wrap mode

70ray:
new function to get wrap mode in pinc/prefs_options.inc
help message for wrap preferences revised

### DIFF
--- a/faq/pophelp/prefs/prefs_pophelp.inc
+++ b/faq/pophelp/prefs/prefs_pophelp.inc
@@ -293,13 +293,13 @@ The lower control frame is not included when calculating this percentage.</p>")
 'set_h_wrap' =>
     array (
     'title' => _("Wrap Text"),
-    'content' => _("<p>Enable/Disable wrapping of lines in the proofreading area that exceed the 'Length of Text Lines' setting.</p>")
+    'content' => _("<p>Lines that exceed the width of the proofreading text area will be wrapped and, in compatible browsers, the end of each line represented by an end-of-line marker.</p>")
     ),
 
 'set_v_wrap' =>
     array (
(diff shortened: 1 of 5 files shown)


cpeel:
@70ray - what browsers have you tested this on? We should at least make sure that it doesn't do something Very Bad on older browsers. (Sorry for the delayed response, I was on vacation)

70ray:
I am on vacation now. I should be back next weekend.

70ray:
I've tested it on various IE emulations:
10 and 11 works ok.
in 8 and below a javascript error occurs in jquery so it doesn't do anything, which is ok.
in 9 it it works incorrectly. I think it is selectionStart and selectionEnd which don't work so it should be fixable but it would perhaps be better to detect IE9 and do nothing rather than adding more complexity.
I'll try it on an old Firefox in windows XP if I can get my old computer working.

cpeel:
@70ray - DP has an account at browserstack.com that you can use to do browser-specific testing. I will send you the credentials in Slack.

70ray:
I've tried it out at browserstack. Firefox is well-behaved: either works correctly or not at all. But old versions of chrome work incorrectly in a different way to IE9.
It would perhaps be best to abandon this.

70ray:
The problems I've found appeared to be bugs rather than lack of features so can't do a feature test to select good browsers. I've added a test for bad IE versions.
The problem with old chrome is not Very Bad and it's unlikely anyone would be using very old versions. I don't know whether it is a good idea to proceed with this or not.

cpeel:
Thanks Ray - I want to take this as-is. Thank you for all your work on it and patience with us to get it in.

PR #111, opened 2019-05-31
Add language redirect for walkthrough

cpeel:
Now that the walkthrough is available in French, lets do some smart language redirection and set up a consistent per-language structure for possible future expansion.


srjfoo:
Looks good to me.

PR #113, opened 2019-06-18
Rename invalid filenames upon upload

cpeel:
When using Remote File Manager to do uploads, rather than aborting an upload due to an invalid filename let's create a valid one instead. This fixes task 1662.

This PR also changes to `zipinfo` to validate that the file uploaded was a valid zip file. We use this in other places in the codebase already and it is more robust than `file`.


srjfoo:
Tested, and correctly replaces each bad character with an underscore character. Looks good to me.

PR #114, opened 2019-07-03
Config tag

70ray:
This branch makes the format preview mode (tags, no tags or re-wrap) configurable when preview is first invoked.


cpeel:
I like the functionality! I think it would be better if the UI always saved the last view mode rather than making the user go into Configure to set it. I think that would be a better UI experience, although I defer to @srjfoo and Formatters for their opinion.

Code-wise this looks great.

70ray:
This change was requested in a forum [https://www.pgdp.net/phpBB3/viewtopic.php?f=34&p=1173375&sid=a9b647dd233cc69f84e8f558d3b90ece#p1173375](url).
Saving the last view mode might not necessarily be desirable but I'll ask what the forum viewers think.

srjfoo:
As I understand it, there's a default view mode, and the only way for a formatter to change that is to go into the configuration dialog. If they customize anything, it's perfectly easy for them to use the "set initial view mode to current" if that's what they want, since they're already configuring how it displays. I think I prefer this (opt-in) rather than automatically saving the current configuration -- however, I agree that it should be up to those who are the primary users.

cpeel:
My counter argument is that the view isn't set in the configure dialog -- it's set in the main interface -- so it's unintuitive to go into the configure dialog to save it.

srjfoo:
> My counter argument is that the view isn't set in the configure dialog -- it's set in the main interface -- so it's unintuitive to go into the configure dialog to save it.

Hm ... I suspect that we're thinking about two different things. I was thinking solely of the things that can be customized in the configure dialog (i.e. colors). I agree that one shouldn't have to go to the configuration dialog to set choices that are made in the main window (i.e. wrap on or off, image and text size, etc). I still would prefer it to be opt-in rather than having it forced on users, though.

cpeel:
This looks great. I will get this merged just as soon as https://github.com/DistributedProofreaders/dproofreaders/pull/122 goes in (122 resolves an issue identified in the prior commit in master, so I'd like to keep them together).

PR #117, opened 2019-07-19
Specify time as well as date for SR deadline

cpeel:
Include the time as well as the date on the SR deadline.

[Task 1718](https://www.pgdp.net/c/tasks.php?action=show&task_id=1718)


cpeel:
Realized that I missed another spot in the code. Cleaned that up a bit too and removed the 'See below' since that was just confusing.

srjfoo:
> Realized that I missed another spot in the code. Cleaned that up a bit too and removed the 'See below' since that was just confusing.

This looks good, too -- thanks!

PR #120, opened 2019-07-24
Use project load to indicate PPer activity when in PP

cpeel:
If a project is in PP, use the project page load by the PPer to indicate
activity rather than waiting until _after_ 90 days have passed. This
will reset the counter for the 90-day notification to the PPer thus
deferring the notification.

[Task 1764](https://www.pgdp.net/c/tasks.php?action=show&task_id=1764)

### DIFF
--- a/project.php
+++ b/project.php
@@ -155,9 +155,8 @@ else
 
 function do_update_pp_activity()
 {
-// If the project has been checked out for more than 90 days and
-// the pp-er loads the page, it's interpreted as a sign of
-// activity. Update the database accordingly.
+    // If the project is checked out and the PPer loads the page, it's
+    // interpreted as a sign of activity. Update the database accordingly.
 
     global $project;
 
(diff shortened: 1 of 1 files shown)


project.php
```diff
@@ -155,27 +155,19 @@
...
-      // activity.
-
-      $projectid = $project->projectid;
-      $now = time();
-      mysqli_query(DPDatabase::get_connection(), "UPDATE projects SET modifieddate=$now WHERE projectid='$projectid'");
-      $project->modifieddate = $now;
-    }
+    $projectid = $project->projectid;
+    $now = time();
+    mysqli_query(DPDatabase::get_connection(), "UPDATE projects SET modifieddate=$now WHERE projectid='$projectid'");
```
jmdyck: Maybe we should stop (ab)using the project's modifieddate for this purpose.

`user_project_info.t_latest_home_visit` already records each user's visit to a project's "home page", so we could just look that up for the PPer when we want to know whether to remind them.

srjfoo:
The main reason I can think of is procedural for the db-req squirrels. I have, in the past (following procedures that originated well before my time) reset modifieddate to UNIX_TIMESTAMP()  to cause message to not be sent the following month.

AFAIK, there's no absolute need for this to happen, so we could just update the documentation for that procedure.

project.php
cpeel: Good idea - that would be a better option. I will get a PR up for that.

PR #122, opened 2019-07-26
Refactor and address bugs in PP notification

cpeel:
This refactors the cumbersome logic used to collate old projects for a PP and notify them into two parts:
* a function to fetch results from the database and collate them to a PP
* a function to notify the PPer of their old projects

This fixes a bug where the last PPer in the DB query was not notified (identified by @jmdyck in https://github.com/DistributedProofreaders/dproofreaders/pull/121).

Given the refactor, it might be easiest to review the full code rather than the diff.

### DIFF
--- a/crontab/notify_old_pp.php
+++ b/crontab/notify_old_pp.php
@@ -5,98 +5,100 @@ include_once($relPath.'project_states.inc');
 include_once($relPath.'maybe_mail.inc');
 include_once($relPath.'misc.inc');
 include_once($relPath.'User.inc');
+include_once($relPath.'post_processing.inc'); // get_pp_projects_past_threshold
 
 // check that caller is localhost or bail
 if(!requester_is_localhost())
     die("You are not authorized to perform this request.");
 
-$old_date = time() - 7776000; // 90 days ago.
+$projects_by_PPer = get_pp_projects_past_threshold();
 
-// send reminder email to PPers with projects checked out for longer than
-// 90 days
+// now send them notifications
+foreach($projects_by_PPer as $PPer => $projects)
+{
+    send_pp_reminders($PPer, $projects);
+}
 
-//get projects that have been checked out longer than old_date
-$sql = "
-    SELECT
-        nameofwork,
-        checkedoutby,
-        projects.projectid,
-        authorsname,
-        DATE_FORMAT(FROM_UNIXTIME(user_project_info.t_latest_home_visit), '%e %M  %Y') as nicedate
-    FROM projects LEFT OUTER JOIN user_project_info ON
-        projects.projectid = user_project_info.projectid AND
-        projects.checkedoutby = user_project_info.username
-    WHERE state = '".PROJ_POST_FIRST_CHECKED_OUT."' AND
-        user_project_info.t_latest_home_visit <= $old_date
-    ORDER BY checkedoutby, user_project_info.t_latest_home_visit;
-";
-$result = mysqli_query(DPDatabase::get_connection(), $sql);
+//---------------------------------------------------------------------------
 
-$rownum = 0;
+function send_pp_reminders($PPer, $projects)
+{
+    global $code_url, $general_help_email_addr, $forums_url,
+           $site_name, $site_signoff, $site_abbreviation;
+    global $pp_alert_threshold_days;
 
-$PPinQuestion = "";
-$lastwork = "";
-$projectslist = "";
-$numprojs = 0;
-$urlbase = "$code_url/project.php?expected_state=proj_post_first_checked_out&id=";
+    $urlbase = "$code_url/project.php?id=";
 
-while ($row = mysqli_fetch_assoc($result)) {
+    $numprojs = count($projects);
 
-    $nameofwork = $row["nameofwork"];
-    $authorsname = $row["authorsname"];
-    $checkedoutby = $row["checkedoutby"];
-    $projectid = $row["projectid"];
-    $nicedate = $row["nicedate"];
+    $projects_list = "";
+    foreach($projects as $project)
+    {
+        $nameofwork = $project["nameofwork"];
+        $authorsname = $project["authorsname"];
+        $projectid = $project["projectid"];
+        $modifieddate = strftime("%e %B %Y", $project["modifieddate"]);
+        $lastvisitdate = strftime("%e %B %Y", $project["lastvisitdate"]);
 
-    if ($PPinQuestion != $checkedoutby) {
-        // have finished the last PPer. Send email to them
-        if ($rownum > 0) {
-
-            $user = new User($PPinQuestion);
-            $email = $user->email;
-
-            echo $PPinQuestion . "\n" . $projectslist ."\n\n";
-
-            if ($numprojs == 1) {
-                $message = "Hello $PPinQuestion,\n\nThis is an automated message.\n\n
-Our database indicates that you have had a PP project checked out for more than 90 days:\n\n
-$projectslist\n\n
-If you haven't yet finished and wish to continue working on this book, please log in to $site_url and visit $url . This will update the status of the project. If you need help please forward a copy of this email (quoting the information on the book, above) with a brief description of the status to $general_help_email_addr.\n\n
-If you have completed your work on the book, please log in to $site_url and visit $url. Select the 'Upload for verification' option and follow the prompts. You will be able to leave a message for the verifier during this process, if you have any special information or comments to pass on.\n\n
-If you are waiting on missing images or page scans, please add the details to the Missing Page Wiki at: $forums_url/viewtopic.php?t=7584\n\n
-If you no longer wish to have this text assigned to you please visit the $site_name website Post Processing section and select Return to Available for this book, or forward this email to $general_help_email_addr and state that you would no longer like to have the book in question assigned to you so that we may return it to the available pool for someone else to work on.\n\n
-$site_signoff";
-            } else {
-                $message = "Hello $PPinQuestion,\n\nThis is an automated message.\n\n
-Our database indicates that you have had $numprojs PP projects checked out for more than 90 days:\n\n
-$projectslist\n\n
-If you wish to continue working on some or all of these books, please log in to $site_url and visit each such project's home-page (copy the URL listed with the project above and paste it into your browser's address-field). Doing this will update the status of the project and let us know that you are still working on it. If you need help please forward this email, quoting the list of books, with a brief description of the status for each of the various books listed above that you need help with to $general_help_email_addr.\n\n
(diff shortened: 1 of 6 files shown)


crontab/notify_old_pp.php
```diff
@@ -10,93 +10,132 @@
...
+
+If you are waiting on missing images or page scans, please add the details to the Missing Page Wiki at: $forums_url/viewtopic.php?t=7584
+
+If you no longer wish to have this text assigned to you please visit the $site_name website Post Processing section and select Return to Available for this book, or forward this email to $general_help_email_addr and state that you would no longer like to have the book in question assigned to you so that we may return it to the available pool for someone else to work on.
+
+$site_signoff
+EOF;
+    }
+    // more than one project
+    else
```
jmdyck: Wouldn't `else // more than one project` read better?

Personally, I'd probably put the comment in the block:
```
else
{
    // more than one project
    ....
```

Looks like [our style quide](https://www.pgdp.net/wiki/DP_Code_Best_Practices#Comments) doesn't address this.

crontab/notify_old_pp.php
```diff
@@ -10,93 +10,132 @@
...
+
+// now send them notifications
+foreach($projects as $checkedoutby => $projects)
+{
+    send_pp_reminders($checkedoutby, $projects);
+}
+
+//---------------------------------------------------------------------------
+
+function fetch_and_collate_projects_by_pper($last_visit_date)
```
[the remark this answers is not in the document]
srjfoo: I wondered about this one, too, but more specifically whether it should include `time` in the variable name.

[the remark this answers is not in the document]
cpeel: Good idea.

crontab/notify_old_pp.php
```diff
@@ -10,93 +10,132 @@
...
+        $projectid = $project["projectid"];
+        $modifieddate = strftime("%e %B %Y", $project["modifieddate"]);
+        $lastvisitdate = strftime("%e %B %Y", $project["lastvisitdate"]);
+
+        $url = $urlbase . $projectid;
+        $work_details  = "$nameofwork by $authorsname ($projectid)";
+        $time_details = "[checked out since $modifieddate, project last visited $lastvisitdate]";
+        $projects_list .= "$work_details\n$time_details\n    $url\n\n";
+
+        echo "    $work_details\n";
```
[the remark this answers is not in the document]
cpeel: Yup, that makes sense particularly after reading Sharon's email regarding the desired changes to this file.

crontab/notify_old_pp.php
```diff
@@ -10,93 +10,132 @@
...
+
+    return $projects;
+}
+
+function send_pp_reminders($checkedoutby, $projects)
+{
+    global $code_url, $general_help_email_addr, $forums_url,
+           $site_name, $site_signoff, $site_abbreviation;
+
+    $urlbase = "$code_url/project.php?id=";
```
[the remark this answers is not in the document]
cpeel: I took it out because I was (and remain) dubious of its value. If you / @srjfoo feel strongly about it I'll happily put it back in, however.

cpeel:
I factored out the logic to query for outdated projects into a separate file. This will be used in additional work I'm doing outside of the crontab. I wanted to avoid duplicating this logic in multiple places.

crontab/notify_old_pp.php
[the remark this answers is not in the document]
jmdyck: I'll defer to @srjfoo's judgement.

cpeel:
The last series of commits implement @srjfoo's updates to the email, including:
* Updated email contents
* Email contents for both the first of the month and mid-month
* Basing alerts on midnight of the first of the current month, not on `now()`
* Translating email content

One *big* note about translating email content: the only locale about the user that we know about when the crontab runs is what they have saved in their user preferences. This does not necessarily reflect what language they see in the UI, however, because the UI defaults to detecting what their browser UI is. This means that users who, for instance, see the UI in French because their browser is requesting French, but have not explicitly set their UI preferences to French will still get an email in English.

cpeel:
Oh, also note that because `$projects_list_string` is translated, dpadmin and dp-req will get possibly-translated strings in this listing. Given that these strings are minor I decided not to worry about it but let me know if I should change that.

crontab/notify_old_pp.php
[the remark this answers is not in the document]
srjfoo: From my personal point of view, I generally find the `expected_state=proj_post_first_checked_out` more annoying than anything else.
However, I can see some value in it for a PPer who may be reviewing the email after an extended delay, and who thinks that the list is current, only to find, upon clicking on a link, that the project is no longer assigned to them. If the `expected_state ...` is in the URL, that triggers a highly noticeable message. If not, they may or may not notice that it's no longer assigned to them. 

However, I also see the value of shortening the URL in the email notification.

My preference is for the shorter URL, but I thought I ought to give the other option a chance, and asked Linda. She voted for the shorter URL, also.

@jmdyck, if you feel that there is a compelling reason to keep it, that's okay, too.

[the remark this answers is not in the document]
cpeel: > However, I can see some value in it for a PPer who may be reviewing the email after an extended delay, and who thinks that the list is current, only to find, upon clicking on a link, that the project is no longer assigned to them.

At best that will indicate to the user if the project is in another state besides PP -- it doesn't do anything to indicate if they are still assigned to the project as the PPer. So for this to be useful the PPer would need to have ignored the email long enough for the project to be assigned to another PPer and for that PPer to finish it and have the project move out of PP.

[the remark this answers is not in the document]
srjfoo: > > However, I can see some value in it for a PPer who may be reviewing the email after an extended delay, and who thinks that the list is current, only to find, upon clicking on a link, that the project is no longer assigned to them.
> 
> At best that will indicate to the user if the project is in another state besides PP -- it doesn't do anything to indicate if they are still assigned to the project as the PPer. So for this to be useful the PPer would need to have ignored the email long enough for the project to be assigned to another PPer and for that PPer to finish it and have the project move out of PP.

There are a few other possibilities, but yes. As long as the PPer reads the email reasonably quickly, it should be irrelevant.

cpeel:
Tonight I will get the email contents tweaked with the new content from @srjfoo. If anyone (@jmdyck, et al) have time to look at the other changes in this PR prior I am happy to address anything else at that time too.

srjfoo:
> Oh, also note that because `$projects_list_string` is translated, dpadmin and dp-req will get possibly-translated strings in this listing. Given that these strings are minor I decided not to worry about it but let me know if I should change that.

I think that will be fine.

crontab/notify_old_pp.php
```diff
@@ -5,98 +5,107 @@
...
-        $PPinQuestion = $checkedoutby;
-    }
+$projects_by_PPer = get_pp_projects_past_threshold();
 
-    $numprojs++;
+// if the script is run before the 14th of the month, send out the first email
+// otherwise send out the second one
+if(strftime("%d", time()) < 14)
+{
+    $which_email = "first";
```
jmdyck: "email" is sometimes a shorthand for "email address", so `$which_message` might be clearer.

crontab/notify_old_pp.php
```diff
@@ -5,98 +5,107 @@
...
+        $message[] = "==" . _("Have you completed your work?") . "==";
+        $message[] = _("If you have completed your post-processing work, please visit the appropriate link listed above, select the 'Upload for verification' option on the Project Page, and follow the prompts. If you have any special information or comments to pass on, you will be able to leave a message for the verifier during this process. If you have Direct Upload capability and no longer use Post-Processing Verification, please upload each completed project directly to PG.");
+        $message[] = "==" . _("Are there any problems?") . "==";
+        $message[] = sprintf(_("If any project is missing page images or illustrations, please contact the Project Manager (PM) for assistance. If the PM does not respond within a reasonable amount of time, please contact db-req at %s concerning any problems."), $db_requests_email_addr);
```
jmdyck: The name "db-req" is pgdp-specific, isn't it? It looks like we have hard-coded it in a few places, but not many. Oh well, PPing is pgdp-specific in the first place.

cpeel: Yes, this is more pgdp.net creep into the codebase. I've (almost) stopped fighting these little intrusions. <sigh>

srjfoo: I have no objection to changing the wording from `please contact db-req at %s concerning` to `please email %s and request help concerning`. I need to occasionally be reminded to try to agnosticize (is that a word) what goes into the code.

cpeel: Nicely done - updated.

jmdyck: Yeah, I was going to suggest something like that, but noticed there were also occurrences of "db-req" in the mid-month message, so decided not to bother.

cpeel:
@jmdyck - are there other changes are looking for?

pinc/post_processing.inc
```diff
@@ -0,0 +1,78 @@
...
+
+// internal function to avoid duplicating the SQL logic that determines
+// projects past the PP alert threshold
+function _run_pp_threshold_query_result($columns, $PPer, $ordering_criteria=NULL)
+{
+    global $pp_alert_threshold_days;
+
+    $column_selector = implode(", ", $columns);
+    $user_selector = $PPer ? "AND checkedoutby = '$PPer'" : "";
+    $ordering_criteria_clause = $ordering_criteria ? "ORDER BY $ordering_criteria" : "";
```
jmdyck: Sorry if I wasn't clear, what I meant was the renamings:
* $order_by -> $ordering_criteria  [as you've done]
* $order_by_sql -> $order_by_clause  [not $ordering_criteria_clause]

jmdyck: That's all.

crontab/notify_old_pp.php
srjfoo: That wording can be tweaked, too. How about:
> $message[] = sprintf(_("If you have already notified %s of an expected absence, you may safely ignore this message. Otherwise, any projects in your queue for which a previous notice has been sent may be reclaimed if you do not respond or renew within the next 7 days."), $db_requests_email_addr);

Post-processing *can* be a pgdp-specific thing, but needn't be. If it is, then whoever is using the code won't need this cron, anyway. If it isn't, they may need something like this.

pinc/post_processing.inc
cpeel: You weren't unclear, I just totally fat-fingered the search/replace.

crontab/notify_old_pp.php
srjfoo: Looks good with the new wording, thanks!

PR #123, opened 2019-07-29
Replace shell commands used for manipulating zip files

mlazaric:
Replaces shell commands for zip validation, listing contents, extracting and compressing zip files with equivalent PHP functions. Adds some basic unit tests for all of them and replaces all occurrences of those shell commands that I found with the appropriate function.

Still requires some discussion and testing to make sure it doesn't break any existing functionality.


tools/project_manager/post_files.inc
```diff
@@ -107,13 +107,14 @@ function generate_interim_file($projectid, $limit_round_id, $which_text, $includ
...
-    {
-        echo "post_files: the zip command failed.";
-    }
-    // cleanup will get called whatever happens  
+    header("Content-disposition: attachment; filename={$filename}.zip");
+
+    $zip_path = "{$dirname}/{$filename}.zip";
+    create_zip_from([$textfile_path], $zip_path);
+    readfile($zip_path);
+    unlink($zip_path);
```
mlazaric: Would it be better to put this in `function clean_up_temp($dirname, $textfile_path)` for consistency?

tools/project_manager/post_files.inc
```diff
@@ -107,13 +107,14 @@ function generate_interim_file($projectid, $limit_round_id, $which_text, $includ
...
-    passthru("zip -q -j - $textfile_path", $return_code);
-    if ($return_code != 0)
-    {
-        echo "post_files: the zip command failed.";
-    }
-    // cleanup will get called whatever happens  
+    header("Content-disposition: attachment; filename={$filename}.zip");
+
+    $zip_path = "{$dirname}/{$filename}.zip";
+    create_zip_from([$textfile_path], $zip_path);
```
mlazaric: This is the contents of the generated zip file:

image

Would it be better to fix this somehow in the `create_zip_from` function or by using [`chdir`](https://www.php.net/manual/en/function.chdir.php) to change the directory to `./tmp/PROJECT_DIR` before creating the zip archive?

mlazaric: This is most likely problematic in a couple places, so it is probably a good idea to think it through.

mlazaric:
One more use of shell commands to create a zip that I haven't replaced:

https://github.com/DistributedProofreaders/dproofreaders/blob/59f2cf193791efedc319cbe55cbb70aef3af959d/tools/download_images.php#L83-L100

Would it be worthwhile to create a whole new function for this kind of stuff (where we are zipping up strings instead of files) or just write the strings into a a temporary file and zip that up using the existing function?

pinc/misc.inc
```diff
@@ -595,6 +595,110 @@ function get_filelist($dirpath, $validext = array(), $with_path = False)
     closedir($d);
     return $filelist;
 }
+
+// -----------------------------------------------------------------------------
+function is_valid_zip_file(string $file_path): bool
+// Return whether the file is a valid zip file.
```
cpeel: Please put these comments above the function declaration instead of between it and the open bracket. I realize that the code is inconsistent in this.

tools/project_manager/post_files.inc
cpeel: I do think we want the ability to create a directory-less zip archive (ie: just the file(s) in question in the zip, not the directory). I think it makes sense to put this functionality within `create_zip_from()`.

tools/project_manager/post_files.inc
cpeel: Yes, the unlink should be in `clean_up_temp()` too.

cpeel:
This is great work, thank you for tackling it!

tools/project_manager/post_files.inc
mlazaric: Turns out that was easier than I thought it was going to be. Fixed in 6ae7e0a8a6467f49381d11f7512e91f6e26b9461. Now all files added to the zip archive *will* go into the root directory, is that okay? 

Anything more complicated is probably best done directly through `ZipArchive` or through a more specific method. 

image

cpeel: I think that's fine. That's certainly how we are using `zip` currently by the looks of it.

cpeel:
> Awesome. Give me a day or two to get the WordCheck change in and we can add `php-zip` to the `.travisci` file and then we'll get this merged in!

WordCheck branch has been merged in. Please rebase and update `.travis.yml` to include `php-zip`, eg:
```
    - sudo apt-get install -y php-zip aspell aspell-en
```

cpeel:
> There is one difference that you might want to resolve, if possible -- comparing the "Choose file" dialog for `project_manager/remote_file_manager.php` with that which, for instance, uploads an SR report as detailed in `tools/upload_text.php`, the former will grey out choices that do not have a `.zip` extension, and if you drag and drop such a file will show a popup with an error message. The latter does not filter out files without a `.zip` extension until after the `Upload` button is pushed.

Good catch! I'll handle that in a separate MR than this one.

PR #125, opened 2019-08-01
Remove sample directory

cpeel:
Code in the sample directory hasn't been usefully touched or modified
in almost a decade. Remove the sample directory entirely from the
codebase. The files are always here in git if anyone ever needs them
in the future.

Also, on pgdp.net we move the directory out of the webspace so they
aren't web-accessible but we've never instructed other sites to do
likewise, which seems bad.


srjfoo:
I have no objections, either.

cpeel:
Breadcrumb added to commit message.

PR #126, opened 2019-08-01
Picker fix

70ray:
New character picker with hard-coded data.


cpeel:
This is way more code change than what I was looking for. What I'd like to see are surgical updates to `pinc/CharacterSelector.inc` and `pinc/ProofreadingToolbox.inc` plus some small JS that change the selectors to your new style. I don't see a need for the API layer or building the HTML for the pickers on-the-fly in JS.

We may eventually move to this more JS-centric method of building our interfaces but I'm not yet ready to do that outside of Format Preview.

70ray:
Thanks Casey, I was wondering if you would prefer not to have the api. I have got such a version somewhere. I'll reinstate it.

PR #127, opened 2019-08-03
Picker min

70ray:
Revised toolbox and character picker with static data.


pinc/CharacterSelector.inc
```diff
@@ -1,241 +1,137 @@
...
-include_once($relPath.'misc.inc'); // attr_safe(), html_safe()
+
+include_once($relPath.'prefs_options.inc'); // get_user_proofreading_font()
 
 class CharacterSelector
 {
-    public function __construct()
+    function __construct ()
     {
+        $this->picker_set =
```
cpeel: This does not include the set of UTF8 characters from the prior version, we should retain that functionality.

pinc/ProofreadingToolbox.inc
```diff
@@ -1,375 +1,111 @@
...
-        );
-        $this->define_tool_link(
-            'blank_page',
-            '[Blank Page]',
-            '[Blank Page]',
-            "top.replaceAllText('[Blank Page]')"
-        );
-    }
+        global $code_url;
+        $formatting = is_formatting_round($round);
```
cpeel: This is confusing here. It would be clearer to determine if we are in a formatting around when we define what buttons to include in `initialize_tool_buttons()` rather than including all of them and relying on the draw function to handle that logic.

styles/toolbox.css
```diff
@@ -0,0 +1,71 @@
+body {
```
cpeel: We should define this as a .less file and incorporate it into the theming rather than define colors here.

pinc/CharacterSelector.inc
```diff
@@ -1,241 +1,137 @@
...
+            $row_string .= $this->draw_row($picker["upper"]);
+            $row_string .= $this->draw_row($picker["lower"]);
+            $row_string .= "</div>\n";
+        }
+
+        $selector_string .= "</div>\n";
+
+        echo "<div id='selector' style='font-family: $font_face;' class='flexbox'>";
+        echo "\n<div id='char-selector' class='nowrap'>$selector_string$row_string</div>\n";
+        echo "<div class='char_box'><input type='text' readonly class='big_text' id='large_char'></div>\n";
```
cpeel: Consider putting this on the left of everything so that it doesn't jump around when the different character classes are selected.

styles/toolbox.css
```diff
@@ -0,0 +1,71 @@
...
+}
+
+.flexbox {
+    display: flex;
+}
+
+/* default buttons have small text = 83% */
+.selector_button {
+    margin-top: 1px;
+    font-size: 16px;
```
cpeel: We should define font sizes in ems. I think it would be better if these were also bigger.

styles/toolbox.css
```diff
@@ -0,0 +1,71 @@
...
+}
+
+.big_text {
+    font-family: inherit;
+    font-size: 300%;
+    width: 0.9em;
+    background-color: #FFF8DC;
+    margin: 0.1em;
+}
+/* specify font-size in px so consistent in different browsers */
```
cpeel: It's a best practice to use ems for specifying font sizes instead of pixels -- even at the expense of inconsistency between browsers.

pinc/ProofreadingToolbox.inc
```diff
@@ -1,375 +1,111 @@
...
+        echo "<div id='tools'>";
+        foreach ($this->tool_buttons as $tool_button)
         {
-            return $source;
+            $tool_button->draw($formatting);
         }
 
-        $result = array();
-        foreach ( $keys as $key )
+        draw_popup_link(_('Search/Replace'), 'srchrep.php', 'dp_searchrepl', 300, 250);
```
cpeel: Consider putting the tool links on a line by themselves.

pinc/ProofreadingToolbox.inc
```diff
@@ -1,375 +1,111 @@
...
-    private function define_popup_link( $id, $label, $url, $window_name, $width, $height )
+    public function __construct($accesskey, $title, $label, $onclick, $formatting = false)
     {
-        $window_features = "width=$width,height=$height,directories=0,location=0,menubar=0,resizable,scrollbars,status=0,toolbar=0";
-        $js = "window.open('$url','$window_name','$window_features'); return false;";
-        $element = "<a href='#' onClick=\"$js\"\n>$label</a>\n";
-
-        $this->_popup_links[$id] = $element;
+        $access_string = $accesskey ? "accesskey='$accesskey'" : '';
+        $this->html = "<button type='button' class='tool-button' onclick='$onclick' $access_string title='" . attr_safe($title) . "'>$label</button>\n";
```
cpeel: These have very small text, lets make them bigger,

pinc/ProofreadingToolbox.inc
```diff
@@ -1,375 +1,111 @@
...
-                // (i.e., currently, hieroglyph_transliterator)
-            }
+            spacer_bar();
+            draw_popup_link(_('Table Maker'), 'mktable.php', 'dp_mktable', 600, 500);
         }
-        return $result;
+        spacer_bar();
+        draw_popup_link(_('Greek Transliterator'), 'greek2ascii.php', 'gkasciiWin', 640, 250);
+
+        $url = get_faq_url($round->document);
```
cpeel: Ditto putting the doc links on a line by themselves

pinc/ProofreadingToolbox.inc
cpeel: I also think we should retain a "Pop-up tools:" prefix to the line too.

pinc/ProofreadingToolbox.inc
cpeel: And lets prefix them with "Documentation:" and remove the `[]`s (why were the brackets even there in the first place!?)

pinc/ProofreadingToolbox.inc
```diff
@@ -1,375 +1,111 @@
...
+        echo "[<a href='$url' class='nowrap' target='roundDoc'>", _('Guidelines'), "</a>]\n";
+
+        echo "[<a href='$code_url/faq/InterfaceDiagram.png' class='nowrap' target='ifaceWin'>", _("Proofreading Diagram"), "</a>]\n";
```
cpeel: I think we should remove this diagram link altogether, but will defer to @srjfoo and @lhamilton1.

srjfoo: Okay by me. It's not up-to-date, anyway (still says that reporting bad pages can be done only from the first round), and doesn't mention format preview.

pinc/CharacterSelector.inc
srjfoo: Currently, it doesn't (at least not in Firefox and Safari). The picker block is sized to fit the widest picker set, and the toolbox changes based on the width of the proofing interface.

pinc/CharacterSelector.inc
70ray: I started on this but it seems like it could be better to change the way the data is stored back to unicode code points rather than utf8 strings. Some characters are difficult to distinguish in a text editor for example for example U+00D0 capital eth, U+0110 D with stroke, and U+0189 African D. For the same reason it would be better to have a tool-tip text (possibly empty) for every character stored with it rather than in a separate array.
These considerations would apply if the data is hard coded or if we were to store configurable data in the database.

pinc/CharacterSelector.inc
70ray: @srjfoo, I think Casey means the box with the large character which jumps around now.
I'll move it to the left.

pinc/ProofreadingToolbox.inc
70ray: This was api-ish. I've simplified it

pinc/ProofreadingToolbox.inc
70ray: Done this.

pinc/ProofreadingToolbox.inc
70ray: Done this.

pinc/ProofreadingToolbox.inc
70ray: Link removed.

pinc/CharacterSelector.inc
srjfoo: Ah, okay, that makes sense.

pinc/CharacterSelector.inc
cpeel: <nod> I think keeping them as unicode code points make sense.

I wonder if for the tool tip we want to use the Unicode character name. They aren't exactly sexy but they are specific and unambiguous -- [IntlChar::charName](https://www.php.net/manual/en/intlchar.charname.php), for instance. That would add a new dependency (`php-intl`) but I'm OK with that.

pinc/ProofreadingToolbox.inc
cpeel: To @srjfoo's point, we might not have a lot of room to make them bigger -- particularly on narrower windows. :/

pinc/CharacterSelector.inc
srjfoo: On thinking this over a bit longer -- yes, I can see the sense of moving it, however, it does provide a distinct visual break. If you move the box to the left of the character picker, and the picker characters fill all the way to the end of that div, the character buttons will be right up against the toolbox, and I think that would be visually confusing. I think you'd need a vertical division with a reasonable margin on both sides for that to work visually.

pinc/CharacterSelector.inc
70ray: Great idea, that will save a lot of work.

70ray:
Some fonts are larger than others so specifying the font size in ems is relative to its basic size. This means that if the font for the picker buttons is made the same as the proofreading font then they can change size quite considerably. I have removed this feature so they always use the default button font (as the present dropdowns do). This means some characters look different to what will appear in the text. (particularly the ordinals which are underlined in some fonts but not others).

70ray:
I'd like to abandon the last four commits because of unexpected problems. I don't know the cleanest way to do this. Shall I push a new branch without them?

cpeel:
You can either:

1) `git revert` the commits (do it in the reverse order you made them) or
2) `git rebase` and drop the last 4 commits

Either way I'll squash your MR into one commit when we merge it.

70ray:
Unfortunately I'd already done a hard reset on my branch and then push would not work. I now done a forced push which seems to have worked.

pinc/slim_header.inc
```diff
@@ -19,7 +19,7 @@ function slim_header($title="", $extra_args=array())
     // Include the slim_header stylesheet
     if(!isset($extra_args['css_files']))
         $extra_args['css_files'] = array();
-    $extra_args['css_files'][] = "$code_url/styles/slim_header.css";
+    $extra_args['css_files'] = array_merge(["$code_url/styles/slim_header.css"], $extra_args['css_files']);
```
cpeel: I don't understand why this change was made. Is it to ensure that `slim_header.css` is always in front? If so, `array_unshift()` would be clearer.

cpeel:
I'm really loving how this is looking - I think this is a major improvement over what we have now and sets us up for Unicode characters in the future too. I'd like to get a conversation going with the larger community with these changes.

pinc/slim_header.inc
70ray: Yes, it is so the body margin in slim_header.css can be over-ridden. I've changed is as you suggest.

70ray:
I also made a change to avoid having to set the toolbox body colour.
And rebased to master so it won't appear that we are reverting anything.

70ray:
Reverted the change to the way the colour of the toolbox works because causes white area to appear when toolbox is scrolled.
The body colour of the frame is set. This isn't a good way to do it but it doesn't cause any strange effects.
It would be good to make the toolbox not in a frame so it always appears the correct size. I tried this out with a flex layout and putting the top frames in an iframe but probably best not to change too much at once.

pinc/CharacterSelector.inc
```diff
@@ -1,241 +1,189 @@
...
-                    {
-                        echo "<option value='$codepoint'>&#{$codepoint};</option>\n";
-                    }
-                }
+                $row .= "<button type='button' class='picker invisible'></button>";
             }
             else
             {
-                assert( 0 );
+                $title_string = attr_safe(IntlChar::charName($character));
```
cpeel: Please update PHP section in `SETUP/installation.txt` to mention that the Intl PHP extension must be installed. On Ubuntu this is provided by the `php-intl` package.

styles/layout.less
```diff
@@ -709,23 +709,63 @@ table.dirlist {
...
-        display: inline;
-        margin: 0 1px 1px 0;
-        padding-top: 1px;
-        cursor: pointer;
+        padding: 1px 7px 0 7px;
+        cursor:pointer;
+    }
+
+    .key-block {
+    display: none;
```
cpeel: Indent this.

styles/layout.less
```diff
@@ -709,23 +709,63 @@ table.dirlist {
...
+    }
+
+    .show {
+    display: table;
```
cpeel: Indent this.

styles/layout.less
```diff
@@ -709,23 +709,63 @@ table.dirlist {
...
+    .table-row {
+        display: table-row;
+    }
+
+    .table-cell {
+        display: table-cell;
+    }
+
+    .char_box {
+    padding-top: 1px;
```
cpeel: Indent this.

styles/toolbox.css
cpeel: Why don't we include this in the regular `.less` files and use `$extra_args['body_attributes']` to add the class to the `<body>` tag? We could drop this file altogether.

70ray:
Implemented these changes.

tools/proofers/ctrl_frame.php
```diff
@@ -9,9 +9,12 @@
...
 $round = get_Round_for_round_id($round_id);
 
-$header_args = array(
-    "body_attributes" => 'onLoad="top.cRef = top.markRef = document.markform;" style="background-color: #CDC0B0; padding: 0; margin: 0;"',
-);
+$header_args = [
+    "js_files" => [
+        "$code_url/tools/proofers/character_selector.js",
+    ],
+    "body_attributes" => 'style="background-color: #CDC0B0; padding: 0; margin: 0;"',
```
cpeel: Please move this CSS into the `layout.less` and `theme.less` and then pull them in via:
```
"body_attributes" => 'class="control-frame'
```
or an ID or somesuch.

pinc/ProofreadingToolbox.inc
```diff
@@ -1,375 +1,88 @@
...
+function draw_tool_button($accesskey, $title, $label, $onclick, $draw = true)
 {
-    public function __construct()
+    if ($draw)
     {
-        $this->initialize_popup_links();
-        $this->initialize_tool_buttons();
-        $this->initialize_tool_links();
-        $this->character_selector = new CharacterSelector();
+        $access_string = $accesskey ? "accesskey='$accesskey'" : '';
```
cpeel: We should wrap `$accesskey` in `attr_safe()` too so the caller can't accidentally pass in invalid HTML.

pinc/ProofreadingToolbox.inc
```diff
@@ -1,375 +1,88 @@
...
-        $window_features = "width=$width,height=$height,directories=0,location=0,menubar=0,resizable,scrollbars,status=0,toolbar=0";
-        $js = "window.open('$url','$window_name','$window_features'); return false;";
-        $element = "<a href='#' onClick=\"$js\"\n>$label</a>\n";
+// Popup Link
+// A link that, when clicked, opens a new window containing some tool.
+function draw_popup_link($label, $url, $window_name, $width, $height)
+{
+    $window_features = "width=$width,height=$height,directories=0,location=0,menubar=0,resizable,scrollbars,status=0,toolbar=0";
+    $js = "window.open('$url','$window_name','$window_features'); return false;";
+    echo "<a href='#' class='nowrap' onClick=\"$js\"\n>$label</a>\n";
```
cpeel: Let's not add a newline at the end of this. If I called this function I would not expect it to add whitespace to the end of the link it output. We should remove the newline in the middle of the `<a>` too.

pinc/ProofreadingToolbox.inc
```diff
@@ -1,375 +1,88 @@
...
+}
 
-        $this->_popup_links[$id] = $element;
-    }
+function spacer_bar()
+{
+    echo "|\n";
```
cpeel: Make this `echo " | ";`. This ensures that the spacer has a space before the bar, regardless of what's before it, and that it is outputting what is visually seen (a space before and after) rather than a newline (which acts the same in HTML but requires more cognitive load on the developer who looks at this next).

70ray:
Thanks Casey, I've implemented these changes.

PR #130, opened 2019-08-16
Prompt the user to upload .zip files

cpeel:
For zip file uploads, tell the browser that we accept .zip files
to prevent the user from selecting a file that isn't a .zip file
(assumimg the browser supports it).

### DIFF
--- a/tools/project_manager/remote_file_manager.php
+++ b/tools/project_manager/remote_file_manager.php
@@ -293,7 +293,7 @@ function do_showupload()
     $form_content .= "<div id='old_uploader'>";
     $form_content .= $submit_blurb;
     $form_content .= _("File to upload") . ":&nbsp;";
-    $form_content .= "<input type='file' name='the_file' size='50' maxsize='50'>";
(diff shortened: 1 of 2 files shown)


mlazaric:
This behaves strangely on vivaldi: 

image

(notice the "Zip archive, **unknown**" and how it doesn't hide the non zip files)

image

Manually selecting "Zip archive" from the dropdown does hide non zip files:

image

These are all from https://www.pgdp.org/~cpeel/c/tools/project_manager/remote_file_manager.php

Not sure why it added the unknown part.

mlazaric:
Chrome seems to have the same strange behavior.

cpeel:
The update in `remote_file_manager.php` is for the non-javascript version of the uploader, which you probably aren't using unless you've disabled JS.

srjfoo:
I'm not seeing that behavior in either FF or Safari, but that may just mean that I don't have any of that kind of file sitting around. Afaik, I have javascript enabled.

PR #131, opened 2019-08-21
Tweak condition for skipping unrecognized filenames

jmdyck:
"preg_match() returns 1 if the pattern matches, 0 if it does not, or FALSE if an error occurred."

So by comparing the result to FALSE, we're only detecting when an error occurs (which is probably never), rather than detecting a simple failure of the pattern to match.
    
Instead, just check if the result is falsy (covering both 0 and FALSE).

[This tweak is already in place on pgdp.net.]


cpeel:
Nice catch. LGTM!

PR #132, opened 2019-08-22
Don't update user_project_info on checkout/return

cpeel:
Don't update user_project_info.t_latest_page_event upon page checkout
or return to round. This allows a user to "try out" a page in a project
and return it without it being added to their My Projects page.

[Task 1208](https://www.pgdp.net/c/tasks.php?action=show&task_id=1208)

### DIFF
--- a/SETUP/upgrade/13/20190929_update_user_project_info.php
+++ b/SETUP/upgrade/13/20190929_update_user_project_info.php
@@ -1,0 +1,119 @@
+<?php
+
+$relPath='../../../pinc/';
+include_once($relPath.'base.inc');
+include($relPath.'udb_user.php'); // $archive_db_name
+
+header('Content-type: text/plain');
+
+// ------------------------------------------------------------
+
+echo "Updating user_project_info to set t_latest_page_event=1 only for\n";
+echo "projects where the user has saved a page as done or in-progress...\n\n";
+
+// Load a list of all users
+$sql = "
+    SELECT username
+    FROM users
+    ORDER BY username
+";
+
+$result = mysqli_query(DPDatabase::get_connection(), $sql)
+    or die( mysqli_error(DPDatabase::get_connection()) );
+
+$users = [];
+while($row = mysqli_fetch_assoc($result))
+{
+    $users[] = $row['username'];
+}
+mysqli_free_result($result);
+
+// Loop though all users getting their user_project_info records to see what
+// projects they have "worked on" (ie: t_latest_page_event > 0)
+foreach($users as $username)
+{
+    echo "Processing $username...\n";
+
+    // load all projects this user has "worked on"
+    $sql = "
+        SELECT projectid
+        FROM user_project_info
+        WHERE
+            username='$username'
+            AND t_latest_page_event > 0
+    ";
+
+    $result = mysqli_query(DPDatabase::get_connection(), $sql)
+        or die( mysqli_error(DPDatabase::get_connection()) );
+
+    $projectids = [];
+    while($row = mysqli_fetch_assoc($result))
+    {
+        $projectids[] = $row['projectid'];
+    }
+    mysqli_free_result($result);
+
+    // For each project, get the distinct event_types from page_events, which
+    // indicates what the user did in that project. If neither the 'saveAsDone'
+    // or the 'saveAsInProgress' events are present then this is a case where
+    // the new code in pinc/DPage.inc would not have updated t_latest_page_event.
+    // For these projects, set user_project_info.t_latest_page_event=0 for this
+    // (username, projectid), to achieve the same effect retroactively.
+    foreach($projectids as $projectid)
+    {
+        // Process both the main page_events table and the archived
+        // page_events table.
+        $current_event_types = get_distinct_event_types("page_events", $username, $projectid);
+        $archived_event_types = get_distinct_event_types("$archive_db_name.page_events", $username, $projectid);
+        $event_types = array_unique(array_merge($current_event_types, $archived_event_types));
+
+        // see if the only event_types are checkout and returnToRound
+        if(in_array('saveAsDone', $event_types) ||
+            in_array('saveAsInProgress', $event_types))
+        {
+            continue;
+        }
+        else
+        {
+            echo "    setting t_latest_page_event=0 for project $projectid\n";
+            $sql = "
+                UPDATE user_project_info
+                SET t_latest_page_event=0
+                WHERE
+                    username='$username'
+                    AND projectid='$projectid';
+            ";
+            $result = mysqli_query(DPDatabase::get_connection(), $sql)
+                or die( mysqli_error(DPDatabase::get_connection()) );
+        }
+    }
+}
+
+// ------------------------------------------------------------
+
+echo "\nDone!\n";
+
+// ------------------------------------------------------------
+
+function get_distinct_event_types($table, $username, $projectid)
+{
+    $sql = "
+        SELECT DISTINCT event_type
+        FROM $table
+        WHERE
+            username='$username'
+            AND projectid='$projectid'
+    ";
+    $result = mysqli_query(DPDatabase::get_connection(), $sql)
+        or die( mysqli_error(DPDatabase::get_connection()) );
+
+    $event_types = [];
+    while($row = mysqli_fetch_assoc($result))
+    {
+        $event_types[] = $row['event_type'];
+    }
+    mysqli_free_result($result);
+    return $event_types;
+}
+
+// vim: sw=4 ts=4 expandtab
--- a/pinc/DPage.inc
+++ b/pinc/DPage.inc
@@ -133,7 +133,7 @@ function project_add_page(
     // If the INSERT raised an error, don't do the other stuff.
     if ($errs) return $errs;
 
-    _log_page_event( $projectid, $image_file_name, 'add', $user, NULL, FALSE );
+    _log_page_event( $projectid, $image_file_name, 'add', $user, NULL );
 
     _project_adjust_n_pages( $projectid, +1 );
     _project_adjust_n_available_pages( $projectid, +1 );
@@ -170,7 +170,7 @@ function Page_delete( $projectid, $image, $user )
     $sql = "DELETE FROM $projectid WHERE image = '$image'";
     mysqli_query(DPDatabase::get_connection(), $sql);
 
-    _log_page_event( $projectid, $image, 'delete', $user, NULL, FALSE );
+    _log_page_event( $projectid, $image, 'delete', $user, NULL );
 
     _project_adjust_n_pages( $projectid, -1 );
     if ($page_was_available)
@@ -188,7 +188,7 @@ function Page_replaceImage( $projectid, $image, $new_image, $user )
         image='$new_image'
     ");
 
-    _log_page_event( $projectid, $image, 'replaceImage', $user, NULL, FALSE ); // $new_image
+    _log_page_event( $projectid, $image, 'replaceImage', $user, NULL ); // $new_image
 }
 
 // -----------------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


srjfoo:
As a db-req squirrel, it has been useful upon rare occasions to have everything that happens to a page logged. I'd honestly rather see these events continue to be logged, but perhaps just not counted?

cpeel:
That's exactly what this does. It continues to log these page events, it just doesn't update the `user_project_info` table for the user saying that they've "interacted" with the project.

srjfoo:
> That's exactly what this does. It continues to log these page events, it just doesn't update the `user_project_info` table for the user saying that they've "interacted" with the project.

Okay, then I misunderstood -- thanks for the clarification. :)

Looks good to me.

cpeel:
I've added a SETUP/upgrade script to update user_project_info to reflect the code update.

jmdyck:
> Don't update user_project_info.t_latest_page_event upon page checkout or return to round.

I'm not objecting, but it makes me a little uncomfortable. We can ignore these two classes of events to obtain a quantity that better supports *this* purpose, but what if there's another purpose that *wants* those events, or another purpose that needs to ignore other events? `review_work.php` is another user of `t_latest_page_event`: is it okay with this change?

Anyhow, with this change, some names will become somewhat misnomers:
- `t_latest_page_event` I can live with, but it might deserve a comment, though I'm not sure where.
- `$is_prooferly` should maybe be changed to `$update_t_latest_page_event`.

SETUP/upgrade/13/20190929_update_user_project_info.php
```diff
@@ -0,0 +1,111 @@
+<?php
+
+$relPath='../../../pinc/';
+include_once($relPath.'base.inc');
+include($relPath.'udb_user.php');
```
jmdyck:
```suggestion
include($relPath.'udb_user.php'); // $archive_db_name
```

(I wondered why the script would need that include.)

SETUP/upgrade/13/20190929_update_user_project_info.php
```diff
@@ -0,0 +1,111 @@
...
+    }
+}
+
+// ------------------------------------------------------------
+
+echo "\nDone!\n";
+
+// ------------------------------------------------------------
+
+function load_events_from_table($table, $username, $projectid)
```
jmdyck:
```suggestion
function get_distinct_event_types($table, $username, $projectid)
```
("from_table" puts undue emphasis on the table, when "for_user" and "for_project" are just as important. Might as well leave them all implicit in the function name.)

SETUP/upgrade/13/20190929_update_user_project_info.php
```diff
@@ -0,0 +1,111 @@
...
+    mysqli_free_result($result);
+
+    // for each project, get unique page_events for the project, if the only
+    // two kinds of event_types are 'checkedout' and 'returnToRound' set
+    // user_project_info.t_latest_page_event=0 for that (username, projectid).
+    foreach($projectids as $projectid)
+    {
+        // Process both the main page_events table and the archived
+        // page_events table.
+        $current_events = load_events_from_table("page_events", $username, $projectid);
```
jmdyck: suggest: $current_event_types, $archived_event_types

cpeel:
> > Don't update user_project_info.t_latest_page_event upon page checkout or return to round.
> 
> I'm not objecting, but it makes me a little uncomfortable. We can ignore these two classes of events to obtain a quantity that better supports _this_ purpose, but what if there's another purpose that _wants_ those events, or another purpose that needs to ignore other events? `review_work.php` is another user of `t_latest_page_event`: is it okay with this change?

`review_work.php` is more than OK with this change seeing as it calls the field `time_of_latest_save` in the resulting query -- but I understand your concern and your points are valid.

Perhaps it would be better to rename the column something like `t_latest_page_update` and tie it less closely to `page_events`. Even now we record `page_event` updates without setting this field, so it is already a misnomer. At least if we called it `t_latest_page_update` we're being already a bit vague on what sets this field.

Prior to this change, we update that field on the following events:
* Page_checkout
* Page_saveAsInProgress
* Page_saveAsDone
* Page_reopen
* Page_returnToRound
* Page_markAsBad

And NOT on:
* project_add_page
* Page_delete
* Page_replaceImage
* Page_replaceText
* Page_modifyText
* Page_reclaim
* Page_clearRound
* Page_eraseBadMark

Perhaps we should only be setting it on:
* Page_saveAsInProgress
* Page_saveAsDone

Then we could call the field `t_latest_page_save` -- because that's what we seem to be really wanting everywhere we use the field.

SETUP/upgrade/13/20190929_update_user_project_info.php
```diff
@@ -0,0 +1,111 @@
...
+        or die( mysqli_error(DPDatabase::get_connection()) );
+
+    $projectids = [];
+    while($row = mysqli_fetch_assoc($result))
+    {
+        $projectids[] = $row['projectid'];
+    }
+    mysqli_free_result($result);
+
+    // for each project, get unique page_events for the project, if the only
```
jmdyck: I suggest moving this comment to *inside* the `foreach` loop. Then make it something like:
```
Consider all of this user's page events for this project,
and fetch just the distinct event_type values for those events.
If the only event_types are (at most) 'checkedout' and 'returnToRound',
then this is a case where the new code in pinc/DPage.inc,
had it been in place back then, would never have updated t_latest_page_event.
So set user_project_info.t_latest_page_event=0 for this (username, projectid),
to achieve the same effect retroactively.
```

jmdyck:
> Perhaps we should only be setting it on:
>     * Page_saveAsInProgress
>     * Page_saveAsDone
> 
> Then we could call the field `t_latest_page_save` -- because that's what we seem to be really wanting everywhere we use the field.

To recap, this PR currently proposes that we ignore (for the purposes of t_latest_page_foo):
* checkout
* returnToRound

and now you're suggesting that we also ignore:
* reopen
* markAsBad

leaving only:
* saveAsInProgress
* saveAsDone

as the event-types that cause `t_latest_page_foo` to be updated.

Seems plausible. @srjfoo, what do you think?

srjfoo:
> > Perhaps we should only be setting it on:
> > * Page_saveAsInProgress
> > * Page_saveAsDone
> > Then we could call the field `t_latest_page_save` -- because that's what we seem to be really wanting everywhere we use the field.
> 
> To recap, this PR currently proposes that we ignore (for the purposes of t_latest_page_foo):
> 
>     * checkout
> 
>     * returnToRound
> 
> 
> and now you're suggesting that we also ignore:
> 
>     * reopen
> 
>     * markAsBad
> 
> 
> leaving only:
> 
>     * saveAsInProgress
> 
>     * saveAsDone
> 
> 
> as the event-types that cause `t_latest_page_foo` to be updated.
> 
> Seems plausible. @srjfoo, what do you think?

It just occurred to me to wonder -- if someone marks a page as bad, should that perhaps be counted? That's not quite the same as just returning a page to the round, even if they don't do another page in the project. However, I don't feel strongly about it -- if the only thing they did was mark a page bad, and then didn't do any other pages in the project, perhaps it *is* best to ignore it.

As long as the only things needed by anything using `t_latest_page_foo` are:
> 
>     * saveAsInProgress
> 
>     * saveAsDone
> 
I'm good.

SETUP/upgrade/13/20190929_update_user_project_info.php
cpeel: I've update the wording based largely on what you have here. I am leaving it outside the foreach, however, as that's how documenting blocks makes the most sense to me (I know you and I differ on this).

cpeel:
Thinking about it from a proofer's perspective -- what actions would you want to do on a project and still have it listed as if you "worked on it"? `saveAsDone`, sure. `saveAsInProgress`, probably. `markAsBad`, maybe? -- presumably you did invest enough in the project/page to indicate it was bad.

I don't feel strongly about `markAsBad` but I do think narrowing the events that count as "worked on" to the set that most users expect makes sense while we're here.

I've researched a bit on the effort to rename a column. MySQL's documentation and the internet at large is unclear if simply renaming a column in the table will result in data shuffle (which we want to avoid) or just a simple record-keeping schema update (which would make sense). Note that updating the column name will also break all TEST sandboxes until they are updated. These two things coupled with the fact that the `t_latest_page_event` is already mis-named because it isn't actually for *all* page_events makes me lean to not updating the name, but rather documenting -- somewhere -- what it is used for.

srjfoo:
> Thinking about it from a proofer's perspective -- what actions would you want to do on a project and still have it listed as if you "worked on it"? `saveAsDone`, sure. `saveAsInProgress`, probably. `markAsBad`, maybe? -- presumably you did invest enough in the project/page to indicate it was bad.
> 
> I don't feel strongly about `markAsBad` but I do think narrowing the events that count as "worked on" to the set that most users expect makes sense while we're here.
> 
I agree that narrowing the events that count as "worked on" is a good idea. I have a slight preference for including `markAsBad`, because the proofer made the decision to mark the page as bad -- and did *not* return it to the round, so they did take an action concerning the page, even if they did not get credited for a save on it.

On the other hand, once a bad page is fixed and returned to available for a proofer to check out, the chances are that same proofer will not get that same page, so I can see justifying it either way. @jmdyck, do you have a preference? And perhaps we should document why that choice was made?

There will be no indication in the page details that the page was ever checked out to that user, once the bad pages is cleared, but neither will there be evidence in the page details if the proofer saved in progress and then tossed it back into the round or had the page reclaimed.

jmdyck:
> 
> I agree that narrowing the events that count as "worked on" is a good idea. I have a slight preference for including `markAsBad`, because the proofer made the decision to mark the page as bad -- and did _not_ return it to the round, so they did take an action concerning the page, even if they did not get credited for a save on it.
> 
> On the other hand, once a bad page is fixed and returned to available for a proofer to check out, the chances are that same proofer will not get that same page, so I can see justifying it either way.

Whether or not the proofer gets the same page again shouldn't matter, as `t_latest_page_event` covers all pages in the project.

So the main question is: if you don't do *any* saves (in-progress or as-done) in a project, but you do mark one or more pages as bad, should that project appear in "My Projects"?

This case seems pretty unlikely. (I suppose we could troll through page_events to see if it's ever happened.) My *guess* is that, if you asked proofers whether they'd expect/want such a project to appear in "My Projects", you wouldn't get a strong consensus, or much concern.

(There's a secondary question: if you *do* save some pages, but the most *recent* thing that you did was mark a page bad, should the time of that event be used, e.g. for sorting projects? But that's subtler and probably even less of a concern.)

> @jmdyck, do you have a preference?

I have a slight preference for ignoring markAsBad, mainly because that makes the semantics easier to describe, both in the code and on the site. (e.g., "This table lists all the projects in which you have saved a page, either in-progress or as-done." without having to tack on "or marked a page as bad".)

srjfoo:
> I have a slight preference for ignoring markAsBad, mainly because that makes the semantics easier to describe, both in the code and on the site. (e.g., "This table lists all the projects in which you have saved a page, either in-progress or as-done." without having to tack on "or marked a page as bad".)
Let's ignore it, then.

jmdyck:
We should probably get GM approval.

cpeel:
@lhamilton1 - can you chime in on the discussion, please?

lhamilton1:
I approve. Save as done and save in progress should cover things.

cpeel:
Code has been updated so that the column is only updated on page saves (saveAsDone, saveAsInProgress). In doing so I pulled the field update out of the `_log_page_events()` function and pulled `upi_set_t_latest_page_event()` directly into the two functions that call it. This should make it much clearer what is going on. I've documented the two table timestamps in the `pinc/user_project_info.inc` file which seemed as good a place as any.

PR #134, opened 2019-09-02
Fix two bugs caused by unshelling zip

mlazaric:
Descriptions of the bugs: https://www.pgdp.net/c/tasks.php?task_id=1767
Deployed at: https://www.pgdp.org/~mlazaric/c/


pinc/misc.inc
```diff
@@ -596,6 +596,21 @@ function get_filelist($dirpath, $validext = array(), $with_path = False)
     return $filelist;
 }
 
+function flatten_directory(string $directory_to_flatten)
```
cpeel: Please add a comment on what this function is intended to do.

SETUP/tests/misc_FlattenDirectoryTest.php
```diff
@@ -0,0 +1,33 @@
...
+
+    /**
+     * @expectedException InvalidArgumentException
+     */
+    public function testFlatteningNonExistentDirectory() {
+        flatten_directory('data/non_existent');
+    }
+
+    public function testFlatteningDirectoryWithNestedSubdirectories() {
+        flatten_directory('data/nested_directories');
```
cpeel: Tests should not alter files (or directories) committed into the repo. Instead, create (and clean up) the files/directories within the testcase itself.

mlazaric: Good catch, too bad PHP doesn't have a nice recursive delete function. :)

PR #136, opened 2019-09-14
Update proofing interface help for character picker

cpeel:
Update the proofreading interface help for the new character picker. Wording by @srjfoo.


70ray:
Should we say also something like "Clicking on a character will insert it into the text, hovering over a character will show an enlarged view of it in the box on the left and a tooltip text will appear describing the character".

srjfoo:
Yes, I like that addition, thanks.

cpeel:
Great idea - that wording has been added.

PR #137, opened 2019-09-15
Remove symbols before preview and restore after

70ray:
This fixes a problem with preview when the line-wrap symbols are shown.


70ray:
There is a better way to do this and another problem needs fixing

PR #138, opened 2019-09-22
Remove line-end symbols before preview and 'show all text'

70ray:


70ray:
Unfortunately this is done in a different way in the two cases because dp_proof.js is in a different frame from mark_wrap.js

70ray:
Having found yet another problem I think we should abandon the end of line marker code altogether so this fix will not be needed

PR #141, opened 2019-09-30
Backend updates for My Projects work

cpeel:
When working on the new My Projects updates, various backend changes were made to support the new code. I pulled these out into their own commit to reduce the review overhead.

The two big changes in this MR:
* New tabbed interface for the Prefs and PM page.
* Move PP/PPV logic into Project class and out of the project.php page.

### DIFF
--- a/pinc/Project.inc
+++ b/pinc/Project.inc
@@ -78,10 +78,14 @@ class Project
                 return $this->can_be_managed_by_user($pguser);
             case "names_can_be_seen_by_current_user":
                 return $this->names_can_be_seen_by_user($pguser);
+            case "PPer":
+                return $this->_get_PPer();
+            case "PPVer":
+                return $this->_get_PPVer();
             case "PPer_is_current_user":
-                return $this->PPer_is_user($pguser);
+                return is_null($pguser) ? False : $this->PPer == $pguser;
             case "PPVer_is_current_user":
-                return $this->PPVer_is_user($pguser);
+                return is_null($pguser) ? False : $this->PPVer == $pguser;
         }
 
         // If we don't recognize the property, raise a notice and return null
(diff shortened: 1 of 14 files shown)


pinc/Project.inc
```diff
@@ -133,40 +137,85 @@ class Project
...
+            $PPVer = $this->ppverifier;
+        }
+        else if ( !empty($this->checkedoutby) &&
+            in_array(
+                $this->state,
+                [
+                    PROJ_POST_SECOND_CHECKED_OUT,
+                    PROJ_POST_COMPLETE,
+                    PROJ_SUBMIT_PG_POSTED,
+                ]
```
srjfoo: This says pretty much the same thing as `private function _get_PPer()`, but `_get_PPer()` assigns the acceptable states to `$checkedout_states`. Functionality is the same, and I realize that the code was just moved from one place to another, but should they follow the same pattern?

cpeel: Yes, it would be better to have them more closely match. Fixed.

cpeel:
@srjfoo, can you give one final look over the following two pages for how the tabs look and work?
* [PM page](https://www.pgdp.org/~cpeel/c/tools/project_manager/projectmgr.php)
* [Prefs page](https://www.pgdp.org/~cpeel/c/userprefs.php)

srjfoo:
> @srjfoo, can you give one final look over the following two pages for how the tabs look and work?
> 
>     * [PM page](https://www.pgdp.org/~cpeel/c/tools/project_manager/projectmgr.php)
> 
>     * [Prefs page](https://www.pgdp.org/~cpeel/c/userprefs.php)

They appear to look and work as the tabs for `My Projects` do, and I don't see a difference in how the old tabs worked for `Prefs` and the updated version works. :)

PR #144, opened 2019-10-07
Add query to count PPV projects over threshold

cpeel:
Add a function to return the count of outstanding PPV projects over
the threshold, much like we have for PP. This is currently only
used in a noncvs script.


cpeel:
@srjfoo - I've updated the variable names and think this is ready for merge.

PR #145, opened 2019-10-07
Text format

70ray:
This adds a "text" mode where inline markup (i, b, f g,u)  is replaced by a single character to show how text will appear in the "plain text" version of the book to get correct alignment of tables etc. Small caps markup is not marked in this way. "Wrap" is now a checkbox so that the Text mode can be used with or without re-wrapping.


cpeel:
This LGTM based on the code review and my testing. Am interested in getting @srjfoo's validation thoughts as well.

srjfoo:
I've tested it, too. And it looks good to me!

My only concern is the way small caps is treated, but it may be irrelevant in this case. Since this is so that the formatters can check alignment for things where alignment is an issue, and there's no one accepted way to do small caps in PP, there's no good way to predict what the PPer will want.

I'd say go with it!

PR #146, opened 2019-10-12
Auto-commit DP updates

cpeel:
The DP code doesn't (currently) use transactions. For systems that have moved to use MySQL engine backends that support transactions (i.e. InnoDB) we should auto commit updates.

This change improves updates on PROD since we have moved some tables (like page_events and user_project_info) to InnoDB tables.

### DIFF
--- a/pinc/DPDatabase.inc
+++ b/pinc/DPDatabase.inc
@@ -28,6 +28,9 @@ final class DPDatabase
         $sql = "SET SESSION sql_mode=(SELECT REPLACE(@@sql_mode,'ONLY_FULL_GROUP_BY',''));";
         if(!mysqli_query(self::$_connection, $sql))
             throw new DBConnectionError("Unable to set sql_mode");
+
+        // Autocommit DB updates
+        mysqli_autocommit(self::$_connection, TRUE);
     }
 
     static public function close()


pinc/DPDatabase.inc
```diff
@@ -28,6 +28,9 @@ final class DPDatabase
         $sql = "SET SESSION sql_mode=(SELECT REPLACE(@@sql_mode,'ONLY_FULL_GROUP_BY',''));";
         if(!mysqli_query(self::$_connection, $sql))
             throw new DBConnectionError("Unable to set sql_mode");
+
+        // Autocommit DB updates
+        mysqli_autocommit(self::$_connection, TRUE);
```
jmdyck: [The MySQL docs](https://dev.mysql.com/doc/refman/5.6/en/innodb-autocommit-commit-rollback.html) say:
> By default, MySQL starts the session for each new connection with autocommit enabled

So it seems like this statement is unnecessary, but I guess it can't hurt.

Also, this ignores the function's return value.Which is maybe okay, if we're not expecting the call to do anything?

cpeel: Yup, thanks - I missed that. I also confirmed that autocommit is being set to 1 already, so this change isn't needed.

PR #148, opened 2019-10-18
Hide picker

70ray:
Character picker can be shown or hidden. State persists using local storage.

### DIFF
--- a/pinc/ProofreadingToolbox.inc
+++ b/pinc/ProofreadingToolbox.inc
@@ -33,7 +33,11 @@ class ProofreadingToolbox
     {
         echo "<div class='display-flex' id='toolbox'>";
         $character_selector = new CharacterSelector;
+        echo "<div id='char-picker'>";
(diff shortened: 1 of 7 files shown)


pinc/ProofreadingToolbox.inc
```diff
@@ -33,7 +33,11 @@ class ProofreadingToolbox
     {
         echo "<div class='display-flex' id='toolbox'>";
         $character_selector = new CharacterSelector;
+        echo "<div id='char-picker'>";
+        echo "<button type='button' id='hide-picker' class='hide-picker-button' title='", _("Hide the character picker"), "'>&lang;</button>";
```
cpeel: Please wrap this in `html_attr()`: `html_attr(_("Hide the character picker"))`. Ditto the show string.

styles/layout.less
```diff
@@ -775,6 +775,21 @@ table.dirlist {
     #char-selector {
         min-width: 430px;
     }
+
+    #char-picker {
+        display: inherit;
+    }
+
+    .hide-picker-button {
+        font-size: 2.5em;
```
cpeel: Let's make this 2 please.

PR #150, opened 2019-10-22
New proof 1

70ray:
This simplifies the javascript code for inserting and transforming text. It removes the complexity to handle very old browsers.
It fixes the issues in tasks 1747 and 1749 and should simplify things we may want to do in the future such as retaining the selection and character insertion in wordcheck.
It should not otherwise make any difference to anything.


cpeel:
This LGTM! I'd like to get this merged in right after we do the next picker-focused push to PROD so that it gets a little bake time in TEST.

PR #151, opened 2019-10-23
Character picker updates

cpeel:
This breaks the symbol character picker into two menus making the overall picker size smaller. It also right-aligns the toolbox tools.


lhamilton1:
I approve

PR #152, opened 2019-10-28
lint config file added

70ray:
This is a fairly basic file.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,24 @@
+module.exports = {
+    env: {
+        browser: true,
+        es6: true,
+    },
+    extends: "eslint:recommended",
+    globals: {
+        Atomics: 'readonly',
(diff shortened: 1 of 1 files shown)


cpeel:
Can we add this to our `Makefile` to have all `.js` files linted? That would also require getting all of the JS files to pass the linter and the Travis config updated to be able to run it -- so it's a bit of a lift -- but  think that's where we want to get to ultimately and thought I'd ask.

It's OK if we want to start with just this and then iterate until we get to full valid JS file linting later.

70ray:
Good plan to check automatically but the older js files give lots of errors: mostly variables not defined or not used. It should be quite easy to fix these. So it would be useful to have the config file merged so it is available while we fix them.

cpeel:
I think I'd rather get this in and have a separate effort to bring the old code in-line. Thanks for getting this going!

PR #154, opened 2019-11-04
Char check

70ray:


styles/layout.less
```diff
@@ -951,5 +951,47 @@ table.search-column {
     grid-row-gap: 2px;
 }
 
+.flex_container {
```
cpeel: Please add a section header like that on lines 944-945.

tools/proofers/char_check.inc
```diff
@@ -0,0 +1,30 @@
+<?php
```
cpeel: We might want to globally rename this from "character check" to "codepoint validator". My worry is that "character check" brings to mind a punctuation checker which this is not.

styles/layout.less
70ray: Also removed unused parts which were pasted in from another branch

PR #155, opened 2019-11-12
Sr1601 1

70ray:
This enables Smooth reading files to be downloaded or viewed individually.

### DIFF
--- a/pinc/Project.inc
+++ b/pinc/Project.inc
@@ -687,24 +687,26 @@ XML;
                $this->smoothread_deadline > time();
     }
 
-    private function delete_file($file)
+    private function delete_file($path)
     {
-        // ensure $file is inside $this->dir
-        if(!startswith(realpath($file), realpath($this->dir)))
+        // ensure $path is inside $this->dir
+        if(!startswith(realpath($path), realpath($this->dir)))
         {
-            throw new UnexpectedValueException("$file is not in $this->dir");
+            throw new UnexpectedValueException("$path is not in $this->dir");
         }
-        if(is_dir($file))
+        if(is_dir($path))
         {
-            foreach(glob("$file/*") as $sub_file)
+            // find also hidden files
+            $files = array_diff(scandir($path), array('.','..'));
+            foreach($files as $sub_file)
             {
-                $this->delete_file($sub_file);
+                $this->delete_file("$path/$sub_file");
             }
-            rmdir($file);
+            rmdir($path);
         }
         else
         {
-            unlink($file);
+            unlink($path);
         }
     }
 
--- a/pinc/Project.inc
+++ b/pinc/Project.inc
@@ -686,6 +686,31 @@ XML;
         return $this->state == PROJ_POST_FIRST_CHECKED_OUT &&
                $this->smoothread_deadline > time();
     }
+
+    private function delete_file($file)
+    {
+        if(is_dir($file))
+        {
+            foreach(glob("$file/*") as $sub_file)
+            {
+                $this->delete_file($sub_file);
+            }
+            rmdir($file);
+        }
+        else
+        {
+            unlink($file);
+        }
+    }
+
+    public function delete_smoothreading_dir()
+    {
+        $dir = "$this->dir/smooth";
+        if(file_exists($this->dir) && file_exists($dir))
+        {
+            $this->delete_file($dir);
+        }
+    }
 }
 
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
--- a/pinc/Project.inc
+++ b/pinc/Project.inc
@@ -1,7 +1,7 @@
 <?php
 include_once($relPath.'user_is.inc');
 include_once($relPath.'project_states.inc');
-include_once($relPath.'misc.inc'); // html_safe(), xmlencode()
+include_once($relPath.'misc.inc'); // html_safe(), xmlencode(), startswith()
 include_once($relPath.'forum_interface.inc'); // topic_create
 include_once($relPath.'SettingsClass.inc');
 include_once($relPath.'User.inc');
--- a/pinc/Project.inc
+++ b/pinc/Project.inc
@@ -689,6 +689,11 @@ XML;
 
     private function delete_file($file)
     {
+        // ensure $file is inside $this->dir
+        if(!startswith(realpath($file), realpath($this->dir)))
+        {
+            throw new UnexpectedValueException("$file is not in $this->dir");
+        }
         if(is_dir($file))
         {
             foreach(glob("$file/*") as $sub_file)
--- a/pinc/misc.inc
+++ b/pinc/misc.inc
@@ -800,13 +800,15 @@ function get_upload_err_msg($upload_error_code)
 // Represent $n bytes as a  $m kB-- $m PB string
 function humanize_bytes($n)
 {
-    $fmt   = "%d B";
-    $units = array("PB", "TB", "GB", "MB", "kB");
-    while($n >= 1024) {
-        $n   /= 1024.0;
-        $fmt  = "%.2f " . array_pop($units);
+    $i = 0;
+    $units = array("B", "KiB", "MiB", "GiB", "TiB", "PiB");
+    while($n >= 1024)
+    {
+        $i += 1;
+        $n /= 1024.0;
     }
-    return sprintf($fmt, $n);
+    $format = (($i == 0) ? "%d " : "%.2f ") . $units[$i];
+    return sprintf($format, $n);
 }
 
 // -----------------------------------------------------------------------------
--- a/pinc/misc.inc
+++ b/pinc/misc.inc
@@ -1015,6 +1015,18 @@ function is_url_for_current_page($url)
     return TRUE;
 }
 
+function return_bytes($size_str)
+// from http://php.net/manual/en/function.ini-get.php
+{
+    switch (substr($size_str, -1))
+    {
+        case 'M': case 'm': return (int)$size_str * 1048576;
+        case 'K': case 'k': return (int)$size_str * 1024;
+        case 'G': case 'g': return (int)$size_str * 1073741824;
+        default: return $size_str;
+    }
+}
+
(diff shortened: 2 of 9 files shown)


project.php
```diff
@@ -1840,52 +1836,50 @@ function do_smooth_reading()
                 "<b>$sr_deadline_str</b>"
             );
 
-            echo "<li>";
-            echo $sr_sentence;
-            echo "</li>\n";
+            echo "<p class='pa'>", $sr_sentence, "</p>";
```
cpeel: If you're putting this on one line, just embed the variable into the string directly:
```
echo "<p class='pa'>$sr_sentence</p>";
```

project.php
```diff
@@ -1974,6 +1969,120 @@ function sr_echo_time_form($label, $min_days, $max_days, $default_days, $extend
     echo "</form>\n";
 }
 
+function echo_smoothreading_options($project)
+{
+    global $projects_url;
+
+    $smooth_dir = "$project->dir/smooth";
+    $project_url = "$projects_url/$project->projectid";
```
cpeel: This is the same as `$project->url`

project.php
```diff
@@ -1974,6 +1969,120 @@ function sr_echo_time_form($label, $min_days, $max_days, $default_days, $extend
...
+
+    // download zipped html
+    $html_files = glob("$smooth_dir/*.{htm,html}", GLOB_BRACE);
+    if($html_files)
+    {
+        // assume only one html file
+        $file = $html_files[0];
+        $file_base_name = basename($file);
+        // zip the html file with images folder for download
+        $zip = new ZipArchive;
```
cpeel: This will recreate the .zip archive on every page load, which we shouldn't be doing. We should create the .zip file of the HTML on upload and only display it on page load.

project.php
```diff
@@ -1974,6 +1969,120 @@ function sr_echo_time_form($label, $min_days, $max_days, $default_days, $extend
...
+    if($files)
+    {
+        echo "<li class='list-head'>", _("Open in browser (for reading only -- no annotations can be made):");
+        echo "<ul>";
+        foreach($files as $file)
+        {
+            $file_base_name = basename($file);
+            $url = "$smooth_url/$file_base_name";
+            echo "<li>";
+            echo "<a href='$url' target='_blank'>$file_base_name</a>";
```
cpeel: We should `html_safe()` the filename to be extra safe.

project.php
```diff
@@ -1974,6 +1969,120 @@ function sr_echo_time_form($label, $min_days, $max_days, $default_days, $extend
...
+        $file_base_name = basename($file);
+        echo_download_item($smooth_url, $file, $file_base_name, $file_base_name);
+    }
+}
+
+function echo_download_item($url, $file, $file_base_name, $text)
+{
+    $url = "$url/$file_base_name";
+    echo "<li>";
+    echo "<a href='$url' download='$file_base_name'>$text</a>";
```
cpeel: We should escape both `$file_base_name` and `$text` to be safe.

styles/layout.less
```diff
@@ -961,5 +961,22 @@ table.search-column {
     grid-row-gap: 2px;
 }
 
+/* ------------------------------------------------------------------------ */
+/* Project Smoothreading */
+
+.h2a {
```
cpeel: These class names aren't at all intuitive. Please rename them to something clearer.

tools/upload_text.php
```diff
@@ -488,4 +515,24 @@ function ensure_path_is_unused( $path )
     }
 }
 
+function delete_contents($dir)
```
cpeel: To make this less dangerous, consider moving it into the Project class as `delete_smoothreading_dir()`. Then validate that the project's directory, and then smooth directory exists. That both limits how this function will be used (to prevent future programmer error) and moves the context closer to the object it is working on (the Project).

project.php
70ray: Good point.

pinc/Project.inc
```diff
@@ -686,6 +686,31 @@ XML;
         return $this->state == PROJ_POST_FIRST_CHECKED_OUT &&
                $this->smoothread_deadline > time();
     }
+
+    private function delete_file($file)
+    {
+        if(is_dir($file))
```
cpeel: Let's add a check to ensure that `$file` is always inside `$this->dir` and throw an UnexpectedValueException if it isn't. Maybe something like (totally untested):
```php
// ensure $file is inside $this->dir
if(! startswith(realpath($file), realpath($this->dir)))
{
    throw new UnexpectedValueException("$file not in $this->dir");
}
```
realpath ensures that no funny path business is going on that might do something outside the directory we're expecting.

project.php
```diff
@@ -1971,54 +1971,30 @@ function sr_echo_time_form($label, $min_days, $max_days, $default_days, $extend
...
     echo "<li class='list-head'>", _("Download a Smooth Reading file");
     echo "<ul>";
     echo_file_downloads(glob("$smooth_dir/*.txt"), $smooth_url);
 
-    // download zipped html
-    $html_files = glob("$smooth_dir/*.{htm,html}", GLOB_BRACE);
-    if($html_files)
+    // zipped htm(l) file
+    $zip_files = glob("$smooth_dir/*.zip");
+    foreach(glob("$smooth_dir/*.zip") as $zip_file) // only one
```
cpeel: What does "only one" mean? What happens if/when there is more than one for some unexpected reason?

tools/upload_text.php
```diff
@@ -269,32 +269,48 @@
...
             {
                 $zip->extractTo($smooth_dir);
                 $zip->close();
+                // if there is a html or htm file zip it (with images) for download
+                $html_files = glob("$smooth_dir/*.{htm,html}", GLOB_BRACE);
+                if($html_files)
+                {
+                    // assume only one html file
+                    $file = $html_files[0];
+                    $zip = new ZipArchive;
```
cpeel: You might consider using `create_zip_from()` in `misc.inc`. That will remove all the paths and have everything within the zip in a single directory, which may or may not be what you want.

tools/upload_text.php
```diff
@@ -269,12 +269,55 @@
...
+        if ($stage == 'smooth_avail')
+        {
+            $project->delete_smoothreading_dir();
+            $smooth_dir = "$project->dir/smooth";
+            if(!mkdir($smooth_dir))
+            {
+                die("Could not create smooth directory");
+            }
+            // unzip into smooth folder
+            $zip = new ZipArchive;
```
cpeel: Consider using `extract_zip_to()` in `misc.inc`.

project.php
70ray: This comment was meant to indicate that we knew that there was not more than one because we made only on on upload so it was shorter to write "foreach ..." than testing if there were any and then selecting one, but as you say there could be more for an unexpected reason such as if the PP had uploaded another zip file inside his main zip. This does raise some other issues so we need to discuss it further. I'll mention it on Slack.

tools/upload_text.php
70ray: Yes, this should work.

70ray:
I also saw the humanize_bytes() function in misc.inc which we could use (and modernise since KiB etc. seems to be preferred now)

pinc/misc.inc
```diff
@@ -800,13 +800,15 @@ function get_upload_err_msg($upload_error_code)
 // Represent $n bytes as a  $m kB-- $m PB string
 function humanize_bytes($n)
 {
-    $fmt   = "%d B";
-    $units = array("PB", "TB", "GB", "MB", "kB");
-    while($n >= 1024) {
-        $n   /= 1024.0;
-        $fmt  = "%.2f " . array_pop($units);
+    $i = 0;
+    $units = array("B", "KiB", "MiB", "GiB", "TiB", "PiB");
```
cpeel: https://ux.stackexchange.com/questions/13815/files-size-units-kib-vs-kb-vs-kb wraps up my thoughts on `KB` vs `KiB`, mostly that while it is technically more accurate, I suspect it's mostly going to just confuse people. Both Mac and Windows represent files in the former way and I think we should too. (that does mean fixing `kB` to `KB` however). I do prefer your updated logic and dropping the decimal if it's zero.

70ray: Fair enough, I have been using a Linux system recently since my windows computer broke down and it uses KiB etc. I'll change it.

cpeel: Thanks, and welcome to all-Linux land!

project.php
```diff
@@ -1840,52 +1836,50 @@ function do_smooth_reading()
...
+                echo_smoothreading_options($project);
                 // We don't allow guests to upload the results of smooth-reading.
                 global $user_is_logged_in;
                 if ( $user_is_logged_in )
                 {
-                    echo "<li>";
+                    echo "<li class='list-head'>";
                     echo "<a href='$code_url/tools/upload_text.php?project=$projectid&stage=smooth_done'>";
-                    echo _("Upload a smooth-read text") ;
+                    echo _("Upload a smoothread report") ;
```
srjfoo: This should be `Upload a smooth read report`. Trying to standardize on one form for display instead of three, and at some point, it went to two separate words, no hyphen.

project.php
```diff
@@ -1974,6 +1969,78 @@ function sr_echo_time_form($label, $min_days, $max_days, $default_days, $extend
...
+    if($files)
+    {
+        echo "<li class='list-head'>", _("Open in browser (for reading only -- no annotations can be made):");
+        echo "<ul>";
+        foreach($files as $file)
+        {
+            $file_base_name = basename($file);
+            $url = "$smooth_url/$file_base_name";
+            echo "<li>";
+            echo "<a href='$url' target='_blank'>", html_safe($file_base_name), "</a>";
```
srjfoo: Do we want to force opening in a new window/tab? My impression was that in general, that was discouraged, and that the user should be the one deciding whether to open a file in the current window, or in a new window or tab (whereas with `target='_blank'`, the user does not even have the option of opening something in the active window/tab).

project.php
70ray: I've changed this and some other similar cases.
There is an inconsistency where we have said "Upload a Smooth Read report" but "dowload Smooth Read _files_ uplaoded by xyz. The 'files' word is in a shared subroutine so more difficult to change.

project.php
70ray: I don't know.

cpeel:
Code-wise this looks good to me!

PR #157, opened 2019-11-13
Enable character pickers to work in wordcheck

70ray:
This implements task 1295 to enable the character pickers to work in the editable wordcheck boxes.


cpeel:
This is an incredibly elegant solution to the problem, very nice!!

PR #158, opened 2019-11-26
Blank option added at start of genres for new project

70ray:
This addresses task 1788
Using `&nbsp;` here and elsewhere rather than nothing to avoid validation errors.
Alternatively we could put some text like "Choose a genre"

### DIFF
--- a/tools/project_manager/edit_common.inc
+++ b/tools/project_manager/edit_common.inc
@@ -56,7 +56,7 @@ function language_list($language)
     echo _("Primary"), ": <select name='pri_language' required>\n";
     maybe_echo_placeholder_option($pri_language);
     echo_language_options($pri_language);
-    echo "</select>\n";
+    echo "</select><br>\n";
     echo _("Secondary"), ": <select name='sec_language'>\n";
     echo "<option value=''>&nbsp;</option>\n";
     echo_language_options($sec_language);
--- a/tools/project_manager/edit_common.inc
+++ b/tools/project_manager/edit_common.inc
@@ -53,29 +53,26 @@ function language_list($language)
         $sec_language = '';
     }
 
-    $array_list = $GLOBALS['lang_list'];
+    global $lang_list;
 
-    echo "<select name='pri_language' required>";
-    echo "<option value=''>" . _("Primary Language") . "</option>";
-    echo "\n";
-    for ($i=0;$i<count($array_list);$i++)
+    echo _("Primary"), ": <select name='pri_language' required>\n";
+    maybe_echo_placeholder_option($pri_language);
+    echo_language_options($lang_list, $pri_language);
+    echo "</select>\n";
+    echo _("Secondary"), ": <select name='sec_language'>\n";
+    echo "<option value=''>&nbsp;</option>\n";
+    echo_language_options($lang_list, $sec_language);
+    echo "</select>\n";
+}
+
+function echo_language_options($lang_list, $default)
+{
+    foreach($lang_list as $language)
     {
-        echo "<option value='".attr_safe($array_list[$i]['lang_name'])."'";
-        if ($pri_language == $array_list[$i]['lang_name']) { echo " SELECTED"; }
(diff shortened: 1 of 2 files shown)


tools/project_manager/edit_common.inc
```diff
@@ -78,7 +78,8 @@ function language_list($language)
 function genre_list($genre)
 {
     $genres = load_genre_translation_array();
-    echo "<select name='genre'>";
+    echo "<select name='genre'>\n";
+    echo "<option value=''>&nbsp;</option>\n";
```
cpeel: For the fields that require a value, we might consider doing something like this instead:
```
"<option value='' disabled>" . _("Select one") . "</option>"\n
```

That way users know what fields are required by looking at them instead of clicking save and changing something until it works.

70ray: Good idea, I'll do that.

srjfoo: That's a good idea.

Would it be reasonable to put shadow text in the "Name of Work" and "Author's Name" input boxes that says "required field". Not that anyone is likely to try to leave at least the title field out. The four fields that currently will generate an error are Name of Work, Author, Project manager, and Primary Language (and once this gets fixed, Genre).

Project manager is auto-filled with the username of the PM who created the project, so the only time that error is triggered is when a squirrel removes the username without adding another.

70ray: Giving the <select> and <input> a required attribute has the required effect.

tools/project_manager/edit_common.inc
```diff
@@ -50,7 +55,7 @@ function language_list($language)
 
     $array_list = $GLOBALS['lang_list'];
 
-    echo "<select name='pri_language'>";
+    echo "<select name='pri_language' required>";
     echo "<option value=''>" . _("Primary Language") . "</option>";
```
cpeel: Looking at this afresh it feels weird for the two language fields to specify Primary/Secondary in the drop-down. I feel like this would be better done like:
```
Primary: [ Select One  ]   Secondary: [            ]
```
then we could put `Select One` in the first one and leave the second one blank like we do the other optional fields.

And yes, I realize we're well into feature creep here, so feel free to punt as this could be done later.

tools/project_manager/edit_common.inc
```diff
@@ -48,37 +53,35 @@ function language_list($language)
         $sec_language = '';
     }
 
-    $array_list = $GLOBALS['lang_list'];
+    echo _("Primary"), ": <select name='pri_language' required>\n";
+    maybe_echo_placeholder_option($pri_language);
+    echo_language_options($pri_language);
+    echo "</select>\n";
```
cpeel: Can you please add a `<br>` here to put these each on their own line?

tools/project_manager/editproject.php
```diff
@@ -855,7 +855,7 @@ function show_form()
         echo     "<input type='submit' name='save' value='".attr_safe(_("Save"))."'>";
         echo     "<input type='submit' name='saveAndQuit' value='".attr_safe(_("Save and Go To PM Page"))."'>";
         echo     "<input type='submit' name='saveAndProject' value='".attr_safe(_("Save and Go To Project"))."'>";
-        echo     "<input type='submit' name='quit' value='".attr_safe(_("Quit Without Saving"))."'>";
+        echo     "<input type='submit' name='quit' formnovalidate value='".attr_safe(_("Quit Without Saving"))."'>";
```
cpeel: Ooohhh, good catch!

cpeel:
Except for my one comment about adding the `<br>`, this LGTM. @srjfoo - have you had a chance to test this?

srjfoo:
Looks good to me! (Apologies if this shows up twice -- I thought I just added a review, but am not seeing it anywhere.)

tools/project_manager/edit_common.inc
70ray: I've done this now.

srjfoo: Looks good to me.

PR #161, opened 2019-11-30
Specify a font-family to fall back to

cpeel:
If the browser doesn't have the specified font, fall back to something sensible.

This is most important for Safari and Chrome on macOS which don't show textareas in monospace fonts.


cpeel:
> Seems alright. We might need to do something more complicated in the future to define a configurable font chain.

I'm not certain that we will, actually. In Unicode-land I can see us specifying a string of fonts to fall back on, absolutely, but we probably don't need it configured per-project. For instance:
```
Noto Sans Mono, Some Other Mono Font With More Glyphs In It, monospace
```

70ray:
Yes, in any case falling back to monospace is good.

srjfoo:
Most definitely!

PR #163, opened 2019-12-01
convert diacritical markup to unicode character

70ray:


tools/proofers/codepoint_validator.js
```diff
@@ -58,9 +62,67 @@ $(function () {
...
+    var above = {
+        "=": "\u0304", // macron
+        ":": "\u0308", // diaeresis
+        ".": "\u0307", // dot
+        "`": "\u0300", // grave
+        "'": "\u0301", // acute
+        "^": "\u0302", // circumflex
+        ")": "\u0306", // breve
+        "~": "\u0303", // tilde
+        ",": "\u0327", // cedilla
```
cpeel: I think this should be `\u0309` for the "hook above" instead of cedilla again.

tools/proofers/codepoint_validator.js
```diff
@@ -58,9 +62,67 @@ $(function () {
...
 
     $("#cc-remove").click(function () {
-        let textarea = $("#text_data");
-        textarea.val(textarea.val().replace(pattern , ""));
+        textArea.value = textArea.value.replace(badPattern , "");
         $("#checker").hide();
         $("#proofdiv").show();
     });
+
+    var above = {
```
cpeel: We might also consider adding the following too, which would extend the current set of the DP diacritical markup:
```
        "*": "\u030A", // ring
        "\\": "\u0338", // stoke or bar
        "(": "\u0311", // inverted breve
```

tools/proofers/codepoint_validator.js
```diff
@@ -58,9 +62,67 @@ $(function () {
...
+        ".": "\u0307", // dot
+        "`": "\u0300", // grave
+        "'": "\u0301", // acute
+        "^": "\u0302", // circumflex
+        ")": "\u0306", // breve
+        "~": "\u0303", // tilde
+        ",": "\u0327", // cedilla
+        "v": "\u030C", // caron
+    };
+    var below = {
```
cpeel: We might also consider adding the following (again, extending our current set):
```
        "*": "\u0325", // ring
        "\\": "\u0338", // stoke or bar
        "(": "\u032F", // inverted breve
```

tools/proofers/codepoint_validator.js
```diff
@@ -33,18 +33,22 @@ $(function () {
             charClass += utf8Chr(code[0]) + "-" + utf8Chr(code[1]);
         }
     });
-    var pattern = new RegExp("[^" + charClass + "]", "ug");
+
+    var badPattern = new RegExp("[^" + charClass + "]", "ug");
```
cpeel: Do you need two of these? I would think that doing a global for where you use badPattern1 would also work.

tools/proofers/codepoint_validator.js
```diff
@@ -58,9 +62,67 @@ $(function () {
...
+        "`": "\u0316", // grave
+        "'": "\u0317", // acute
+        "^": "\u032D", // circumflex
+        ")": "\u032E", // breve
+        "~": "\u0330", // tilde
+        ",": "\u0327", // cedilla
+        "v": "\u032C", // caron
+    };
+
+    $(textArea).on("input", function() {
```
cpeel: Please add a comment within anonymous function on what this is doing.

tools/proofers/codepoint_validator.js
70ray: This was because of the lower case g with cedilla which appears above the letter. I can change it or add another symbol for hook.

tools/proofers/codepoint_validator.js
70ray: I thought it would work but on successive test functions it starts from the lastIndex from the previous test so doesn't find anything. I've now reset the lastIndex before test instead of using a different regex.

tools/proofers/codepoint_validator.js
70ray: I haven't added the stroke like this because it gives a stroke sloping the other way.

cpeel: I'm fine with other symbols, `/` or `-`. Unicode characters WITH STROKE go a few different directions: https://unicode-search.net/unicode-namesearch.pl?term=STROKE

@srjfoo - thoughts?

tools/proofers/codepoint_validator.js
cpeel: Thanks for the update, but I was thinking something a little more high-level, eg:
```
Detect if the user has input one of the diacritical markups [..]
and convert it to a Unicode codepoint. If the codepoint is not
valid for the project, leave it as the markup. This also converts
a few common ligatures to their Unicode codepoints.

This event fires on every input into the textarea, but only
takes action if the last character entered is a ].
```

tools/proofers/codepoint_validator.js
```diff
@@ -58,9 +61,96 @@ $(function () {
...
+        // if text changed and character before caret is ]
+        // (i.e. just inserted or pasted ] or deleted following character)
+        // check for markup like [..] for ligature or diacritical
+        // i0, i1, i3 point to these characters
+        var text = textArea.value;
+        var startPos = textArea.selectionStart;
+        var i0, i1, i3, uchar;
```
cpeel: Please use more descriptive names for these, like:
```
startIndex, char1Index, char2Index, endIndex
```

tools/proofers/codepoint_validator.js
```diff
@@ -58,9 +61,96 @@ $(function () {
...
+
+        i3 = startPos - 1;
+        if(text[i3] != "]") {
+            // if out of range would get ""
+            return;
+        }
+        i0 = startPos - 4;
+        if(text[i0] === "[") {
+            i1 = startPos - 3;
+            uchar = ligatures[text.slice(i1, i3)];
```
cpeel: Ligatures aren't guaranteed to be valid for the project's codepoints, we should validate those too. You could just move the validation into the substitute function.

tools/proofers/codepoint_validator.js
```diff
@@ -58,9 +61,96 @@ $(function () {
...
+    $(textArea).on("input", function() {
+        // if text changed and character before caret is ]
+        // (i.e. just inserted or pasted ] or deleted following character)
+        // check for markup like [..] for ligature or diacritical
+        // i0, i1, i3 point to these characters
+        var text = textArea.value;
+        var startPos = textArea.selectionStart;
+        var i0, i1, i3, uchar;
+
+        function substitute() {
```
cpeel: It's a bit unintuitive to have a function like this operate only "global" variables. Consider passing in the values you want the function to operate over:
```
function substitute(startIndex, replaceChar, startPos)
```

tools/proofers/codepoint_validator.js
70ray: Good point.

tools/proofers/codepoint_validator.js
70ray: Douglas Crockford says it is one of the 'very good' features of javascript that functions defined within another function have access to the outer function's local variables, but I agree with you that passing the parameters would make it more obvious how it is meant to work, although it could be less efficient, but probably no-one would notice the difference.

cpeel: IANAJavascriptDeveloper in any fashion, so if that's common practice that's fine with me.

tools/proofers/codepoint_validator.js
srjfoo: Is this similar to the cedilla vs. the hook issue discussed above, and which code is used? If there's a one-to-one correspondence in our code between the codepoint value and what we use for to stroke type, we probably need different symbols for each one. Otherwise, don't you need a lookup chart for which letters take what kind of stroke, and can't map the symbol the proofer types to just one code value that gets combined with the letter it modifies? (If I'm understanding the issue correctly, and I hope I'm saying it correctly.)

cpeel: There's a one-to-one correspondence on symbol to Unicode diacritical combining character, but how those are represented visually may differ depending on the character. So if `-` maps to STROKE then "capital L with stroke" will have a diagonal slash (Ł) while "capital H with stroke" will have a horizontal slash (Ħ).

IMHO, we should map a symbol to the Unicode diacritical combining character, not how it is visually represented in the glyph, which might also vary by font. That is most straightforward to implement and for people to reference by Unicode name.

srjfoo: If the Unicode diacritical combining character is the same, regardless of how it's represented in the glyph, then I don't see a problem with representing the "stroke" with a single character for our purposes. As for `/` fs `-`, which are we more likely to need for a different combining character?

70ray: I'm not sure stroke would be very useful. Ø can't be composed like this.

cpeel: Can stroke be used to compose the stroked letters in Extended Latin? I'm interested in adding that composite if it's generally useful, even if it's not useful for Basic Latin.

70ray: I checked the characters in extended latin that have 'stroke' in their description using a tool `https://minaret.info/test/normalize.msp` but none of them would decompose. I tested some with 'bar' and 'hook' which did not decompose either. This seems to imply you can't compose them from combining components either but  I'm not entirely sure about this.
'Ogonek' and 'horn' seem to always decompose so we could perhaps add those.
There are several characters with two diacritical marks. These can be entered as for example [=[~O]] or ['[ae]] or ['[c,]] or we could extend the code to accept e.g. [=~O] as well.

cpeel: So much for consistency in Unicode I guess. Thanks for verifying that. I think we should go with what you have right now. We can always tweak it more later.

srjfoo: Sounds good to me, too.

cpeel:
This LGTM, @70ray. Unless you have other changes planned for this I will merge it in!

tools/proofers/codepoint_validator.js
70ray: I don't want to change it any more now. As Casey says, we can tweak it later.

PR #166, opened 2019-12-06
Standardize on Smooth Reading

cpeel:
Here are all of the user-facing instances of `smooth reading`, `smooth-reading`, and `smoothreading` (even `smooth read`) that I found in the code.

I do wonder if in some of these places we should use `smooth read(ing)` with lowercase as the verb and `Smooth Reading` as a stage/noun. The former seems to read better alongside `proofread`/`proofreading` on the main page, for instance.


cpeel:
I rolled back the changes to `default.php` for the moment. I'd rather get those in along with the translated versions into French so the front page continues to be translated into French.

srjfoo:
Makes sense.

PR #169, opened 2019-12-11
Fix wording about character picker in help

cpeel:
Remove outdated paragraph in the proofreading interface help that discusses the old character selection UI and try to clarify the new character picker interface.

### DIFF
--- a/faq/prooffacehelp.php
+++ b/faq/prooffacehelp.php
@@ -670,16 +670,12 @@ Proofreading Guidelines for suggestions).
 
 <p>
 As an alternate option, the Proofreading interfaces have a tiled character
-picker section. Hovering over a character will show an enlarged view of it
-in the box on the left and a tooltip text will appear describing the
-character. Clicking on a character will insert it into the text.
-</p>
-
-<p>
-The top row shows the available character categories. The ones labeled
-A,E,I,O,U contain various accented versions, upper and lower case, of
-those respective characters. The one labeled "+" contains other special
-symbols, and accented versions of some consonants like Y, C, D, N etc.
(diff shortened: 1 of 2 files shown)


faq/prooffacehelp.php
```diff
@@ -670,18 +670,12 @@ function echo_row( $name, $tooltip, $button_image_base, $accelerator )
 
 <p>
 As an alternate option, the Proofreading interfaces have a tiled character
-picker section. Hovering over a character will show an enlarged view of it
-in the box on the left and a tooltip text will appear describing the
+picker section. Selecting one of the menus from the top row will show an
+array of characters. Hovering over a character will show an enlarged view
+of it in the box on the left and a tooltip text will appear describing the
 character. Clicking on a character will insert it into the text.
 </p>
```
srjfoo: Good changes. Should we also mention that hovering over the menu tiles will show a description of the content of each menu?

wf49670: Yes, good changes. But I agree that mentioning the tooltip for the tiles would be useful.

70ray: We could change the tooltips for the letter menu tiles to say something like `A variants` and so on.

srjfoo: Yes, also a good idea.

cpeel: Added tooltip mentions and added revised tooltips.

70ray: There is a typo. hoving for hovering.

cpeel: Good catch - thanks!

PR #172, opened 2019-12-16
Bad character warning area now scrolls correctly in all cases

70ray:
and message improved with highlighting.
The important part is changing the height of the flex container to 100% instead of 100vh which makes it work correctly in the enhanced interface.
Had to add an extra div with height 100vh in the standard interface for it to work there also.


cpeel:
I manually merged this in, because I once again futz with things upstream messing this merge up.

Thanks for the fix Ray!

PR #173, opened 2019-12-17
Adjust so fixedbox does not grow

70ray:


cpeel:
For my own edification, what problem is this solving Ray?

70ray:
This corrects a fault I unthinkingly introduced on the earlier change to correct the scrolling problem and puts the fixedbox css back how it was originally so that the box with the warning text and buttons only takes up as much space as it needs and does not grow bigger if there is only a small amount of text in the main box.

PR #175, opened 2019-12-18
Improve markup and ligature conversion

70ray:
Revised.


tools/proofers/codepoint_validator.js
```diff
@@ -34,15 +34,17 @@ $(function () {
         }
     });
 
-    var badPattern = new RegExp("[^" + charClass + "]", "ug");
+    var badClass = "[^" + charClass + "]";
+    var badPattern = new RegExp(badClass, "ug");
+    var badPattern0 = new RegExp(badClass, "u");
```
cpeel: I *really* dislike variables with numbers at the end. Can we rename these to what they are?
```
badPatternGlobal
badPattern
```

tools/proofers/codepoint_validator.js
```diff
@@ -149,11 +148,24 @@ $(function () {
...
+                    return match;
                 }
             }
             replaceChar = replaceChar.normalize("NFC");
-            maybeSubstitute();
+            return maybeReplace();
         }
-    });
+
+        text = text.replace(/\[(..)\]/gu, replacer);
```
cpeel: This makes me very wary. So for every 'input' event we will be firing a regex across all of the text. I'm wary of how this will perform on older browsers or slower machines.

Also, this will prevent proofers from inputting some PM-specified diacriticals. For instance, if the PM asked for a certain character be coded as `[:am]` and the user first inserts `[]` and then tries to fill in `:am`, the diacritical will be converted into the a+umlaut before they can insert the `m`. The way it is now there is a single predictable way that the conversion happens (upon `]`) which seems like a boon.

tools/proofers/codepoint_validator.js
```diff
@@ -149,11 +148,24 @@ $(function () {
...
+            caretIndex -= displacement;
+            textArea.setSelectionRange(caretIndex, caretIndex);
+        }
+    }
+
+    // event fires on every input into the textarea
+    $(textArea).on("input", replaceMarkup);
+
+    // also replace on initial load
+    replaceMarkup();
```
cpeel: I like this!

tools/proofers/codepoint_validator.js
70ray: Sorry about that.

tools/proofers/codepoint_validator.js
70ray: It would be interesting to find how much time this really takes. I'll look into it. Perhaps could underclock a machine. 
There was a lot of discussion about the brackets. Perhaps if it hadn't been mentioned that the [ ] key would not work for this purpose no-one would have thought about it. Perhaps we should discuss which option is preferable.

tools/proofers/codepoint_validator.js
70ray: Perhaps we should keep this part and have the character entry work only on closing bracket.

70ray: If it is useful to convert on load this could be done more simply. The extra complexity to put the caret in the right place would not be needed.

70ray:
We don't want to do this now.

PR #177, opened 2019-12-29
Drop redundant indexes

cpeel:
past_tallies and page_events are the two largest tables on pgdp.net because they are constantly written to. Both contain indexes that are no longer used but are taking up disk space. For instance, the page_events index size is 2x the size of the data itself.

Drop the indexes on these tables that aren't used in queries against them as confirmed by explain plans.

This is done with an eye to the InnoDB conversion and later the Unicode conversion (although page_events is already InnoDB on PROD).

### DIFF
--- a/SETUP/upgrade/13/20191228_alter_page_events.php
+++ b/SETUP/upgrade/13/20191228_alter_page_events.php
@@ -1,0 +1,24 @@
+<?php
+$relPath='../../../pinc/';
+include_once($relPath.'base.inc');
+
+header('Content-type: text/plain');
+
+// ------------------------------------------------------------
+
+echo "Dropping redundant keys on page_events...\n";
+$sql = "
+    ALTER TABLE page_events
+        DROP INDEX projectid,
+        DROP INDEX username_projectid_round_time;
+";
+
+echo "$sql\n";
+
+mysqli_query(DPDatabase::get_connection(), $sql) or die( mysqli_error(DPDatabase::get_connection()) );
+
+// ------------------------------------------------------------
+
+echo "\nDone!\n";
+
+// vim: sw=4 ts=4 expandtab


SETUP/upgrade/13/20191228_alter_page_events.php
```diff
@@ -0,0 +1,25 @@
...
+include_once($relPath.'base.inc');
+
+header('Content-type: text/plain');
+
+// ------------------------------------------------------------
+
+echo "Dropping redundant keys on page_events...\n";
+$sql = "
+    ALTER TABLE page_events
+        DROP INDEX projectid,
```
jmdyck: I'm pretty sure the 'projectid' key is being used. Look at `archive_ancillary_data_for_project`, which (via `move_project_rows_to_archive_db`) archives rows for a given project from `page_events`. `EXPLAIN` says that it uses the `projectid` index.

Moreover, when a proofer reports something odd relating to pages checked out/saved/returned/whatever, we typically do ad hoc queries like `select * from page_events where project='...' and ...`, which will probably use the `projectid` index.

Side note: the `projectid` index is an index on (`projectid`,`image`,`round_id`), but it probably doesn't need to be that granular. For a given project, there's probably only about a couple thousand page events, and if a query is only interested in a subset of them, I'm guessing MySQL can filter them pretty quickly even without a more granular index. So we could maybe save space by dropping `image` and `round_id` from the index.

SETUP/upgrade/13/20191228_alter_past_tallies.php
```diff
@@ -0,0 +1,23 @@
...
+include_once($relPath.'base.inc');
+
+header('Content-type: text/plain');
+
+// ------------------------------------------------------------
+
+echo "Dropping redundant key on past_tallies...\n";
+$sql = "
+    ALTER TABLE past_tallies
+        DROP INDEX tallyboard_time;
```
jmdyck: `explain select holder_id, tally_value from past_tallies where timestamp = 1577509200 and tally_name='P1' and holder_type = 'U';` says that it uses the `tallyboard_time` index.

SETUP/upgrade/13/20191228_alter_page_events.php
cpeel: I missed that (yay for code reviews!). MySQL seems happy to use either the `projectid` index or the `projectid_username` index for that query for whichever exists. Perhaps we keep these two and remove the other two?
```
  KEY `username` (`username`,`round_id`),
  KEY `projectid_username` (`projectid`,`username`),
```

SETUP/upgrade/13/20191228_alter_past_tallies.php
cpeel: Yup, thanks. I didn't see that one come through when I looked through the query logs. I'll drop the change on this table altogether.

SETUP/upgrade/13/20191228_alter_page_events.php
jmdyck: So removing:
```
KEY `projectid` (`projectid`,`image`,`round_id`),
KEY `username_projectid_round_time` (`username`,`projectid`,`round_id`,`timestamp`)
```
That seems like a good plan to me, but I've been wrong about indexes before. Is there a way we can test their removal on prod such that if it doesn't work, we can reinstate them quickly?

jmdyck: (I'm assuming that rebuilding them from scratch would not be quick.)

cpeel: Dropping the index should be very fast as a simple metadata operation. Adding it back would be slow, but the table would be available the entire time because that's an online DDL operation (see https://dev.mysql.com/doc/refman/5.7/en/innodb-online-ddl-operations.html). I don't know how we can test it on prod without doing the changes and I also don't know how we could do a backup such that we could quickly fall back.

cpeel:
@jmdyck - any further thoughts on this after my updates?

jmdyck:
No further thoughts.

PR #178, opened 2019-12-30
Use the $pi_tools_for_P and _F from stages.inc to define tools

70ray:
This should now work much the same as it did originally.
I originally changed it when working on an api version but it was probably not such a good idea there either.


pinc/ProofreadingToolbox.inc
```diff
@@ -4,31 +4,69 @@ include_once($relPath.'faq.inc');
 include_once($relPath.'stages.inc');
```
cpeel: Is this still needed? I suspect this was for `is_formatting_round()`, but I'm not 100% certain.

70ray: I've removed it. Still works.

PR #181, opened 2019-12-31
Extend SR deadline now done using extend_sr.php

70ray:
instead of upload_text.php. This includes validity checks
and both use subroutine in smooth_read_change.inc


pinc/smooth_read_change.inc
```diff
@@ -0,0 +1,57 @@
+<?php
```
cpeel: Please put this new function in `pinc/smoothread.inc` with the rest of the smoothreading functions.

tools/extend_sr.php
```diff
@@ -0,0 +1,58 @@
...
+
+require_login();
+
+$projectid = validate_projectID('project', @$_REQUEST['project']);
+$project = new Project($projectid);
+$days = get_integer_param($_REQUEST, 'days', 0, 0, 56);
+
+$title = _("Extend the Smooth Reading deadline");
+
+output_header($title);
```
cpeel: You might consider using `slim_header()` here with: if an error occurs do what you do now. If success use `metarefresh()` with 1 second to route the user back to the project page with just the success message showing briefly (it passes in as an argument to `metarefresh()`). That's similar to how other interstitial pages work and I think would be a little more consistent.

pinc/smooth_read_change.inc
70ray: Done this. A reason for keeping them separate was that the function is only needed in upload_text.php and the other functions in project.php. But of course this logic would lead to hundreds more small include files.

tools/extend_sr.php
70ray: Done this. It is consistent with the way upload_text.php works, but other upload cases such as handle_bad_page and remote file manager show a persistent confirmation.

cpeel: Thanks. I've created https://www.pgdp.net/c/tasks.php?action=show&task_id=1797 to track standardizing how we do this, because you're right - it's all across the board.

PR #182, opened 2020-01-02
Allow the user to specify a local font by name

cpeel:
This MR:
* Adds the license for DPCM2
* Adds DejaVu Sans Mono as a web font

But the biggest thing is that it changes the font selection in preferences so that the only fonts we offer by name are the two monospace web fonts (DejaVu Sans Mono and DPCM2). It also allows the user to type in a local font name to use. If the user was using one of the now-removed fonts, the font name is populated into this 'other' field during migration.

The upside is that encourages users to use the two monospace web fonts that we think provide a better ability to distinguish between characters while still enabling them to use a local font of their choice.


faq/font_sample.php
```diff
@@ -47,35 +50,62 @@
...
-echo "<p>" . _("This site provides DPCustomMono2 as a web font and browsers that support it do not require the font to be installed locally.") . "</p>";
+if($show_user_custom_font)
+{
+    echo "<h2 style='clear: both;'>" . _("Custom Proofreading Font") . "</h2>";
+    echo "<p>" . _("Your current proofreading font is one you've specified by name. This is what a specimen looks like in that font.") . "</p>";
+    show_font_specimen($proofreading_font, $proofreading_font);
+}
 
-echo "<p style='font-family: DPCustomMono2'>" . _("If your browser supports web fonts or you already have DPCustomMono2 installed, you will see this paragraph in that typeface. If this paragraph's font doesn't look radically different to that of the paragraph above, you can download DPCustomMono2 from <a href='DPCustomMono2.ttf'>here</a> (right click the link, and choose Save Target As..) After you have installed the font please refresh this page to make sure DPCustomMono2 is installed correctly.") . "</p>\n";
+echo "<h2 style='clear: both;'>" . _("DPCustomMono2") . "</h2>";
```
srjfoo: Not sure we should be localizing font names, so I thought I'd ask. (I realize that it's embedded in other strings that do need to be localized, though.)

cpeel: No, we shouldn't - good catch!

cpeel:
A reminder to myself that when this merges to update #1771 and related tickets.

PR #191, opened 2020-01-16
Pass back phpBB login failure reason to DP login

cpeel:
Pass back the underlying phpBB failure reason to the DP login page.
Route the user to the forums if their maximum number of login attempts
were exceeded so they can log in through the CAPTCHA.

This is available for testing in the [phpbb3-login-improvements](https://www.pgdp.org/~cpeel/c.branch/phpbb3-login-improvements/) sandbox.

### DIFF
--- a/accounts/login_failure.php
+++ b/accounts/login_failure.php
@@ -58,6 +58,13 @@ echo "<li>" . _("Caching set to off (or: refresh page every visit).") . "</li>\n
 echo "<li>" . _("Ensure your PC clock is set to the correct date &amp; time.") . "</li>\n";
 echo "</ol>";
 echo "<p>" . sprintf( _("If all of this fails, contact a <a href='%s'>site manager</a>."), "mailto:$site_manager_email_addr") . "</p>";
-echo "<p>" . sprintf( _("Note: If you have just registered, you will need to wait for the welcome mail to arrive to your mailbox. Once it does, please click the activation link to complete the registration (this is to prevent others from signing you up to the site without your knowledge). If you have waited for an hour or so and have still not received any mail from us (please check any spam filters!), it is likely that you misentered your email-address. Please contact a <a href='%s'>site manager</a> to solve the problem."), "mailto:$site_manager_email_addr") . "</p>";
+if($testing)
+{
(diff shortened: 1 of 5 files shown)


accounts/login_failure.php
```diff
@@ -26,6 +26,8 @@
     'no_username'      => _("You did not supply a username."),
     'no_password'      => _("You did not supply a password."),
     'auth_failure'     => _("Unable to authenticate. The username/password may be incorrect or your account may be locked."),
+    'unknown_failure'  => sprintf(_("An unexpected failure occurred, please contact the <a href='%s'>site manager</a>."), "mailto:$site_manager_email_addr"),
```
srjfoo: I'd probably say `a` site manager instead of `the` site manager

cpeel: I'm fine either way, this is using the same wording as the other paragraph at the bottom of the file.

srjfoo: Just grammatical pickiness -- `the` is specific, and implies that there's only one.

accounts/login_failure.php
```diff
@@ -55,7 +57,7 @@
...
+echo "<p>" . sprintf( _("If all of this fails, contact a <a href='%s'>site manager</a>."), "mailto:$site_manager_email_addr") . "</p>";
+echo "<p>" . sprintf( _("Note: If you have just registered, you will need to wait for the welcome mail to arrive to your mailbox. Once it does, please click the activation link to complete the registration (this is to prevent others from signing you up to the site without your knowledge). If you have waited for an hour or so and have still not received any mail from us (please check any spam filters!), it is likely that you misentered your email-address. Please contact a <a href='%s'>site manager</a> to solve the problem."), "mailto:$site_manager_email_addr") . "</p>";
```
srjfoo: This para probably ought to have a separate $testing option, also. It gives the false impression that users will actually receive an activation email. Possibly something along the lines of (further wordsmithing welcome):

> sprintf( _("Note: This is a testing site, and is not set up to send emails. If you have just registered, the page generated should have had an activation link on the page for you to copy and paste into a new browser tab. If you closed that page and can no longer access the information on it, please contact a <a href='%s'>site manager</a> to solve the problem."), "mailto:$site_manager_email_addr")

cpeel: Good call and good wording. Sandbox updated.

PR #193, opened 2020-01-19
Remote upload

70ray:
This unifies the uploads for upload_text and remote_file_manager using subroutines in upload_file.inc, the upload handler upload_file.php and the javascript file file_resume.js. The upload handler code has been revised since it is possible to defeat the lock file used before. This uses the flock() function. The documentation says that on some operating systems this cannot be relied on either. The other upload cases should also be considered but it seemed best to get some more opinions on these changes first. The zip checks are called at a high level so the lower level functions can be used for non-zip uploads.


tools/upload_text.php
```diff
@@ -167,80 +148,66 @@
...
 }
 
+$return_anchor = "<a href='$back_url'>$back_blurb</a>";
+$return_message = "<p>". sprintf(_("Return to the %s"), $return_anchor). "</p>";
+
 if (!isset($action))
 {
     // Present the upload page.
-
-    output_header($title);
```
cpeel: We should keep the themed output on this page.

tools/upload_file.php
```diff
@@ -0,0 +1,122 @@
+<?php
```
cpeel: If this is only for resumable files, calling it `upload_resumable_file.php` would be more accurate and informative.

pinc/upload_file.inc
```diff
@@ -0,0 +1,255 @@
...
+        {
+            $file_info = $_FILES['uploaded_file'];
+
+            // If a user hits the "Upload" button without first selecting a file,
+            // it appears that most browsers send a request containing a file whose
+            // name and content are empty. But I think it's also legal for a browser
+            // to send a request that doesn't contain a file at all (in which case
+            // $file_info would be null.  Check both possibilities.
+            if (is_null($file_info) || $file_info['name'] == '') {
+                return NULL;
```
cpeel: Is there a reason we shouldn't just throw an exception here anyway? That would streamline how this function is used -- it either returns a valid $file_info or it raises an exception.

cpeel: You could even raise and throw separate exception when no files was uploaded, that way you could catch and handle that expected case in `upload_text.php`.

pinc/upload_file.inc
```diff
@@ -0,0 +1,255 @@
...
+                You should perform a complete virus scan on your computer as soon as possible."), $av_test_result[0]));
+        }
+        else if($av_retval > 1)
+        {
+            throw new FileException(_("AntiVirus error:") . $av_retval . "<br>" . $av_test_result[0]);
+        }
+    }
+    catch(FileException $e)
+    {
+        unlink($file_path);
```
cpeel: It feels weird to have this function remove the thing that was uploaded. I would expect the calling function to do that.

pinc/upload_file.inc
```diff
@@ -0,0 +1,255 @@
...
+        // The extension was already checked and the file is not properly named (it has some temporary name), so we should
+        // disable the extension check.
+        if (!is_valid_zip_file($file_path, true))
+        {
+            throw new FileException(sprintf(_("%s is not a valid zip file"),  $original_name));
+        }
+    }
+    catch(FileException $e)
+    {
+        unlink($file_path);
```
cpeel: It feels weird to have this function remove the thing that was uploaded. I would expect the calling function to do that

pinc/upload_file.inc
```diff
@@ -0,0 +1,255 @@
...
+    return sprintf(_("If you are trying to upload a very big zip file
+        and the upload does not succeed, upload a small placeholder zip file
+        instead and email %s for assistance."), "<span class='nowrap'>$db_requests_email_addr</span>");
+}
+
+function get_upload_args()
+{
+    global $code_url;
+
+    $upload_messages = json_encode([
```
70ray: This should be utf8_encoded

tools/upload_text.php
70ray: Done this but I haven't put the statsbar back.

cpeel:
@70ray - after you've addressed my comments, can you please resolve (or comment & resolve) them so I know you've seen and looked at them?

70ray:
Thanks Casey, I'm having another look at the code.

pinc/upload_file.inc
```diff
@@ -0,0 +1,255 @@
+<?php
+
+define("RESUMABLE_UPLOAD_SIZE", 1024 * 1024 * 1024);  // 1GB
+
+class FileException extends Exception {}
```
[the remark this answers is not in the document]
70ray: Done this

pinc/upload_file.inc
```diff
@@ -0,0 +1,276 @@
...
+{
+    // create a valid filename re-using the extension provided
+    $path_parts = pathinfo($name);
+    // replace starting hyphen or anything except alphanumeric or - or _ by _
+    $new_name = preg_replace('/^-|[^a-zA-Z0-9_-]/', '_', $path_parts['filename']) . "." . $path_parts["extension"];
+    $message = sprintf(_('File was renamed to "%1$s" because "%2$s" is not a valid filename'), $new_name, $name);
+    return [$new_name, $message];
+}
+
+function is_valid_filename($filename, $restrict_extension=False)
```
cpeel: It feels like `is_valid_filename()` should use `make_valid_filename()` to ensure we don't have two places that define what a valid filename is. Right now the regexes are different -- there is no limit on the size of the filename in `make_valid_filename()` but there is in `is_valid_filename()`.

Basically
```
function is_valid_filename($filename, $restrict_extension=False)
{
    return $filename == make_valid_filename($filename, $restrict_extension);
}
```
With appropriate updates in `make_valid_filename()` to handle `$restrict_extension`.

tools/upload_text.php
```diff
@@ -249,164 +217,138 @@
...
-                    if(!$zip_ok)
+                    // extract any zips in smooth_dir
+                    $zips = glob("$smooth_dir/*.zip");
+                    foreach($zips as $zip)
                     {
-                        break;
+                        $zip_ok = extract_zip_to($zip, $smooth_dir);
+                        if(!$zip_ok)
+                        {
+                            break;
```
cpeel: Is there a reason we don't just raise the exception here instead of setting $zip_ok and raising later?

tools/upload_text.php
```diff
@@ -249,164 +217,138 @@
...
-        // replace filename
-        $zipext = ".zip";
-        $name = $projectid.$indicator.$zipext;
-        $location = "$project->dir/$name";
-        ensure_path_is_unused( $location );
-        copy($uploaded_file, $location);
-        unlink($uploaded_file);
-        $have_file = TRUE;
-        if ($stage == 'smooth_avail')
+        try
```
cpeel: Consider putting all of this inner-try logic into its own function. I think it's a fairly intact piece of code. That function thus either works or raises an exception. I think it only needs to be passed in `$returning_to_pool` and could return `$have_file`.

pinc/upload_file.inc
```diff
@@ -0,0 +1,276 @@
+<?php
+include_once($relPath.'misc.inc'); // get_upload_err_msg(), attr_safe(), return_bytes()
+
+// This contains functions for uploading files
```
cpeel: Can you please re-word-wrap this (and really all updated comment blocks) at 80 columns? I think it would be easier to read. It won't always be perfect but lets make use of the full 80-characters when we can.

pinc/upload_file.inc
70ray: Perhaps it would be better to remove both of these functions and have a javascript function to validate the filename before uploading (or whatever). Then the user would have the opportunity to rename the file himself.

cpeel: No, sorry, that won't work as a replacement. We can't replace server-side validation with client-side validation that can be bypassed. We could add a supplementary JS-side check, but that would require duplicating the validation logic -- which is not onerous but yet another thing to keep updated in two places.

70ray:
I've made some more changes.

70ray:
Oops... forgot to use translatable strings in the js file name validator.

70ray:
Translatable strings fixed now.

tools/project_manager/remote_file_manager.php
```diff
@@ -268,172 +252,33 @@ function do_showupload()
...
-}
-
-function handle_file_upload($file_info)
-{
-    global $curr_abspath, $hce_curr_displaypath, $antivirus_executable;
+    global $curr_abspath, $hce_curr_displaypath;
     global $pguser, $despecialed_username;
     global $commons_dir;
 
+    slim_header();
```
cpeel: This page should have some sort of title.

pinc/upload_file.inc
```diff
@@ -0,0 +1,276 @@
...
+class NoFileUploadedException extends Exception {}
+
+// This is called to process the posted information from the upload form
+// It returns a file_info with 'name' (the original file name)
+// and 'tmp_name' (the path to the file).
+// It checks for an empty file and for virus infection.
+// throws NoFileUploadedException if there is no file
+// throws a FileUploadException if there is a problem
+// if an exception is thrown any file will have been deleted.
+function validate_uploaded_file()
```
cpeel: One downside to this new implementation is that it doesn't give the user any feedback on upload status until everything is completed. In the current code, remote_file_manager pushes messages to the browser at each stage (file is uploaded, running AV check, etc). For the user's experience we should retain that more verbose output, at least for remote_file_manager.

Consider adding a `$verbose=FALSE` flag to this function. If `$verbose==TRUE`, output additional messages on the status, possibly with a helper function:
```
(in validate_uploaded_file)
        show_message($verbose, _("File successfully uploaded"));
        empty_check($file_size);
        show_message($verbose, _("Starting virus check"));
        virus_check($file_path);

function show_message($verbose, $message)
{
    if(!$verbose)
        return;
    echo "<p>$message</p>";
    flush();
}
```

To make this work, we need to disable the gzip encoding for the pages that use this. I would put that in `get_upload_args()`:
```
    // Disable gzip compression so we can flush the buffer after each step
    // in the process to give the user some progress details. Note that this
    // doesn't necessarily work for all browsers.
    apache_setenv('no-gzip', '1');
```

70ray: I think it gives as much feedback as is useful. The only two things that take a significant amount of time are the upload itself and the virus check. It gives upload progress percentage messages and says "Upload complete. Running file checks, please wait..." while running the virus check. Any other messages would get immediately superseded.

pinc/upload_file.inc
```diff
@@ -0,0 +1,276 @@
...
+        }
+        else
+        {
+            // resumable upload
+            $root_staging_dir = "/tmp/resumable_uploads";
+            $original_name = array_get($_POST, "resumable_filename", "");
+            $identifier = array_get($_POST, "resumable_identifier", "");
+            $hashed_filename = md5($identifier);
+            $file_path = "$root_staging_dir/$hashed_filename";
+            $file_size = filesize($file_path);
```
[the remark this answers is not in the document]
70ray: Done this.

pinc/upload_file.inc
cpeel: That's a good point, but it does presume that all uploads will be JS-based. That isn't an unreasonable expectation these days. It does also require that the browser leave the current page up while the virus scan is running. Chrome does, I'm not sure about the other browsers. This might also be impacted by the web server's gzip setting. It's on in TEST & PROD, but if it's off the server (which is the Apache default) posting to the next page may return a blank page while the virus scan runs.

All of that to say that there are a fair amount of unknowns around the user's experience in different scenarios.

tools/project_manager/remote_file_manager.php
70ray: I've given it the same title as the show_upload page now.

tools/upload_text.php
```diff
@@ -167,80 +149,66 @@
     exit;
 }
 
+$return_anchor = "<a href='$back_url'>$back_blurb</a>";
+$return_message = "<p>". sprintf(_("Return to the %s"), $return_anchor). "</p>";
```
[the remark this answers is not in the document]
70ray: Done this.

pinc/upload_file.inc
70ray: If it is uploading the old way it says "Working, please wait..." until everything is finished.
I don't know much about what is going on at lower levels but flushing would mean we were deliberately outputting invalid html in the sense that the end body and html tags were missing but browsers seem to handle it.

cpeel: > I don't know much about what is going on at lower levels but flushing would mean we were deliberately outputting invalid html

Less "invalid" and more "not valid yet", and this as always worked (for some definition of "always"). Think of a big 4MB HTML file - the browser will start rendering it to the user as soon as it has some content, refreshing as more content became available, well before the last body or html tag showed up.

70ray: Ok, I don't know what would be best then. Let's see what srjfoo's testing reveals.

srjfoo: I've been mulling this over, and I'm not sure what you mean @cpeel. I suspect that I just don't understand where it would be used where it might be trying to display a 4MB HTML file (or is that hypothetical and aimed at future-proofing?).

pinc/upload_file.inc
```diff
@@ -0,0 +1,276 @@
...
+
+    // if resumable, these are shown instead
+    echo "<div id='resumable_uploader' style='display: none;'>
+    <input type='hidden' name='resumable_filename' value=''>
+    <input type='hidden' name='resumable_identifier' value=''>
+    <p>", sprintf(_("Your browser supports uploading large files (up to %s) via javascript. If the upload fails,
+        or you navigate away from this page before it finishes, uploading the file again will pick up where it left off."), humanize_bytes(RESUMABLE_UPLOAD_SIZE)), "</p>
+    <p>", _("File to upload"), ": <span id='resumable_selected_file'></span> &nbsp;
+    <span id='resumable_browse' class='button'>", _("Choose File"), "</span>
+    <span id='resumable_submit' class='button'>", _("Upload"), "</span></p>
```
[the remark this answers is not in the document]
70ray: Good find. I'll fix it so the button has the same label as for non-resumable case.

tools/upload_text.php
```diff
@@ -134,7 +117,7 @@
 {
     $title = _("Upload smooth read version");
     $intro_blurb = _("This page allows you to upload a smooth read version of the project. One version per user per project is allowed: additional uploads by the same user will overwrite their previous upload.");
```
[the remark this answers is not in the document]
70ray: Good. Done this.

pinc/upload_file.inc
70ray: Casey was explaining to me why my objection to outputting incomplete html pages did not stand up. The original issue was considering whether it would be possible/useful to show any more messages describing the progress of the operation.

70ray:
> I tested a variety of combinations, both with resumables turned on and off so that I could see what a user with an browser that doesn't support the resumable uploads would see. The uploads succeeded when they were supposed to, and failed when they were supposed to.
> 
> The concerns that I mentioned in my comments have more to do with layout and wording than functionality.
> 
> Just to clarify -- this covers
> 
>     * Remote File Manager
> 
>     * PPer uploading to SR
> 
>     * PPer replacing SR file
> 
>     * PPer uploading to PPV
> 
>     * PPer uploading an interim file (or not) when returning a book to the PP pool
> 
>     * SRer uploading a SR report
> 
>     * PPVer uploading an interim file (or not) when returning a book to the PPV pool
>       Did I miss anything?

There is also the 'Upload corrected edition' case but I think it is not implemented on this site.

pinc/upload_file.inc
```diff
@@ -0,0 +1,279 @@
...
+}
+
+// This supplies the arguments for the header in the page containing the upload
+// form below
+function get_upload_args()
+{
+    global $code_url, $utf8_site;
+
+    $upload_messages = [
+        'working' => _("Working, please wait..."),
```
chrismiceli: I'm sure there may be many other instances of this, but I prefer to use the single '…' character to get better rendering with fonts. Just minor suggestion

tools/upload_text.php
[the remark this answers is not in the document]
70ray: Comments have got interleaved incorrectly but we could move the "return to wherever" link to the bottom of the page, where it is on the remote file manager.

scripts/file_resume.js
```diff
@@ -0,0 +1,100 @@
...
+        if(validate(filename)) {
+            // won't work if we disable browse button
+            $("#old_submit").prop( "disabled", true);
+            showProgress(uploadMessages.working);
+        } else {
+            ev.preventDefault();
+        }
+    });
+
+    $("#resumable_submit").click(function() {
```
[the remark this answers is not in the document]
70ray: It seems not. [(https://github.com/23/resumable.js/)] recommends using a span for the browse button but this is the submit button. I think cpeel wrote the original version of this.

pinc/upload_file.inc
70ray: Not sure I understand your meaning.

chrismiceli: instead of using 3 '.' characters, use the single '' unicode character https://www.fileformat.info/info/unicode/char/2026/index.htm to represent ellipsis

srjfoo: I think that we tend to use the three separate dots in general because of our proofing guidelines. In fact, that's one of the characters that will continue to be rejected even once we move to UTF-8. The problem is that in a monospace font, the three dots are crammed into the space a single character takes up, so in fact, you don't necessarily get better rendering across fonts.

pinc/upload_file.inc
srjfoo: Okay, thanks. I think I understand now. As I said, my testing seemed to show successful or unsuccessful uploads depending on whether or not I was knowingly uploading a too-large zip. I do appreciate the verbose messages in Remote File Manager.

pinc/upload_file.inc
cpeel: @chrismiceli - Purely from a technical perspective, right now the site is limited to outputting Latin-1 characters (we explicitly set the charset we report to the browser), so we couldn't use the `…` character without using the HTML entity for it. After the Unicode deployment we can start using UTF-8 characters in the code and this becomes possible.

scripts/file_resume.js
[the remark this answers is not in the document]
cpeel: I did, but I have no memory of why I made this decision vs another one.

pinc/upload_file.inc
srjfoo: > @chrismiceli - Purely from a technical perspective, right now the site is limited to outputting Latin-1 characters (we explicitly set the charset we report to the browser), so we couldn't use the `…` character without using the HTML entity for it. After the Unicode deployment we can start using UTF-8 characters in the code and this becomes possible.

Also true, though I think we will still need to be aware of where the output will be shown as monospace text and where a proportional font will be displayed.

scripts/file_resume.js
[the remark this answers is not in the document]
chrismiceli: It looks like they recommend span for styling reasons since they add children to the element. So if we stick with span, we should probably add an aria role as button and add keyboard handlers for enter.

pinc/upload_file.inc
chrismiceli: I agree entirely with @cpeel and @srjfoo. Not important right now, but something we can consider in the future depending on context.

scripts/file_resume.js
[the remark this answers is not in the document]
70ray: Perhaps we could create a new task for this. I've put in a note about why the span is used for the browse button.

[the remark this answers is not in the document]
chrismiceli: Sure, shouldn't hold this up I would imagine since the defect already exists.

pinc/upload_file.inc
70ray: Done something very like this. Had to change name to avoid conflict.

scripts/file_resume.js
```diff
@@ -48,9 +48,15 @@ $(function() {
 
     // Before we start the upload, prevent the user from hitting upload again.
     $("#old_submit").click(function(ev) {
-        let filename = $("#old_browse")[0].files[0].name;
-        if(validate(filename)) {
-            // won't work if we disable browse button
+        let file = $("#old_browse")[0].files[0];
+        if(file.size >= maxNormSize) {
```
cpeel: I don't understand this change. If the user has Javascript enabled they will very likely be using the resumable uploader. If they don't, this code will never run and this check will never happen. I don't think it's worth supporting the very small set of users who will have JS enabled but won't support resumable uploads.

tools/upload_text.php
```diff
@@ -221,6 +221,11 @@
     echo "<h1>$title</h1>";
     echo "<h2>", sprintf("Project: %s", $project->nameofwork), "</h2>";
 
+    // Disable gzip compression so we can flush the buffer after each step
+    // in the process to give the user some progress details. Note that this
+    // doesn't necessarily work for all browsers.
+    apache_setenv('no-gzip', '1');
```
cpeel: I think this needs to be set before we start outputting any text, otherwise Apache will have already enabled gzip for the buffer. That's why I suggested you set this in `get_upload_args()` which is almost guaranteed to be called before anything is output.

scripts/file_resume.js
70ray: This was to avoid the difficult situation that happens when a too-big file is uploaded and is handled by the function 'detect_too_large()' on line 23 of upload_file.inc. But as you say it is unlikely to happen. Sharon's testing came across it but only because I hacked the code to disable resumable in order to test the non-resumable case.

tools/upload_text.php
70ray: I've moved it up above the slim_header()
Does it only affect an individual visit to the server? If so then it would have to be in this section of the code where the header with `get_upload_args()` is not executed.
I've moved it in remote_file_manager also.

cpeel: Ah, good point. Moving it before `slim_header()` works fine.

cpeel:
This works perfectly for me, including outputting messages about it starting the virus check. I consider this good to go.

70ray:
There are two commits I haven't yet pushed up. I'll do it now.
Srjfoo had some concerns about where the return link on the fnal page of upload_text.php points to, expecially when something goes wrong, which have not been addressed yet.

srjfoo:
> There are two commits I haven't yet pushed up. I'll do it now.
> Srjfoo had some concerns about where the return link on the fnal page of upload_text.php points to, expecially when something goes wrong, which have not been addressed yet.

If uploading a file to dpscans via Remote File Manager fails because of a bad zip, the return link points back to the correct place (i.e. the folder within the user's dpscans that they were trying to upload to). Same thing if they forget to choose a file before clicking on the "Upload" button.

However, for example, if I attempt to upload a project for PPV, and neglect to choose a file (or choose a bad file), I'm sent back to the PP pool page, not to the project I was trying to upload for verification.

Was there also an issue with there being no return link at all if an upload failed for non-resumables?

tools/upload_text.php
```diff
@@ -249,164 +217,152 @@
...
-    if (is_file($uploaded_file))
+        $error_msg = project_transition( $projectid, $new_state, $pguser, $extras );
+        if ($error_msg)
+        {
+            throw new FileUploadException($error_msg);
+        }
+
+        // special handling for smooth reading, which does not involve a state change
+        // so project_transition() will do nothing
+        // but still needs some changes recorded in project table
```
srjfoo: Should `project table` be `project_events` table or `projects` table, or both? `smoothread_deadline` is stored in the `projects` table, but the assorted events recorded in the project history are recorded in `project_events`.

70ray: > However, for example, if I attempt to upload a project for PPV, and neglect to choose a file (or choose a bad file), I'm sent back to the PP pool page, not to the project I was trying to upload for verification.
For upload_text.php I've changed the links to 'Go Back' when the operation has not succeeded.

70ray: > Was there also an issue with there being no return link at all if an upload failed for non-resumables?

There was. Although it is not possible to do anything better after it has happened I've added a check for file size before the upload starts which should prevent it ever happening.

70ray: > Should project table be project_events table or projects table, or both? smoothread_deadline is stored in the projects table, but the assorted events displayed in the project history are recorded in project_events.

Yes, it should be both. I've changed the comment.

PR #194, opened 2020-01-22
Correct README link to intallation

chrismiceli:
Looking to contribute where necessary and noticed this while trying to familiarize myself with the codebase.


cpeel:
Awesome, thanks for catching and fixing it!

PR #195, opened 2020-01-22
Move the target-changing code into codepont_validator.js

70ray:
This addresses task DPUT 2250 where next image and text loads into right pane. This was because the target-changing code is executed but then wordcheck is not actually run. This task only executes the code if the validator succeeds and we actually do
wordcheck. Also move the h-v switch confirm so we do the
validation first.

### DIFF
--- a/tools/proofers/proof_frame_enh.inc
+++ b/tools/proofers/proof_frame_enh.inc
@@ -13,7 +13,8 @@ function echo_proof_frame_enh( $ppage )
     global $code_url, $userP;
 
     $json_code_points = get_json_codepoints($ppage->projectid());
-    $switch_confirm = javascript_safe(_('Are you sure you want to change layout & save the current document?'));
+    $switch_confirm = javascript_safe(_('Are you sure you want to save the current page and change layout?'));
+    $revert_confirm = javascript_safe(_("Are you sure you want to save the current page and revert to the original text for this round?"));
 
     $header_args = array(
         "css_files" => array("$code_url/styles/preview.css"),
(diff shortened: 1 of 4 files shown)


tools/proofers/proof_frame_enh.inc
```diff
@@ -13,6 +13,8 @@ function echo_proof_frame_enh( $ppage )
     global $code_url, $userP;
 
     $json_code_points = get_json_codepoints($ppage->projectid());
+    $switch_confirm = javascript_safe(_('Are you sure you want to change layout & save the current document?'));
```
cpeel: I hate that this confirmation message exists. But it does, so that's fine. Can we reorder the wording to put the important part the user might care about confirming first?

`Are you sure you want to save the current page and change layout?`

70ray: Done this.

70ray:
I suppose to be consistent we should validate before confirm for the revert to original text case also, and perhaps change the message to be "Are you sure you want to save the current page and revert to the original text for this round?

cpeel:
> I suppose to be consistent we should validate before confirm for the revert to original text case also, and perhaps change the message to be "Are you sure you want to save the current page and revert to the original text for this round?

If that's the way it works (save and then revert) then yes, I guess so.

/me mutters things about our confusing "enhanced" interface.

70ray:
I've made these changes. I don't like repeating the same code in two different places but hopefully in the next few years we'll be able to unify the two different interfaces so the repetition will go away.

cpeel:
> I've made these changes. I don't like repeating the same code in two different places but hopefully in the next few years we'll be able to unify the two different interfaces so the repetition will go away.

I agree completely.

PR #196, opened 2020-01-25
Update style and theme docs and initial config

cpeel:
This updates the theming documentation and adds the 3 built-in themes to the site upon initial creation.

### DIFF
--- a/styles/README.md
+++ b/styles/README.md
@@ -11,7 +11,7 @@ recent version of [node.js](https://nodejs.org). They have
 operating systems, including Ubuntu.
 
 After you have a recent version of node.js installed, use npm to install less:
-```
+```bash
 sudo npm install -g less
 ```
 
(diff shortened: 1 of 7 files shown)


styles/README.md
```diff
@@ -20,17 +20,27 @@ CSS is broken down into three major sections:
...
+Check in both the `.css` and `.less` files. While it isn't ideal to have generated
 code checked into source control, doing so means that consumers of the DP code
 don't also have to install and use less to generate the CSS.
+
+## Style design philosophy and demo
+The `design_philosophy.php` and `style_demo.php` pages discuss some of the
+thinking behind the site's design. You can view these pages rendered at pgdp.net
+at:
+
+* [Design Philosophy](https://www.pgdp.net/c/styles/design_philosophy.php)
```
srjfoo: I don't see any recommended changes for the Design Philosophy, but in looking at it, it was written before the current styling was rolled out. Some of the wording for that should probably be tweaked, as well. For example, under Design Goals/Consistency, the last sentence of that para is "This means that many pages will change such that all look alike.". We probably don't need that sentence any more.

cpeel: Good call. I added a commit to update the tenses on that page.

PR #197, opened 2020-01-27
Encode forum subjects when creating topics

cpeel:
phpBB encodes forum subjects before saving them to the database. We haven't been doing this since we moved over to phpBB3. Browsers appear to be forgiving the sin of not encoding the double quotes, but they draw the line at `>`s which breaks the rendering of project names with those in them.

### DIFF
--- a/SETUP/upgrade/13/20200127_fix_forum_topic_encoding.php
+++ b/SETUP/upgrade/13/20200127_fix_forum_topic_encoding.php
@@ -1,0 +1,143 @@
+<?php
+$relPath='../../../pinc/';
+include_once($relPath.'base.inc');
+
+# Force the DB connection into UTF8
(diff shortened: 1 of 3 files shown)


cpeel:
A downside to this is that any projects with `&` entities in the title or author is going to get double-encoded. This is less likely to happen when we move to UTF-8, but is a problem before then. This could be fixed by manually editing the topic after it is created.

One could argue that this would still be better than what we are doing now which is creating invalid HTML in the forum with these DP-created topics.

cpeel:
AH, I just need to set double_encode=False to fix that, nevermind.

cpeel:
I've added an upgrade script that will fix these unencoded strings in the forum database. Turns out there are a few places these subjects end up and I'm pretty sure I got them all.

PR #198, opened 2020-01-28
Pre-release doc & tooling updates

cpeel:
This is a collection of documentation and release-related tooling updates. After https://github.com/DistributedProofreaders/dproofreaders/pull/197 goes in, I believe this is the last thing to merge before we do a release.

### DIFF
--- a/SETUP/apache2.conf.example
+++ b/SETUP/apache2.conf.example
@@ -7,6 +7,19 @@
 # DP code uses 'default.php' as an alternative index file
 DirectoryIndex index.php default.php index.html default.html
 
+# To ensure that word lists and other project artifacts are downloaded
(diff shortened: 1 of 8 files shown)


SETUP/apache2.conf.example
```diff
@@ -7,6 +7,19 @@
 # DP code uses 'default.php' as an alternative index file
 DirectoryIndex index.php default.php index.html default.html
 
+# To ensure that word lists and other project artifacts are downloaded
+# with the correct character set, tell Apache to use ISO-8850-1 for
```
srjfoo: > \# with the correct character set, tell Apache to use ISO-8850-1 for

`ISO-8859-1`?

cpeel: Thanks - fixed!

PR #201, opened 2020-01-30
Handle combining diacritical marks

70ray:
This correctly highlights invalid characters with diacritical marks in the
validation view and removes them completely if required.
It also has code to enable use of valid glyphs with combining marks.


cpeel:
*Excellent* work Ray! Thank you for working on this!

PR #202, opened 2020-01-31
1449: Enhanced Image button

chrismiceli:
Create display_image url for image button to allow quick navigation

### DIFF
--- a/tools/proofers/PPage.inc
+++ b/tools/proofers/PPage.inc
@@ -87,10 +87,15 @@ class PPage
         return "$projects_url/{$this->lpage->projectid}/{$this->lpage->imagefile}";
     }
 
-    function url_for_display_image()
+    function url_for_display_image($escape_amp=FALSE)
     {
         global $code_url;
-        return "$code_url/tools/project_manager/displayimage.php?project={$this->lpage->projectid}&amp;imagefile={$this->lpage->imagefile}";
+        if($escape_amp) $amp = '&amp;';
+        else $amp = '&';
+        return "$code_url/tools/project_manager/displayimage.php"
+            . "?project={$this->lpage->projectid}"
+            . $amp
+            . "imagefile={$this->lpage->imagefile}";
     }
 
     function url_for_do_another_page($escape_amp=FALSE)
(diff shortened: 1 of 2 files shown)


chrismiceli:
Let me know if this change is not wanted in this way. All it does is open the image viewer instead of the actual image file. This allows peaking at the previous and next page quickly, which as the Task description mentions, makes reviewing some things easier. I need to request a sandbox to demonstrate the change though.

70ray:
Looks good to me. It should be really useful.

tools/proofers/PPage.inc
```diff
@@ -87,6 +87,12 @@ class PPage
         return "$projects_url/{$this->lpage->projectid}/{$this->lpage->imagefile}";
     }
 
+    function url_for_display_image()
```
cpeel: For consistency, this should take a `$escape_amp` argument and do the escaping (or not) like the other functions.

cpeel:
To be fair, what really needs to happen to those functions in PPage that have an `$escape_amp` arg is to have that arg removed, never return an escaped URL, and use `html_safe()` / `attr_safe()` on the returned value wherever that URL needs to be escaped in an HTML page. But that's outside the scope of this PR.

tools/proofers/PPage.inc
```diff
@@ -87,6 +87,17 @@ class PPage
         return "$projects_url/{$this->lpage->projectid}/{$this->lpage->imagefile}";
     }
 
+    function url_for_display_image($escape_amp=FALSE)
+    {
+        global $code_url;
+        if($escape_amp) $amp = '&amp;';
```
70ray: I would prefer the conditional statement on a new line. like
`if($escape_amp)
    $amp = '&amp;';`
or even
`if($escape_amp)
{
    $amp = '&amp;';
}`

chrismiceli: Sure I can make that change. I was just trying to be consistent with line 116. I see another open PR to autoformat PHP. That will help here :).

cpeel: Consistency with the rest of the file is probably the right move here. As-is getting rid of escaping in the function as well (but again, not in this PR).

PR #203, opened 2020-02-02
MARC search now returns empty string for genre

70ray:
and language if results are not in DP genre or langage lists. This addresses task 1817.

In the literary_form_array I've put a hyphen in Non Fiction but haven't changed anything else.


pinc/MARCRecord.inc
```diff
@@ -154,26 +155,33 @@ class MARCRecord
...
-        $marc_language = "";
         $key = $this->_key_search("008", "");
         if(isset($this->record[$key]))
         {
             $marc_language = substr($this->record[$key][1], 35, 3);
-            $marc_language = convert_short_lang($marc_language);
+            $lang_name = langname_for_langcode3($marc_language);
+            if($lang_name)
+            {
+                return $lang_name;
```
cpeel: Consider keeping `$marc_language` and only having one `return` in the function (and `$marc_literary_form` below). When possible, having one return makes the code flow easier to understand and avoids bugs in the future where the dev is unaware that there are multiple returns (clearly no the case here on these small functions, but a general good rule-of-thumb).

cpeel:
Also note that your build failed the CI/CD pipeline, looks like you need to update the MARC tests: https://travis-ci.org/DistributedProofreaders/dproofreaders/jobs/645019851

70ray:
There is something odd happening here with the language test expecting 'ng'. If I just make it expect an empty string it will just hide the problem. I'll try to fix it properly, (perhaps not immediately).

pinc/MARCRecord.inc
70ray: Done this.

70ray:
The language now returns English correctly. The literary form gives Fiction which is not one of our genres so it gives empty string.
I comented out the last test because I didn't know how to make the file independently of the method used to test it.

cpeel:
> I commented out the last test because I didn't know how to make the file independently of the method used to test it.

Using the method to create the string is valid, it will at least detect if changes are made to that function in the future.

Is there a reason you created/loaded a new MARC record for the test vs using the exiting one? Then you wouldn't need to recreate the existing string.

70ray:
The existing MARC record seemed to be faulty; there is a one character offset in the string that defines the language and genre. In the part  "1beng" 1 means fiction, b we don't use and eng is english, but we were getting b for biography which is plausible but not correct and ng instead of eng.

cpeel:
Weird - I know I just pulled it from whatever YAZ returned. Thanks for fixing it.

SETUP/tests/MARCRecordTest.php
```diff
@@ -1,4 +1,6 @@
 <?php
+$relPath='../../pinc/';
```
cpeel: You shouldn't need either of these lines. `phpunit_bootstrap.php` pulls in base.inc which installs an autoloader which will pull in MARCRecord on first use of the class. These tests should always be run with the bootstrap:
```
To run them, use the `phpunit_bootstrap.php` file to pull in pre-requisits.
    phpunit --bootstrap phpunit_bootstrap.php --verbose .
```

70ray:
Thanks I've removed them.

PR #204, opened 2020-02-02
Daily Page Limits

jmdyck:
I developed this code about 4 years ago (yikes), and it received some testing, but I abandoned it when other things came up. I'm creating a draft pull request so that it exists somewhere other than my home repo, and can maybe get some discussion.

A lot of the code is production-ready, I think (modulo intervening changes in house style). The main unresolved question is how the system should represent limits. The upgrade script (`SETUP/upgrade/09/01_daily_page_limits.php`) creates a `daily_page_limit` table, but none of the other new code refers to it, because for testing purposes, it was easier to create a representative DailyPageLimit object "by hand". (See `dpls_get_all` in `pinc/DailyPageLimit.inc`.)

### DIFF
--- a/SETUP/upgrade/09/01_daily_page_limits.php
+++ b/SETUP/upgrade/09/01_daily_page_limits.php
@@ -1,0 +1,39 @@
+<?php
+
+$relPath='../../../pinc/';
+include($relPath.'connect.inc');
+new dbConnect();
+
+header('Content-type: text/plain');
+
+
+echo "Creating 'daily_page_limits' table...\n";
+$sql = "
+    CREATE TABLE daily_page_limit
+    (
+        name             VARCHAR(50) NOT NULL,
+        project_selector VARCHAR(256) NOT NULL,
+        daily_limit      INT(8) NOT NULL,
+        PRIMARY KEY (name)
+    )
+";
+echo "$sql\n";
+mysqli_query(DPDatabase::get_connection(), $sql) or die( mysqli_error(DPDatabase::get_connection()) );
+
+
+echo "Creating 'daily_page_limit_user_counts' table...\n";
+$sql = "
+    CREATE TABLE daily_page_limit_user_counts
+    (
+        limit_name    VARCHAR(50) NOT NULL,
+        username      VARCHAR(25) NOT NULL,
+        current_count INT(8) NOT NULL,
+        PRIMARY KEY (limit_name, username)
+    )
+";
+echo "$sql\n";
+mysqli_query(DPDatabase::get_connection(), $sql) or die( mysqli_error(DPDatabase::get_connection()) );
+
+echo "\nDone!\n";
+
+// vim: sw=4 ts=4 expandtab
--- a/pinc/DailyPageLimit.inc
+++ b/pinc/DailyPageLimit.inc
@@ -1,0 +1,132 @@
+<?php
+include_once($relPath."Project.inc");
+include_once($relPath."misc.inc"); // functions that might be used in project_selector
+
+class DailyPageLimit
+{
+    function DailyPageLimit( $name, $project_selector, $daily_limit )
+    {
+        $this->name = $name;
+        $this->project_selector = $project_selector;
+        $this->daily_limit = $daily_limit;
+    }
+
+    function applies_to_project($project)
+    {
+        $str_to_eval = "return $this->project_selector;";
+        // echo "<br>\nevaluating: $str_to_eval<br>\n";
+        $result = eval($str_to_eval);
+        // echo "returns "; var_dump($result);
+        return $result;
+    }
+
+    function get_current_count_for_user($username)
+    {
+        $res = mysqli_query(DPDatabase::get_connection(), "
+            SELECT current_count
+            FROM daily_page_limit_user_counts
+            WHERE
+                limit_name = '$this->name'
+                AND
+                username = '$username'
+        ") or die(mysqli_error(DPDatabase::get_connection()));
+        if (mysqli_num_rows($res) == 0)
+        {
+            return 0;
+        }
+        else
+        {
+            list($count) = mysqli_fetch_row($res);
+            return $count;
+        }
+    }
+
+    function user_count_has_reached_limit($username)
+    {
+        $user_count = $this->get_current_count_for_user($username);
+        return ($user_count >= $this->daily_limit);
+    }
+
+    function adjust_user_count($username, $adjustment)
+    {
+        $res = mysqli_query(DPDatabase::get_connection(), "
+            INSERT INTO daily_page_limit_user_counts
+            SET
+                limit_name = '$this->name',
+                username = '$username',
+                current_count = $adjustment
+            ON DUPLICATE KEY UPDATE
+                current_count = current_count + $adjustment
+        ") or die(mysqli_error(DPDatabase::get_connection()));
+        // $n_rows = mysqli_affected_rows($res);
+        // $n_rows == 0 means an existing row was set to its current values
+        // $n_rows == 1 means a new row was inserted
+        // $n_rows == 2 means an existing row was updated
+    }
+}
+
+// We assume that there will be very few DPLs in the system,
+// and a project will rarely if ever be targeted by more than one DPL.
+
+function dpls_get_all()
+{
+    static $all_dpls = NULL;
+    if (is_null($all_dpls))
+    {
+        // XXX get from DB
+        $all_dpls = array(
+            new DailyPageLimit('F1', '$project->state == "F1.proj_avail" && !startswith($project->nameofwork, "F1 self-evaluation")', 3)
+        );
+    }
+    return $all_dpls;
+}
+
+function dpls_get_applicable($project)
+// Return an array of DailyPageLimit objects,
+// one for each page limit that currently applies to this project.
+{
+    $all_dpls = dpls_get_all();
+
+    $applicable_dpls = array();
+    foreach ($all_dpls as $dpl)
+    {
+        if ($dpl->applies_to_project($project))
+        {
+            $applicable_dpls[] = $dpl;
+        }
+    }
+    return $applicable_dpls;
+}
+
+function dpls_reset_user_counts()
+{
+    mysqli_query(DPDatabase::get_connection(), "
+        TRUNCATE TABLE daily_page_limit_user_counts
+    ") or die( mysqli_error(DPDatabase::get_connection()) );
+}
+
+function dpls_check_user_count($project, $username)
+// Of the DPLs (if any) applicable to $project,
+// return those (if any) for which the user has reached the limit.
+{
+    $dpls_reached = array();
+    foreach ( dpls_get_applicable($project) as $dpl )
+    {
+        if ($dpl->user_count_has_reached_limit($username))
+        {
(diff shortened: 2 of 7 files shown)


SETUP/upgrade/09/01_daily_page_limits.php
```diff
@@ -0,0 +1,39 @@
...
+        project_selector VARCHAR(256) NOT NULL,
+        daily_limit      INT(8) NOT NULL,
+        PRIMARY KEY (name)
+    )
+";
+echo "$sql\n";
+mysqli_query(DPDatabase::get_connection(), $sql) or die( mysqli_error(DPDatabase::get_connection()) );
+
+
+echo "Creating 'daily_page_limit_user_counts' table...\n";
```
cpeel: Why can't we use the TallyBoard for this instead of creating yet another table to track pretty much the same thing?

pinc/DailyPageLimit.inc
```diff
@@ -0,0 +1,132 @@
...
+// and a project will rarely if ever be targeted by more than one DPL.
+
+function dpls_get_all()
+{
+    static $all_dpls = NULL;
+    if (is_null($all_dpls))
+    {
+        // XXX get from DB
+        $all_dpls = array(
+            new DailyPageLimit('F1', '$project->state == "F1.proj_avail" && !startswith($project->nameofwork, "F1 self-evaluation")', 3)
```
cpeel: Rather than make this so extensible and requiring an eval(), why not make it narrower for the few cases that pgdp.net needs?

tools/proofers/processtext.php
```diff
@@ -306,6 +306,85 @@ function leave_spellcheck_mode( $ppage )
     }
 }
 
+function attempt_to_save_as_done($ppage, $text_data, $and_do_another)
+// This is only an attempt, because a daily page limit might block the save,
+// or prevent further saves.
+// If there's a problem, this function does not return to the caller.
+{
+    global $code_url, $pguser, $projectid, $proj_state;
```
cpeel: The projectid should be available in $ppage->lpage->projectid.

SETUP/upgrade/09/01_daily_page_limits.php
jmdyck: Because it's only "pretty much" the same thing, not the same thing. In particular, the F1 limit has to ignore pages saved in the F1 self-evaluation project.

pinc/DailyPageLimit.inc
jmdyck: If we don't represent (the details of) the daily page limits in the database, then we can certainly avoid the eval.

SETUP/upgrade/09/01_daily_page_limits.php
cpeel: But is it "close enough" to not matter? If we put this in people will then want to see how close they are to a given limit which means we have to put it in some page for them to view. It cascades quickly.

cpeel: Basically, can we go back to Linda et al and say "hey, the code is simpler if we implement a per-round per-person per-day limit". Because then we can use the TallyBoard and we don't need an `eval()` at all. Yes, it's less flexible but it's also less complex and will probably still address the core need.

jmdyck: There was also a request to allow for the possibility of having a higher daily limit for specific projects (e.g. projects that have been available longer, or special day projects). And I think there was also talk of using this mechanism to limit pages in beginner projects. So none of the anticipated use cases are as simple as "the number of pages saved in a given round".

cpeel: Of course not, nothing we do is simple.

Then this needs a way for a user to see what limits impact them and how close they are to those limits. It also needs a crontab or something to reset the contents of this table, presumably at midnight.

pinc/DailyPageLimit.inc
cpeel: We should consider putting the creation of `DailyPageLimit` objects for a site into a separate file. If the file exists we `include()` the file at the bottom of `pinc/DailyPageLimit.inc` which would then create the objects. This would allow us (and others) to have a file we can slide into place that would define these without patch hacks or putting them into the codebase.

I'm doing something similar for Glyphsets and it feels like a good approach.

jmdyck: Oh certainly, if DailyPageLimits were going to be created explicitly (rather than slurped in from a database table), then I'd define them in a separate file. (That just didn't seem like an option 4 years ago.) Putting the file of declarations outside the code-base seems fine too. Would we look for it in $dyn_dir?

cpeel: Well, we could do it either like we do for `$site_registration_protection_code` (having `site_vars.php` explicitly point to the file) or have the code assume some filename within `pinc/` and if it exists, include it.

SETUP/upgrade/09/01_daily_page_limits.php
jmdyck: >  this needs a way for a user to see what limits impact them

See the new code in project.php. Whenever they visit a project that contributes to a daily page limit, the "blurbs" will tell them so, and also tell them their current count toward that limit.

> It also needs a crontab or something to reset the contents of this table, presumably at midnight.

Yup. (For testing, we just "simulated midnight" whenever.)

cpeel: > See the new code in project.php. Whenever they visit a project that contributes to a daily page limit, the "blurbs" will tell them so, and also tell them their current count toward that limit.

That will be a poor user experience. A user will stop being able to proof a project and get redirected back to a project page where they are told they can't work on it anymore. Then they hunt around to try and find another project where they can work on, clicking through projects until they find one they can work on. If there are multiple limits in place this could be a very frustrating process.

If would be far better if the limits, and how closer the user was to them, was available somewhere for the user to look at -- such as on their stats page.

pinc/DailyPageLimit.inc
jmdyck: So if the file were in `pinc`, it'd be put there via `customize_deployment` (i.e., in our case, by copying from `c_hacks.newfiles`)?

SETUP/upgrade/09/01_daily_page_limits.php
jmdyck: > Then they hunt around to try and find another project where they can work on, clicking through projects until they find one they can work on.

I suspect there wouldn't be *that* much hunting involved. If you hit the limit for "X" projects, then you say to yourself "Okay, no more X projects for me today." So if you click on a project, thinking you'll be able to work on it, only to be told that you've already hit your limit for "X" projects, then either:
* you simply forgot (in which case, now you've been reminded), or
* you didn't realize that this project is an "X" project (contributes to the "X" daily limit). This is plausible, depending on how well the name of the limit conveys the set of projects that it covers (E.g., for "F1", it's pretty clear, but (say) "F1 older" and "F1 newer" wouldn't tell you exactly where the cutoff is.)

So that suggests a couple things:
* A DailyPageLimit should have a precise human-readable description of the project-selector that can be shown to the user when appropriate (e.g., as a tooltip or at the end of a link).
* In project listings, it might be helpful to mark (somehow) those projects that contribute to a limit that the user has already maxed out today.

> If would be far better if the limits, and how closer the user was to them, was available somewhere for the user to look at -- such as on their stats page.

(Do you mean "if instead" or "if in addition"?) Yup, something like that would be good. It would help people who've forgotten if/which limits they've reached today. But apart from that, I don't think it would improve the experience of 'hunting' for a project they can work on.

pinc/DailyPageLimit.inc
cpeel: Correct

SETUP/upgrade/09/01_daily_page_limits.php
cpeel: Yes - in addition to the project pages.

srjfoo: > > If would be far better if the limits, and how closer the user was to them, was available somewhere for the user to look at -- such as on their stats page.
> 
> (Do you mean "if instead" or "if in addition"?) Yup, something like that would be good. It would help people who've forgotten if/which limits they've reached today. But apart from that, I don't think it would improve the experience of 'hunting' for a project they can work on.

What about the "My Projects" page?

jmdyck: It sort-of relates, in that your counts-for-today would all be from projects on your "Recent" tab, but I don't think it really fits, because it's not a list of projects. Conceivably it *could* be a list of projects (i.e., the ones where your saves today contributed to your current count for limit X), but that wasn't requested. (It would require either different record-keeping or some slightly dubious queries.)

srjfoo: One of the problems with the user's Stats (or Member Details) page, assuming that's what is meant by stats page, is that it's not quick and easy to get to. The user has to go to Statistics Central and type their username into the search box (or is there an easier way that I've just missed for the last 10 years? :D). I've no problem with using that page, but it might be a good idea to make it more easily and obviously accessible, possibly with a direct link to "My Stats" at the top of Statistics Central?

cpeel: I think the daily limits contents belongs on the user's stats page. Yes, I've wanted to update how users access their own stats for a while, but I think that should be done outside of this PR.

srjfoo: That's reasonable -- I just thought that it was worth mentioning. :)

jmdyck: One other place that might be good for your-progress-toward-daily-limits is in the stats-bar on round pages. E.g., the F1 limit would make sense on the F1 page, under "Personal Statistics", say. (So each DailyPageLimit would have to know the round(s) for which it's relevant.)

jmdyck: > (or is there an easier way that I've just missed for the last 10 years? :D).

(If you're on a round page, you can go to the sidebar, under "Personal Statistics", and click the "<round-id> Details..." link. From there you can go to your other round-specific stats pages, but I don't think there's a link back to the round-agnostic stats page where your access chart is, though you can get there by stripping the round param from the URL.)

srjfoo: > > (or is there an easier way that I've just missed for the last 10 years? :D).
> 
> (If you're on a round page, you can go to the sidebar, under "Personal Statistics", and click the "[round-id] Details..." link. From there you can go to your other round-specific stats pages, but I don't think there's a link back to the round-agnostic stats page where your access chart is, though you can get there by stripping the `tally_name` parameter from the URL.)

Thanks. I've probably tripped over that occasionally in the past, but hadn't remembered it.

mannyack:
My two cents on the subject of hunting for projects that one can work on...it would be nice if projects for which one has gone beyond todays limit were grayed out on the F1 home page. Probably a follow on task, not part of initial implementation.

cpeel:
@jmdyck - can we close this request now that we have the smaller-scoped version in?

jmdyck:
(Odd, I didn't get a notification of this question.) Sure, I'll close.

PR #205, opened 2020-02-03
Add 'u' flag to regular expression.

70ray:
This addresses task 2255

### DIFF
--- a/tools/proofers/srchrep.php
+++ b/tools/proofers/srchrep.php
@@ -24,7 +24,7 @@ function do_replace()
     {
         search = preg_quote(search);
     }
-    opener.parent.docRef.editform.text_data.value=opener.parent.docRef.editform.text_data.value.replace(new RegExp(search,'g'),replacetext.value);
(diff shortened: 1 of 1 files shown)


tools/proofers/srchrep.php
```diff
@@ -24,7 +24,7 @@ function do_replace()
     {
         search = preg_quote(search);
```
chrismiceli: I'm new here and this isn't your change, but isn't this backwards? Don't we want to escape the regex if it is a regex and contains user entered data to avoid any issues creating the RegExp?

70ray: It took me a while to figure out what's going on here. The problem they're trying to solve is to 'replace all'. If you just use a string rather than a regex it will only replace the first occurrence so you have to use a regex but escape all the special characters. Stack Overflow has a much simpler way of doing it.

cpeel: The search/replace interface supports both regex searches and non-regex searches. If the user says this isn't a regex (`!is_regex`) we want to escape all of the characters so it's a straightforward search/replace. If it is a regex, we just want to pass it in unchanged. When being used as a regex we rely on the user to get their regex correct.

PR #206, opened 2020-02-07
Add warning for images larger than 100 kibibytes.

chrismiceli:
Example where I forced every warning for > 50 bytes (so every image). It shows off rows with multiple warnings as well.
image

### DIFF
--- a/pinc/image_check.inc
+++ b/pinc/image_check.inc
@@ -1,0 +1,15 @@
+<?php
+
+$page_image_size_limit = 100; // kb
+
+function get_image_size_error($imageSize)
+{
+    global $page_image_size_limit;
+    $result = NULL;
+
+    if($imageSize > ($page_image_size_limit * 1024))
+    {
+        $result = sprintf(_("Image > %dkb: %dkb"), $page_image_size_limit, $imageSize / 1024);
+    }
+    return $result;
+}
--- a/pinc/image_check.inc
+++ b/pinc/image_check.inc
@@ -2,14 +2,14 @@
 
 $page_image_size_limit = 100; // kb
 
-function get_image_size_error($imageSize)
+function get_image_size_error($image_size)
 {
     global $page_image_size_limit;
     $result = NULL;
 
-    if($imageSize > ($page_image_size_limit * 1024))
+    if($image_size > ($page_image_size_limit * 1024))
     {
-        $result = sprintf(_("Image > %dkb: %dkb"), $page_image_size_limit, $imageSize / 1024);
+        $result = sprintf(_("Image > %dkb: %dkb"), $page_image_size_limit, $image_size / 1024);
     }
     return $result;
 }
--- a/pinc/project_quick_check.inc
+++ b/pinc/project_quick_check.inc
@@ -5,6 +5,7 @@ include_once('links.inc'); // new_window_link()
 include_once($relPath.'../tools/project_manager/post_files.inc'); // page_info_query page_info_fetch
 include_once('bad_bytes.inc');
 include_once('wordcheck_engine.inc');
+include_once('image_check.inc');
 
 $test_functions  = array(
     // Automatic tests
@@ -41,19 +42,18 @@ function _test_project_for_large_page_images($projectid)
         $details .= "<th>" . _("Size") . "</th>";
         $details .= "</tr>";
 
-        $page_image_size_limit = 100; // kb
-
         $num_large_images = 0;
         while(list($image) = mysqli_fetch_row($result_res)) {
             // the the file's info
             $file_info = get_file_info_object($image,"$projects_dir/$projectid","$projects_url/$projectid");
 
-            // check to see if the file is > $page_image_size_limit kb
-            if($file_info->size > ($page_image_size_limit * 1024))
+            // check to see if the file is > size limit
+            $error = get_image_size_error($file_info->size);
+            if(isset($error))
             {
                 $details .= "<tr>";
                 $details .= "<td><a href='{$file_info->abs_url}'>$image</a></td>";
-                $details .= "<td>" . sprintf(_("Image > %dkb: %dkb"), $page_image_size_limit, $file_info->size / 1024) . "</td>";
+                $details .= "<td>" . $error . "</td>";
                 $details .= "</tr>";
     
                 $num_large_images++;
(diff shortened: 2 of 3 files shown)


tools/project_manager/add_files.php
```diff
@@ -407,7 +407,17 @@ function analyze()
...
+                $warning_msgs = array_merge($warning_msgs, $warnings);
+            }
+
+            if (isset($row['image']['src'][0]))
+            {
+                $image_filename = $base . $row['image']['src'][0];
+                if (filesize($image_filename) > 204800 /* 200 Kibibytes */)
+                {
+                    $this->n_warnings += 1;
+                    array_push($warning_msgs, pgettext("image too large", "Image may be larger than necessary."));
```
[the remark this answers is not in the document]
cpeel: https://github.com/DistributedProofreaders/dproofreaders/blob/master/pinc/project_quick_check.inc#L24 has the logic Sharon is referring to. Please re-use the file size and strings (string re-use when possible is a boon to our translators) in there.

tools/project_manager/add_files.php
```diff
@@ -383,7 +383,7 @@ function analyze()
             $row =& $this->page_file_table[$base];
 
             $error_msgs = '';
-            $warning_msgs = '';
+            $warning_msgs = array();
```
cpeel: I like this change. I would also like to have consistency in the file between `$error_msgs` and `$warning_msgs`. Could you add a second commit to this MR to change `$error_msgs` to use the same array-based syntax?

chrismiceli: Sure, sounds great.

tools/project_manager/add_files.php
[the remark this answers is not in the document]
chrismiceli: I moved the logic to a shared inc and call it from both locations now to share the max size and strings now.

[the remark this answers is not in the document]
chrismiceli: new strings with a value of 30kb for testing: 
image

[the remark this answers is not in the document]
chrismiceli: and in project_quick_check mainly to show I didn't break it:
image
image

chrismiceli:
> ?

personally I think that may be too restrictive, but if the community wants it. Do we know the average size of our image?

pinc/project_quick_check.inc
```diff
@@ -63,6 +63,7 @@ function _test_project_for_large_page_images($projectid)
         mysqli_free_result($result_res);
 
         $details .= "</table>";
+        global $page_image_size_limit;
```
cpeel: This isn't used anywhere in this file, leftover from a prior iteration?

pinc/image_check.inc
```diff
@@ -0,0 +1,15 @@
+<?php
+
+$page_image_size_limit = 100; // kb
+
+function get_image_size_error($imageSize)
```
cpeel: s/$imageSize/$image_size/ -- we generally prefer snake_case over camelCase

pinc/project_quick_check.inc
```diff
@@ -41,19 +42,18 @@ function _test_project_for_large_page_images($projectid)
...
         $num_large_images = 0;
         while(list($image) = mysqli_fetch_row($result_res)) {
             // the the file's info
             $file_info = get_file_info_object($image,"$projects_dir/$projectid","$projects_url/$projectid");
 
-            // check to see if the file is > $page_image_size_limit kb
-            if($file_info->size > ($page_image_size_limit * 1024))
+            // check to see if the file is > size limit
+            $error = get_image_size_error($file_info->size);
+            if(isset($error))
```
cpeel: You want `if($error)` here. If `$error` is NULL, this will still be true. PHP is weird.

tools/project_manager/add_files.php
```diff
@@ -407,7 +408,18 @@ function analyze()
...
                 $this->n_warnings += count($warnings);
-                $warning_msgs .= implode("\n", $warnings);
+                $warning_msgs = array_merge($warning_msgs, $warnings);
+            }
+
+            if (isset($row['image']['src'][0]))
+            {
+                $image_filename = $base . $row['image']['src'][0];
+                $warning_msg = get_image_size_error(filesize($image_filename));
+                if (isset($warning_msg))
```
cpeel: Ditto `if($warning_msg)`.

tools/project_manager/add_files.php
70ray: I haven't looked at this in detail and it's perhaps beyond the scope of this task but if you are keeping warnings and errors in arrays do you still need to keep a count of them in n_warnings etc.?

pinc/project_quick_check.inc
cpeel: And I am wrong!
> isset — Determine if a variable is declared and is different than NULL

From https://www.php.net/manual/en/function.isset.php

You can leave this as-is :)

pinc/project_quick_check.inc
chrismiceli: Used on line 71 and 77 for summary message.

chrismiceli:
@70ray, I was going to use your n_warnings  advice, but warning_msg is per page (txt/image pair) whereas n_warnings is for all pages. So I still think it makes sense to track separately.

tools/project_manager/add_files.php
chrismiceli: @70ray, I was going to use your n_warnings advice, but warning_msg is per page (txt/image pair) whereas n_warnings is for all pages. So I still think it makes sense to track separately.

pinc/project_quick_check.inc
chrismiceli: Yeah I didn't want to rely on a falsy check, so I looked up isset seems to be the best way to check for not null to me.

pinc/project_quick_check.inc
srjfoo: Not sure whether this needs to be differentiated or not, but I thought that I'd point out that images have two sizes -- the file size and the image dimensions. There are times when it's appropriate to show one or the other, and there are times when it's appropriate to show both, but "size" by itself may need context.

cpeel: Ah, that's right you moved the declaration. Please move this to the top of the functions - we keep all global variable declarations at the start of a function.

chrismiceli: Thanks for pointing that out.

PR #207, opened 2020-02-09
#1430: Firefox - retaining focus for multiple selections

chrismiceli:
select modified text to allow further operations without reselecting text.


chrismiceli:
We'll just want to make sure we don't regress https://www.pgdp.net/c/tasks.php?action=show&task_id=1747 that @cpeel fixed.

cpeel:
Great fix!

PR #209, opened 2020-02-10
Jquery min

chrismiceli:
Use min version of jquery to save ~180kb
downloaded from https://code.jquery.com/jquery-3.3.1.min.js


chrismiceli:
I'm not sure if this was discussed before, but this should save some bytes when visiting pages. I know it is most of the time cached anyway though. For reviewing, you can md5 the file and compare to https://code.jquery.com/jquery-3.3.1.min.js to ensure I'm not injecting anything malicious.

cpeel:
This is a good idea. I don't want to take it at this time only because I've already upgraded the version in the Unicode branch to 3.4.1 and used the minified version there and I don't think it's worth dealing with the merge conflict on that since it is most often cached. See also https://github.com/cpeel/dproofreaders/commit/68f62549db6715c82bcdab5a9970bfdd38d4843e

chrismiceli:
Thanks, that makes sense @cpeel. I don't want to make those merged any more difficult than I'm sure it's going to be. I checked the unicode branch though and still see 3.3.1. Did you not push upstream yet? https://github.com/DistributedProofreaders/dproofreaders/tree/unicode/pinc/3rdparty/jquery

cpeel:
I have not. I am waiting to land Ray's upload branch (code complete as of today), then rebase and force push. Should be done in a day or two when I find time.

PR #213, opened 2020-02-16
#1822: Change the default PM page view from 'Basic Page'

chrismiceli:
change default when creating users table
create script to alter existing table for upgrade


SETUP/upgrade/13/20200215_alter_pm_default.php
```diff
@@ -0,0 +1,19 @@
+<?php
```
cpeel: `SETUP/upgrade/13` was for the R202002 release that just went out, create a new `SETUP/upgrade/14` directory and put this in there.

PR #215, opened 2020-02-16
Add wordcheck status to diff page

mlazaric:
Adds an indicator to `/c/tools/project_manager/page_detail.php` of the WC status of each page. Useful for checking whether you ran wordcheck on each of your proofread pages.

 - Relevant task: https://www.pgdp.net/c/tasks.php?action=show&task_id=1174
 - Deployed at: https://www.pgdp.org/~mlazaric/c/

## Design

Currently it just prints `0` if it has not been wordchecked and `1` if it has:
image

### Checkmarks and xs 

Alternatively we could add a &check; (`&check;`) or &times; (`&times;`):
image
image

### Bold items which weren't wordchecked

We could bold the items which weren't wordchecked or the ones that were:
image

### New column in table

We could add a new column which contains the WC status:
image
image

**Note:** this would mean one additional WC column per round so it would bloat the table.

## Miscellaneous questions

 - Should we add the WC status to the recently proofed pages?

image

 - Should we treat it as a binary value (has been wordchecked / hasn't been wordchecked) or display the number of wordcheck events by the current user? 

 - Should we display it for formatting rounds?

### DIFF
--- a/project.php
+++ b/project.php
@@ -798,8 +798,13 @@ function recentlyproofed( $wlist )
     $recentNum=5; // if this is > 5 more rows will be shown
 
     $sql = "
-        SELECT image, state, {$round->time_column_name}
+        SELECT DISTINCT $projectid.image, $projectid.state, {$round->time_column_name}, (wordcheck_events.projectid IS NOT NULL) AS `wordcheck_status`
         FROM $projectid
+        LEFT OUTER JOIN wordcheck_events
+        ON (wordcheck_events.projectid = '$projectid' AND 
+            wordcheck_events.image = $projectid.image AND 
+            wordcheck_events.username = $round->user_column_name AND 
+            wordcheck_events.round_id = '$round->id')
         WHERE {$round->user_column_name}='$pguser' AND $state_condition
         ORDER BY {$round->time_column_name} DESC
         LIMIT $recentNum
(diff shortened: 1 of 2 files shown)


cpeel:
Some thoughts (not definitive decisions):
* I don't think users need to know the number of times they've WordChecked a page.
* I like having a simple check or X (but probably https://www.compart.com/en/unicode/U+2718 rather than the multiplication symbol).
* For simplicity, having it included in the formatting rounds makes sense to me.
* Having it on the recently proofread pages would be nice too. Part of me has wanted to rethink how that part of the table is laid out, but not enough to do anything about it. Having the check and X next to the name is probably sufficient for now.

chrismiceli:
I think a check or x is fine, and I agree we should use the proper symbol. I think a column is nice too, but then I would expect to sort by that column.

srjfoo:
> Some thoughts (not definitive decisions):
> 
>  * I don't think users need to know the number of times they've WordChecked a page.
> 
>  * I like having a simple check or X (but probably https://www.compart.com/en/unicode/U+2718 rather than the multiplication symbol).
> 
>  * For simplicity, having it included in the formatting rounds makes sense to me.
> 
I agree with all of the above. My main concern is screen real-estate, but dealing with that would be a different task altogether. Because of that, I have a slight preference for bolding the username where WC isn't used, but visually, the check mark and x (U+2718) are much more informative, and I like them, too. (We can't use the actual characters, though, until after we roll out UTF-8.)

>  * Having it on the recently proofread pages would be nice too. Part of me has wanted to rethink how that part of the table is laid out, but not enough to do anything about it. Having the check and X next to the name is probably sufficient for now.
>
I agree -- rethinking the layout of "recently proofread pages" section of the project page sounds like a good idea, but not part of this task.

Looking back to when I was a newbie, even with the link to the Page Details right above that line, I never noticed it initially, so I wouldn't have gotten any WC feedback. Is this a non-starter?

srjfoo:
> I think a check or x is fine, and I agree we should use the proper symbol. I think a column is nice too, but then I would expect to sort by that column.

The Page Detail display isn't sortable, so that's not particularly an issue. It takes up an awful lot of extra horizontal space, too, which isn't much of an issue for P1 and P2, but when you get much beyond that, it can become so.

70ray:
I like the tick or cross. Perhaps there should be some text on the page explaining what they mean.

mlazaric:
Switched to &check; and &#10008; and added a hover message explaining the icons. 

image

image

image

mlazaric:
Added wordcheck status to recently proofed pages:

image
image

cpeel:
This looks great to me. Thoughts @srjfoo?

srjfoo:
> This looks great to me. Thoughts @srjfoo?

I agree.

cpeel:
I intend to merge this in tomorrow. The feedback in the forums has been very favorable.

PR #217, opened 2020-02-18
Allow \n and treat as newline character

chrismiceli:
image

### DIFF
--- a/tools/proofers/srchrep.php
+++ b/tools/proofers/srchrep.php
@@ -17,7 +17,7 @@ function do_replace()
     var search_textbox = document.getElementById('search');
     var search = search_textbox.value;
     var regex_checkbox = document.getElementById('is_regex');
-    var replacetext = document.getElementById('replace').value.replace(new RegExp('\\\\n', 'g'), '\n');
+    var replacetext = document.getElementById('replace').value.replace(new RegExp('\\\\n', 'g'), '\r\n');
     save_text();
     var is_regex = regex_checkbox.checked;
     if (!is_regex)
@@ -131,7 +131,6 @@ window.onload = test_and_enable_nonregex;
     <input type="button" id='undo' value="<?php echo attr_safe(_("Undo.")); ?>" onClick="restore_saved_text()" disabled>
 </center>
 </form>
-<p><?php echo _("Note: \\n will be treated as a newline in the replace field."); ?></p>
 <p><?php echo _("Warning: Undo is only possible for the most recent replace!"); ?></p>
 <p id='regex_help_title' onclick='toggle_regex_help();'><span id='regex_arrow'>&#9654;</span>
 <?php echo _('Regular Expression?'); ?></p>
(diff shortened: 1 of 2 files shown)


chrismiceli:
I added a small note to increase the discoverability of this feature. I can remove if desired.

tools/proofers/srchrep.php
```diff
@@ -131,6 +131,7 @@ function preg_quote(str)
     <input type="button" id='undo' value="<?php echo attr_safe(_("Undo.")); ?>" onClick="restore_saved_text()" disabled>
 </center>
 </form>
+<p><?php echo _("Note: \\n will be treated as a newline in the replace field."); ?></p>
```
srjfoo: Am I correct that \n is available only for regex s/r? If so, this note might be better off in the "Regular Expression?" help area. Otherwise, I have a slight preference for removing the note.

chrismiceli: No, checking the regular expression checkbox is not required. That checkbox controls what text will get selected to be replaced, not what it will be replaced with. This change would disallow replacing with the text "\n" and always interpreting that as a newline character.

Thanks for your opinion on the note.

chrismiceli:
I just realized I should replace with \r\n so please don't merge until I fix that.

tools/proofers/srchrep.php
srjfoo: > No, checking the regular expression checkbox is not required. That checkbox controls what text will get selected to be replaced, not what it will be replaced with. This change would disallow replacing with the text "\n" and always interpreting that as a newline character.
> 
Thanks for the clarification -- I was forgetting that (I'm used to `\n` being used only in regexes, both for search and replace).
> Thanks for your opinion on the note.
It's not that I don't like things to be documented -- I am just very leery about adding text to the s/r window. I'd rather keep the actual s/r form simple and slim. I notice, though, that search/replace is not documented in the prooffacehelp for either the standard or enhanced version -- I think this would be a good place to document it. We could also consider adding something to the "Proofing Interface tips and tricks" DP wiki page.

srjfoo: > ... search/replace is not documented in the prooffacehelp for either the standard or enhanced version -- I think this would be a good place to document it.

There appears to be a request to add documentation for search/replace as [task 1740](https://www.pgdp.net/c/tasks.php?action=show&task_id=1740)

cpeel: I would prefer we punt documenting this into task 1740 and just do the replacement. The odds of someone wanting to do a replace with a '\n' and not mean a newline is almost nil.

Let's remove the hint para and add comment in task 1740 on what needs to be included for it in the search/replace doc update.

srjfoo: Works for me.

chrismiceli: I updated the code to remove the note and I can take a look at 1740.

chrismiceli:
> 
> 
> I just realized I should replace with \r\n so please don't merge until I fix that.

This is resolved.

PR #218, opened 2020-02-20
Rename glyphsets to character suites

cpeel:
@jmdyck asserts that 'glyphset' is inaccurate and could be confusing. This is a proposal to change them to "character suites" or charsuites for short. I'm interested in other developers' thoughts on this change.

I'm on the fence. We've already established the term Glyphset with the community and it is referenced in several places in the wiki and forums. I also dislike this monolithic change-the-world commit but the effort to rewrite history to be consistent is *way* more work than I am interested in doing. I also don't find the term "glyph" to be inaccurate or confusing, but am willing to entertain that it might be for others.


70ray:
Perhaps just change 'glyph' to 'character' and so on in the translatable strings.

lhamilton1:
I approve

lhamilton1:
I suspect we'll be better off switching to character suite since it is more accurate. Glyph sounds lovely and sort of mysterious and magical but over time, being accurate will probably be best.

PR #219, opened 2020-02-20
Bad page check

70ray:
Addresses task 2247. It validates the characters in the replace text textarea before submitting and shows bad characters in the same way as in the proofing interface.
Some javascript and other files have been re-arranged, renamed and moved in order to avoid having paths to them in a different subdirectory.


70ray:
Now running in https://www.pgdp.org/~rp31/preview sandbox.

tools/proofers/text_frame_std.inc
```diff
@@ -3,7 +3,7 @@ include_once($relPath.'button_defs.inc');
 include_once($relPath.'slim_header.inc');
 include_once($relPath.'misc.inc');
 include_once('preview.inc');
-include_once('codepoint_validator.inc');
+include_once($relPath.'codepoint_validator.inc');
```
cpeel: Move this up two lines before the `preview.inc` with the others that pull from `pinc/`

tools/proofers/proof_frame_enh.inc
```diff
@@ -6,7 +6,7 @@ include_once('PPage.inc');
 include_once('image_block_enh.inc');
 include_once('button_menu.inc');
 include_once('preview.inc');
-include_once('codepoint_validator.inc');
+include_once($relPath.'codepoint_validator.inc');
```
cpeel: Move this up before `PPage.inc` so it's with the other files that are in `pinc/`.

tools/project_manager/update_validate.js
```diff
@@ -0,0 +1,9 @@
+/*global $ validateText */
```
cpeel: Please rename this to something clearer. I would not be able to figure out what this file is and how it is suppose to be used based on the filename and contents alone.

You might just call it `handle_bad_page.js` since it is so closely tied to this page.

tools/proofers/proofing.js
```diff
@@ -1,70 +1,8 @@
-/*global $ goodChars charMatch standardInterface switchConfirm revertConfirm XRegExp */
```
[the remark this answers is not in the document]
70ray: I think there is a good case for splitting this file into two which do different things:
`proof_validate.js` and `process_diacritical_marks.js`.

tools/proofers/text_frame_std.inc
70ray: Done this.

tools/proofers/proof_frame_enh.inc
70ray: Done this.

tools/project_manager/update_validate.js
70ray: Done this.

tools/proofers/proofing.js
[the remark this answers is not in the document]
70ray: Done this split now.

pinc/wordcheck_engine.inc
```diff
@@ -8,7 +8,7 @@ include_once($relPath.'unicode.inc');
 
 // Handle dp-specific notation for non-Latin-1 characters
 // See faq/proofreading_guidelines.php#d_chars
-// and tools/proofers/codepoint_validator.js
+// and tools/proofers/proofing.js
```
cpeel: `tools/proofers/process_diacritcal_markup.js`?

70ray: Done thus

PR #220, opened 2020-02-21
Check combination

70ray:
This addresses task 2260.
For adding files to a  project, if a file contains any extended characters (with combining diacritical marks) the whole character is shown in the list of bad characters and a title tooltip added to indicate what it is.
Similarly for Project Quick Check the bad character table shows the whole character and the hex codepoints from which it is composed.
This is running in the https://www.pgdp.org/~rp31/pm sandbox.
I made a bad project to test it projectID5c13a64918b30.
Perhaps we should make a project with a more comprehensive set of faults to give a better test if such a thing does not already exist.
bad_bytes.inc contains some reference to combining marks. I haven't investigated whether this will do anything in any circumstances.

### DIFF
--- a/pinc/DPage.inc
+++ b/pinc/DPage.inc
@@ -716,11 +716,13 @@ function get_normalize_page_text_changes($text, $projectid)
 
     if($invalid_chars)
     {
-        // this will contain the characters and decomposed codepoint arrays.
+        // for each element in this array the key will be the character and
+        // the value will be an array of utf8 strings for each component
+        // if not printable the key is U+xxxx codepoint instead.
         $processed_invalid_chars = [];
         foreach(array_keys($invalid_chars) as $char)
         {
-            // find the components if it is an extended character
+            // find the components; maybe an extended character
             preg_match_all("/./u", $char, $matches);
             $char0 = $matches[0][0];
             // if not defined or is a control character or soft hyphen
--- a/pinc/DPage.inc
+++ b/pinc/DPage.inc
@@ -716,23 +716,33 @@ function get_normalize_page_text_changes($text, $projectid)
 
     if($invalid_chars)
     {
-        $invalid_char_names = [];
+        // this will contain the characters and decomposed codepoint arrays.
+        $processed_invalid_chars = [];
         foreach(array_keys($invalid_chars) as $char)
         {
-            $name = IntlChar::charName($char);
-            # if it doesn't have a name it's a control character which can't
-            # be printed anyway, or if it's a soft hyphen
-            if(!$name || utf8_chr_to_hex($char) == 'U+00ad')
+            // find the components if it is an extended character
+            preg_match_all("/./u", $char, $matches);
+            $char0 = $matches[0][0];
+            // if not defined or is a control character or soft hyphen
+            if(!IntlChar::isdefined($char0) || IntlChar::iscntrl($char0) || ($char0 === "\x{ad}"))
             {
-                $name = utf8_chr_to_hex($char);
+                // put U+xxxx instead of utf8 string
+                $processed_invalid_chars[utf8_chr_to_hex($char0)] = $matches[0];
             }
             else
             {
-                $name = $char;
+                // extended characters will get here unless first component is caught above
+                $processed_invalid_chars[$char] = $matches[0];
             }
-            $invalid_char_names[] = $name;
         }
-        $invalid_char_string = implode(", ", $invalid_char_names);
+        // put the characters in spans with a title
+        $char_spans = [];
+        foreach($processed_invalid_chars as $char => $codepoints)
+        {
+            $title = make_title($codepoints);
+            $char_spans[] = "<span class='warning' title='$title'>$char</span>";
+        }
+        $invalid_char_string = implode(", ", $char_spans);
         $changes[] = sprintf(_("Invalid characters will be removed: <span class='mono'>%s</span>"), $invalid_char_string);
     }
 
--- a/pinc/unicode.inc
+++ b/pinc/unicode.inc
@@ -61,6 +61,26 @@ function utf8_filter_out_codepoints($string, $remove_codepoints, $replacement=""
     return preg_replace($pattern_string, $replacement, $string);
 }
 
+// make a string for use in a regular expression to match any of the characters
+// including the ranges
+function make_regex_string($characters)
+{
+    $char_class = "";
+    foreach($characters as $character)
+    {
+        if(strpos($character, '-') === False)
+        {
+            $char_class .= utf8_chr($character);
+        }
+        else
+        {
+            list($start, $end) = explode('-', $character);
+            $char_class .= utf8_chr($start) . '-' . utf8_chr($end);
+        }
+    }
+    return "[$char_class]";
+}
+
 # Take an array of Unicode codepoints (U+####), or codepoint ranges
 # (U+####-U+####) and return an array of unicode characters.
 function convert_codepoint_ranges_to_characters($codepoints)
--- a/pinc/unicode.inc
+++ b/pinc/unicode.inc
@@ -25,7 +25,6 @@ function utf8_filter_to_codepoints($string, $valid_codepoints, $replacement="")
         function ($matches) use($pattern_string)
         {
             $match = $matches[0];
-//            var_dump($match);
             if(1 === preg_match($pattern_string, $match))
             {
                 return $match;
@@ -49,7 +48,6 @@ function utf8_filter_out_codepoints($string, $remove_codepoints, $replacement=""
         function ($matches) use($pattern_string)
         {
             $match = $matches[0];
-//            var_dump($match);
             if(1 === preg_match($pattern_string, $match))
             {
                 return "";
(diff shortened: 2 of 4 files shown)


cpeel:
From the bug:
> If I add a page to a project containing a valid base character with a combining diacritical mark that does not normalise to a single unicode codepoint the diacritical mark is removed and reported as bad but the base character remains. Both should removed and reported as a single unit.

I disagree. I think PMs will want the combining character removed and the character to remain.

pinc/DPage.inc
```diff
@@ -716,23 +716,35 @@ function get_normalize_page_text_changes($text, $projectid)
...
             }
-            $invalid_char_names[] = $name;
         }
-        $invalid_char_string = implode(", ", $invalid_char_names);
+        // put the characters in spans with a title
+        $char_spans = [];
+        foreach($processed_invalid_chars as $char => $codepoints)
+        {
+            $title = make_title($codepoints);
+            $char_spans[] = "<span class='warning' title='$title'>$char</span>";
```
cpeel: I don't think these need to be a warning class, they are already in the Warnings column.

pinc/DPage.inc
```diff
@@ -716,23 +716,35 @@ function get_normalize_page_text_changes($text, $projectid)
...
+        $processed_invalid_chars = [];
         foreach(array_keys($invalid_chars) as $char)
         {
-            $name = IntlChar::charName($char);
-            # if it doesn't have a name it's a control character which can't
-            # be printed anyway, or if it's a soft hyphen
-            if(!$name || utf8_chr_to_hex($char) == 'U+00ad')
+            // find the components; maybe an extended character
+            preg_match_all("/./u", $char, $matches);
+            $char0 = $matches[0][0];
```
cpeel: Please stop appending numbers to the end of variables to make new variables. `$char0` tells me literally nothing about what it is and how it differs from `$char`.

pinc/DPage.inc
cpeel: I do like that these include the title now.

pinc/unicode.inc
```diff
@@ -170,3 +193,43 @@ function guess_string_encoding($text)
     # Give up and return ISO-8859-1
     return 'ISO-8859-1';
 }
+
+function make_title($codepoints)
```
cpeel: `get_character_titles()` would be a more descriptive title.

pinc/unicode.inc
```diff
@@ -170,3 +193,43 @@ function guess_string_encoding($text)
...
+{
+    // make title from array of base codepoint and possible combining marks
+    $title = IntlChar::charName($codepoints[0]);
+    $count = count($codepoints);
+    if($count > 1)
+    {
+        $title .= " with " . IntlChar::charName($codepoints[1]);
```
cpeel: Creating the titles like this won't work for non-English speakers. It might simply be easier translation-wise if we removed the with/and altogether and just a `+` symbol between all of the titles instead.

pinc/unicode.inc
```diff
@@ -170,3 +193,43 @@ function guess_string_encoding($text)
...
+    }
+    for($index = 2; $index < $count; $index++)
+    {
+        $title .= " and " . IntlChar::charName($codepoints[$index]);
+    }
+    return $title;
+}
+
+function extended_to_hex($char)
```
cpeel: Is "extended" a wording in Unicode to represented combined characters? I feel like "combined" is a better description and how we've talked about it so far.

`combined_character_to_hex()` perhaps.

pinc/unicode.inc
```diff
@@ -170,3 +193,43 @@ function guess_string_encoding($text)
...
+    $codepoints = [];
+    preg_match_all("/./u", $char, $matches);
+    foreach($matches[0] as $match)
+    {
+        $codepoints[] = str_pad(dechex(IntlChar::ord($match)), 4, "0", STR_PAD_LEFT);
+    }
+    return implode(" ", $codepoints);
+}
+
+function extended_strlen($string)
```
cpeel: I can't work out what this does -- does it return the length of the string treating all combined characters as separate? Isn't that just the number of codepoints? If it does something different we need to rename it.

pinc/unicode.inc
```diff
@@ -170,3 +193,43 @@ function guess_string_encoding($text)
...
+{
+    // how many 'extended Unicode sequences'?
+    return preg_match_all("/\X/u", $string);
+}
+
+function is_extended($string)
```
cpeel: So this returns a true or false if the string strictly contains only combined characters?

pinc/unicode.inc
```diff
@@ -18,47 +18,64 @@ function utf8_normalize($string)
...
+            {
+                return $match;
+            }
+            return "";
+        },
+        $string);
 }
 
 # Filter out any characters in $string that are in $remove_codepoints
 function utf8_filter_out_codepoints($string, $remove_codepoints, $replacement="")
```
cpeel: I don't understand why this function needs to change. The purpose of this function is to remove specific codepoints from a string. If, instead, we want to remove specific characters, which might include combined characters, that would be a different function with a different name.

pinc/unicode.inc
```diff
@@ -18,47 +18,64 @@ function utf8_normalize($string)
...
+                return "";
+            }
+            return $match;
+        },
+        $string);
+}
+
+// make a string for use in a regular expression to match any of the characters
+// including the ranges
+function make_regex_string($characters)
```
cpeel: When naming functions, keep in mind that these are in the global namespace, not as part of a class, and in that light this function name isn't very helpful or clear on what it does.

`create_regex_from_characters()`, perhaps.

srjfoo:
> From the bug:
> 
> > If I add a page to a project containing a valid base character with a combining diacritical mark that does not normalise to a single unicode codepoint the diacritical mark is removed and reported as bad but the base character remains. Both should removed and reported as a single unit.
> 
> I disagree. I think PMs will want the combining character removed and the character to remain.

Of those two options, I don't know which I'd prefer. Right now, we delete unwanted characters, so a case could be made for either one. Either way, you wind up with something that is not what was intended. As long as the project load or page save is able to identify what's bad and make sure that the PM/proofer knows about the problem, and what will happen, that's what's important.

srjfoo:
> I'm also concerned that this is a spot-fix for a specific issue (that I'm not too worried about) and doesn't fully address how we want to handle combining characters in general. We should figure out how we want to address these and if we want to address them before July.

I admit that I haven't really studied combining characters enough to understand them. I wouldn't think we would need anything in place to handle them before July, except as individual characters that are not being combined by the DP software. (Or am I missing something, here?)

70ray:
> From the bug:
> 
> > If I add a page to a project containing a valid base character with a combining diacritical mark that does not normalise to a single unicode codepoint the diacritical mark is removed and reported as bad but the base character remains. Both should removed and reported as a single unit.
> 
> I disagree. I think PMs will want the combining character removed and the character to remain.
I think it is important for the PM to know to which character the diacritical mark was attached; this code enables him/her to see the whole character.

pinc/DPage.inc
70ray: True, they don't need to be in the warning class.

pinc/DPage.inc
70ray: Sorry, I had forgotten this was one of your pet hates, I'll think of something better.

70ray:
> I didn't comment on everything, but I have a problem how we're suddenly co-opting functions that operate on codepoints to now operate on characters.
> 
> I'm also concerned that this is a spot-fix for a specific issue (that I'm not too worried about) and doesn't fully address how we want to handle combining characters in general. We should figure out how we want to address these and if we want to address them before July.

True, it's a fairly fundamental issue so we should think about it now to avoid having to make code changes later.

pinc/unicode.inc
70ray: Good plan, I hadn't thought of translation. That would make the code simpler also.

pinc/unicode.inc
70ray: I think you are right. 'Combining character sequence' seems to be the official phrase according to [http://www.unicode.org/versions/Unicode12.1.0/ch03.pdf#G30602](url).
I've changed this and all similar things.

pinc/unicode.inc
70ray: This is a bit confusing. It returns the number of characters as it would appear if printed.
So the character Q̣̇ (with dot above and below) would have length 1 although it consists of 3 codepoints: U+0051 U+0323 U+0307 and 5 bytes: 51 cc a3 cc 87.
I've renamed it to combined_strlen.

pinc/unicode.inc
70ray: Yes, there could be a better way of fixing the problem it was intended to solve. PQC checks for an initial '&' to distinguish the sort of problem. Althought it is totally unrealistic I put in & with a combining mark in the test project to see what would happen.

pinc/unicode.inc
70ray: You are right, it doesn't. It was formerly used in `get_invalid_characters()` but I cghanged that. I've put back the original.

pinc/unicode.inc
70ray: Done this

70ray:
This has now been addressed by other branches.

PR #224, opened 2020-02-23
Change way translated strings are transferred to javascript

70ray:
This avoids error message in console about MIME type
and simplifies code

### DIFF
--- a/tools/proofers/preview.inc
+++ b/tools/proofers/preview.inc
@@ -2,14 +2,14 @@
 $relPath="./../../pinc/";
 include_once($relPath.'misc.inc');
 
-function shared_strings($escape_function="html_safe")
+function get_style_strings()
 {
     return array(
-        "italic" => $escape_function(_("Italic")),
-        "bold" => $escape_function(_("Bold")),
-        "gesperrt" => $escape_function(_("Gesperrt")),
-        "small_caps" => $escape_function(_("Small caps")),
-        "font_change" => $escape_function(_("Font change")),
+        "italic" => _("Italic"),
+        "bold" => _("Bold"),
+        "gesperrt" => _("Gesperrt"),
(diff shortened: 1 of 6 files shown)


tools/proofers/preview.inc
```diff
@@ -71,25 +71,20 @@ function output_preview_strings()
...
     {
-        $value = javascript_safe($value);
-        $json_entries[] = "\"$key\": \"$value\"";
+        array_walk($messages, function(&$item) { $item = utf8_encode($item);} );
     }
-    $preview_msg = "{" . implode(",", $json_entries) . "}";
+    $preview_msg = json_encode($messages);
 
-    echo "var previewMessages = JSON.parse('$preview_msg'); var previewDemo = '$demo_string'; var ieWarn = '$ie_warn'";
+    return "var previewMessages = $preview_msg; var previewDemo = '$demo_string'; var ieWarn = '$ie_warn';";
```
cpeel: We should run `javascript_safe()` on `$demo_string` and `$ie_warn` in case the translation contains a single quote.

tools/proofers/preview.inc
```diff
@@ -58,11 +58,11 @@ function get_preview_strings()
     );
 }
 
-function output_preview_strings()
+function preview_strings()
```
cpeel: Consider renaming this `get_preview_js_vars()` or something similar.

tools/proofers/preview.inc
70ray: I think it is, but the code is written in such a strange way it's not obvious. Who wrote this stuff? Oh wait ...
I'll perhaps rewrite it.

cpeel: Ah, yes, this is more straightforward - thanks.

PR #225, opened 2020-02-23
Allow users to skip empty difference pages (task 1518)

mlazaric:
Adds a checkbox which allows users to skip empty difference pages when going to the previous or the next difference page. 

 - Relevant task: https://www.pgdp.net/c/tasks.php?action=show&task_id=1518
 - Deployed here: https://www.pgdp.org/~mlazaric/c/

### Screenshot
image

### Dropdown
Currently it hides the empty differences from the dropdown as well, but we can change it to display them normally or display them as disabled options.

Page `079.png` is an empty difference and it is not shown.

image

### DIFF
--- a/tools/project_manager/diff.php
+++ b/tools/project_manager/diff.php
@@ -16,6 +16,7 @@ $image       = validate_page_image_filename('image', @$_GET['image'], true);
 $L_round_num = get_integer_param($_GET, 'L_round_num', null, 0, MAX_NUM_PAGE_EDITING_ROUNDS);
 $R_round_num = get_integer_param($_GET, 'R_round_num', null, 0, MAX_NUM_PAGE_EDITING_ROUNDS);
 $format = get_enumerated_param($_GET, "format", null, array("keep", "remove"), true);
+$only_nonempty_diffs = @$_GET['only_nonempty_diffs'] === 'on';
 
 $project = new Project( $projectid );
 $state = $project->state;
@@ -115,8 +116,8 @@ output_header("$title: $project_title", NO_STATSBAR, $extra_args);
 echo "<h1>$project_title</h1>\n";
 echo "<h2>$image_link</h2>\n";
 
-do_navigation($projectid, $image, $L_round_num, $R_round_num, 
-              $L_user_column_name, $L_user, $format);
+do_navigation($projectid, $image, $L_round_num, $R_round_num, $L_user_column_name, $L_user, $format,
+              $L_text_column_name, $R_text_column_name, $only_nonempty_diffs);
 echo $navigation_text;
 
 $url = "$code_url/project.php?id=$projectid&amp;expected_state=$state";
@@ -161,8 +162,8 @@ if ($L_text != $R_text)
 
 // build up the text for the navigation bit, so we can repeat it
 // again at the bottom of the page
-function do_navigation($projectid, $image, $L_round_num, $R_round_num, 
-                       $L_user_column_name, $L_user, $format) 
+function do_navigation($projectid, $image, $L_round_num, $R_round_num, $L_user_column_name, $L_user, $format,
+                       $L_text_column_name, $R_text_column_name, $only_nonempty_diffs)
 {
     global $navigation_text;
     $jump_to_js = "this.form.image.value=this.form.jumpto[this.form.jumpto.selectedIndex].value; this.form.submit();";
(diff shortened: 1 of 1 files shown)


chrismiceli:
I feel like I would expect it to be present in drop-down even if checked. That is, unless the checkbox was labelled something like filter empty diffs.

tools/project_manager/diff.php
```diff
@@ -116,7 +117,7 @@
 echo "<h2>$image_link</h2>\n";
 
 do_navigation($projectid, $image, $L_round_num, $R_round_num, 
-              $L_user_column_name, $L_user, $format);
+              $L_user_column_name, $L_user, $format, $L_text_column_name, $R_text_column_name, $only_nonempty_diffs);
```
cpeel: Please move `$L_user_column_name, $L_user, $format,` to the line above and even these out. Ditto in the `do_navigation()` definition too.

mlazaric:
image image

srjfoo:
> I think we should leave the empty diffs listed in the drop-down and have them disabled if "skip empty diffs" is selected. I think that would be less confusing to users.

I like that idea. I think it would be too confusing to have them enabled in the dropdown, but not in the previous/next navigation.

tools/project_manager/diff.php
```diff
@@ -176,46 +176,49 @@ function do_navigation($projectid, $image, $L_round_num, $R_round_num,
...
     $navigation_text .= "\n" . _("Jump to") . ": <select name='jumpto' onChange='$jump_to_js'>\n";
 
-    $condition = "";
-
-    if ($only_nonempty_diffs) {
-        $condition = "WHERE image = '$image' OR $L_text_column_name <> $R_text_column_name";
-    }
-
-    $query = "SELECT image, $L_user_column_name  FROM $projectid $condition ORDER BY image ASC";
+    $query = "SELECT image, $L_user_column_name, ($L_text_column_name = $R_text_column_name) AS is_empty_diff FROM $projectid ORDER BY image ASC";
```
mlazaric: Not sure how expensive this is to run every time:

`($L_text_column_name = $R_text_column_name) AS is_empty_diff`

We only need the values calculated when `$only_nonempty_diffs` is set to true, so we could set it to something like `1 AS is_empty_diff` when it is false. That might be an unnecessary optimization and it would make the code less clear though.

cpeel: Why don't you run an explain on each version of the query, post the results here, and we can evaluate what the database will do to fulfill the query.

mlazaric: Here are the results from the EXPLAIN command:

```
mysql> EXPLAIN SELECT image, round1_user, (master_text = round1_text) AS is_empty_diff FROM projectID463e6a9454485 ORDER BY image ASC;
+----+-------------+------------------------+------------+------+---------------+------+---------+------+------+----------+----------------+
| id | select_type | table                  | partitions | type | possible_keys | key  | key_len | ref  | rows | filtered | Extra          |
+----+-------------+------------------------+------------+------+---------------+------+---------+------+------+----------+----------------+
|  1 | SIMPLE      | projectID463e6a9454485 | NULL       | ALL  | NULL          | NULL | NULL    | NULL |  608 |   100.00 | Using filesort |
+----+-------------+------------------------+------------+------+---------------+------+---------+------+------+----------+----------------+
1 row in set, 1 warning (0.03 sec)
```

```
mysql> EXPLAIN SELECT image, round1_user, 0 AS is_empty_diff FROM projectID463e6a9454485 ORDER BY image ASC;
+----+-------------+------------------------+------------+------+---------------+------+---------+------+------+----------+----------------+
| id | select_type | table                  | partitions | type | possible_keys | key  | key_len | ref  | rows | filtered | Extra          |
+----+-------------+------------------------+------------+------+---------------+------+---------+------+------+----------+----------------+
|  1 | SIMPLE      | projectID463e6a9454485 | NULL       | ALL  | NULL          | NULL | NULL    | NULL |  608 |   100.00 | Using filesort |
+----+-------------+------------------------+------------+------+---------------+------+---------+------+------+----------+----------------+
1 row in set, 1 warning (0.00 sec)
```

I guess there is no meaningful difference? I ran it on this project: https://www.pgdp.org/~mlazaric/c/tools/project_manager/page_detail.php?project=projectID463e6a9454485&show_image_size=0 which has 608 pages and quite a few of the pages are 1000-2000 bytes.

cpeel: Yes, it's doing a tablescan for the query anyway, so they are largely equivalent.

cpeel:
I intend to merge this in tomorrow or Monday.

PR #226, opened 2020-02-25
Make toolbox buttons at least 2ems wide

cpeel:
Make the buttons in the proofreading toolbox at least 2ems wide. This most notably affects the italics button.

[Task 1840](https://www.pgdp.net/c/tasks.php?action=show&task_id=1840)

### DIFF
--- a/styles/layout.css
+++ b/styles/layout.css
@@ -741,6 +741,7 @@ table.dirlist caption {
   margin: 1px 0px;
   cursor: pointer;
   font-size: 0.9em;
+  min-width: 2em;
 }
 #toolbox .key-block {
   display: none;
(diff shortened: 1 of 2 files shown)


cpeel:
> Why do we check in both files? Should we only commit the less and generate the css at build?

Good question. There is no "build" required currently -- the code can be used as-is from the checkout. Adding `less` as a "after you check it out run this" dependency is one more thing that other sites would need to install just to use the code. This is documented in https://github.com/DistributedProofreaders/dproofreaders/blob/master/styles/README.md

chrismiceli:
> 
> 
> > Why do we check in both files? Should we only commit the less and generate the css at build?
> 
> Good question. There is no "build" required currently -- the code can be used as-is from the checkout. Adding `less` as a "after you check it out run this" dependency is one more thing that other sites would need to install just to use the code. This is documented in https://github.com/DistributedProofreaders/dproofreaders/blob/master/styles/README.md

got it thanks!

PR #228, opened 2020-02-25
Remove R* references from the code

cpeel:
This MR is broken out into 3 separate commits, depending on how much of the R* code we want to pull out.

* dde2ab9 does the minimum amount needed for [Task 1842: P2 and F1 requirements still have old R1 and R2 pages as objectives](https://www.pgdp.net/c/tasks.php?action=show&task_id=1842).
* 90de3dc goes one step further and removes the references from the R* rounds from the beginner project exclusion and the "how many pages have you done" exclusion. This one makes sense to pull in as well.
* aee25fc is the last and final step of removing all reference to the R* rounds from the code -- this pulls out all access to R* statistics. I'm dubious of how useful these stats are at this point, but I can see some arguments for keeping them in.

### DIFF
--- a/pinc/Activity.inc
+++ b/pinc/Activity.inc
@@ -15,12 +15,6 @@ $ACCESS_CRITERIA = array(
     'quiz/p_mod2'    => sprintf(_('<a href="%1$s">moderate proofreading quiz %2$d</a> pass'),"$code_url/quiz/start.php?show_level=P_MOD", 2),
     'quiz/f_only'    => sprintf(_("<a href='%s'>formatting quiz</a> pass"),"$code_url/quiz/start.php?show_only=format"),
 );
-if ( TRUE )
-{
-    // kludge
-    // TRANSLATORS: %s is a round ID, such as P1 or P1+P2
-    $ACCESS_CRITERIA['R*+P1'] = sprintf( _("'%s' pages completed"), 'R1+R2+P1' );
-}
 
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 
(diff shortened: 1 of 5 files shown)


srjfoo:
>     * [aee25fc](https://github.com/DistributedProofreaders/dproofreaders/commit/aee25fce116820faba6d82626e6bb2ff4e2881b8) is the last and final step of removing all reference to the R* rounds from the code -- this pulls out all access to R* statistics. I'm dubious of how useful these stats are at this point, but I can see some arguments for keeping them in.

If I were someone who had done several thousand pages in R1/R2, I think I would want to be able to still see that documented somewhere, and the user's details page is the most appropriate place for that. It doesn't matter that they don't count toward any other access, but it's part of the user's history.

lhamilton1:
I agree that we should be able to see how many R* pages were done both in total and for specific users, though we should no longer mention R* rounds as qualifying criteria. I use the Statistics Central information, including the R* info, myself when memorializing volunteers for example.

cpeel:
That makes sense. I will drop the last commit and check out Michael's feedback on the first one.

srjfoo:
> That makes sense. I will drop the last commit and check out Michael's feedback on the first one.

Will you just leave the code the way it is, or make the appearance of the R*+P1 column conditional on a proofer having R* pages? If the former, we'll still get questions on "what does that mean".

cpeel:
I will leave the code the way it is. This means that the only place R* will show up is on the stats pages.

lhamilton1:
That sounds good. Thanks, Casey!

jmdyck:
> ... make the appearance of the R*+P1 column conditional on a proofer having R* pages?

Offhand, I suspect that would involve some fairly ugly code.

cpeel:
> * first commit: looks correct to me.
> 
> In get_user_score(), the first block of code handles the case where a $criterion_code contains a '+'. I believe this commit removes the only use of this code, so you could remove it or not, depending on whether you think it might be useful in future (or is being used by other DP sites). But if it stays, it should maybe get a comment saying that it was there to handle 'R*+P1' and isn't currently used (at pgdp.net).

I opted to remove it.

> longabsent.php could conceivably get a sentence saying that your R* tally no longer counts toward satisfying access-minima, except that it never said that R* _did_ count toward minima, so maybe just leave it.

I went for "don't fix what ain't broke" and left it alone.

Thanks for the input.

srjfoo:
> > ... make the appearance of the R*+P1 column conditional on a proofer having R* pages?
> 
> Offhand, I suspect that would involve some fairly ugly code.

Good point. We'll just deal with the questions as they arise, then.

jmdyck:
Revised PR looks good to me.

PR #229, opened 2020-02-27
Multi point

70ray:
This shows multiple codepoints as well as bytes in the Bad Bytes tables.
Bad sequences are handled separately from invalid codepoints; this is to pre-empt a problem that would occur if considering combined characters rather than codepoints.
BOM handling is also improved.

### DIFF
--- a/pinc/bad_bytes.inc
+++ b/pinc/bad_bytes.inc
@@ -786,24 +786,16 @@ define_bad_byte_sequences();
 
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 
-function find_bad_byte_sequences_in_text($text, $projectid)
+function find_bad_byte_sequences_in_text($text)
 {
     global $_unanchored_bad_bytes_regex, $_bad_byte_sequences;
 
     $occurrences = array();
 
-    // find any characters in the text that are not valid for the project
     if(utf8_string_has_bom($text))
     {
         $occurrences['BOM'] = 1;
         $text = utf8_str_replace(utf8_bom(), '', $text);
-    }
-
-    $project = new Project($projectid);
-    $invalid_chars = get_invalid_characters($text, $project->get_valid_codepoints());
-    foreach($invalid_chars as $char => $num)
-    {
-        $occurrences[$char] = $num;
     }
 
     $n_matches = preg_match_all($_unanchored_bad_bytes_regex, $text, $matches);
@@ -898,30 +890,22 @@ function get_remarks_for_bad_byte_sequence($raw)
 
         $why_bad = "HTML numeric character reference";
     }
-    elseif(utf8_strlen($raw) == 1 && is_utf8($raw))
-    {
-        $why_bad = "Codepoint not valid for project";
-        $intended_character = "";
-    }
     else
     {
         list($code_point, $why_bad) = $_bad_byte_sequences[$raw];
     }
 
-    if(!isset($intended_character))
+    @$info = $_character_data[$code_point];
+    if ($info)
     {
-        @$info = $_character_data[$code_point];
-        if ($info)
-        {
-            $char_name = $info['name'];
-        }
-        else
-        {
-            // $_character_data doesn't have an entry for that code point
-            $char_name = '???';
-        }
-        $intended_character = sprintf("U+%04x %s", $code_point, $char_name);
+        $char_name = $info['name'];
     }
+    else
+    {
+        // $_character_data doesn't have an entry for that code point
+        $char_name = '???';
+    }
+    $intended_character = sprintf("U+%04x %s", $code_point, $char_name);
 
     return array($intended_character, $why_bad);
 }
--- a/pinc/bad_bytes.inc
+++ b/pinc/bad_bytes.inc
@@ -788,6 +788,7 @@ define_bad_byte_sequences();
 
 define("BOM", "\u{feff}");
 
+// note that this can change $text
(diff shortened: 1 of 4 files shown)


70ray:
Running in [https://www.pgdp.org/~rp31/pm/tools/project_manager/project_quick_check.php?projectid=projectID5c13a64918b30](url) (with a project with two bad pages)

pinc/unicode.inc
```diff
@@ -170,3 +170,15 @@ function guess_string_encoding($text)
...
 }
+
+function codepoints_to_hex($string)
+{
+    // find the components of combined character as hex strings
+    $codepoints = [];
+    preg_match_all("/./u", $string, $matches);
+    foreach($matches[0] as $match)
+    {
+        $codepoints[] = str_pad(dechex(IntlChar::ord($match)), 4, "0", STR_PAD_LEFT);
```
cpeel: Please check in `portable-utf8.php` as I suspect there is already a function that does this.

tools/project_manager/bad_bytes_explainer.php
```diff
@@ -153,6 +153,7 @@ function show_a_table_of_bads($desired_badness)
         <table class='basic striped'>
         <tr>
             <th>" . _("Raw") . "</th>
+            <th>" . _("Codepoints") . "</th>
             <th>" . _("Bytes") . "</th>
```
cpeel: I would prefer we put the Bytes column ahead of the Bytes column. I think we should always show both columns instead of jumping through the hoops with the colspans (for HTML entity fields the Codepoints field would be empty). We should retain the 'U+' syntax for the codepoint column so it's clearer these are Unicode character values.

pinc/bad_bytes.inc
```diff
@@ -786,26 +786,21 @@ define_bad_byte_sequences();
 
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 
-function find_bad_byte_sequences_in_text($text, $projectid)
+define("BOM", "\u{feff}");
+
+function find_bad_byte_sequences_in_text(&$text)
```
cpeel: I really don't like pass-by-reference in PHP because it's unexpected and confusing since almost everything else is pass-by-value. Please add a comment that this will modify the value of `$text`, both here and where it is called.

pinc/bad_bytes.inc
```diff
@@ -898,30 +897,22 @@ function get_remarks_for_bad_byte_sequence($raw)
...
-        $why_bad = "Codepoint not valid for project";
-        $intended_character = "";
-    }
     else
     {
         list($code_point, $why_bad) = $_bad_byte_sequences[$raw];
     }
 
-    if(!isset($intended_character))
+    @$info = $_character_data[$code_point];
```
cpeel: `if(isset($_character_data[$code_point]))` would be clearer than using `@`.

pinc/project_quick_check.inc
```diff
@@ -323,19 +323,24 @@ function _test_project_for_bad_bytes($projectid)
...
 
     $n_bad_pages = 0;
+    $valid_codepoints = $project->get_valid_codepoints();
+
     while (list($text, $imagename, $proofers) = page_info_fetch($pages_res))
     {
         $round_num = get_round_num_from_proofers($proofers);
-        $occurrences = find_bad_byte_sequences_in_text($text, $projectid);
+        $occurrences = find_bad_byte_sequences_in_text($text);
+        $invalid_chars = get_invalid_characters($text, $valid_codepoints);
```
cpeel: I understand that you want to output characters different than bytes. Can you explain why you needed to split this out to do that rather than change `tds_for_bad_bytes()` to output different data for characters vs other bad bytes? Doing it this way neuters `find_bad_byte_sequences_in_text()`.

pinc/unicode.inc
70ray: There is something useful.

tools/project_manager/bad_bytes_explainer.php
70ray: Done this.

pinc/bad_bytes.inc
70ray: Done this.

pinc/bad_bytes.inc
70ray: I agree. Found the null coalescing operator.

pinc/project_quick_check.inc
70ray: The problem was that `find_bad_byte_sequences_in_text()` just returns the strings which you then have to interpret in `tds_for_bad_bytes()` to work out what the problem was. An ambiguity can arise with a string starting with '&' which could in principle (though unlikely) be a combining sequence (when that part gets written) rather than a html entity so in the combining sequence branch I had to put in a 'hack' to distinguish that case. The former code used the `elseif(utf8_strlen($raw) == 1 && is_utf8($raw))`  line to distinguish invalid codepoints which also needed further adjustment to distinguish invalid combining characters.
Dealing with invalid characters and bad sequences separately avoids these problems.
Another approach would be to record more information in the '$occurrences' array so we did not  have to re-interpret the results but this would probably be more complicated.

70ray:
Following my first attempt which screwed up the 'rowspan' in the table, while looking to see how this should work I came across this comment  [https://developer.mozilla.org/en-US/docs/Web/HTML/Element/table#Accessibility_concerns](url)   which says that tables with rowspans are not only difficult to write but also difficult to read and should be avoided, and wondered whether the page and round columns would be better as a caption with a separate table for each page.

70ray:
This has been re-made to use new unicode features.

PR #230, opened 2020-02-28
implemented quiz diacritical conversion

70ray:
Introduced new trait CharSuiteSet used in class Project and new class
QuizCharsuiteSet implementing get_pickersets and get_valid_codepoints
renamed projectid to charsuitesid where appropriate.
Now gets basic latin for quizes but could be extended to give different charsuite sets for different quizes.

### DIFF
--- a/pinc/CharSuites.inc
+++ b/pinc/CharSuites.inc
@@ -178,3 +178,44 @@ foreach($charsuite_files as $charsuite_file)
 {
     include_once($charsuite_file);
 }
+
+trait CharSuiteSet
+{
+    public function get_pickersets()
+    {
+        $pickersets = [];
+        foreach($this->get_charsuites() as $charsuite)
+        {
+            $pickersets[] = $charsuite->pickerset;
+        }
+        return $pickersets;
+    }
+
+    public function get_valid_codepoints()
+    {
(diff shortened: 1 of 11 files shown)


70ray:
Now running in sandbox [https://www.pgdp.org/~rp31/preview/quiz/start.php](url)

pinc/CharSuites.inc
```diff
@@ -178,3 +178,44 @@ foreach($charsuite_files as $charsuite_file)
 {
     include_once($charsuite_file);
 }
+
+trait CharSuiteSet
```
cpeel: I don't think this level of complexity is necessary just for quizzes. The hack in `pinc/CharacterSelector.inc` isn't great, but it's at least localized there instead of spilling out into multiple places that is only used for the quizzes.

70ray: Perhaps I have been over-assiduous in renaming variables in various files which I will undo.

cpeel: Your variable names are all good here, I just don't think we need this level of abstraction to solve this one-off case for quizzes.

70ray:
Revised branch made.

PR #231, opened 2020-02-28
change styling so french format preview lines do not wrap

70ray:
remove defined width for confiure style lines and change to flex style


cpeel:
Can you give an example of the problem this is fixing? I wasn't able to replicate any difference between TEST and this code in my sandbox for the Preview config page using French, but I'm probably looking at the wrong thing.

70ray:
This is what it looks like before:
image

and after

PR #233, opened 2020-02-29
Enable diacritical markup conversion in quizes

70ray:
pickerset and valid_codepoint functions moved out of project.inc
so we can use them if there isn't a project.
New function to construct charsuites for quizes.

### DIFF
--- a/pinc/CharSuites.inc
+++ b/pinc/CharSuites.inc
@@ -208,23 +208,11 @@ trait CharSuiteSet
     }
 }
 
-class QuizProject
-{
-    use CharSuiteSet;
-
-    public function get_charsuites()
-    {
-        $charsuites[] = CharSuites::get('basic-latin');
-        $charsuites[] = CharSuites::get('extended-european-latin');
-        return $charsuites;
-    }
-}
-
 function get_real_or_quiz_project($identifier=NULL)
 {
     if($identifier == "quiz")
     {
-        return new QuizProject();
+        return new QuizCharSuites();
     }
     else
     {
--- a/pinc/CharSuites.inc
+++ b/pinc/CharSuites.inc
@@ -179,44 +179,55 @@ foreach($charsuite_files as $charsuite_file)
     include_once($charsuite_file);
 }
 
-function get_pickersets($charsuites)
+trait CharSuiteSet
 {
-    $pickersets = [];
-    foreach($charsuites as $charsuite)
+    public function get_pickersets()
     {
-        $pickersets[] = $charsuite->pickerset;
+        $pickersets = [];
+        foreach($this->get_charsuites() as $charsuite)
+        {
+            $pickersets[] = $charsuite->pickerset;
+        }
+        return $pickersets;
     }
-    return $pickersets;
-}
 
-function get_valid_codepoints($charsuites)
-{
-    # Codepoints applicable to all projects
-    $global_codepoints = [
-        'U+000A', # new lines
-        'U+000D', # carriage returns
-    ];
-
-    $codepoints = $global_codepoints;
-    foreach($charsuites as $charsuite)
+    public function get_valid_codepoints()
     {
-        $codepoints = array_merge($codepoints, $charsuite->codepoints);
-    }
-    return $codepoints;
-}
-
-function get_charsuites($identifier=NULL)
-{
-    try
-    {
-        $project = new Project($identifier);
-        return $project->get_charsuites();
-    }
-    catch (NonexistentProjectException $e)
-    {
-        return [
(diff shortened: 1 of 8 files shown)


tools/proofers/ctrl_frame.php
```diff
@@ -8,11 +8,12 @@
...
 
-
-$projectid = validate_projectID($_GET, $_GET["projectid"]);
 $round_id = get_enumerated_param($_GET, 'round_id', null, array_keys($Round_for_round_id_));
 $round = get_Round_for_round_id($round_id);
-$projectid = validate_projectID('projectid', @$_GET['projectid']);
+// if this is used in a quiz there will not be a projectid, 'quiz' will be used
+// for the MRU local storage prefix and the toolbox output will discover the
+// projectid is invalid and provide the quiz charsuites
+$projectid = array_get($_GET, 'projectid', 'quiz');
```
cpeel: We should still validate that `$projectid` is either a valid projectid or "quiz" here. Could be done after this line with:
```
if($projectid != "quiz")
{
    $projectid = validate_projectID('projectid', $projectid);
}
```

pinc/CharSuites.inc
```diff
@@ -178,3 +178,45 @@ foreach($charsuite_files as $charsuite_file)
...
+
+function get_charsuites($identifier=NULL)
+{
+    try
+    {
+        $project = new Project($identifier);
+        return $project->get_charsuites();
+    }
+    catch (NonexistentProjectException $e)
+    {
```
cpeel: Since we're passing in "quiz" for the identifier (which I think is great) we should explicitly check against that:
```
    if($identifier == "quiz")
    {
        return [
            CharSuites::get('basic-latin'),
            CharSuites::get('extended-european-latin'),
        ];
    }
    else
    {
        $project = new Project($identifier);
        return $project->get_charsuites();
    }
```
Then we don't even need to add a comment of why we return something different if the project doesn't exist.

pinc/Project.inc
```diff
@@ -243,32 +243,6 @@ class Project
         return $charsuites;
     }
 
-    public function get_valid_codepoints()
```
cpeel: You need to update the other instances where this is called -- `pinc/DPage.inc` and `pinc/bad_bytes.inc`

70ray: Yes, this occured to me later also.

pinc/CharSuites.inc
70ray: Good idea. I've done this slightly differently.

pinc/Project.inc
70ray: New commit should avoid needing to do this.

tools/proofers/ctrl_frame.php
70ray: Done this.

pinc/CharSuites.inc
```diff
@@ -178,3 +178,45 @@ foreach($charsuite_files as $charsuite_file)
 {
     include_once($charsuite_file);
 }
+
+function get_pickersets($charsuites)
```
cpeel: I also wonder if these should be static functions inside the `CharSuites` class to help indicate what they are getting the pickersets from:
```
// Given a character suite (by object or name, or an array of either)
// return the picker sets associated with them.
public static function get_pickersets($charsuites)
{
    if(!is_array($charsuites))
    {
        $charsuites = [ $charsuites ];
    }

    $pickersets = [];
    foreach($charsuites as $charsuite)
    {
        $charsuite = CharSuites::resolve($charsuites);
        $pickersets[] = $charsuite->pickerset;
    }
    return $pickersets;
}
```

Ditto `get_valid_codepoints()`.

70ray: They will be in effect inside a project class which will have the same sort of effect.

pinc/CharSuites.inc
```diff
@@ -179,44 +179,55 @@ foreach($charsuite_files as $charsuite_file)
...
+        {
+            $codepoints = array_merge($codepoints, $charsuite->codepoints);
+        }
+        return $codepoints;
     }
-    return $pickersets;
 }
 
-function get_valid_codepoints($charsuites)
+class QuizProject
```
cpeel: This should be in its own `pinc/QuizProject.inc` file. And QuizProject doesn't make much sense -- why not just call it `Quiz` and, in the future, we can move other quiz-related stuff in here.

70ray: There is already a Quiz class and Quiz.inc. I moved it to Quiz.inc and called it QuizCharSuites.

PR #234, opened 2020-02-29
add char suites to project info page

chrismiceli:
add edit for char suites in edit project page

### DIFF
--- a/project.php
+++ b/project.php
@@ -404,15 +404,9 @@ function do_project_info_table()
     $charsuites = [];
     foreach($project_charsuites as $project_charsuite)
     {
-        array_push($charsuites, $project_charsuite->title);
+        array_push($charsuites, "<a href='tools/charsuites.php?charsuite=" . attr_safe($project_charsuite->name) . "'>" . html_safe($project_charsuite->title) . "</a>");
     }
-    echo_row_a(
-        _("Character Suites"),
-        implode(", ", $charsuites) .
-          " &middot; <a href='tools/charsuites.php?projectid=$projectid'>".
-          _("View project's character suites") .
-          "</a>",
-        FALSE);
+    echo_row_a(_("Character Suites"), implode(", ", $charsuites));
 
     echo_row_a( _("Difficulty"),      _($project->difficulty) );
 
@@ -565,6 +559,10 @@ function do_project_info_table()
 
         echo_row_a( _("Word Lists"), $links );
     }
+
+    echo_row_a( _("Character Suites"),
+        "<a href='tools/charsuites.php?projectid=$projectid'>" . _("View project's character suites") . "</a>",
+        FALSE);
 
     // -------------------------------------------------------------------------
 
--- a/project.php
+++ b/project.php
@@ -409,7 +409,7 @@ function do_project_info_table()
     echo_row_a(
         _("Character Suites"),
         implode(", ", $charsuites) .
-          " | <a href='tools/charsuites.php?projectid=$projectid'>".
+          " &middot; <a href='tools/charsuites.php?projectid=$projectid'>".
           _("View project's character suites") .
           "</a>",
         FALSE);
--- a/project.php
+++ b/project.php
@@ -406,7 +406,13 @@ function do_project_info_table()
     {
         array_push($charsuites, $project_charsuite->title);
     }
-    echo_row_a( _("Character Suites"), implode(", ", $charsuites));
+    echo_row_a(
+        _("Character Suites"),
+        implode(", ", $charsuites) .
+          " | <a href='tools/charsuites.php?projectid=$projectid'>".
+          _("View project's character suites") .
+          "</a>",
+        FALSE);
 
     echo_row_a( _("Difficulty"),      _($project->difficulty) );
 
@@ -559,10 +565,6 @@ function do_project_info_table()
 
         echo_row_a( _("Word Lists"), $links );
     }
-
-    echo_row_a( _("Character Suites"),
-        "<a href='tools/charsuites.php?projectid=$projectid'>" . _("View project's character suites") . "</a>",
-        FALSE);
 
     // -------------------------------------------------------------------------
 
--- a/project.php
+++ b/project.php
@@ -399,6 +399,15 @@ function do_project_info_table()
     echo_row_a( _("Author"),          $project->authorsname );
     echo_row_a( _("Language"),        $project->language );
     echo_row_a( _("Genre"),           _($project->genre) );
+
+    $project_charsuites = $project->get_charsuites();
+    $charsuites = [];
+    foreach($project_charsuites as $project_charsuite)
+    {
+        array_push($charsuites, $project_charsuite->title);
+    }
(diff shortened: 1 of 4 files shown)


chrismiceli:
Project page:
image

edit project page:
image

srjfoo:
Currently, the "Character Suites" line on the project page shows a link: 
image
The page displayed shows all character suites enabled. It's not a bad idea to actually list the different character suites instead of a generic phrase, but it would probably be a good idea to provide a link to the charsuite display page. (At least it doesn't look like you have links at this point.)

tools/project_manager/edit_common.inc
```diff
@@ -94,6 +95,39 @@ function genre_list($genre)
...
+    {
+        $existing_charsuites[$charsuite] = TRUE;
+    }
+
+    echo "<select name='charsuites[]' required multiple size='3'>\n";
+    foreach($all_charsuites as $k=>$v)
+    {
+        echo "<option value='".attr_safe($k)."'";
+        if ($existing_charsuites[$k]) { echo " SELECTED"; }
+        echo ">$v</option>";
```
cpeel: We should `html_safe($v)` the title too.

tools/project_manager/editproject.php
```diff
@@ -440,9 +448,21 @@ function set_from_post()
...
             : $pri_language );
 
+        $this->charsuites = [];
+        foreach(@$_POST['charsuites'] as $charsuite)
+        {
+            array_push($this->charsuites, $charsuite);
+        }
+        if (sizeof($this->charsuites) == 0)
+        {
+            $errors .= _("Character Suites is required.")."<br>";
```
cpeel: Perhaps phrase as "At least one Character Suite is required."?

tools/project_manager/edit_common.inc
```diff
@@ -94,6 +95,39 @@ function genre_list($genre)
 
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 
+function charsuite_list($charsuites, $project_charsuites)
+{
+    $global_charsuites = CharSuites::get_enabled();
```
cpeel: Consider renaming this to `$enabled_charsuites` for clarity.

tools/project_manager/editproject.php
```diff
@@ -951,6 +970,14 @@ function show_visible_controls()
         }
         $this->row( _("Language"),                    'language_list',       $this->language         );
         $this->row( _("Genre"),                       'genre_list',          $this->genre            );
+
+        $project_charsuites = [];
+        if (isset($this->projectid))
+        {
+            $project = new Project($this->projectid);
+            $project_charsuites = $project->get_charsuites();
+        }
```
cpeel: We should probably add an `else` here that adds the basic-latin character suite to `$project_charsuites` for new projects 
 (if basic-latin is enabled on the site) for two reasons:
1. Upon initial roll-out we will only have basic-latin enabled and having it selected makes sense for 100% of the time.
2. After roll-out, it is very likely that for pgdp.net 99% of the projects will always use basic-latin.

One downside is that for the 1% of pgdp.net projects that don't use basic-latin they will have to unselect it, but that seems reasonable given the common case.

tools/project_manager/edit_common.inc
```diff
@@ -94,6 +95,39 @@ function genre_list($genre)
...
+        $all_charsuites[$project_charsuite->name] = $project_charsuite->title;
+    }
+    $existing_charsuites = [];
+    foreach($charsuites as $charsuite)
+    {
+        $existing_charsuites[$charsuite] = TRUE;
+    }
+
+    echo "<select name='charsuites[]' required multiple size='3'>\n";
+    foreach($all_charsuites as $k=>$v)
```
cpeel: We should sort `$all_charsuites` by value (title of the charsuite) so they are always presented in a consistent fashion to the user.

cpeel:
I agree that listing the individual character suites on the project page makes more sense than just having the link. I'm torn if we should link to the individual charsuites or just to the current "all suites for this project" however. One thing that is nice about the latter is that you can see all of the characters valid for the project on one page without having to open up each charsuite.

srjfoo:
> One thing that is nice about the latter is that you can see all of the characters valid for the project on one page without having to open up each charsuite.

True -- I like that idea.

tools/project_manager/editproject.php
chrismiceli: Yeah I thought that might come up. I think it makes sense so defaulting to basic-latin now as it was before.

tools/project_manager/edit_common.inc
```diff
@@ -94,6 +95,39 @@ function genre_list($genre)
 
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 
+function charsuite_list($charsuites, $project_charsuites)
+{
+    $enabled_charsuites = CharSuites::get_enabled();
+
+    $all_charsuites = [];
+    foreach($enabled_charsuites as $global_charsuite)
```
cpeel: We should also rename `$global_charsuite` to `$enabled_charsuite` so it matches the plural version.

tools/project_manager/editproject.php
```diff
@@ -169,6 +169,7 @@ function set_from_nothing()
         $this->comments         = '';
         $this->clearance        = '';
         $this->postednum        = '';
+        $this->charsuites       = ['basic-latin'];
```
cpeel: While it will for pgdp.net, we can't guarantee that 'basic-latin' is always going to be enabled for all sites and this would cause problems. You could either do the "if basic-latin is an enabled charsuite then use it" logic here and on line 210 or in a helper function in this file that does the same thing.

tools/project_manager/edit_common.inc
cpeel: Looks like you missed updating this one.

project.php
```diff
@@ -399,6 +399,21 @@ function do_project_info_table()
     echo_row_a( _("Author"),          $project->authorsname );
     echo_row_a( _("Language"),        $project->language );
     echo_row_a( _("Genre"),           _($project->genre) );
+
```
cpeel: Let's move this back down to where it was prior. This bit of info isn't super useful to most proofers whereas some of the other bits below it are.

chrismiceli:
> Currently, the "Character Suites" line on the project page shows a link:
> image
> The page displayed shows all character suites enabled. It's not a bad idea to actually list the different character suites instead of a generic phrase, but it would probably be a good idea to provide a link to the charsuite display page. (At least it doesn't look like you have links at this point.)

I did that in the latest iteration:
image

tools/project_manager/editproject.php
chrismiceli: ok, went ahead and use empty if basic-latin isn't enabled. Then they'll get an error if they save without one.

PR #238, opened 2020-03-07
Test combine

70ray:
This consists of parts from my and @cpeel 's branches to enable combining characters to be used. It is not intended to be merged as it is. It contains an override to enable testing of projects with valid combining characters.
It was really intended to put everything together I've been looking at before I forget how it is supposed to work.
It works for all the aspects I could think of. There are two issues I came across while testing (one pre-existing) which I'll look into.


70ray:
This has been handled in another branch.

PR #240, opened 2020-03-07
UTF-8 helper functions and intro to combined chars

cpeel:
Add additional UTF-8 helper functions when dealing with codepoints
and strings. This also introduces the concept of combined characters
to the code and a format to specify them, although does not include
the full implementation.

This was a collaboration between cpeel and 70ray.

### DIFF
--- a/SETUP/tests/UnicodeTest.php
+++ b/SETUP/tests/UnicodeTest.php
@@ -5,12 +5,60 @@ class UnicodeTests extends PHPUnit\Framework\TestCase
     private $a_to_z_codepoints = [
         'U+0061-U+007a',  # a-z
     ];
+    private $combined_codepoints = [
+        'U+004E',  # N
+        'U+004E>U+0305',  # N̅
+        'U+004E>U+0306',  # N̆
+        'U+004E>U+0307',  # Ṅ
+    ];
 
     public function testSubstrReplace()
     {
         $string = "ЖЖaac";
         $new_string = utf8_substr_replace($string, "bb", 2, 2);
         $this->assertEquals("ЖЖbbc", $new_string);
+    }
+
+    public function testUtf8CombinedChr()
+    {
+        $codepoint = 'U+004E';  # N
+        $char = utf8_combined_chr($codepoint);
+        $this->assertEquals("N", $char);
+    }
+
+    public function testUtf8CombinedChrCombined()
+    {
+        $codepoint = 'U+004E>U+0305';  # N̅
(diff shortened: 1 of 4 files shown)


pinc/unicode.inc
```diff
@@ -61,27 +87,31 @@ function utf8_filter_out_codepoints($string, $remove_codepoints, $replacement=""
...
+            # we have a character range
+            list($start, $end) = explode('-', $codepoint);
+            $return_array = array_merge($return_array, utf8_range($start, $end));
         }
         else
         {
-            list($start, $end) = explode('-', $codepoint);
-            $return_array = array_merge($return_array, utf8_range($start, $end));
+            # a single character or a combined character
+            $return_array[] = utf8_combined_chr($codepoint);
```
70ray: Need to deal with single codepoints separately since the  version of utf8_combined_chr() in this file doesn't check for single characters.

SETUP/tests/UnicodeTest.php
```diff
@@ -13,6 +19,48 @@ public function testSubstrReplace()
...
+        $codepoint = 'U+004E';  # N
+        $char = utf8_combined_chr($codepoint);
+        $this->assertEquals("N", $char);
+    }
+
+    public function testUtf8CombinedChrCombined()
+    {
+        $codepoint = 'U+004E>U+0305';  # N̅
+        $char = utf8_combined_chr($codepoint);
+        $this->assertEquals("N̅", $char);
```
70ray: We should really write this as
```
$this->assertEquals("\u{4e}\u{305}", $char);
```
It doesn't make any difference in this case but if we did the test for 'U+004E>U+0307'
and wrote the assertion for
```
 "\u{4e}\u{307}"
```
then it would fail because the function would normalise it to 
```
"\u{1e44}"
```
although both look like "Ṅ"

pinc/unicode.inc
cpeel: That's the beautify of your `utf8_combined_chr()` version -- it does!

SETUP/tests/UnicodeTest.php
cpeel: Good call - fixed.

cpeel:
> It would be nice to have an automatic test to check that the combined character definitions in the charsuite...inc files are in normalised form.

Agreed. I created [a task](https://www.pgdp.org/c.unicode/tasks.php?action=show&task_id=2268) for it and will get to that today probably.

cpeel:
@70ray confirmed in Slack that this was good to merge.

PR #243, opened 2020-03-09
Make validator window full height even if little text.

70ray:
This is how it was before the changes for handle_bad_page.

### DIFF
--- a/styles/global.css
+++ b/styles/global.css
@@ -45,6 +45,9 @@
 .full-width {
   width: 100%;
 }
+.full-height {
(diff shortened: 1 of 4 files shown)


chrismiceli:
Can you share screenshots of before and after?

chrismiceli:
Looks good to me. Can you share screenshots of before and after though demonstrating why the change is needed?

70ray:
Here are two screenshots showing the effect of the change. 
Here is 'before':
image

Here is screenshot for after:
image

PR #244, opened 2020-03-10
Validate codepoints in a charsuite are normalized

cpeel:
This validates that all codepoints in a character suite are normalized to prevent non-normalized characters from being inserted into proofreading texts and confusing the user when validation fails, among other things.

### DIFF
--- a/SETUP/lint_charsuites.php
+++ b/SETUP/lint_charsuites.php
@@ -1,0 +1,19 @@
+#!/usr/bin/env php
+<?php
+$relPath="../pinc/";
+include_once($relPath."CharSuites.inc");
+
+foreach(CharSuites::get_all() as $charsuite)
+{
+    echo "Validating charsuite $charsuite->name...\n";
+    $nonnormalized_codepoints = $charsuite->get_nonnormalized_codepoints();
+    if($nonnormalized_codepoints)
+    {
+        echo "ERROR: codepoints are not normalized:\n";
+        foreach($nonnormalized_codepoints as $orig => $norm)
+        {
+            echo sprintf("    %s normalized is %s\n", $orig, $norm);
+        }
+        exit(1);
+    }
+}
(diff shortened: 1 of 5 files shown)


cpeel:
> Looks good but I can't see the 'lint_charsuites.php' file.

UGH. I missed adding it and our CI/CD passed because it wasn't calling the `Makefile` to do the tests. I've fixed the CI/CD and added the file and confirmed it is working.

pinc/CharSuites.inc
```diff
@@ -83,6 +83,13 @@ class CharSuite
         $enabled_sets = CharSuites::get_enabled();
         return in_array($this, $enabled_sets);
     }
+
+    // Validate that combined characters in the charsuite are normalized
```
70ray: I can't connect to unicode.org at the moment to check but I seem to remember there could be single codepoints that normalise to others such as Å. If this is so we could change this comment to leave out combined.

SETUP/lint_charsuites.php
```diff
@@ -0,0 +1,14 @@
...
+<?php
+$relPath="../pinc/";
+include_once($relPath."CharSuites.inc");
+
+foreach(CharSuites::get_all() as $charsuite)
+{
+    echo "Validating charsuite $charsuite->name...\n";
+    if(!$charsuite->are_codepoints_normalized())
+    {
+        echo "    ERROR: codepoints are not normalized\n";
```
70ray: It would be useful to print out what the character was.

cpeel: That's true. I struggled with how to best do that given the framework that I have right now but decided to punt on that for the time being.

pinc/CharSuites.inc
cpeel: You are absolutely correct and we should be checking for all the characters in the range, not just specific points. I've updated the code accordingly.

SETUP/lint_charsuites.php
cpeel: 7 hours later I figured out how I wanted to do it and just did it.

PR #246, opened 2020-03-11
rename replace_images to update_illos

chrismiceli:
restrict update_illos to only work with illustrations
add delete operation to update_illos page

image index page:
image

Delete confirmation:
image

delete success:
image

non illustration or invalid file:
image

### DIFF
--- a/tools/project_manager/update_illos.php
+++ b/tools/project_manager/update_illos.php
@@ -22,7 +22,7 @@ $image     = @$_REQUEST['image'];
 if (!preg_match('/^\w[\w.-]*\.(png|jpg)$/', $image)) // see _check_file() in add_files.php
 {
     // This should only happen if someone is URL-tweaking.
-    die(_("Parameter 'image' is not a valid image filename."));
+    die(_("Parameter 'image' is not a valid illustration filename."));
 }
 
 $project = new Project($projectid);
@@ -30,9 +30,9 @@ $operation = get_enumerated_param($_REQUEST, 'operation', 'replace', array('repl
 
 $operation_image_str;
 if ($operation == 'replace') {
-    $operation_image_str = _('Replace Image');
+    $operation_image_str = _('Replace Illustration');
 } else {
-    $operation_image_str = _('Delete Image');
+    $operation_image_str = _('Delete Illustration');
 }
 
 output_header("$operation_image_str: {$project->nameofwork}");
--- a/tools/project_manager/update_illos.php
+++ b/tools/project_manager/update_illos.php
@@ -7,6 +7,7 @@ include_once($relPath.'theme.inc');
 include_once($relPath.'Project.inc');
 include_once($relPath.'misc.inc'); // get_upload_err_msg, get_enumerated_param, attr_safe, html_safe
 include_once($relPath.'project_states.inc'); // PROJ_NEW
+include_once($relPath.'abort.inc'); // provide_escape_links
 
 require_login();
 
@@ -14,24 +15,35 @@ require_login();
 // They should perhaps be refactored.)
 
 $projectid = validate_projectID('projectid', @$_REQUEST['projectid']);
+$operation = get_enumerated_param($_REQUEST, 'operation', 'replace', array('replace', 'delete', 'delete_all'));
 
-$image     = @$_REQUEST['image'];
-// Note that using validate_page_image_filename() here would be inappropriate,
-// because it's specific to *page* images (which this isn't),
-// and so enforces a maximum filename length.
-if (!preg_match('/^\w[\w.-]*\.(png|jpg)$/', $image)) // see _check_file() in add_files.php
+if ($operation !== 'delete_all')
 {
-    // This should only happen if someone is URL-tweaking.
-    die(_("Parameter 'image' is not a valid illustration filename."));
+    $image     = @$_REQUEST['image'];
+    // Note that using validate_page_image_filename() here would be inappropriate,
+    // because it's specific to *page* images (which this isn't),
+    // and so enforces a maximum filename length.
+    if (!preg_match('/^\w[\w.-]*\.(png|jpg)$/', $image)) // see _check_file() in add_files.php
+    {
+        // This should only happen if someone is URL-tweaking.
+        die(_("Parameter 'image' is not a valid illustration filename."));
+    }
 }
 
 $project = new Project($projectid);
-$operation = get_enumerated_param($_REQUEST, 'operation', 'replace', array('replace', 'delete'));
 
-if ($operation == 'replace') {
+if ($operation == 'replace')
+{
     $operation_image_str = _('Replace Illustration');
-} else {
+}
+elseif ($operation == 'delete')
+{
     $operation_image_str = _('Delete Illustration');
+}
+else
+{
+    $operation_image_str = _('Delete All Illustrations');
+    $is_delete_all_operation = true;
 }
 
 output_header("$operation_image_str: {$project->nameofwork}");
@@ -42,7 +54,12 @@ echo "<p>";
 echo "<b>" . _("Project") . ":</b> " . html_safe($project->nameofwork) . "<br>";
 echo "<b>" . _("Project ID") . ":</b> {$project->projectid}<br>";
 echo "<b>" . _("Project state") . ":</b> {$project->state}<br>";
-echo "<b>" . _("Illustration") . ":</b> $image<br>";
+if (!$is_delete_all_operation)
+{
+    echo "<b>" . _("Illustration") . ":</b> $image<br>";
+    $image_link = "<a href='$projects_url/$projectid/" . rawurlencode($image) . "'>" . _("Illustration") . "</a>";
+    echo "<b>" . _("View") . ":</b> $image_link";
+}
 echo "</p>";
 
 echo "<h2>$operation_image_str</h2>\n";
@@ -50,22 +67,25 @@ echo "<h2>$operation_image_str</h2>\n";
 if (!$project->can_be_managed_by_current_user)
 {
     echo "<p>", _('You are not authorized to manage this project.'), "</p>\n";
+    provide_escape_links();
     exit;
 }
 
 if ($operation == 'delete' && $project->state != PROJ_NEW)
 {
     echo "<p>", _('You can only delete illustrations for a project in the new state.'), "</p>\n";
+    provide_escape_links();
     exit;
 }
 
 $nonpage_image_names = $project->get_illustrations();
-if (!in_array($image, $nonpage_image_names) || !is_file("$projects_dir/$projectid/$image"))
+if (!$is_delete_all_operation && (!in_array($image, $nonpage_image_names) || !is_file("$projects_dir/$projectid/$image")))
 {
     // This too should only happen if someone is URL-tweaking.
     // (Or conceivably, the file was recently deleted, and the user
     // has an image_index that was generated before the deletion.)
     echo "<p>", _('There is no such illustration in the project.'), "</p>\n";
+    provide_escape_links();
     exit;
 }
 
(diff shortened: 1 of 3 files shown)


chrismiceli:
> 
> 
> As per a couple of comments in the task, the ability to delete illustration images should be restricted to the New Project state, or possibly that and P1 Unavailable (Those are the only states in which a PM can load images, themselves). I don't see anything that implies there is such a restriction (though, obviously, I may be missing it). Images can be _replaced_ at any time. You might also consider adding the ability for a PM "delete all" illustration images (same restrictions) -- deleting them one image at a time is fine if there are only 3 or 4, but if there are many more than that, the PM is probably better off asking db-req to delete them in bulk.

Great catch. I've added this check. I also found out that if a user deep linked to the replace image page, they would get a message saying you weren't allowed, but you could still replace the image. I resolved that as well.

chrismiceli:
any comments @cpeel?

cpeel:
Sorry, this is still in my TODO list. I promise to get to it shortly.

srjfoo:
I admit I'd like to test it out -- any chance of getting a sandbox version to play with?

chrismiceli:
@srjfoo, I forgot I can share with my sandbox. I'll set something up tonight. And no rush @cpeel, just making sure you saw it since I can't add reviewers. You do more than enough!

tools/project_manager/update_illos.php
```diff
@@ -0,0 +1,210 @@
...
+    {
+        return _('The uploaded file cannot be moved into the project directory for some reason.');
+    }
+}
+
+function handle_delete()
+// If there's a problem, return a string containing an error message.
+// If no problem, return the empty string.
+{
+    global $projectid, $image;
```
cpeel: We should pass these two global variables into the function instead of having them be globals.

tools/project_manager/update_illos.php
```diff
@@ -0,0 +1,210 @@
...
+if (!in_array($image, $nonpage_image_names) || !is_file("$projects_dir/$projectid/$image"))
+{
+    // This too should only happen if someone is URL-tweaking.
+    // (Or conceivably, the file was recently deleted, and the user
+    // has an image_index that was generated before the deletion.)
+    echo "<p>", _('There is no such illustration in the project.'), "</p>\n";
+    return;
+}
+
+$err_msg;
```
cpeel: What are these two lines here for? They aren't being set to anything.

chrismiceli:
> 
> 
> I admit I'd like to test it out -- any chance of getting a sandbox version to play with?

Sandbox available here with sample project created with 1 illustration.
https://www.pgdp.org/~chrismiceli/c/project.php?id=projectID5e6af75f2d779&detail_level=4

chrismiceli:
> 
> 
> If we are changing this file to be just for illustrations, we should go through all of the user-visible strings and use 'illustration' instead of 'image'.
> 
> Because of the rename, github isn't showing just your changes but the entire file so I apologize if some of my comments are on whatever code you got stuck with originally.

I also resolve this comment I believe. Let me know if I missed any references.

chrismiceli:
> 
> 
> > If we are changing this file to be just for illustrations, we should go through all of the user-visible strings and use 'illustration' instead of 'image'.
> > Because of the rename, github isn't showing just your changes but the entire file so I apologize if some of my comments are on whatever code you got stuck with originally.
> 
> I also resolve this comment I believe. Let me know if I missed any references.

Actually I still see some myself. I'll resolve those tonight. Sorry for the delay.

chrismiceli:
> > > If we are changing this file to be just for illustrations, we should go through all of the user-visible strings and use 'illustration' instead of 'image'.
> > > Because of the rename, github isn't showing just your changes but the entire file so I apologize if some of my comments are on whatever code you got stuck with originally.
> > 
> > 
> > I also resolve this comment I believe. Let me know if I missed any references.
> 
> Actually I still see some myself. I'll resolve those tonight. Sorry for the delay.

@cpeel, updating all user facing strings on the illustration page. I believe all comments have been addressed.

tools/project_manager/update_illos.php
```diff
@@ -0,0 +1,206 @@
...
+        return '';
+    }
+    else
+    {
+        return _('The uploaded file cannot be deleted from the project directory for some reason.');
+    }
+}
+
+// vim: sw=4 ts=4 expandtab
+?>
```
cpeel: Let's remove this at the end too, per our best practices.

tools/project_manager/update_illos.php
```diff
@@ -0,0 +1,206 @@
...
+        echo _('Please try again.'), "\n";
+        echo "</p>\n";
+        // Fall through to try again.
+    }
+}
+
+if ($operation == 'replace') {
+    echo "<p>", _('Select a replacement illustration to upload:'), "</p>\n";
+    echo "
+        <form enctype='multipart/form-data' action='update_illos.php' method='post'>
```
cpeel: For HTML5, which we are using, you can drop the `action` attribute: `<form enctype='multipart/form-data' method='POST'>`.

tools/project_manager/update_illos.php
```diff
@@ -0,0 +1,206 @@
...
+        <input type='hidden' name='operation' value='$operation'>
+        <input type='file' name='replacement_image' size='50'>
+        <br>
+        <input type='submit' value='", attr_safe(_("Upload Illustration")), "'>
+        </form>
+    ";
+} else {
+    echo "<p>", _('Are you sure you want to delete this illustration?'), "</p>\n";
+    echo "
+        <form enctype='multipart/form-data' action='update_illos.php' method='post'>
```
cpeel: Since we're not uploading a file here, this can just be `<form method='POST'>`

tools/project_manager/update_illos.php
```diff
@@ -0,0 +1,206 @@
...
+
+output_header("$operation_image_str: {$project->nameofwork}");
+
+echo "<h1>" . html_safe($project->nameofwork) . "</h1>\n";
+echo "<h2>$operation_image_str: $image</h2>\n";
+
+if (!$project->can_be_managed_by_current_user)
+{
+    echo "<p>", _('You are not authorized to manage this project.'), "</p>\n";
+    return;
```
cpeel: For consistency with other scripts, lets use `exit;` here instead of `return;`. Ditto in the next check too.

tools/project_manager/update_illos.php
```diff
@@ -0,0 +1,206 @@
...
+}
+
+if ($operation == 'delete' && $project->state != PROJ_NEW)
+{
+    echo "<p>", _('You can only delete illustrations for a project in the new state.'), "</p>\n";
+    return;
+}
+
+$page_image_names = array();
```
cpeel: Consider moving lines 55-71 into their own function (`get_project_illustrations()`?). This not only helps bundle all of the functionality together but it helps document what this code is doing as well.

Bonus: put this function in the `Project` class as `get_illustrations()`.

tools/project_manager/update_illos.php
```diff
@@ -0,0 +1,206 @@
...
+$operation = get_enumerated_param($_REQUEST, 'operation', 'replace', array('replace', 'delete'));
+
+$operation_image_str;
+if ($operation == 'replace') {
+    $operation_image_str = _('Replace Illustration');
+} else {
+    $operation_image_str = _('Delete Illustration');
+}
+
+output_header("$operation_image_str: {$project->nameofwork}");
```
cpeel: Let's make this page mirror handle_bad_page a bit more:

* Make the HTML header and `<h1>` title `_("Update Illustration")`.
* Add a small block below this that give some project info, including the illustration being updated (ala https://github.com/DistributedProofreaders/dproofreaders/blob/master/tools/project_manager/handle_bad_page.php#L81)
* The `<h2>` can then be `$operation_image_str`

chrismiceli: Here is what it looks like now:
image
image

chrismiceli:
Sandbox is updated as well: https://www.pgdp.org/~chrismiceli/c/tools/project_manager/update_illos.php?projectid=projectID5e6af75f2d779&image=pic.jpg&operation=replace

srjfoo:
Two main things.
1: I don't see an option for deleting all illustrations -- am I just missing something?

2: Should there be a "cancel" option (a.k.a. "Oops -- I don't want to delete that illustration, after all, I wanted to delete the one next to it.")?

tools/project_manager/update_illos.php
```diff
@@ -0,0 +1,196 @@
...
+$project = new Project($projectid);
+$operation = get_enumerated_param($_REQUEST, 'operation', 'replace', array('replace', 'delete'));
+
+if ($operation == 'replace') {
+    $operation_image_str = _('Replace Illustration');
+} else {
+    $operation_image_str = _('Delete Illustration');
+}
+
+output_header("$operation_image_str: {$project->nameofwork}");
```
cpeel: Let's use `_("Update Illulstration")` here too, please.

chrismiceli:
ok, I added a delete all operation with a button next to download illustrations:
image

chrismiceli:
> 
> 
> Testing shows that the bulk delete can be done in any round. It should be restricted to New Project or P1 Unavailable, just as the individual deletions are. When I deleted the images from a test project in P2 Available, I got the following message:
> 
> > Notice: Undefined variable: has_error in /home/chrismiceli/public_html/dproofreaders/tools/project_manager/update_illos.php on line 257
> 
> Edit: Technically, individual deletions are restricted to New Project. It is not unreasonable to allow the deletions to happen in P1 Unavailable, also, as that would match PMs' ability to load new images, themselves (i.e. if they can load new illustrations when a project is in P1 Unavailable, it's reasonable for them to be able to delete them then, also).

Sandbox updated. I made delete all match the individual delete which is currently only the new state. I don't think I fully understand the states that should allow it based on the code states, perhaps @cpeel could tell me which code states to allow and the best way to select those states.

srjfoo:
> Sandbox updated. I made delete all match the individual delete which is currently only the new state. I don't think I fully understand the states that should allow it based on the code states, perhaps @cpeel could tell me which code states to allow and the best way to select those states.

I think the database states that are relevant are `project_new` and `P1.proj_unavail`. Not sure about the best way to select the states. 

I get

> Notice: Undefined variable: is_delete_all_operation in /home/chrismiceli/public_html/dproofreaders/tools/project_manager/update_illos.php on line 57

and

>Notice: Undefined variable: is_delete_all_operation in /home/chrismiceli/public_html/dproofreaders/tools/project_manager/update_illos.php on line 82**

when deleting a single image. After deleting the image, I get

> Notice: Undefined variable: is_delete_all_operation in /home/chrismiceli/public_html/dproofreaders/tools/project_manager/update_illos.php on line 99.

I am still getting a

> Notice: Undefined variable: has_error in /home/chrismiceli/public_html/dproofreaders/tools/project_manager/update_illos.php on line 257

when deleting all images.

image

-----

For displaying the image, instead of 

image
could the filename be the link?

chrismiceli:
> P1.proj_unavail

Ok, sorry for the issues. I believe I have resolved them all. I am new to PHP so I made some pretty dumb mistakes regarding scope inside conditionals.

cpeel:
This looks great to me. Let's give a few days to get any community input and then we can get this merged in.

cpeel:
@chrismiceli - would you mind rebasing this off the most recent master then I'll get it merged in. I can do the rebasing if you'd rather.

chrismiceli:
> 
> 
> @chrismiceli - would you mind rebasing this off the most recent master then I'll get it merged in. I can do the rebasing if you'd rather.

Done here:
https://github.com/DistributedProofreaders/dproofreaders/pull/261
Sorry, I didn't realize I couldn't reopen this PR after deleting the remote branch. I am used to atlassian's bitbucket where I can reopen PRs after rebasing. It seems github does support this but I have to do a force push. I wasn't aware. Sorry for the confusion.

PR #248, opened 2020-03-12
Remove the WordCheck status from formatting rounds

cpeel:
Remove the WordCheck status from formatting rounds. Also make the X less bold.

[Task 1855](https://www.pgdp.net/c/tasks.php?action=show&task_id=1855)

Available in [no-wc-indicators-in-formatting](https://www.pgdp.org/~cpeel/c.branch/no-wc-indicators-in-formatting) sandbox.

### DIFF
--- a/project.php
+++ b/project.php
@@ -797,14 +797,27 @@ function recentlyproofed( $wlist )
 
     $recentNum=5; // if this is > 5 more rows will be shown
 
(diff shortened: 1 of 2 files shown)


cpeel:
After much discussion, we are going to go with the current implementation while realizing there may need to be be further refinements later.

PR #249, opened 2020-03-13
Allow for combined characters in project_quick_check and Add files

70ray:
This is a re-make of a similar branch but using the new unicode functions.
It shows multiple codepoints as well as bytes in the Bad Bytes tables in PQC.
Bad sequences are handled separately from invalid codepoints; this simplifies distinguishing the nature of the badness.
For Add Page it shows complete combined characters if present and title names for the bad characters.
BOM handling is also improved.

### DIFF
--- a/pinc/bad_bytes.inc
+++ b/pinc/bad_bytes.inc
@@ -786,24 +786,20 @@ define_bad_byte_sequences();
 
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 
-function find_bad_byte_sequences_in_text($text, $projectid)
+define("BOM", "\u{feff}");
+
+// note that this can change $text
+function find_bad_byte_sequences_in_text(&$text)
 {
     global $_unanchored_bad_bytes_regex, $_bad_byte_sequences;
 
     $occurrences = array();
 
-    // find any characters in the text that are not valid for the project
     if(utf8_string_has_bom($text))
     {
-        $occurrences['BOM'] = 1;
+        $occurrences[BOM] = 1;
+        //remove it so not reported again as an invalid character
         $text = utf8_str_replace(utf8_bom(), '', $text);
-    }
-
-    $project = new Project($projectid);
-    $invalid_chars = get_invalid_characters($text, $project->get_valid_codepoints());
-    foreach($invalid_chars as $char => $num)
-    {
-        $occurrences[$char] = $num;
     }
 
     $n_matches = preg_match_all($_unanchored_bad_bytes_regex, $text, $matches);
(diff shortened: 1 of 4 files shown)


70ray:
Add files after:
image

70ray:
Now running in sandbox [https://www.pgdp.org/~rp31/preview](url)

pinc/bad_bytes.inc
```diff
@@ -898,30 +898,22 @@ function get_remarks_for_bad_byte_sequence($raw)
...
-        $why_bad = "Codepoint not valid for project";
-        $intended_character = "";
-    }
     else
     {
         list($code_point, $why_bad) = $_bad_byte_sequences[$raw];
     }
 
-    if(!isset($intended_character))
+    $info = $_character_data[$code_point] ?? null;
```
cpeel: What you have is fine, but you might just use a ternary:
```
$char_name = isset($_character_data[$code_point]["name"]) ? $_character_data[$code_point]["name"] : '???';
```

pinc/project_quick_check.inc
```diff
@@ -446,6 +472,17 @@ function tds_for_bad_bytes($raw)
     return $tds;
 }
 
+function tds_for_invalid_chars($raw)
+{
+    $tds = "<td>$raw</td>";
+    $tds .= "<td>" . string_to_hex($raw) . "</td>"; // bytes
+    $tds .= "<td>" . string_to_codepoints_string($raw) . "</td>"; // codepoints
+    $tds .= "<td></td>"; // intended character
+    $tds .= "<td>" . "Codepoint not valid for project" . "</td>";
```
cpeel: Wrap the string in `_()`.

pinc/bad_bytes.inc
70ray: Then there could be an even shorter way:

pinc/project_quick_check.inc
70ray: Done this.

pinc/bad_bytes.inc
cpeel: I had to go look up the `??` operator -- very nice! Thanks for teaching me some thing new.

srjfoo:
> This LGTM and worked in my testing. Would appreciate @srjfoo's thoughts and testing as well.

After poking around to figure out how to do combining diacritics on a Mac, and especialy multiple combining diacritics on one character -- it looks good to me, too. Screenshot:
combining_diacritics

Caveat -- you both know a whole lot more about combining diacritics than I do, so if there's something specific you want me to poke at and test, let me know. I'm just beginning to get a handle on it.

I started out by reloading various of my earlier test sets (which are simple compared to the combining diacritic issue) and everything continues to work well as far as I can see. After that, I created the new test files from which the above screenshot is taken.

70ray:
Now running in sandbox

[https://www.pgdp.org/~rp31/preview](url)

PR #250, opened 2020-03-14
Add checks to ensure project tables are UTF-8

cpeel:
Prevent changes to project tables that have not yet been converted to
UTF-8. This is mostly for the rare circumstances where projects created
before the conversion are unarchived. For these tables, provide an admin
tool to convert the table.

You can see how this works with the [Aspazio project](https://www.pgdp.org/~cpeel/c.branch/project-table-encoding-checks/project.php?id=projectID4fd62bf9ac045&detail_level=3) in my [project-table-encoding-checks](https://www.pgdp.org/~cpeel/c.branch/project-table-encoding-checks) sandbox.

### DIFF
--- a/tools/project_manager/handle_bad_page.php
+++ b/tools/project_manager/handle_bad_page.php
@@ -27,6 +27,14 @@ if(user_can_edit_project($projectid) != USER_CAN_EDIT_PROJECT)
     die("You are not authorized to manage this project.");
 }
 
+$project = new Project($projectid);
+
+// prevent changes to the project table if it isn't UTF-8
+if(!$project->is_utf8)
+{
+    die(_("Project table is not UTF-8."));
+}
+
 // If the user hit a cancel button, return them to the starting form
 if($cancel)
     $modify = '';
@@ -39,8 +47,6 @@ if (!$resolution) {
     $b_User = $page['b_user'];
     $b_Code = $page['b_code'];
 
-    $project = new Project($projectid);
-    
     $round = get_Round_for_page_state($state);
     $current_round_num = $round->round_number;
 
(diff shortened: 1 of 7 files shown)


tools/project_manager/handle_bad_page.php
```diff
@@ -27,6 +27,14 @@
     die("You are not authorized to manage this project.");
 }
 
+$project = new Project($projectid);
+
+// prevent changes to the project table if it isn't UTF-8
+if(!$project->is_utf8)
+{
+    die(_("Project table is not UTF-8."));
+}
```
srjfoo: I'm assuming that the way this message is provided, there's no way to do `class='error'`? Is the same true of including a back-link?

tools/site_admin/copy_pages.php
```diff
@@ -233,6 +233,14 @@ function do_stuff( $projectid_, $from_image_, $page_name_handling,
 
     foreach ( array( 'from', 'to' ) as $which )
     {
+        $projectid = $projectid_[$which];
+        $project = new Project($projectid);
+
+        if(!$project->is_utf8)
+        {
+            die("Project table $projectid is not UTF-8.");
```
srjfoo: I'm not terribly concerned about that siteadmin scripts be translatable, but this should probably have gettext() markup to be consistent? Ditto above question about back-link.

cpeel:
Because having a project table that isn't UTF-8 encoded should be a very rare edge case caused by Squirrel manipulations, I'm mostly viewing these as "emergency breaks" to prevent bad data getting into the table and, therefore, not really concerned about a lot of the nuances of back links et al. If you really feel these should be more user-friendly though I can see about updating them.

srjfoo:
Okay, good point. The likelihood of anyone other than a Squirrel seeing the warning is pretty small. I withdraw the suggestions. :) And we'll document for Squirrels.

PR #252, opened 2020-03-16
lastpass bug

Guy-Sowden:
fixed bug with nav-bar for lastpass users on firefox users where username and password input box would appear unstyled
image

here's how it appear after the patch
image


cpeel:
Thank you for the MR! Can you describe more the problem you are solving? Currently we don't do styling on most (all?) of our form inputs and I want to make sure we are thinking of how these changes will differ from other form inputs. Also, please revisit your commits and ensure that only the files you want changed are actually included as you have some extra changes (including reverting an existing commit).

cpeel:
Closing after not getting feedback or updates to the MR.

PR #253, opened 2020-03-16
For smooth reading upload singly-zipped files

70ray:
construct zipped html and images folder after upload

### DIFF
--- a/tools/upload_text.php
+++ b/tools/upload_text.php
@@ -335,12 +335,33 @@ function process_file($project, $indicator, $stage, $returning_to_pool)
                 throw new FileUploadException("failed to extract files");
             }
             // extract any zips in smooth_dir
+            // there should not now be any but keep for compatibility for now
             $zips = glob("$smooth_dir/*.zip");
             foreach($zips as $zip)
             {
                 if(!extract_zip_to($zip, $smooth_dir))
                 {
                     throw new FileUploadException("failed to extract files");
+                }
+            }
+            // if there is an htm or html file, zip it with images directory
+            $htm_files = glob("$smooth_dir/*.{htm,html}", GLOB_BRACE);
+            if($htm_files)
+            {
+                $htm_file = $htm_files[0];
+                $files_to_zip[] = $htm_file;
+                // make the zip file with same name but zip extension.
+                // we don't know if it is htm or html so can't use basename()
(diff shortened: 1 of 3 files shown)


tools/upload_text.php
```diff
@@ -343,6 +344,26 @@ function process_file($project, $indicator, $stage, $returning_to_pool)
                     throw new FileUploadException("failed to extract files");
                 }
             }
+            // if there is an htm or html file, zip it with images directory
+            $htm_files = glob("$smooth_dir/*.{htm,html}", GLOB_BRACE);
+            if($htm_files)
+            {
+                $htm_file = $htm_files[0];
```
cpeel: What happens if there is more than one file? This would take the first one which could be a random one. Should we put all of them in? Throw an error if there is more than one?

srjfoo: I would say that if there is more than one HTML file, zip them all into the `HTML+images` zip for download. The problem is that PPers still haven't gotten the hang of zipping the HTML and images together for inclusion into the master SR upload zip. For PPV, they don't have to, and for PG they don't have to. This would bring the SR upload more in line with what PPV and PG accept.

It's not going to be perfect, whatever is done, but at least this will align PPers' expectations more with what they're used to.

70ray: If they had names like "book.html" and "book_no_images.html" then the smooth reader would know what they were for when the file was unzipped. But what would we name the zip file? Perhaps <projectID>.zip.

srjfoo: Hm. Good point. Do we keep a copy of what they named the zip to begin with? If we do, we could use that, with `-html` appended to the filename, preceding the extension. That would be my first choice. IIf that's too annoying to code, then `projectID-html.zip` would be my choice.

70ray: We do know the original name. I'll try what you suggest.

70ray: This change has now been made.

cpeel:
This LGTM! @srjfoo?

srjfoo:
> This LGTM! @srjfoo?

Me too!

cpeel:
@srjfoo - have you had a chance to test out Ray's latest change?

srjfoo:
> @srjfoo - have you had a chance to test out Ray's latest change?

Apologies for getting distracted. I've tested it now, and it looks good!

PR #254, opened 2020-03-17
1777: Show timestamp for uploaded SR reports

chrismiceli:
add timestamp next to SR report for download

### DIFF
--- a/project.php
+++ b/project.php
@@ -1763,6 +1763,7 @@ function echo_download_zip( $link_text, $discriminator )
 
         $url = "$project->url/$p";
         $filesize_b = filesize("$project->dir/$p");
+        $last_modified = filemtime("$project->dir/$p");
     }
 
     echo "<li>";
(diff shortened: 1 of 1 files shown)


chrismiceli:
I expect some feedback on this, but thought I'd start the conversation with a PR. Here is what this looks like:
image
@cpeel @srjfoo

project.php
```diff
@@ -1763,17 +1763,29 @@ function echo_download_zip( $link_text, $discriminator )
...
     echo "\n";
 }
 
+function echo_last_modified($last_modified)
+{
+    global $datetime_format;
+
+    if (isset($last_modified))
+    {
+        echo " (", strftime($datetime_format, filemtime($file)), ")";
```
cpeel: `s/filemtime($file)/$last_modified/`

chrismiceli: This is what happens when I do a small refactor after testing but before pushing like a fool.

cpeel: I assumed as much - it happens, don't worry about it! 😁

srjfoo: Join the club. 😁

PR #256, opened 2020-03-22
Allow sites to define default character suites

cpeel:
Instead of defaulting to basic-latin, allow sites to specify their own default character suites for new projects.

This seemed like a small but useful change for other sites. `SETUP/UNICODE.md` doesn't exist yet but will ... soon.


chrismiceli:
FYI Travis says there is an issue with the site vars file.

cpeel:
@chrismiceli, thanks for setting this up to be a very-simple change. I appreciate that!

chrismiceli:
> 
> 
> @chrismiceli, thanks for setting this up to be a very-simple change. I appreciate that!

Glad to be of any use.

PR #257, opened 2020-03-22
Unicode docs

cpeel:
Here is at least a start at some code-level documentation about the new unicode changes.

What else should we include in here?

### DIFF
--- a/SETUP/UNICODE.md
+++ b/SETUP/UNICODE.md
@@ -1,25 +1,34 @@
 # Unicode support
 
-The DP codebase fully supports UTF-8, both in the HTML rendered to the user
-as well as in page texts.
+The DP codebase fully supports UTF-8, in the HTML rendered to the user as well
+as in page texts.
 
 Rather than open up the floodgates and allow proofreaders to input any of
 the millions of Unicode character into the proofreading interface, we allow
-project managers to restrict what Unicode characters are valid for a project.
-This ensures that proofreaders don't insert characters that don't rightfully
-belong in the project but look similar to those that do (Kappa for a 'K'
-as an example).
+project managers to restrict what Unicode characters are valid for project
+pages. This ensures that proofreaders don't insert characters that don't
+rightfully belong in the project but look similar to those that do (Kappa
+for a 'K' as an example).
 
 Both the UI and the server-side normalization enforce that only supported
-characters are used. This allows the code to support the full set of Unicode
-characters but reduces the possibility of invalid characters in a given project.
+characters are used for project pages. This allows the code to support the full
+set of Unicode characters but reduces the possibility of invalid characters in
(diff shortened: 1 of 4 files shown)


chrismiceli:
> 
> 
> Here is at least a start at some code-level documentation about the new unicode changes.
> 
> What else should we include in here?

The only thing I could think of is some of the extraneous stuff, such as database, OS, or apache type stuff. Not sure where that is appropriate however.

cpeel:
Added a bit more data on the upgrade aspect of it.

SETUP/UNICODE.md
```diff
@@ -0,0 +1,64 @@
...
+are created. Disabled character suites can still be used by projects that are
+using them, but cannot be added to new projects.
+
+If you want all new projects to use one or more character suites by default,
+set `_DEFAULT_CHAR_SUITES` in your `configuration.sh` file.
+
+## Upgrading from Latin-1 projects
+
+`upgrade/14` includes scripts that will convert a prior Latin-1-based
+version of the DP database to support UTF-8.
```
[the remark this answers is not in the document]
cpeel: Good idea - added.

SETUP/UNICODE.md
```diff
@@ -0,0 +1,102 @@
+# Unicode support
+
+The DP codebase fully supports UTF-8, both in the HTML rendered to the user
+as well as in page texts.
```
srjfoo: Grammar police comment: you don't need "both" and "as well as" -- I recommend either `, ... as well as ...` or `both ... and ...`.

SETUP/UNICODE.md
```diff
@@ -0,0 +1,102 @@
...
+Rather than open up the floodgates and allow proofreaders to input any of
+the millions of Unicode character into the proofreading interface, we allow
+project managers to restrict what Unicode characters are valid for a project.
+This ensures that proofreaders don't insert characters that don't rightfully
+belong in the project but look similar to those that do (Kappa for a 'K'
+as an example).
+
+Both the UI and the server-side normalization enforce that only supported
+characters are used. This allows the code to support the full set of Unicode
+characters but reduces the possibility of invalid characters in a given project.
```
srjfoo: Should there be some mention here of the fact that this enforcement applies only to the text in the projects, and that project title and author are not held to the same restrictions? (Even once we force normalization on title and author, we're not going to force them to be held to the same charsuites that the project text is, are we?)

SETUP/UNICODE.md
cpeel: Fixed.

SETUP/UNICODE.md
cpeel: Good idea. I've added more details as well as info around the phpBB limitations.

PR #259, opened 2020-03-23
1849: Changing wait time before non-newcomers can proof new French P1…

chrismiceli:
… projects

change wait time to 3 days

### DIFF
--- a/pinc/stages.inc
+++ b/pinc/stages.inc
@@ -508,8 +508,8 @@ function get_reserve_length($project, $round)
         // so are Children's Book Week projects (in 2017)
         if ($project->special_code == 'ChBookWk') return 0;
 
-        // Otherwise, reserve for 7 days.
-        return 7;
+        // Otherwise, reserve for 3 days.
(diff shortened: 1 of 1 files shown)


pinc/stages.inc
```diff
@@ -508,8 +508,8 @@ function get_reserve_length($project, $round)
         // so are Children's Book Week projects (in 2017)
         if ($project->special_code == 'ChBookWk') return 0;
 
-        // Otherwise, reserve for 7 days.
-        return 7;
+        // Otherwise, reserve for 3 days.
+        return 3;
```
srjfoo: This looks good, but lvl's comment in the task actually recommended making the French exception the same as the English That is, replace the two current special day exceptions (for anniversary and children's book week) with just:
```
// Same treatment as the English
// All 'special day' projects are exempted.
if ($project->special_code != '') return 0;
```

chrismiceli: Yup, you're right!

PR #260, opened 2020-03-23
Ensure proper gettext translations

cpeel:
This is a small collection of commits that came out of some gettext troubleshooting from another site that is trying to use our code. The last commit is the most important one, the others are small things found along the way.


cpeel:
> Not knowing anything about this I looked at the documentation but it seems very confusing.

Yes - it is very very (very) confusing. Like black magic, except darker.

PR #262, opened 2020-03-25
1769: Add a text box for the "Other" choice on registration page

chrismiceli:
Add column to users for other explanation
add column to non_activated_users for other explanation
add javascript to show new field when other is selected
update sql to handle new columns


chrismiceli:
Sandbox is updated here https://www.pgdp.org/~chrismiceli/c/accounts/addproofer.php

accounts/addproofer.php
```diff
@@ -105,7 +107,7 @@
 // to this file & run the above commands.
 
 $header = _("Create An Account");
-output_header($header);
+output_header($header, true, array("js_files" => array("$code_url/accounts/addproofer.js")));
```
cpeel: Please use the constant  `SHOW_STATSBAR` (which is just `TRUE`) for the second argument as it's a little clearer on what that is.

accounts/addproofer.php
```diff
@@ -263,7 +270,7 @@ function _validate_fields($real_name, $username, $userpass, $userpass2, $email,
     // forum software which, if used, will cause account creation to fail in
     // activate.php.
 
-    if(!$referrer)
+    if(!$referrer || ($referrer == 'other' && !$referrer_explanation))
```
cpeel: Checking that `$referrer_explanation` isn't empty makes sense but we should run it past Linda. It's one thing to give the user an option to fill out `other` and another to require them to do so before continuing. Consider pinging her in the #pgdpdev channel.

cpeel:
Don't forget to update the tests -- that's why TravisCI is failing :)

srjfoo:
> I haven't yet tested it myself but the code looks _really good_, thank you. I'm on the fence regarding the `referrer_explanation` label. My gut wants to rename that to `referrer_other` but I'm not sure that's any clearer. I mention this to have a discussion about it, not a request to change it.

`referrer_details`, maybe?

I've tested it, and it works well.

accounts/addproofer.php
chrismiceli: switched it to referrer_details per @srjfoo.

accounts/addproofer.js
```diff
@@ -8,7 +8,7 @@ function AddProofer() {
     "use strict";
 
     that.referrerOption = $('select[name=referrer]');
-    that.referrerExplanation = $('#referrer_explanation');
+    that.referrerExplanation = $('#referrer_details');
```
cpeel: Can we also rename this `referrerDetails` to match?

accounts/addproofer.js
```diff
@@ -8,7 +8,7 @@ function AddProofer() {
...
     that.setOtherExplanation();
```
cpeel: And this to `setOtherDetails`.

SETUP/upgrade/14/20200324_alter_referrer_explanation.php
```diff
@@ -0,0 +1,37 @@
+<?php
```
cpeel: And rename this file to `20200324_alter_referrer_details.php`.

accounts/addproofer.php
```diff
@@ -184,6 +186,11 @@
 echo "  </td>";
 echo "</tr>\n";
 
+echo "<tr id='referrer_details'>";
+echo "  <th>" . _("Other Explanation") . ":</th>";
```
srjfoo: The more I look at it, the more awkward "Other Explanation" sounds to me, but I don't think "Other Details" sounds any better. Can we make the label just "Other" and, in the text field add a placeholder attribute with something like "Please provide a short description"? Or possibly do away with the label altogether, and make that whole row into text input with placeholder text describing what's needed?

chrismiceli: I think Other details makes sense. I don't like putting instruction or some sort of label in the placeholder for various reasons: https://www.nngroup.com/articles/form-design-placeholders/.

srjfoo: Interesting article, and I can definitely agree with their points, though I guess I hadn't given placeholder text a lot of thought, because I've actually appreciated it when I've seen it. And we do have it in other places in the code (at least some of it at my instigation).

Part of my concern is just wondering if we need to give more of a hint than "Other Details". Should there be a comma? -- it seems to me that there's a difference between "Other Details" and "Other, Details". And yes, I'm splitting hairs.

cpeel: I like "Other Details" with the comma, personally.

chrismiceli: Updated. Thanks for the advice. I am so bad at these kinds of decisions.

srjfoo: It's a group effort -- we all need proofreaders!

accounts/addproofer.js
```diff
@@ -0,0 +1,29 @@
...
+    "use strict";
+
+    that.referrerOption = $('select[name=referrer]');
+    that.referrerDetails = $('#referrer_details');
+    that.setOtherDetails();
+    that.referrerOption.change(that.setOtherDetails);
+  };
+
+  this.setOtherDetails = function() {
+    "use string";
```
cpeel: Should this be `"use strict";` instead?

Sorry, just noticed this.

accounts/addproofer.js
```diff
@@ -0,0 +1,29 @@
+/*global $*/
```
cpeel: Can you run this through eslint too? IIRC we are using 4-space indents instead of 2- and that would validate this.

cpeel: We have an outstanding task for getting that into our CI/CD too.

accounts/addproofer.js
chrismiceli: /facepalm.

PR #264, opened 2020-03-27
Add a sitemap; also add metadata to project pages

cpeel:
Add a sitemap we can submit to search engines so they can reasonably index our site. Note that most of our pages require a login, so mostly these are project pages.

On the project pages, include LD+JSON metadata so search engines can glean a bit more structured data about the page.

### DIFF
--- a/SETUP/apache2.conf.example
+++ b/SETUP/apache2.conf.example
@@ -29,6 +29,12 @@ DirectoryIndex index.php default.php index.html default.html
 
     ErrorLog ${APACHE_LOG_DIR}/error.log
 
+    <IfModule mod_rewrite.c>
+        # make the sitemap work with an .xml extension
+        RewriteEngine On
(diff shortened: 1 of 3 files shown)


chrismiceli:
Can you share the output of the sitemap.php and the header? I'm curious what it outputs. Also, will we need a robots.txt?

srjfoo:
> Can you share the output of the sitemap.php and the header? I'm curious what it outputs. Also, will we need a robots.txt?

I'd like to see this, too. It sounds like a good idea, but most of the code is over my head.

I thought we already had robots.txt.

cpeel:
You can see the output at https://www.pgdp.org/~cpeel/c.branch/sitemaps/sitemap.php for TEST.

[PROD has a robots.txt](https://www.pgdp.net/robots.txt) that (should) allow well-behaved crawlers to access files in the code root, which includes `project.php`.

PR #265, opened 2020-03-28
1856: Add eslint to TravisCI

chrismiceli:
add npm init for eslint
turn off rules causing failures
fix autofixable issues

### DIFF
--- a/tools/proofers/dp_proof.js
+++ b/tools/proofers/dp_proof.js
@@ -38,11 +38,8 @@ function setText() {
 }
 
 function fixText() {
-    //if (document.all)
-    //{
     docRef.editform.text_data.style.width = ieW;
     docRef.editform.text_data.style.height = ieH;
-    //}
 }
 
 function changeFontFamily(font_index, font, fallback) {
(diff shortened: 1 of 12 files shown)


chrismiceli:
https://github.com/DistributedProofreaders/dproofreaders/pull/265/checks?check_run_id=541124478 You can see the node and php 7.0 passed. It seems 7.1 and 7.2 are missing though. I'll have to debug that.

tools/proofers/dp_proof.js
```diff
@@ -1,55 +1,51 @@
...
+        ieW = docRef.editform.text_data.style.width;
+        ieH = docRef.editform.text_data.style.height;
+        ieSt = 1;
     }
 }
 
-function fixText()
-{
+function fixText() {
     //if (document.all)
```
cpeel: Let's just remove these comments -- not sure why they are even there.

tools/proofers/dp_proof.js
```diff
@@ -72,103 +67,88 @@ function changeFontSize(font_size_index, font_size)
...
 }
 
-function showNW_safe(str)
-// Entity-encode str's HTML-special characters,
-// but reinstate <i> and <b> and <hr> tags,
-// because we want the browser to interpret them (but nothing else) as markup.
-// Also convert <sc></sc> to <spans> that do the right thing.
-{
+function showNW_safe(str) {
+    // Entity-encode str's HTML-special characters,
```
cpeel: Let's move these above the function definition so they remain aligned left.

tools/proofers/dp_proof.js
```diff
@@ -72,103 +67,88 @@ function changeFontSize(font_size_index, font_size)
...
-// in an HTML document.
-// That is, make the following replacements:
-//    &  ->  &amp;
-//    <  ->  &lt;
-//    >  ->  &gt;
-// This should be equivalent to PHP's
-//     htmlspecialchars($str,ENT_NOQUOTES)
-{
+function html_safe(str) {
+    // Return a version of str that is safe to send as element-content
```
cpeel: Ditto place above function decl.

tools/proofers/dp_proof.js
```diff
@@ -72,103 +67,88 @@ function changeFontSize(font_size_index, font_size)
     fixText();
 }
 
-function showAllText()
```
cpeel: Did we drop this function intentionally?

chrismiceli: it was not referenced anywhere and eslint was complaining. It looks like it was added for debugging to be honest.

cpeel:
Nothing the CI/CD changes will get into source control, so it doesn't matter if the CI/CD auto-changes the files -- they'll just result in a linting error which is the goal.

I'm hoping we can create a `SETUP/format_js.sh` file (or something similarly named that can later have a PHP equiv) that will run eslint against the JS files and make local updates for users to run locally before they commit files.

cpeel:
> https://github.com/DistributedProofreaders/dproofreaders/pull/265/checks?check_run_id=541124478 You can see the node and php 7.0 passed. It seems 7.1 and 7.2 are missing though. I'll have to debug that.

Interesting - I guess matrix builds only run against a single version? If it comes down to the tradeoff of duplicating the whole PHP segment three times in the `.travis.yml` or running against a single version, let's just run it against a single version (the one we have installed on TEST and PROD which is 7.0). To my knowledge we've never had a CI/CD build that worked for one version and failed on another.

chrismiceli:
exactly. Travis runs the lint script which will issue warnings and errors only. the lint-fix script is just for developer convenience to fix their issues.

cpeel:
> the lint-fix script is just for developer convenience to fix their issues.

Err.. did you intend to commit a linting file in `SETUP/` and just missed the `git add`?

chrismiceli:
> > the lint-fix script is just for developer convenience to fix their issues.
> 
> Err.. did you intend to commit a linting file in `SETUP/` and just missed the `git add`?

No, the script is in the package.json file. There are 2 scripts, 1 travis runs which only reports errors and another which is for devleopers to autofix any issues that eslint knows how to fix, such as spacing.  You can run it by "npm run lint-fix".

cpeel:
Hahahaha. This node n00b totally missed that.

/me wonders how to best document that for the next person...

cpeel:
Ok, I think I just need to create some dev-centric styling docs in `SETUP/` -- I'll take care of that separately.

PR #268, opened 2020-03-31
Revise mentors/view_page_text_image etc.

70ray:
mentors/view_page_text_image.php now uses and persists settings in local storage.
project_manager/displayimage.php can switch to displaying text or image + text and can select a project.


70ray:
This version loads image and text without reloading the whole page. Experiment shows there seems to be little advantage in doing this and because of the way the image is drawn, it appears slower in some circumstances. Making a version which reloads the whole page.

PR #269, opened 2020-03-31
Redirect from homepage to Activity Hub on login

cpeel:
When a user logs in from a page, we generally redirect them back to the
page they were on. Most new users will login from the home page and
get redirected back there -- which doesn't provide them with any good
links for where to go next. The Activity Hub, however, provides text
and links for new users to get them to the right place. The AH also
provides not-so-new users with useful News, so redirecting there from
the homepage for everyone is useful.


cpeel:
Let's start with this and get it rolled out and we can adjust the wording in a later MR.

PR #270, opened 2020-04-02
Copy all WordCheck data on project copy

cpeel:
Copy all data (files and events) when doing project copy/merges.

[Task 1862](https://www.pgdp.net/c/tasks.php?action=show&task_id=1862)

### DIFF
--- a/tools/site_admin/copy_pages.php
+++ b/tools/site_admin/copy_pages.php
@@ -41,7 +41,7 @@ $action = get_enumerated_param($_POST, 'action', 'showform', array('showform', '
 $page_name_handling = get_enumerated_param($_POST, 'page_name_handling', null, array('PRESERVE_PAGE_NAMES', 'RENUMBER_PAGES'), true);
 $transfer_notifications = get_integer_param($_POST, 'transfer_notifications', 0, 0, 1);
 $add_deletion_reason = get_integer_param($_POST, 'add_deletion_reason', 0, 0, 1);
-$merge_wordcheck_files = get_integer_param($_POST, 'merge_wordcheck_files', 0, 0, 1);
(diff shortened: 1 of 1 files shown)


tools/site_admin/copy_pages.php
```diff
@@ -519,13 +519,13 @@ function do_stuff( $projectid_, $from_image_, $page_name_handling,
...
-    if ($merge_wordcheck_files) 
+    if ($merge_wordcheck_data)
     {
-        echo _("The WordCheck files from the source project WILL be merged into the destination project");
+        echo _("The WordCheck data from the source project WILL be merged into the destination project");
     }
     else
     {
-        echo _("The WordCheck files WILL NOT be merged");
+        echo _("The WordCheck data WILL NOT be merged");
```
srjfoo, line 528: Above, for the button, you have the text: `Merge WordCheck files and events into destination project`. It might be good to use that same wording for thest two messages.

tools/site_admin/copy_pages.php
```diff
@@ -679,64 +679,15 @@ function do_stuff( $projectid_, $from_image_, $page_name_handling,
         }
     }
 
-    if ($merge_wordcheck_files) {
-        echo "<p>" . _("Merging wordcheck files.") . "</p>\n";
+    if ($merge_wordcheck_data) {
+        echo "<p>" . _("Merging WordCheck data.") . "</p>\n";
```
srjfoo, line 683: Probably a good idea to be consistent and use "files and events" here, too. That gets all the squirrel-facing text consistent.

tools/site_admin/copy_pages.php
cpeel: Good idea - fixed.

PR #271, opened 2020-04-03
Revise mentors/view_page_text_image etc.

70ray:
Add new display image and or text using functions
in page_controls.inc

### DIFF
--- a/scripts/text_control.js
+++ b/scripts/text_control.js
@@ -1,0 +1,53 @@
+/*global $ */
+// this operates the font face, font size and wrap controls
+$(function () {
+    var fontFaceSelector = document.getElementById("fntFace");
+    var fontSizeSelector = document.getElementById("fntSize");
+    var wrapCheck = document.getElementById("wrap");
+
+    function setFontFace() {
+        $("#text_data").css("font-family", fontFaceSelector.value);
+    }
+
+    var fontFaceIndex = localStorage.getItem("font_face_index");
+    if(fontFaceIndex) {
+        fontFaceSelector.selectedIndex = fontFaceIndex;
+    }
+    setFontFace();
+
+    $(fontFaceSelector).change(function () {
+        setFontFace();
+        localStorage.setItem("font_face_index", fontFaceSelector.selectedIndex);
+    });
+
+    function setFontSize() {
+        $("#text_data").css("font-size", fontSizeSelector.value);
+    }
+
+    var fontSizeIndex = localStorage.getItem("font_size_index");
+    if(fontSizeIndex) {
+        fontSizeSelector.selectedIndex = fontSizeIndex;
+    }
+    setFontSize();
+
+    $(fontSizeSelector).change(function () {
+        setFontSize();
+        localStorage.setItem("font_size_index", fontSizeSelector.selectedIndex);
+    });
+
+    function setWrap() {
+        $("#text_data").attr('wrap', wrapCheck.checked ? 'soft' : 'off');
+    }
+
+    var textWrap = JSON.parse(localStorage.getItem("text_wrap"));
+    if(!textWrap) {
+        textWrap = false;
+    }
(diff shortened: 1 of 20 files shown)


cpeel:
Functionally, I really like how this works from a UI perspective.

It's also a lot of code that combines a lot of JS with a lot of PHP. What happens if we think about this as a single JS application? It would be a collection of static HTML + CSS + a single JS file that is passed in a JSON blob with the necessary data (including translated strings) and the JS app would handle everything else. Intermixing the JS and PHP makes things hard to follow and hard to maintain.

70ray:
Interesting idea. I'll have a look at it.

70ray:
This is really five different php pages so it is not really so complicated as it first appears. The three linked pages: displayimage, display_text and display_image_text are separate pages. The mentors page is independent of them. The project_selector page is used by all of them. They all use functions from page_controls.inc. The javascript files do specific things e.g. image_size.js is used for pages with an image and image-sizing controls.

cpeel:
The interlaced PHP and HTML with JS is not the model that I'd like for us to move to for these JS-based applications. If we're going to build a JS app, let's build a JS app. Yes, this is new territory for the DP code but it's how modern web development is done and what we need the future proofreading interface to look like.

scripts/text_control.js
```diff
@@ -0,0 +1,53 @@
+/*global $ */
+// this operates the font face, font size and wrap controls
+$(function () {
+    var fontFaceSelector = document.getElementById("fntFace");
```
chrismiceli, line 4: why jquery sometimes and other time just normal DOM APIs? For instance here could've been $('#...').

70ray: No particular reason. Some of this was taken from old code. It would probably be better to be more consistent as you suggest.

70ray:
> The interlaced PHP and HTML with JS is not the model that I'd like for us to move to for these JS-based applications. If we're going to build a JS app, let's build a JS app. Yes, this is new territory for the DP code but it's how modern web development is done and what we need the future proofreading interface to look like.

I think I see what you mean. For displaying text and/or image you would use a single 'app' which could load images directly and get texts by an ajax function. To switch between showing text and image the app would change the controls it showed (image size or round) rather than loading a different app. Have I understood this correctly? I'll look at doing something like this.

cpeel:
> I think I see what you mean. For displaying text and/or image you would use a single 'app' which could load images directly and get texts by an ajax function. To switch between showing text and image the app would change the controls it showed (image size or round) rather than loading a different app. Have I understood this correctly? I'll look at doing something like this.

AJAX would be nice, but let's just start with an HTML page that includes the JSON of data we want to build with and have the JS do the rest. Most JS widgets attach to a single entity (`<div>`, etc) and then create whatever HTML inside that entity are necessary to render their page -- including building forms, populating drop-downs, etc. This places all of the building logic in one place (the JS) and creates a clean separation from the data layer and the rendering layer. The page could still GET/POST via a form to the same HTML page to load the next set of images.

This also lets us place the entire JS building logic into a single contained JS file instead of a much of smaller files that are single-purpose.

In the future this could be enhanced to use AJAX, but it shouldn't start out that way.

So basically: let's create a JS "text & image browser" that works just like yours does, but is initialized with the following HTML
```html
<script>
// JSON populated by PHP for the project in question
var page_browser_data = {
    "projectid": "projectIDxxxx",
    "currentPage": "010.png",
    "showText": True,
    "pages": [
        "010.png",
        "011.png"
    ]
};

// when the page is ready, we instantiate the JS widget and it takes
// care of building out all of the UI bits
$(function() {
    var page_browser = PageBrowser({
        bindto: '#page_browser',
        data: browser_data
    });
});

</script>
<div id='page_browser'></div>
```

70ray:
Thanks Casey, this is useful.

70ray:
Revised version made.

PR #272, opened 2020-04-04
Encourage users to work in P2 where possible

cpeel:
_This code is up for review and feedback (particularly around wording) but is not intended to be, and will not be, merged._

With the crazy influx of new users, we want to really encourage people to work in P2 if they can. This hack shows a call-out box on the P1 page. It has 3 different messages:
1. If the user already has P3 or P2 access, it will encourage them to proof in those rounds instead.
2. If the user is already eligible to work in P2, but just haven't requested access, it will encourage them to do so and proof in P2.
3. If the user has done the minimum number of pages and been around long enough but not taken/passed the quiz it will encourage them to take/pass the quiz and work in P2.

This is available for testing in the [covid19-hacks](https://www.pgdp.org/~cpeel/c.branch/covid19-hacks) sandbox.

One oddity about the test site is that it reduces the minimum page and time criteria to get into the rounds (3 days and 10 pages rather than 21 days and 300 pages) -- but the gradual welcome messages are hard-coded at PROD-level numbers. This means that on TEST you may see two callouts (existing and the new) whereas on PROD you will only see one or the other.

### DIFF
--- a/pinc/gradual.inc
+++ b/pinc/gradual.inc
@@ -148,7 +148,7 @@ function encourage_p1_to_p2($round)
 
         echo "<div class='callout'>";
         echo "<div class='calloutheader'>";
-        echo _("Consider working on projects in $round_target!");
+        echo "Consider working on projects in $round_target!";
         echo "</div>";
 
         echo "<p>Please consider proofreading a project in $round_target rather than here in P1. With the large influx of new users, having eligible users working in <a href='$code_url/tools/proofers/round.php?round_id=$round_target'>$round_target</a> helps significantly.</p>";
@@ -164,7 +164,7 @@ function encourage_p1_to_p2($round)
     {
         echo "<div class='callout'>";
         echo "<div class='calloutheader'>";
-        echo _("You are eligible to work in P2!");
+        echo "You are eligible to work in P2!";
         echo "</div>";
 
         echo "<p>You are eligible to work on projects in P2! Simply <a href='$code_url/tools/proofers/round.php?round_id=P2#Entrance_Requirements'>request access</a> and find a <a href='$code_url/tools/proofers/round.php?round_id=P2'>P2</a> project that interests you.</p>";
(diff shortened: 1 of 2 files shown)


cpeel:
We could also do something similar on the P2 page to encourage users to format in F1.

pinc/gradual.inc
```diff
@@ -127,4 +127,70 @@ function maybe_output_new_proofer_message()
...
+    // take and pass the quiz, so nudge them along
+    $uao = $round->user_access($pguser);
+    if($uao->minima_table["P1"][3] && $uao->minima_table["days since reg"][3])
+    {
+        echo "<div class='callout'>";
+        echo "<div class='calloutheader'>";
+        echo _("You are a quiz away from working in P2!");
+        echo "</div>";
+
+        echo "<p>You are almost eligible to work in P2! The only thing left is to take and pass the <a href='$code_url/quiz/start.php?show_level=P_MOD'>moderate proofreading quiz</a>. After you pass the quiz you can <a href='$code_url/tools/proofers/round.php?round_id=P2#Entrance_Requirements'>request access</a> and you'll be able to work on projects in <a href='$code_url/tools/proofers/round.php?round_id=P2'>P2</a>.</p>";
```
srjfoo, line 188: Shouldn't that be `quizzes`? Also, are all 10 moderate quizzes required, or just the first group of 5?

lhamilton1: How about "just a few short quizzes away"? Also could we say "working in the P2 round"? These are beginners and may not recognize what P2 means. The second use could be just "P2" though.

lhamilton1: Also -- "If you're qualified to work in a higher round, please consider working in that round." might be better for the first sentence "please consider...." since we would like P3s and F2s especially to work in those round.

cpeel: I'll get the first changes in.

Proofreaders who are already able to work in P2 or P3 will get the first message in this change so we don't need to tweak this message to apply to them.

cpeel: I've adjusted the wording. I'm trying to keep the first bold line short and succinct.

srjfoo: Looks good to me.

pinc/gradual.inc
```diff
@@ -127,4 +127,70 @@ function maybe_output_new_proofer_message()
...
+            $round_target = "P3";
+        else
+            $round_target = "P2";
+
+        echo "<div class='callout'>";
+        echo "<div class='calloutheader'>";
+        echo _("Consider working on projects in $round_target!");
+        echo "</div>";
+
+        echo "<p>Please consider proofreading a project in $round_target rather than here in P1. With the large influx of new users, having eligible users working in <a href='$code_url/tools/proofers/round.php?round_id=$round_target'>$round_target</a> helps significantly.</p>";
```
chrismiceli, line 154: Are we not localizing as this is temporary?

srjfoo: Hm ... may just have missed it, as the title line is marked up. I didn't notice it, either.

cpeel: The markup in the title line was, frankly, accidental. This is Super Temporary so I don't want to bother with proper gettext markup.

I am noodling on ways to abstract this out to be less temporary which will get proper markup.

cpeel:
Closing as this was never intended to be merged in.

PR #274, opened 2020-04-04
2274: Normalize text input for nameofwork (Title) and authorsname (Au…

chrismiceli:
…thor) fields

normalize fields from user in project
add script to normalize fields in project table

### DIFF
--- a/SETUP/upgrade/14/20200404_normalize_project_details.php
+++ b/SETUP/upgrade/14/20200404_normalize_project_details.php
@@ -1,7 +1,6 @@
 <?php
 $relPath='../../../pinc/';
 include_once($relPath.'base.inc');
-include_once($relPath.'misc.inc');
 include_once($relPath.'unicode.inc');
 
 header('Content-type: text/plain');
@@ -24,43 +23,28 @@ while($row = mysqli_fetch_assoc($result))
     $projects[] = $row;
 }
 
-$stop_file = '/tmp/utf8_normalization';
-
-echo <<<EOF
-This will now start normalizing all of the project fields.
-This will likely take a while.
-To stop the process gracefully create $stop_file
-(eg: touch $stop_file) and it will stop after it
-finishes the table it is working on.
-
-EOF;
-
 echo count($projects) . " possible projects to normalize.\n";
 
 foreach($projects as $project)
 {
-    if(file_exists($stop_file))
-    {
-        echo "\n\n";
-        echo "Detected stop file ($stop_file), so stopping before starting " . $project["projectid"] . ".\n";
-        echo "Delete this file and restart to continue.\n";
-        exit();
-    }
+    $sql = sprintf("
+            UPDATE projects SET
+            nameofwork    = '%s',
+            authorsname   = '%s',
+            comments      = '%s',
+            extra_credits = '%s'
+            WHERE projectid = '%s';
+        ",
+        mysqli_real_escape_string(DPDatabase::get_connection(), utf8_normalize(html_entity_decode($project["nameofwork"]))),
+        mysqli_real_escape_string(DPDatabase::get_connection(), utf8_normalize(html_entity_decode($project["authorsname"]))),
+        mysqli_real_escape_string(DPDatabase::get_connection(), utf8_normalize($project["comments"])),
+        mysqli_real_escape_string(DPDatabase::get_connection(), utf8_normalize(html_entity_decode($project["extra_credits"]))),
+        $project["projectid"]
+    );
 
-    
-    $sql = "
-        UPDATE projects SET
-        nameofwork    = '".mysqli_real_escape_string(DPDatabase::get_connection(), utf8_normalize($project["nameofwork"]))."',
-        authorsname   = '".mysqli_real_escape_string(DPDatabase::get_connection(), utf8_normalize($project["authorsname"]))."',
-        comments      = '".mysqli_real_escape_string(DPDatabase::get_connection(), utf8_normalize($project["comments"]))."',
-        extra_credits = '".mysqli_real_escape_string(DPDatabase::get_connection(), utf8_normalize($project["extra_credits"]))."'
-        WHERE projectid = '".$project["projectid"]."';
-    ";
-
+    echo "Processing " . $project["projectid"] . "\n";
     mysqli_query(DPDatabase::get_connection(), $sql) or die( mysqli_error(DPDatabase::get_connection()) );
 }
-
-echo "\n";
 
 // ------------------------------------------------------------
 
--- a/SETUP/upgrade/14/20200404_normalize_project_details.php
+++ b/SETUP/upgrade/14/20200404_normalize_project_details.php
@@ -1,0 +1,69 @@
+<?php
+$relPath='../../../pinc/';
+include_once($relPath.'base.inc');
+include_once($relPath.'misc.inc');
+include_once($relPath.'unicode.inc');
+
+header('Content-type: text/plain');
+
+// ------------------------------------------------------------
+
+echo "Normalizing project details...";
+
+$sql = "
+    SELECT projectid, nameofwork, authorsname, comments, extra_credits
+    FROM projects
+    ORDER BY projectid;
+";
+
+$result = mysqli_query(DPDatabase::get_connection(), $sql) or die( mysqli_error(DPDatabase::get_connection()) );
+
+$projects = array();
+while($row = mysqli_fetch_assoc($result))
+{
+    $projects[] = $row;
+}
+
+$stop_file = '/tmp/utf8_normalization';
+
+echo <<<EOF
+This will now start normalizing all of the project fields.
+This will likely take a while.
+To stop the process gracefully create $stop_file
+(eg: touch $stop_file) and it will stop after it
+finishes the table it is working on.
+
+EOF;
+
+echo count($projects) . " possible projects to normalize.\n";
+
+foreach($projects as $project)
+{
+    if(file_exists($stop_file))
+    {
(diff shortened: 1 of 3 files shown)


SETUP/upgrade/14/20200404_normalize_project_details.php
```diff
@@ -0,0 +1,69 @@
...
+    if(file_exists($stop_file))
+    {
+        echo "\n\n";
+        echo "Detected stop file ($stop_file), so stopping before starting " . $project["projectid"] . ".\n";
+        echo "Delete this file and restart to continue.\n";
+        exit();
+    }
+
+    
+    $sql = "
```
cpeel, line 51: Let's use `sprintf()` to make this more readable:
```
$sql = sprintf("
    UPDATE projects
    SET
        nameofwork = '%s',
        authorsname = '%s',
        comments = '%s',
        extra_credits = '%s'
    WHERE projectid = '%s'
",
    mysqli_real_escape_string(DPDatabase::get_connection(), utf8_normalize($project["nameofwork"])),
    mysqli_real_escape_string(DPDatabase::get_connection(), utf8_normalize($project["authorsname"])),
    mysqli_real_escape_string(DPDatabase::get_connection(), utf8_normalize($project["comments"])),
    mysqli_real_escape_string(DPDatabase::get_connection(), utf8_normalize($project["extra_credits"])),
    $projectid
);
```

SETUP/upgrade/14/20200404_normalize_project_details.php
```diff
@@ -0,0 +1,69 @@
...
+
+$result = mysqli_query(DPDatabase::get_connection(), $sql) or die( mysqli_error(DPDatabase::get_connection()) );
+
+$projects = array();
+while($row = mysqli_fetch_assoc($result))
+{
+    $projects[] = $row;
+}
+
+$stop_file = '/tmp/utf8_normalization';
```
cpeel, line 27: Because we're just operating on a single table (projects) and not over thousands of tables, I feel comfortable doing this update without a stop_file.

SETUP/upgrade/14/20200404_normalize_project_details.php
```diff
@@ -0,0 +1,69 @@
...
+        echo "\n\n";
+        echo "Detected stop file ($stop_file), so stopping before starting " . $project["projectid"] . ".\n";
+        echo "Delete this file and restart to continue.\n";
+        exit();
+    }
+
+    
+    $sql = "
+        UPDATE projects SET
+        nameofwork    = '".mysqli_real_escape_string(DPDatabase::get_connection(), utf8_normalize($project["nameofwork"]))."',
```
cpeel, line 53: For nameofwork, authorsname, and extra_credits it would be a big help to run these strings through `html_entity_decode()` first, so that any HTML entities are converted into their Unicode equiv first. Turns out we have several projects on PROD that have HTML entities in them. We *don't* want to do that for comments since it's expected those actually do have HTML entities we want to retain.

SETUP/upgrade/14/20200404_normalize_project_details.php
```diff
@@ -0,0 +1,69 @@
+<?php
+$relPath='../../../pinc/';
+include_once($relPath.'base.inc');
+include_once($relPath.'misc.inc');
```
cpeel, line 4: I don't _think_ this is using `misc.inc`.

SETUP/upgrade/14/20200404_normalize_project_details.php
```diff
@@ -0,0 +1,69 @@
...
+(eg: touch $stop_file) and it will stop after it
+finishes the table it is working on.
+
+EOF;
+
+echo count($projects) . " possible projects to normalize.\n";
+
+foreach($projects as $project)
+{
+    if(file_exists($stop_file))
```
cpeel, line 42: Would be nice to output the projectid being worked on in the loop so we can see progress as it goes.

SETUP/upgrade/14/20200404_normalize_project_details.php
srjfoo: There are a small handful of projects on DPUT that have entities in the titles, but none in the `authorsname` or `extra_credits` fields. Should I mung a few of them so that we can test all fields (probably not strictly necessary).

I agree that if PMs use entities in the Project Comments, they should be left as they are. What about the `postcomments`? I show none on TEST, but 38 in the `projects` table on PROD. By comparison, only 2 in the `extra_credits` field.

cpeel: As long as there are HTML entities in some fields in TEST that the script can convert for validation I don't think we need something in all of them -- so I wouldn't bother.

Ah, good questions. We should be normalizing `postcomments` as well, but I think we should treat them like we do `comments` as far as HTML entities (that is: leave them as entities).

SETUP/upgrade/14/20200404_normalize_project_details.php
```diff
@@ -24,44 +23,29 @@
...
-        UPDATE projects SET
-        nameofwork    = '".mysqli_real_escape_string(DPDatabase::get_connection(), utf8_normalize($project["nameofwork"]))."',
-        authorsname   = '".mysqli_real_escape_string(DPDatabase::get_connection(), utf8_normalize($project["authorsname"]))."',
-        comments      = '".mysqli_real_escape_string(DPDatabase::get_connection(), utf8_normalize($project["comments"]))."',
-        extra_credits = '".mysqli_real_escape_string(DPDatabase::get_connection(), utf8_normalize($project["extra_credits"]))."'
-        WHERE projectid = '".$project["projectid"]."';
-    ";
-
+    $sql = sprintf("
+            UPDATE projects SET
```
cpeel, line 31: I know this is minor, but let's out-dent the UPDATE and WHERE line so that they are indented less. That maps to how we try to lay out SQL for better readability elsewhere.

SETUP/upgrade/14/20200404_normalize_project_details.php
```diff
@@ -0,0 +1,53 @@
...
+
+$projects = array();
+while($row = mysqli_fetch_assoc($result))
+{
+    $projects[] = $row;
+}
+
+echo count($projects) . " possible projects to normalize.\n";
+
+foreach($projects as $project)
```
cpeel, line 28: To prevent loading all of the project data into memory (the `comments` field in particular is rather large) we can move the query iteration to this part and do:
```
while($project = mysqli_fetch_assoc($result))
{
```

We lose the "X possible projects.." message, but I think the trade-off is worth it.

SETUP/upgrade/14/20200404_normalize_project_details.php
srjfoo: As long as they are treated like `comments` in terms of display, that should be fine. And yes, definitely normalize them, too!

chrismiceli: Ok updated on save and added postcomments too. However, I'm curious about the decoding stuff. Isn't that for html display? Why do we care if they have html like data in the database. Shouldn't that only be a display issue?

srjfoo: It's okay if it's in the database for the `postcomments` and `comments` fields, but not for `nameofwork` and `authorsname`. I'm not sure about `extra_credits`, but I'd err on the side of No, unless we really want active links in that field. Do a project search on the test site for `<b>` or `<i>` in the title field. I think there's also `<b>` in the author field. See the results of what happens on the project page for this project: [Death's blue screen {test}](https://www.pgdp.org/c/project.php?id=projectID44dcfc8b0b855).

srjfoo: (It's not the presence of the markup that messes with the display of the project page, but the fact that the element is not closed.)

chrismiceli: but should we just encode before rendering so that isn't an issue? that is how usually apps handle persisted XSS issues.

srjfoo: Hm. Okay, I think I misunderstood your original question. Are you not questioning the fact that it needs to be done, but the order in which it should be done?

chrismiceli: That will be resolved with the changes that @cpeel backed out recently where we encode properly before displaying the name in the UI.

srjfoo: Then I obviously went off on a complete tangent and don't know what I'm talking about!

pinc/smoothread.inc
```diff
@@ -283,7 +283,7 @@ function handle_smooth_reading_change($project, $postcomments, $days, $extend)
         SET $smoothread_deadline_sql
             postcomments = CONCAT(postcomments, '%s')
         WHERE projectid = '$projectid'
-    ", mysqli_real_escape_string(DPDatabase::get_connection(), $postcomments));
+    ", mysqli_real_escape_string(DPDatabase::get_connection(), html_entity_decode($postcomments)));
```
cpeel, line 286: It sounds like this was hashed out in the Slack discussion, but we only need `html_entity_decode()` in the update script to fix problematic database entries. [This forum post](https://www.pgdp.net/phpBB3/viewtopic.php?f=16&t=16898&p=1197231#p1197231) discusses how they got in there, why we need to convert them, and why we don;t have to worry about this after we switch to Unicode.

cpeel:
This looks perfect. I will do a few test runs to make sure that this works the way we all think it does and get it merged in.

cpeel:
This worked _beautifully_.

PR #275, opened 2020-04-06
Require transitions to P1.unavail to pass PQC

cpeel:
Prevent projects from transitioning from NEW to P1.unavailable if PQC finds any errors.

This therefore changes the Bad Bytes check to output 'Error' instead of 'Warning'. This does involve a fair amount of out and back again redirecting using encoded URLs within encoded URLs and there is a nominal maximum URL length of ~2k characters. I don't think we're getting close to this limit but it is something I worry about.

This is available in the [pqc-in-transition](https://www.pgdp.org/~cpeel/c.branch/pqc-in-transition/) sandbox.

### DIFF
--- a/pinc/project_quick_check.inc
+++ b/pinc/project_quick_check.inc
@@ -2,8 +2,53 @@
 include_once('RoundDescriptor.inc'); // get_Round_for_round_number()
 include_once('wordcheck_engine.inc'); // get_file_info_object()
 include_once('links.inc'); // new_window_link()
+include_once('slim_header.inc'); // slim_header()
 include_once($relPath.'../tools/project_manager/post_files.inc'); // page_info_query page_info_fetch
 include_once('bad_bytes.inc');
+
+// Transition callback function to prevent a project state change if PQC
+// results return an error.
+function gate_on_pqc($projectid)
+{
+    global $code_url;
+
+    $results = get_pqc_test_results($projectid);
+
+    // loop thorugh results looking for errors
+    $found_error = FALSE;
+    foreach($results as $function => $test_result)
+    {
+        if($test_result["status"] == _("Error"))
+        {
+            $found_error = TRUE;
+            break;
+        }
+    }
+
+    if($found_error)
+    {
+        $title = _("Change Project State");
+        slim_header($title);
+        echo "<h1>$title</h1>";
+
+        echo "<p>";
+        echo sprintf(_("One or more errors were found when checking the project. Review the <a href='%s'>project quick check results</a>, address any problems found, and try the project transition again."), "$code_url/tools/project_manager/project_quick_check.php?projectid=$projectid");
+        echo "</p>";
+
+        echo "<p>";
+        echo _("If you believe these results are incorrect, please contact db-req for assistance.");
+        echo "</p>";
+
+        echo "<p>";
+        echo sprintf(_("Return to the <a %s>project page</a>"), "href='$code_url/project.php?id=$projectid'");
+        echo "</p>";
+
+        echo "<h2>" . _("Project Quick Check Summary") . "</h2>";
+        show_pqc_result_summary($results, FALSE);
+
+        exit();
+    }
+}
 
 function show_pqc_result_summary($results, $output_links = True)
 {
@@ -463,7 +508,7 @@ function _test_project_for_bad_bytes($projectid)
     }
     else
     {
-        $r['status'] = _('Warning');
+        $r['status'] = _('Error');
         $r['summary'] = sprintf(_("%d pages had bad characters, see detail."), $n_bad_pages);
         $r['details'] = $details_for_this_project;
     }
--- a/pinc/project_quick_check.inc
+++ b/pinc/project_quick_check.inc
@@ -1,11 +1,81 @@
 <?php
-include_once('stages.inc'); // $Round_for_round_id_
-include_once('wordcheck_engine.inc'); // $word_characters $puncCharacters get_file_info_object()
+include_once('RoundDescriptor.inc'); // get_Round_for_round_number()
+include_once('wordcheck_engine.inc'); // get_file_info_object()
 include_once('links.inc'); // new_window_link()
 include_once($relPath.'../tools/project_manager/post_files.inc'); // page_info_query page_info_fetch
 include_once('bad_bytes.inc');
-include_once('wordcheck_engine.inc');
 
+function show_pqc_result_summary($results, $output_links = True)
+{
+    echo "<table class='basic striped'>";
+    echo "<tr>";
+    echo "<th>" . _("Test") . "</th>";
+    echo "<th>" . _("Status") . "</th>";
+    echo "<th>" . _("Summary") . "</th>";
+    echo "</tr>";
+    foreach($results as $function => $test_result)
+    {
+        echo "<tr>";
+        echo "<td>" . $test_result["name"] . "</td>";
+        echo "<td>";
+        if($output_links)
+            echo "<a href='#$function'>";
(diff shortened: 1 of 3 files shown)


tools/project_manager/project_quick_check.php
```diff
@@ -14,31 +14,70 @@
...
-echo   "<td>" . _("Project ID") . "</td>";
-echo   "<td><input name='projectid' type='text' value='$projectid' size='40' required></td>";
-echo  "</tr>";
-echo "</table>";
-echo "<input type='submit' value='Test'>";
-echo "</form>";
+if($callback)
+{
+    echo "<p>";
+    echo _("One or more errors were found when checking the project. Review the project quick check results, address any problems found, and try the project transition again.");
```
srjfoo, line 54: The only thing that will trigger displaying the PQC results at this stage is an error in loading the text. How about:
"One or more errors were found when loading the text files into the project. Please review the errors listed in the Bad Bytes in Page Text (add link) section, below.

tools/project_manager/project_quick_check.php
```diff
@@ -14,31 +14,70 @@
...
+    echo "</p>";
 
-// stop if no projectid was specified
-if(empty($projectid))
+    echo "<p>";
+    echo sprintf(_("If you feel this project quick check result is incorrect, you can <a href='%s'>continue with the transition</a>."), $callback);
+    echo "</p>";
```
srjfoo, line 59: I can understand wanting to give them the opportunity to override, but I'd move the ability to override to the bottom of the list of errors. If it's at the top of the page, they're not really motivated to look. Or may click on it accidentally.

Perhaps put a link back to the project page here, instead? "Return to Project Page (link) to reload corrected text files."?

(This is my opinion, only and I'm open to other suggestions.)

tools/project_manager/project_quick_check.php
cpeel: There are a few checks that return Errors, this MR just elevates the bad bytes check to that level. I don't want to focus the language on any one check.

tools/project_manager/project_quick_check.php
cpeel: Given Michael's suggestion below, this might be moot but I will see.

cpeel:
I don't see the point in moving the PQC detour to another script. I also don't like how the other detour scripts change state -- that's literally the whole point of `changestate.php`.

I *do* like the idea of making detour a callable and having the `changestate.php` script. I will investigate that.

tools/project_manager/project_quick_check.php
srjfoo: Understood.

jmdyck:
> I don't see the point in moving the PQC detour to another script.

The point would be to have 2 simple control-flows rather than one complicated control-flow.

> I also don't like how the other detour scripts change state -- that's literally the whole point of `changestate.php`.

While `changestate.php` is concerned with changing project state, that doesn't mean that all state-changes should go through it. Its job is to respond to someone pressing a state-change button. State-changes that aren't the immediate result of such a button-press don't involve `changestate.php`, and instead call `project_transition()`. (Of course, those two mechanisms should share as much code as possible, mainly via ProjectTransition.)

> I _do_ like the idea of making detour a callable and having the `changestate.php` script. I will investigate that.

Ah good.

(Theoretically, the detour could *always* be a callable (or null), and the status quo detours could be handled by a function that does the necessary url-prep and metarefresh. But I don't think the purity would be worth the churn.)

cpeel:
Code updated to use a callback function -- this approach I like much better.

This version does not have an override. If we really feel that the PQC errors are errors there probably shouldn't be one. Sandbox has been updated.

srjfoo:
I like having the override removed.

I do like that the initial display focuses on the PQC summary (it definitely stands out more), but I'm not entirely sure that it makes sense to do it that way -- PMs don't see this *unless* there's an error. And, while they can hit the back button and go back to the project page, they don't see the details of any errors if they do that: they have to go to the full PQC page to see what the error(s) is(are). Does that make sense?

cpeel:
No, sorry, it doesn't. If a project fails PQC during a transition, what would you like/expect to see that you don't?

srjfoo:
When I try to move a project that contains bad characters to P1 Unavailable, I see:
image

That just tells me that there's at least one bad character on one page, and to go to the PQC results to review the results. If I'm going to have to go to PQC to review the results, anyway, why not just show it to me in the first place?

Or if you don't want to show the full PQC page by default, maybe provide a link back to the project page, as well, for those who facepalm because they realize what they did wrong and don't need to look? Something like: "For details, review the project quick check results (with link). Otherwise, return to the project page (with link), address any problems, and try the project transition again."

I'm making the assumption that it would not be easy to include the detailed section of the area that triggered the error. along with the PQC summary table, and have the link to PQC be there just in case they want to check the full report.

cpeel:
If we're going to show all of the results why not just take them to the PQC page if there is an error? We can certainly do that. I don't want to just recreate the PQC page here (whether via functions or not). One downside to this approach is the the transition would effectively have to run PQC twice (once to see if there is an error before we redirect them, once after we redirect them) so that could take a while time-wise.

We could also add a "Return to the project page" link instead.

Either one is fine and I can make work, just let me know which you'd prefer I do.

srjfoo:
> If we're going to show all of the results why not just take them to the PQC page if there is an error? We can certainly do that. I don't want to just recreate the PQC page here (whether via functions or not). One downside to this approach is the the transition would effectively have to run PQC twice (once to see if there is an error before we redirect them, once after we redirect them) so that could take a while time-wise.

Thanks for the explanation, that helps. 

> 
> We could also add a "Return to the project page" link instead.
> 
> Either one is fine and I can make work, just let me know which you'd prefer I do.

Given your description of what's involved in just showing them the PQC results page instead of the summary, I vote for adding a link back to the project page, as well as the link to PQC results. There's a good chance that if they get the error message, they'll know what the problem is without looking at the PQC results directly, and a link back to the project page doesn't force them to go to PQC.

That's why I asked the question in the first place -- it appeared that the options were PQC, close the window or use the browser back button. If we're going to try to steer away from encouraging using the browser back button, having an explicit link back to the project page seems like a good idea.

cpeel:
Return to project link added.

srjfoo:
Looks good to me!

pinc/project_quick_check.inc
```diff
@@ -1,11 +1,121 @@
...
+    }
+    return $result;
+}
+
+// ---------------------------------------------------------------------------
+// Test functions
+
+// $test_functions specifies which project test functions are available.
+// The keys are function names (defined below) which will be called with
+// a projectid. Functions return an associative array of the form:
```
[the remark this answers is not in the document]
cpeel: I've massaged the wording slightly from what you have, but I like it.

Yes, I think a lot of the PQC code needs to be refactored into a class, but that's for a separate MR.

pinc/project_quick_check.inc
```diff
@@ -1,11 +1,121 @@
...
+        slim_header($title);
+        echo "<h1>$title</h1>";
+
+        echo "<p>";
+        echo sprintf(_("One or more errors were found when checking the project. Review the <a href='%s'>project quick check results</a>, address any problems found, and try the project transition again."), "$code_url/tools/project_manager/project_quick_check.php?projectid=$projectid");
+        echo "</p>";
+
+        echo "<p>";
+        // string taken from tools/changestate.php
+        echo _("If you believe this is an error, please contact db-req for assistance.");
```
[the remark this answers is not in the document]
cpeel: Good point. I've updated the wording to be less ambiguous.

PR #276, opened 2020-04-07
Daily Page Limit Per Round

jmdyck:
Each round optionally gets a daily limit on the (net) number of pages saved-as-done in that round.

This is a simplified version of PR #204. There, each DPL defines the set of projects that it applies to, based on any project properties; here, the only option is for a DPL to apply to all the projects of a round. So the 'back-end' is simplified, but the 'front-end' is, in practice, pretty much identical.

The messages to the user could probably be made shorter. I'll leave it to the reviewers to suggest specific wording.

### DIFF
--- a/pinc/RoundDescriptor.inc
+++ b/pinc/RoundDescriptor.inc
@@ -197,15 +197,8 @@ class Round extends Stage
         $round_tallyboard = new TallyBoard($this->id, 'U');
 
         $user = new User($username);
-        $u_id = $user->u_id;
 
-        $current_tally = $round_tallyboard->get_current_tally($u_id);
-        $snapshot_info = $round_tallyboard->get_info_from_latest_snapshot($u_id);
-        $latest_snapshot_tally = $snapshot_info['tally_value'];
-        $tally_delta_since_snapshot = $current_tally - $latest_snapshot_tally;
-        // (Alternatively, could extract it from the result of 'get_vitals')
-
-        return $tally_delta_since_snapshot;
+        return $round_tallyboard->get_delta_since_latest_snapshot($user->u_id);
     }
 
     // -----------
--- a/pinc/RoundDescriptor.inc
+++ b/pinc/RoundDescriptor.inc
@@ -187,20 +187,6 @@ class Round extends Stage
         return !is_null($this->daily_page_limit);
     }
 
-    // XXX This probably doesn't belong here, but I'm not sure where it does.
-    function get_dpl_count_for_user($username)
-    // The current "dpl count" for a user (in a given round)
-    // is the delta in their page-tally for that round
-    // since the last snapshot.
-    // (i.e., the net number of save-as-dones in that round today.)
-    {
-        $round_tallyboard = new TallyBoard($this->id, 'U');
-
-        $user = new User($username);
-
-        return $round_tallyboard->get_delta_since_latest_snapshot($user->u_id);
-    }
-
     // -----------
 
     function get_honorific_for_page_tally( $page_tally )
--- a/pinc/RoundDescriptor.inc
+++ b/pinc/RoundDescriptor.inc
@@ -53,6 +53,10 @@ class Round extends Stage
             // See pinc/ProofreadingToolbox.inc for which tools are avaialble
             // and their names.
 
+        $daily_page_limit,
+            // This is either NULL (indicating no daily page limit for this round)
+            // or a non-negative (though likely positive) integer.
+
         $other_rounds_with_visible_usernames,
             // An array of round_ids.
             // If user X worked on a page in this round, they can see the
@@ -87,6 +91,7 @@ class Round extends Stage
         $this->round_number       = $n_rounds;
 
         $this->pi_tools = $pi_tools;
+        $this->daily_page_limit = $daily_page_limit;
         $this->other_rounds_with_visible_usernames = $other_rounds_with_visible_usernames;
         $this->honorifics = $honorifics;
         krsort($this->honorifics);
@@ -174,6 +179,36 @@ class Round extends Stage
 
     function is_a_mentee_round() { return !is_null($this->mentor_round); }
     function is_a_mentor_round() { return !is_null($this->mentee_round); }
+
+    // -----------
+
+    function has_a_daily_page_limit()
+    {
+        return !is_null($this->daily_page_limit);
+    }
+
+    // XXX This probably doesn't belong here, but I'm not sure where it does.
+    function get_dpl_count_for_user($username)
+    // The current "dpl count" for a user (in a given round)
+    // is the delta in their page-tally for that round
+    // since the last snapshot.
+    // (i.e., the net number of save-as-dones in that round today.)
+    {
+        $round_tallyboard = new TallyBoard($this->id, 'U');
+
+        $user = new User($username);
+        $u_id = $user->u_id;
+
+        $current_tally = $round_tallyboard->get_current_tally($u_id);
+        $snapshot_info = $round_tallyboard->get_info_from_latest_snapshot($u_id);
+        $latest_snapshot_tally = $snapshot_info['tally_value'];
+        $tally_delta_since_snapshot = $current_tally - $latest_snapshot_tally;
+        // (Alternatively, could extract it from the result of 'get_vitals')
+
+        return $tally_delta_since_snapshot;
+    }
+
+    // -----------
 
     function get_honorific_for_page_tally( $page_tally )
     {
--- a/pinc/daily_page_limit.inc
+++ b/pinc/daily_page_limit.inc
@@ -1,4 +1,26 @@
 <?php
+
+// This include file is dedicated to the "Daily Page Limit" (DPL) feature:
+// each round can optionally impose a daily limit on the 
+// number of pages that each user can save-as-done in a server day.
+//
+// (More precisely, the limit is on the *net* number of pages saved-as-done;
+// reopens and clears decrement the count. This is the same quantity tracked
+// by the page-tally system, so we just consult it for the user's "dpl count".)
+
+// There's only one function here because the rest of the implementation
+// of the feature fits fairly well elsewhere.
+//
+// Limits are set in invocations of the Round constructor in stages.inc.
+//
+// Limits are consulted and imposed at 2 points:
+// - on the project page:
+//   See decide_blurbs() in project.php.
+//
+// - in the proofing interface:
+//   See LPage::attemptSaveAsDone() in LPage.inc
+//   and PPage::attempt_to_save_as_done() in PPage.inc
+
 include_once($relPath.'User.inc');
 include_once($relPath.'TallyBoard.inc');
 
--- a/pinc/daily_page_limit.inc
+++ b/pinc/daily_page_limit.inc
@@ -1,0 +1,18 @@
+<?php
+include_once($relPath.'User.inc');
+include_once($relPath.'TallyBoard.inc');
+
+function get_dpl_count_for_user_in_round($username, $round)
+// The current "dpl count" for a user in a given round
+// is the delta in their page-tally for that round
+// since the last snapshot.
+// (i.e., the net number of save-as-dones in that round today.)
+{
+    $user = new User($username);
+
+    $round_tallyboard = new TallyBoard($round->id, 'U');
+
+    return $round_tallyboard->get_delta_since_latest_snapshot($user->u_id);
+}
+
+// vim: sw=4 ts=4 expandtab
--- a/project.php
+++ b/project.php
@@ -267,10 +267,10 @@ function decide_blurbs()
                 $round->daily_page_limit
             );
         }
-        else
+        elseif ($user_dpl_count >= 0.9 * $round->daily_page_limit)
         {
             // User hasn't reached this round's DPL,
-            // but they should be warned that there *is* a DPL for this round?
+            // but they should be warned that there *is* a DPL for this round.
             $msg = sprintf(
                 _("Warning: This project contributes to the '%s' limit of %d page-saves per day. Since server midnight, your current count is <b>%d</b>."),
                 $round->id,
@@ -278,6 +278,11 @@ function decide_blurbs()
                 $user_dpl_count
             );
             $page_limit_warning = "<br>\n$msg\n";
+        }
+        else
+        {
+            // They're not that close to reaching the DPL,
+            // so don't bother warning them about it.
         }
     }
     if ($blocked_by_limit_message)
--- a/project.php
+++ b/project.php
@@ -260,11 +260,12 @@ function decide_blurbs()
         if ($user_dpl_count >= $round->daily_page_limit)
         {
             // User has reached this round's DPL.
-            $blocked_by_limit_message = sprintf(
+            $msg = sprintf(
                 _("This project contributes to the '%s' limit of %d page-saves per day, and you have already reached that limit today."),
                 $round->id,
                 $round->daily_page_limit
             );
+            $blocked_by_limit_message = "<span class='warning'>$msg</span>";
             return array($blocked_by_limit_message, $blocked_by_limit_message);
         }
         elseif ($user_dpl_count >= 0.9 * $round->daily_page_limit)
@@ -272,12 +273,12 @@ function decide_blurbs()
             // User hasn't reached this round's DPL,
             // but they should be warned that there *is* a DPL for this round.
             $msg = sprintf(
-                _("Warning: This project contributes to the '%s' limit of %d page-saves per day. Since server midnight, your current count is <b>%d</b>."),
+                _("This project contributes to the '%s' limit of %d page-saves per day. Since server midnight, your current count is <b>%d</b>."),
                 $round->id,
                 $round->daily_page_limit,
                 $user_dpl_count
             );
-            $page_limit_warning = "<br>\n$msg\n";
+            $page_limit_warning = "<br>\n<span class='warning'>$msg</span>\n";
         }
         else
         {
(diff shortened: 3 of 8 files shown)


jmdyck:
Oh, also, I've only given this a small amount of testing.

jmdyck:
Something else: the "user experience" was originally designed with the F1 limit in mind, where it would only appear to fairly seasoned users. If we're thinking of putting this in P1, then it would (as currently coded) appear to even the newest of newbies, so think about their experience.  (E.g., should the site be silent about a DPL until the user is within some percentage of reaching it?)

tools/proofers/processtext.php
```diff
@@ -306,6 +306,82 @@ function leave_spellcheck_mode( $ppage )
...
+        .   "<li>"
+        .     sprintf(
+                _("Return to <a %s>the Activity Hub</a>"),
+                "href='$code_url/activity_hub.php' target='_top'"
+              )
+        .   "</li>\n"
+        . "</ul>\n"
+        ;
+
+    slim_footer();
```
cpeel, line 380: No need to output footers - the shutdown function takes care of that for us.

tools/proofers/processtext.php
```diff
@@ -306,6 +306,82 @@ function leave_spellcheck_mode( $ppage )
     }
 }
 
+function attempt_to_save_as_done($ppage, $text_data, $and_do_another)
+// This is only an attempt, because a daily page limit might block the save,
+// or prevent further saves.
+// If there's a problem, this function does not return to the caller.
+{
+    global $code_url, $pguser, $projectid, $proj_state;
```
cpeel, line 314: It would be nice if we passed in file-scope-globals like `$projectid` and `$proj_state`. The fewer global variables we have lying around the better. This isn't always easy, but it appears to be in this case.

tools/proofers/processtext.php
```diff
@@ -306,6 +306,82 @@ function leave_spellcheck_mode( $ppage )
...
+        .   "<li>"
+        .     sprintf(
+                _("Return to <a %s>round %s</a>"),
+                "href='round.php?round_id={$round->id}' target='_top'",
+                $round->id
+              )
+        .   "</li>\n"
+        .   "<li>"
+        .     sprintf(
+                _("Return to <a %s>the Activity Hub</a>"),
```
cpeel, line 373: Consider using `_("Return to the <a %s>Activity Hub</a>.")` to match the string in `report_bad_page.php`.

tools/proofers/processtext.php
```diff
@@ -306,6 +306,82 @@ function leave_spellcheck_mode( $ppage )
...
+        . sprintf(
+            _("You will not be allowed to save any more pages in %s until the next server midnight."),
+            $round->id
+          )
+        . "</p>\n";
+
+    echo "<ul>\n"
+        .   "<li>"
+        .     sprintf(
+                _("Return to <a %s>the Project Page</a>"),
```
cpeel, line 360: The code is all over the place with how we mark up this type of string for translation:
```
_("Return to Project Page for %s") -- entire string is the link
_("Return to the Project Page") -- entire string is the link
_('Return to Project Page') -- entire string is the link
_("Return to the <a %s>project page</a>.")
_("Return to <a href='%s'>the Project Page</a>")
```
And more. We should pick the form that we want to use everywhere and use it here and create a task to normalize all of the other ones together.

I have a preference for not capitalizing `Project Page` just the same as we don't capitalize `round page` -- we have multiple project pages just as we have multiple round pages.

My vote is therefore on:
```
_("Return to the <a %s>project page</a>")  // no period
```

tools/proofers/processtext.php
```diff
@@ -306,6 +306,82 @@ function leave_spellcheck_mode( $ppage )
...
+    slim_header( $title );
+
+    echo "<p>$sentence</p>\n";
+    
+    echo "<p>"
+        . sprintf(
+            _("You will not be allowed to save any more pages in %s until the next server midnight."),
+            $round->id
+          )
+        . "</p>\n";
```
cpeel, line 355: This seems like a good place to plug SR, as well as adding a link to the SR page in the listing below.

pinc/RoundDescriptor.inc
```diff
@@ -175,6 +180,36 @@ class Round extends Stage
...
     function is_a_mentor_round() { return !is_null($this->mentee_round); }
 
+    // -----------
+
+    function has_a_daily_page_limit()
+    {
+        return !is_null($this->daily_page_limit);
+    }
+
+    // XXX This probably doesn't belong here, but I'm not sure where it does.
```
cpeel, line 190: Yeah, not sure where is the best place for this. It seems odd to have it here. It might make more sense to have it in `User` and pass in the round.

project.php
```diff
@@ -248,6 +248,43 @@ function decide_blurbs()
...
+                _("This project contributes to the '%s' limit of %d page-saves per day, and you have already reached that limit today."),
+                $round->id,
+                $round->daily_page_limit
+            );
+        }
+        else
+        {
+            // User hasn't reached this round's DPL,
+            // but they should be warned that there *is* a DPL for this round?
+            $msg = sprintf(
```
cpeel, line 274: I think we should only warn the user when they are at some percentage of the limit. Perhaps 90%?

We should also output this in our warning CSS class and probably with a `<p>` tag so it stands out from the surrounding text.

project.php
```diff
@@ -248,6 +248,43 @@ function decide_blurbs()
...
+                _("Warning: This project contributes to the '%s' limit of %d page-saves per day. Since server midnight, your current count is <b>%d</b>."),
```
cpeel, line 275: Perhaps:
```
This project contributes to the %s limit of %d page-saves per day since server midnight. Your current count is <b>%d</b>.
```
If it's in the warning CSS class it shouldn't need the label and I don't think we need the single quotes around the round.

cpeel:
I have this code in the [daily_page_limit_per_round](https://www.pgdp.org/~cpeel/c.branch/daily_page_limit_per_round) sandbox if you want to poke at it @srjfoo. I've set a P1 limit of 2 so we can easily hit it 😁

lhamilton1:
I recall that when we tested years ago, Michael had put in a Midnight function to help us switch over to a new day. It helped with the testing. Not sure if it still exists.

project.php
srjfoo: 90% sounds good to me.

jmdyck:
Something else occurred to me: the midnight "snapshot" of page counts is not instantaneous, it takes about 10 minutes for the whole thing. If someone is proofing shortly after server midnight in a page-limited round, they would expect their dpl-count to have been reset minutes ago, but the system will still be using yesterday's dpl-count, so they might be confused/annoyed by the DPL messages they get. One mitigation is that the messages will at least be consistent with their "Pages Today" shown in the sidebar.

Some possible solutions:
- Document and apologize.
- Include some timing info in messages (e.g., how long ago your dpl-count was reset).
- Change the snapshot process so that it *is* instantaneous. (Not in this PR!)

cpeel:
I like #1 and #3. We should be able to make it instantaneous using transactions since those are InnoDB tables so I'd prefer we lean that direction and just document this corner case for now.

tools/proofers/processtext.php
jmdyck: Done!

tools/proofers/processtext.php
jmdyck: Done. (Mind you, that has a period where the other bullets don't.)

tools/proofers/processtext.php
jmdyck: Done.

tools/proofers/processtext.php
jmdyck: Eliminated those globals, but via extraction from $ppage rather than passing as separate parameters.

Which makes me wonder: should  `attempt_to_save_as_done` be a method of PPage?

project.php
jmdyck: 90%: Done.

'warning' class: Done. (And also for the "already reached the limit" message.)

`<p>` tag: I used `<span>` instead, because the "blurbs" mostly use `<br>` rather than `<p>`. (Not that I object to `<p>`, but it feels like that should be the subject of a separate refactoring.)

project.php
jmdyck: I've changed it to @srjfoo's wording.

I also changed the other message to eliminate the "This project contributes to" phrase; have a look.

jmdyck: I kept the "<b>" for the user's current dpl-count, but note that it has no effect, because layout.css says that 'warning' has font-weight bold.

jmdyck:
> Testing worked well. The one thing I couldn't test (because the test IDs I used to test didn't have any in progress pages from previous days) was what happens if I do my daily allotment and then go open an in progress page from a previous day. I've set up some pages so I can do that tomorrow.

Resuming an in-progress page should have no effect on your dpl-count. But reopening a saved-as-done should decrease your count (for 'today', regardless of when you saved-as-done) by 1.
 
> I am assuming that it's correct that as soon as I hit my limit, the Start Proofreading button goes away.

That's how the code is meant to work, and I believe how we agreed it should work. But that was a while ago, and in a different context (F1 rather than P1), so I don't know what's correct.

> Also that it's valid that if I go back and return a page done on a previous day to the round, because I opened it `today`, if I return it `today` it decrements `today`'s DPL. Thus, I can proof another page.

Right. The return-to-round itself has no effect on your count; the fact that you first have to reopen it is what decreases your count.

jmdyck:
> I recall that when we tested years ago, Michael had put in a Midnight function to help us switch over to a new day. It helped with the testing. Not sure if it still exists.

Yeah, it simulated the effect of server midnight as far as DPLs were concerned. That was pretty easy in the original design, because it just emptied the table that stored everyone's dpl-count, and so was basically only visible to people who were testing DPLs. In this design, the corresponding action would affect everyone's "actual" page-counts, which is probably not a big deal on the test server, but would be visible to people who are testing other things, and could possibly cause problems/consternation for them.

I suppose I could restrict the page-count-hacking to accounts that people are using to test this feature. Not sure I really want to bother though. Let me know if it would help.

lhamilton1:
We should be OK without the Midnight function. Thanks.

project.php
cpeel: LGTM!

tools/proofers/processtext.php
cpeel: That's a really good point. It does feel like PPage is a reasonable place for that.

project.php
cpeel: This LGTM!

srjfoo:
> > I recall that when we tested years ago, Michael had put in a Midnight function to help us switch over to a new day. It helped with the testing. Not sure if it still exists.
> 
> Yeah, it simulated the effect of server midnight as far as DPLs were concerned. That was pretty easy in the original design, because it just emptied the table that stored everyone's dpl-count, and so was basically only visible to people who were testing DPLs. In this design, the corresponding action would affect everyone's "actual" page-counts, which is probably not a big deal on the test server, but would be visible to people who are testing other things, and could possibly cause problems/consternation for them.
> 
> I suppose I could restrict the page-count-hacking to accounts that people are using to test this feature. Not sure I really want to bother though. Let me know if it would help.

I don't think I need it, but thanks!

> Testing worked well. The one thing I couldn't test (because the test IDs I used to test didn't have any in progress pages from previous days) was what happens if I do my daily allotment and then go open an in progress page from a previous day. I've set up some pages so I can do that tomorrow.

I've now tested this and gotten the expected message, so that worked fine, too.

>> Resuming an in-progress page should have no effect on your dpl-count. But reopening a saved-as-done should decrease your count (for 'today', regardless of when you saved-as-done) by 1.
>
> That's how the code is meant to work, and I believe how we agreed it should work. But that was a while ago, and in a different context (F1 rather than P1), so I don't know what's correct.

I'm fine with it working that way.

Given the variety of ways a page refresh from the proofing interface works, though, we may be in for interesting times (explaining to confused users why they're seeing what they're seeing when they get different responses to what they think are the same actions):
1. Page newly opened for proofing (at some point before leaving the page), refresh will save the current page in progress and move to the next checked out page. Basically acts as a "save in progress and check out another page".
2. Page saved as done, reopened, refresh will generate an error.
3. Page saved in progress and later reopened, refresh will do nothing (no message is generated).

>> Also that it's valid that if I go back and return a page done on a previous day to the round, because I opened it today, if I return it today it decrements today's DPL. Thus, I can proof another page.
>
> Right. The return-to-round itself has no effect on your count; the fact that you first have to reopen it is what decreases your count.

Good distinction, thanks.

If you can think of additional tests I should run, let me know.

cpeel:
[daily_page_limit_per_round](https://www.pgdp.org/~cpeel/c.branch/daily_page_limit_per_round) sandbox updated with Michael's latest code.

tools/proofers/processtext.php
jmdyck: Done.

pinc/RoundDescriptor.inc
jmdyck: I looked at putting it in `User`, that didn't feel right either.

I carved the guts of it off to a TallyBoard method; that seemed to make sense.

But then the function itself I just stuck in a new `inc` file.

srjfoo:
With a new DPL of 10 to test out the 90% threshold, it works well if you visit the project page between pages. Perhaps it's just my font, but because the width of the project table is constrained, the orphaned page count looks a bit out of place. Is it reasonable to force a line break, or is that not an option?

image

However, if you're doing "Save as Done and Proofread Next Page", you don't get a warning. I suspect that something similar to the `You may save....` warning on the project page should be added somehow. Otherwise, proofers may not actually ever see the warning. The question is whether to give them that warning on every page save after they reach the warning threshold, or perhaps just there and just preceding the checkout of the last allowable page. If the DPL is, for instance, 200 pages, I can see that getting a warning for each of the last 20 pages might get kind of annoying.

pinc/RoundDescriptor.inc
cpeel: Roger that. 👍

tools/proofers/processtext.php
cpeel: Fair point. Let's leave this without SR.

cpeel:
> With a new DPL of 10 to test out the 90% threshold, it works well if you visit the project page between pages. Perhaps it's just my font, but because the width of the project table is constrained, the orphaned page count looks a bit out of place. Is it reasonable to force a line break, or is that not an option?

I'd prefer not to force a line break. In my browser it all shows up easily on one line and I'd rather we not try to tweak the layout too much (particularly in light of the eventual page redesign).

> However, if you're doing "Save as Done and Proofread Next Page", you don't get a warning.

The message between pages is only 1 second, so adding much text is going to be more annoying than anything I think as it will be too much to read. I agree it would be ideal to have some more data about approaching the page limit between pages, but I don't think it's a ship-blocker. I suggest we roll out what we have as-is and tweak it further later as-needed.

pinc/daily_page_limit.inc
```diff
@@ -0,0 +1,18 @@
+<?php
+include_once($relPath.'User.inc');
```
cpeel, line 2: Can you please add some doc-level comment about this file and perhaps leave some breadcrumbs about where one might poke to see other parts of the daily_page_limit functionality? (Also adding a "dpl == daily_page_limit" together would help if someone is grepping the code and trying to figure out WTH "dpl" is 😁

srjfoo:
> > With a new DPL of 10 to test out the 90% threshold, it works well if you visit the project page between pages. Perhaps it's just my font, but because the width of the project table is constrained, the orphaned page count looks a bit out of place. Is it reasonable to force a line break, or is that not an option?
> 
> I'd prefer not to force a line break. In my browser it all shows up easily on one line and I'd rather we not try to tweak the layout too much (particularly in light of the eventual page redesign).

That's fair.

> 
> > However, if you're doing "Save as Done and Proofread Next Page", you don't get a warning.
> 
> The message between pages is only 1 second, so adding much text is going to be more annoying than anything I think as it will be too much to read. I agree it would be ideal to have some more data about approaching the page limit between pages, but I don't think it's a ship-blocker. I suggest we roll out what we have as-is and tweak it further later as-needed.

I agree it's more important to get it out than it is to worry with it now. I think we'll hear complaints, though.

cpeel:
> I agree it's more important to get it out than it is to worry with it now. I think we'll hear complaints, though.

Please point to anything we've deployed that hasn't generated complaints! 🤣

tools/proofers/PPage.inc
```diff
@@ -260,6 +260,84 @@ class PPage
...
+            );
+        }
+
+        slim_header( $title );
+
+        echo "<p>$sentence</p>\n";
+        
+        echo "<p>"
+            . sprintf(
+                _("You will not be allowed to save any more pages in %s until the next server midnight."),
```
srjfoo, line 307: We've had a request to tweak the wording here to `"You will be able to save more pages in %s after server midnight."`

srjfoo: Actually, that adjustment should be made everywhere the "You will not be allowed to ..." wording appears.

pinc/daily_page_limit.inc
jmdyck: Done!

tools/proofers/PPage.inc
jmdyck: Done.

jmdyck:
Once the latest version gets in the sandbox, a little testing would be good, as I've moved some code around.

pinc/daily_page_limit.inc
cpeel: Great - thank you!

tools/proofers/PPage.inc
jmdyck: > Actually, that adjustment should be made everywhere the "You will not be allowed to ..." wording appears.

This is the only place where that wording appears. However, the project page is another place where that wording *could* have appeared, and it's possible you might want the replacement text to be added there. (Currently, it says `<round> limits you to <n> page-saves per day, and you have already reached that limit today`.)

PR #277, opened 2020-04-07
Update CSS for much better mobile support

cpeel:
Update the CSS to better handle very narrow viewports such as those from mobile devices. This is not an attempt to make the site totally usable on mobile devices, but it is an attempt to make the main page, the Activity Hub, and project pages viewable on cell phones. I tried to make as few visible changes for desktop users as possible.

Note that when the screen size is small enough, the stats bar drops to below the page content. Unless the user has the stats bar set to be on the left, in which case the stats bar gets dropped entirely. I think I could make this work in all cases by using `flex` layout, but I worry what it would look like on older browsers.

This can be viewed in the [mobile-layout](https://www.pgdp.org/~cpeel/c.branch/mobile-layout) sandbox.

This was tested against a physical Pixel and several iPhones via the DP BrowserStack account. Assuming this passes the sniff test I will float it out to the larger community to poke at.


cpeel:
I also tested this against our list of minimum supported browsers -- and several that are much older -- and this code worked well on those as well.

PR #278, opened 2020-04-10
Increase PM limit for auto-releases

jmdyck:
Note that there are 2 commits: the first is a refactor to express the limit more sensibly, the second is the actual limit change.

### DIFF
--- a/tools/project_manager/autorelease.inc
+++ b/tools/project_manager/autorelease.inc
@@ -421,8 +421,9 @@ class ReleaseRestrictor
         }
 
         // ---------------------------------------------------------------------
-        // If an author is Anonymous, Unknown, or Various, we should allow 4 books by that 'Author'
-        // Otherwise we only allow one book by the author to be available in this round
+        // Block the project's release if there are projects by its author available in this round.
+        // If the author is 'Anonymous', 'Unknown', or 'Various', we block when there are 4 or more
+        // by that author available, but otherwise, we block at even a single project by that author.
 
         // TODO: Add a flag to some queues that all have identical authors that allows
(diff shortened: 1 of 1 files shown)


tools/project_manager/autorelease.inc
```diff
@@ -421,28 +421,28 @@ class ReleaseRestrictor
         }
 
         // ---------------------------------------------------------------------
-        // If an author is Anonymous, Unknown, or Various, we should allow 3 books by that 'Author'
+        // If an author is Anonymous, Unknown, or Various, we should allow 4 books by that 'Author'
```
cpeel, line 424: Can we reword this and the PM comment so that they are parallel? The PM talks about skips and this one talks about allowing. Having them parallel would make this easier to wrap one's head around.

Consider for here:
```
// Only allow 1 book by the author to be available in this round unless the "author" is 
// one of [Anonymous, Unknown, Various] in which case allow 4 books.
// If they have at least that many, skip their books.
```

And for the PM:
```
// Allow a PM to have 13 books in this round, if they have at least that many skip their books.
```

jmdyck: I'm okay with parallelizing, but I'm not sure about the "allow in the round" wording. E.g., it's possible for a PM to have, say, 15 projects in a round. This code doesn't say "that's not allowed!", it just says "I won't release any more for that PM".

I'll suggest a different wording...

cpeel: The new wording is much clearer for me, thank you. Do you want these three commits merged in as-is or are you planning on squashing some of them? I'm happy with either answer.

jmdyck: Keep them separate, I'd say.

PR #279, opened 2020-04-11
1850: Search-and-replace dies between pages

chrismiceli:
close opened toolbar popups when unloading the frame


chrismiceli:
Available for testing in my sandbox: https://www.pgdp.org/~chrismiceli/c/

PR #281, opened 2020-04-14
New method of removing overall directory in zip file

70ray:
applied to main zip and any zips inside it

### DIFF
--- a/pinc/misc.inc
+++ b/pinc/misc.inc
@@ -713,7 +713,7 @@ function list_files_in_zip(string $file_path): array
 // If the output directory does not exist or is not a valid directory, an InvalidArgumentException is thrown.
 //
 // Returns whether the extraction was successful.
-function extract_zip_to(string $path_to_zip, string $output_directory): bool
+function extract_zip_to(string $path_to_zip, string $output_directory, bool $remove_top_dir = false): bool
 {
     if (!is_valid_zip_file($path_to_zip)) {
         throw new InvalidArgumentException(sprintf(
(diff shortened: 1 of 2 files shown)


pinc/misc.inc
```diff
@@ -732,6 +732,30 @@ function extract_zip_to(string $path_to_zip, string $output_directory): bool
     // No need to check the result of the function since it is checked in is_valid_zip_file
     $zip->open($path_to_zip);
 
+    if($remove_top_dir)
```
cpeel, line 735: This logic alters the contents of the zip file which I would not expect from this function.

Instead, consider putting this logic in a separate function in this file that you can call before the extract, something like `remove_common_basedir_from_zip()`.

pinc/misc.inc
```diff
@@ -732,6 +732,30 @@ function extract_zip_to(string $path_to_zip, string $output_directory): bool
...
+    {
+        // find if there is a common prefix implying a common directory, remove it
+        $filenames = [];
+        $last_index = $zip->numFiles - 1;
+        for ($i = 0; $i <= $last_index; $i++)
+        {
+            $filenames[] = $zip->getNameIndex($i);
+        }
+        sort($filenames);
+        $prefix = $filenames[0];
```
cpeel, line 745: Since `$prefix` here is a base directory, `$basedir` would make more sense and help make the rest of the logic clearer.

pinc/misc.inc
```diff
@@ -732,6 +732,30 @@ function extract_zip_to(string $path_to_zip, string $output_directory): bool
...
+        if(endswith($prefix, "/") && startswith($filenames[$last_index], $prefix))
```
cpeel, line 746: Adding a comment above this line on what it is doing would help further devs. something like:
```
// If the first entry is a directory (ending in /) and the last entry
// starts with that directory, then all files in the archives are within
// the directory and we remove it from all files in the archive.
```

pinc/misc.inc
70ray: Good plan.

70ray:
Made these changes.

PR #283, opened 2020-04-20
Further WordCheck updates for Unicode

cpeel:
This implements two updates to WordCheck:
* Correctly parse words that contain combining characters. (oops)
* Detect if words contain a mix of Unicode script characters and if so, flag them as bad words.

I _really_ expected aspell to detect that second one, but no, it ignores them as "invalid words". Thanks aspell.

While I was here I broke out the diacritical markup flagging into its own function.

This is available in the [wc-multi-script](https://www.pgdp.org/~cpeel/c.branch/wc-multi-script) sandbox.

### DIFF
--- a/pinc/wordcheck_engine.inc
+++ b/pinc/wordcheck_engine.inc
@@ -15,7 +15,7 @@ $mark = '[=:.`\\\\\'/v)(~,^\\\\\*]'; // pattern for: diacritical mark
 $bracketed_character_pattern = "\\[(?:oe|OE|$mark$base|$base$mark)\\]";
 
 // This is used when splitting a text into words.
-$char_pattern = "(?:\w|$bracketed_character_pattern)";
+$char_pattern = "(?:\w\pM*|$bracketed_character_pattern)";
 $word_pattern = "!$char_pattern+(?:'$char_pattern+)*!uS";
 // (The pattern is delimited by exclamation marks rather than the usual slashes,
 // because slash is a character within the pattern.)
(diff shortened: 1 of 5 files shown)


pinc/wordcheck_engine.inc
```diff
@@ -393,6 +413,42 @@ function get_bad_words_via_pattern($input_words_w_freq, $languages)
...
+// Return a list of "bad" words where characters within a word are not all
+// of the same Unicode script type. For instance, this detects words that
+// contain a mix of Latin and Cyrillic characters.
+// Arguments:
+//   input_words_w_freq - an array whose keys are the distinct words of the input text
+//
+// Returns:
+//       -- an array of bad words
+//
+function get_bad_words_with_multi_scripts($input_words_w_freq)
```
chrismiceli, line 425: since you're not using the frequency value, would this make sense in unicode.inc or some other shared location? Or at least some portion of it?

cpeel: I did think about this. I struggled with:
1. What would a useful, abstracted interface of this look like? "given a string return the scripts that compose it"? Would that return an array of strings or IntlChar enums?
2. I suspect performance of this is pretty abysmal since we're iterating over every. single. character (at least until we early-exit) and I was wary about considering this a useful generic algorithm to build off of. For individual pages, where it is used most often, this is probably OK. For bigger texts we could probably get the script for the first letter, via regex remove all characters with that script, and iterate until we're done. Actually, that might be better anyway - I will ponder this.

Open to ideas on either one of those, particularly #1!

chrismiceli: Maybe just the unicode stuff could be in unicode.inc, like a function that takes a word and either null or the $unicode_script_enum. But not necessary if you don't see this concept being used anywhere else.

cpeel: I decided pulling two useful functions out into unicode.inc, and making it more efficient, was a good idea. Thank you for suggesting it.

cpeel:
Now with unit tests!

chrismiceli:
LGTM

cpeel:
No, I just need to merge it in and get sandboxes updated.

PR #284, opened 2020-04-22
1590: Wordcheck: Be able to reject a suggestion

chrismiceli:
add add words to bad word list
remember cutoff between POSTs

### DIFF
--- a/pinc/wordcheck_engine.inc
+++ b/pinc/wordcheck_engine.inc
@@ -10,7 +10,7 @@ include_once($relPath.'misc.inc'); // str_contains, array_get
 // NOTE: This should probably be changed when moving to UTF-8
 //       at that time we can probably move to using \W instead of
 //       the mess below
-$word_letters = "a-zA-ZÃ€-Ã–Ã˜-Ã¶Ã¸-Ã¿";
+$word_letters = "a-zA-ZÀ-ÖØ-öø-ÿ";
 $word_characters = "0-9$word_letters";
 
 // Handle dp-specific notation for non-Latin-1 characters
@@ -30,7 +30,7 @@ $word_pattern = "!$char_pattern+(?:'$char_pattern+)*!S";
 
 // string of punctuation characters
 // this is only used for punctuation highlighting
-$puncCharacters = '.,;:?!*/()#@%+=[]{}<>\"$|_Â¬Â¢Â£Â¥Â©Â®Â§Â°Â±Â¶Â·Â´Â¸ÂºÃ—Â¦Â¡Â¿-Â»Â«Â¯Ã·Â¹Â²Â³Â¼Â½Â¾Â¤';
+$puncCharacters = '.,;:?!*/()#@%+=[]{}<>\"$|_¬¢£¥©®§°±¶·´¸º×¦¡¿-»«¯÷¹²³¼½¾¤';
 
 // -----------------------------------------------------------------------------
 
@@ -793,7 +793,6 @@ function save_word_list($path, $words)
         // skip any blank words
         if ($word == "") continue;
 
-        var_dump($word);
         fputs($fh, "$word\r\n");
     }
     fclose($fh);
--- a/pinc/wordcheck_engine.inc
+++ b/pinc/wordcheck_engine.inc
@@ -10,7 +10,7 @@ include_once($relPath.'misc.inc'); // str_contains, array_get
 // NOTE: This should probably be changed when moving to UTF-8
 //       at that time we can probably move to using \W instead of
 //       the mess below
-$word_letters = "a-zA-Z�-��-��-�";
+$word_letters = "a-zA-ZÀ-ÖØ-öø-ÿ";
 $word_characters = "0-9$word_letters";
 
 // Handle dp-specific notation for non-Latin-1 characters
@@ -30,7 +30,7 @@ $word_pattern = "!$char_pattern+(?:'$char_pattern+)*!S";
 
 // string of punctuation characters
 // this is only used for punctuation highlighting
-$puncCharacters = '.,;:?!*/()#@%+=[]{}<>\"$|_��������������צ��-�����������';
+$puncCharacters = '.,;:?!*/()#@%+=[]{}<>\"$|_¬¢£¥©®§°±¶·´¸º×¦¡¿-»«¯÷¹²³¼½¾¤';
 
 // -----------------------------------------------------------------------------
 
--- a/pinc/wordcheck_engine.inc
+++ b/pinc/wordcheck_engine.inc
@@ -455,6 +455,12 @@ function save_project_good_words($projectid, $words)
     return save_word_list($fileObject->abs_path, $words);
 }
 
+function touch_project_good_words($projectid)
+{
+    $fileObject = get_project_word_file($projectid, "good");
+    touch($fileObject->abs_path);
+}
+
 // -----------------------------------------------------------------------------
 
 function load_project_bad_words($projectid)
@@ -776,6 +782,17 @@ function save_word_list($path, $words)
     // pull out the values to reset the numeric keys;
     // needed for the array == array comparison below
     $words = array_values($words);
+
+     // see if list has changed before saving it
+    $oldWords = load_word_list($path);
+
+    // if no change was made, return Success
+    // without modifying file
+    if($words == $oldWords)
+    {
+        return "Success";
+    }
+
 
     // now open the file
     $fh = @fopen($path, "w");
--- a/pinc/wordcheck_engine.inc
+++ b/pinc/wordcheck_engine.inc
@@ -10,7 +10,7 @@ include_once($relPath.'misc.inc'); // str_contains, array_get
 // NOTE: This should probably be changed when moving to UTF-8
 //       at that time we can probably move to using \W instead of
 //       the mess below
-$word_letters = "a-zA-ZÀ-ÖØ-öø-ÿ";
+$word_letters = "a-zA-Zï¿½-ï¿½ï¿½-ï¿½ï¿½-ï¿½";
 $word_characters = "0-9$word_letters";
 
 // Handle dp-specific notation for non-Latin-1 characters
@@ -30,7 +30,7 @@ $word_pattern = "!$char_pattern+(?:'$char_pattern+)*!S";
 
 // string of punctuation characters
 // this is only used for punctuation highlighting
-$puncCharacters = '.,;:?!*/()#@%+=[]{}<>\"$|_¬¢£¥©®§°±¶·´¸º×¦¡¿-»«¯÷¹²³¼½¾¤';
+$puncCharacters = '.,;:?!*/()#@%+=[]{}<>\"$|_ï¿½ï¿½ï¿½ï¿½ï¿½ï¿½ï¿½ï¿½ï¿½ï¿½ï¿½ï¿½ï¿½ï¿½×¦ï¿½ï¿½-ï¿½ï¿½ï¿½ï¿½ï¿½ï¿½ï¿½ï¿½ï¿½ï¿½ï¿½';
 
 // -----------------------------------------------------------------------------
 
@@ -777,17 +777,6 @@ function save_word_list($path, $words)
     // needed for the array == array comparison below
     $words = array_values($words);
 
-    // see if list has changed before saving it
-    $oldWords = load_word_list($path);
-
-    // if no change was made, return Success
-    // without modifying file
-    if($words == $oldWords)
-    {
-        return "Success";
-    }
-
-
     // now open the file
     $fh = @fopen($path, "w");
 
@@ -804,6 +793,7 @@ function save_word_list($path, $words)
         // skip any blank words
         if ($word == "") continue;
 
+        var_dump($word);
         fputs($fh, "$word\r\n");
     }
     fclose($fh);
--- a/tools/project_manager/show_all_good_word_suggestions.php
+++ b/tools/project_manager/show_all_good_word_suggestions.php
@@ -37,9 +37,14 @@ $frame = get_enumerated_param($_REQUEST, 'frame', 'master', array('master', 'lef
 
 if($frame=="update") {
     $newProjectWords=array();
+    $rejectProject = false;
     $destination_list = isset($_POST[BAD_WORD_LIST]) ? BAD_WORD_LIST : GOOD_WORD_LIST;
     foreach($_POST as $key => $val) {
-        if(preg_match("/cb_(projectID[0-9a-f]{13})_(\d+)/",$key,$matches)) {
+        if(preg_match("/" . REJECT_SUGGESTIONS . "_(projectID[0-9a-f]{13})/", $key, $matches))
+        {
+            $rejectProject = $matches[1];
+        }
+        elseif(preg_match("/cb_(projectID[0-9a-f]{13})_(\d+)/",$key,$matches)) {
             $projectid=$matches[1];
             $word=decode_word($val);
             if(!is_array(@$newProjectWords[$projectid]))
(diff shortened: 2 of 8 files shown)


chrismiceli:
Available in sandbox here:
https://www.pgdp.org/~chrismiceli/c/

Screenshot of new submit button:
image

cpeel:
What you have is good and certainly something that has been requested although I'm not 100% certain it addresses task 1590 as-described. We should probably do the same updates to `tools/project_manager/show_all_good_word_suggestions.php` so the two interfaces are the same (/me rants to self of why those are using different codepaths...).

I think this page could also use a "Reject all suggestions" button, placing it below the "Number of pages with suggestions: X" line. When clicked it would just re-save the good words list. That would reset the this-project-has-suggestions marker and is an oft-requested task. It too would need updating on `show_all_good_word_suggestions.php` for each project -- which would solve Task 1651. The "Reject all suggestions" changes could easily be a separate MR.

chrismiceli:
> 
> 
> What you have is good and certainly something that has been requested although I'm not 100% certain it addresses task 1590 as-described. We should probably do the same updates to `tools/project_manager/show_all_good_word_suggestions.php` so the two interfaces are the same (/me rants to self of why those are using different codepaths...).
> 
> I think this page could also use a "Reject all suggestions" button, placing it below the "Number of pages with suggestions: X" line. When clicked it would just re-save the good words list. That would reset the this-project-has-suggestions marker and is an oft-requested task. It too would need updating on `show_all_good_word_suggestions.php` for each project -- which would solve Task 1651. The "Reject all suggestions" changes could easily be a separate MR.

Great input. I pushed a second commit, but this is still work in progress I think to address all the input in regards to the word check pages. I appreciate the feedback as we work on getting this right.

pinc/wordcheck_engine.inc
```diff
@@ -10,7 +10,7 @@ include_once($relPath.'misc.inc'); // str_contains, array_get
 // NOTE: This should probably be changed when moving to UTF-8
 //       at that time we can probably move to using \W instead of
 //       the mess below
-$word_letters = "a-zA-Z�-��-��-�";
```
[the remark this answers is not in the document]
cpeel: I suspect your editor saved the file as UTF-8 instead of ISO-8859-1. We actually need the file to stay the latter until the Unicode branch merges (the file was already converted to UTF-8 in that branch).

tools/project_manager/show_good_word_suggestions.php
```diff
@@ -10,6 +10,7 @@
 
 require_login();
 
+define("REJECT_SUGGESTIONS", "reject_suggestions");
```
cpeel, line 13: I would probably put this in `word_freq_table.inc` with the other new defines.

tools/project_manager/show_good_word_suggestions.php
```diff
@@ -37,9 +38,27 @@
...
+        $words = load_project_good_words($projectid);
+        $words = array_merge($words,$postedWords);
+        save_project_good_words($projectid,$words);
+    }
+    else
+    {
+        var_dump("bad word list");
+        $words = load_project_bad_words($projectid);
+        $words = array_merge($words,$postedWords);
+        save_project_bad_words($projectid,$words);
```
cpeel, line 60: Consider moving this outside the if/else block, since it's common to both.

Also, please add spaces after the commas in the function calls (I realize this isn't consistent with the rest of the file).

tools/project_manager/show_good_word_suggestions.php
```diff
@@ -37,9 +38,27 @@
...
+    }
```
cpeel, line 61: Oh hey - here is a fun bug in the old file that is more evident with your code. We determine the last change for the good words file at the top of the file:
```php
$fileObject = get_project_word_file($projectid,"good");
$timeCutoff = get_integer_param($_REQUEST, 'timeCutoff', $fileObject->mod_time, 0, null);
```
But we save the good words list later, changing when it was last updated. This means that after the page is submitted, it will still show the old suggestions instead of the new ones based on the updated time. This is most evident when we "reject all". If we move the above two lines below this `if($format=="update") {` block that should fix it.

pinc/wordcheck_engine.inc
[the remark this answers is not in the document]
chrismiceli: Fixed it by changing visual studio code to save in iso-8859-1 for now until unicode comes over.

tools/project_manager/show_good_word_suggestions.php
chrismiceli: I refactored some, but still needed 2 conditionals for save_project_bad vs save_project_good and load_project_bad vs load_project_good.

tools/project_manager/show_good_word_suggestions.php
chrismiceli: I noticed that and was planning on investigating. This makes perfect sense. Updated.

pinc/wordcheck_engine.inc
```diff
@@ -777,17 +777,6 @@ function save_word_list($path, $words)
     // needed for the array == array comparison below
     $words = array_values($words);
 
-    // see if list has changed before saving it
```
cpeel, line 780: Instead of taking this out and doing a load/save, let's add a new function to this file that uses `touch()` to update the good word list file. That does less data shuffling and creates a single function we can use where we want to reject suggestions. Perhaps call it simply `touch_project_good_words_file($projectid)`.

tools/project_manager/show_good_word_suggestions.php
```diff
@@ -35,15 +27,33 @@
...
 
+$fileObject = get_project_word_file($projectid,"good");
+$timeCutoff = get_integer_param($_REQUEST, 'timeCutoff', $fileObject->mod_time, 0, null);
+
+if($timeCutoff==0)
+    $time_cutoff_text = _("<b>All proofreader suggestions</b> are included in the results.");
+else
+    $time_cutoff_text = sprintf(_("Only proofreader suggestions made <b>after %s</b> are included in the results."),strftime($datetime_format,$timeCutoff));
+
 list($all_suggestions_w_freq,$all_suggestions_w_occurrences,$round_suggestions_w_freq,$round_suggestions_w_occurrences,$rounds,$round_page_count,$messages) =
```
cpeel, line 57: Here's another fun edge case. If the user uses the new "add to bad list" function to add all suggestions to the bad words list, when the page loads it will say "There are no suggestions in the given time frame that aren't already on the Good Words List." -- which is 100% correct. However, the timestamp of the Good Words List file won't be reset, so the alert on the PM page will still show up. The ideal solution would be to have the alert on the PM listing use the same logic here to determine if there really are good words to show, but that's rather resource intensive. Instead, we could:
```
if($is_reject && $timeCutoff == $fileObject->mod_time && count($all_suggestions_w_occurrences) == 0)
{
    // touch good word list function call
}
```

But trying to implement that in the other file is going to be complicated, and really the whole thing is a mess (which is all my fault 13 years ago).

So for now let's not scope creep the hell out of this task and live with the problem for now.

tools/project_manager/word_freq_table.inc
```diff
@@ -51,6 +55,11 @@ function get_cutoff_script($cutoffOptions,$instances) {
         }
         if(document.getElementById("freqCutoffValue")) document.getElementById("freqCutoffValue").value=newCutoff;
         document.getElementById("current_cutoff").innerHTML=newCutoff;
+
+        // set new value for forms so the value is persisted through actions
+        $('form input[name=freqCutoff').each(function (index, value) {
```
cpeel, line 60: Should this have a `]` in it? i.e. `$('form input[name=freqCutoff]')`?

pinc/wordcheck_engine.inc
chrismiceli: great idea. I thought of it just after I submitted the last change realizing it is such a waste otherwise.

tools/project_manager/word_freq_table.inc
chrismiceli: Yes, I'm actually very surprised it works without it.

tools/project_manager/show_good_word_suggestions.php
chrismiceli: Wouldn't the touch get called during the is_reject update processing?

cpeel: Not in the case where they add the last word to the bad words list. It would if they rejected all the suggestions,

chrismiceli: I see, I was getting confused between "add to bad list" and reject all suggestions.

tools/project_manager/show_good_word_suggestions.php
```diff
@@ -15,38 +15,58 @@
...
     _get_word_list($projectid,$timeCutoff);
 
+if ($destination_list == BAD_WORD_LIST && $timeCutoff == $fileObject->mod_time && $all_suggestions_w_occurrences == 0)
```
cpeel, line 65: You're missing a `count()` on the array:
```
if ($destination_list == BAD_WORD_LIST && $timeCutoff == $fileObject->mod_time && count($all_suggestions_w_occurrences) == 0)
```

cpeel: The following diff will do effectively the same thing for the `show_all_good_word_suggestions.php` file:
```diff
diff --git a/tools/project_manager/show_all_good_word_suggestions.php b/tools/project_manager/show_all_good_word_suggestions.php
index 34fc53f..db5d9e4 100644
--- a/tools/project_manager/show_all_good_word_suggestions.php
+++ b/tools/project_manager/show_all_good_word_suggestions.php
@@ -186,7 +186,17 @@ echo "<br>";

         // if no words are returned (probably because something was
         // suggested but is no longer in the text) skip this project
-        if(count($suggestions_w_freq)==0) continue;
+        if(count($suggestions_w_freq)==0)
+        {
+            // If the time cutoff is the same as the good words file and
+            // we still have no suggestions, touch the good words file to
+            // reset the time so no alert appears for this project.
+            if($timeCutoffActual == $goodFileObject->mod_time)
+            {
+                touch_project_good_words($projectid);
+            }
+            continue;
+        }

         $projectsNeedingAttention++;
```

pinc/wordcheck_engine.inc
```diff
@@ -777,7 +783,7 @@ function save_word_list($path, $words)
...
+     // see if list has changed before saving it
```
cpeel, line 786: Minor: extra space here (I assume from putting this back in).

tools/project_manager/show_all_good_word_suggestions.php
```diff
@@ -79,7 +99,9 @@
     // get all projects for this PM
     $projects = _get_projects_for_pm($pm);
 
+    $rejectAlllabel = _("Reject all suggestions");
     $submitLabel = _("Add selected words to Good Words List");
+    $rejectLabel = _("Add selected words to Bad Words List");
```
cpeel, line 104: Consider changing this to `$submitBadLabel` instead, since putting them on the bad list isn't rejecting them.

tools/project_manager/show_good_word_suggestions.php
```diff
@@ -151,6 +171,7 @@
 echo_cutoff_text( $initialFreq,$cutoffOptions );
 
 $submit_label=_("Add selected words to Good Words List");
+$reject_label = _("Add selected words to Bad Words List");
```
cpeel, line 174: Suggest changing this to `$submit_bad_label` instead.

tools/project_manager/show_project_possible_bad_words.php
```diff
@@ -107,11 +107,11 @@
 $checkbox_form["freqCutoff"]=$freqCutoff;
 echo_checkbox_selects(count($bad_words_w_freq));
 echo_checkbox_form_start($checkbox_form);
-echo_checkbox_form_submit(_("Add selected words to Bad Words List"));
+echo_checkbox_form_submit(_("Add selected words to Bad Words List"), BAD_WORD_LIST);
```
srjfoo, line 110: The possibility exists that a PM might want to add a word to the GWL that is identified as a possible bad word. If I looked at the following, I'd just as soon add "tip" to the GWL and not bother the proofers with worrying about it.

image

and 

image

I realize that the task really deals only with the GWL, and this could be considered scope creep -- should it be a different task?

cpeel: Yes, that's scope creep and we should punt that. I've already scope-creeped this a bit much already for poor Chris.

tools/project_manager/show_all_good_word_suggestions.php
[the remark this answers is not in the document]
cpeel: > Am I correct that "Reject all suggestions" is just a means of clearing the flag so that a PM knows they've finished with this word list? That is, the words that were rejected wholesale are not on a list to be used as a filter for future suggestions and never seen again?

Correct.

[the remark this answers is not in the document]
srjfoo: Thanks for the confirmation. Either way, it needs to be documented exactly how it works, so there's no misunderstanding. Possibly as part of the recent additions announcement.

tools/project_manager/show_project_possible_bad_words.php
srjfoo: <grin> Understood.

It should be documented, then, and perhaps a new task created after it's rolled out. I could always be wrong, but this strikes me as the sort of inconsistency that will be noticed and questioned.

tools/project_manager/show_all_good_word_suggestions.php
```diff
@@ -11,6 +11,9 @@
 
 require_login();
 
+define("GOOD_WORD_LIST", "good_word_list");
+define("BAD_WORD_LIST", "bad_word_list");
```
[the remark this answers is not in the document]
chrismiceli: some of the code has it plural, others don't. I must've most recently seen one singular and went with that. I'll update.

tools/project_manager/show_project_possible_bad_words.php
chrismiceli: Create the ticket. This shouldn't be hard to add. Just trying to not go past the upcoming code freeze.

cpeel:
Right now, if a PM rejects words we still show the words they rejected in the form because the `timeCutoff` submitted with the form is the old GWL save time. This patch fixes that such that if the user is looking at the "since last saved time", they will be after submitting the form:
```patch
diff --git a/tools/project_manager/show_good_word_suggestions.php b/tools/project_manager/show_good_word_suggestions.php
index c6195bc..bba5bc2 100644
--- a/tools/project_manager/show_good_word_suggestions.php
+++ b/tools/project_manager/show_good_word_suggestions.php
@@ -175,7 +175,7 @@ $submit_bad_label = _("Add selected words to Bad Words List");

 $checkbox_form["projectid"]=$projectid;
 $checkbox_form["freqCutoff"]=$freqCutoff;
-$checkbox_form["timeCutoff"]=$timeCutoff;
+$checkbox_form["timeCutoff"]=$timeCutoff == $fileObject->mod_time ? "" : $timeCutoff;
 echo_checkbox_form_start($checkbox_form);

 $context_array["[[TITLE]]"]=_("Show Context");
diff --git a/tools/project_manager/word_freq_table.inc b/tools/project_manager/word_freq_table.inc
index 3782071..6dd9221 100644
--- a/tools/project_manager/word_freq_table.inc
+++ b/tools/project_manager/word_freq_table.inc
@@ -423,7 +423,8 @@ function echo_checkbox_form_start($checkbox_form) {
     echo "<form action='" . attr_safe($_SERVER['PHP_SELF']) . "' method='post'>";
     echo "<input type='hidden' name='format' value='update'>";
     foreach($checkbox_form as $name => $value) {
-        echo "<input type='hidden' name='$name' value='$value'>";
+        if($value != '')
+            echo "<input type='hidden' name='$name' value='$value'>";
     }
 }
```

chrismiceli:
Sandbox updated to latest commit

cpeel:
This works great in all of the variations that I tested!

PR #286, opened 2020-04-27
Ie mru fix

70ray:
IE doesn't support findIndex(), so use another method.


tools/proofers/character_selector.js
```diff
@@ -69,9 +69,14 @@ $(function () {
     }
 
     function addMru(char, titletext) {
-        var index = mru.findIndex(function (element) {
```
cpeel, line 72: Let's add this block back in, but commented out, and a comment line above it:
```
// IE HACK - IE11 doesn't support .findIndex()
```

70ray: Done this

PR #288, opened 2020-04-27
improve performance internet explorer

70ray:
IE doesn't implement "u" flag in regular expressions. We don't need it for the validCharacterPattern if we're not using astral plane characters.
IE also doesn't do normalise. This means in IE the square bracket diacritical conversion won't work (but ligature conversion will). Also the character validation will not normalise, so if for example you put in an Angstrom symbol it would normally get normalized to A with ring above which could be a valid character but this will not happen.
This fix just avoids javascript errors.

### DIFF
--- a/scripts/character_test.js
+++ b/scripts/character_test.js
@@ -13,7 +13,9 @@ var goodChar;
 $(function () {
     // need to define this after the page has loaded so validCharacterPattern
     // is available
-    goodChar = RegExp(validCharacterPattern, "u");
+    // need "u" flag if using astral plane but IE doesn't handle it
+    goodChar = RegExp(validCharacterPattern);
+    console.log(goodChar);
 });
 
 function testChar(character) {
@@ -22,7 +24,9 @@ function testChar(character) {
 
 // return false if text contains any bad characters
 function testText(text) {
-    text = text.normalize("NFC");
+    if("normalize" in String.prototype) {
+        text = text.normalize("NFC");
+    }
     let result;
     charMatch.lastIndex = 0;
     while(null != (result = charMatch.exec(text))) {
--- a/scripts/character_test.js
+++ b/scripts/character_test.js
@@ -13,18 +13,21 @@ var goodChar;
 $(function () {
     // need to define this after the page has loaded so validCharacterPattern
     // is available
-    // need "u" flag if using astral plane but IE doesn't handle it
-    goodChar = RegExp(validCharacterPattern);
-    console.log(goodChar);
+    goodChar = XRegExp(validCharacterPattern, "A");
 });
 
 function testChar(character) {
+    if(!goodChar) {
+        // IE HACK - IE sometimes says goodChar is undefined
+        goodChar = XRegExp(validCharacterPattern, "A");
+    }
     return goodChar.test(character);
 }
 
 // return false if text contains any bad characters
 function testText(text) {
-    if("normalize" in String.prototype) {
+    // IE HACK - IE11 does not support string normalization
+    if(String.prototype.normalize) {
         text = text.normalize("NFC");
     }
     let result;
--- a/scripts/text_validator.js
+++ b/scripts/text_validator.js
@@ -47,7 +47,9 @@ $(function () {
 
     function _validateText() {
         var text = textArea.value;
-        text = text.normalize("NFC");
+        if("normalize" in String.prototype) {
+            text = text.normalize("NFC");
+        }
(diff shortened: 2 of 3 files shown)


scripts/character_test.js
```diff
@@ -13,7 +13,9 @@ var goodChar;
 $(function () {
     // need to define this after the page has loaded so validCharacterPattern
     // is available
-    goodChar = RegExp(validCharacterPattern, "u");
```
chrismiceli, line 16: Not sure how long we will keep having to support IE, but is it worth looking into using babel to help alleviate these kinds of issues:
https://babeljs.io/docs/en/babel-plugin-transform-es2015-unicode-regex
We could transform similar to our less tool since we don't have a full build pipeline.

70ray: Thanks, I'll look into it.

cpeel: IE 11 is the only available Microsoft browser for Windows 7 and Windows 8. Windows 7 went out of extended support in Jan 2020, so we can happily put that OS behind us. Windows 8 goes out of extended support in Jan 2023 -- so it will continue getting OS updates until that time. While I would love to deep-6 IE11 support, I don't feel that we can do that without a little more heads-up to our community.

All that to say that we should continue some level of IE11 support for the Unicode roll-out and then start messaging to our community (both at pgdp.net and  a heads-up to our other code users) that we are dropping IE11 support at a specific time.

cpeel: Can't we use XRegEx here? That would be a more forward-looking approach and we use it elsewhere.

scripts/character_test.js
```diff
@@ -13,7 +13,9 @@ var goodChar;
...
+    // need "u" flag if using astral plane but IE doesn't handle it
+    goodChar = RegExp(validCharacterPattern);
+    console.log(goodChar);
```
cpeel, line 18: Don't forget to remove this debug logging.

scripts/character_test.js
```diff
@@ -22,7 +24,9 @@ function testChar(character) {
 
 // return false if text contains any bad characters
 function testText(text) {
-    text = text.normalize("NFC");
+    if("normalize" in String.prototype) {
```
cpeel, line 27: I've seen `if(String.prototype.normalize)` elsewhere on the internet for this check -- which seems a bit more straightforward.

For all of these, please add a comment `// IE HACK - IE11 does not support string normalization` above them so we can go in and remove them when we eventually drop IE11 support.

scripts/character_test.js
chrismiceli: exactly, babel will allow us to use modern javascript with less IE issues. Also, for what it is worth, the new chromium edge is available for windows 7, 8, and 10 as well as mac and soon linux:
https://en.wikipedia.org/wiki/Microsoft_Edge#Anaheim_(2019%E2%80%93present)

cpeel: We are way too close to release date to bring in more middleware. I'm happy to consider that for after the roll-out, but I'd just rather drop IE11 in 6 months instead.

70ray:
I ran into another problem when I tried XRegExp at first but perhaps gave up too quickly. I'll have another look at it.

70ray:
I ran into another problem when I tried XRegExp at first but perhaps gave up too quickly. I'll have another look at it.

70ray:
Using XRegExp for this now. A console message would sometimes appear in IE saying that goodChar was undefined. I couldn't track down why this happened but put in a fix to define it again if it should occur.

chrismiceli:
> 
> 
> Using XRegExp for this now. When starting wordcheck a console message would sometimes appear in IE saying that goodChar was undefined. I couldn't track down why this happened but put in a fix to define it again if it should occur.

Instead of the way you have it here, why not always leave it undefined and initialize it if undefined in testChar always. Sort of like lazy initialization. That way it is not a hack and optimized if testChar is never called.

cpeel:
> Instead of the way you have it here, why not always leave it undefined and initialize it if undefined in testChar always. Sort of like lazy initialization. That way it is not a hack and optimized if testChar is never called.

I don't know how variable scoping works in JS. If we define it here like this does it make it local or global? Because `testChar()` will be called a lot and ideally we want to instantiate `goodChar` only once, which is why it's suppose to be created after document load (unless you're IE, I guess).

cpeel:
And to be clear: I'm fine with this code as-is. It's not ideal but neither is IE11.

PR #293, opened 2020-05-03
1874: Expand ability to add words to Bad Words List or Good Words List

chrismiceli:
add options for stealth scannos and currently flagged words to add to bad words list

### DIFF
--- a/tools/project_manager/show_current_flagged_words.php
+++ b/tools/project_manager/show_current_flagged_words.php
@@ -25,11 +25,20 @@ enforce_edit_authorization($projectid);
 $format = get_enumerated_param($_REQUEST, 'format', 'html', array('html', 'file', 'update'));
 
 if($format=="update") {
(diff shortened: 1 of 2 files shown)


chrismiceli:
Available in my sandbox: https://www.pgdp.org/~chrismiceli/c

cpeel:
Thanks Chris. I still need to actually review this, but note that we won't be getting it merged until after the Unicode roll-out. I expect it will land without merge conflicts though.

cpeel:
We rolled back the original commit based on issues that were found. Given what the prior commit revealed (ie. poorly-understood use cases by the dev team, including myself), I am going to close this MR and suggest we revisit the overall WordCheck PM UX as a whole rather than patch this functionality in.

I appreciate you working on it, Chris. I'm sad that the code you're building on top of sucks so much.

srjfoo:
I would deny that it sucks all that badly! It was a badly needed tool when it was rolled out, and has been used extensively over the years. As time goes on, we always understand things better. Spell checking has been a lot better at DP for all these years because of this tool!

PR #294, opened 2020-05-03
Move "Reject all suggestions" above round info

cpeel:
Move the "Reject all suggestions" button above the round data on both management pages to reinforce that the button applies to all rounds.

Done based on feedback from PMs.


cpeel:
Instead, we rolled back the original commit to address some other issues. Closing this request.

PR #298, opened 2020-05-11
Construct displayimage.php etc. using javascript.

70ray:
displaymage.php has options to display image, text or text+image. (text is read-only)
mentors/view_page_text_image.php displays image and modifiable text. This is kept separate so we can add more features later.
Running in [https://www.pgdp.org/~rp31/pm](https://www.pgdp.org/~rp31/pm) sandbox


scripts/page_browse.js
```diff
@@ -0,0 +1,212 @@
...
+    });
+
+    hSwitchButton.click(function () {
+        changeSplit(false);
+    });
+
+    mainSplit = initSplit({paneContainer: container, verticalSplit: vSplit, splitPercent: splitPercent});
+    setSplitControls(vSplit);
+
+    $(window).resize(mainSplit.reLayout);
```
cpeel, line 83: The splitter should register this event callback itself.

scripts/splitControl.js
```diff
@@ -0,0 +1,129 @@
+/*global $ */
+/* exported initSplit */
+
+function initSplit(newData) {
+    // use these defaults if any newData not set
+    let data = {paneContainer: "#window", verticalSplit: false, splitPercent: 50, minSiz0: 50, minSiz1: 50};
```
cpeel, line 6: paneContainer should be a required argument, not an optional one, so the function definition should look more like:
```
function initSplit(container, config) {
```

scripts/splitControl.js
```diff
@@ -0,0 +1,129 @@
...
+    for(let key in newData) {
+        data[key] = newData[key];
+    }
+    // paneContainer is the id of a div which will contain three divs:
+    // referred to in this function as pane1, dragBar and pane2.
+    // More split views can be set up within the panes.
+    // minSiz0, minSiz1 minimum allowed size of panes top/left and bottom/right
```
cpeel, line 13: `minSizePane1` and `minSizePane2` would be clearer names and also map to how you refer to them later.

scripts/splitControl.js
cpeel: I think it would be more intuitive if `verticalSplit` was `splitDirection` and one of `horizontal`/`vertical`.

cpeel: This should also take an optional ID for each pane and set it when creating the `<div>`s to allow programmatic access to them later. Actually, that may not work quite right as other JS (and possibly the PHP) will need to interact with these divs and we might get into a race condition on when the IDs were defined.

How do other splitters handle this? Do they require the `<div>`s to be defined already and just pass in the IDs?

scripts/page_browse.js
70ray: > Focused this review primarily on the splitter.

Thanks Casey.

What we really want to do is to re-layout the splitter when the contaner resizes. This would cover both the case of the window resizing and for a subsidiary splitter (like the one in the mentors page) needing to be re-layed out when the splitter bar in the containing splitter is moved (which is now done by a callback). It would be possible to do this using https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver. But this is not implemented in IE. I'll try coding it.

scripts/splitControl.js
70ray: > paneContainer should be a required argument, not an optional one, so the function definition should look more like:
> 
> ```
> function initSplit(container, config) {
> ```

I agree.


> I think it would be more intuitive if `verticalSplit` was `splitDirection` and one of `horizontal`/`vertical`.

Yes, I'm afraid my desire to do things most efficiently from the time when memory and speed were only a tiny fraction of what they are now
`if(verticalSplit)` rather than `if(splitDirection === `vertical`)

70ray: > This should also take an optional ID for each pane and set it when creating the `<div>`s to allow programmatic access to them later. Actually, that may not work quite right as other JS (and possibly the PHP) will need to interact with these divs and we might get into a race condition on when the IDs were defined.
> 
> How do other splitters handle this? Do they require the `<div>`s to be defined already and just pass in the IDs?

The previous version of this, (until a few days ago) needed the divs to be pre-defined and identified them as children of the container. This might be a better way to do it. It might be best to still create the splitter-bar div inside the splitter code but possibly not.
It would be easy to configure the width and colour of the splitter bar and it would be nice to have an image in it (like the three dots in your wire-frames). I'll look at doing this. And look at other splitters again.

scripts/splitControl.js
70ray: Yes.

scripts/splitControl.js
cpeel: > The previous version of this, (until a few days ago) needed the divs to be pre-defined and identified them as children of the container. This might be a better way to do it. It might be best to still create the splitter-bar div inside the splitter code but possibly not.

I do like the idea of saying "to use this, pass in the ID of a `<div>` that contains **exactly two other immediate-children** `<div>`s, eg: for `<div id='container'><div></div><div></div></div>` pass in 'container'". Then the code will add the splitter bar between the two child divs and off we go. But this is also me thinking out loud a bit of how I would want to use it as a developer.

cpeel:
I was remiss in not saying that the UX for what you have is really awesome!!

scripts/page_browse.js
70ray: Changed to adding another parameter for relayout. Not sure if it's an improvement.

scripts/splitControl.js
70ray: changed to using two pre-made divs.

srjfoo:
I really like the smooth operation of the resizing of the image pane in the image/text mode.

One thing that jumped out at me, though -- if I want to resize the image, I _*must*_ use the `Resize` button for it to take effect. If I'm typing in the percentage value, I would expect to be able to press the `Return` key and have it take effect, but if I do that currently, I'm sent into Image-only mode, and if I press it again, I'm sent into text only mode.

This seems to happen only if the cursor is in the text box that allows user input on the image size (i.e. I tried using the arrows to adjust the image size percentage, but `Return` did nothing at all unless I placed the cursor inside the input box.

If I use the tab key to get to the `Resize` button, and then press `Return`, it works as it should.

I think that's the only place on the page that allows for keyed-in input. Is there a way to force `Return` to apply the setting if the cursor is in that field, or to disable it altogether there? What concerns me is the unexpected results I currently get when I press`Return` while the cursor is in the box. (Note -- this is on Firefox on a Mac. I have not yet tried Chrome or Safari.)

cpeel:
I would propose deferring additional work on the image manipulation bits, until we iron out some of the Page Image Widget work. Ideally we would be able to plug that into this code and have the zoom controls handled by it instead.

srjfoo:
> I would propose deferring additional work on the image manipulation bits, until we iron out some of the Page Image Widget work. Ideally we would be able to plug that into this code and have the zoom controls handled by it instead.

That's entirely fair. What I outlined was just something that turned up in testing that didn't feel quite right.

scripts/splitControl.js
```diff
@@ -0,0 +1,132 @@
...
+    var dragEnd = $.Callbacks();
+
+    function moveSplit() {
+        if (splitPos < minPos) {
+            splitPos = minPos;
+        }
+        if (splitPos > maxPos) {
+            splitPos = maxPos;
+        }
+        var sp6 = splitPos + 6;
```
cpeel, line 44: What is `sp6` and what are the 6s for in this function?

scripts/splitControl.js
```diff
@@ -0,0 +1,132 @@
+/*global $ */
```
cpeel, line 1: This splitter is something that I want to start building with in multiple places in the site. I would prefer we take this file out into its own MR so we can iterate on it and get it in as a widget we can use with this MR and others later.

To that end, it would be good to get some simple, manual tests in `SETUP/tests/manual_web` we can use to test it, both as part of its development and later as we make changes to it. I'm envisioning at least three test files:
1. Two pane vertical split
2. Two pane horizontal split
3. Two pane vertical split with the right pane also split horizontally.

scripts/splitControl.js
```diff
@@ -0,0 +1,132 @@
...
+/* exported initSplit */
+
+function initSplit(container, reDraw, config) {
```
cpeel, line 4: `splitControl` would perhaps be a better name for this control as reflected in the filename.

Please move the parts of the comment within the function outside of it and make it clearer what the arguments are and what the function returns, eg:
```js
/*
Create a splitter between two <div>s within a container.
Arguments:
* container - ID of a <div> which contains two <div>s (herein referred
  to as pane1 and pane2). The splitter will be created between these.
* reDraw - jQuery callback which is fired when the container sizes change.
* config - dictionary that controls the div:
  {
    splitDirection: "horizontal" or "vertical",
    splitPercent: 50, // what is this?
    minSizePane1: 50,  // describe this
    minSizePane2: 50, // describe this
  }
Returns:
???
*/
function splitControl(container, reDraw, config) {
```

scripts/splitControl.js
70ray: The 6 is the width/height of the dragbar. It should be configurable.

scripts/splitControl.js
70ray: Thanks Casey. I'll do that.

scripts/splitControl.js
70ray: I'll do that.

cpeel:
@70ray - now that we have splitControl merged, can you rebase this branch off of master?

70ray:
> @70ray - now that we have splitControl merged, can you rebase this branch off of master?

@cpeel - Ok , I'll do that.

70ray:
> @70ray - now that we have splitControl merged, can you rebase this branch off of master?

@cpeel - Ok , I'll do that.

Pushing a new branch and closing this one.

PR #302, opened 2020-05-19
Unicode Support

cpeel:
Convert the site to full Unicode support.

From the `SETUP/UNICODE.md`:
```
The DP codebase fully supports UTF-8, in the HTML rendered to the user as well
as in page texts.

Rather than open up the floodgates and allow proofreaders to input any of
the millions of Unicode character into the proofreading interface, we allow
project managers to restrict what Unicode characters are valid for project
pages. This ensures that proofreaders don't insert characters that don't
rightfully belong in the project but look similar to those that do (Kappa
for a 'K' as an example).

Both the UI and the server-side normalization enforce that only supported
characters are used for project pages. This allows the code to support the full
set of Unicode characters but reduces the possibility of invalid characters in
a given project.

Project managers select which characters are valid for a project by associating
it with a Character Suite when creating or editing the project.

Character suites are only used to enforce project page texts. Project metadata,
such as title and author, are not restricted and will accept any valid Unicode
character. It's recommended not to use Unicode characters that are encoded as 4
bytes in UTF-8 in project titles or authors. These fields are used when creating
project discussion threads and phpBB doesn't support UTF-8 characters encoded
with more than 3 bytes. Using such characters will result in an error when a
user attempts to create a discussion thread.
```

This is a reference MR for later github spelunkers wanting to know when this branch merged in. This is not intended for a review and will be merged in within the hour.

Reviewers are welcome to "approve" this review for posterity if they wish 😁


lhamilton1:
Approved

PR #303, opened 2020-05-19
Filename validation during uploads

cpeel:
These are updates that came out of some investigation on why a user's SR uploads were giving invalid filename errors (and yet were still being uploaded successfully).

This does 3 things:
1. Validates the filename server-side before doing virus checks.
2. Prevents files from being uploaded that fail the client-side validation.
3. Shows the filename that failed validation (helps when a user's OS is hiding extensions so `file.zip` is actually `file.zip.zip` and is failing).

Live in [filename-upload-validation](https://www.pgdp.org/~cpeel/c.branch/filename-upload-validation) sandbox.


scripts/file_resume.js
```diff
@@ -72,10 +72,7 @@ $(function() {
             // a file has been selected
             resumable.upload();
         } else {
-            // no file selected, the resumable submit button will do nothing
-            // so submit the old form in 'upload' mode, with no file
-            showProgress(uploadMessages.working);
-            $("#upload_form").submit();
+            alert(uploadMessages.noFile);
```
70ray, line 75: For returning a project to PP pool it is ok to not upload a file so we need to allow for this.
If there should be a file this will get caught at the server afterwards.

cpeel: Yes, thanks. Sharon caught this too and is testing a fix that accounts for this.

cpeel: New code update adds a second submit button for the forms that can be submitted without a file. This ensures that all file uploads actually come with uploads.

pinc/upload_file.inc
```diff
@@ -279,20 +288,40 @@ function show_upload_message($verbose, $message)
...
+function is_valid_dirname($dirname)
+{
+    // Directory names can't be longer than 200 characters total
+    if(strlen($dirname) > 200)
+    {
+        return false;
     }
 
+    // We allow filenames to start with an alphanumeric, followed by 0 more
+    // alphanumerics, dashes, and periods.
```
srjfoo, line 324: This is only a suggestion -- when I first saw the regex, I had to go look up `\w` to discover that the underscore is part of that character class. I don't think of the term "alphanumerics" as including the underscore, though. The current comment also says "periods", plural, which is incorrect.

```
// ... followed by 0 or more
// alphanumerics, underscores, hyphens and one dot separating
// the filename from the extension"
```

cpeel: I'll update to include underscores. For directories it is one or more periods, however.

PR #304, opened 2020-05-20
Revise character checks.

70ray:
Revise what characters should not appear before or after inline tags.

### DIFF
--- a/tools/proofers/preview.js
+++ b/tools/proofers/preview.js
@@ -59,7 +59,8 @@ var makePreview = function (txt, viewMode, wrapMode, styler) {
 
     // "A" means can use astral plane
     let puncChar = XRegExp("\\pP", "A");
-    let wordChar = XRegExp("\\pL", "A");
+    // wordChar is letter or number
+    let wordChar = XRegExp("\\pL|\\pN", "A");
 
     // ILTags can have "u" for underline added. Used for constructing regexes
     var ILTags = "[ibfg]|sc";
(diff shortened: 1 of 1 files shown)


70ray:
We should make some tests to exercise the format preview and check that any changes do not break anything.

cpeel:
> We should make some tests to exercise the format preview and check that any changes do not break anything.

Agreed. I've created [Task 1890](https://www.pgdp.net/c/tasks.php?action=show&task_id=1890) for this, but I don't think it blocks fixing this code.

cpeel:
Code-wise this looks good but I would love for @srjfoo to test it if she has time.

tools/proofers/preview.js
```diff
@@ -424,7 +425,10 @@ var makePreview = function (txt, viewMode, wrapMode, styler) {
                 if (postChar === "\n") {
                     reportIssue(start, tagLen, "nlAfterStart");
                 }
-                if (/\S/.test(preChar)) { // non-space before start tag
+
+                // disallow letter or ,.;: before inline start tag
+                let disallowBeforeStart = XRegExp("\\pL|[,.;:]", "A");
```
srjfoo, line 430: Just making sure that I'm understanding correctly that letters and `,.;:` cannot precede an inline start tag, but numbers are okay? So, something like `6<i>s</i>` would be okay, but `i<i>magine</i>` would not?

Should `!?` be included in that prohibition?

70ray: This change is mostly reverting the change I made for unicode with some other changes, one of which is to let numbers be ok in this siuation.
I think it is not really a good idea to change to different features at the same time so I am proposing to close this MR and instead make a new MR which just reverts the previous one. Then we can make any changes for unicode or other things later.

70ray:
Making a new MR which just reverts the earlier one.

PR #306, opened 2020-05-22
Revert "Handle non-Latin punc & chars in format preview (#289)"

70ray:
This reverts commit a3c117b73e450a6f16e04efbea8bb48483a7eef4.


70ray:
Since javascript works in 16-bit unicode changing the site to unicode won't make any difference to the format preview function.
We can make some changes later to allow for more characters. For example, in checking what characters should not appear before an in-line start tag the character class 'w' matches only A-Za-z0-9_ so will not flag À, but it didn't do this before.

PR #307, opened 2020-05-22
remove common sub-directory always

70ray:
It does not itself appear in a zip made in windows but does in
Linux and macOS


70ray:
Running in https://www.pgdp.org/~rp31/preview/

PR #308, opened 2020-05-22
1844: No truncation for GWL context list or option to see ALL if wanted

chrismiceli:
Add ability to change results count without manually modifying query param

### DIFF
--- a/tools/project_manager/show_word_context.php
+++ b/tools/project_manager/show_word_context.php
@@ -81,19 +81,17 @@ if($frame=="left") {
     echo "</a>";
     echo "</p>";
 
-    echo "<form method='GET'>";
+    echo "<form method='GET' id='wordInstancesForm'>";
     echo "<input type='hidden' name='projectid' value='$projectid' />";
     echo "<input type='hidden' name='word' value='$encWord' />";
     echo "<input type='hidden' name='layout' value='$layout' />";
     echo "<input type='hidden' name='frame' value='left' />";
     echo "<label for='wordInstancesSelect'>" . _("Number of word context results: ") . "</label>";
-    echo "<select id='wordInstancesSelect' name='wordInstances' style='margin-left: 2px;'>";
+    echo "<select id='wordInstancesSelect' name='wordInstances' style='margin-left: 2px;' onchange='$(\"#wordInstancesForm\").submit()'>";
     foreach(range(10, 100, 10) as $option) {
         echo "<option value='$option'" . ($option == $wordInstances ? "selected" : "") . ">$option</option>";
     }
     echo "</select>";
-    echo "<br />";
-    echo "<input type='submit' value='" . _("Update") . "' />";
     echo "</form>";
 
 
--- a/tools/project_manager/show_word_context.php
+++ b/tools/project_manager/show_word_context.php
@@ -16,6 +16,7 @@ require_login();
 
 define("LAYOUT_HORIZ", "horizontal");
 define("LAYOUT_VERT",  "vertical");
+define("MAX_WORD_INSTANCES", 100);
 
 set_time_limit(0); // no time limit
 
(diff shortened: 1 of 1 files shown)


chrismiceli:
image
sandbox is updated https://www.pgdp.org/~chrismiceli/c/ but not working because I need to run some scripts for unicode @cpeel

tools/project_manager/show_word_context.php
```diff
@@ -81,6 +81,22 @@
...
+    echo "<input type='hidden' name='layout' value='$layout' />";
+    echo "<input type='hidden' name='frame' value='left' />";
+    echo "<label for='wordInstancesSelect'>" . _("Number of word context results: ") . "</label>";
+    echo "<select id='wordInstancesSelect' name='wordInstances' style='margin-left: 2px;'>";
+    foreach(range(10, 100, 10) as $option) {
+        echo "<option value='$option'" . ($option == $wordInstances ? "selected" : "") . ">$option</option>";
+    }
+    echo "</select>";
+    echo "<br />";
+    echo "<input type='submit' value='" . _("Update") . "' />";
```
srjfoo, line 96: I'd either leave the "Update" button on the same line as the pulldown, or make it an auto-update pulldown. When I tested it, my eyes didn't immediately see the update button, because it was all the way back at the beginning of the line, hidden under existing text. (Not sure whether the auto-update is more appropriate, or the "execute" button, from a UX design standpoint.)

chrismiceli: Sure, I think updating automatically is better.

tools/project_manager/show_word_context.php
```diff
@@ -81,6 +81,20 @@
...
 
+    echo "<form method='GET' id='wordInstancesForm'>";
+    echo "<input type='hidden' name='projectid' value='$projectid' />";
+    echo "<input type='hidden' name='word' value='$encWord' />";
+    echo "<input type='hidden' name='layout' value='$layout' />";
+    echo "<input type='hidden' name='frame' value='left' />";
+    echo "<label for='wordInstancesSelect'>" . _("Number of word context results: ") . "</label>";
+    echo "<select id='wordInstancesSelect' name='wordInstances' style='margin-left: 2px;' onchange='$(\"#wordInstancesForm\").submit()'>";
+    foreach(range(10, 100, 10) as $option) {
+        echo "<option value='$option'" . ($option == $wordInstances ? "selected" : "") . ">$option</option>";
```
cpeel, line 92: Add a space after `'$option'` so there's one before `selected`.

tools/project_manager/show_word_context.php
```diff
@@ -112,7 +126,7 @@
     mysqli_free_result($pages_res);
 
     if($foundInstances>=$wordInstances)
-        echo "<p>" . _("More instances were found, stopping after a small sample.") . "</p>";
+        echo "<p>" . _("More instances were found; please choose how many to show from the drop-down, up to 100.") . "</p>";
```
cpeel, line 129: I would remove `, up to 100` so if we change the max size later we don't have to have the string retranslated. You could also just `sprintf()` the string but I don't think it's important to include the number here.

srjfoo: It just occurred to me, should the max size be a named constant, anyway, or is that irrelevant here?

cpeel: I think it's worth making a constant since we use it in `get_integer_param()` and in the `range()`. I still think removing it from here is fine.

tools/project_manager/show_word_context.php
```diff
@@ -81,6 +81,22 @@
     echo "</a>";
     echo "</p>";
 
+    echo "<form method='GET'>";
+    echo "<input type='hidden' name='projectid' value='$projectid' />";
```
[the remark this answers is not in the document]
chrismiceli: Ugh, I've been coding too much in React's JSX lately which doesn't have these rules.

tools/project_manager/show_word_context.php
chrismiceli: I put it before the word selected later on so it is only present if we are on the selected one. Not that it matters much.

tools/project_manager/show_word_context.php
chrismiceli: Removed from string and used constant for range and get_integer_param.

PR #311, opened 2020-05-25
Split test

70ray:
splitControl.js implements a splitter. The tests use it in various ways.


cpeel:
I'm happy to have the contents of `manual_web/` restructured, but it needs to stay in the `tests/` directory because they're tests.

SETUP/manual_web/split_test/horizontal_split_flex.php
```diff
@@ -0,0 +1,59 @@
+<?php
+$relPath="../../../pinc/";
+include_once($relPath.'base.inc'); // require_login()
+include_once($relPath.'slim_header.inc');
+
+// Demonstrate two pane vertical split
+
+require_login();
```
cpeel, line 8: I would remove this from all files and allow unauthenticated people to use it.

SETUP/manual_web/split_test/h_split.js
```diff
@@ -0,0 +1,12 @@
+/* global $ splitControl */
```
cpeel, line 1: Consider renaming the test-centric JS files so they match the name of the PHP files they are used it so it's clear what goes with what at a glance.

SETUP/manual_web/split_test/vertical_horizontal_split_flex.php
```diff
@@ -0,0 +1,43 @@
...
+
+echo "<div style='display: flex; flex-direction: column; height: 100vh;'>\n";
+
+echo "<div style='background-color: aquamarine;'>";
+echo "<h1>$title</h1>\n";
+echo "</div>\n";
+
+echo "<div id='top-container' style='background-color: beige; flex: auto;'>\n";
+
+echo "<div>This is pane1</div>\n";
```
cpeel, line 34: Please include some more content in each pane. You might just throw the text from `/horizontal_split_flex.php` into a .txt file in the directory and `file_get_contents()` it into each pane in each test.

scripts/splitControl.js
```diff
@@ -0,0 +1,172 @@
+/*global $ */
+/* exported splitControl */
+
+/*
+ * Create a splitter between two <div>s within a container.
+ * Arguments:
+ * container - ID of a <div> which contains two <div>s (herein referred
+ * to as pane1 and pane2). The splitter will be created between these.
```
cpeel, line 8: Add space before the text so it's clear that it's associated with the `container` label above it, eg:
```
 * container - ID of a <div> which contains two <div>s (herein referred
 *     to as pane1 and pane2). The splitter will be created between these.
```
Ditto other argument descriptions and returns.

chrismiceli:
We could setup a JS test suite that runs on a headless browser perhaps using karma, jest, or chai. I can look into it as part of this if we want.

70ray:
> We could setup a JS test suite that runs on a headless browser perhaps using karma, jest, or chai. I can look into it as part of this if we want.

Aside from this, I was looking at JS tests in relation to https://www.pgdp.net/c/tasks.php?action=show&task_id=1890 and had aquick search of what was available. Qunit seemed to be suitable but I'd be interested to know what you would suggest since I know nothing about this.

SETUP/manual_web/split_test/h_split.js
```diff
@@ -0,0 +1,12 @@
...
+$(function () {
+    let reDraw = $.Callbacks();
+
+    $(window).resize(function () {
```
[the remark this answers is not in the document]
70ray: That could be a good approach. It seems possible that there could be other ways (apart from window.resize() that the container could get resized so the splitter would break but I haven't found any.
It would be best to detect the container resizing directly by ResizeObserver. This would also remove the need for the resize return parameter but as Casey observed in https://github.com/DistributedProofreaders/dproofreaders/pull/298 it is not supported in all the browsers we want to support. I tried it and it works well in Firefox but only partially in Edge and Chrome: it doesn't get triggered by a resize induced by javascript. Another way to do it I found on stackoverflow is to continuously check the size with a timer function but I didn't want to do that.

SETUP/manual_web/split_test/horizontal_split_flex.php
```diff
@@ -0,0 +1,59 @@
...
+$js_files = [
+    "$code_url/scripts/splitControl.js",
+    "h_split.js",
+];
+
+// overflow: hidden prevents getting incorrect container size
+// when using flex layout and window size is reducing
+$header_args = [
+    "js_files" => $js_files,
+    "body_attributes" => "style='margin: 0; overflow: hidden;'",
```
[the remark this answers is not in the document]
70ray: `margin: 0;` is to override the `margin: 0.5em` in slim_header.css. There could be a case for removing the margin from slim_header and putting it in some other way where it is needed.
`overflow: hidden;` is only needed with flex layout. Could there be a bug in flex layout?

SETUP/manual_web/split_test/v_split.js
```diff
@@ -0,0 +1,12 @@
+/* global $ splitControl */
+$(function () {
+    let reDraw = $.Callbacks();
+
+    $(window).resize(function () {
+        reDraw.fire();
+    });
+
+    let mainSplit = splitControl("#container", reDraw, {splitDirection: "vertical"});
```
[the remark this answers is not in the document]
70ray: That seems a good idea.

scripts/splitControl.js
```diff
@@ -0,0 +1,172 @@
+/*global $ */
+/* exported splitControl */
+
+/*
+ * Create a splitter between two <div>s within a container.
+ * Arguments:
```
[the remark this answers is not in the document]
70ray: I don't know anything about this but I'll have a look at it.

scripts/splitControl.js
```diff
@@ -0,0 +1,172 @@
...
+ * drawing any divs surrounding the container.
+ * reSize: this jquery callback is fired after relayout has been called
+ * and after moving the dragbar. It can be used as the reDraw parameter for
+ * subsidiary splitControls.
+ * dragEnd: this jquery callback is fired at the end of a drag resize with
+ * a percentage parameter. It enables the split percentage to be stored so
+ * that when splitControl is used again the split ratio can be persisted.
+*/
+function splitControl(container, reDraw, config) {
+    let theConfig = {splitDirection: "horizontal", splitPercent: 50, dragBarSize: 6, dragBarColor: "darkgray"};
```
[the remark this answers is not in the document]
70ray: Thanks, It will take me a day or two to learn all these new tricks.

scripts/splitControl.js
```diff
@@ -0,0 +1,172 @@
...
+            range = 0;
+        }
+        splitPos = base + (range * splitRatio);
+        // mouse sets top/left of dragbar
+        maxPos = base + range;
+        moveSplit();
+    }
+
+    function dragMove(event) {
+        splitPos = ((theConfig.splitDirection === "vertical") ? event.pageX : event.pageY);
```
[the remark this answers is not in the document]
70ray: No indeed.

SETUP/manual_web/split_test/horizontal_split_flex.php
70ray: done this.

SETUP/manual_web/split_test/h_split.js
70ray: done this.

SETUP/manual_web/split_test/vertical_horizontal_split_flex.php
70ray: done this with text and image

scripts/splitControl.js
70ray: done this.

SETUP/manual_web/split_test/h_split.js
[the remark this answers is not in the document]
70ray: This is now default and moved into config

SETUP/manual_web/split_test/v_split.js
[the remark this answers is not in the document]
70ray: done something like this.

scripts/splitControl.js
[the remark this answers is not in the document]
chrismiceli: I guess both aren't supported in IE so what we have is probably fine.

chrismiceli:
Looks much better to me, I need to play with it more some though.

SETUP/tests/manual_web/split_test/vertical_split.php
```diff
@@ -0,0 +1,28 @@
...
+include_once($relPath.'slim_header.inc');
+
+// Demonstrate two pane vertical split
+
+$title = _('Two Pane Vertical Split');
+
+$js_files = [
+    "$code_url/scripts/splitControl.js",
+    "vertical_split.js",
+    ];
```
cpeel, line 13: nit: un-indent this closing bracket

SETUP/tests/manual_web/split_test/switchable_split.php
```diff
@@ -0,0 +1,36 @@
...
+
+$title = _('Two Pane Switchable Split with Flex Layout');
+
+$js_files = [
+    "$code_url/scripts/splitControl.js",
+    "switchable_split.js",
+];
+
+$header_args = [
+    "js_files" => $js_files,
```
cpeel, line 16: What you have here is fine. Note also you can do:
```
$header_args = [
    "js_files" => [
        "$code_url/scripts/splitControl.js",
        "switchable_split.js",
    ],
    "css_files" => ["split_test.css"],
    "body_attributes" => "style='margin: 0; overflow: hidden;'",
];
```

SETUP/tests/manual_web/split_test/vertical_split.php
70ray: done this

SETUP/tests/manual_web/split_test/switchable_split.php
70ray: Changed all files to be like this.

cpeel:
@chrismiceli - any additional thoughts / feedback on this MR?

chrismiceli:
> @chrismiceli - any additional thoughts / feedback on this MR?

I'll review it tonight. Apologies on the delay. I'll try to update my test PR too with some real tests.

chrismiceli:
Looks good to me!

PR #312, opened 2020-05-25
Better charsuite validation

cpeel:
Add two new validation checks for character suites in the code:
1. The suites do not contain characters that we convert on page save.
2. The suites use lowercase hex values to match what portable-utf8 provides with `utf8_chr_to_hex()`.

This MR is all non-functional changes outside of the linting. I have also confirmed via case-insensitive diff that the character suites before and after are identical outside of case.

These two functions in `unicode.inc` will be used in the per-project custom suites as well.

### DIFF
--- a/SETUP/lint_charsuites.php
+++ b/SETUP/lint_charsuites.php
@@ -35,6 +35,9 @@ foreach(CharSuites::get_all() as $charsuite)
         exit(1);
     }
 
+    // Confirm that all hex codes are in lowercase
+    validate_hex_codes_lowercase($charsuite->codepoints, "uppercase hex values found in charsuite codepoints");
+
     // Validate pickersets only contain characters within the suite
     foreach($charsuite->pickerset->get_subsets() as $title => $picker_subset)
     {
@@ -51,6 +54,31 @@ foreach(CharSuites::get_all() as $charsuite)
                 }
                 exit(1);
             }
+
+            // Confirm hex codes are in lowercase
+            validate_hex_codes_lowercase($codepoints, "uppercase hex values found in picker set");
         }
+    }
+}
+
+function validate_hex_codes_lowercase($codepoints, $error_message)
+{
+    // Confirm that all hex codes are in lowercase
+    $uppercase_codepoints = [];
+    foreach($codepoints as $codepoint)
+    {
+        if(preg_match('/[ABCDEF]/', $codepoint))
+        {
+            $uppercase_codepoints[] = $codepoint;
+        }
+    }
+    if($uppercase_codepoints)
+    {
+        echo "ERROR: $error_message\n";
+        foreach($uppercase_codepoints as $codepoint)
+        {
+            echo sprintf("    %s\n", $codepoint);
+        }
+        exit(1);
     }
 }
--- a/SETUP/lint_charsuites.php
+++ b/SETUP/lint_charsuites.php
@@ -2,6 +2,7 @@
 <?php
 $relPath="../pinc/";
 include_once($relPath."CharSuites.inc");
+include_once($relPath."unicode.inc");
 
 foreach(CharSuites::get_all() as $charsuite)
 {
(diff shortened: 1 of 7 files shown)


SETUP/lint_charsuites.php
```diff
@@ -19,6 +20,24 @@
...
+    // to ASCII
+    $disallowed_characters = convert_codepoint_ranges_to_characters(get_disallowed_codepoints());
+    $charsuite_characters = convert_codepoint_ranges_to_characters($charsuite->codepoints);
+    $characters = array_intersect($disallowed_characters, $charsuite_characters);
+    if($characters != [])
+    {
+        echo "ERROR: disallowed characters found in suite\n";
+        foreach($characters as $char)
+        {
+            echo sprintf("    %s: %s\n", $char, utf8_chr_to_hex($char));
```
chrismiceli, line 33: should we use \t here?

pinc/charsuite-polytonic-greek.inc
```diff
@@ -4,87 +4,87 @@ include_once($relPath."CharSuites.inc");
...
+    'U+1fec',
+    'U+1ff2-U+1ff4',
+    'U+1ff6-U+1ff8',
+    # skip oxia U+1ff9
+    'U+1ffa',
+    # skip oxia U+1ffb
+    'U+1ffc',
 ];
 $charsuite->reference_urls = [
     "https://www.pgdp.net/wiki/DP_Code_-_Unicode/Greek",
```
chrismiceli, line 90: not sure the usual approach here, but should we use site vars to make this not pgdp.net specific?

SETUP/lint_charsuites.php
cpeel: Sorry, use `\t` where in the string?

pinc/charsuite-polytonic-greek.inc
cpeel: Good question. In this case the above-reference page goes into more detail about the complications when creating a Polytonic Greek character suite so it will be useful for other consumers too.

SETUP/lint_charsuites.php
chrismiceli: I meant for alignment. Instead of using 4 spaces before %s, use a \t character. Nothing critical. Just curious.

chrismiceli:
Looks good to me.

SETUP/lint_charsuites.php
srjfoo: My understanding is that using `\t` can be somewhat fraught, because it means different things in different places, depending on where tab stops are set.

Another question, though -- should it be `%1$s` and `%2$s`, or is that irrelevant because there's no translation to worry about?

SETUP/lint_charsuites.php
```diff
@@ -19,6 +20,24 @@
...
+        }
+        exit(1);
+    }
+
+    // Confirm that all hex codes are in lowercase
+    validate_hex_codes_lowercase($charsuite->codepoints, "uppercase hex values found in charsuite codepoints");
```
srjfoo, line 39: And speaking of translation, does this string need to be wrapped in `_()`?

SETUP/lint_charsuites.php
```diff
@@ -35,6 +54,31 @@
                 }
                 exit(1);
             }
+
+            // Confirm hex codes are in lowercase
+            validate_hex_codes_lowercase($codepoints, "uppercase hex values found in picker set");
```
srjfoo, line 59: As with line 39, `_()` for the string?

pinc/charsuite-polytonic-greek.inc
srjfoo: Would it be better to have it as a markdown file within the code base? (Or protect the wiki page and document why it should stay where it is.)

pinc/unicode.inc
```diff
@@ -310,3 +310,68 @@ function guess_string_encoding($text)
     # Give up and return ISO-8859-1
     return 'ISO-8859-1';
 }
+
+# Return a list of Unicode codepoints we want to replace with some ASCII
+# equivalence. We convert these codepoints to ASCII up on page save in
+# DPage.inc.
+function get_utf8_to_ascii_codepoints()
+{
+    return [
```
srjfoo, line 319: The first three replacement characters below are delimited with apostrophes; the rest of them with  double quotes. All of the codepoints are delimited by apostrophes. Just wondering whether it should be all one or the other?

SETUP/lint_charsuites.php
cpeel: > Another question, though -- should it be `%1$s` and `%2$s`, or is that irrelevant because there's no translation to worry about?

It's irrelevant because it isn't translated.

pinc/charsuite-polytonic-greek.inc
cpeel: It's fine where it is. I'm not too worried about someone coming in and changing the wiki page, although if we wanted to protect it that wouldn't be a bad idea.

SETUP/lint_charsuites.php
cpeel: We don't translate files in `SETUP/` since they will never be seen by the user -- just site admins when they set up the site or developers.

pinc/unicode.inc
cpeel: It's mostly irrelevant. If we used all of one we'd need to escape the character in one of the strings.

PR #316, opened 2020-05-31
Base API framework

cpeel:
This creates a framework for an API to access DP data. This MR will merge into an upstream **api** branch -- similar to what we had for unicode -- until we have enough base functionality to go into master.

The framework includes:
* JSON-based RESTful API
* Authentication via API key linked to user account
* Authorization via user account associated with the API key
* Optional rate limiting using memcached

Initial API functionality in this MR includes:
* GET project metadata
* GET project word lists (good & bad)
* PUT project word lists (good & bad)

You can see this for yourself on the TEST server using `curl` (the `review` API key is linked to one of my unprivileged TEST accounts):
```
curl -i -H "Accept: application/json" -H "X-API-KEY: review" \
    "https://www.pgdp.org/~cpeel/c.branch/api-part-0/api/index.php?url=v1/projects/projectID5e672cf6b96c9"

curl -i -H  "Accept: application/json" -H "X-API-KEY: review" \
    "https://www.pgdp.org/~cpeel/c.branch/api-part-0/api/index.php?url=v1/projects/projectID5e672cf6b96c9/wordlists/good"
```

I have additional functionality that builds on top of this code (project search, site and round statistics) that further proves out the design.

My initial target for this API is to surface the data used in the Archer's Gap game since we have a developer willing to make those modifications (/me waves to Brian). Eventually this will be used to provide API access for client-side JS.

This post was particularly insightful when getting started creating the API:
* https://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api


pinc/base.inc
```diff
@@ -1,84 +1,25 @@
 <?php
+// This file sets up code infrastructure for pages that output HTML.
+// ../api/index.php handles similar infrastructure for the API.
```
bpfoley, line 3: s/API/JSON API/ ?

pinc/base.inc
```diff
@@ -1,84 +1,25 @@
 <?php
+// This file sets up code infrastructure for pages that output HTML.
```
bpfoley, line 2: s/HTML/content/, since presumably this will include raw JSON too.

pinc/base.inc
```diff
@@ -1,84 +1,25 @@
...
 // * (re)establishing the session
 // * setting up gettext
 // * enforcing maintenance mode
 // * providing the require_login() function
 //
-// This file should be included at the top of EVERY .php page. This
-// is not a catch-all. The kitchen sink does not belong in here!
-// Keeping this and its dependencies slim improves overall page
+// This file should be included at the top of EVERY .php page that renders
+// HTML. Keeping this and its dependencies slim improves overall page
```
bpfoley, line 13: s/HTML/content/, as above

pinc/bootstrap.inc
```diff
@@ -0,0 +1,76 @@
...
+
+// Exception handlers are defined differently for HTML and the API
+if($testing)
+{
+    set_exception_handler('test_exception_handler');
+}
+else
+{
+    set_exception_handler('production_exception_handler');
+}
```
bpfoley, line 38: A nit (feel free to ignore), but would this be less verbose as
```
set_exception_handler($testing ? 'test_exception_handler' : 'production_exception_handler');
```

SETUP/API.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Access Control
+
+API access is limited to individuals with API keys. API keys are associated
+with users and the API calls are restricted to the same information as the
+user who owns it.
+
+To enable an API key for a user, add a 32-character GUID to the user's
+`api_key` column in the database.
+
```
bpfoley, line 40: Cool! So this means a user doesn't need to be logged in to access the API?

api/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+`api_v1_projects()` checks the URI to see if there are additional parts on
+the URI. If there are, it loads the `projectid` off the URI, validates it
+and passes the Project object back to the router for the next handler.
+
+The router then calls `api_v1_projects_PROJECTID_wordlists()` which takes the
+project, determines which words list to use (`good`) and based on the method
+type either returns the word list (GET) or updates the word list (PUT). In
+both cases it returns a data structure that will be output to the user in
+JSON format.
```
bpfoley, line 35: A thought for future development here: We might want to be able to associate API keys with ACLs which only allow particular ACLs, and perhaps allow a user to generate multiple API keys each with different ACL sets.
Specifically, it would be good to allow users to read but not write the wordlist.

api/README.md
```diff
@@ -0,0 +1,69 @@
...
+### URI-handling function responsibilities
+
+URI-handling functions are each responsible for managing the following
+items:
+
+1. Pulling any resource variables off the URI and validating them.
+2. If the URI still contains other elements, calling `api_router()` to handle
+   them, passing along resources variables or objects created from them.
+3. Validating that the requested method is supported by calling
+   `api_validate_method()` with the supported methods.
```
bpfoley, line 46: Insert permission check here (later)?

api/dp-openapi.yaml
```diff
@@ -0,0 +1,245 @@
...
+      - title
+      type: object
+      properties:
+        projectid:
+          type: string
+        title:
+          type: string
+        author:
+          type: string
+        languages:
```
bpfoley, line 143: A nit here, but I find the YAML for the property types really verbose and hard to eyeball. YAML has an alternate denser syntax that might make this specific case a bit better.

```yaml
properties:
 projectid: {type: string}
 title: {type: string}
 author: {type: string}
 languages: {type: array, items: {type: string}}
 genre: {type: string}
```

and
```yaml
 credits_so_far: {type: string, readOnly: true}
 modified: {type: string, format: dateTime, readOnly: true}
```

What do you think?

api/index.php
```diff
@@ -0,0 +1,203 @@
...
+{
+    $username = api_authenticate();
+
+    api_rate_limit($username);
+
+    $query_params = $_GET;
+    $path = explode('/', array_get($query_params, "url", ""));
+    unset($query_params["url"]);
+
+    // if the path ended with a /, pop it off for correct routing
```
bpfoley, line 38: Is this valid? For example, I'd expect "/projects" (which would list all projects) vs "/projects/" (which should error because you're missing a {projectID}), to behave differently.

api/index.php
```diff
@@ -0,0 +1,203 @@
...
+        return;
+    }
+
+    $memcache = new Memcached();
+    $memcache->addServer('localhost', 11211);
+
+    // increment (or initialize) our count
+    $count = $memcache->get("$key:count");
+    $count = $count === FALSE ? 0 : $count + 1;
+    if($memcache->set("$key:count", $count) === FALSE)
```
bpfoley, line 86: As a microoptimization, in the common case we could replace get/set with increment to avoid a second round trip...
https://www.php.net/manual/en/memcache.increment.php

api/index.php
```diff
@@ -0,0 +1,203 @@
...
+
+    $next_hop = array_shift($path);
+    $next_hop_function = "${last_hop}_${next_hop}";
+    if(!$next_hop || !function_exists($next_hop_function))
+    {
+        $path_url = str_replace("_", "/", $next_hop_function);
+        throw new InvalidAPI();
+    }
+
+    return $next_hop_function($path, $query_params, $data);
```
bpfoley, line 135: This aspect of PHP terrifies me, but hey, it is what it is. At least we have api_validate_method below to stop API clients from calling arbitrary functions...

api/index.php
```diff
@@ -0,0 +1,203 @@
...
+    {
+        throw new MethodNotAllowed();
+    }
+
+    return $method;
+}
+
+function api_get_request_body()
+{
+    return json_decode(file_get_contents('php://input'));
```
bpfoley, line 151: Oh wow. TIL. php://input. :)

pinc/base.inc
cpeel: No, `base.inc` is just for HTML files.

SETUP/API.md
cpeel: Correct. I have an idea for how to support the UI accessing the API via logged-in sessions as well later.

api/index.php
cpeel: I started trying that, but increment only works when using the binary protocol (see first comment on that page) which it looks like the version of PHP we use on TEST doesn't support. I banged my head against this for longer than I want to admit.

api/index.php
cpeel: I am very open to other approaches to an implementation abstraction that doesn't require every endpoint to validate arguments. For instance:
```
v1/projects/:projectid
v1/projects/:projectid/wordlists/good
```
^^^ I don't want to have both of those do independent validations of the projectid.

api/index.php
cpeel: Ah, yes, that would be a better approach. I'll get this updated.

api/dp-openapi.yaml
cpeel: Oh, cool! Honestly, I was just building this in the Swagger editor which uses the verbose syntax by default. It looks like it happily supports the denser syntax as well. I'll check out using the denser syntax for some of the schema bits, although I don't want to use it for everything.

api/README.md
cpeel: This list should absolutely call out the authorization bits - I will get that updated.

api/README.md
cpeel: Conceptually, I totally agree with you. In practice, the object authorization model in the code is completely decentralized right now, which makes this almost impossible. Examples:
* The project page limits the user's ability to see the clearance line based on who they are, but the Project object doesn't know anything about such limitations.
* The user's ability to edit the project word lists is embedded in the word list editing page not in the underlying code that saves the word lists.
Both of the above determine this by the currently-logged-in user (`$pguser`).

The word list example is a good one -- I mis-coded the enforcement in the API as it currently doesn't match the rest of the site as all users can see word lists.

I think having a 1:1 mapping for users is the least-bad option right now to get a working API. We have a fair amount of work to do before we can break that into 1:many with different authorization levels.

pinc/bootstrap.inc
cpeel: Very nice - thanks!

pinc/base.inc
cpeel: Fixed!

api/README.md
70ray: Anterior Cruciate Ligament?

cpeel: In this case, [Access Control Lists](https://en.wikipedia.org/wiki/Access-control_list).

jmdyck:
I was initially puzzled by the reference to "the Archer's Gap game". A link to [the relevant forum topic](https://www.pgdp.net/phpBB3/viewtopic.php?f=11&t=32137) might be helpful.

cpeel:
> I was initially puzzled by the reference to "the Archer's Gap game". A link to [the relevant forum topic](https://www.pgdp.net/phpBB3/viewtopic.php?f=11&t=32137) might be helpful.

See also the [P3 Archers](https://www.pgdp.net/wiki/P3_Archers) wiki page.

Right now the Archer's Gap game is done via HTML scraping. Doing this by API would be less fragile and have less load on the system since it is just pulling the information it needs rather than the full HTML page it doesn't.

cpeel:
I've changed the router to be **much** clearer -- explicitly defining the routes and where they go. I'm much happier with this version.

api/index.php
cpeel: The new implementation is much better and doesn't rely on dynamic functions.

cpeel:
> I'm trying to visualize how parts work here. The implementation of `add_route` and `route` seems to require that all parts are unique, so you wouldn't be able to distinguish between (eg) `project/info` and `user/info` because `add_route` will insert 'project', 'info', 'user' into `url_map` and then the last 'info' will be ignored.

`url_map` is a tree, so `v1/project/info` and `v1/user/info` represent different paths:
```
v1
  project
    info
  user
    info
```

> A similarl issue occurs with param placeholders, if for example you were to use the same param name for different types (eg maybe a project `id` is different from a user `id`.)

Parameters must uniquely map to a validator, so `:user_id` and `:project_id` can both exist and be validate by different functions, but `:id` would need to map to just one of them.

PR #318, opened 2020-06-02
1895: Search and Replace does not correctly find and replace formatting tags

chrismiceli:
update regular expression escaping to match javascript regex engine.

### DIFF
--- a/tools/proofers/srchrep.php
+++ b/tools/proofers/srchrep.php
@@ -22,7 +22,7 @@ function do_replace()
     var is_regex = regex_checkbox.checked;
     if (!is_regex)
     {
-        search = preg_quote(search);
+        search = escapeRegExp(search);
     }
     opener.parent.docRef.editform.text_data.value=opener.parent.docRef.editform.text_data.value.replace(new RegExp(search,'gu'),replacetext);
     set_undo_button_disabled(false);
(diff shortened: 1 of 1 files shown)


chrismiceli:
The function we had to escape regular expressions did not match what javascript needed. I'm not sure why this just started failing, perhaps it was just noticed? I took the regular expression escaping from MDN, the most reliable JS place on the internet in my opinion:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions

There was also logic to disable the regular expression checkbox I assume in case your browser didn't support regular expressions, but I'm not sure why that would ever be the case so I removed it. Let me know if I should restore it.

chrismiceli:
sandbox updated here: https://www.pgdp.org/~chrismiceli/c/ @srjfoo

chrismiceli:
> 
> 
> A quick spot of testing -- the initial problem that was noticed seems to be fixed.
> 
> However, `\n` is treated as a newline indicator, whether in normal mode or in regex mode. In normal mode, shouldn't it be treated as just any other character pair, and only treated as a newline in regex mode?

We discussed this last time (if I recall correctly) an decided that \n isn't a regex and shouldn't require the checkbox, especially given it shouldn't show up in normal text often anyways.

srjfoo:
> > A quick spot of testing -- the initial problem that was noticed seems to be fixed.
> > However, `\n` is treated as a newline indicator, whether in normal mode or in regex mode. In normal mode, shouldn't it be treated as just any other character pair, and only treated as a newline in regex mode?
> 
> We discussed this last time (if I recall correctly) an decided that \n isn't a regex and shouldn't require the checkbox, especially given it shouldn't show up in normal text often anyways.

Oops ... my mistake. For some reason, I thought that if I did a non-grep s/r in my regular text editor, it would read just the characters, but I just checked, and I was totally wrong.

PR #319, opened 2020-06-02
If javascript crashes in previewControl.js, do not submit the form.

70ray:
This prevents possible loss of work. js crash can occur as a result
of third-party code which we might use such as MathJax

### DIFF
--- a/pinc/button_defs.inc
+++ b/pinc/button_defs.inc
@@ -204,7 +204,7 @@ function echo_button( $button_id, $which_interface )
                 'name'      => "button14",
                 $label      => attr_safe(_("Preview")),
                 'title'     => attr_safe(_("Format preview")),
-                'onclick'   => "previewControl.show(); return false;",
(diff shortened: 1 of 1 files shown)


chrismiceli:
Why do we expect crashes? Would it be possible to wrap code we can't assume to not fail with a try catch and handle the issue there?

70ray:
> Why do we expect crashes? Would it be possible to wrap code we can't assume to not fail with a try catch and handle the issue there?

Good idea! I'll try it out. Then we can show an error message. It happened when I was using MathJax to preview Latex equations and hit a bug in MathJax.

70ray:
> Why do we expect crashes? Would it be possible to wrap code we can't assume to not fail with a try catch and handle the issue there?

Good idea! I'll try it out. Then we can show an error message. It happened when I was using MathJax to preview Latex equations and hit a bug in MathJax.
There is probably little point in making this change now.

PR #320, opened 2020-06-04
Math preview 2

70ray:
This previews Latex math markup. It is for use in projects which will have the markup converted to images as described in https://www.pgdp.net/wiki/User:RP31/Math_conversion_tool

The function interprets equations inside the tags `\[ \] and \( \)`. It also checks that these tags are correctly matched.

It would be possible to make it so that it only operates if it is explicitly enabled in the preview configuration. I haven't done this yet but can do it if you think it would be best.

It uses MathJax to do the preview. In the final book the equations will get converted to images (first svg then svg to png) using a related MathJax function. It would thus be best to use the svg version of MathJax here but it has a known bug which should be fixed in the next release so in the meantime a different version of Mathjax is used.

### DIFF
--- a/tools/proofers/previewControl.js
+++ b/tools/proofers/previewControl.js
@@ -100,7 +100,11 @@ function initPrev() {
         );
         prevWin.innerHTML = preview.txtout;
         if (preview.ok) {
-            MathJax.typeset([prevWin]);
+            try {
+                MathJax.typeset([prevWin]);
+            } catch(exception) {
+                alert("MathJax error: " + exception);
+            }
         }
         issBox.value = preview.issues;
         possIssBox.value = preview.possIss;
--- a/tools/proofers/previewControl.js
+++ b/tools/proofers/previewControl.js
@@ -1,6 +1,6 @@
 /* eslint-disable no-use-before-define, camelcase */
 /* exported previewControl, initPrev */
-/* global $ previewDemo, makePreview, ieWarn */
+/* global $ previewDemo, makePreview, ieWarn MathJax */
 /*
 This file controls the user interface functions. Initially nothing is displayed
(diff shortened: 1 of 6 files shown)


tools/proofers/previewControl.js
```diff
@@ -99,6 +99,13 @@ function initPrev() {
                 : "pre"
         );
         prevWin.innerHTML = preview.txtout;
+        if (preview.ok) {
+            try {
```
cpeel, line 103: Rather than load the MathJax JS file for every proofreading page -- may of which won't ever use Format Preview -- consider only loading the JS when needed, possibly with https://api.jquery.com/jQuery.getScript/. I also see merit in having this be an option in Format Preview so that we only attempt loading the JS and running MathJax if the user wants to preview math.

70ray: Good point, I'll do that.

70ray:
Should we have a local copy of MathJax and add it to the credits?

cpeel:
> Should we have a local copy of MathJax and add it to the credits?

I would be fine with that since it appears to be a single-file import. If we bundle it locally, should we be using the `tex-chtml-full` file instead?

70ray:
> > Should we have a local copy of MathJax and add it to the credits?
> 
> I would be fine with that since it appears to be a single-file import. If we bundle it locally, should we be using the `tex-chtml-full` file instead?

I suppose that would be best. I don't know what the extensions are that it includes but it wouldn't hurt to have them. I incorrectly have `tex-mml-chtml.js` since we do not need mml input.

70ray:
Closing this because it would probably be difficult to merge with the "analyse" changes to preview.js. I'll re-make it after that is merged.

PR #322, opened 2020-06-07
exclude node_modules for json checks

chrismiceli:


chrismiceli:
These json are not controlled by us and slow our build time.

PR #324, opened 2020-06-12
Analyse

70ray:
This is not intended to be merged in its present form but is going through the process to see if the tests run correctly.
It revises preview.js in various ways and adds tests for it.


70ray:
I think it failed: "UnhandledPromiseRejectionWarning".
I forgot to include preview.js.
Trying again

70ray:
I think it worked this time!

SETUP/tests/manual_web/qunit.html
```diff
@@ -0,0 +1,19 @@
+<!DOCTYPE html>
```
chrismiceli, line 1: do we want to duplicate this file? We can move the other one and update the npm run script if we want the location here.

SETUP/tests/jsTests/formatPreviewTests.js
```diff
@@ -0,0 +1,113 @@
+/* global QUnit analyse */
+
+QUnit.test("Format preview test", function( assert ) {
+    let inLineTags = "[ibfg]|sc";
```
chrismiceli, line 4: const since we don't seem to modify this. What is this? Should it come from the preview.js code?

SETUP/tests/jsTests/formatPreviewTests.js
```diff
@@ -0,0 +1,113 @@
...
+    let suppress = {};
```
chrismiceli, line 5: const?

SETUP/tests/jsTests/formatPreviewTests.js
```diff
@@ -0,0 +1,113 @@
...
+    let text;
+    let issArray;
```
chrismiceli, line 7: could we pass this to issueTest each time instead of a sort of global?

SETUP/tests/jsTests/formatPreviewTests.js
```diff
@@ -0,0 +1,113 @@
...
+
+    // assert start, length code and type correct for issue at index
+    function issueTest(index, start, length, code, type) {
+        let issue = issArray[index];
+        assert.deepEqual(issue.start, start);
```
chrismiceli, line 12: do we need deepEqual? Are these objects? Could we use strictEquals instead?

SETUP/tests/jsTests/formatPreviewTests.js
```diff
@@ -0,0 +1,113 @@
...
+        assert.deepEqual(issue.len, length);
+        assert.deepEqual(issue.code, code);
+        assert.deepEqual(issue.type, type);
+    }
+
+    // missing inline start tag
+    text = "df</b> ab";
```
chrismiceli, line 19: I think it makes more sense to make each one of these their own QUnit.test case, no?

70ray:
> The change in previewControl is really large and difficult to review. What was your strategy there? Knowing may make it easier to understand.

I think you meant preview.js. As usual I forgot to mention the most important things: At present the preview makes a text with html markup indicating issues with style and colouring and translated strings so it seemed it would be really difficult to make tests. In order to run tests more easily I separated out the analyse function. This just makes an array of issues. Each entry has the position in the text and the name and severity of the issue and in some cases an extra string indicating what caused the issue. The changes required for this mainly involve moving functions around. I also wrapped it all in a `$(function()` to avoid putting more functions into global scope unnecessarily.
It would also be possible to test the other parts of preview independently: putting in the markup and doing the rewrap but I haven't addressed these yet.

70ray:
> I think it failed: "UnhandledPromiseRejectionWarning".
> I forgot to include preview.js.
> Trying again

@chrismiceli It seemed like this caused it to not run any js tests so it said it had passed (all tests that it ran). Do you think it would be possible to catch this problem and indicate a failure?

SETUP/tests/jsTests/.htaccess
```diff
@@ -0,0 +1 @@
+Allow from all
```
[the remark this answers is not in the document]
70ray: I ran the tests manually while working on this using the duplicated qunit.html. It seemed I had to put the .htaccess in to enable it to work. This was a hacking job in order to get it working. What I wanted to do was to be able to run just the new tests manually as well as incorporating them into your new automatic testing regime. If you can show me how to do this in a better way I will change it accordingly.

SETUP/tests/manual_web/qunit.html
70ray: This is just a hack as above.

SETUP/tests/jsTests/formatPreviewTests.js
70ray: I think the paramters I chose for the anayse function may not be the best. Maybe taking the whole `style` into it would be better and allow for the optional math markup in another MR.

SETUP/tests/jsTests/formatPreviewTests.js
70ray: That would make it clearer. I have perhaps taken too literally Douglas Crockford's opinion that it is a "very good thing" that "inner functions get access to the variable of functions they are defined within".

SETUP/tests/jsTests/formatPreviewTests.js
70ray: Yes, I should have read the documentation more carefully.

SETUP/tests/jsTests/formatPreviewTests.js
70ray: I agree. This is not really the right way do do it. I'll reconsider this.

SETUP/tests/jsTests/formatPreviewTests.js
70ray: Maybe, I was tring to make it more compact.

SETUP/tests/jsTests/formatPreviewTests.js
```diff
@@ -0,0 +1,113 @@
...
+    issueTest(1, 4, 3, "misMatchTag", 1);
+    issueTest(0, 9, 4, "misMatchTag", 1);
+
+    // character before start tag, possible issue
+    text = "as<i>df</i>";
+    issArray = analyse(text, inLineTags, suppress);
+    issueTest(0, 1, 1, "charBeforeStart", 0);
+
+    // - before start tag, ok
+    text = "as-<i>df</i>";
```
contramu, line 46: Testing this special case seems to have limited value unless you test all characters individually.

70ray: > It would be impossible to test all cases for positive or negative results so does this really do anything useful?

There is some truth in this but it is better than nothing and, provided one example of each positive and negative test is made, it could catch any inadvertent breaking changes. Making it testable also has the side-effect of forcing the preview code to be written in a better structured way.

SETUP/tests/jsTests/.htaccess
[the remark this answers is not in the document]
chrismiceli: Sent a PR to your branch to resolve this:
https://github.com/70ray/dproofreaders/pull/1

tools/proofers/previewControl.js
```diff
@@ -16,7 +16,8 @@ var previewControl;
 // this is called (in proof_frame_enh.inc or text_frame_std.inc) when the
 // preview button is pressed, the returned functions are then members
 // of previewControl
-function initPrev() {
+//function initPrev() {
```
cpeel, line 19: Remove commented-out line.

SETUP/tests/jsUnit.js
```diff
@@ -14,15 +12,19 @@ const assert = require('assert').strict;
             new firefox.ServiceBuilder(
                 `./node_modules/geckodriver/geckodriver${process.platform === 'win32' ? '.exe' : ''}`))
         .build();
+    let succeeded = false;
     try {
-        await driver.get(`file:///${process.cwd()}/SETUP/tests/qunit.html`);
+        await driver.get(`file:///${process.cwd()}/SETUP/tests/manual_web/qunit.html`);
```
cpeel, line 17: qunit tests aren't manual tests -- they don't require a user to do anything, they're part of our CI/CD process. Only tests that require a user to manually do something to validate it should be in `manual_web`.

chrismiceli: That was my opinion as well. You can load the file from file:// and not apache if you want to test locally. We should move it under test in my opinion and avoid duplicate.

tools/proofers/previewControl.js
70ray: done this now.

SETUP/tests/jsUnit.js
70ray: Moved tests out of manual tests. Had to put in htaccess to allow running manually.

70ray:
@chrismiceli I still have the deliberately failing test. Is there a way to provide an output log of the failing tests in Travis? At present the only way I can see to discover what went wrong is to run qunit locally.

chrismiceli:
> 
> 
> @chrismiceli I still have the deliberately failing test. Is there a way to provide an output log of the failing tests in Travis? At present the only way I can see to discover what went wrong is to run qunit locally.

Yeah I can look into that, but I think it shouldn't block what we have here since at least CI will fail.

cpeel:
The CI/CD tests are passing. I did a cursory check of the Format Preview UI in my sandbox and it looks good.

Is this ready to merge or is there other work needed?

70ray:
> The CI/CD tests are passing. I did a cursory check of the Format Preview UI in my sandbox and it looks good.
> 
> Is this ready to merge or is there other work needed?

I think it is ready to merge. preview.js has been considerably re-arranged but the tests should help to prove nothing has gone wrong.

PR #326, opened 2020-06-19
Remove global variable $userP

cpeel:
A user's current proofreading profile is saved in the user's session upon login. Sessions are then persisted across page accesses in the database and loaded into global variable `$userP` on every page load.

Persisting the proofreading sessions in the user's session is problematic when we want to add or remove settings from the profile. For instance, earlier this year we rolled out the profile updates that allow a user to specify a custom proofreading font. To roll this out we had to log out all users to ensure they logged back in and picked up the new profile options.

This also means that if a user is accessing DP with two different browsers and makes a changes to their proofreading profile, that change is not picked up by the second browser until they log out and back in.

This MR entirely removes the `$userP` global and instead accesses the user's profile whenever it is needed. This means we no longer need to store profile information in the user's session.

This change touches a lot of code because we used `$userP` in a number of different places for a variety of things. This code is live in the [remove-userP](https://www.pgdp.org/~cpeel/c.branch/remove-userP) sandbox.

### DIFF
--- a/pinc/User.inc
+++ b/pinc/User.inc
@@ -51,7 +51,37 @@ class User
             );
         }
 
-        $this->table_row[$name] = $value;
+        switch($name)
+        {
+            case "profile":
+                // set the user's profile, but first make sure it's a valid
+                // profile and they own it
+                if((! $value instanceof UserProfile) || $value->u_ref != $this->u_id)
+                {
+                    throw new NonexistentUserProfileException(
+                        "Invalid profile"
+                    );
+                }
+                $this->current_profile = $value;
+
+                // persist the current profile ID to the database
+                $sql = sprintf("
+                    UPDATE users
+                    SET u_profile = %d
+                    WHERE u_id = %d
+                ", $this->current_profile->id, $this->u_id);
+                mysqli_query(DPDatabase::get_connection(), $sql)
+                    or DPDatabase::log_error();
+
+                // now set the profile cookie if this is for the current user
+                if($this->username == User::current_username())
+                {
+                    setcookie("profile", $this->current_profile->id);
+                }
+                break;
+            default:
+                $this->table_row[$name] = $value;
+        }
     }
 
     public function __get($name)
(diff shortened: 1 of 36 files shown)


tasks.php
```diff
@@ -17,7 +17,8 @@
 
 $tasks_url = $code_url . "/" . basename(__FILE__);
 
-$requester_u_id = $userP['u_id'];
+$user = User::load_current();
```
chrismiceli, line 20: I don't think I see any issues, but here you introduce $user. If it was already in scope you could be overriding it. Would it be safer to do $requested_u_id = User::load_current()->u_id;
Again, probably not an issue.

cpeel: If `$user` was already in scope here we have problems as that shouldn't be a global variable -- we actually set that variable in various locations in the code assuming it will be used in a local context.

tools/proofers/proof_frame_std.inc
```diff
@@ -1,29 +1,28 @@
 <?php
 // The following variables must be defined by the file that includes this one:
 // $relPath
```
jmdyck, line 3: Might as well just delete this now.

tools/proofers/spellcheck.inc
```diff
@@ -53,10 +55,10 @@ if($userP['i_type']==1)
 slim_header(_("WordCheck"), $header_args);
 
 // print basic image html
-if ($userP['i_type']==1) {
-    if ($userP['i_layout']==1)
-       { $iWidth = $userP['v_zoom']; }
-    else { $iWidth = $userP['h_zoom']; }
+if ($user->profile->i_type = 1) {
```
jmdyck, line 58: Changed `==` to `=`!

70ray:
> > [Currently,] if a user is accessing DP with two different browsers and makes a changes to their proofreading profile, that change is not picked up by the second browser until they log out and back in.
> 
> Whereas with this PR, it'll be picked up by the second browser on the next request+response, right? But that might not be what people want either (e.g., if they're proofing on 2 devices with very different screens, or they're proofing 2 projects with very different image characteristics). It seems like at least `u_profile` should be session-specific, though I'm not sure how that would work.

Making the profile session-specific would not be the ideal solution to having different settings on different browsers because if a user logs out and in again their profile will change to whatever the current profile is. Could you have a separate profile cookie independent of the session?

pinc/User.inc
```diff
@@ -191,6 +191,37 @@ class User
 
     // static functions
 
+    // Return the username of the currenty-logged-in user
+    public static function current()
```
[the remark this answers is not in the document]
cpeel: This function is what will eventually phase out the global `$pguser` in the ever-present quest to kill global variables. I've renamed it to `User::current_username()`.

pinc/theme.inc
```diff
@@ -197,13 +197,14 @@ function html_navbar()
...
-    global $code_url, $site_abbreviation, $pguser, $forums_url, $blog_url, $wiki_url;
+    global $code_url, $site_abbreviation, $forums_url, $blog_url, $wiki_url;
 
     echo "<div id='navbar-outer'>";
     echo "<div id='navbar'>";
 
     //Code to display login form if not logged in
-    if (!isset($pguser)) {
+    $current_user = User::current();
+    if (!$current_user) {
```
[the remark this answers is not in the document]
cpeel: Good catch and suggestions -- am making these updates.

cpeel:
The updated commits should address the code issues @jmdyck found (thanks Michael!).

I'm still noodling over how to support other devices / browsers.

cpeel:
I've updated the branch to persist the active profile in a cookie. This allows using specific profiles across different browsers/devices. We jump through some extra hoops to ensure that the profile specified in the cookie is both a valid profile *and* belongs to the user. If either of those things are false the code uses the profile referenced from the user record. This also then handles the case where a user deletes a profile in use in another browser.

cpeel:
I would appreciate some folks poking at how this works for them across multiple browser/devices.

PR #327, opened 2020-06-22
fix for IE no default parameters

70ray:
if parameters are undefined use defaults.


cpeel:
I can't wait until we dump IE...

PR #328, opened 2020-06-23
Change format preview to use flexbox

70ray:
This should not make any visible difference. Previously the same
effect was achieved using javascript.
This is a simplification.
One user reported that the preview controls were not visible in the new Edge browser which prompted me to look at this but the problem subsequently went away spontaneously.


cpeel:
Nice update, Ray!

PR #330, opened 2020-06-25
Fix some html5 validation issues

chrismiceli:


pinc/theme.inc
```diff
@@ -184,7 +184,7 @@ function html_logobar() {
             $id = 'testbar';
         }
         if($alert_message)
-            $messages[] = $alert_message;
+            $messages[] = html_safe($alert_message);
```
cpeel, line 187: We allow HTML tags (like `<a>`) in `$alert_message` so we assume(!) that the admins who set the alert message make it HTML safe and should not escape it here.

faq/doc-copy.php
```diff
@@ -26,10 +23,10 @@
     Many people consider this very boring stuff but it is important that we get it
     right. <i>Thanks for proofreading it!</i>
 </p>
-<table border="0" cellspacing="0" width="100%">
+<table style="width: 100%">
```
cpeel, line 26: Oh geeze. Let's just replace this whole table with a simple `<hr>`.

faq/doc-copy.php
```diff
@@ -204,21 +199,18 @@
...
-        <a href="..">Distributed Proofreaders home page</a>,
-        &nbsp;&nbsp;&nbsp;
-        <a href="faq_central.php">DP FAQ Central page</a>,
-        &nbsp;&nbsp;&nbsp;
-        <a href="<?php echo $PG_home_url; ?>">Project Gutenberg home page</a>.
-        </font>
-    </td>
-</tr>
-</table>
+<div style='background-color: silver; width: 100%; text-align: center; font: 10px verdana, helvetica, sans-serif'>
```
cpeel, line 202: I would simply delete this entire block.

faq/doc-copy.php
```diff
@@ -4,15 +4,12 @@
 include_once($relPath.'pg.inc');
 include_once($relPath.'theme.inc');
 
-output_header('Proofreading Guidelines -- Copyright Renewal Books', NO_STATSBAR);
-?>
+$theme_args["css_data"] = "tt { color: red } pre { color: red }";
```
cpeel, line 7: `<tt>` isn't used anywhere in this file (anymore) so let's delete the style for it. I failed to clean that up in e06a364b1.

faq/doc-copy.php
```diff
@@ -4,15 +4,12 @@
...
 
-<style type='text/css'>
-    tt { color: red }
-    pre { color: red }
-</style>
+output_header('Proofreading Guidelines -- Copyright Renewal Books', NO_STATSBAR, $theme_args);
+?>
 
-<h2 align="center">Proofreading Guidelines -- Special Rules for Copyright Renewal Books</h2>
+<h2 style='text-align: center'>Proofreading Guidelines -- Special Rules for Copyright Renewal Books</h2>
```
cpeel, line 12: Let's remove the centering altogether.

pinc/ProjectSearchResults.inc
```diff
@@ -622,7 +622,7 @@ class ProjectSearchResults
         if(!empty($warning))
             echo "\n<p class='warning'>$warning</p>\n";
 
-        echo "<p id='head'>" . _("To set primary sort term, click on the column name; to set secondary sort term, click on the <u>^</u> that follows the column name.") . "</p>";
+        echo "<p id='head'>" . _("To set primary sort term, click on the column name; to set secondary sort term, click on the <span style='text-decoration: underline'>^</span> that follows the column name.") . "</p>";
```
srjfoo, line 625: If the `text-decoration: underline` is likely to be needed in multiple places, shouldn't it be defined in the site CSS somewhere?

tools/project_manager/remote_file_manager.php
```diff
@@ -214,7 +214,7 @@ function do_showdir($action_message)
     echo "<p>" . _("This page allows you to manage content in this uploads folder.") . "</p>\n";
 
     if (get_access_mode($pguser) == 'common') {
-        show_message('info', _("Your files are located in a common, shared, area.<br><u>Please take care to avoid affecting other users' files.</u>"));
+        show_message('info', _("Your files are located in a common, shared, area.<br><span style='text-decoration: underline'>Please take care to avoid affecting other users' files.</span>"));
```
srjfoo, line 217: ditto above

pinc/theme.inc
chrismiceli: I only noticed because "<<ALERT_MESSAGE>>" was in the default and it wasn't encoded and the validator complained, but that makes sense. I can change the default to &lt;&lt;...but probably overkill.

faq/doc-copy.php
chrismiceli: There are lots of places we use tables for layout that we probably shouldn't, but that's a different day.

pinc/ProjectSearchResults.inc
chrismiceli: That's possible, though I wasn't sure the standard for using css vs inline style. I see inline styles frequently so followed that approach.

pinc/theme.inc
cpeel: That value is replaced by `configure` as part of set-up. If it's still that value in your sandbox you might want to reconfigure your sandbox on TEST with:
```
SETUP/configure ~cpeel/bin/TEST_setup.sh .
```

If you're testing locally, check the updated `SETUP/configuration.sh` script.

pinc/ProjectSearchResults.inc
cpeel: We actually have a class for underline already as one of our mixins. You should be able to use:
```
<span class='underlined'>
```

chrismiceli: Nice, I missed that one somehow. I updated to use it here and in some other places. I hope it is ok.

PR #331, opened 2020-06-25
1836: Browser should enforce the input for required form fields

chrismiceli:
add required to report bad image reason

### DIFF
--- a/tools/proofers/report_bad_page.php
+++ b/tools/proofers/report_bad_page.php
@@ -50,9 +50,15 @@ if (!isset($_POST['submitted']) || $_POST['submitted'] != 'true')
     echo "<input type='hidden' name='submitted' value='true'>\n";
 
     echo "<p><b>" . _("Reason") . ":</b> ";
-    echo "<select name='reason'>";
+    echo "<select name='reason' required>";
     foreach($PAGE_BADNESS_REASONS as $i => $reason)
-        echo "<option value='$i'>$reason</option>";
+    {
+        if ($i == 0) {
+            echo "<option value=''></option>";
+        } else {
+            echo "<option value='$i'>$reason</option>";
+        }
+    }
     echo "</select>";
     echo "</p>";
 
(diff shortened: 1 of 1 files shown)


chrismiceli:
image

tools/proofers/report_bad_page.php
```diff
@@ -50,9 +50,15 @@
     echo "<input type='hidden' name='submitted' value='true'>\n";
 
     echo "<p><b>" . _("Reason") . ":</b> ";
-    echo "<select name='reason'>";
+    echo "<select name='reason' required>";
     foreach($PAGE_BADNESS_REASONS as $i => $reason)
-        echo "<option value='$i'>$reason</option>";
+    {
+        if ($i == 0) {
+            echo "<option value=''></option>";
```
cpeel, line 57: Can we do one-better and make this option disabled so they can't select it? I forget how browsers treat fields like that.

chrismiceli: I can remove the empty option and default to the first reason (Bad Image)
image
I can disable the selection for this option too as you suggest, but it is selected by default so it doesn't prevent submitting. Let me know what you think.

srjfoo: I definitely woiuldn't set one of the options as a default. We've seen what happens when we do that, and it's why we don't want a default value set. The user should be forced to choose one of the options. Can the Submit button be disabled unless an option is chosen?

Testing in the sandbox, showing the error message, and not allowing the report to be submitted unless a value is chosen seems to work fine.

chrismiceli: I can add that logic, but that would involve adding JavaScript and I think it achieves the same thing we have here with the required attribute. Let me know though if you want me to pursue that option.

cpeel: This is actually working great as-is. Thank you for fixing it!

PR #332, opened 2020-06-25
Log SQL errors to disk

cpeel:
Rather than printing SQL error messages to the user, which might reveal information we'd rather not show, log them to the `php_errors` file.

### DIFF
--- a/SETUP/check_security.sh
+++ b/SETUP/check_security.sh
@@ -1,0 +1,45 @@
+#!/bin/bash
+
+if [ "_$1" != "_" ]; then
+    BASE_DIR=$1
+elif [ -d SETUP ]; then
+    BASE_DIR=.
+elif [ -d ../SETUP ]; then
+    BASE_DIR=../
+else
+    echo "Unable to determine base code directory"
+    exit 1
+fi
+
+echo "Checking PHP files for known security issues..."
+
+for file in $(find $BASE_DIR -name "*.php" -o -name "*.inc"); do
+    # skip files in SETUP
+    echo $file | grep -q SETUP
+    if [[ $? -eq 0 ]]; then
+        continue;
+    fi
+
+    echo $file
+
+    # die(mysqli_error(
+    egrep -q 'die\s*\(\s*mysqli_error\s*\(' $file
+    if [[ $? -eq 0 ]]; then
+        echo "ERROR: found die() with mysqli_error output"
+        exit 1
+    fi
+
+    # echo mysqli_error(
+    egrep -q 'echo\s*mysqli_error\s*\(' $file
+    if [[ $? -eq 0 ]]; then
+        echo "ERROR: found echo with mysqli_error output"
+        exit 1
+    fi
+
+    # mysqli_error(
+    egrep -q 'mysqli_error\s*\(' $file
+    if [[ $? -eq 0 ]]; then
+        echo "WARNING: mysqli_error found - confirm it does not leak data to user"
+    fi
+
+done
--- a/pinc/DPDatabase.inc
+++ b/pinc/DPDatabase.inc
@@ -59,6 +59,17 @@ final class DPDatabase
         return self::$_connection;
     }
 
+    static public function log_error()
+    {
+        // Log the SQL error to the PHP error log and return a generic error
+        // for display to the user
+        $backtrace = debug_backtrace();
+        $caller = $backtrace[0]['file'] . ":" . $backtrace[0]['line'];
+        $error = str_replace("\n", "\\n", mysqli_error(DPDatabase::get_connection()));
+        error_log("DPDatabase.inc - log_error from $caller: $error");
+        return _("An error occurred during a database query and has been logged.");
+    }
+
     static public function get_default_db_charset()
     {
         $sql = sprintf("
--- a/pinc/DPDatabase.inc
+++ b/pinc/DPDatabase.inc
@@ -80,7 +80,7 @@ final class DPDatabase
 
         $result = mysqli_query(DPDatabase::get_connection(), $sql);
         if(!$result)
-            throw new DBQueryError(mysqli_error(DPDatabase::get_connection()));
+            throw new DBQueryError(DPDatabase::log_error());
 
         $row = mysqli_fetch_assoc($result);
         $default_charset = $row['DEFAULT_CHARACTER_SET_NAME'];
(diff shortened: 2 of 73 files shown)


pinc/DPDatabase.inc
```diff
@@ -59,6 +59,17 @@ final class DPDatabase
...
 
+    static public function log_error()
+    {
+        // Log the SQL error to the PHP error log and return a generic error
+        // for display to the user
+        $backtrace = debug_backtrace();
+        $caller = $backtrace[0]['file'] . ":" . $backtrace[0]['line'];
+        $error = str_replace("\n", "\\n", mysqli_error(DPDatabase::get_connection()));
+        error_log("DBDatabase.inc - log_error from $caller: $error");
+        return _("An error occurred during a DB query and has been logged.");
```
jmdyck, line 70:
```suggestion
        return _("An error occurred during a database query and has been logged.");
```

SETUP/install_db.php
```diff
@@ -18,10 +18,10 @@
 
 mysqli_query(DPDatabase::get_connection(), "
     CREATE DATABASE IF NOT EXISTS $db_name
-") or die(mysqli_error(DPDatabase::get_connection()));
+") or die(DPDatabase::log_error());
 mysqli_query(DPDatabase::get_connection(), "
     USE $db_name
-") or die(mysqli_error(DPDatabase::get_connection()));
+") or die(DPDatabase::log_error());
```
jmdyck, line 24: There's also the occurrence on line 49.

pinc/DPDatabase.inc
```diff
@@ -59,6 +59,17 @@ final class DPDatabase
...
     }
 
+    static public function log_error()
+    {
+        // Log the SQL error to the PHP error log and return a generic error
+        // for display to the user
+        $backtrace = debug_backtrace();
+        $caller = $backtrace[0]['file'] . ":" . $backtrace[0]['line'];
+        $error = str_replace("\n", "\\n", mysqli_error(DPDatabase::get_connection()));
+        error_log("DBDatabase.inc - log_error from $caller: $error");
```
jmdyck, line 69:
```suggestion
        error_log("DPDatabase.inc - log_error from $caller: $error");
```

pinc/projectinfo.inc
```diff
@@ -9,8 +9,7 @@ function Project_getNumPagesInState( $projectid, $page_state , $counter = "*")
     $res = mysqli_query(DPDatabase::get_connection(), "SELECT COUNT($counter) FROM $projectid WHERE state='$page_state'");
     if (!$res)
     {
-        echo mysqli_error(DPDatabase::get_connection());
-        return 0;
+        die(DPDatabase::log_error());
```
jmdyck, line 12: You're changing the behavior from `return 0` to `die`. I'd advise against, unless you've convinced yourself that it's appropriate.

pinc/projectinfo.inc
```diff
@@ -23,8 +22,7 @@ function Project_getNumPages( $projectid )
     $res = mysqli_query(DPDatabase::get_connection(), "SELECT COUNT(*) FROM $projectid");
     if (!$res)
     {
-        echo mysqli_error(DPDatabase::get_connection());
-        return 0;
+        die(DPDatabase::log_error());
```
jmdyck, line 25: Ditto above.

userprefs.php
```diff
@@ -446,7 +446,6 @@ function save_general_tab() {
 
     $userSettings->set_boolean('hide_special_colors', $_POST["show_special_colors"]=='no');
 
-    echo mysqli_error(DPDatabase::get_connection());
```
jmdyck, line 449: Why delete? We don't want an unconditional
   `echo DPDatabase::log_error();`
of course, but we could say
   `mysqli_query(...) or print(DPDatabase::log_error())`
to reproduce the current behavior.

SETUP/check_security.sh
```diff
@@ -0,0 +1,45 @@
...
+        continue;
+    fi
+
+    echo $file
+
+    # die(mysqli_error(
+    egrep -q 'die\s*\(\s*mysqli_error\s*\(' $file
+    if [[ $? -eq 0 ]]; then
+        echo "die() with mysqli_error output found."
+        exit 1
```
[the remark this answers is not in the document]
cpeel: See comment on other MR regarding consistency with other linting.

SETUP/install_db.php
cpeel: Ah - actually I want to roll back my changes and leave the original. I anticipate that the user will want the raw SQL error when running this script on the command line. Thanks for catching this.

pinc/projectinfo.inc
cpeel: Good call - fixed.

userprefs.php
cpeel: This one in particular is only going to print and return the error from the last DB call in the `$userSettings->set_bool()` which is arguably not what the user actually intended -- probably because I or someone else didn't clean this up after a refactor. I'll re-attach this one to the query call and fix the others.

jmdyck: I checked the history: the `mysql_query` got separated from the `mysql_error` way back on 2005-03-19, when the userSettings code was inserted.

SETUP/check_security.sh
```diff
@@ -0,0 +1,45 @@
...
+    if [[ $? -eq 0 ]]; then
+        continue;
+    fi
+
+    echo $file
+
+    # die(mysqli_error(
+    egrep -q 'die\s*\(\s*mysqli_error\s*\(' $file
+    if [[ $? -eq 0 ]]; then
+        echo "die() with mysqli_error output found."
```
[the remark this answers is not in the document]
cpeel: Fixed.

pinc/DPDatabase.inc
```diff
@@ -59,6 +59,17 @@ final class DPDatabase
         return self::$_connection;
     }
 
+    static public function log_error()
+    {
+        // Log the SQL error to the PHP error log and return a generic error
+        // for display to the user
+        $backtrace = debug_backtrace();
+        $caller = $backtrace[0]['file'] . ":" . $backtrace[0]['line'];
```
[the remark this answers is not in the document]
cpeel: I thought so too at first, but no:
```
[26-Jun-2020 18:38:21 UTC] DPDatabase.inc - log_error from /home/cpeel/u-cpeel-dp/pinc/Project.inc:31: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'asdf\n                FROM projects\n                WHERE projectid = 'projectID4' at line 1
```
^^ after intentionally borking the SQL in Project constructor

PR #333, opened 2020-06-25
Require login for various pages

cpeel:
Reduce our public-facing surface area by requiring log-in on various pages. In particular this MR changes all statistics pages to require the user to be logged in to view. It leaves the jpgraph files as the browser is expecting those to be images, not an HTML redirect.

The list in `SETUP/check_require_login.php` is going to be the authoritative source of PHP files that can be publicly accessed as it will be added to our CI/CD as an additional validation check after https://github.com/DistributedProofreaders/dproofreaders/pull/332 merges.

### DIFF
--- a/SETUP/check_require_login.php
+++ b/SETUP/check_require_login.php
@@ -1,0 +1,189 @@
+#!/usr/bin/env php
+<?php
+$relPath="../pinc/";
+include_once($relPath."misc.inc");
+
+# List of accepted unauthenticated files
+$ok_files = [
+    # Base website
+    "default.php",
+    "credits.php",
+    "sitemap.php",
+    "project.php",
+    "list_etexts.php",
+    "locale/debug_ui_language.php",
+    # RSS and XML feeds
+    "feeds/backend.php",
+    # Account login / registration pages
+    "accounts/addproofer.php",
+    "accounts/login_failure.php",
+    "accounts/activate.php",
+    "accounts/logout.php",
+    "accounts/require_login.php",
+    "accounts/login.php",
+    # FAQs
+    "faq/cp.php",
+    "faq/default.php",
+    "faq/dev_process.php",
+    "faq/doc-copy.php",
+    "faq/dochist.php",
+    "faq/dpf.php",
+    "faq/DPflow.php",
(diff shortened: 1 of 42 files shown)


SETUP/check_require_login.php
```diff
@@ -0,0 +1,186 @@
...
+
+function file_requires_auth($filename)
+{
+    $contents = file_get_contents($filename);
+
+    # "regular" pages use require_login();
+    $login = preg_match("/^require_login\(\);$/m", $contents);
+
+    # crontab pages use requester_is_localhost()
+    $localhost = preg_match("/requester_is_localhost\(\)/m", $contents);
```
jmdyck, line 183: Side thought: we should maybe define (say) `require_localhost_request()` as a parallel to `require_login()`.

SETUP/check_require_login.php
```diff
@@ -0,0 +1,186 @@
...
+    # If it's in the SETUP directory, skip it
+    if(startswith($file, "SETUP/"))
+        continue;
+
+    # If it's on the OK list, skip it
+    if(in_array($file, $ok_files))
+        continue;
+
+    echo "$file does not require authentication and is not on the known-good list\n";
+    exit(1);
```
jmdyck, line 156: Maybe abort after finding all, rather than after just the first.

cpeel: All of our linting checks currently stop after the first failure, otherwise it's hard to find the ones with the errors since it outputs all of the filenames. And while we could stop outputting the names unless there was an error, we'd only output the failures in a single check script instead of all of them and then stopping. Either way, for this and the other MR it's better to be consistent with our current lint check behaviour.

SETUP/check_require_login.php
cpeel: I like this idea a lot. Standby for update.

cpeel: And added.

SETUP/check_require_login.php
jmdyck: Ah, okay.

lhamilton1:
I approve

PR #337, opened 2020-06-30
Handle missing project tables better

cpeel:
`displayimage.php` and `images_index.php` return a database error if the project table for the requested project doesn't exist. This MR abstracts out querying for project pages into the Project class and raises appropriate exceptions if the project table or directory doesn't exist.

This is available in the [no-project-table](https://www.pgdp.org/~cpeel/c.branch/no-project-table) sandbox.

### DIFF
--- a/tools/proofers/images_index.php
+++ b/tools/proofers/images_index.php
@@ -14,63 +14,28 @@ $zip_type = get_enumerated_param( $_GET, 'zip_type', NULL, array('pages', 'illos
 
 $project = new Project($projectid);
 
-$image_index_str = _('Image Index');
-
-// Suppress display if we're streaming back a zipfile
-// so that we can control the http headers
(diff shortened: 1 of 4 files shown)


tools/proofers/images_index.php
```diff
@@ -79,13 +44,38 @@
 }
 else
 {
+    $image_index_str = _('Image Index');
+
+    output_header("$image_index_str: {$project->nameofwork}");
+
+    echo "
+        <h1>" . html_safe($project->nameofwork) . "</h1>
```
chrismiceli, line 52: Did we ever remerge my changes around sanitizing nameofwork throughout the project?

cpeel: Yes, that was merged into the Unicode branch before it merged into master.

PR #339, opened 2020-07-02
Replace DPCustomMono2 with DP Sans Mono

cpeel:
This replaces the DPCustomMono2 web font with DP Sans Mono, its spiritual successor. This is a straight replacement and all users who were using DPCM2 before will be using DPSM after this is rolled out without action from users.

This is available in the [dp-sans-mono](https://www.pgdp.org/~cpeel/c.branch/dp-sans-mono/) sandbox. Of note is the pared down version of the [font_sample](https://www.pgdp.org/~cpeel/c.branch/dp-sans-mono/faq/font_sample.php) page.

@70ray - I have noticed that this doesn't remove DPCustomMono from the Format Preview listing, presumably because it is loading the config from local storage. Can you take a look into what will be required to make that work? Users should be restricted to the set of fonts we specify, even if that set changes.


chrismiceli:
Looks good to me so far.

chrismiceli:
Should the font samples have some unicode in them?

faq/font_sample.php
```diff
@@ -85,16 +55,16 @@
...
+echo "<p style='clear: both;'>" . _("DP Sans Mono is a font adapted by DP volunteers, based on the suggestions and ideas of many experienced proofreaders. Using DP Sans Mono as your proofreading font can help you find mistakes."). "</p>\n";
 
-echo "<p style='clear: both'>" . _("This site provides DPCustomMono2 as a web font and browsers that support web fonts do not require the font to be installed locally.") . " ";
-echo sprintf(_("For more information on installing and using the font locally for offline activitlies like Post-Processing and Smooth Reading, see the <a href='%s'>DPCustomMono2</a> wiki page at pgdp.net."), "http://www.pgdp.net/wiki/Installing_DPCustomMono") . "</p>\n";
+echo "<p style='clear: both'>" . _("This site provides DP Sans Mono as a web font and browsers that support web fonts do not require the font to be installed locally.") . " ";
+echo sprintf(_("For more information on installing and using the font locally for offline activitlies like Post-Processing and Smooth Reading, see the <a href='%s'>DP Sans Mono</a> wiki page at pgdp.net."), "https://www.pgdp.net/wiki/DP_Sans_Mono") . "</p>\n";
```
srjfoo, line 63: The "[DPCustomMono2 Font](https://www.pgdp.net/wiki/DP_Official_Documentation:Proofreading/DPCustomMono2_Font)" wiki page is in the DP Official Documentation namespace; probably the `DP Sans Mono` page ought to be, too? (Or at least protected.)

styles/fonts/README.md
```diff
@@ -2,16 +2,16 @@
 
 This directory holds web fonts web make available to all DP users.
```
srjfoo, line 3: >This directory holds web fonts web make

[** we make?]

styles/fonts/details.json
```diff
@@ -0,0 +1,16 @@
...
+    {
+        "name": "DejaVu Sans Mono",
+        "url": "https://dejavu-fonts.github.io/",
+        "license": "DejaVu Fonts license",
+        "license_url": "https://dejavu-fonts.github.io/License.html",
+        "license_file": "DejaVu Fonts License.txt"
+    },
+    {
+        "name": "DP Sans Mono",
+        "url": "https://www.pgdp.net/wiki/DP_Sans_Mono",
```
srjfoo, line 11: Same as above -- if this page is not in the DP Official documentation, it should be protected.

70ray:
> Can you take a look into what will be required to make that work? Users should be restricted to the set of fonts we specify, even if that set changes.

We could make it work like the main preferences with 3 predefined and one user selectable option.

faq/font_sample.php
cpeel: I want a short URL so we'll go with this one (See also: https://www.pgdp.net/wiki/DPCustomMono2&redirect=no). I haven't decided if I want to make protected or a redirect to an DPOD yet but we can do one or the other.

cpeel:
> > Can you take a look into what will be required to make that work? Users should be restricted to the set of fonts we specify, even if that set changes.
> 
> We could make it work like the main preferences with 3 predefined and one user selectable option.

At this point I would like for it to enforce only the set of fonts we pass it -- adding a user-selectable font seems overkill (and if we went that route we should just reuse whatever font they have specified in their preferences and not let them select yet another font in Format Preview).

Can you please put together an MR that ensures that the fonts available in Format Preview are only those that we pass into Format Preview and if a user had selected a font that is no longer available it falls back to the initial default?

cpeel:
> Should the font samples have some unicode in them?

That's a good idea. I will noodle over which ones to pull in.

styles/fonts/README.md
cpeel: Fixed.

credits.php
```diff
@@ -8,33 +8,52 @@
 
 echo "<h1>$title</h1>";
 
+echo "<p>" . sprintf(_("This site is powered by the <a href='%s'>dproofreaders</a> source code and licensed under the GPL v2."), "https://github.com/DistributedProofreaders/dproofreaders") . "</p>";
+
 echo "<h2>" . _("Developers") . "</h2>";
 
-echo "<p>" . sprintf(_("This software brought to you by many volunteer developers across the world and years. Recent developer contribution after the move to git can be seen on the project's <a href='%s'>Github Contributors page</a>."), "https://github.com/DistributedProofreaders/dproofreaders/graphs/contributors"). "</p>";
+echo "<p>" . sprintf(_("This software brought to you by many volunteer developers across the world over many years. Recent developer contribution after the move to git can be seen on the project's <a href='%s'>Github Contributors page</a>."), "https://github.com/DistributedProofreaders/dproofreaders/graphs/contributors"). "</p>";
```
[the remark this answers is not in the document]
cpeel: Fixed!

70ray:
> > > Can you take a look into what will be required to make that work? Users should be restricted to the set of fonts we specify, even if that set changes.
> > 
> > 
> > We could make it work like the main preferences with 3 predefined and one user selectable option.
> 
> At this point I would like for it to enforce only the set of fonts we pass it -- adding a user-selectable font seems overkill (and if we went that route we should just reuse whatever font they have specified in their preferences and not let them select yet another font in Format Preview).
> 
> Can you please put together an MR that ensures that the fonts available in Format Preview are only those that we pass into Format Preview and if a user had selected a font that is no longer available it falls back to the initial default?

I've made a branch so the user can select from the same fonts available for the proofreading interface. I need to check it over a bit more before pushing it.

cpeel:
This now has the latest DPSM files from Laurent.

I wasn't able to come up with any additional unicode characters to include on the font sample page so I think this is ready to merge as-is after review.

styles/fonts/DP Sans Mono License.txt
```diff
@@ -1,38 +1,42 @@
...
 The above copyright and trademark notices and this permission notice
 shall be included in all copies of one or more of the Font Software
 typefaces.
 
 The Font Software may be modified, altered, or added to, and in
 particular the designs of glyphs or characters in the Fonts may be
-modified and additional glyphs or characters may be added to the
-Fonts, only if the fonts are renamed to names not containing either
-the words “Bitstream” or the word “Vera”.
+modified and additional glyphs or or characters may be added to the
```
srjfoo, line 24: "or or"?

cpeel: I fixed this. It's still wrong in the font but I'm not too worried about it.

lhamilton1:
I approve

PR #340, opened 2020-07-03
addProofer.js tests

chrismiceli:


cpeel:
Nice! Thanks for adding these unit tests!

PR #342, opened 2020-07-04
Preview font

70ray:
The indentation of the changes to button_menu.inc is wrong but it is all so chaotic anyway I haven't changed it.


70ray:
Trying a way to do it withou such wide-ranging changes.

PR #344, opened 2020-07-05
Reorganized Basic Latin pickersets

srjfoo:
Added florin sign to the Basic Latin character suite.

Removed micro sign and capital and small letter S and Z with caron
from pickersets. Reorganized pickersets.

Added title text display to pickerset display on character suite
details pages.

The changes to the picker sets were organized by DP user 
bunny-crunch and reviewed by the DP community.

### DIFF
--- a/pinc/charsuite-basic-latin.inc
+++ b/pinc/charsuite-basic-latin.inc
@@ -6,7 +6,7 @@ $charsuite->codepoints = [
     # https://en.wikipedia.org/wiki/Basic_Latin_(Unicode_block)
     'U+0020-U+007e',
     # https://en.wikipedia.org/wiki/Latin-1_Supplement_(Unicode_block)
-    # without the soft hyphen and micro sign
+    # without the soft hyphen
     'U+00a1-U+00ac',
     'U+00ae-U+00ff',
     # codepoints that map to Windows-1252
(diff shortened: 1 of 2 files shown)


pinc/charsuite-basic-latin.inc
```diff
@@ -30,26 +31,26 @@ $pickerset->add_subset("A-Z", [
...
-$pickerset->add_subset(utf8_chr("U+00c0"), [
-    [ 'U+00c0-U+00c7', 'U+00d0', 'U+00c8-U+00cf', 'U+00d1' ],
-    [ 'U+00e0-U+00e7', 'U+00f0', 'U+00e8-U+00ef', 'U+00f1' ],
-]);
-$pickerset->add_subset(utf8_chr("U+00d2"), [
-    [ 'U+00d2-U+00d6', 'U+00d8', 'U+0152', 'U+0160', NULL, 'U+00d9-U+00dd', 'U+00de-U+00de', 'U+0178', 'U+017d' ],
-    [ 'U+00f2-U+00f6', 'U+00f8', 'U+0153', 'U+0161', 'U+00df', 'U+00f9-U+00fd', 'U+00fe-U+00ff', 'U+017e' ],
-]);
+$pickerset->add_subset("ˆ", [
+    [ 'U+00c1', 'U+00c9', 'U+00cd', 'U+00d3', 'U+00da', 'U+00dd', 'U+00c0', 'U+00c8', 'U+00cc', 'U+00d2', 'U+00d9', 'U+00c2', 'U+00ca', 'U+00ce', 'U+00d4', 'U+00db' ],
```
cpeel, line 35: I did a poor job of this in the original suite, but keeping the lines a bit shorter makes it easier to edit. The `charsuite-polytonic-greek.inc` does a much better job of this.

I don't see any need to change the MR, but something to think about in the future.

pinc/charsuite-basic-latin.inc
```diff
@@ -6,7 +6,7 @@ $charsuite->codepoints = [
     # https://en.wikipedia.org/wiki/Basic_Latin_(Unicode_block)
     'U+0020-U+007e',
     # https://en.wikipedia.org/wiki/Latin-1_Supplement_(Unicode_block)
-    # without the soft hyphen
+    # without the soft hyphen and micro sign
```
srjfoo, line 9: Need to remove reference to removing micro sign from character suite -- it was only removed from the pickers.

pinc/charsuite-basic-latin.inc
srjfoo: I noticed that the lines in EEL are shorter, and followed that pattern there. Good idea.

srjfoo: Done.

pinc/charsuite-basic-latin.inc
srjfoo: Done.

PR #346, opened 2020-07-09
PPV report: Change lines re cover image requirements

jmdyck:
(as requested by pgdp.net GM)


lhamilton1:
I approve moving this to production once we do a release

PR #347, opened 2020-07-09
Improve quiz query escaping

cpeel:
This improves escaping for the quiz queries. It does so by introducing two new static methods in `DPDatabase` that wrap some common functionality. I wanted to get a wider set of eyes on it to see if these look reasonable to build off of elsewhere in the code before pulling them in.

Notably we can now escape queries in SQL with:
```php
DPDatabase::escape($value)
```
instead of:
```php
mysqli_real_escape_string(DPDatabase::get_connection(), $value)
```

And the new `DPDatabase::query($sql)` method will automatically log query errors and optionally raise an exception if the query fails.

So we could use this:
```php
```
instead of:
```php
$result = mysqli_query(DPDatabase::get_connection(), $sql) or die(DPDatabase::log_error());
```

### DIFF
--- a/pinc/DPDatabase.inc
+++ b/pinc/DPDatabase.inc
@@ -59,12 +59,32 @@ final class DPDatabase
         return self::$_connection;
     }
 
-    static public function log_error()
+    static public function escape($value)
+    {
+        return mysqli_real_escape_string(self::$_connection, $value);
+    }
+
+    static public function query($sql, $throw_on_failure=TRUE)
+    {
+        $result = mysqli_query(self::$_connection, $sql);
+        if(!$result)
+        {
+            // include this function's caller in the backtrace
+            $error = self::log_error(1);
+            if($throw_on_failure)
+            {
+                throw new DBQueryError($error);
+            }
+        }
+        return $result;
+    }
+
+    static public function log_error($backtrace_level=0)
     {
         // Log the SQL error to the PHP error log and return a generic error
         // for display to the user
         $backtrace = debug_backtrace();
-        $caller = $backtrace[0]['file'] . ":" . $backtrace[0]['line'];
+        $caller = $backtrace[$backtrace_level]['file'] . ":" . $backtrace[$backtrace_level]['line'];
         $error = str_replace("\n", "\\n", mysqli_error(DPDatabase::get_connection()));
         error_log("DPDatabase.inc - log_error from $caller: $error");
         return _("An error occurred during a database query and has been logged.");
--- a/pinc/Quiz.inc
+++ b/pinc/Quiz.inc
@@ -68,10 +68,12 @@ class Quiz
             $pages_required_results[$quiz_page_id] = "no";
         }
 
-        $result = mysqli_query(DPDatabase::get_connection(), "
-            SELECT * FROM quiz_passes
-            WHERE username='$username'
-        ");
+        $sql = sprintf("
+            SELECT *
+            FROM quiz_passes
+            WHERE username = '%s'
+        ", DPDatabase::escape($username));
+        $result = DPDatabase::query($sql);
         while ($attempt = mysqli_fetch_object($result))
         {
             if($attempt->result != "pass") continue;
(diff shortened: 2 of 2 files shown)


cpeel:
> Looking around it seems some people recommend using PDO. I don't know if this is relevant. I believe BookBuff knows about sql.

PDO is being recommended because of the use of prepared statements which bypasses the need for escaping. mysqli supports [prepared statements](https://www.php.net/manual/en/mysqli.quickstart.prepared-statements.php) directly but it introduces a lot of additional overhead both for the developer and the database (every query turns into effectively two queries -- see the bottom of the prior link). If we were developing DP from scratch it would probably be the way to go, but I'm wary of trying to retrofit such a different DB interaction pattern at this point.

pinc/DPDatabase.inc
```diff
@@ -59,12 +59,32 @@ final class DPDatabase
         return self::$_connection;
     }
 
-    static public function log_error()
+    static public function escape($value)
+    {
+        return mysqli_real_escape_string(self::$_connection, $value);
+    }
+
+    static public function query($sql, $except_on_failure=TRUE)
```
jmdyck, line 67:
```suggestion
    static public function query($sql, $exception_on_failure=TRUE)
```
... because "except" could be read as "Do something, *except* when there's a failure."

Or maybe change it to `$throw_on_failure`.

pinc/Quiz.inc
```diff
@@ -240,36 +240,50 @@ function get_Quiz_containing_page($quiz_page_id)
...
     else
     {
-        mysqli_query(DPDatabase::get_connection(), "
+        $sql = sprintf("
             INSERT INTO quiz_passes
-            VALUES('$username', '$now', '$quiz_page_id', '$result')
-        ");
+            VALUES ('%s', %d, '%s', '%s')
+        ", time(),
+            DPDatabase::escape($username),
```
jmdyck, line 269: Swap these 2 args.

pinc/DPDatabase.inc
cpeel: I like `$throw_on_failure` better, good idea.

pinc/Quiz.inc
```diff
@@ -284,10 +300,12 @@ function user_has_passed_quiz_page($username, $quiz_page_id, $pass_requirements
 
 function get_last_attempt_date_for_quiz_page($username, $quiz_page_id)
 {
-    $res = mysqli_query(DPDatabase::get_connection(), "
-        SELECT date FROM quiz_passes
-        WHERE username='$username' AND quiz_page='$quiz_page_id'
-    ");
+    $sql = sprintf("
+        SELECT *
```
srjfoo, line 304: Showing my ignorance, here, but is there a reason for changing from `SELECT date` to `SELECT *` here? (Or does it not matter?)

cpeel: It doesn't actually matter since we fetch the column by name from the row. Fetching the whole row has the slight benefit that we might pull it from the MySQL query cache since the query is identical to the other queries in this file.

pinc/DPDatabase.inc
```diff
@@ -59,12 +59,32 @@ final class DPDatabase
         return self::$_connection;
     }
 
-    static public function log_error()
+    static public function escape($value)
```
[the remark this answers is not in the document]
cpeel: This is a good idea and I'm bummed that it's a few months late. I'm loath to change it at this time. One way to think about this though is that we are escaping in a DPDatabase namespace (since it's a static function) and the only escaping one would do for a database is for SQL.

pinc/Quiz.inc
```diff
@@ -240,36 +240,50 @@ function get_Quiz_containing_page($quiz_page_id)
...
-    $res = mysqli_query(DPDatabase::get_connection(), "
-        SELECT * FROM quiz_passes
-        WHERE username='$username' AND quiz_page='$quiz_page_id'
-    ");
-    $now = time();
+    $sql = sprintf("
+        SELECT *
+        FROM quiz_passes
+        WHERE username = '%s' AND quiz_page = '%s'
+    ", DPDatabase::escape($username), DPDatabase::escape($quiz_page_id));
```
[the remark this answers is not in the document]
cpeel: Yes, we are very inconsistent. While we can know that all real usernames do not need to be encoded, we might have usernames passed to us from the UI and we can't guarantee that they are all safe. We should generally err on the side of caution and always escape strings going into SQL.

pinc/DPDatabase.inc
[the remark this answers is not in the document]
chrismiceli: Yeah, I don't think people will get confused using it given the scoped naming.

PR #348, opened 2020-07-12
1894: Do not convert [,x] to character with cedilla.

chrismiceli:
Update code that handles this


chrismiceli:
Note, we'd need to update the guidelines here:
https://www.pgdp.net/wiki/DP_Official_Documentation:Proofreading/Proofreading_Guidelines#Accented.2C_Diacritical.2C_and_Non-ASCII_Characters
but I don't have permission (rightfully so)
available in sandbox here: https://www.pgdp.org/~chrismiceli/c

cpeel:
This LGTM but I'd appreciate @70ray's eyes on it.

PR #349, opened 2020-07-13
1880: Move Post-Processing Projects from the basement of every My Pro…

chrismiceli:
…jects tab

Add quick jump links to my projects page.

### DIFF
--- a/styles/layout.css
+++ b/styles/layout.css
@@ -201,10 +201,16 @@ footer small,
   display: table-cell;
 }
 .quick-links {
-  padding: 0.5em 0;
+  padding: 0;
 }
-.quick-links .quick-links-divider {
-  padding: 0 5px;
+.quick-links li {
+  display: inline;
+}
+.quick-links li::before {
+  content: " | ";
+}
+.quick-links li:first-child::before {
+  content: "";
 }
 /*
  If the screen is "narrow enough" we move the statsbar below the page
--- a/styles/layout.css
+++ b/styles/layout.css
@@ -200,6 +200,12 @@ footer small,
   padding: 0 0.5em;
   display: table-cell;
 }
+.quick-links {
+  padding: 0.5em 0;
+}
+.quick-links .quick-links-divider {
+  padding: 0 5px;
+}
 /*
  If the screen is "narrow enough" we move the statsbar below the page
  content using a flex layout. We don't use a flex layout for the
--- a/styles/layout.less
+++ b/styles/layout.less
@@ -154,6 +154,14 @@ footer, #footer {
     display: table-cell;
 }
 
+.quick-links {
+    padding: .5em 0;
+
+    .quick-links-divider {
+        padding: 0 5px;
+    }
+}
+
 /*
  If the screen is "narrow enough" we move the statsbar below the page
  content using a flex layout. We don't use a flex layout for the
--- a/styles/layout.less
+++ b/styles/layout.less
@@ -155,10 +155,16 @@ footer, #footer {
 }
 
 .quick-links {
-    padding: .5em 0;
+    padding: 0;
+    li {
+        display: inline;
+        &::before {
+            content: " | ";
+        }
 
-    .quick-links-divider {
-        padding: 0 5px;
+        &:first-child::before {
+            content: "";
+        }
     }
 }
 
--- a/styles/style_demo.php
+++ b/styles/style_demo.php
@@ -9,16 +9,26 @@ output_header("Style Demos");
 
 <h1>Style Demos</h1>
 
+<div class='quick-links'>
+  <a href='#headings'>Headings</a>
+  <span class='quick-links-divider'>|</span>
+  <a href='#paragraphs'>Paragraphs</a>
+  <span class='quick-links-divider'>|</span>
+  <a href='#links'>Links</a>
+  <span class='quick-links-divider'>|</span>
+  <a href='#table-formats'>Table formats</a>
+</div>
+
 <p>This page demonstrates how common page elements are rendered in the current theme. Logged-in users can change their theme in their <a href='../userprefs.php'>Preferences</a>.</p>
 
-<h2>Headings</h2>
+<h2 id='headings'>Headings</h2>
 <h1>Heading 1</h1>
 <h2>Heading 2</h2>
 <h3>Heading 3</h3>
 <h4>Heading 4</h4>
 <h5>Heading 5</h5>
 
-<h2>Paragraphs</h2>
+<h2 id='paragraphs'>Paragraphs</h2>
 <p>This is a paragraph of standard text.</p>
 
 <p>This is a long paragraph of standard text. Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aliquam in felis eget lectus volutpat imperdiet eget accumsan urna. Duis vitae dictum justo. Sed blandit massa sed auctor convallis. Donec lobortis tortor eget lectus auctor eleifend. Nulla dignissim tristique aliquam. Class aptent taciti sociosqu ad litora torquent per conubia nostra, per inceptos himenaeos. Proin pellentesque purus eu nisl varius tempor. Integer consequat pharetra velit, a lobortis enim. Mauris nisi risus, euismod ut mollis sit amet, convallis sit amet nulla. Sed molestie nisi et tellus luctus, a dictum eros pulvinar.</p>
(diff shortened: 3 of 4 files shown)


srjfoo:
It doesn't really make sense to me to have a jump to link to Proofreading & Formatting Projects at the top of the page, when the Proofreading & Formatting heading is immediately below the jump-to section. It makes more sense to have it down below. Something like this:

image

cpeel:
> It doesn't really make sense to me to have a jump to link to Proofreading & Formatting Projects at the top of the page, when the Proofreading & Formatting heading is immediately below the jump-to section.

And I don't really like that because of the precedent it sets -- if we have 3 sections on a page do we put the 2 other links under every header? What if we have 5 sections?

Stepping back a bit: while I thought we did, it doesn't look like any of our pages use this type of intra-page link (we have anchors mid-page, but we don't have links to them on the page). Whatever we define here needs to work for other pages and we should include it in the Style Guide. I'm open to design suggestions and it might be worth looking at how other sites do it.

In a past life a site I worked on used this for intra-page links:
```
Header Title 1 | Header Title 2 | Header Title 3
```

@chrismiceli - sorry this opened up into a bit of a broader design issue, although I think this is a very valuable exercise.

chrismiceli:
> 
> 
> > It doesn't really make sense to me to have a jump to link to Proofreading & Formatting Projects at the top of the page, when the Proofreading & Formatting heading is immediately below the jump-to section.
> 
> And I don't really like that because of the precedent it sets -- if we have 3 sections on a page do we put the 2 other links under every header? What if we have 5 sections?
> 
> Stepping back a bit: while I thought we did, it doesn't look like any of our pages use this type of intra-page link (we have anchors mid-page, but we don't have links to them on the page). Whatever we define here needs to work for other pages and we should include it in the Style Guide. I'm open to design suggestions and it might be worth looking at how other sites do it.
> 
> In a past life a site I worked on used this for intra-page links:
> 
> ```
> Header Title 1 | Header Title 2 | Header Title 3
> ```
> 
> @chrismiceli - sorry this opened up into a bit of a broader design issue, although I think this is a very valuable exercise.

I quite like this approach. I'll try it out tonight.

chrismiceli:
> 
> 
> It doesn't really make sense to me to have a jump to link to Proofreading & Formatting Projects at the top of the page, when the Proofreading & Formatting heading is immediately below the jump-to section. It makes more sense to have it down below. Something like this:
> 
> image



> 
> 
> It doesn't really make sense to me to have a jump to link to Proofreading & Formatting Projects at the top of the page, when the Proofreading & Formatting heading is immediately below the jump-to section. It makes more sense to have it down below. Something like this:
> 
> image

There may also be more content between them. We define some callouts. Currently they are all geared towards new proofers, but still something to think about if we ever expanded in the future.

cpeel:
Some other thoughts: I know some sites have ways to quickly jump back to the top of the page, which seems like it could be useful. I like how Github includes the heading anchor link on hover-over for Markdown files. Not sure if these are ideas or things we want to incorporate into DP but they seem applicable to the discussion at least.

srjfoo:
> > > It doesn't really make sense to me to have a jump to link to Proofreading & Formatting Projects at the top of the page, when the Proofreading & Formatting heading is immediately below the jump-to section.
> > 
> > 
> > And I don't really like that because of the precedent it sets -- if we have 3 sections on a page do we put the 2 other links under every header? What if we have 5 sections?

Yes, that could be a problem.

> > Stepping back a bit: while I thought we did, it doesn't look like any of our pages use this type of intra-page link (we have anchors mid-page, but we don't have links to them on the page). Whatever we define here needs to work for other pages and we should include it in the Style Guide. I'm open to design suggestions and it might be worth looking at how other sites do it.
> > In a past life a site I worked on used this for intra-page links:
> > ```
> > Header Title 1 | Header Title 2 | Header Title 3
> > ```
> > 
> > 
> > @chrismiceli - sorry this opened up into a bit of a broader design issue, although I think this is a very valuable exercise.
> 
> I quite like this approach. I'll try it out tonight.

I like it, too. One thing that should be taken into account, though, is navigation elsewhere on the page. That would become even more of an issue if there were 4 or 5 headings.  Is
```
Header Title 1 | Header Title 2 | Header Title 3
```
included below each header? Do all the headers have a
```
Back to Top
```
link? Or are the links one way, and we expect the user to get themselves back to the top of the page?

cpeel:
> Is
> 
> ```
> Header Title 1 | Header Title 2 | Header Title 3
> ```
> 
> included below each header?

No, just once at the top. I expect the most common use case is getting to data in the page, not getting to the top [although we cross-posted about that issue 😁 ].

chrismiceli:
> > Is
> > ```
> > Header Title 1 | Header Title 2 | Header Title 3
> > ```
> > 
> > 
> > included below each header?
> 
> No, just once at the top. I expect the most common use case is getting to data in the page, not getting to the top [although we cross-posted about that issue 😁 ].

Here is the latest which is available in sandbox (https://www.pgdp.org/~chrismiceli/c):
with callout:
image

without callout:
image

tools/proofers/my_projects.php
```diff
@@ -83,6 +83,18 @@
 
 echo "<h1>" . get_usertext($page_header) . "</h1>";
 
+$can_view_post_processing = in_array("PP", $allowed_stages) or in_array("PPV", $allowed_stages);
```
cpeel, line 86: `$allowed_stages` isn't defined until line 249.

tools/proofers/my_projects.php
```diff
@@ -83,6 +83,18 @@
...
+$proof_heading = _("Proofreading & Formatting Projects");
+$pool_heading = _("Post-Processing Projects");
+if ($can_view_post_processing)
+{
+    echo "<div style='padding: .5em 0'>";
```
cpeel, line 91: We should make this a first-class style and add it to the main style sheet in `layout.less`. I would also like to see this element defined in a way that we don't have to specify the vertical bar, something like:
```
<div class='quicklinks'>
<a href=...>Link1</a>
<a href=...>Link2</a>
</div>
```
That should be doable with the right CSS. I don't really care too much if we need to define this as a `<ul>` instead of a `<div>`.

We should add this to the `styles/style_demo.php` page as well.

tools/proofers/my_projects.php
chrismiceli: I was testing by commenting out this line :/

tools/proofers/my_projects.php
chrismiceli: Here it is on the demo page:
image

styles/layout.css
```diff
@@ -200,6 +200,12 @@ footer small,
   padding: 0 0.5em;
   display: table-cell;
 }
+.quick-links {
+  padding: 0.5em 0;
```
srjfoo, line 204: Here and in `layout.less`, I'm curious about why padding is used instead of margin (mind you, I'm totally not an expert). I realize that we want some space top and bottom, but padding -- even a half em top and bottom -- adds empty space and is not collapsible.

I don't think the call-out box should be a concern; at least not the one for newbies. Except on the test site, newbies with fewer than 100 pages proofed are extremely unlikely to ever see the links, as they won't have proofed enough pages to have access to Post-Processing.

styles/layout.less
```diff
@@ -154,6 +154,14 @@ footer, #footer {
     display: table-cell;
 }
 
+.quick-links {
```
cpeel, line 157: I was rather thinking something like this:
CSS:
```css
    ul.quick-links {
        padding: 0;
    }
    ul.quick-links li {
        display: inline;
    }
    ul.quick-links li::before {
        content: " | ";
    }
    ul.quick-links li:first-child::before {
        content: "";
    }
```
HTML:
```html
    <ul class='quick-links'>
        <li><a href='blah'>Heading 1</a></li>
        <li><a href='blah'>Heading 2</a></li>
        <li><a href='blah'>Heading 3</a></li>
    </ul>
    <h1>Heading 1</h1>
```
That way we can style and the separator (which is really only textual styling) independent of the code.

styles/style_demo.php
```diff
@@ -9,16 +9,26 @@
 
 <h1>Style Demos</h1>
 
+<div class='quick-links'>
```
cpeel, line 12: This is great, although I was thinking more in the "here's how to do quick links" sense. I think all you need to do is add an `<h2>Quick Links</h2>` above this line.

styles/layout.less
chrismiceli: I like this better. I usually avoid ::before due to accessibility concerns, but here it is something we don't want screen readers reading anyways.

styles/layout.css
chrismiceli: using margin now indirectly through ul/li tags now.

styles/style_demo.php
chrismiceli: Added and moved below introductory text explaining the page.

cpeel:
This looks great to me Chris!

PR #350, opened 2020-07-14
View page image

70ray:
Construct displayimsge and view_page_image using javascript.

### DIFF
--- a/pinc/page_controls.inc
+++ b/pinc/page_controls.inc
@@ -85,7 +85,7 @@ function get_proofreading_interface_data_js()
         "faces" => get_font_styles(),
         "faceFamilies" => get_full_font_families(),
         "sizes" => get_font_sizes(),
-        "sizeFamilies" => get_font_size_families(),
+        "sizeFamilies" => get_font_sizes_css(),
     ];
 
     $strings = [
--- a/pinc/page_controls.inc
+++ b/pinc/page_controls.inc
@@ -2,21 +2,20 @@
 include_once($relPath.'stages.inc'); // $Round_for_round_id_
 include_once($relPath.'metarefresh.inc');
 
-function get_page_data_js($projectid, $page, $round_id)
+function get_page_data_js($projectid, $page, $round_id, $error_message)
 {
     global $code_url;
 
     $heading = "";
     $text = "";
-    $error_message = "";
     $page_names = array();
     $image_url = "";
 
-    if($projectid)
+    if(!$error_message && $projectid)
     {
         try
         {
-            $projectid = validate_projectID('projectid', $projectid); // throws InvalidProjectIDException
+            // $projectid has been validated before we get here
             $project = new Project($projectid); // throws NonexistentProjectException
 
             $page_names = $project->get_page_names_from_db();
@@ -25,23 +24,8 @@ function get_page_data_js($projectid, $page, $round_id)
                 throw new UnexpectedValueException(sprintf(_("There are no pages in %s"), $projectid));
             }
 
-            // page names are sorted so will find first match
-            $page_match = false;
-            if($page !== "")
+            if(!in_array($page, $page_names))
             {
-                foreach($page_names as $page_name)
-                {
-                    if(1 === preg_match("/.*$page.*/", $page_name))
-                    {
-                        $page = $page_name;
-                        $page_match = true;
-                        break;
-                    }
-                }
-            }
-            if(!$page_match)
-            {
-                // no data given or couldn't find a match
                 $page = $page_names[0];
             }
 
@@ -59,7 +43,7 @@ function get_page_data_js($projectid, $page, $round_id)
             $result = DPDatabase::query("SELECT $text_column_name FROM $projectid WHERE image = '$esc_page'");
             $row = mysqli_fetch_assoc($result);
             $text = $row[$text_column_name];
-            $image_url = $project->url/$page;
+            $image_url = "$project->url/$page";
         }
         catch(Exception $exception)
         {
--- a/pinc/page_controls.inc
+++ b/pinc/page_controls.inc
@@ -11,7 +11,7 @@ function get_page_data_js($projectid, $page, $round_id, $error_message)
     $page_names = array();
     $image_url = "";
 
-    if(!$error_message && $projectid)
+    if($projectid)
     {
         try
         {
@@ -24,7 +24,8 @@ function get_page_data_js($projectid, $page, $round_id, $error_message)
                 throw new UnexpectedValueException(sprintf(_("There are no pages in %s"), $projectid));
             }
 
-            if(!in_array($page, $page_names))
+            // if page is not null check if it exists else select first page
+            if($page && !in_array($page, $page_names))
             {
                 $page = $page_names[0];
             }
@@ -39,16 +40,18 @@ function get_page_data_js($projectid, $page, $round_id, $error_message)
             {
                 $text_column_name = get_Round_for_round_id($round_id)->text_column_name;
             }
-            $esc_page = DPDatabase::escape($page);
-            $result = DPDatabase::query("SELECT $text_column_name FROM $projectid WHERE image = '$esc_page'");
-            $row = mysqli_fetch_assoc($result);
-            $text = $row[$text_column_name];
-            $image_url = "$project->url/$page";
+            if($page)
+            {
+                $esc_page = DPDatabase::escape($page);
+                $result = DPDatabase::query("SELECT $text_column_name FROM $projectid WHERE image = '$esc_page'");
+                $row = mysqli_fetch_assoc($result);
+                $text = $row[$text_column_name];
+                $image_url = "$project->url/$page";
+            }
         }
         catch(Exception $exception)
         {
             $error_message = html_safe($exception->getMessage());
-            $projectid = "";
         }
     }
 
@@ -105,6 +108,7 @@ function get_proofreading_interface_data_js()
         "optional" => _("(optional)"),
         "reset" => _("Select a different project"),
         "projectid" => _("Project ID"),
+        "selectAPage" => _("Please select a page"),
     ];
 
     global $code_url;
--- a/pinc/page_controls.inc
+++ b/pinc/page_controls.inc
@@ -10,6 +10,7 @@ function get_page_data_js($projectid, $page, $round_id)
     $text = "";
     $error_message = "";
     $page_names = array();
+    $image_url = "";
 
     if($projectid)
     {
@@ -58,6 +59,7 @@ function get_page_data_js($projectid, $page, $round_id)
             $result = DPDatabase::query("SELECT $text_column_name FROM $projectid WHERE image = '$esc_page'");
             $row = mysqli_fetch_assoc($result);
             $text = $row[$text_column_name];
+            $image_url = $project->url/$page;
         }
         catch(Exception $exception)
         {
@@ -74,7 +76,7 @@ function get_page_data_js($projectid, $page, $round_id)
         "text" => $text,
         "heading" => $heading,
         "errorMessage" => $error_message,
-        "imageUrl" => "$project->url/$page"
+        "imageUrl" => "$image_url"
     ];
     $page_browser_data = json_encode($page_browser_data);
 
--- a/pinc/page_controls.inc
+++ b/pinc/page_controls.inc
@@ -1,0 +1,138 @@
+<?php
+include_once($relPath.'stages.inc'); // $Round_for_round_id_
+include_once($relPath.'metarefresh.inc');
+
+function get_page_data_js($projectid, $page, $round_id)
+{
+    global $code_url;
+
+    $heading = "";
+    $text = "";
+    $error_message = "";
+    $page_names = array();
+
+    if($projectid)
+    {
+        try
+        {
+            $projectid = validate_projectID('projectid', $projectid); // throws InvalidProjectIDException
+            $project = new Project($projectid); // throws NonexistentProjectException
+
+            $page_names = $project->get_page_names_from_db();
+            if(!$page_names)
+            {
+                throw new UnexpectedValueException(sprintf(_("There are no pages in %s"), $projectid));
+            }
+
+            // page names are sorted so will find first match
+            $page_match = false;
+            if($page !== "")
+            {
+                foreach($page_names as $page_name)
+                {
+                    if(1 === preg_match("/.*$page.*/", $page_name))
+                    {
+                        $page = $page_name;
+                        $page_match = true;
+                        break;
+                    }
+                }
+            }
+            if(!$page_match)
+            {
+                // no data given or couldn't find a match
+                $page = $page_names[0];
+            }
+
+            $heading = html_safe($project->nameofwork) . "<a href='$code_url/project.php?id=$projectid'>" . _("Go to Project Page") . "</a>";
+
+            if ($round_id == "OCR")
+            {
+                $text_column_name = 'master_text';
+            }
+            else
+            {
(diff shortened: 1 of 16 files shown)


pinc/page_controls.inc
```diff
@@ -0,0 +1,152 @@
...
+    $error_message = "";
+
+    if($projectid)
+    {
+        try
+        {
+            $projectid = validate_projectID('projectid', $projectid); // throws InvalidProjectIDException
+            $project = new Project($projectid); // throws NonexistentProjectException
+
+            $result = DPDatabase::query("SELECT image FROM {$projectid} ORDER BY image ASC");
```
cpeel, line 22: You can now replace this whole block with:
```php
$images = $project->get_page_names_from_db();
```

pinc/page_controls.inc
```diff
@@ -0,0 +1,152 @@
+<?php
+include_once($relPath.'stages.inc'); // $Round_for_round_id_
+include_once($relPath.'metarefresh.inc');
+
+function get_page_data()
+{
+    $projectid = array_get($_GET, "project", "");
```
cpeel, line 7: These should all be arguments to `get_page_data()`. Ideally only the top-level `.php` pages should be accessing the `$_GET` and `$_POST` variables.

styles/global.less
```diff
@@ -126,6 +126,14 @@
     display: none;
 }
 
+.overflow-auto {
+    overflow:auto;
```
cpeel, line 130: Space after `:`.

tools/proofers/previewControl.js
```diff
@@ -168,12 +168,9 @@ $( function() {
         // set up the font selector
         Object.keys(fontStyles).forEach(function(index) {
             let fontStyle = fontStyles[index];
-            if(fontStyle) {
-                // do not add empty string
-                let selected = (index === previewStyles.defFontIndex);
-                let option = new Option(fontStyle, index, selected, selected);
-                $fontSelector[0].add(option, null);
-            }
+            let selected = (index === previewStyles.defFontIndex);
```
cpeel, line 171: This doesn't belong in this MR since it's unrelated to the image control.

pinc/page_controls.inc
```diff
@@ -0,0 +1,152 @@
...
+            {
+                $images[] = $row["image"];
+            }
+            mysqli_free_result($result);
+
+            if($page !== "")
+            {
+                // find a match for the page
+                $esc_page = DPDatabase::escape($page);
+                $result = DPDatabase::query("SELECT image FROM $projectid WHERE image LIKE '%$esc_page%'");
```
cpeel, line 37: Why are we doing a substring search here? Shouldn't the image name always be one of the ones we've defined? Seems like this would just be:
```
if(!in_array($page, $images))
{
    $page = $images[0];
}
```

pinc/page_controls.inc
```diff
@@ -0,0 +1,152 @@
+<?php
+include_once($relPath.'stages.inc'); // $Round_for_round_id_
+include_once($relPath.'metarefresh.inc');
+
+function get_page_data()
```
cpeel, line 5: Since this is returning a JS string and not an object, `get_page_data_js()` would be more accurate.

pinc/page_controls.inc
```diff
@@ -0,0 +1,152 @@
...
+        "errorMessage" => $error_message,
+    ];
+    $page_browser_data = json_encode($page_browser_data);
+
+    return "var pageBrowserData = $page_browser_data;";
+}
+
+function get_mode()
+{
+    $mode = get_enumerated_param($_GET, 'mode', 'image', ['image', 'text', 'imageText']);
```
cpeel, line 92: Ditto earlier comment that this file shouldn't be accessing the `$_GET` array.

pinc/page_controls.inc
```diff
@@ -0,0 +1,152 @@
...
+function expanded_rounds()
+{
+    global $Round_for_round_id_;
+
+    $expanded_rounds = array_keys($Round_for_round_id_);
+    array_unshift($expanded_rounds, 'OCR');
+    return $expanded_rounds;
+}
+
+function get_dp_data()
```
cpeel, line 105: We should be more specific than `dp_data`. `get_proofreading_interface_data_js()`?

scripts/page_browse.js
```diff
@@ -0,0 +1,247 @@
...
+    const imagePercentID = "image_percent";
+    let percent = localStorage.getItem(imagePercentID);
+    if(!percent) {
+        percent = 100;
+    }
+
+    let percentInput = $("<input>", {type: 'number', min: '1', max: '999', value: percent});
+    let resizeButton = $("<input>", {type: 'button', value: dpData.strings.resize});
+
+    let image = $("<img>", {src: dpData.projectsUrl + "/" + pageBrowserData.projectid + "/" + pageBrowserData.currentPage});
```
cpeel, line 134: We should include the `$project->url` in the `pageBrowserData` object and not try to reconstruct it here from the `projectsUrl`.

scripts/page_browse.js
```diff
@@ -0,0 +1,247 @@
+/* global $ pageBrowserData dpData splitControl mode */
```
cpeel, line 1: We need to change `dpData` to something more specific as dpData could mean anything. `proofIntData`?

scripts/page_browse.js
```diff
@@ -0,0 +1,247 @@
...
+
+    let splitPercent = localStorage.getItem(splitPercentID);
+    if(!splitPercent) {
+        splitPercent = 50;
+    }
+
+    let mainSplit = splitter.setup(container, {splitDirection: splitDirection, splitPercent: splitPercent});
+
+    // must set the image width, otherwise it will at first be small, then when image loads later elements can wrap and disappear
+    let vSwitchButton = $("<button>", {type: 'button', class: 'img-button', title: dpData.strings.switchVert}).append($("<img>", {src: dpData.buttonImages.imgVSplit, width: 27}));
```
cpeel, line 83: Ideally we wouldn't include the size of the images here, otherwise if we ever change them we have to remember to come back here and change it, and since it isn't colocated with the name of the image itself we'll likely never find it.

styles/layout.less
```diff
@@ -1158,5 +1158,69 @@ table.search-column {
...
+    height: 100%;
+    box-sizing: border-box;
+    padding-left: .25em;
+    resize: none;
+    border: 0;
+    margin: 0;
+    display: block;
+}
+
+.image-back {
```
cpeel, line 1221: This feels like it belongs in `theme.less` instead.

pinc/page_controls.inc
70ray: This was for a request made in a forum https://www.pgdp.net/phpBB3/viewtopic.php?f=4&t=66584&p=1178166&hilit=page+match#p1178168 to use a partial match but now we have all the pages we can do it with php

cpeel: This is going to return unexpected things when multiple pages match. If a project has page "044" and "144" and someone searches for "44" they will get the first entry of whatever the database returns. And since we do not order the select it is up to the database the order that the rows return -- which may or may not be deterministic.

70ray: I'll write a php function to search in the images we downloaded earlier.

pinc/page_controls.inc
70ray: changed this.

scripts/page_browse.js
70ray: Changed to a relaying out after images are loaded whic seems to work.

tools/proofers/previewControl.js
70ray: Unfortunately this branch changed the get_font_styles() function to remove empty strings which seemed like a good idea at the time since it avoids the function building the selector having to check for empty strings but it meant that preview control.js also had to be changed.

70ray: Unfortunately this branch changed the get_font_styles() function to remove empty strings which seemed like a good idea at the time since it avoids the function building the selector having to check for empty strings but it meant that preview control.js also had to be changed.

pinc/page_controls.inc
```diff
@@ -19,38 +18,32 @@ function get_page_data()
...
-                $result = DPDatabase::query("SELECT image FROM $projectid WHERE image LIKE '%$esc_page%'");
-                if (mysqli_num_rows($result) !== 0)
-                {
-                    $page = mysqli_fetch_assoc($result)["image"];
-                }
-                else
+                foreach($page_names as $page_name)
                 {
-                    $page = "";
+                    if(1 === preg_match("/.*$page.*/", $page_name))
```
cpeel, line 33: While I still don't think this is a good idea, using `stripos($page_name, $page) !== FALSE` is a better way to check for a substring search than doing a regex match.

pinc/page_controls.inc
```diff
@@ -75,24 +68,19 @@ function get_page_data()
...
     $page_browser_data = [
         "projectid" => $projectid,
-        "pages" => $images,
+        "pages" => $page_names,
         "currentPage" => $page,
         "currentRound" => $round_id,
         "text" => $text,
         "heading" => $heading,
         "errorMessage" => $error_message,
+        "imageUrl" => "$projects_url/$projectid/$page"
```
cpeel, line 77: Change `$projects_url/$projectid` to `$project->url` and drop the global for `$projects_url` altogether.

tools/mentors/view_page_text_image.php
```diff
@@ -6,6 +6,10 @@
 
 require_login();
 
+$projectid = array_get($_GET, "project", "");
```
cpeel, line 9: This should use `validate_projectID()` (although I have an MR out where it would be `get_projectID_param()`).

tools/mentors/view_page_text_image.php
```diff
@@ -6,6 +6,10 @@
...
+$page = array_get($_GET, "imagefile", "");
```
cpeel, line 10: This should use `validate_page_image_filename()` (although my MR changes this to `get_page_image_param()`).

tools/project_manager/displayimage.php
```diff
@@ -7,6 +7,16 @@
 
 require_login();
 
+$projectid = array_get($_GET, "project", "");
```
cpeel, line 10: Ditto `validate_projectID()` and `validate_page_image_filename()`.

pinc/prefs_options.inc
```diff
@@ -143,6 +148,25 @@ function get_full_font_families($interface = NULL)
...
+function get_font_sizes()
+{
+    $font_sizes = get_available_proofreading_font_sizes();
+    $font_sizes[0] = BROWSER_DEFAULT_STR;
+    return $font_sizes;
+}
+
+// return an array of strings to use to render text
+// The array is indexed like get_available_proofreading_font_sizes()
+function get_font_size_families()
```
cpeel, line 163: Font sizes don't have families. `get_font_sizes_css()` perhaps?

pinc/page_controls.inc
70ray: I'll change it as you suggest. Is your objection to doing this at all that it adds unnecessary complexity?
Another way to approach it could be to remove the page input when selecting a project, then do not show any page view at first but allow the user to select one.

pinc/page_controls.inc
70ray: good idea.

tools/mentors/view_page_text_image.php
70ray: I'll use your new functions.

pinc/page_controls.inc
70ray: If we validate the imagename we can't get here with a partial name.

cpeel: It's like you're making my case for why accepting partial names is a bad idea :D

70ray:
The red error message doesn't show up well on the form background. Perhaps we should change the form background colour to @header-background.

cpeel:
> The red error message doesn't show up well on the form background. Perhaps we should change the form background colour to @header-background.

That will make it hard to read on the blue theme (red on blue is hard to read). Can you provide a screenshot of the error message you're talking about?

70ray:
> > The red error message doesn't show up well on the form background. Perhaps we should change the form background colour to @header-background.
> 
> That will make it hard to read on the blue theme (red on blue is hard to read). Can you provide a screenshot of the error message you're talking about?

Here is a screen shot 
image
On my screen it loks ok at some angles but not others.

cpeel:
I'm inclined to leave it as-is for now. It is an error state and that styling does match elsewhere on the site.

styles/layout.less
```diff
@@ -1172,5 +1172,62 @@ table.search-column {
...
+        margin-right: 0.4em;
+        margin-left: 0.4em;
+    }
+}
+
+.flex_100vh {
+    height: 100vh;
+    display: flex;
+    flex-direction: column;
+    .fix {
```
cpeel, line 1212: This could use a more descriptive name. Fix? Fix what?

tools/project_manager/display_image.js
```diff
@@ -0,0 +1,96 @@
...
+                    modeInput.val(modeObj.type);
+                });
+                returnValue.push(modeButton);
+            }
+        });
+        return returnValue;
+    };
+
+    let topDiv = $("#top-div");
+    let fixHead = $("<div>", {class: 'fix control-form'});
```
cpeel, line 34: What is 'fixHead'?

scripts/page_browse.js
```diff
@@ -0,0 +1,259 @@
...
+
+// Construct the image sizing controls.
+var imageControl = function() {
+    const imagePercentID = "image_percent";
+    let percent = localStorage.getItem(imagePercentID);
+    if(!percent) {
+        percent = 100;
+    }
+
+    let percentInput = $("<input>", {type: 'number', min: '1', max: '999', value: percent});
```
cpeel, line 143: If I change the image % number and hit Enter it doesn't apply the new image size, but rather submits the form and wants me to enter a new Project ID.

tools/mentors/page_text_image.js
```diff
@@ -0,0 +1,45 @@
...
+        let blankDiv = $("<div>");
+        let textDiv = $("<div>").append(topTextDiv, blankDiv);
+
+        stretchDiv.append(imageDiv, textDiv);
+        let theSplitter = viewSplitter(stretchDiv);
+
+        pageControlForm.append(topLine(), pageChanger(), " ", roundSelect(true), " ", theImageControl.controls,
+            theSplitter.buttons, theTextControl.controls);
+
+        let subSplitter = splitControl();
```
cpeel, line 40: What is the bottom half of the right column for? I wasn't able to get anything to ever show up in there and the original doesn't have it.

pinc/prefs_options.inc
cpeel: Reminder about this unaddressed comment.

70ray: Now changed as suggested.

scripts/page_browse.js
70ray: I've rearranged the controls to avoid this and improved the way it works.

tools/mentors/page_text_image.js
70ray: This is so that mentors can restrict the number of lines of text visible which is currently possible in most browsers as requested in this comment: https://www.pgdp.net/phpBB3/viewtopic.php?f=4&t=66584&hilit=page+match&start=15#p1178414

cpeel: I question the utility of this, but let's see how the other mentors like it.

PR #353, opened 2020-07-16
Updated EEL character suite and picker sets

srjfoo:
Based on research done by DP user bunny-crunch, and with input from
the forums, trimmed several characters from the Extended European
Latin character suite, and reorganized the picker sets.

See [extended-european-latin](https://www.pgdp.org/~srjfoo/c/tools/charsuites.php?charsuite=extended-european-latin) in my sandbox.


srjfoo:
> Can you link to the forum post if possible in case we come upon this PR one day in the future.

Good idea: [Character suites and picker sets: Extended European Latin](https://www.pgdp.net/phpBB3/viewtopic.php?f=4&t=70715). There are bits and pieces of it elsewhere, but this is a large part of it.

PR #354, opened 2020-07-16
Allow a per-project custom character suite

cpeel:
Allow each project to include a set of characters that can be included as adhoc characters for the project. These appear in a special 'Proj' menu in the character picker.

This is available in the [per-project-charsuite](https://www.pgdp.org/~cpeel/c.branch/per-project-charsuite/) sandbox.

### DIFF
--- a/tools/project_manager/editproject.php
+++ b/tools/project_manager/editproject.php
@@ -184,6 +184,7 @@ class ProjectInfoHolder
         $this->text_preparer    = $pguser;
         $this->extra_credits    = '';
         $this->deletion_reason  = '';
+        $this->custom_chars     = '';
         // $this->year          = '';
         $this->state            = '';    
     }
@@ -228,6 +229,7 @@ class ProjectInfoHolder
         $this->text_preparer    = $pguser;
         $this->extra_credits    = '';
         $this->deletion_reason  = '';
+        $this->custom_chars     = '';
         $this->state            = '';
 
         $this->original_marc_array_encd = $encoded_marc_array;
@@ -276,6 +278,7 @@ class ProjectInfoHolder
         $this->text_preparer    = $up_info['d_text_preparer'];
         $this->extra_credits    = $up_info['d_extra_credits'];
         $this->deletion_reason  = '';
+        $this->custom_chars     = $up_info['d_custom_chars'];
         $this->state            = '';
 
         // $this->year          = $up_info['d_year'];
@@ -364,13 +367,14 @@ class ProjectInfoHolder
             $this->clone_projectid = $project->projectid;
             $this->state            = '';
         }
-        $this->up_projectid     = $project->up_projectid;
-        $project_charsuites = $project->get_charsuites();
+        $this->up_projectid = $project->up_projectid;
+        $project_charsuites = $project->get_charsuites(FALSE);
         $this->charsuites = [];
         foreach($project_charsuites as $project_charsuite)
         {
             array_push($this->charsuites, $project_charsuite->name);
         }
+        $this->custom_chars  = $project->custom_chars;
     }
 
     // -------------------------------------------------------------------------
@@ -599,6 +603,34 @@ class ProjectInfoHolder
             }
         }
 
+        $this->custom_chars = utf8_normalize(@$_POST['custom_chars']);
+        if($this->custom_chars)
+        {
+            $codepoints = utf8_codepoints_combining($this->custom_chars);
+
+            // all characters must be unique
+            if($codepoints != array_unique($codepoints))
+            {
+                $errors .= _("The set of custom characters must be unique.")."<br>";
+            }
+
+            // only allow 32 characters
+            if(count($codepoints) > 32)
+            {
+                $errors .= _("A maximum of 32 custom characters are allowed.")."<br>";
+            }
+
+            // prevent disallowed characters from being added
+            $disallowed_codepoints = array_intersect(get_disallowed_codepoints(), $codepoints);
+            if($disallowed_codepoints != [])
+            {
+                $errors .= sprintf(
+                    _("The following custom characters are not allowed: %s"),
+                    implode(", ", array_map('utf8_chr', $disallowed_codepoints))
+                )."<br>";
+            }
+        }
+
         $this->scannercredit    = @$_POST['scannercredit'];
         $this->comments         = @$_POST['comments'];
         $this->clearance        = @$_POST['clearance'];
@@ -646,7 +678,8 @@ class ProjectInfoHolder
             image_preparer = '".mysqli_real_escape_string(DPDatabase::get_connection(), $this->image_preparer)."',
             text_preparer  = '".mysqli_real_escape_string(DPDatabase::get_connection(), $this->text_preparer)."',
             extra_credits  = '".mysqli_real_escape_string(DPDatabase::get_connection(), utf8_normalize($this->extra_credits))."',
-            deletion_reason= '".mysqli_real_escape_string(DPDatabase::get_connection(), $this->deletion_reason)."'
+            deletion_reason= '".mysqli_real_escape_string(DPDatabase::get_connection(), $this->deletion_reason)."',
+            custom_chars   = '".mysqli_real_escape_string(DPDatabase::get_connection(), utf8_normalize($this->custom_chars))."'
         ";
         $pm_setter = '';
         if ( user_is_a_sitemanager() )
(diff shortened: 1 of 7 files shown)


srjfoo:
I'm not entirely sure what's going on here, or if we need to worry about it at this point. This is a screenshot of some testing I did in the [Tiger Lily](https://www.pgdp.org/~cpeel/c.branch/per-project-charsuite/project.php?id=projectID5df912d3cae6b) project in P1:
image

I went through my OS's character viewer and picked some random Astral Plane charcters in addition to including a few of the zodiac symbols with VS-15. Also some plain ASCII characters.

1. WC flagging the characters as needing review is inconsistent.
2. WC flagging characters seems to depend on where they are in relation to the big filled-in circle.
3. I think the two characters on the far left for lines 2-4 are RTL -- when I was adding that character to the per-project character suite, the cursor kept re-positioning to the left when I tried to add another character to the right of it. Note that in the proofing interface, they just go where I put them. (Yes, I know we're not dealing with RTL right now -- I'm mentioning it only because of the positioning weirdness when I was entering that character in the project edit interface.)

tools/project_manager/editproject.php
```diff
@@ -599,6 +603,40 @@ function set_from_post()
...
+            if($codepoints != array_unique($codepoints))
+            {
+                $errors .= _("The set of custom characters must be unique.")."<br>";
+            }
+
+            // all characters must be normalized
+            if(count(get_nonnormalized_codepoints($codepoints)))
+            {
+                $errors .= _("Custom characters must be in NFC normalized form.")."<br>";
+            }
```
srjfoo, line 621: Requiring characters to be normalized is legitimate, but for those entering combining diacritics, they may not even realize that what they're entering is not normalized. I deliberately added an `ā` both in normalized and non-normalized forms to see if that would trigger the message, and it did. Would it be realistic to normalize anything being added to a project custom character suite?

If all they're doing is copying a few characters from an existing character suite, or possibly from another web page, it should come to them normalized, but I'm thinking back to the few characters we found in project titles and authors' names that were non-normalized combining characters. It may be rare, but it will happen, and we'll likely wind up with confused PMs.

tools/project_manager/editproject.php
```diff
@@ -599,6 +603,40 @@ function set_from_post()
...
+
+            // only allow 32 characters
+            if(count($codepoints) > 32)
+            {
+                $errors .= _("A maximum of 32 custom characters are allowed.")."<br>";
+            }
+
+            // prevent disallowed characters from being added
+            $disallowed_codepoints = array_intersect(get_disallowed_codepoints(), $codepoints);
+            if($disallowed_codepoints != [])
```
srjfoo, line 631: Maybe add the space to the disallowed codepoints? This is what the picker tile looks like for a space: 
image

tools/project_manager/editproject.php
cpeel: We can easily transparently normalize them and indeed that might be simpler for the users. My only worry is that some PM is going to put in a character that doesn't NFC normalize down to what they expect.

tools/project_manager/editproject.php
cpeel: We could special-case a space, but we can't add it to `get_disallowed_codepoints()` because it's allowed.

srjfoo: Oops. Right. Silly me. I was focused on "disallowed for picker sets" not the fact that it would be disallowing the space completely. Yeah, bad idea.

cpeel:
> I did some more testing on WC and the special characters. As things are right now, they don't appear to be flagged as bad characters even if added as individual characters in both WC and the proofreading interface -- so if them not being flagged is going to be a problem we need to figure out how to deal with it (i.e. -- try to come up with a fix, or document?)/

Project custom characters are effectively treated the same way as those from character suites as far as WC is concerned. So while you may have found a bug, it is unlikely to be in the custom character code. Please do the same testing with those character in a character suite. I also don't understand how you expect them to be flagged -- did you add words with them to a project's bad list? Are you expecting them to be flagged because they are different scripts?

tools/project_manager/editproject.php
srjfoo: Either they have to go out and find the normalized form, or they put in what they think is the right character (or character combination) and we normalize it for them. If it doesn't look right, they still have to go out and find the normalized form. As it stands now, it's slightly more work for them, because if they get the normalization error, they have to go out and find a normalized form when they might not have to if we do it for them.

I noticed one character yesterday where the character was vastly different, depending on the font, so it's also possible that any problems in shape might be due to font variations.

srjfoo:
> > I did some more testing on WC and the special characters. As things are right now, they don't appear to be flagged as bad characters even if added as individual characters in both WC and the proofreading interface -- so if them not being flagged is going to be a problem we need to figure out how to deal with it (i.e. -- try to come up with a fix, or document?)/
> 
> Project custom characters are effectively treated the same way as those from character suites as far as WC is concerned. So while you may have found a bug, it is unlikely to be in the custom character code. Please do the same testing with those character in a character suite. I also don't understand how you expect them to be flagged -- did you add words with them to a project's bad list? Are you expecting them to be flagged because they are different scripts?

I'll do some more research.

70ray:
Would it be better to specify the characters as their unicode value rather than pasting in the actual character? For (probably unrealistic) example the three characters 00D0 (capital eth), 0110 (D with stroke), and 0189 (African D) look pretty much identical so the user might not get what he really intends.

cpeel:
> Would it be better to specify the characters as their unicode value rather than pasting in the actual character? 

I thought about that, but then we have to teach them how to specify compound characters and also expect them to get the syntax right. Direct insert of the character seemed like the best UX.

srjfoo:
> > Would it be better to specify the characters as their unicode value rather than pasting in the actual character?
> 
> I thought about that, but then we have to teach them how to specify compound characters and also expect them to get the syntax right. Direct insert of the character seemed like the best UX.

I don't know about Linux, but both Macs and Windows machines have the ability to show a table of Unicode characters, I think? I know Macs do, but I'm not sure about the precise implementation on Windows). Taking another possible situation, hopefully someone actually going to the trouble of looking up characters in a table would be unlikely to confuse the ASCII letter `A` with its Greek or Cyrillic counterpart.

tools/project_manager/editproject.php
cpeel: Characters are now normalized on save.

cpeel:
I added "N̅" to a project's custom character list (project was in English). I then put "N̅graze" into the proofreading interface and it did not get flagged -- which is expected since "N̅" is not a character in the English alphabet so aspell ignores the word (`<roll eyes>` thanks aspell). I added that word to the project's Bad Words list and WC correctly flagged it.

It's certainly possible we have some other WC-related issues in the code, but I don't think it's related to the per-project character suites

tools/project_manager/editproject.php
srjfoo: Looks like the normalization is done after the check for duplicates:
image

Reference custom charsuite in [UTF-8 practice: Tiger Lily and other stories.](https://www.pgdp.org/~cpeel/c.branch/per-project-charsuite/tools/charsuites.php?projectid=projectID5df912d3cae6b#projectID5df912d3cae6b).

cpeel: Yes, it was on save to the DB. I've moved it up to before the validation. This is probably going to make it confusing for the user if they add a normalized letter and a non-normalized letter and then get a report that there is a duplicate. `<shrugs>`

srjfoo: If you think it best, we can go back to just issuing the warning that characters in the custom suite have to be normalized, and not force the normalization. I'm trying to decide what I'd prefer as a PM, and I think I'd prefer the script to take care of it, especially if the only way I know to input it is non-normalized, but others might feel differently, and even if I get a duplicate character message.

Did the original version identify the non-normalized character(s)? (Short memory, here -- I don't remember!) If not, that's a potential problem if the PM doesn't have a clue what's normalized and what's not.

srjfoo:
> I added "N̅" to a project's custom character list (project was in English). I then put "N̅graze" into the proofreading interface and it did not get flagged -- which is expected since "N̅" is not a character in the English alphabet so aspell ignores the word (`<roll eyes>` thanks aspell). I added that word to the project's Bad Words list and WC correctly flagged it.
> 
> It's certainly possible we have some other WC-related issues in the code, but I don't think it's related to the per-project character suites

And I just now tried `Αño` (cap Greek alpha), and WC correctly identified as a possible bad word when I enabled the Spanish dictionary. On the other hand āño slid through without a whimper. I think that PM education is going to be very important. My earliest tests were aimed at seeing if WC would identify possible bad characters just on the basis of them not being part of a "real word".

srjfoo:
> > ... My earliest tests were aimed at seeing if WC would identify possible bad characters just on the basis of them not being part of a "real word".

Whoops ... posted too quickly. Those early tests can be discounted, I think, because I was trying a bunch of fairly bizarre stuff that should not happen in practice.

PR #356, opened 2020-07-20
Separate projectid/page param validation

cpeel:
Create two new projectID and page name param fetching functions that
more closely mirror the syntax for the other param fetching functions
in misc.inc. This lets us make the validate_*() functions purely
for validation and more generally useful elsewhere.

### DIFF
--- a/pinc/Project.inc
+++ b/pinc/Project.inc
@@ -15,6 +15,7 @@ class UTF8ConversionException extends Exception { }
 class ProjectException extends Exception { }
 class NonexistentProjectException extends ProjectException { }
 class InvalidProjectIDException extends ProjectException { }
+class InvalidPageException extends ProjectException { }
 class NoProjectPageTable extends ProjectException { }
 class NoProjectDirectory extends ProjectException { }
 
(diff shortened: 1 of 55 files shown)


pinc/Project.inc
```diff
@@ -1068,37 +1069,80 @@ function does_project_page_table_exist($projectid)
...
+    }
+    return $projectid;
 }
 
-// Checks if $value contains only legitimate characters for
-// the filename of a page image file in a project.
-// If it does, it returns it, otherwise dies with a warning.
-function validate_page_image_filename($param_name, $value, $allownull = false)
+// Confirm that $page_name is a well-formed page filename or raise an exception.
+function validate_page_image($page_name)
```
chrismiceli, line 1111: do we have unit tests for this? It seems simple enough to write one.

cpeel: No we don't but we should - I will fix that.

cpeel: Tests added.

chrismiceli: Looks great thanks!

cpeel:
Note to self: update the noncvs code when this gets rolled out to TEST and then PROD.

PR #360, opened 2020-07-21
Improve SQL in some pinc/ files

cpeel:
Improve the SQL in some files in `pinc/`, including removing unused code from `pinc/SortUtility.inc`.

### DIFF
--- a/pinc/authors.inc
+++ b/pinc/authors.inc
@@ -1,6 +1,11 @@
 <?php
   function get_biography($id) {
-    $result = mysqli_query(DPDatabase::get_connection(), "SELECT bio FROM biographies WHERE bio_id = $id;");
+    $sql = sprintf("
+        SELECT bio
+        FROM biographies
+        WHERE bio_id = %d
+    ", $id);
+    $result = DPDatabase::query($sql);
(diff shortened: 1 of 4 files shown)


pinc/authors.inc
```diff
@@ -1,6 +1,11 @@
 <?php
   function get_biography($id) {
-    $result = mysqli_query(DPDatabase::get_connection(), "SELECT bio FROM biographies WHERE bio_id = $id;");
+    $sql = sprintf("
+        SELECT bio
+        FROM biographies
+        WHERE bio_id = %d
+    ", $id);
+    $result = DPDatabase::query($sql);
```
chrismiceli, line 8: are we intentionally now throwing with new default behavior?

cpeel: Yes, particularly for SELECTs. The only reason we should ever get a DB error from a SELECT is if we have some kind of poorly formed query in which case we want to know about it and stop. INSERTs are going to be trickier as I've seen at least one case where the query fails the insert due to a duplicate primary key. That represents a logic error we should be handling upstream of the query, but I'm unsure if we should die in that case.

PR #361, opened 2020-07-24
Convert authors pages

chrismiceli:


tools/authors/add.php
```diff
@@ -54,34 +54,46 @@
         // insert into the database
         if ($author_id) {
             // edit existing author
-            $result = mysqli_query(DPDatabase::get_connection(), sprintf("
+            $result = DPDatabase::query(sprintf("
```
cpeel, line 57: Per the [best practices guide](https://www.pgdp.net/wiki/DP_Code_Best_Practices#Building_SQL), please make this:
```
$sql = sprintf("
    ...
);
```
Ditto elsewhere.

tools/authors/add.php
```diff
@@ -54,34 +54,46 @@
...
-                mysqli_real_escape_string(DPDatabase::get_connection(), $dcomments)
+            ",  DPDatabase::escape($last_name),
+                DPDatabase::escape($other_names),
+                DPDatabase::escape($byear),
+                DPDatabase::escape($bmonth),
+                DPDatabase::escape($bday),
+                DPDatabase::escape($dyear),
+                DPDatabase::escape($dmonth),
+                DPDatabase::escape($dday),
+                DPDatabase::escape($bcomments),
```
cpeel, line 71: This needs to be after `DPDatabase::escape($bday),`

tools/authors/add.php
```diff
@@ -121,11 +133,9 @@
...
         // edit specified author
         // get the values from the database
-        $result = mysqli_query(DPDatabase::get_connection(), "SELECT * FROM authors WHERE author_id = $author_id;");
-        if (!$result) {
-            echo "That author doesn't exist!";
-            exit();
-        }
+        $result = DPDatabase::query(sprintf(
+            "SELECT * FROM authors WHERE author_id = %d;",
+            DPDatabase::escape($author_id)));
```
cpeel, line 138: `$author_id` is an int, you don't need to escape it with the `%d`.

tools/authors/addbio.php
```diff
@@ -19,7 +19,9 @@
 elseif (isset($_GET['bio_id'])) {
     // init edit of existing bio
     $bio_id = get_integer_param($_GET, 'bio_id', null, null, null, TRUE);
-    $result = mysqli_query(DPDatabase::get_connection(), "SELECT * FROM biographies WHERE bio_id = $bio_id;");
+    $result = DPDatabase::query(sprintf(
+        "SELECT * FROM biographies WHERE bio_id = %d;",
+        DPDatabase::escape($bio_id)));
```
cpeel, line 24: Don't escape, it's an int.

tools/authors/addbio.php
```diff
@@ -36,20 +38,20 @@
...
             // edit existing bio
             $bio_id = get_integer_param($_POST, 'bio_id', null, null, null, TRUE);
-            $result = mysqli_query(DPDatabase::get_connection(), sprintf("
+            $result = DPDatabase::query(sprintf("
                 UPDATE biographies
                 SET bio = '%s'
-                WHERE bio_id = %s
-            ", mysqli_real_escape_string(DPDatabase::get_connection(), $bio), $bio_id));
+                WHERE bio_id = %d
+            ", DPDatabase::escape($bio), DPDatabase::escape($bio_id)));
```
cpeel, line 45: `$bio_id` is an int.

cpeel: You're escaping ints in various places, please review all of your changes and address these.

tools/authors/add.php
chrismiceli: I was worried that they came from the query, which is user controlled, but I see we run them through get_integer_param which validates the interger-ness of it. I'll all validated integers and remove escaping.

cpeel: That actually doesn't matter though -- sprintf("%d") handles that for us and forces anything to an integer which doesn't need escaping which is why we use it for integers.
```
sprintf("%d", "blah") == 0
```

tools/authors/add.php
```diff
@@ -54,35 +54,49 @@
...
-                    byear=$byear, bmonth=$bmonth, bday=$bday, bcomments='%s',
-                    dyear=$dyear, dmonth=$dmonth, dday=$dday, dcomments='$%s'
-                WHERE author_id = $author_id
-            ", mysqli_real_escape_string(DPDatabase::get_connection(), $last_name),
-                mysqli_real_escape_string(DPDatabase::get_connection(), $other_names),
-                mysqli_real_escape_string(DPDatabase::get_connection(), $bcomments),
-                mysqli_real_escape_string(DPDatabase::get_connection(), $dcomments)
-            ));
+                    byear=%d, bmonth=%d, bday=%d, bcomments='%s',
+                    dyear=%d, dmonth=%d, dday=%d, dcomments='$%s'
```
cpeel, line 61: s/$%s/%s/

tools/authors/authorxml.php
```diff
@@ -32,7 +32,8 @@
 header("Content-Type: text/xml; charset=$charset");
 echo "<?xml version=\"1.0\" encoding=\"$charset\" ?>\n";
 
-$result = mysqli_query(DPDatabase::get_connection(), "SELECT * FROM authors $clause");
+$sql = sprintf("SELECT * FROM authors %s", $clause);
```
cpeel, line 35: You don't need the sprintf() here since we aren't enforcing formatting. The original with the embedded $clause is fine.

tools/authors/bio.php
```diff
@@ -24,7 +24,8 @@
 }
 
 // try to get bio
-$result = mysqli_query(DPDatabase::get_connection(), "SELECT author_id, bio FROM biographies WHERE bio_id=$id;");
+$sql = sprintf("SELECT author_id, bio FROM biographies WHERE bio_id=%d;", $id);
+$result = DPDatabase::query($sql, false);
```
cpeel, line 28: We don't need `false` here. It's not an error if select returns zero rows.

tools/authors/bioxml.php
```diff
@@ -39,7 +39,8 @@
 header("Content-Type: text/xml; charset=$charset");
 echo "<?xml version=\"1.0\" encoding=\"$charset\" ?>\n";
 
-$result = mysqli_query(DPDatabase::get_connection(), "SELECT * FROM biographies $clause");
+$sql = sprintf("SELECT * FROM biographies %s", $clause);
```
cpeel, line 42: No need for sprintf() here.

tools/authors/harvest.php
```diff
@@ -33,12 +33,14 @@
     exit;
 }
 
-$result = mysqli_query(DPDatabase::get_connection(), 'SELECT * FROM authors');
+$sql = 'SELECT * FROM authors';
+$result = DPDatabase::query($sql, false);
```
cpeel, line 37: No need for `false`.

tools/authors/harvest.php
```diff
@@ -33,12 +33,14 @@
...
 if (mysqli_num_rows($result) > 0) {
     echo _("The table 'authors' is not empty. Please empty it and try again.");
     exit;
 }
-$result = mysqli_query(DPDatabase::get_connection(), 'SELECT * FROM biographies');
+$sql = 'SELECT * FROM biographies';
+$result = DPDatabase::query($sql, false);
```
cpeel, line 43: No need for `false`.

tools/authors/harvest.php
```diff
@@ -196,37 +198,30 @@
                         byear, bmonth, bday,
                         dyear, dmonth, dday,
                         bcomments, dcomments, enabled)
-                VALUES ('%s', '%s', $date_fields_str '', '', 'no')
-            ", mysqli_real_escape_string(DPDatabase::get_connection(), $last_name),
-                mysqli_real_escape_string(DPDatabase::get_connection(), $other_names));
+                VALUES ('%s', '%s', %s '', '', 'no')
```
cpeel, line 201: Leave `$date_fields_str` inside the string since its contents are already validated prior.

tools/authors/harvest.php
```diff
@@ -196,37 +198,30 @@
...
                 $author_id = mysqli_insert_id(DPDatabase::get_connection());
                 echo '    ' . sprintf( _("The author was inserted into the database with the id %d."), $author_id) . "\n";
             }
             $query = sprintf("
                 INSERT INTO biographies
                     (author_id, bio)
-                VALUES(%s, '%s')
-            ", $author_id, mysqli_real_escape_string(DPDatabase::get_connection(), $bio));
+                VALUES(%d, '%s')
+            ", DPDatabase::escape($author_id), DPDatabase::escape($bio));
```
cpeel, line 219: $author_id doesn't need the escape()

tools/authors/harvest.php
```diff
@@ -242,7 +237,8 @@ function ensure_digits($digits_or_question_mark) {
 
 // Returns true if the table exists in the current database, false otherwise.
 function table_exists($tableName) {
```
cpeel, line 239: You can remove this function and the block at the top of the file that validates these exist. They will always exist since they are part of the base install. This file must have predated how we do table creations.

tools/authors/manage.php
```diff
@@ -468,7 +478,11 @@ function evaluateForm(form) {
 
 while ($i++ < $count && $author = @mysqli_fetch_array($result)) {
     $id = $author['author_id'];
-    $bioresult = mysqli_query(DPDatabase::get_connection(), "SELECT bio_id FROM biographies WHERE author_id = $id ORDER BY bio_id;");
+    $bioresult = DPDatabase::query(
```
cpeel, line 481: Break this out into an `$sql=`. $id doesn't need to be escape()d. query() doesn't need a 'false'.

tools/authors/search.inc
```diff
@@ -173,7 +173,10 @@ function get_search_title() {
 function search() {
     global $sortUtility, $sql_what_to_list, $search_result;
     $orderby = $sortUtility->getOrderBy();
-    $search_result = mysqli_query(DPDatabase::get_connection(), "SELECT * FROM authors WHERE $sql_what_to_list ORDER BY $orderby");
```
cpeel, line 176: Leave the two variables in the string directly, but do break it out into an `$sql=`. Also, remove the `false` from query().

tools/authors/search.inc
```diff
@@ -173,7 +173,8 @@ function get_search_title() {
...
+    $sql = sprintf("SELECT * FROM authors WHERE $sql_what_to_list ORDER BY $orderby");
```
cpeel, line 176: You don't need the `sprintf()` -- this is currently throwing an error.

chrismiceli:
> Please be sure and test your changes (feel free to use the TEST server). I found this bug first thing after checking out your branch and doing a search.

Yes, sorry I was having trouble with my TEST account, but could've used the VM. I hope I don't end up being more burden than I'm worth.

cpeel:
> Yes, sorry I was having trouble with my TEST account, but could've used the VM. I hope I don't end up being more burden than I'm worth.

Not at all - I appreciate your work!!

PR #365, opened 2020-08-06
Unicode strict identity escaping

chrismiceli:
It is due to unicode changes. As part of unicode changes, we updated search replace to support unicode characters. We accomplish this search by using regular expressions (even for non regular expression searches). To prevent searches in non regular expression searches from accidentally stumbling into a valid or invalid regular expression, we used an encoding algorithm defined here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#Escaping. You can see (not obviously) that it encodes '-' to '\\-'. This encoding is not necessary since we will never be in a character class restriction. Thus, it is called an identity escape. Here is where it gets fun. For unicode regular expressions, the javascript specification makes this escape an error (more strict). I can't find much documentation on this besides the spec itself: https://github.com/v8/v8/blob/master/src/regexp/regexp-parser.cc#L1673 and from the reg exp engine source code for chrome and firefox:
https://github.com/v8/v8/blob/master/src/regexp/regexp-parser.cc#L1673
https://github.com/mozilla/gecko-dev/blob/baf1cd492406a9ac31d9ccb7a51c924c7fbb151f/js/src/irregexp/imported/regexp-parser.cc#L390

### DIFF
--- a/tools/proofers/srchrep.js
+++ b/tools/proofers/srchrep.js
@@ -24,7 +24,9 @@ var srchrep = (function() {
         }
         opener.parent.docRef.editform.text_data.value = opener.parent.docRef.editform.text_data.value.replace(
             new RegExp(search,'gu'),
-            replacetext);
+            function () {
+                return replacetext;
+            });
         setUndoButtonDisabled(false);
     }
 
--- a/tools/proofers/srchrep.js
+++ b/tools/proofers/srchrep.js
@@ -8,7 +8,7 @@ var srchrep = (function() {
     }
 
     function escapeRegExp(string) {
-        return string.replace(/[.*+\-?^${}()|[\]\\]/g, '\\$&'); // $& means the whole matched string
+        return string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'); // $& means the whole matched string
     }
 
     function setUndoButtonDisabled(state) {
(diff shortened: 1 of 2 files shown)


tools/proofers/srchrep.js
```diff
@@ -8,7 +8,7 @@ var srchrep = (function() {
     }
 
     function escapeRegExp(string) {
-        return string.replace(/[.*+\-?^${}()|[\]\\]/g, '\\$&'); // $& means the whole matched string
+        return string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'); // $& means the whole matched string
```
srjfoo, line 11: Testing shows that there's one other character in the replace string above that misbehaves when used as a s/r. More specifically, as a replacement string. The `$` doesn't behave as expected.

I've probably gone about my testing backwards, but first tests show that:
s: `$`, r: `$$` shows no visible effect.
However, s: `$`, r: `$$$$`, shows the replacement as `$$`. 

So, what's happening with the `$` is that it's only doing one of every two. I.e. in the first example, it probably was actually replacing the intital `$`, but since it was only taking one of the ones in the replacement string, it didn't appear that anything was happening.

srjfoo: postscript: when I tried an odd number (5), the replacement string was 3 characters long.

chrismiceli: Interesting. I learn something new every day. Apparently the replace value has special "[replacement patterns](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#Specifying_a_string_as_a_parameter)" all revolving around the '$' character. I've sent a fix, but this has been around forever it seems. I've sent a fix that avoids these replacement patterns by using a replacement function.

cpeel:
Great investigative work and fixes for both of these issues!!

PR #367, opened 2020-08-07
Pp upload

70ray:
Not ready yet. Just checking it passes tests.

### DIFF
--- a/tools/upload_text.php
+++ b/tools/upload_text.php
@@ -3,7 +3,7 @@ $relPath="../pinc/";
 include_once($relPath.'base.inc');
 include_once($relPath.'metarefresh.inc');
 include_once($relPath.'project_states.inc');
-include_once($relPath.'project_trans.inc');
+include_once($relPath.'project_trans.inc'); // project_transition()
 include_once($relPath.'theme.inc');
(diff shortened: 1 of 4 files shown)


tools/upload_text.php
```diff
@@ -270,6 +295,15 @@
             notify_project_event_subscribers( $project, 'sr_reported' );
         }
 
+        if (($stage == "in_prog_1") || ($stage == "in_prog_2"))
+        {
+            // record postcomments in projects table
```
cpeel, line 300: We're trying to standardize on the following format:
```php
$sql = sprintf("
    UPDATE projects
    SET postcomments = CONCAT(postcomments, '%s')
    WHERE projectid = '%s'
", DPDatabase::escape($postcomments),
    DPDatabase::escape($projectid));
DPDatabase::query($sql);
```

See the updated [coding best practices](https://www.pgdp.net/wiki/DP_Code_Best_Practices#Escaping_in_SQL).

srjfoo:
Tests look good!

tools/upload_text.php
70ray: Done this.

PR #370, opened 2020-08-11
Updates for addproofer.php and privacy.php

srjfoo:
Per GM, clarification updates for registration hints paragraph in
accounts/addproofer.php, and updates to faq/privacy.php.

Available for review at [https://www.pgdp.org/~srjfoo/c.branch/addproofer-privacy/accounts/addproofer.php](https://www.pgdp.org/~srjfoo/c.branch/addproofer-privacy/accounts/addproofer.php)

### DIFF
--- a/faq/privacy.php
+++ b/faq/privacy.php
@@ -41,7 +41,8 @@ output_section(
     array(
         _('When you register, %s will send a confirmation/introduction message to your e-mail address.'),
         _('You may also receive optional e-mail alerts from our discussion forums and occasional notifications and feedback related to your proofreading activities.'),
-        _('%1$s may also invite you by e-mail to vote in periodic Board Trustee elections and could contact you for site-related purposes authorized by %1$s management.'),
+        _('%1$s may also invite you by e-mail to vote in periodic Board of Trustees elections.'), 
+        _('You may also be contacted for site-related purposes by %1$s administrators.'),
     )
 );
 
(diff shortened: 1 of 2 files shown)


faq/privacy.php
```diff
@@ -51,51 +52,57 @@
...
+        _('Your User Name and any information you post on the wiki are visible to both volunteers and unregistered guests using the wiki.'),
     )
 );
 
 output_section(
     _('Tracking information'),
     array(
         _('%s tracks your User Name, the date your account was created and your last login date.'),
-        _('This tracking information will be available to other volunteers.'),
+        _('If you set your statistics to "anonymous," the date your account was created and your last login date will be visible only to administrators.'),
```
cpeel, line 83: > `"anonymous,"`

I know this is how it's "suppose to be" in American English, but I cringe with the comma being inside the quote, because `anonymous,` is not the setting -- it's `anonymous`. If we aren't willing to put the comma outside, can we sidestep it and use something besides quotes to emphasize the setting name?

srjfoo: For what it's worth, I agree with you. That's the way I got it, and since I admitted that it was a legitimate construction (even though it looks wrong to me), I left it.

srjfoo: Fixed.

lhamilton1:
I approve

PR #371, opened 2020-08-11
Revised charsuite-polytonic-greek.inc

srjfoo:
Removed ASCII semicolon and mid-dot from character suite. Revised
picker sets: removed capital vowels with grave or acute accents
that don't have breathing marks; removed capital Alpha, Eta and Omega
with iota subscript/adscript alone or in combination with any other
marks. Changed picker labels for the pickers with diacriticals or other
marks to use the initial accented lower case picker letter. The characters
that were removed from the pickers remain in the full character suite,
just not available via a picker.

Can be viewed in the [polytonic-greek](https://www.pgdp.org/~srjfoo/c.branch/polytonic-greek/tools/charsuites.php?charsuite=polytonic-greek&font=DP+Sans+Mono) branch sandbox.

### DIFF
--- a/pinc/charsuite-polytonic-greek.inc
+++ b/pinc/charsuite-polytonic-greek.inc
@@ -99,8 +99,7 @@ $pickerset->add_subset(utf8_chr('U+0391') . '-' . utf8_chr('U+03a9'), [
 
 $pickerset->add_subset(utf8_chr('U+03ac'), [
     [ NULL, NULL, NULL, 'U+1f08', 'U+1f09', 'U+1f0c', 'U+1f0d', 'U+1f0a', 
-      'U+1f0b', 'U+1f0e', 'U+1f0f', NULL, NULL, NULL, NULL, NULL, NULL,
-      NULL, NULL, NULL, NULL, NULL, NULL ],
+      'U+1f0b', 'U+1f0e', 'U+1f0f' ],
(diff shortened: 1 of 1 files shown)


pinc/charsuite-polytonic-greek.inc
```diff
@@ -93,58 +91,68 @@ $charsuite->reference_urls = [
...
     [ 'U+03b1-U+03c9' ], # lowercase alpha through omega
 ]);
-$pickerset->add_subset(utf8_chr('U+0386'), [
-    [ 'U+0386', 'U+1fba', NULL, 'U+1f08', 'U+1f09', 'U+1f0c', 'U+1f0d', 'U+1f0a', 
-      'U+1f0b', 'U+1f0e', 'U+1f0f', 'U+1fbc', NULL, NULL, NULL, 'U+1f88', 'U+1f89',
-      'U+1f8c', 'U+1f8d', 'U+1f8a', 'U+1f8b', 'U+1f8e', 'U+1f8f'  ],
+
+$pickerset->add_subset(utf8_chr('U+03ac'), [
+    [ NULL, NULL, NULL, 'U+1f08', 'U+1f09', 'U+1f0c', 'U+1f0d', 'U+1f0a', 
+      'U+1f0b', 'U+1f0e', 'U+1f0f', NULL, NULL, NULL, NULL, NULL, NULL,
```
cpeel, line 102: I would prefer we not try to fill in the right side of the picker sets with `NULL`s. If we want the picker set rows "filled out" on the `charsuite.php` page we should do it there.

PR #372, opened 2020-08-12
Allow PFs to see Anonymous user stats

srjfoo:
Allow Project Facilitators to see Anonymous user stats. Updated the message that is displayed to reflect the change.


cpeel:
Nicely done!

PR #373, opened 2020-08-13
Show just the image and size control where appropriate.

70ray:
There are some other cases where this could be used which I can incorporate if you think this is a good way to go.


cpeel:
After thinking about it, let's move both of these files to the top-level `tools/` directory named thusly:
* tools/display_image.php
* tools/display_image.js

In a future commit we can rename the current `tools/project_manager/displayimage.php` to `tools/explore_images.php`.

tools/project_manager/displayimageonly.php
```diff
@@ -0,0 +1,40 @@
...
+{
+    $error_message = $exception->getMessage();
+}
+
+$title = _("Display Image for page");
+
+$js_files = [
+    "$code_url/scripts/page_browse.js",
+    "$code_url/tools/project_manager/display_image_only.js",
+    ];
```
cpeel, line 28: Outdent this so it matches the `$header_args` def.

70ray: done this

PR #374, opened 2020-08-14
Add project page access to APIs

cpeel:
This adds (very) basic project page access to the APIs and allows the APIs to be accessed via session making them useful to JS UIs. Specifically it exposes the APIs needed to allow @70ray's new view images code to populate the data via AJAX instead of page loads.

I've also included a new global JS function to assist with creating the `$.ajax()` calls.

API examples:

## All the available page rounds (including OCR)
```
curl -i -g -H "Accept: application/json" -H "X-API-KEY: review" \
    "https://www.pgdp.org/~cpeel/c.branch/api-project-pages/api/index.php?url=v1/projects/pagerounds"
```
Results:
```json
[
    "OCR",
    "P1",
    "P2",
    "P3",
    "F1",
    "F2"
]
```

## All the pages for a project
```
curl -i -g -H "Accept: application/json" -H "X-API-KEY: reviw" \
    "https://www.pgdp.org/~cpeel/c.branch/api-project-pages/api/index.php?url=v1/projects/projectID45c225f598e32/pages"
```
Results:
```json
[
    "001.png",
    "002.png",
    "003.png",
    "004.png",
...
    "417.png",
    "418.png"
]
```

## Page details
```
curl -i -g -H "Accept: application/json" -H "X-API-KEY: review" \
    "https://www.pgdp.org/~cpeel/c.branch/api-project-pages/api/index.php?url=v1/projects/projectID45c225f598e32/pages/006.png/pagerounds/OCR"
```
Results:
```json
{
    "pagename": "006.png",
    "image_url": "https://www.pgdp.org/projects/projectID45c225f598e32/006.png",
    "text": "\"The social revolution which is impending in Europe is\r\nchiefly concerned with the future of the workers and the\r\nwomen. It is for this that I hope and wait, and for this I\r\nwill work with all my powers.\"--IBSEN.",
    "state": "P1.page_saved"
}
```

### jQuery AJAX example:
```js
settings = makeApiAjaxSettings("v1/projects/projectID45c225f598e32/pages");
$(document).ready(function() {
    $.ajax(
        settings
    ).then(function(data) {
    	console.log(data)
    });
});
```

### DIFF
--- a/api/ApiRouter.inc
+++ b/api/ApiRouter.inc
@@ -52,7 +52,7 @@ class ApiRouter
             {
                 list($param_name, $validator) = $this->get_validator($url_map);
                 $url_map = &$url_map[$param_name];
-                $data[$param_name] = $validator($part);
+                $data[$param_name] = $validator($part, $data);
             }
         }
         if(!isset($url_map["endpoint"]))
--- a/api/README.md
+++ b/api/README.md
@@ -91,7 +91,7 @@ into JSON error messages and HTTP error codes as appropriate. See
 
 The `dp-openapi.yaml` file should always contain the definitive OpenAPI on what
 is currently implemented. This file is best edited via
-[Swagger Editor](http://editor.swagger.io/). Indeed, that's a great place
+[Swagger Editor](https://editor.swagger.io/). Indeed, that's a great place
 to test out the APIs as well.
 
 ## References
--- a/api/dp-openapi.yaml
+++ b/api/dp-openapi.yaml
@@ -166,6 +166,29 @@ paths:
         default:
           $ref: '#/components/responses/UnexpectedError'
 
+  /projects/pagerounds:
+    get:
+      tags:
+      - project
+      description: Returns a list of valid page rounds
+      responses:
+        200:
+          description: List of page rounds
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  type: string
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '429':
+          $ref: '#/components/responses/RateLimitExceeded'
+        default:
+          $ref: '#/components/responses/UnexpectedError'
+
   /projects/states:
     get:
       tags:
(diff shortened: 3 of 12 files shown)


cpeel:
This still needs the Swagger docs updated.

@70ray - does this look like something you can use in the new image viewer?

scripts/api.js.php
```diff
@@ -0,0 +1,30 @@
+<?php
+$relPath="../pinc/";
+include_once($relPath."base.inc");
+
+/* This file is a JavaScript file, but to pull in variables in
+ * site_vars.php it has a PHP extension. It's important that we
+ * return the right MIME type, however, so the browser reads it
+ * as a JS file.
+ */
```
70ray, line 9: I think you could do this by assigning the function to a variable after the page has loaded, so code_url was then available as a js_data variable. I'll try it  out to see if this makes any sense.

70ray: I've tried it out. In the example it will work if you call makeApiAjaxSettings inside the $(function(){ ... }); and have a line in html_page_common like
```
    echo "<script>\n codeUrl = '$code_url'; </script>\n";
```
so api.js is
```javascript
/*global $ codeUrl */
/* exported makeApiAjaxSettings */

function makeApiAjaxSettings(apiUrl, queryParams) {
    let queryParamString = "";
    if(queryParams)
    {
        queryParamString = "&" + $.param(queryParams);
    }
    settings = {
        "url": codeUrl + "/api/index.php?url=" + apiUrl + queryParamString,
        "dataType": "json",
        "headers": {
            "X-API-KEY": "SESSION"
        }
    }
    return settings;
}
```
and the test function is
```javascript
$(function () {
    let settings = makeApiAjaxSettings("v1/projects/projectID45c225f598e32/pages");
    $.ajax(
        settings
    ).then(function(data) {
        console.log(data)
    });
});
```
butther could be a reason for not doing it like that.

cpeel: I like that approach better. I was trying to avoid putting JS code into every page if I could avoid it -- mostly because of some memory around it being not-ideal for use with [Content Security Policies](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy) if we wanted to enable those.

I wonder, though, if we we should be instantiating a global settings object that contains a lot of what's in `site_vars.php` already. If we did that I would probably implement it similar to what I did here -- PHP masquerading as JS -- but that's also probably not as bad as this as it would be a simple object creation (probably using JSON) and less actual code. I may play with this approach a bit but will fall back to yours if mine doesn't seem like it pans out.

cpeel: After looking at the contents of `site_vars.php` I decided not to over-optimize the solution and went with your approach. If we decide we need more variables from that file exposed in JS we might move to an object later. Thanks for the suggestion!

cpeel:
I've added the Swagger docs and this branch is ready to merge if it looks good to others.

PR #390, opened 2020-08-24
Move round colors to theme

cpeel:
This moves the per-stage table colors from the code into `theme.less` which allows them to be (optionally) overridden by the theme. There are zero visual changes to the user in this MR, but this enables some work @srjfoo is doing.

[mini-rant]Why are these different? Why are P1, P2, and F1 the same but different than P3 and F2? Why are PP, PPV, and SR all different from everything else?[/mini-rant]

This is viewable in the [move-round-colors-to-styles](https://www.pgdp.org/~cpeel/c.branch/move-round-colors-to-styles) sandbox.

### DIFF
--- a/pinc/Pool.inc
+++ b/pinc/Pool.inc
@@ -20,10 +20,6 @@ class Pool extends Stage
         $access_change_callback,
         $description,
         $document,
-        $listing_bgcolors,
-            // These parameters have the same semantics as the
-            // corresponding parameters of the Stage constructor.
-            // See Stage.inc for documentation.
 
         $project_checkedout_state,
         $project_available_state,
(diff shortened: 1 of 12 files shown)


srjfoo:
> [mini-rant]Why are these different? Why are P1, P2, and F1 the same but different than P3 and F2? Why are PP, PPV, and SR all different from everything else?[/mini-rant]

At a guess, it's historical. No clue whether one color pair was for R1 and the other for R2, but P3 and F2 correspond to the expert rounds, and P1, P2 and F1 to the "learning and practicing" rounds. At least that's my guess.

Looks good to me!

70ray (approved):

PR #391, opened 2020-08-25
Define font sizes in javascript.

70ray:
The order of the elements in the font sizes array defined in php
is lost when transferred to javascript so the sizes in the
selector appear in the order of the keys which is different to the font sizes. Rather than adding more
complexity to fix this define the sizes in a simple way in js.

### DIFF
--- a/pinc/page_controls.inc
+++ b/pinc/page_controls.inc
@@ -84,8 +84,6 @@ function get_proofreading_interface_data_js()
     $font_data = [
         "faces" => get_font_styles(),
         "faceFamilies" => get_full_font_families(),
-        "sizes" => get_font_sizes(),
-        "sizeFamilies" => get_font_sizes_css(),
     ];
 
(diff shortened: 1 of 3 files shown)


pinc/page_controls.inc
```diff
@@ -84,8 +84,6 @@ function get_proofreading_interface_data_js()
     $font_data = [
         "faces" => get_font_styles(),
         "faceFamilies" => get_full_font_families(),
-        "sizes" => get_font_sizes(),
```
cpeel, line 87: If you do:
```
"sizes" => [ get_font_sizes() ],
"sizeFamilies" => [ get_font_sizes_css() ],
```
The dicts inside the arrays will be sorted like they are in PHP.

70ray: Or we could just transfer the sizes without the indices.

PR #393, opened 2020-08-27
Update queries in some tools files

chrismiceli:
This was all manual and is probably error prone so I've deployed it to my sandbox here:  https://www.pgdp.org/~chrismiceli/c.branch/tools-query-update

I did notice what looked like 2 issues in the existing uber queries that may happened to have worked out, note the missing ending ' character:
https://github.com/DistributedProofreaders/dproofreaders/blob/master/tools/project_manager/edit_uberproject.php#L347
https://github.com/DistributedProofreaders/dproofreaders/blob/master/tools/project_manager/edit_uberproject.php#L384
I hope I didn't update them erroneously.

### DIFF
--- a/tools/project_manager/add_files.php
+++ b/tools/project_manager/add_files.php
@@ -249,7 +249,8 @@ class Loader
 
         // Get the set of all 'image' fields in the page-table.
         $db_entries = array();
-        $sql = sprintf("SELECT image,master_text FROM %s", DPDatabase::esacpe($this->projectid));
+        validate_projectID($this->projectid);
+        $sql = sprintf("SELECT image,master_text FROM %s", $this->projectid);
         $res = DPDatabase::query($sql);
         while( list($image,$text) = mysqli_fetch_row($res) )
         {
@@ -382,11 +383,12 @@ class Loader
         // --------------------------------
 
         // Find out how long the 'image' field is.
+        validate_projectID($this->projectid);
         $sql = sprintf("
             SELECT image
             FROM %s
             LIMIT 0
-        ", DPDatabase::escape($this->projectid));
+        ", $this->projectid);
         $res = DPDatabase::query($sql);
         $field_data = mysqli_fetch_field_direct($res, 0);
         $this->image_field_len = $field_data->length;
--- a/tools/project_manager/add_files.php
+++ b/tools/project_manager/add_files.php
@@ -157,7 +157,10 @@ if ( $loading_tpnv )
     system("cp *.jpg " . escapeshellarg($dest_project_dir));
     echo "</pre>\n";
 
-    $result = mysqli_query(DPDatabase::get_connection(), "UPDATE projects SET state = 'project_new_waiting_app' WHERE projectid = '$projectid'");
+    $sql = sprintf(
+        "UPDATE projects SET state = 'project_new_waiting_app' WHERE projectid = '%s'",
+        DPDatabase::escape($projectid));
+    $result = DPDatabase::query($sql);
 }
 else
 {
@@ -246,7 +249,8 @@ class Loader
 
         // Get the set of all 'image' fields in the page-table.
         $db_entries = array();
-        $res = mysqli_query(DPDatabase::get_connection(), "SELECT image,master_text FROM $this->projectid");
+        $sql = sprintf("SELECT image,master_text FROM %s", DPDatabase::esacpe($this->projectid));
+        $res = DPDatabase::query($sql);
         while( list($image,$text) = mysqli_fetch_row($res) )
         {
             $db_entries[$image] = $text;
@@ -378,11 +382,12 @@ class Loader
         // --------------------------------
 
         // Find out how long the 'image' field is.
-        $res = mysqli_query(DPDatabase::get_connection(), "
+        $sql = sprintf("
             SELECT image
-            FROM $this->projectid
+            FROM %s
             LIMIT 0
-        ") or die(DPDatabase::log_error());
+        ", DPDatabase::escape($this->projectid));
+        $res = DPDatabase::query($sql);
         $field_data = mysqli_fetch_field_direct($res, 0);
         $this->image_field_len = $field_data->length;
 
--- a/tools/project_manager/add_files.php
+++ b/tools/project_manager/add_files.php
@@ -250,7 +250,7 @@ class Loader
         // Get the set of all 'image' fields in the page-table.
         $db_entries = array();
         validate_projectID($this->projectid);
-        $sql = sprintf("SELECT image,master_text FROM %s", $this->projectid);
+        $sql = "SELECT image,master_text FROM $this->projectid";
         $res = DPDatabase::query($sql);
         while( list($image,$text) = mysqli_fetch_row($res) )
         {
@@ -384,11 +384,10 @@ class Loader
 
         // Find out how long the 'image' field is.
         validate_projectID($this->projectid);
-        $sql = sprintf("
+        $sql = "
             SELECT image
-            FROM %s
-            LIMIT 0
-        ", $this->projectid);
+            FROM $this->projectid
+            LIMIT 0";
         $res = DPDatabase::query($sql);
         $field_data = mysqli_fetch_field_direct($res, 0);
         $this->image_field_len = $field_data->length;
--- a/tools/project_manager/automodify.php
+++ b/tools/project_manager/automodify.php
@@ -230,16 +230,15 @@ while ( $project = mysqli_fetch_assoc($allprojects) ) {
             SELECT image
             FROM $projectid
             WHERE state IN ('%s','%s')
-                AND %d <= %d
+                AND $round->time_column_name <= %d
             ORDER BY image ASC",
-            DPDatabase::escape($round->page_out_state),
-            DPDatabase::escape($round->page_temp_state),
-            $round->time_column_name,
+            $round->page_out_state,
+            $round->page_temp_state,
             $max_reclaimable_time);
-        $res = DPDatabase::query($sql);
-        if ( !$res )
-        {
-            echo DPDatabase::log_error(), "\n";
+        try {
+            $res = DPDatabase::query($sql);
+        } catch(DPQueryError $error) {
+            echo "$error->message\n";
             echo "Skipping further processing of this project.\n";
             continue;
         }
--- a/tools/project_manager/automodify.php
+++ b/tools/project_manager/automodify.php
@@ -102,12 +102,16 @@ echo "<pre>\n";
 
 if ($one_project) {
     $verbose = $GLOBALS['testing'];
-    $condition = "projectid = '$one_project'";
+    $condition = sprintf("projectid = '%s'", DPDatabase::escape($one_project));
 
     // log tracetimes
     $tracetime = time();
-    mysqli_query(DPDatabase::get_connection(), "INSERT INTO job_logs (filename, tracetime, event, comments)
-               VALUES ('automodify.php', $tracetime, 'BEGIN', 'running for single proj $one_project')");
+    $sql = sprintf("
+        INSERT INTO job_logs (filename, tracetime, event, comments)
+        VALUES ('automodify.php', %d, 'BEGIN', 'running for single proj %s')",
+        $tracetime,
+        DPDatabase::escape($one_project));
+    DPDatabase::query($sql);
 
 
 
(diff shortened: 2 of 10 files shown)


chrismiceli:
sandbox up to date at  https://www.pgdp.org/~chrismiceli/c.branch/tools-query-update

tools/project_manager/add_files.php
```diff
@@ -378,11 +382,12 @@ function analyze()
         // --------------------------------
 
         // Find out how long the 'image' field is.
-        $res = mysqli_query(DPDatabase::get_connection(), "
+        $sql = sprintf("
             SELECT image
-            FROM $this->projectid
```
cpeel, line 383: We can't escape table names like this. Instead, we should call `validate_projectID()` before we build the SQL to validate it, eg:
```
validate_projectID($this->projectid);
$sql = "
    SELECT image
    FROM $this->projectid
";
```

tools/project_manager/automodify.php
```diff
@@ -217,13 +226,17 @@ function ensure_project_blurb( $project )
...
-        $res = mysqli_query(DPDatabase::get_connection(), "
+        $sql = sprintf("
             SELECT image
             FROM $projectid
-            WHERE state IN ('$round->page_out_state','$round->page_temp_state')
-                AND $round->time_column_name <= $max_reclaimable_time
-            ORDER BY image ASC
-        ");
+            WHERE state IN ('%s','%s')
+                AND %d <= %d
```
cpeel, line 233: We can't escape column names with `DPDatabase::escape()` . Since the column name is defined in the round, I wouldn't escape it but leave it directly into the string.

tools/project_manager/automodify.php
```diff
@@ -217,13 +226,17 @@ function ensure_project_blurb( $project )
...
+            ORDER BY image ASC",
+            DPDatabase::escape($round->page_out_state),
+            DPDatabase::escape($round->page_temp_state),
+            $round->time_column_name,
+            $max_reclaimable_time);
+        $res = DPDatabase::query($sql);
```
cpeel, line 239: This is a case where we don't want to raise an exception because the below handles it. It would be better to rewrite this and the next few lines in a try block:
```php
try {
} catch(DPQueryError $error) {
    echo "$error->message\n";
    echo "Skipping further processing of this project.\n";
    continue;
}
```
The above is utterly untested and may be using slightly wrong syntax as I'm doing it from memory.

tools/project_manager/autorelease.inc
```diff
@@ -563,33 +566,37 @@ function AP_setup( $round )
...
+    $sql = sprintf("
         INSERT INTO active_page_counts
         SELECT
-            '$projectid',
-            SUM( state != '{$round->page_save_state}' ) as pages
-        FROM $projectid
-    ") or die(DPDatabase::log_error());
+            '%s',
+            SUM( state != '%s' ) as pages
+        FROM %s",
```
cpeel, line 580: Test for valid projectid rather than escaping it here.

tools/project_manager/autorelease.inc
```diff
@@ -563,33 +566,37 @@ function AP_setup( $round )
...
-    $res = mysqli_query(DPDatabase::get_connection(), "
+    $sql = sprintf("
         SELECT
-            SUM(projects.state='{$round->project_available_state}') as projects,
+            SUM(projects.state='%s') as projects,
             SUM(active_page_counts.pages) as pages
         FROM projects NATURAL JOIN active_page_counts
-        WHERE $cooked_project_selector
-    ") or die(DPDatabase::log_error());
+        WHERE %s",
```
cpeel, line 596: Leave `$cooked_project_selector` in the string since it is built elsewhere and can't be escaped.

tools/project_manager/diff.php
```diff
@@ -74,13 +74,19 @@
...
 }
 
-$query = "
-    SELECT $L_text_column_name, $R_text_column_name,
-           $L_user_column_name, $R_user_column_name
-    FROM $projectid
-    WHERE image='$image'";
-
-$res = mysqli_query(DPDatabase::get_connection(), $query) or die(DPDatabase::log_error());
+$query = sprintf("
```
cpeel, line 77: We can't escape column or table names. We should keep the embedded column names as-is and validate the projectID before building the query.

cpeel: Ditto below two changes as well.

cpeel: Actually this is true for several other spots too.

tools/project_manager/autorelease.inc
chrismiceli: The only reason I pulled all of these out was in case there was a % in the text that would be expanded before being passed to sprintf. Then it could run into an issue trying to sprintf the string. However, in these cases I think we are ok.

chrismiceli:
I didn't update uber project file as we will delete it before merging this MR.

tools/project_manager/autorelease.inc
cpeel: Just insert `$projectid` here directly since it has been validated and is the table name.

tools/project_manager/diff.php
```diff
@@ -74,13 +74,16 @@
...
 
-$query = "
+validate_projectID($projectid);
+$query = sprintf("
     SELECT $L_text_column_name, $R_text_column_name,
-           $L_user_column_name, $R_user_column_name
-    FROM $projectid
-    WHERE image='$image'";
+        $L_user_column_name, $R_user_column_name
+    FROM %s
```
cpeel, line 81: Insert `$projectid` directly.

tools/project_manager/diff.php
```diff
@@ -176,8 +179,14 @@ function do_navigation($projectid, $image, $L_round_num, $R_round_num, $L_user_c
...
     $navigation_text .= "\n" . _("Jump to") . ": <select name='jumpto' onChange='$jump_to_js'>\n";
 
-    $query = "SELECT image, $L_user_column_name, ($L_text_column_name = $R_text_column_name) AS is_empty_diff FROM $projectid ORDER BY image ASC";
-    $res = mysqli_query(DPDatabase::get_connection(),  $query) or die(DPDatabase::log_error());
+    validate_projectID($projectid);
+    $query = sprintf("
+        SELECT image,
+        $L_user_column_name,
+            ($L_text_column_name = $R_text_column_name) AS is_empty_diff
+        FROM %s ORDER BY image ASC",
```
cpeel, line 187: Insert `$projectid` directly.

tools/project_manager/diff.php
```diff
@@ -280,8 +289,14 @@ function can_see_names_for_page($projectid, $image)
             }
             $fields .= $round->user_column_name;
         }
-        $query = "SELECT $fields from $projectid WHERE image = '$image'";
-        $res = mysqli_query(DPDatabase::get_connection(), $query) or die(DPDatabase::log_error());
+
+        validate_projectID($projectid);
+        $query = sprintf("
+            SELECT $fields from %s WHERE image = '%s'",
```
cpeel, line 295: Insert `$projectid` directly (and let's also capitalize `FROM` while we're here).

tools/project_manager/downloadproofed.php
```diff
@@ -18,7 +18,12 @@
     $text_column_name = $round->text_column_name;
 }
 
-$result = mysqli_query(DPDatabase::get_connection(), "SELECT $text_column_name FROM $project WHERE image = '$image'"); 
+validate_projectID($project);
+$sql = sprintf("
+    SELECT $text_column_name FROM %s WHERE image = '%s'",
```
cpeel, line 23: Use `$projectid` directly.

tools/project_manager/edit_project_word_lists.php
```diff
@@ -500,9 +500,12 @@ function number_of_pages_in_round($round)
         } else {
             $round_column="master_text";
         }
-        $sql = "select count(*) from $this->projectid where $round_column <> ''";
+        validate_projectID($this->projectid);
+        $sql = sprintf(
+            "select count(*) from %s where $round_column <> ''",
```
cpeel, line 505: Use `$this->projectid` directly.

tools/project_manager/handle_bad_page.php
```diff
@@ -41,7 +41,12 @@
 
 if (!$resolution) {
     //Find out information about the bad page report
-    $result = mysqli_query(DPDatabase::get_connection(), "SELECT * FROM $projectid WHERE image='$image'");
+    validate_projectID($projectid);
+    $sql = sprintf(
+        "SELECT * FROM %s WHERE image='%s'",
```
cpeel, line 46: Use `$projectid` directly.

tools/project_manager/page_compare.php
```diff
@@ -106,15 +110,18 @@ function render()
...
 
-        $res = mysqli_query(DPDatabase::get_connection(), "
+        validate_projectID($this->projectid);
+        $sql = sprintf("
             SELECT image, $L_text_column_name, $R_text_column_name
-            FROM $this->projectid
-            WHERE $condition
-            ORDER BY image ASC
-        ") or die(DPDatabase::log_error());
+            FROM %s
```
cpeel, line 119: Use `$this->projectid` directly.

tools/project_manager/page_compare.php
```diff
@@ -106,15 +110,18 @@ function render()
...
+            WHERE %s
```
cpeel, line 120: Use `$condition` directly.

tools/project_manager/page_operations.inc
```diff
@@ -10,11 +10,14 @@ include_once($relPath.'page_tally.inc');
...
 function page_del( $projectid, $image )
 {
-    $result = mysqli_query(DPDatabase::get_connection(), "
+    validate_projectID($projectid);
+    $sql = sprintf("
         SELECT image
-        FROM $projectid
-        WHERE image = '$image'
-    ");
+        FROM %s
```
cpeel, line 16: Use `$projectid` directly.

tools/project_manager/page_operations.inc
```diff
@@ -48,11 +51,15 @@ function page_clear( $projectid, $image )
...
 
-    $result = mysqli_query(DPDatabase::get_connection(), "
-        SELECT state, {$round->user_column_name}
-        FROM $projectid
-        WHERE image = '$image'
-    ");
+    validate_projectID($projectid);
+    $sql = sprintf("
+        SELECT state, %s
+        FROM %s
```
cpeel, line 57: Here too.

cpeel (commented):
This is looking good. To make the SQL more readable, I don't want to use sprintf-placeholders for table or column names. Since we're validating the project table names before building the SQL. We can use the table names directly:
```php
validate_projectid($projectid);
$sql = "
    SELECT *
    FROM $projectid
";
```

cpeel:
I'm not entirely sure what you did, but you've pulled in commits that are already in the code and now there is a merge conflict.

When I update a branch, I rebase it instead of merging, eg:
```bash
git fetch upstream
git checkout $BRANCHNAME
git rebase upstream/master
```

chrismiceli:
> I'm not entirely sure what you did, but you've pulled in commits that are already in the code and now there is a merge conflict.
> 
> When I update a branch, I rebase it instead of merging, eg:
> 
> ```shell
> git fetch upstream
> git checkout $BRANCHNAME
> git rebase upstream/master
> ```

Went ahead and rebased. Hope that helps. Where I work they don't like rebases since you can hide some changes, but I prefer the cleaner commit history of rebases, so I'll use them here.

tools/project_manager/diff.php
```diff
@@ -176,8 +178,13 @@ function do_navigation($projectid, $image, $L_round_num, $R_round_num, $L_user_c
     $navigation_text .= "\n<input type='hidden' name='format' value='$format'>";
     $navigation_text .= "\n" . _("Jump to") . ": <select name='jumpto' onChange='$jump_to_js'>\n";
 
-    $query = "SELECT image, $L_user_column_name, ($L_text_column_name = $R_text_column_name) AS is_empty_diff FROM $projectid ORDER BY image ASC";
-    $res = mysqli_query(DPDatabase::get_connection(),  $query) or die(DPDatabase::log_error());
+    validate_projectID($projectid);
+    $query = "
+        SELECT image,
+        $L_user_column_name,
```
cpeel, line 184: Please indent this 4 spaces.

tools/project_manager/diff.php
```diff
@@ -176,8 +178,13 @@ function do_navigation($projectid, $image, $L_round_num, $R_round_num, $L_user_c
...
+            ($L_text_column_name = $R_text_column_name) AS is_empty_diff
+        FROM $projectid ORDER BY image ASC";
```
cpeel, line 186: Move `ORDER BY` onto next line.

tools/project_manager/diff.php
```diff
@@ -280,8 +287,13 @@ function can_see_names_for_page($projectid, $image)
             }
             $fields .= $round->user_column_name;
         }
-        $query = "SELECT $fields from $projectid WHERE image = '$image'";
-        $res = mysqli_query(DPDatabase::get_connection(), $query) or die(DPDatabase::log_error());
+
+        validate_projectID($projectid);
+        $query = sprintf("
+            SELECT $fields from $projectid WHERE image = '%s'",
+            $fields,
```
cpeel, line 294: Delete this line.

cpeel (commented):
This looks really good. I found 2 small (very minor) formatting things and one incorrect `sprintf()` and this is ready to merge. Thanks for working on it!

PR #399, opened 2020-08-30
Fix quiz character suites

srjfoo:
Now that the comprehensive Extended European Latin character suite is
no longer available, update `Quiz.inc` to point to Extended European
Latin B.

Issue discovered when trying to test an unrelated update. Can be tested at [fix-quiz-charsuites](https://www.pgdp.org/~srjfoo/c.branch/fix-quiz-charsuites/quiz/start.php)

### DIFF
--- a/pinc/Quiz.inc
+++ b/pinc/Quiz.inc
@@ -203,7 +203,7 @@ class QuizCharSuites
     public function get_charsuites()
     {
         $charsuites[] = CharSuites::get('basic-latin');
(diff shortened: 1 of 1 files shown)


chrismiceli:
No other references are discovered with a global search of files?

cpeel:
> No other references are discovered with a global search of files?

Correct, this is the only place we've hard-coded a character suite besides basic-latin. The quizzes are a bit of a snowflake.

PR #400, opened 2020-08-30
Update F1 access email

srjfoo:
Per request from the GM, clarify the pages-per-day portion of the
email that is sent automatically to new F1 formatters. This is just a small wording change.

Per suggestion from @cpeel, updated to include code to automatically include the daily page limit, since that's now stored in the code.

Test in [update-f1-access-email](https://www.pgdp.org/~srjfoo/c.branch/update-f1-access-email/). To test, go to your stats page. If you already have F1 access, disable it, and re-enable it. If you don't, just enable it. You should be shown a copy of the email that would have been sent if email were enabled for the test site. The daily page limit has been set to 5 pages for the purposes of this test.

### DIFF
--- a/pinc/access_change_emails.inc
+++ b/pinc/access_change_emails.inc
@@ -82,7 +82,7 @@ function f1_access_change_email($user, $access_change)
 
     if($daily_limit)
     {
-        $message[] = "***" . sprintf( _("Please keep in mind that we have a limit in F1 of no more than %d pages per day."), $daily_limit) . "*** " .
+        $message[] = "***" . sprintf( _("Please keep in mind that we currently have a limit in F1 of no more than %d pages per day."), $daily_limit) . "*** " .
             _("(This is a daily limit, not an average.) The limit is there to make sure that we have a variety of projects available in the round at all times.");
     // The above paragraph will be added to the email only if the daily limit
     // is a postiive integer. Currently, if the daily limit value is zero or
--- a/pinc/access_change_emails.inc
+++ b/pinc/access_change_emails.inc
@@ -84,10 +84,10 @@ function f1_access_change_email($user, $access_change)
     {
         $message[] = "***" . sprintf( _("Please keep in mind that we currently have a limit in F1 of no more than %d pages per day."), $daily_limit) . "*** " .
             _("(This is a daily limit, not an average.) The limit is there to make sure that we have a variety of projects available in the round at all times.");
-    // The above paragraph will be added to the email only if the daily limit
-    // is a postiive integer. Currently, if the daily limit value is zero or
-    // NULL, the paragraph is not added. This code will need to be adjusted
-    // if there is ever a need for them to be treated differently.
+        // The above paragraph will be added to the email only if the daily limit
+        // is a postiive integer. Currently, if the daily limit value is zero or
+        // NULL, the paragraph is not added. This code will need to be adjusted
+        // if there is ever a need for them to be treated differently.
     }
 
     $message[] = sprintf(_("Before you begin formatting, please read through the Formatting Guidelines: %1\$s . You may also view and print our two-page Formatting Summary: %2\$s to keep on hand while you're formatting. The Guidelines contain a lot of information, so please take your time and ask questions if there's something you don't understand."),
--- a/pinc/access_change_emails.inc
+++ b/pinc/access_change_emails.inc
@@ -58,7 +58,9 @@ function pp_access_change_email($user, $access_change)
 
 function f1_access_change_email($user, $access_change)
 {
-    global $site_name, $site_abbreviation;
+    global $site_name, $site_abbreviation, $Round_for_round_id_;
+
+    $daily_limit = $Round_for_round_id_['F1']->daily_page_limit;
 
     // only send emails on grant
     if($access_change != "grant")
@@ -78,8 +80,15 @@ function f1_access_change_email($user, $access_change)
 
     $message[] = _("After all of the pages of a project are completed in F1, the project moves into F2 for review and any needed corrections. F2 is the final round of formatting before Post-Processing.");
 
-    $message[] = "***" . _("Please keep in mind that we have a 35 page-per-day limit in F1.") . "*** " .
-        _("(This is a daily limit, not an average.) The limit is there to make sure that we have a variety of projects available in the round at all times.");
+    if($daily_limit)
+    {
+        $message[] = "***" . sprintf( _("Please keep in mind that we have a limit in F1 of no more than %d pages per day."), $daily_limit) . "*** " .
+            _("(This is a daily limit, not an average.) The limit is there to make sure that we have a variety of projects available in the round at all times.");
+    // The above paragraph will be added to the email only if the daily limit
+    // is a postiive integer. Currently, if the daily limit value is zero or
+    // NULL, the paragraph is not added. This code will need to be adjusted
+    // if there is ever a need for them to be treated differently.
+    }
 
     $message[] = sprintf(_("Before you begin formatting, please read through the Formatting Guidelines: %1\$s . You may also view and print our two-page Formatting Summary: %2\$s to keep on hand while you're formatting. The Guidelines contain a lot of information, so please take your time and ask questions if there's something you don't understand."),
         "https://www.pgdp.net/wiki/DP_Official_Documentation:Formatting/Formatting_Guidelines",


srjfoo:
> No sandbox link, because this can't be tested without that being merged.

Still no sandbox link, but I did figure out that one of my test accounts did have up-to-date quizzes, so I was able to test it.

pinc/access_change_emails.inc
```diff
@@ -78,7 +78,7 @@ function f1_access_change_email($user, $access_change)
 
     $message[] = _("After all of the pages of a project are completed in F1, the project moves into F2 for review and any needed corrections. F2 is the final round of formatting before Post-Processing.");
 
-    $message[] = "***" . _("Please keep in mind that we have a 35 page-per-day limit in F1.") . "*** " .
+    $message[] = "***" . _("Please keep in mind that we have a limit in F1 of no more than 35 pages per day.") . "*** " .
```
cpeel, line 81: Now that we have this enforced in the code, we should consider making this dynamic, eg:
```
$daily_limit = $Round_for_round_id_['F1'];
if($daily_limit)
{
    $message[] = "***" . sprintf(_("Please keep in mind that we have a limit in F1 of no more than %d pages per day."), $daily_limit) . "*** " .
        _("(This is a daily limit, not an average.) The limit is there to make sure that we have a variety of projects available in the round at all times.");
}
```

$Round_for_round_id_ is a global so it would need to be declared at the top of the function.

cpeel: Note that to test this on TEST you'll need to update `pinc/stages.inc` thusly:
```diff
diff --git a/pinc/stages.inc b/pinc/stages.inc
index a99b993..9b2a68c 100644
--- a/pinc/stages.inc
+++ b/pinc/stages.inc
@@ -203,7 +203,7 @@ new Round(
     _("The page-texts have already been proofread, but now need to be formatted with markup which may be specific to the project."),
     'formatting_guidelines.php',
     $pi_tools_for_F,
-    NULL, /* daily_page_limit */
+    35, /* daily_page_limit */
     array(),
     array(
              0 => _('Formatting Neophyte'),
```

jmdyck: > ```
> $daily_limit = $Round_for_round_id_['F1'];
> if($daily_limit)
> {
>     ... $daily_limit ...
> }
> ```

more like:
```
$round = $Round_for_round_id_['F1'];
if ($round->has_a_daily_page_limit())
{
    ... $round->daily_page_limit ...
}
```

cpeel: Well, really what I meant (and didn't get right) was:
```php
$daily_limit = $Round_for_round_id_['F1']->daily_page_limit;
if($daily_limit)
{
    ... $daily_limit ...
}
```

jmdyck: It depends on whether you want to handle the possibility of a daily limit being set to zero. (Which, granted, seems unlikely to ever occur, but the existing code allows for it.) `if($daily_limit)` would treat a limit of 0 the same as a NULL limit (i.e., no limit), but they're very different things.

cpeel: In either case I would not include that paragraph, which my code does. But I leave it up to Sharon.

srjfoo: I felt like the question of whether there should be a paragraph there or not should be up to @lhamilton1, and ran the question past her. I don't think that the paragraph should be there if there's not a limit, but the question was whether there should be an `else` clause there to mention the possibility of a daily page limit. She chose just omitting the paragraph altogether if there is no daily page limit.

As for the question of whether to treat NULL and zero the same (I've seen multiple cases where zero means "no limit"), I don't know. It seems that that should be a design decision. Do we have (or want) another way to quickly and easily disable all proofing in any or all rounds without bringing the rest of the site down?

jmdyck: > As for the question of whether to treat NULL and zero the same (I've seen multiple cases where zero means "no limit"),

In other (non-DP) contexts, you mean?

> I don't know. It seems that that should be a design decision. Do we have (or want) another way to quickly and easily disable all proofing in any or all rounds without bringing the rest of the site down?

To clarify, the existing code that enforces the daily limits *does* treat NULL and zero differently. NULL means "there is no limit" (which you might think of as limit=infinity), and zero means "there is a limit, and its value is zero" (so you're not going to be able to save-as-done anything, unless you first reopen some previously saved pages).

With respect to this email, it makes sense that in the NULL case (no limit), you wouldn't add a paragraph about a limit. The question is what to do in the zero case.
- Certainly, it would be odd to say "Please keep in mind that we have a limit of 0 pages per day."
- A human would probably say something like "... we normally have a limit of 35 pages per day, though it's currently set to 0 for some reason." But programmatically, if it's set to 0, the code doesn't know that it's normally 35.
- So probably the best thing (in the case of limit=0) is, as cpeel suggests, to just say nothing about a limit, and rely on the site to convey (e.g., via a news item on the F1 page) what's going on.

However, I don't think  it's a good idea to simply use `if ($daily_limit)` to accomplish the latter, as it doesn't convey that it's handling (excluding) two distinct cases that are both represented by 'falsey' values. A developer that isn't familiar with the daily limits might copy that idiom to somewhere where it *doesn't* make sense to treat the two cases the same. And a developer that *is* familiar with the daily limits might look at the condition and think it's a bug. So at the very least there should be a comment that conveys the intended semantics. But I think it would be better to use the established idiom for dealing with daily limits, i.e. don't access `$round->daily_page_limit` unless you've established that `$round->has_a_daily_page_limit()` is true. So, e.g.:
```
$round = $Round_for_round_id_['F1'];
if ($round->has_a_daily_page_limit() && $round->daily_page_limit > 0)
{
    ... $round->daily_page_limit ...
}
// If an F1 daily page limit exists, but is zero,
// we don't say anything about a limit here, because
// (a) mentioning a limit of 0 would be odd, and
// (b) it's probably a temporary situation.
```
or
```
$round = $Round_for_round_id_['F1'];
if ($round->has_a_daily_page_limit())
{
    if ($round->daily_page_limit > 0)
    {
        $message[] = ...
    }
    else
    {
        // An F1 daily page limit exists, but it's zero.
        // In this case, we don't say anything about a limit here,
        // because: [etc]
    }
}

```

(If PHP had a distinguished 'infinity' value, like JavaScript does, I would probably have used that instead of NULL, and this would have been a somewhat different discussion.)

jmdyck:
Also, if the idea is that they might keep this email for reference, you might want the message to say that the limit is *currently* 35.

pinc/access_change_emails.inc
cpeel: I find these code samples overly verbose and unnecessary -- but we've often disagreed on things like this. I defer to Sharon, who is making the code change.

lhamilton1:
I like the idea of using "currently".

pinc/access_change_emails.inc
srjfoo: > ```
> // If an F1 daily page limit exists, but is zero,
> // we don't say anything about a limit here, because
> // (a) mentioning a limit of 0 would be odd, and
> // (b) it's probably a temporary situation.
> ```

I'm slow sometimes, but hopefully I'll get there eventually. As things work currently, if the daily page limit is set to `0`, it's set to `0`, and no pages can be proofed in that round. So, as implemented, `NULL` and `0` are not the same. If we wish for `0` and `NULL` to be treated the same, that needs to be changed elsewhere in the code.

If we're going to leave it as-is, with the two meaning different things, I don't think that it should be ignored, but that the message sent ought to be a reassurance that it's a temporary situation. So, perhaps the question should be, "Should they be treated the same?".

jmdyck: > I'm slow sometimes, but hopefully I'll get there eventually. As things work currently, if the daily page limit is set to `0`, it's set to `0`, and no pages can be proofed in that round.

Roughly, yes, but not precisely. If you first reopen some pages that you previously saved in that round, then you'll be able to save-as-done that many. But your 'net' pages saved won't be allowed to go above zero.

> So, as implemented, `NULL` and `0` are not the same.

Right.

> If we wish for `0` and `NULL` to be treated the same, that needs to be changed elsewhere in the code.

Yes. But, to be clear, that's not what's being discussed above (at least, I don't think so). Rather, the main question is how the two situations (no limit vs limit of zero) should be treated *by this script*. (And then there's the subordinate question: *if* they are to be treated the same here, how to express that in code.)

> So, perhaps the question should be, "Should they be treated the same?".

If you mean "Should they be treated the same *by this script*?": That's up to you.

If you mean "Should they be treated the same *throughout the code*?": Presumably, we're not going to get rid of the ability to represent a round with no limit (because pgdp currently has 4 such). So the question is equivalent to "Should `0` represent `no limit` (the same as `NULL` does currently)?", which is then equivalent to "Should we give up the ability to represent limit=0?" I'd say no: you already have it, and you might never use it, but on the other hand, you might (or some other DP site might).

srjfoo: I actually really appreciate all the discussion. It makes me go digging around to figure things out and helps me learn!

srjfoo:
> I like the idea of using "currently".

Done.

pinc/access_change_emails.inc
```diff
@@ -78,8 +80,15 @@ function f1_access_change_email($user, $access_change)
 
     $message[] = _("After all of the pages of a project are completed in F1, the project moves into F2 for review and any needed corrections. F2 is the final round of formatting before Post-Processing.");
 
-    $message[] = "***" . _("Please keep in mind that we have a 35 page-per-day limit in F1.") . "*** " .
-        _("(This is a daily limit, not an average.) The limit is there to make sure that we have a variety of projects available in the round at all times.");
+    if($daily_limit)
+    {
+        $message[] = "***" . sprintf( _("Please keep in mind that we currently have a limit in F1 of no more than %d pages per day."), $daily_limit) . "*** " .
+            _("(This is a daily limit, not an average.) The limit is there to make sure that we have a variety of projects available in the round at all times.");
+    // The above paragraph will be added to the email only if the daily limit
```
cpeel, line 87: Please indent this paragraph 4 more spaces so it aligns with the line above.

srjfoo: Done.

PR #403, opened 2020-09-01
Refinements to the CSRF token

cpeel:
Small set of refinements for the CSRF token from @chrismiceli. Note that we can't set the samesite attribute on the cookie because that was only added to `setcookie()` in PHP 7.3.0.

Available in the [improve-csrf](https://www.pgdp.org/~cpeel/c.branch/improve-csrf) sandbox.


srjfoo:
Tested logout/login, both generally, and pasting in a destination URL while logged out, and then logging in. Also tested changing userprefs. Everything looks good.

70ray (approved):

PR #404, opened 2020-09-02
Title png

70ray:
This makes some adjustments to the view page and text pages.

### DIFF
--- a/scripts/page_browse.js
+++ b/scripts/page_browse.js
@@ -1,6 +1,6 @@
 /* global $ pageBrowserData proofIntData splitControl */
 /* exported viewSplitter hiddenProject projectSelector projectReset
-pageChanger imageControl textControl roundSelect */
+pageChanger imageControl textControl roundSelect pageTitle */
 
 var projectSelector = function() {
     return [
--- a/scripts/page_browse.js
+++ b/scripts/page_browse.js
@@ -9,6 +9,10 @@ var projectSelector = function() {
     ];
 };
 
+function pageTitle() {
+    document.title = proofIntData.strings.displayImageX.replace("%s", pageBrowserData.currentPage);
+}
+
 var hiddenProject = function() {
     return $("<input>", {type: 'hidden', name: 'project', value: pageBrowserData.projectid});
 };
--- a/pinc/page_controls.inc
+++ b/pinc/page_controls.inc
@@ -102,7 +102,6 @@ function get_proofreading_interface_data_js()
         "showImage" => _("Show Image only"),
         "showText" => _("Show Text only"),
         "showImageText" => _("Show Image & Text"),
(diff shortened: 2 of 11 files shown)


tools/display_image.php
```diff
@@ -20,7 +20,7 @@
     $error_message = $exception->getMessage();
 }
 
-$title = _("Display Image for page");
+$title = sprintf(_("Display Image: %s"), $imagefile);
```
cpeel, line 23: `$imagefile` may be blank here. If so we should just have the title be "Display Image" or something similar.

tools/mentors/view_page_text_image.php
```diff
@@ -35,7 +35,8 @@
     "body_attributes" => 'class="no-margin overflow-hidden"',
 ];
 
-slim_header(_("Image and text for page"), $header_args);
+$title = sprintf(_("Display Image: %s"), $imagefile);
```
cpeel, line 38: Ditto `$imagefile` possibly being blank here.

tools/project_manager/displayimage.php
```diff
@@ -28,7 +28,7 @@ function get_mode_js()
     return "var mode = '$mode';";
 }
 
-$title = _("Display Image for page");
+$title = sprintf(_("Display Image: %s"), $imagefile);
```
cpeel, line 31: Here too.

tools/display_image.php
chrismiceli: Is title encoded by the header function? If not may want to encode image here

chrismiceli: Just checked, it is encoded: https://github.com/DistributedProofreaders/dproofreaders/blob/master/pinc/html_page_common.inc#L58 and

70ray: Changed the way this is done.

scripts/page_browse.js
```diff
@@ -9,6 +9,10 @@ var projectSelector = function() {
     ];
 };
 
+function pageTitle() {
+    document.title = proofIntData.strings.displayImageX.replace("%s", pageBrowserData.currentPage);
```
chrismiceli, line 13: replace does interesting things with $ when you use a string replace function as outlined here:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#Specifying_a_string_as_a_parameter
We may want to use a replace function.

70ray: Good point. I think we will get away with it because get_page_image_param() wouldn't let $ through. But it wouldn't hurt to be extra safe.

cpeel: I think this is OK as-is. The warning about `$` is valid but I feel very confident we won't allow special characters like that in page image names.

srjfoo (approved):

PR #406, opened 2020-09-04
1390: Remove Edit link from a page marked bad.

chrismiceli:
remove edit link for bad pages

I didn't see comments on the issue, so if it is not desired, I can close this MR. FYI @srjfoo

### DIFF
--- a/tools/project_manager/page_table.inc
+++ b/tools/project_manager/page_table.inc
@@ -348,12 +348,14 @@ function echo_page_table(
 
         // (It's annoying that we have to pass all those args,
         // or else make them global, which I don't want to do.)
+        $page_is_in_bad_state = page_state_is_a_bad_state($page_state);
 
         // --------------------------------------------
         // Link to proofreading interface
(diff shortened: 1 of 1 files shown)


chrismiceli:
Sandbox here: https://www.pgdp.org/~chrismiceli/c.branch/issue/1390

srjfoo (approved):
Looks good.

Yes, thank you, still wanted. I kept tripping over it when I went looking through old tasks, and meaning to verify whether it's still an issue. This MR prompted me to finally do that testing and verify that it is still the case.

Thanks for fixing it!

tools/project_manager/page_table.inc
```diff
@@ -383,7 +385,6 @@ function echo_page_table(
             echo "<td>$clear_string</td>";
 
             // Bad Page
```
cpeel, line 387: Let's move the comment along with the line and make sure there's a space between this comment and the comment block on line 350.

PR #355, opened 2020-07-17
Implement API for site statistics

cpeel:
This builds on top of the other two MRs (again, into the `api` branch, not `master`). This adds very basic site and round statistics.

To get some basic statistics about the site:
```bash
curl -i -g -H "Accept: application/json" -H "X-API-KEY: review" \
    "https://www.pgdp.org/~cpeel/c.branch/api-part-2/api/index.php?url=v1/stats/site"
```

To get round statistics:
```bash
curl -i -g -H "Accept: application/json" -H "X-API-KEY: review" \
    "https://www.pgdp.org/~cpeel/c.branch/api-part-2/api/index.php?url=v1/stats/site/P1"

```

My primary question is: should the round statistics URL be `v1/stats/site/P1` or `v1/stats/site/round/P1`? Now that I'm looking at it afresh the second one seems better (although I have implemented the first one).

PR #408, opened 2020-09-06
Add a RESTful API

cpeel:
The purpose of a DP API is two-fold:
1. Allow future DP-designed javascript user interfaces to access data - this will be key for the next-generation proofreading interface. This iteration provides functionality needed for @70ray's display image interface to remove the need for page reloads.
2. Allow *approved* volunteers access to the API for building tools and other automation. This iteration provides the functionality for @bpfoley's [Archer's Gap game](https://www.pgdp.net/wiki/P3_Archers).
 
Some important details I have accounted for in the design:
1. The API can be enabled/disabled via `site_vars.php`.
2. API access is restricted per-user and is tied to each user. If The Management decides to grant a user API access, an API key is created and assigned to the user.
3. API access is limited to what the user can see. The API key is used by the code to "become" the user and thus only provide data to the user that they would be able to access on the DP site.
4. API access is rate limited. To prevent automation overwhelming the system, we use rate limiting to prevent more than some number of API requests per unit of time. Right now this is limited to 3600 requests over the course of an hour, so averaging 1 query per second. This is easily configurable.

This branch is the cumulation of several smaller MRs (https://github.com/DistributedProofreaders/dproofreaders/pull/316, https://github.com/DistributedProofreaders/dproofreaders/pull/335, https://github.com/DistributedProofreaders/dproofreaders/pull/355, https://github.com/DistributedProofreaders/dproofreaders/pull/374) although those didn't get a lot of feedback. This code can be accessed via the [api](https://www.pgdp.org/~cpeel/c.branch/api) sandbox either via the web or via the RESTful API endpoint.


70ray (approved):
I haven't looked through all of this in detail but I'll try it with the display image and text functions when it's in the master branch.

PR #412, opened 2020-09-11
Update URLs in feed SDK

cpeel:
This is just to update the URLs in this document. [Task 1927](https://www.pgdp.net/c/tasks.php?action=show&task_id=1927)

You can check it out in the [update-feed-sdk](https://www.pgdp.org/~cpeel/c.branch/update-feed-sdk/faq/feeds_sdk/) sandbox.

### DIFF
--- a/faq/feeds_sdk/index.php
+++ b/faq/feeds_sdk/index.php
@@ -49,14 +49,12 @@ Proofreaders XML Feeds SDK</font></b></td>
 files for Project Gutenberg's Distributed Proofreaders are created as 
 well as focusing mostly on how to implement these XML feeds into your 
 site.</p>
-<p>If you are further interested in XML you may want to look into the 
-follow sites as well:</p>
+<p>If you are further interested in XML and RSS you may want to look into the 
+following sites as well:</p>
 <ul>
-<li><a href="http://www.xml.org">http://www.xml.org</a></li>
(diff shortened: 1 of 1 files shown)


faq/feeds_sdk/index.php
```diff
@@ -49,14 +49,12 @@
 files for Project Gutenberg's Distributed Proofreaders are created as 
 well as focusing mostly on how to implement these XML feeds into your 
 site.</p>
-<p>If you are further interested in XML you may want to look into the 
+<p>If you are further interested in XML and RSS you may want to look into the 
 follow sites as well:</p>
```
srjfoo, line 53: "following" sites (I realize that was in the original)

chrismiceli:
Looks good to me.im guessing localization isn't on this page yet. Out of curiosity, do we have an idea how many people use the feeds?

srjfoo (approved):

faq/feeds_sdk/index.php
cpeel: Good catch - fixed.

cpeel:
> Looks good to me.im guessing localization isn't on this page yet. Out of curiosity, do we have an idea how many people use the feeds?

I'm dubious of the value of this page, but thought we might as well fix the URLs after Sharon found them. According to our web logs, `feeds/backend.php` was hit 8,541 times in August -- so someone still uses it.

PR #414, opened 2020-09-11
Add search forms to the Members and Team listings

cpeel:
This adds a search form the to Members and Team listings. In a separate commit I styled the search forms on the Stats Central page. I'm not enamored with that one so I'm happy if folks think we should just drop that commit. [Task 1926](https://www.pgdp.net/c/tasks.php?action=show&task_id=1926)

See the [member-team-search](https://www.pgdp.org/~cpeel/c.branch/member-team-search/) sandbox.

### DIFF
--- a/stats/members/mbr_list.php
+++ b/stats/members/mbr_list.php
@@ -16,61 +16,70 @@ $order = get_enumerated_param(
 $direction = get_enumerated_param(
     $_GET, 'direction', 'asc', array('asc', 'desc') );
 $mstart = get_integer_param( $_GET, 'mstart', 0, 0, null );
-$uname = @$_REQUEST['uname'];
-$uexact = @$_REQUEST['uexact'];
+$uname = normalize_whitespace(array_get($_GET, 'uname', ''));
+$uexact = array_get($_GET, 'uexact', '') == 'yes';
 
-if (!empty($uname)) {
-    if ($uexact == 'yes')
+if ($uname) {
+    if ($uexact)
     {
-        $where_clause = sprintf("
-            WHERE username='%s'
-        ", mysqli_real_escape_string(DPDatabase::get_connection(), $uname));
+        $where_clause = sprintf("WHERE username = '%s'", DPDatabase::escape($uname));
     }
     else
     {
-        $where_clause = sprintf("
-            WHERE username LIKE '%%%s%%'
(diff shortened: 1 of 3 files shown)


srjfoo:
> ... In a separate commit I styled the search forms on the Stats Central page. I'm not enamored with that one so I'm happy if folks think we should just drop that commit.

I agree. I think the real problem, though, is that there are two tables, side-by-side, that take up more horizontal space than the table below them (which is narrower, if not restricted by the browser window width, than the tables below it). The current version doesn't draw the eye because it doesn't have the color and the extra width from the table layout.

Very cool for the team and member list pages to have a search form on them. Looks good!

srjfoo (approved):

cpeel:
> > ... In a separate commit I styled the search forms on the Stats Central page. I'm not enamored with that one so I'm happy if folks think we should just drop that commit.
> 
> I agree. I think the real problem, though, is that there are two tables, side-by-side, that take up more horizontal space than the table below them (which is narrower, if not restricted by the browser window width, than the tables below it). The current version doesn't draw the eye because it doesn't have the color and the extra width from the table layout.

Sorry, is that a "you agree we should drop the changes to the stats central"?

srjfoo:
> > > ... In a separate commit I styled the search forms on the Stats Central page. I'm not enamored with that one so I'm happy if folks think we should just drop that commit.
> > 
> > 
> > I agree. I think the real problem, though, is that there are two tables, side-by-side, that take up more horizontal space than the table below them (which is narrower, if not restricted by the browser window width, than the tables below it). The current version doesn't draw the eye because it doesn't have the color and the extra width from the table layout.
> 
> Sorry, is that a "you agree we should drop the changes to the stats central"?

Yes, I agree we should drop the changes to the stats central page. My apologies for the blather -- I was just trying to analyze why it didn't quite fit.

cpeel:
I've updated the commit to do better parameter validation and try to align the files closer together style-wise. This needs to be retested to make sure I didn't horribly break anything.

srjfoo (approved):
>This needs to be retested to make sure I didn't horribly break anything.

I don't see any brokenness -- everything seems to work as it should!

chrismiceli (approved):

70ray (approved):

PR #415, opened 2020-09-11
Give user a page they've never seen if possible

cpeel:
PROD and TEST have `$preceding_proofer_restriction` set to `not_immediately_preceding`, which means that when a proofreader hits "Start Proofreading" or "Save and Done and Proofread Next" we will only give them pages they have not proofed in the prior round. The value can also be set to `not_previously_proofed` which will only give them pages they have never proofed before.

[Task 1137](https://www.pgdp.net/c/tasks.php?action=show&task_id=1137) requests that even if the site is set to `not_immediately_preceding`, we first try to see if we can satisfy `not_previously_proofed` and give them a page they have never seen before. If we can't we fall back to `not_immediately_preceding`. That seems very reasonable and this MR implements that logic.

This is testable in the [get-best-page](https://www.pgdp.org/~cpeel/c.branch/get-best-page/) sandbox.

### DIFF
--- a/pinc/LPage.inc
+++ b/pinc/LPage.inc
@@ -15,15 +15,17 @@ include_once($relPath.'daily_page_limit.inc'); // get_dpl_count_for_user_in_roun
 
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 
-function get_available_page( $projectid, $proj_state, $pguser, &$err )
-// Returns an LPage, unless no page is available,
-// in which case returns NULL and sets $err.
+function get_available_page_array(
+    $projectid,
+    $proj_state,
+    $pguser,
+    $preceding_proofer_restriction,
+    &$err
+)
+// Returns the best page entry for the user given the specified preceding
+// project restriction. If successful, it returns [ $imagename, $state ],
+// on failure it returns NULL and sets $err;
 {
-    global $preceding_proofer_restriction;
-    // Make sure project is still in same state.
-    $err = project_continuity_test($projectid, $proj_state, TRUE);
-    if ( $err ) return NULL;
-
     $round = get_Round_for_project_state($proj_state);
 
     // Normally, pages are served in order of "page number"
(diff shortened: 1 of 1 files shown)


srjfoo (approved):
Looks good.

jmdyck (changes requested):
I have an alternative approach that I think it would be easier to understand. It would also be slightly more efficient, in that it avoids the possibility of needing 2 queries to get the next page. Specifically, for 'not_immediately_preceding', specify an `$order` that puts never-seen pages before previously-seen pages (with a secondary sort on `image`). So if there are any never-seen pages, the SELECT will return the next of those, and if there aren't, it'll return the next previously-seen page (if any).

The 'not_previously_proofed' case is already (basically) constructing the never-seen condition (though for WHERE rather than ORDER), so you could pull that code out either as a preliminary or a utility function.

(And you'd need to move the mentor-specific chunk later, so that its `$order` overrides the one from 'not_immediately_preceding'.)

(BTW, I had a long think about the race condition mentioned in the comment before `$dbQuery`, as I thought it might be entangled in how this task should be implemented, but eventually decided that it's probably orthogonal.)

cpeel:
> I have an alternative approach that I think it would be easier to understand.

Interesting, because to me that approach seems much more difficult to understand from a code perspective. Currently we have two modes that are easy to conceptualize what they do and we can reuse the modes in understandable ways. Making the already-complex `not_immediately_preceding` even more complex seems like a step in the wrong direction.

I see the value in having one query instead of two, but I don't think making the code more difficult to understand and maintain is worth the return.

70ray (commented):
I suppose, in principle, it would be better to do something about the race condition mentioned by jmdyck before doing this. But that is probably difficult and it does not seem to be a problem in practice but it could be borne in mind for the api version of this part of the system.

I approve of Casey's wish to use the simplest possible code.

Wouldn't the usual convention put "get_available_page" first?

cpeel:
> I suppose, in principle, it would be better to do something about the race condition mentioned by jmdyck before doing this. But that is probably difficult and it does not seem to be a problem in practice but it could be borne in mind for the api version of this part of the system.

The API to do page checkouts would use this same function. I feel pretty confident we can solve this with some sensible DB locking but don't feel like we need to figure this out right now.

> Wouldn't the usual convention put "get_available_page" first?

The code is all over the place on if dependent functions come first or second. I did it this way to minimize the diff, not because I think it's particularly better or worse order-wise.

70ray (approved):

PR #416, opened 2020-09-12
Position the text_preview div

70ray:
so hover messages scroll with it.
This addresses task 1929 [https://www.pgdp.net/c/tasks.php?action=show&task_id=1929](url)

### DIFF
--- a/styles/preview.css
+++ b/styles/preview.css
@@ -14,6 +14,7 @@
 #text_preview {
     padding: 0.2em;
     white-space: pre;
+    position: relative;
 }
 .no_display {
     display: none;


cpeel:
Just FYI: Your markdown for the Task in the MR description is wrong and doesn't work:
```
This addresses task 1929 [https://www.pgdp.net/c/tasks.php?action=show&task_id=1929](url)
```
should be:
```
This addresses [Task 1929](https://www.pgdp.net/c/tasks.php?action=show&task_id=1929)
```
eg:
> This addresses [task 1929](https://www.pgdp.net/c/tasks.php?action=show&task_id=1929)

note the URL being in the `()`s instead of in the `[]`s.

or even just use the raw URL without markup:
```
This addresses task 1929: https://www.pgdp.net/c/tasks.php?action=show&task_id=1929
```
eg:
> This addresses task 1929: https://www.pgdp.net/c/tasks.php?action=show&task_id=1929

cpeel:
This works great for me on Chrome on Mac.

70ray:
Thanks for info. Casey. I just clicked the icon and pasted the url but that's claearly not how it works.

PR #418, opened 2020-09-12
Remove support for TEI

cpeel:
TEI is considered a dead format (according to [the wiki](https://www.pgdp.net/wiki/PGTEI) and [this forum post](https://www.pgdp.net/phpBB3/viewtopic.php?p=1144915#p1144915)) so stop generating the files and show already-generated files in the "other files" section of a project. [Task 1709](https://www.pgdp.net/c/tasks.php?action=show&task_id=1709)

This can be played with in the [remove-tei](https://www.pgdp.org/~cpeel/c.branch/remove-tei/) sandbox.


chrismiceli:
Should we update import_pg_catalog.php?

cpeel:
> Should we update import_pg_catalog.php?

The only entry I saw in there is a MIME type which seems fine to leave as-is (because it is still accurate).

faq/post_proof.php
```diff
@@ -82,7 +82,6 @@
                 <li><a href="#html2">HTML</a></li>
                 <li><a href="#lilypond">Lilypond</a></li>
                 <li><a href="#latex">LaTeX</a></li>
-                <li><a href="#tei">TEI</a></li>
```
srjfoo, line 85: This and other changes to `faq/post_proof.php` will need to be updated in the DPOD, as well, since this document is redirected to the wiki for the production site.

srjfoo (approved):
Verified that only the `.txt` and its zipped version are now generated.

lhamilton1:
I approve

PR #419, opened 2020-09-13
Handle queue selectors containing '%'s

cpeel:
Queue selectors can contain '%'s which means we can't use sprintf() to build the full query. Instead, we build it piecemeal to ensure other arguments are correctly escaped.

This is a regression from 5a21c6a8a.

I found and confirmed this is fixed on TEST.

### DIFF
--- a/tools/project_manager/autorelease.inc
+++ b/tools/project_manager/autorelease.inc
@@ -228,14 +228,21 @@ function maybe_release_projects(
     // Projects that are in this round's waiting state,
     // don't have a hold in that state,
     // and, if $qd is set, satisfy its project selector.
-    $sql = sprintf("
+
+    // $cooked_project_selector can contain '%'s so we need to build the
+    // escaped state selector separately.
+    $escaped_state = sprintf(
+        "state = '%s'",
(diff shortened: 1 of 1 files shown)


chrismiceli (approved):

chrismiceli:
This is the kind of thing that was worrying me and why I was using %s for any placeholder, even column names. Is there a better way to handle this? Perhaps another encoding function that encodes % for sprintf uses?

cpeel:
> This is the kind of thing that was worrying me and why I was using %s for any placeholder, even column names.

And that's a fair point (although column names can't have %s in them) but doing so makes it even less readable than it is already. It's a hard balance.

> Is there a better way to handle this? Perhaps another encoding function that encodes % for sprintf uses?

The "right" way to deal with this is to use prepared statements so that we don't need to escape anything, but that gets particularly challenging with our dynamically-built queries.

cpeel:
The original designers were not DB designers and the entire site is built around very non-standard SQL practices (storing queue selectors to be injected into SQL strings, each project having it's own table, each project table having columns that encode specific rounds in them) all of which make  things much more difficult than they need to be. But here we are.

PR #420, opened 2020-09-13
Pull page interface styling out into theme.less

cpeel:
Pull page interface styling out of `layout.less` and into `theme.less`. This now centralizes all of the color for the various interfaces that deal with pages (proofreading, WordCheck, image display, format preview) into a single location. This is to enable some theming work that @srjfoo is working on.

This standardizes some of the UI style which makes for some small changes compared to what was there before:
* Buttons in control panes / control panels now all look the same. They have a small drop-shadow to help convey that they are a button. This includes the buttons in WordCheck.
* The background color of the image pane in the standard interface has changed to match the image pane color in the enhanced interface.
* Slight color changes to the control panels of Format Preview.

To review it's best to largely ignore the generated .css and focus on reviewing the other code changes.

This can be played with in the [standardize-proofing-theme](https://www.pgdp.org/~cpeel/c.branch/standardize-proofing-theme/) sandbox.

This does introduce some CSS classes and less variables that use some new nomenclature. Please see, review, and provide feedback on [DP Code - Page Interface Nomenclature](https://www.pgdp.net/wiki/DP_Code_-_Page_Interface_Nomenclature).

### DIFF
--- a/scripts/page_browse.js
+++ b/scripts/page_browse.js
@@ -193,7 +193,7 @@ var imageControl = function() {
 
 // Construct the font-face, font-size and wrap controls
 var textControl = function() {
-    let textArea = $("<textarea>", {class: "full-text"});
+    let textArea = $("<textarea>", {class: "text-pane"});
     textArea.val(pageBrowserData.text);
     let fontFaceSelector = document.createElement("select");
     fontFaceSelector.title = proofIntData.strings.changeFontFace;
--- a/styles/layout.css
+++ b/styles/layout.css
@@ -742,7 +742,6 @@ table.dirlist caption {
 #standard_interface {
   padding: 0;
   margin: 0;
-  background-color: #CDC0B0;
 }
 #standard_interface #imagedisplay {
   text-align: center;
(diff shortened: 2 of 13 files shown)


srjfoo:
> Please see, review, and provide feedback on [DP Code - Page Interface Nomenclature](https://www.pgdp.net/wiki/DP_Code_-_Page_Interface_Nomenclature).

The only thing that I initially found confusing on that page was using "Page Explorer" as a name to describe the view-only interface. It's too close for my liking to "Windows Explorer". I'd prefer something somewhat more generic ("Page Browser" would be okay, though that, also, may be confusing to some). I'll admit, though, that I may be unreasonably prejudiced -- I'll get used to whatever is decided on. :D

srjfoo (approved):
I doubt that I've tested everything possible, but I've looked at both proofreading interfaces, `view_page_text_image.php`, `displayimage.php`, WordCheck, codepoint validation, and Preview, and I'm fine with the changes (some of them will undoubtedly be noticed, though).

cpeel:
> > Please see, review, and provide feedback on [DP Code - Page Interface Nomenclature](https://www.pgdp.net/wiki/DP_Code_-_Page_Interface_Nomenclature).
> 
> The only thing that I initially found confusing on that page was using "Page Explorer" as a name to describe the view-only interface. It's too close for my liking to "Windows Explorer". I'd prefer something somewhat more generic ("Page Browser" would be okay, though that, also, may be confusing to some). I'll admit, though, that I may be unreasonably prejudiced -- I'll get used to whatever is decided on. :D

After thinking about this some more, and seeing `scripts/page_browse.js`, I think "Page Browser" is a better name for this. Still looking for naming feedback from @70ray and @chrismiceli.

chrismiceli (approved):
Looks good to me code wise.

70ray (commented):
I noticed two things while playing with it. Not sure if they're related to these changes:

When toolbox buttons are pressed the text changes from black to white which is almost invisible against a light background.

Some of the buttons in the enhanced interface can be dragged. This isn't really a problem but sems a bit disconcerting.

Another thing peripherally related which I came across while working on the page browser is that the button for horizontal/vertical split (and presumably all the others in the enhanced interface) have the form background colour built-in to them. If they had a transparent background then it would be possible to change the form colour.

cpeel:
Thanks for looking at it!

> I noticed two things while playing with it. Not sure if they're related to these changes:
> 
> When toolbox buttons are pressed the text changes from black to white which is almost invisible against a light background.

I'm not able to reproduce this which makes me think it is probably browser specific. What browser are you using?
 
> Some of the buttons in the enhanced interface can be dragged. This isn't really a problem but sems a bit disconcerting.

In Chrome I'm able to click and drag the images in the Enhanced interface but they don't go anywhere or do anything. It isn't any different to me than clicking and dragging any image on any web page however. What behavior are you seeing exactly?

70ray:
> > I noticed two things while playing with it. Not sure if they're related to these changes:
> > When toolbox buttons are pressed the text changes from black to white which is almost invisible against a light background.
> 
> I'm not able to reproduce this which makes me think it is probably browser specific. What browser are you using?

I only see it on Firefox in Linux (unrelated to this branch). Looks like it's a Firefox bug.
> 
> > Some of the buttons in the enhanced interface can be dragged. This isn't really a problem but sems a bit disconcerting.
> 
> In Chrome I'm able to click and drag the images in the Enhanced interface but they don't go anywhere or do anything. It isn't any different to me than clicking and dragging any image on any web page however. What behavior are you seeing exactly?
Playing with it some more, they can be dragged into the text area where they paste some text. (This is also unrelated to this branch). I'll make a new task for fixing it.

srjfoo:
> Thanks for looking at it!
> 
> > I noticed two things while playing with it. Not sure if they're related to these changes:
> > When toolbox buttons are pressed the text changes from black to white which is almost invisible against a light background.
> 
> I'm not able to reproduce this which makes me think it is probably browser specific. What browser are you using?

I didn't see it in either Chrome or Firefox, but did see it in Safari, so, yes, browser-specific. It's also not a change from the old behavior -- I tried it on the main test site, and got the same behavior. I'm not terribly fussed, because it's high contrast in the original button that you need to see when choosing it. Once you press it, it's less important to be able to read it, and once you let it back up again, it's back to normal.

(Added upon seeing Ray's comment above (he beat me to it) -- as I've noted, I don't see it in FF and Chrome, but do in Safari. So it's not just specific to FF.)


> > Some of the buttons in the enhanced interface can be dragged. This isn't really a problem but sems a bit disconcerting.
> 
> In Chrome I'm able to click and drag the images in the Enhanced interface but they don't go anywhere or do anything. It isn't any different to me than clicking and dragging any image on any web page however. What behavior are you seeing exactly?

I've tried this now in Safari, Firefox and Chrome. In Chrome, there's no effect other than being able to see the URL in the status bar for the browser window. In Firefox and Safari, if you drag and drop it over the text box, it inserts the URL into the text.

This, also, is not new behavior. And it's only for some of the buttons.

cpeel:
I've rebased this branch after merging https://github.com/DistributedProofreaders/dproofreaders/pull/421. I've also updated the wording to use "page-browser" instead of "page-explorer".

PR #421, opened 2020-09-16
One page browser to rule them all

cpeel:
This consolidates the 3 page browser interfaces into one set of files (one JS and one PHP) instead of three sets. If I have done my job correctly, they will look and act identically to how they did before -- the goal was zero user-visible changes.

The mentors interface has two features that were, and remain, mentors-only:
* the "ruler" div
* the textarea being editable

If a user is a mentor in any round, these will still be true for them in the consolidated interface.

After this goes in there will be a follow-up MR that moves this tool up a level and renames it -- probably `page_browser.php`.

This is testable in [one-page-browser](https://www.pgdp.org/~cpeel/c.branch/one-page-browser/).


tools/project_manager/display_image.js
```diff
@@ -31,70 +35,106 @@ $(function () {
...
+            if(mentorMode) {
+                let topTextDiv = textDiv;
+                let bottomTextDiv = $("<div>", {class: 'image-back'});
+                textDiv = $("<div>").append(topTextDiv, bottomTextDiv);
             }
+            stretchDiv.append(imageDiv, textDiv);
+            let theSplitter = viewSplitter(stretchDiv);
+            if(mentorMode) {
+                const subSplitID = "sub_split_percent";
+                let subSplitPercent = localStorage.getItem(subSplitID);
```
chrismiceli, line 120: This was already present, but since local storage only stores strings, here subSplitPercent can be a string or a number (line 123). While javascript allows this, it isn't ideal for JIT (not big deal here) or consistency. I'd use a `parseFloat` on the response from `localStorage.getItem`

chrismiceli (approved):

srjfoo (approved):
Looks good to me.

One question -- it's present in the code currently in use, as well as this version, and not a bug, just a matter of alignment: Is the indent on the first line of the project title deliberate: 
image
It might make more sense visually to make it a hanging indent, but that's definitely a matter of opinion.

I believe the changes will be user visible, if I'm understanding correctly. They won't be visible to users who are not mentors, but for mentors, PFs and squirrels, there will be a difference. There will no longer be `mentors/view_page_text_image.php`, so mentors will have to learn to access the pages from a project's page detail page. PFs and squirrels will have that interface by default.

If what you mean is that there will be no difference in how the page looks and operates, then I agree -- I didn't see any difference.

tools/project_manager/display_image.js
```diff
@@ -31,70 +35,106 @@ $(function () {
...
+    } else if(!pageBrowserData.currentPage) {
+        if(!simpleHeader) {
+            // show project name
+            fixHead.append($("<p>").append(pageBrowserData.heading));
+        }
+        fixHead.append(pageControlForm);
+
+        // show page selector
+        pageControlForm.append(hiddenProject(), hiddenMode(), hiddenSimpleHeader(), pageChanger(pageControlForm));
+        if(mode !== "image") {
```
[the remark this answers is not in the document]
cpeel: I'm not opposed to this, but this isn't the MR to make that change as all the other parts of this code use the strings.

cpeel:
> One question -- it's present in the code currently in use, as well as this version, and not a bug, just a matter of alignment: Is the indent on the first line of the project title deliberate:
> image
> It might make more sense visually to make it a hanging indent, but that's definitely a matter of opinion.

I don't know, but I'm not going to change it.

> I believe the changes will be user visible, if I'm understanding correctly. They won't be visible to users who are not mentors, but for mentors, PFs and squirrels, there will be a difference. There will no longer be `mentors/view_page_text_image.php`, so mentors will have to learn to access the pages from a project's page detail page. PFs and squirrels will have that interface by default.

On PROD I will put in a "this page has moved" page so people will update their bookmarks. So yes, they will notice and will need to update their bookmarks. As you surmised, I was mostly focused on keeping the look-and-feel identical.

tools/project_manager/display_image.js
cpeel: Good call -- I think that's clearer.

70ray (approved):
Looks good to me.
Although this is unchanged, perhaps put 1 or 2px bottom margin on the select contols so they are not hard up against the bottom of the control pane.

PR #422, opened 2020-09-17
Fix PHP warnings for merged projects

cpeel:
Don't output pages available or percent done for merged projects. This fixes the sprintf() warning from trying to force '' into a number.

@srjfoo found this bug on TEST. We don't show warnings on PROD so this isn't visible there.

Testable in the [fix-php-warning](https://www.pgdp.org/~cpeel/c.branch/fix-php-warning/) sandbox.


srjfoo (approved):
looks good to me

chrismiceli (approved):

PR #425, opened 2020-09-21
Improve special day color selection performance

cpeel:
Currently, for *every row* in a project listing on the round pages, we do a query to the database, load up the *entire special days table*, and return just a single color which may or may not be from the database query. Why? I have no idea.

This change does one query and stores the special days table into a static array that is populated just once on a page load. It also abstracts out the "common" special days into a single location rather than being spread across two places. This is an all-backend change and there are no user-visible differences.

Available in the [special-colors-perf](https://www.pgdp.org/~cpeel/c.branch/special-colors-perf/) sandbox.

### DIFF
--- a/pinc/special_colors.inc
+++ b/pinc/special_colors.inc
@@ -1,6 +1,45 @@
 <?php
 include_once($relPath.'misc.inc');
 
+// return a set of hard-coded special days keyed by spec_code
+function load_common_special_days()
+{
+    return [
+        "_BIRTHDAY_RECENT" => [
+            'color' => "CCFFFF",
+            'display_name' => "Authors with recent birthdays",
+        ],
+        "_BIRTHDAY_TODAY" => [
+            'color' => "33CCFF",
+            'display_name' => "Authors with birthdays today",
+        ],
+        "_OTHER_SPECIAL" => [
+            'color' => "FFFF66",
+            'display_name' => "Other Special",
+        ],
+    ];
+}
+
+function load_special_days()
+{
+    $specials_array = load_common_special_days();
+
+    $sql = "
+        SELECT spec_code, color, display_name
+        FROM special_days
+        ORDER BY display_name
+    ";
+    $specs_result = DPDatabase::query($sql);
+
+    while ($row = mysqli_fetch_assoc($specs_result)) {
+        $specials_array[$row['spec_code']] = [
+            "color" => $row['color'],
+            "display_name" => $row['display_name'],
+        ];
+    }
+
+    return $specials_array;
+}
 
 // Returns the special color associated with this
 // project, or null if no such color is specified.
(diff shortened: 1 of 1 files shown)


pinc/special_colors.inc
```diff
@@ -78,30 +106,19 @@ function echo_special_legend( $projects_where_clause)
...
-        }
+        if(isset($specials_array[$cs_row['spec']]))
+            $day_array[] = $specials_array[$cs_row['spec']];
     }
-    $common_array[] = array('color' => "CCFFFF", 'display_name' => "Authors with recent birthdays");
-    $common_array[] = array('color' => "33CCFF", 'display_name' => "Authors with birthdays today");
-    $common_array[] = array('color' => "FFFF66", 'display_name' => "Other Special");
+    // sort day_array by display_name
+    array_multisort(array_column($day_array, 'display_name'), SORT_ASC, $day_array);
+    // now append the hard-coded special day names
```
cpeel, line 118: This retains the current behavior of keeping the "common" special days at the end of the legend. To me it makes more sense to have them be alpha-sorted with the rest of them and that change would be easy to make.

pinc/special_colors.inc
```diff
@@ -1,63 +1,91 @@
...
-    $bgcolor = null;
-
+// that the key 'special_code' exists.
+function get_special_color_for_project($book)
+{
     $special_code = $book['special_code'];
 
-    // first three are (for now) special cases, the rest will be data-driven
-    // from the SPECIAL_DAYS table
+    static $specials_array = [];
```
chrismiceli, line 55: would it make sense to set this outside the function, that way we could avoid setting up a similar variable on line 109?

chrismiceli (approved):

pinc/special_colors.inc
cpeel: I don't want to make it a global variable (global variables = evil).

I would have liked to pull the `load_common_special_days()` array directly into `load_special_days()` and be done with it, but the logic in `get_special_color_for_project()` requires we keep them separate since we can't rely solely on the project's `special_code` value to determine which special day it goes with. ☹️

srjfoo (approved):
I didn't see any visible differences, which, I assume, is what's intended!

pinc/special_colors.inc
cpeel: I'm not even sure if Other Special is used (or should be) but we don't have a foreign key on the tables to ensures that all projects map to existing values in the `special_codes` table. I'm just going to leave it all as-is for now I think.

srjfoo: Sounds good.

I have seen Other Special be used. It's to fill in the gap left between the defined special days and the PM-definable birthdays. It, like the PM-definable birthday, is easily abused.

PR #426, opened 2020-09-22
Rename displayimage page to page_browser

cpeel:
`displayimage.php` is much more than just an image displayer. Rename it to reflect its new capabilities and move it a level up into the `tools/` directory since everyone has access to it, not just PMs.

This is testable in the [rename-page-browser](https://www.pgdp.org/~cpeel/c.branch/rename-page-browser/) sandbox. This will be seamless for users except for the new location.

This sandbox has redirects for the older pages to ease the transition for people with bookmarks. These will be going onto PROD as part of our local hacks as well:
* https://www.pgdp.org/~cpeel/c.branch/rename-page-browser/tools/project_manager/displayimage.php
* https://www.pgdp.org/~cpeel/c.branch/rename-page-browser/tools/mentors/view_page_text_image.php
* https://www.pgdp.org/~cpeel/c.branch/rename-page-browser/tools/display_image.php


srjfoo (approved):

chrismiceli (approved):

70ray (commented):
Would it be appropriate to use the simpleHeader option for some other cases where the old displayimage was used? Such as PQC, handle_bad_page etc.

srjfoo:
> Would it be appropriate to use the simpleHeader option for some other cases where the old displayimage was used? Such as PQC, handle_bad_page etc.

If I'm correct that the simpleHeader is what's used for WC context, and excludes the ability to choose to show the text and image both, then no. When fixing bad pages, it's quite useful to be able to display both text and image side-by-side. I have customarily brought up both image and text when fixing bad pages, but have had to do it separately.

cpeel:
> Would it be appropriate to use the simpleHeader option for some other cases where the old displayimage was used? Such as PQC, handle_bad_page etc.

I think no, at least not right now. Now that we have a very flexible single tool I'm hoping we can start using it in other places where we link directly to an image or page text (with some community discussion).

70ray (approved):

PR #427, opened 2020-09-23
Move color CSS from layout.less to theme.less

srjfoo:
This moves almost everything that can take a color out of `layout.less` and into `theme.less`. This includes border*, color and background-color. After experimenting with just breaking out the border color, I came to the conclusion that it was better to move the whole border property instead of breaking it up.  There are still a few border: none; and background-color: transparent; declarations that should probably be moved. The purpose is to make it easier to build additional site themes.

Instead of moving the stragglers, we might want to consider merging `layout.less` into `theme.less`: the problem with the current setup is that many structures have declaration blocks in both files, which can be confusing.

None of the class/mixin names I've choses are set in stone -- more logical suggestions are welcome.

If I've done what I should, there should be no user visible changes (except for a small one on the SR pool page). Compare [move-color-to-theme](https://www.pgdp.org/~srjfoo/c.branch/move-color-to-theme/) with the main test site.

### DIFF
--- a/styles/themes/theme.less
+++ b/styles/themes/theme.less
@@ -1,5 +1,61 @@
 @import (reference) "../global.less";
 
+/* ------------------------------------------------------------------------ */
+/* mixins with color */
+
+.solid-border(@color: black) {
+    border: thin solid @color;
+}
+
+.solid-border-bottom(@color: black) {
+    border-bottom: thin solid @color;
+}
+
+/* standalone classes that can be used as mixins, but that should not
+ * be given parameters, as they can't be used as classes in the code
+ * directly if they have parameters that the lessc compiler has to 
+ * sort out
+ */
+
+.default-border {
+    .solid-border;
+}
+
+.default-border-bottom {
+    .solid-border-bottom;
+}
+
+/* ------------------------------------------------------------------------ */
+/* Attention-getting structures */
+.error, .test_warning {
+    color: red;
+}
+
+.warning {
+    color: blue;
+}
+
+.highlight {
+    background-color: yellow;
+    color: black;
+}
+
+.button {
+    .sans-serif;
+}
+
+kbd {
+    color: #cc0000;
+}
+
+
+div.callout {
+    background-color: #EEE;
+    .solid-border;
+}
+
+/* ------------------------------------------------------------------------ */
+
 body {
     background-color: @page-background;
     color: @page-text;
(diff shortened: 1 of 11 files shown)


srjfoo:
Postscript -- I also removed some inline style color information that was in the code, if it was a simple replacement of a style with a class.

styles/themes/theme.less
```diff
@@ -101,26 +157,11 @@ h1, h2, h3, h4, h5, h6 {
...
-.button {
-    .sans-serif;
-}
 
-kbd {
-    color: #cc0000;
+/* ------------------------------------------------------------------------ */
+/* News items */
+div.newsitem {
+    box-shadow: -5px -5px 5px lightgrey; //srj: moved from layout
```
cpeel, line 164: Remove your note to self.

styles/themes/theme.less
```diff
@@ -285,9 +409,76 @@ table.register {
...
 
 table.snapshottable {
+    td, th {
+        .solid-border;
+    }
     th {
         background-color: @header-background;
     }
+    th.activity-icon-header {
+        background-color: @page-background;  // moved from layout
```
cpeel, line 419: Remove note to self.

cpeel (commented):
What you have looks good to me.

Waaay back when we moved the theming to CSS we only pulled things directly related to the different themes into `theme.less`. Now I think that we would be better served by combining `layout.less` and `theme.less` into one file -- probably `layout.less` -- and having the theme files just import it.

We should also probably `@import global.less` and `statsbar.less` into the theme files as well. This way we would have just one CSS file for the browser to import (and just one for less to generate and to check into source control) rather than the 4 we do now. I'm interested in the other devs' thoughts on this.

I think I would like to see the refactoring be in a separate MR than this one which is really focused on abstracting the color into mixins. (Yes, that will mean we then just move things back into `layout.less` but I think that's ok.)

srjfoo:
> I think I would like to see the refactoring be in a separate MR than this one which is really focused on abstracting the color into mixins. (Yes, that will mean we then just move things back into `layout.less` but I think that's ok.)

I agree. My time wasn't wasted. I learned quite a bit, and bouncing back and forth between the two was what made me wonder if it would be a good idea to combine them in the first place.

styles/themes/theme.less
srjfoo: Done. (Darn, I thought I'd gotten them all!)

styles/themes/theme.less
srjfoo: And done.

70ray (approved):

PR #428, opened 2020-09-24
Replace stage access and enligibility icons

cpeel:
To better support themes with non-white backgrounds, replace the stage access icons and stage eligibility icons with text.

Accessible in the [remove-stage-access-icons](https://www.pgdp.org/~cpeel/c.branch/remove-stage-access-icons) sandbox.


activity_hub.php
```diff
@@ -265,33 +265,29 @@ function summarize_stage($stage, $desired_states, $show_filtered_projects=FALSE,
 {
     global $pguser, $n_projects_in_state_, $n_projects_transitioned_to_state_;
 
-    // Get the stage identifier.
-    $stage_icon_url = get_dyn_image_url_for_file("stage_icons/{$stage->id}");
```
chrismiceli, line 269: side note, are code_images aren't in revision control it seems. Is that intentional?

chrismiceli (approved):

styles/themes/theme.less
```diff
@@ -290,6 +290,33 @@ table.snapshottable {
     }
 }
 
+.stage-access(@color: black) {
+    display: inline-block;
+    min-width: 1.5em;
+    .center-align;
+    font-size: 1.1em;
+    font-weight: bold;
```
srjfoo, line 298: I'd like to see all three of the access indicator characters slightly larger, if possible. The green checkmark isn't as heavy as the red x to begin with, so could stand to be somewhat larger, if possible. However, that would require passing a size parameter in the mixin, as well as the color. If it's a reasonable thing to do, I suggest 1.4em for the green checkmark, and 1.2 em for the other two. Otherwise, I'd suggest taking all of them to 1.2 instead of 1.1.

srjfoo (approved):

activity_hub.php
cpeel: Actually, yes. The graphics in `code_images` are a hacky way to maintain the text-in-graphics that have been on the site for a **long** time for English users while providing fallbacks to the actual text for non-English users and other consumers of the code. [I told you it was hacky.]

I hope/expect to be changing many of these to use text as we work to support the styling work @srjfoo is working on. As you can imagine that's likely going to be a contentious change.

styles/themes/theme.less
cpeel: I'm amenable to making them all 1.2em, but then I'd like help on figuring out how the @#$! to center the text contents within the cell. The Info icon already makes those cells taller than the others and the alignment is worse with the larger font size.`vertical-align: middle` isn't working and for the life of me I can't seem to figure it out.

srjfoo: Oh. Well, bah.

Looking at PROD, though, the current icons aren't vertically centered, either: 
image

Also, comparing it against the narrower version for narrow window/mobile devices, it looks like you're not slimming that column down to just the round ID? Was that deliberate?

Just looking at the code (and remember, my HTML is rusty), `$access_icon` is inside a `<div>` element, and the rest of the text in that table cell is outside that. (And yes, I realize that the div is `inline-block`. Instead of `vertical-align: middle`, have you tried shifting the baseline by a specific amount?

cpeel: > Looking at PROD, though, the current icons aren't vertically centered, either:
> image

It's not the icon centering that I'm worried about -- it's the text.

> Also, comparing it against the narrower version for narrow window/mobile devices, it looks like you're not slimming that column down to just the round ID? Was that deliberate?

It was that time, yes.

activity_hub.php
```diff
@@ -380,19 +376,17 @@ function summarize_stage($stage, $desired_states, $show_filtered_projects=FALSE,
 
     // Every row gets a label, name, and access information.
     echo "<td class='stage-column' $span_rows>";
-    echo "<div class='stage-icon'>$stage_id_bit</div>";
 
     // Output the access status icon. If the user does not yet have access
     // make the image a link to the access requirements.
-    echo "<div class='stage-access'>";
+    echo "<div class='stage-access $access_class'>";
```
70ray, line 382: Should there be a title with the access text in here?

cpeel: Yes indeed - thank you for catching that!

styles/themes/theme.less
cpeel: I've added a WIP (work-in-progress) commit to this MR to increase the font size and updated the sandbox. To best highlight the weird not-vertically-centered stuff look with an account that is eligible for an activity but not access to it (like one that has PP but not PPV).

srjfoo: > I'm amenable to making them all 1.2em, but then I'd like help on figuring out how the @#$! to center the text contents within the cell. The Info icon already makes those cells taller than the others and the alignment is worse with the larger font size.`vertical-align: middle` isn't working and for the life of me I can't seem to figure it out.

Okay, with the increase in size to 1.2em, what I'm seeing is that, when the icon overlaps with the descriptive text, it causes an increase in line height for that top line, pushing the second line down:
image

But if it doesn't overlap, there's no change:
image

And if there's only one line, it's aligned to the middle based on the tallest line. Which is the icon?
image

Short of putting the icon in a separate table cell, is there a way to cause it to center on the normal text, and not have the icon cause perturbations?

So yeah, I see why you sized it at 1.1 initially. I'm willing to go back to that.

srjfoo: > I've added a WIP (work-in-progress) commit to this MR to increase the font size and updated the sandbox. To best highlight the weird not-vertically-centered stuff look with an account that is eligible for an activity but not access to it (like one that has PP but not PPV).

Yes, that makes it even more obvious, though I did finally figure out what you meant, once you increased the size of the icon. Short of disassociating them from each other completely, I'm not sure I see a way to force the text to be centered vertically, independently of the icon.

cpeel: I moved it back down to 1.1em for now.

cpeel:
@srjfoo - I fixed the issues you identified with the mobile layout too.

srjfoo:
> @srjfoo - I fixed the issues you identified with the mobile layout too.

Looks good!

70ray (approved):

PR #430, opened 2020-09-25
Use a unicode star instead of graphics

cpeel:
Instead of graphics (which have a white background) use a Unicode star styled with CSS for the stars on the homepage. Also add these to the list_etexts page instead of changing the header color.

Available in the [star-icons](https://www.pgdp.org/~cpeel/c.branch/star-icons/) sandbox.


srjfoo (approved):

pinc/showstartexts.inc
```diff
@@ -27,8 +27,9 @@ function showstartexts($etext_limit,$type) {
     echo "<div class='star-text-summary'>";
 
     echo "<div class='star-header'>";
-    // Display star graphic
-    echo "<img src='$code_url/graphics/{$type}_star.jpg' height='38' width='40' alt='$type star'><br>\n";
+
+    // Display star
+    echo "<div class='star-$type'>★</div>";
```
chrismiceli, line 32: We're losing some accessibility without the alt text. A styled star may not read well with a screen reader. Perhaps add aria img role with aria label
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Role_Img

70ray: Or there could be a title with text as alt was. But it should be translated.

styles/themes/classic_grey.css
```diff
@@ -750,6 +750,26 @@ table.snapshottable th {
...
+/* Star texts */
+.star-gold {
+  font-size: 3em;
+  color: #FFD700;
+  text-align: center;
+  text-shadow: 3px 2px 5px black;
+}
+.star-silver {
+  font-size: 3em;
+  color: gray;
```
70ray, line 763: I think the lighter 'silver' colour looks better.

pinc/showstartexts.inc
srjfoo: In this particular case, would that be value-added?

Mind you, I'm not against accessibility accommodations, but the stars are essentially visual attention-getters. No useful information is missed by leaving out alt text that says "Gold star", "Silver star", or "Bronze star".

chrismiceli: Take a look at this screenshot simulating color blindness:
image
We do explain the sections underneath the star, but if they were using a screen reader, it may help to know what star they are on before having the screen reader read the text below it. Now i know the site probably fails in a million ways accessibility related, but we had alt text before so in this case, we are making it worse.

srjfoo: Fair point. And I was thinking primarily of the linked pages, not at the blurbs down at the bottom of the home page. Definitely more useful on the home page, whereas the stars are being added to the linked pages.

cpeel: All good points - thank you for catching this, @chrismiceli.

I've added an `aria-label` to the divs.

styles/themes/classic_grey.css
cpeel: I think so too and have changed it.

pinc/showstartexts.inc
chrismiceli: Thanks

chrismiceli (approved):

PR #434, opened 2020-09-25
Separate round images from their titles

cpeel:
Some history: before the site was translatable into something besides English, each round page had a graphic at the top for its header. The graphics (the same ones we have today) have English text in them that we can't easily translate. During the effort to provide a translatable site, the compromise was to pull the images out of the codebase but make them accessible via `code_images`. If these images-with-English are present we show them  if the user's locale was English and otherwise we output the translated round title.

The downside is that this presents two very different versions of the Activity Hub and round pages for English and non-English speakers. It also means that we have a weird thing where the Activity Hub header is in a font distinctly different from everything else and it has a white background which limits our ability to theme the site.

This MR allows us to have proper page text for the round headers and also display some graphics for each round. This way we get some header consistency and still have some more eye-catching images. The MR also introduces a way to have images for the front page without putting them in a news item which I think @lhamilton1 will appreciate.

This is a stepping stone to us pulling in an interesting font to use for the headers in a later MR.

Viewable in the [separate-page-header-images](https://www.pgdp.org/~cpeel/c.branch/separate-page-header-images/) sandbox.

### DIFF
--- a/pinc/theme.inc
+++ b/pinc/theme.inc
@@ -1039,18 +1039,23 @@ function get_dyn_image_url_for_file($filename)
     return NULL;
 }
 
-// Given an image, see if we have a version that is in the current
-// user's locale. If so, return the <img> tag with $text as alt text.
-// If not, return $text.
-function get_translated_graphic_or_text($directory, $image, $text)
+// Given a page basename, see if we have a header image for the page.
+// If so, return the <img>. If not, return "";
+function get_page_header_image($image)
 {
-    $locale = get_desired_language();
-    $dyn_url = get_dyn_image_url_for_file("$directory/$locale/$image");
-
+    # Try a locale-less image first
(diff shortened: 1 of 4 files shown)


pinc/theme.inc
```diff
@@ -1042,15 +1042,21 @@ function get_dyn_image_url_for_file($filename)
 // Given an image, see if we have a version that is in the current
```
chrismiceli, line 1042: is this comment still accurate?

chrismiceli (approved):

pinc/theme.inc
cpeel: Not at all. Fixed!

chrismiceli (approved):

srjfoo (approved):
Just verifying that adding the page title even on the pages that have a header image (P1-F2, PP, PPV and SR) was intentional. 

It all looks good to me.

PR #435, opened 2020-09-26
Allow news entries to have a header and a type

cpeel:
Currently news items are a single blob of text. Admins have gotten in the habit of adding formatted header-lines (larger and with color) over the years. This MR pulls the header out into a proper field and allows the news item to have a type that influences the styling. This enables the header to be styled with the theme rather than hard-coded.

The header is optional and if not present won't show up at all so this is backwards compatible with all existing news items.

The item types and colors are mostly adhoc and I'm looking to @lhamilton1, @wf49670, and @srjfoo for recommendations on what these should be.

This also converts the news admin code to use the new DPDatabase functions. It's in a separate commit though so if it is preferable for it to be in a separate MR I can easily do so.

Viewable in the [news-headers](https://www.pgdp.org/~cpeel/c.branch/news-headers/) sandbox.

### DIFF
--- a/SETUP/db_schema.sql
+++ b/SETUP/db_schema.sql
@@ -176,6 +176,8 @@ CREATE TABLE `news_items` (
   `ordering` smallint(6) NOT NULL default '0',
   `content` text NOT NULL,
   `locale` varchar(8) NOT NULL DEFAULT '',
+  `header` varchar(256) NOT NULL DEFAULT '',
+  `item_type` varchar(16) NOT NULL DEFAULT '',
   PRIMARY KEY  (`id`),
   KEY `pageid_locale` (`news_page_id`,`locale`)
 );
--- a/tools/site_admin/sitenews.php
+++ b/tools/site_admin/sitenews.php
@@ -111,51 +111,58 @@ function handle_any_requested_db_updates($news_page_id, $action, $item_id, $cont
         case 'add':
             // Save a new site news item
             $content = strip_tags($content, $allowed_tags);
-            $date_posted = time();
-            $insert_news = mysqli_query(DPDatabase::get_connection(), sprintf("
+            $sql = sprintf("
                 INSERT INTO news_items
                 SET
                     id           = NULL,
-                    news_page_id = '$news_page_id',
+                    news_page_id = '%s',
                     status       = '%s',
-                    date_posted  = '$date_posted',
+                    date_posted  = %d,
                     content      = '%s',
                     locale       = '%s'
-            ", mysqli_real_escape_string(DPDatabase::get_connection(), $item_status),
-               mysqli_real_escape_string(DPDatabase::get_connection(), $content),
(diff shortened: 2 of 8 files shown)


cpeel:
Added 2 more levels of "announcement" type and adjusted the colors per feedback from @lhamilton1

srjfoo (approved):
Looks good, works well.

tools/site_admin/sitenews.php
```diff
@@ -173,17 +194,29 @@ function handle_any_requested_db_updates($news_page_id, $action, $item_id, $cont
...
-            $result = mysqli_query(DPDatabase::get_connection(), sprintf("
+            $sql = sprintf("
                 UPDATE news_items
-                SET content='%s', locale='%s', status='%s'
-                WHERE id=$item_id
-            ", mysqli_real_escape_string(DPDatabase::get_connection(), $content),
-               mysqli_real_escape_string(DPDatabase::get_connection(), $locale),
-               mysqli_real_escape_string(DPDatabase::get_connection(), $item_status)));
-            $result = mysqli_query(DPDatabase::get_connection(), "SELECT status FROM news_items WHERE id=$item_id");
+                SET header = '%s', content = '%s', locale = '%s', status = '%s', item_type = '%s'
```
chrismiceli, line 199: did we want to do the LEFT stuff here?

tools/site_admin/sitenews.php
```diff
@@ -330,8 +383,9 @@ function show_all_news_items_for_page( $news_page_id )
             }
             echo "</td>";
             echo "<td class='items default-border-bottom'>";
+            echo "<div class='news-header news-{$news_item['item_type']}'>" . html_safe($news_item['header']) . "</div>";
```
chrismiceli, line 386: intentionally no HTML header, right?

chrismiceli (approved):

SETUP/db_schema.sql
```diff
@@ -176,6 +176,8 @@ CREATE TABLE `news_items` (
   `ordering` smallint(6) NOT NULL default '0',
   `content` text NOT NULL,
   `locale` varchar(8) NOT NULL DEFAULT '',
+  `header` varchar(256) NOT NULL DEFAULT '',
+  `item_type` varchar(16) NOT NULL DEFAULT '',
```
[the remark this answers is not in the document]
cpeel: SQL enums are a PITA and in this case having more flexibility to adjust them without a DB update is going to be better as well.

tools/site_admin/sitenews.php
cpeel: Yes we should, update incoming.

tools/site_admin/sitenews.php
cpeel: Correct - I want to discourage SAs from putting HTML into the header. (I'd love for them to not put it in the content, but that ship has sailed.)

PR #437, opened 2020-09-26
Consolidate styles into layout.less

cpeel:
Move contents of `theme.less` into `layout.less` and consolidate all generated CSS into the theme CSS thereby reducing the number of CSS files the browser needs to download. This also reduces the number of generated CSS files we need to check in. It doesn't measurably reduce the total CSS download size, however.

This also separates out the page interface styling into a new `page_inteface.less` file. This will enable page interface-specific theming separate from site theming which we know we will need to do at some point for the new single proofreading interface.

Proof that I didn't break anything is via [mega-less-file](https://www.pgdp.org/~cpeel/c.branch/mega-less-file) sandbox.


styles/page_interfaces.less
```diff
@@ -0,0 +1,374 @@
...
+        padding: 0 1px 0 0;
+    }
+
+    #hide-picker {
+        box-shadow: -1px -1px grey;
+    }
+
+    .invisible {
+        visibility: hidden;
+    }
```
srjfoo, line 282: Would this be good as a global mixin?

srjfoo (commented):
The border of the text pane (actually, the button bar at the top) overlaps the image pane in the enhanced interface. Comparing against TEST, there isn't an overlap there. Other than that, while there may be small differences, I didn't notice anything.

cpeel:
> The border of the text pane (actually, the button bar at the top) overlaps the image pane in the enhanced interface.

Can you provide some screenshots? I'm not seeing a difference and I suspect I'm not looking in the right place.

styles/README.md
```diff
@@ -19,8 +19,8 @@ sudo npm install -g less
 CSS is broken down into three major sections:
 * `global.less` primarily includes mix-ins and is unlikely to change much
 * `layout.less` governs page layout details
-* `themes/theme.less` governs colors, fonts, and other things impacted by themes
-  See the [Themes README](themes/README.md) for more information.
+
```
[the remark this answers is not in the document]
cpeel: It should, thanks for catching that.

styles/page_interfaces.less
cpeel: Yup! Moved.

srjfoo:
In advance, apologies for the "length" of this comment -- but long, narrow screenshots take a lot of vertical space 😁 

> > The border of the text pane (actually, the button bar at the top) overlaps the image pane in the enhanced interface.
> 
> Can you provide some screenshots? I'm not seeing a difference and I suspect I'm not looking in the right place.

For posterity, 
image

However, I did some additional investigation. The effect is not as a result of this MR. It exists live on PROD, as well. If I have my proofing window more or less full screen-height, I don't see it: 
image

But if I shorten it, I do:
image

70ray (approved):

srjfoo (approved):

PR #438, opened 2020-09-30
Separate pagedetail data from presentation

cpeel:
Separate pulling the pagedetail data from the database and how it is presented to the user. This is key to making this data available via the API which is the motivation for this change.

Move `page_table.inc` to `pinc/` as the functions in page_table are used by more than PM pages.

This is purely backend work and there should be no changes to upstream callers or the UI.

Testable in the [abstract_pagedetails](https://www.pgdp.org/~cpeel/c.branch/abstract_pagedetails/) sandbox.

### DIFF
--- a/pinc/page_table.inc
+++ b/pinc/page_table.inc
@@ -8,6 +8,216 @@ include_once($relPath.'../tools/proofers/PPage.inc'); // url_for_pi_do_particula
 include_once($relPath.'misc.inc'); // surround_and_join, attr_safe
 include_once($relPath.'links.inc'); // private_message_link
 
+// Given a project, return the rounds that contain useful data
+function get_rounds_to_display($project)
+{
+    $project_round = get_Round_for_project_state($project->state);
+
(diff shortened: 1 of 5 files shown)


pinc/page_table.inc
```diff
@@ -8,6 +8,215 @@ include_once($relPath.'../tools/proofers/PPage.inc'); // url_for_pi_do_particula
...
+        }
+        else
+        {
+            $wordcheck_query = "
+                (
+                    SELECT count(*)
+                    FROM wordcheck_events
+                    WHERE projectid = '$project->projectid' AND
+                        image = {$project->projectid}.image AND
+                        username = $round->user_column_name AND
```
chrismiceli, line 104: do we want to use the sprintf and encoding here

chrismiceli (approved):

pinc/page_table.inc
cpeel: Yes & done.

chrismiceli (approved):

srjfoo (approved):

PR #439, opened 2020-10-02
Move quiz colors into layout.less

srjfoo:
We need to be able to theme the passed/not passed colors for the quizzes. This moves them out of `Quiz.inc` into `layout.less`. I also removed an unused class reference from `Quiz.inc` and added a mixin for `x-large` text to `global.less`.

Testable here: [quiz-color-to-layout](https://www.pgdp.org/~srjfoo/c.branch/quiz-color-to-layout/quiz/start.php).

### DIFF
--- a/pinc/Quiz.inc
+++ b/pinc/Quiz.inc
@@ -138,7 +138,7 @@ class Quiz
         if ($this->description != "" || count($stage_required_for) > 0)
         {
             echo "
-                <tr><td colspan='$n_columns' class='info'>";
+                <tr><td colspan='$n_columns'>";
             if (count($stage_required_for) > 0)
             {
                 echo "<p>" . _("Passing this quiz is required for:") . " ";
@@ -168,18 +168,27 @@ class Quiz
             {
                 $passed = user_has_passed_quiz_page($username, $quiz_page_id);
                 $text = $passed ? _("Passed") : _("Not passed");
-                $bgcolor = $passed ? '#ccffcc' : '#ffcccc';
-                echo "<td style='background-color: $bgcolor'>$text</td>";
+                $class = $passed ? 'quiz-passed' : 'quiz-not-passed';
+                echo "<td class='$class'>$text</td>";
+
                 $date = get_last_attempt_date_for_quiz_page($username, $quiz_page_id);
                 $text = ($date != 0) ? strftime("%d-%b-%y", $date) : _("Not attempted");
                 $max = $this->pass_requirements['maximum_age'];
                 $date_ok = ((time() - $date) < $max) || empty($max);
-                $bgcolor = $date_ok ? '#ccffcc' : '#ffcccc';
-                echo "<td style='background-color: $bgcolor'>$text</td>";
-                $text = ($passed && $date_ok) ?	"<span style='font-size: 1.5em' title='Quiz page passed!'>&check;</span>"
-                    : "<span style='font-size: 1.5em' title='Quiz page not passed'>&#x2717;</span>";
-                $bgcolor = ($passed && $date_ok) ? '#88ff88' : '#ff8888';
-                echo "<td style='background-color: $bgcolor'>$text</td>";
+                $class = $date_ok ? 'quiz-date-ok' : 'quiz-date-not-ok';
+                echo "<td class='$class'>$text</td>";
+
+                if($passed && $date_ok)
+                {
+                    $text = "<span class='x-large' title='" . attr_safe(_("Quiz page passed!")) . "'>✓</span>";
+                    $class = 'quiz-ok';
+                }
+                else
+                {
+                    $text = "<span class='x-large' title='" . attr_safe(_("Quiz page not passed!")) . "'>✗</span>";
+                    $class = 'quiz-not-ok';
+                }
+                echo "<td class='$class'>$text</td>";
             }
             echo "</tr>";
         }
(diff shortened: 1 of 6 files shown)


styles/global.less
```diff
@@ -40,6 +40,10 @@
     font-size: large;
 }
 
+.x-large {
```
cpeel, line 43: Consider moving this before the `.large` so they are ordered biggest to smallest in the file.

pinc/Quiz.inc
```diff
@@ -168,28 +168,30 @@ class Quiz
             {
                 $passed = user_has_passed_quiz_page($username, $quiz_page_id);
                 $text = $passed ? _("Passed") : _("Not passed");
-                $bgcolor = $passed ? '#ccffcc' : '#ffcccc';
-                echo "<td style='background-color: $bgcolor'>$text</td>";
+                $bgcolor = $passed ? 'pass' : 'no-pass';
```
cpeel, line 171: Rename `$bgcolor` to `$class` or someting similar since it no longer contains a color (here and below).

cpeel: Also consider prefixing the class names with `quiz-` so `quiz-pass` & `quiz-no-pass` (or `quiz-not-passed`?).

pinc/Quiz.inc
```diff
@@ -168,28 +168,30 @@ class Quiz
...
                 $text = ($date != 0) ? strftime("%d-%b-%y", $date) : _("Not attempted");
                 $max = $this->pass_requirements['maximum_age'];
                 $date_ok = ((time() - $date) < $max) || empty($max);
-                $bgcolor = $date_ok ? '#ccffcc' : '#ffcccc';
-                echo "<td style='background-color: $bgcolor'>$text</td>";
-                $text = ($passed && $date_ok) ?	"<span style='font-size: 1.5em' title='Quiz page passed!'>&check;</span>"
-                    : "<span style='font-size: 1.5em' title='Quiz page not passed'>&#x2717;</span>";
-                $bgcolor = ($passed && $date_ok) ? '#88ff88' : '#ff8888';
-                echo "<td style='background-color: $bgcolor'>$text</td>";
+                $bgcolor = $date_ok ? 'date-ok' : 'date-not-ok';
```
cpeel, line 178: Similar as the above, consider prefixing with `quiz-`: `quiz-date-ok` & `quiz-date-not-ok`.

pinc/Quiz.inc
```diff
@@ -168,28 +168,30 @@ class Quiz
...
+                echo "<td class='$bgcolor'>$text</td>";
+
+                $text = ($passed && $date_ok) ?	"<span class='x-large' title='Quiz page passed!'>&check;</span>"
```
cpeel, line 181: I realize this was like this already, but it's a bit unwieldy as a ternary (not to mention the title isn't translated -- wonder if that was my fault in an earlier commit!).

Perhaps:
```php
if($passed && $date_ok)
{
    $text = "<span class='x-large' title='" . attr_safe(_("Quiz page passed!")) . "'>&check;</span>";
    $class = 'quiz-ok';
}
else
{
    $text = "<span class='x-large' title='" . attr_safe(_("Quiz page not passed!")) . "'>&#x2717;</span>";
    $class = 'quiz-not-ok';
}
```

chrismiceli: agree, if we go this way though, can we put unicode characters directly instead of their encodings? so '✗' and '✓'.

srjfoo: > agree, if we go this way though, can we put unicode characters directly instead of their encodings? so '✗' and '✓'.

I have no foundation knowledge to know which is the better way to go -- however, after a bit of research, I see that the entities are used elsewhere in the code for WC purposes (pre-Unicode?).

styles/global.less
srjfoo: Whoops, I should have noticed that to begin with. Done

pinc/Quiz.inc
srjfoo: Done. Coming up with meaningful, relatively short class names can be a challenge, but you're right that it's better to be more detailed. I was also trying to keep from giving the color variables and the classes the same names, but ...

pinc/Quiz.inc
srjfoo: > I realize this was like this already, but it's a bit unwieldy as a ternary (not to mention the title isn't translated -- wonder if that was my fault in an earlier commit!).
> 
> Perhaps:
> 
> ```
> if($passed && $date_ok)
> {
>     $text = "<span class='x-large' title='" . attr_safe(_("Quiz page passed!")) . "'>&check;</span>";
>     $class = 'quiz-ok';
> }
> else
> {
>     $text = "<span class='x-large' title='" . attr_safe(_("Quiz page not passed!")) . "'>&#x2717;</span>";
>     $class = 'quiz-not-ok';
> }
> ```

Done.

I have not yet followed @chrismiceli's suggestion of using the actual characters for the '✗' and '✓', but I can do that easily enough if that's the way we want to go. If we do, we should probably also update it elsewhere in the code.

srjfoo: > agree, if we go this way though, can we put unicode characters directly instead of their encodings? so '✗' and '✓'.

And now done.

styles/layout.less
```diff
@@ -1044,6 +1044,40 @@ table.dirlist {
     }
 }
 
+/* ------------------------------------------------------------------------ */
+/* Quizzes */
+
+@quiz-passed: #ccffcc;
+@quiz-not-passed: #ffcccc;
+@quizdate-ok: #ccffcc;
```
cpeel, line 1052: Any particular reason why this isn't `@quiz-date-ok`?

chrismiceli (approved):

styles/layout.less
srjfoo: 😆 -- None except an ancient irrational attempt to minimize the number of characters and avoid duplicate names. In light of some of the earlier changes, I've updated those two, also.

PR #440, opened 2020-10-03
Properly escape <meta>-refreshed URLs

cpeel:
Properly escape redirect URLs in `metarefresh()` that use the `<meta>` tag. Also ensure we disallow redirects off-site when using the `<meta>` tag unless explicitly allowed. This addresses a XSS vulnerability.

This can be played with using the https://www.pgdp.org/~cpeel/c.branch/escape-metarefresh-urls/pinc/test.php page and passing in a URL, eg:
https://www.pgdp.org/~cpeel/c.branch/escape-metarefresh-urls/pinc/test.php?url=../project.php?id=projectID45c225f598e32

### DIFF
--- a/pinc/metarefresh.inc
+++ b/pinc/metarefresh.inc
@@ -59,12 +59,13 @@ function metarefresh($seconds, $url, $title="", $body="", $allow_external=FALSE)
     // If we can't do the redirection using HTTP headers (which is the common
     // case), fall back to using a <meta> tag.
 
-    // See if ambersands are present and if not escaped, escape them
-    if(strpos($url, '&') !== FALSE && strpos($url, '&amp;') === FALSE) {
-        $url = str_replace('&', '&amp;', $url);
+    // If ampersands are present and already escaped, unescape them so
+    // attr_safe() will re-escape them along with anything else
+    if(strpos($absolute_url, '&amp;') !== FALSE) {
+        $absolute_url = str_replace('&amp;', '&', $absolute_url);
     }
 
-    $meta_tag_refresh = "<meta http-equiv=\"refresh\" content=\"$sec ;URL=$url\">";
+    $meta_tag_refresh = "<meta http-equiv='refresh' content='$sec ;URL=" . attr_safe($absolute_url) . "'>";
 
     global $relPath;
     include_once($relPath."slim_header.inc");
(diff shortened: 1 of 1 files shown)


srjfoo (approved):
> This can be played with using the https://www.pgdp.org/~cpeel/c.branch/escape-metarefresh-urls/c/pinc/test.php page and passing in a URL, eg:
> https://www.pgdp.org/~cpeel/c.branch/escape-metarefresh-urls/pinc/test.php?url=../project.php?id=projectID45c225f598e32

The first of these links doesn't work because of the `/c` in the URL. When I removed that, I first got "Notice: Undefined index: url in /home/cpeel/public_html/c.branch/escape-metarefresh-urls/pinc/test.php on line 6".

Following the other link seems to work, though, and trying to generate the message just by carrying out an action that I would expect to generate it, I also don't see any difference in behavior from expected.

pinc/metarefresh.inc
```diff
@@ -59,12 +59,13 @@ function metarefresh($seconds, $url, $title="", $body="", $allow_external=FALSE)
     // If we can't do the redirection using HTTP headers (which is the common
     // case), fall back to using a <meta> tag.
 
-    // See if ambersands are present and if not escaped, escape them
-    if(strpos($url, '&') !== FALSE && strpos($url, '&amp;') === FALSE) {
-        $url = str_replace('&', '&amp;', $url);
+    // If ambersands are present and already escaped, unescape them so
```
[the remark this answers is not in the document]
cpeel: Fixed!

cpeel:
> The first of these links doesn't work because of the `/c` in the URL. When I removed that, I first got "Notice: Undefined index: url in /home/cpeel/public_html/c.branch/escape-metarefresh-urls/pinc/test.php on line 6".

Whoops - sorry about that! I fixed the MR description.

pinc/metarefresh.inc
```diff
@@ -59,12 +59,13 @@ function metarefresh($seconds, $url, $title="", $body="", $allow_external=FALSE)
     // If we can't do the redirection using HTTP headers (which is the common
     // case), fall back to using a <meta> tag.
 
-    // See if ambersands are present and if not escaped, escape them
-    if(strpos($url, '&') !== FALSE && strpos($url, '&amp;') === FALSE) {
-        $url = str_replace('&', '&amp;', $url);
+    // If ampersands are present and already escaped, unescape them so
+    // attr_safe() will re-escape them along with anything else
+    if(strpos($absolute_url, '&amp;') !== FALSE) {
+        $absolute_url = str_replace('&amp;', '&', $absolute_url);
```
chrismiceli, line 65: Should we find the callers to this method and try to remove the encoding?

chrismiceli (approved):

pinc/metarefresh.inc
cpeel: We could, but I'm not sure it's worth the effort.

70ray (approved):

PR #441, opened 2020-10-03
Use EB Garamond for page headers

cpeel:
Use EB Garamond for page headers, providing a bit more visual pizzaz to them.

This is viewable in the [eb-garamond-headers](https://www.pgdp.org/~cpeel/c.branch/eb-garamond-headers/) sandbox.

### DIFF
--- a/faq/faq_central.php
+++ b/faq/faq_central.php
@@ -220,7 +220,7 @@ class FAQ
                 echo "<td class='column'>";
                 continue;
             }
-            echo "\n<div class='faqheader'>$section->title</div>\n";
+            echo "\n<h2>" . html_safe($section->title) . "</h2>\n";
 
             $section->output();
         }
--- a/styles/global.less
+++ b/styles/global.less
@@ -190,3 +190,8 @@
     font-weight: normal;
     font-style: oblique;
 }
+
+
+/* EB Garamond
+   https://fonts.google.com/specimen/EB+Garamond */
+@import url('https://fonts.googleapis.com/css2?family=EB+Garamond:wght@700&display=swap');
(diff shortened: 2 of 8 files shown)


lhamilton1:
Looks good to me

srjfoo (approved):

faq/faq_central.php
```diff
@@ -220,7 +220,7 @@ function output()
                 echo "<td class='column'>";
                 continue;
             }
-            echo "\n<div class='faqheader'>$section->title</div>\n";
+            echo "\n<h2>$section->title</h2>\n";
```
chrismiceli, line 223: should we encode here?

styles/global.less
```diff
@@ -190,3 +190,8 @@
     font-weight: normal;
     font-style: oblique;
 }
+
+
+/* EB Garamond
+   https://fonts.google.com/specimen/EB+Garamond */
+@import url('https://fonts.googleapis.com/css2?family=EB+Garamond:wght@700&display=swap');
```
chrismiceli, line 197: I'm worried about external requests from css in case someone has their browser block 3rd party requests, especially from a google domain (to help prevent tracking). Does their license allow us to check it in like jquery?

cpeel: Even if it does allow it I don't want to check it in. If their browser blocks the load it will fall back to the browser default font. With jQuery, however, if their browser blocks it the site breaks, which is why I included it in the code.

In general I would prefer we embed less in the codebase, not more.

faq/faq_central.php
cpeel: Indeed! Fixed.

chrismiceli (approved):

styles/global.less
chrismiceli: That makes sense, just clarifying I suppose our philosophy here.

cpeel: Yup, totally fair question. 👍

PR #442, opened 2020-10-04
API updates: correct project fields; page details; more info in pages index

cpeel:
This updates the API with the following changes:
* corrects project fields so the OpenAPI docs match the fields returned for a project
* updates the `/pages` endpoint to include more information about the page instead of just a listing of names
* adds a `/pagedetails` endpoint that returns largely the same data as the UI page with the same name

Testable via:
```bash
curl -i -g -X GET "https://www.pgdp.org/~cpeel/c.branch/api-updates/api/index.php?url=v1/projects/projectID45c225f598e32" \
    -H "accept: application/json" \
    -H "X-API-KEY: review"

curl -i -g -X GET "https://www.pgdp.org/~cpeel/c.branch/api-updates/api/index.php?url=v1/projects/projectID45c225f598e32/pagedetails" \
    -H "accept: application/json" \
    -H "X-API-KEY: review"

curl -i -g -X GET "https://www.pgdp.org/~cpeel/c.branch/api-updates/api/index.php?url=v1/projects/projectID45c225f598e32/pages" \
    -H "accept: application/json" \
    -H "X-API-KEY: review"
```

The updates to the `/pages` and project endpoints break compatibility with what was out there before, but currently we have no consumers and the API is not included in any DP code release, so this seems entirely reasonable at this time.

### DIFF
--- a/api/dp-openapi.yaml
+++ b/api/dp-openapi.yaml
@@ -332,7 +332,7 @@ paths:
               schema:
                 type: array
                 items:
-                  type: string
+                  $ref: '#/components/schemas/project_page_index'
         '401':
           $ref: '#/components/responses/Unauthorized'
         '404':
@@ -496,6 +496,22 @@ components:
         includes a great deal of information, some fields of which are calculated
         and not directly modifyable. Not all fields are available to all users. Fields
         the user does not have access to will not be returned.
+
+    project_page_index:
+      required:
+      - image
+      - image_url
+      - image_size
+      type: object
+      properties:
+        image:
+          type: string
+        image_url:
+          type: string
+          format: url
+        image_size:
+          type: integer
+          description: Image size in bytes
 
     project_page:
       required:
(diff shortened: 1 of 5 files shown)


chrismiceli (approved):

srjfoo (approved):

api/dp-openapi.yaml
```diff
@@ -497,6 +513,69 @@ components:
         and not directly modifyable. Not all fields are available to all users. Fields
```
70ray, line 513: perhaps "modifiable"?

70ray (approved):

api/dp-openapi.yaml
cpeel: Indeed, thank you. @srjfoo found some other description updates that I've also incorporated.

bpfoley (commented):
SGTM!

PR #443, opened 2020-10-04
Style subscriptions and hold checkboxes

cpeel:
Pull the formatting for event subscriptions and project holds out of `project.php` and into classes so they can be styled. Very minimal visual changes, just prep work for what @srjfoo is working on.

Viewable in the [checkbox-selected](https://www.pgdp.org/~cpeel/c.branch/checkbox-selected/) sandbox.

### DIFF
--- a/project.php
+++ b/project.php
@@ -818,7 +818,7 @@ function echo_row_a( $left, $right, $escape_right=FALSE )
 function echo_row_b( $top, $bottom, $class='' )
 {
     echo "<tr>";
-    echo "<td colspan='5' class='$class' style='text-align: center;'>";
+    echo "<td colspan='5' class='$class center-align'>";
     echo "<span class='bold large'>$top</span>";
     if ($bottom)
     {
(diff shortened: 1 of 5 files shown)


project.php
```diff
@@ -1223,8 +1223,8 @@ function do_holds()
     echo "<table>\n";
     echo "<tr>\n";
     echo "<th></th>\n";
-    echo "<th style='padding: 0em 1em'>", _("hold in Waiting"), "</th>\n";
-    echo "<th style='padding: 0em 1em'>", _("hold in Available"), "</th>\n";
+    echo "<th>", _("hold in Waiting"), "</th>\n";
+    echo "<th>", _("hold in Available"), "</th>\n";
```
srjfoo, line 1227: After looking at this and comparing it to what Event Subscriptions already was, I think it will work fine without the padding. I would definitely expect some people to notice it, though, since the two table headers are crammed right up against each other, without even a punctuation mark.

srjfoo (approved):

project.php
cpeel: <nod> I added a second commit to put some spacing between the columns on both of these tables.

chrismiceli (approved):

PR #444, opened 2020-10-04
Explicitly set the standard interface colors

cpeel:
Rather than rely on the standard interface inheriting from the site style, explicitly set the colors. This is a stepping stone to before we have themes that apply just to the proofreading interface. This is in prep for theming work that @srjfoo is doing.

Note that one downside to this is that buttons in the standard proofreading interface are now styled rather than simply whatever the browser agent is using.

Testable in the [set-std-interface-colors](https://www.pgdp.org/~cpeel/c.branch/set-std-interface-colors/) sandbox.

### DIFF
--- a/styles/page_interfaces.less
+++ b/styles/page_interfaces.less
@@ -35,6 +35,7 @@
 
         input, select, button {
             background-color: @text-pane-background;
+            color: @text-pane-text;
             .solid-border;
         }
 
@@ -55,7 +56,7 @@
 /* ------------------------------------------------------------------------ */
 /* Standard Editing Interface */
 
-#standard_interface {
+#standard_interface_image {
     .no-padding;
     .no-margin;
     .page-interface;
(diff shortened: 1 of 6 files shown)


srjfoo (approved):
> Note that one downside to this is that buttons in the standard proofreading interface are now styled rather than simply whatever the browser agent is using.

They would have needed to be styled, anyway. If there's a particular look we want for the buttons, perhaps we should design the style we want?

chrismiceli (approved):

70ray:
The buttons look a bit angular now. A border-radius of 2 or 3px softens the appearance a bit.

srjfoo:
> The buttons look a bit angular now. A border-radius of 2 or 3px softens the appearance a bit.

That's what I assumed we'd do -- just haven't gotten around to it yet.

cpeel:
But if we do this, do we do this everywhere? The buttons in the toolbox don't have rounded corners (nor have they ever to my knowledge).

srjfoo:
> But if we do this, do we do this everywhere? The buttons in the toolbox don't have rounded corners (nor have they ever to my knowledge).

Also true.

PR #448, opened 2020-10-06
fix-bg1b-template -- mentor-requested updates

srjfoo:
Changes to the BEGIN template BG1b.txt.

Sandbox available: [fix-bg1b-template](https://www.pgdp.org/~srjfoo/c.branch/fix-bg1b-template/). See [this post](https://www.pgdp.net/phpBB3/viewtopic.php?p=1221643#p1221643) and following for context. There are two BEGIN projects in P1 in the sandbox for verification.


srjfoo:
These initial changes are the only requested changes, but it occurred to me to ask if the links ought to be updated to `https:` and to point directly to the referenced sections of the guidelines, instead continuing as redirects. `BG2r.txt` also still has an `http:` link.

cpeel:
Please update the links to be `https`. I would leave the redirects in because IIRC we redirect to the guidelines in the proofreader's language if that's available.

srjfoo:
> Please update the links to be `https`. I would leave the redirects in because IIRC we redirect to the guidelines in the proofreader's language if that's available.

Done. Also removed a username from the `BGr2.txt` template. I'd forgotten we hadn't done that yet.

chrismiceli (approved):

PR #450, opened 2020-10-09
Redirect to task after acting upon it

cpeel:
Redirect back to the task's page after acting on a task. This clears the POST data and allows the user to refresh the page without resubmitting the form. [Task 1839](https://www.pgdp.net/c/tasks.php?action=show&task_id=1839)

Testable in the [task-center-redirect](https://www.pgdp.org/~cpeel/c.branch/task-center-redirect/) sandbox.

### DIFF
--- a/tasks.php
+++ b/tasks.php
@@ -635,7 +635,6 @@ function handle_action_on_a_specified_task()
     $pre_task = mysqli_fetch_object($result);
     if (!$pre_task)
     {
-        TaskHeader("Task #$task_id does not exist");
         ShowError("Task #$task_id was not found!");
         return;
     }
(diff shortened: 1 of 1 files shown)


mlazaric (approved):
Looks great :+1: 

One thing I've noticed while testing this is that a couple actions still lead to a broken state (they send you to https://www.pgdp.org/~cpeel/c.branch/task-center-redirect/tasks.php with form data, so it either shows the task details of the task from the POST form data or just an error like "You must supply a valid related task id number." and refresh shows you the form resubmit alert). They're probably out of scope for this PR, but it might be a good idea to create a task for them, here are the ones I've found:

 - Trying to post an empty comment
 - Trying to add a non-existent related task (there's no task with task id 1 on both production and the test server, so it's a good way to test)
 - Trying to add a non-existent related forum thread (for example 99999)

cpeel:
> [H]ere are the ones I've found:
> 
> * Trying to post an empty comment
> * Trying to add a non-existent related task (there's no task with task id 1 on both production and the test server, so it's a good way to test)
> * Trying to add a non-existent related forum thread (for example 99999)

All true. I also don't know how to fix any of those error cases and still show the error message. Any ideas?

mlazaric:
That's a really good point. I can't think of a simple/straightforward way to do it and the way we handle empty comments seems to be mostly in line with the way the post/redirect/get pattern handles errors, so it might be good enough to just show the error message in the same way for a non-existent related task/forum thread?

cpeel:
I actually went the other route and made the 'user hit Add Comment but without a comment' error case do what the other error cases do now: print the error and tell the user to go back with a link. I think this will provide a better UX because that back button should still work (in most all cases) and the page they go back to should be a GET-requested page for the task itself.

cpeel:
After some testing I added another commit that improves the validation when adding new topics to a task.

srjfoo (approved):

PR #451, opened 2020-10-09
Standardize completed title listings

cpeel:
Standardize the completed title listings between the sidebar and the list_ebooks pages.

Kick the tires in the [book-listings](https://www.pgdp.org/~cpeel/c.branch/book-listings/default.php) sandbox.

### DIFF
--- a/pinc/list_projects.inc
+++ b/pinc/list_projects.inc
@@ -19,17 +19,14 @@ function list_projects_tersely( $where_condition, $order_clause, $limit_clause )
 
     while (list($title,$author,$language,$postednum) = mysqli_fetch_row($result))
     {
-        echo "<li><span class='title-listing'>";
+        echo "<li class='title-listing'>";
 
-        echo "<span class='book-title'>", '"', html_safe($title), '"', "</span>";
-        echo "&mdash;" . html_safe($author) . " ($language)";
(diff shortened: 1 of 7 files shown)


srjfoo (approved):

pinc/list_projects.inc
```diff
@@ -19,17 +19,14 @@ function list_projects_tersely( $where_condition, $order_clause, $limit_clause )
 
     while (list($title,$author,$language,$postednum) = mysqli_fetch_row($result))
     {
-        echo "<li><span class='title-listing'>";
+        echo "<li class='title-listing'>";
 
-        echo "<span class='book-title'>", '"', html_safe($title), '"', "</span>";
-        echo "&mdash;" . html_safe($author) . " ($language)";
+        echo "<span class='book-title'>" . html_safe($title) . "</span>";
+        echo ", " . html_safe($author) . " ($language)";
```
chrismiceli, line 25: Not that I anticipate it, but we may want to include html_safe on language if we don't validate it on insertion into db.

chrismiceli (approved):

pinc/list_projects.inc
cpeel: That's a good idea. It certainly can't hurt!

PR #452, opened 2020-10-10
padding for highlighted text removed

70ray:
This addresses task 1936 https://www.pgdp.net/c/tasks.php?action=show&task_id=1936.
The problem occurs because the text with padding is greater than the line height.

### DIFF
--- a/pinc/DifferenceEngineWrapper.inc
+++ b/pinc/DifferenceEngineWrapper.inc
@@ -181,6 +181,10 @@ function get_DifferenceEngine_css_data() {
     font-family: $font_family;
     $font_size
 }
+.diff-deletedline .diffchange,
+.diff-addedline .diffchange {
+    padding: 0;
+}
 ";
 }
 
--- a/pinc/DifferenceEngineWrapper.inc
+++ b/pinc/DifferenceEngineWrapper.inc
@@ -181,6 +181,7 @@ function get_DifferenceEngine_css_data() {
     font-family: $font_family;
     $font_size
 }
+/* Adjust padding to prevent descenders from being chopped off. Task 1936 */
 .diff-deletedline .diffchange,
 .diff-addedline .diffchange {
     padding: 0.1em 0;
--- a/pinc/DifferenceEngineWrapper.inc
+++ b/pinc/DifferenceEngineWrapper.inc
@@ -183,7 +183,7 @@ function get_DifferenceEngine_css_data() {
 }
 .diff-deletedline .diffchange,
 .diff-addedline .diffchange {
-    padding: 0;
+    padding-top: 0;
 }
 ";
 }
--- a/pinc/DifferenceEngineWrapper.inc
+++ b/pinc/DifferenceEngineWrapper.inc
@@ -183,7 +183,7 @@ function get_DifferenceEngine_css_data() {
 }
 .diff-deletedline .diffchange,
 .diff-addedline .diffchange {
-    padding-top: 0;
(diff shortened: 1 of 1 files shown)


cpeel:
@70ray - can you post a TEST URL of a repro of the bug so we can see before/after?

srjfoo:
> @70ray - can you post a TEST URL of a repro of the bug so we can see before/after?

I just happened to have had my tests handy. Note that I saw only a minimal chopping in Chrome, and none at all in FF. The screenshots below are from Chrome.

Before:
image

After (shows the gap after removing all padding):
image

This shows the boxes overlapping, if you look at the corners.
image


I did wonder whether all padding should be removed, but I also don't know what the value of the padding was in the first place.

cpeel:
Hmm... and I'm wondering if increasing line-height slightly would fix it.

srjfoo:
> Hmm... and I'm wondering if increasing line-height slightly would fix it.

I would think so.

70ray:
Looking at this again, with no padding, the tails of 'g' etc. are hard against the bottom of the coloured area but there is still a gap at the top of capitals and tall letters. Setting padding-top to 0 fixes the problem without introducing any white space between the coloured lines. (Firefox on Linux). Alternatively inceasing line-height from 1.6 to 1.7 fixes it.
On Chromium on Linux, no clipping can be seen even with the original padding of 0.25em.
I think setting padding-top to zero would be the best option although there could be some combination of font, browser and operating system where it doesn't work.

cpeel:
Let's go with the padding-top option then, thanks for exploring the options!

70ray:
I've put the padding-top option in the sandbox https://www.pgdp.org/~rp31/c.branch/diff_pad (using cpeel's new method).
The example given by jjz is worse than any I have seen so it might not fix it completely. If so we can adjust it some more.

srjfoo:
> I've put the padding-top option in the sandbox https://www.pgdp.org/~rp31/c.branch/diff_pad (using cpeel's new method).
> The example given by jjz is worse than any I have seen so it might not fix it completely. If so we can adjust it some more.

image

Deliberately pushing the limits, if there are tall diacriticals over upper case letters, those go all the way to the top, when `padding-top` is set to `0`. If you want *some* padding to show that the character is not being clipped, possibly a slight adjustment in the line hight is a better option?

cpeel:
FWIW, I checked the most recent MediaWiki styles and it doesn't look like they've changed the styling of this since 1.31.1, so no help on that front.

70ray:
> > I've put the padding-top option in the sandbox https://www.pgdp.org/~rp31/c.branch/diff_pad (using cpeel's new method).
> > The example given by jjz is worse than any I have seen so it might not fix it completely. If so we can adjust it some more.
> 
> image
> 
> Deliberately pushing the limits, if there are tall diacriticals over upper case letters, those go all the way to the top, when `padding-top` is set to `0`. If you want _some_ padding to show that the character is not being clipped, possibly a slight adjustment in the line hight is a better option?

Good point. I have also asked jjz to say what browser and OS she's using so we can check it out properly.

70ray:
Changed to 0.1em padding at top and bottom. This looks good on all the browsers jjz mentioned.

srjfoo (approved):
Looks clean to me on all three of FF, Chrome and Safari on macOS.

cpeel:
I added a comment to the CSS to denote why we're changing the padding so we don't undo it later inadvertantly.

PR #453, opened 2020-10-10
Improve button styling

cpeel:
This does three minor things:
1. Add styling for disabled buttons in WordCheck
2. Rounds button corners -- this is global for all page interface buttons, including the toolbar. If we think this is too dramatic I can lower it to 2px which is very subtle but still better than none, to @70ray's point.
3. Sets a non-white background color for buttons in the standard interface to make them blend in less. I used the button color that Chrome on Mac uses.

Testable in the [button-disabled](https://www.pgdp.org/~cpeel/c.branch/button-disabled/) sandbox.

### DIFF
--- a/styles/page_interfaces.less
+++ b/styles/page_interfaces.less
@@ -39,8 +39,13 @@
             .solid-border;
         }
 
+        input[type=submit]:disabled, input[type=button]:disabled, button:disabled {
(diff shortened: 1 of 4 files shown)


chrismiceli (approved):

srjfoo (approved):
I think the button radius is fine as it is.

PR #454, opened 2020-10-11
Update Task Center style to match site

cpeel:
This updates the Task Center styling to match that of the rest of the site. This involves quite a lot of changes since it was a beast unto itself.

Testable in the [style-task-center](https://www.pgdp.org/~cpeel/c.branch/style-task-center/) sandbox.

### DIFF
--- a/tasks.php
+++ b/tasks.php
@@ -340,7 +340,7 @@ function SearchParams_echo_controls()
 // initializing it with any (valid) value that the current request
 // has supplied for the parameter.
 {
-    global $SearchParams_choices;
+    global $SearchParams_choices, $tasks_url;
 
     if (isset($_REQUEST['search_text']) && !empty($_REQUEST['search_text'])) {
         $st = $_REQUEST['search_text'];
@@ -348,14 +348,23 @@ function SearchParams_echo_controls()
     }
     else $search_text = "";
 
+    echo "<form action='$tasks_url' method='get'>";
+    echo "<table class='themed'>\n";
+    echo "<tr><th>" . _("Search") . "</th>\n";
+    echo "<td>";
+
     echo "<input type='hidden' name='action' value='search'>\n";
-    echo "<input type='text' value='$search_text' name='search_text' size='50'>\n";
+    echo "<input type='text' value='$search_text' name='search_text' style='width: 20em'>\n";
 
     foreach ($SearchParams_choices as $param_name => $choices)
     {
         $value = (int) get_enumerated_param($_REQUEST, $param_name, '999', array_keys($choices));
         echo dropdown_select($param_name, $value, $choices);
     }
+
+    echo "<input type='submit' value='Search'></td>\n";
+    echo "</tr>\n";
+    echo "</table></form><br>\n";
 }
 
 function SearchParams_get_sql_condition($request_params)
@@ -571,7 +580,8 @@ if (!isset($_REQUEST['task_id'])) {
                 $header .= ": " . $_REQUEST['search_text'];
             }
             TaskHeader($header);
-
+            echo "<h1 style='margin-top: 0;'>" . _("Task Center") . "</h1>";
+            SearchParams_echo_controls();
             search_and_list_tasks($_REQUEST);
             break;
 
@@ -579,6 +589,8 @@ if (!isset($_REQUEST['task_id'])) {
             // The user just entered the Task Center
             // (e.g., by clicking the "Report a Bug" link).
             TaskHeader("All Open Tasks", true);
+            echo "<h1 style='margin-top: 0;'>" . _("Task Center") . "</h1>";
+            SearchParams_echo_controls();
             list_all_open_tasks();
             break;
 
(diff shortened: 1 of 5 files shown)


tasks.php
```diff
@@ -340,22 +339,31 @@ function SearchParams_echo_controls()
...
-    echo "<input type='text' value='$search_text' name='search_text' size='50'>\n";
+    echo "<input type='text' value='$search_text' name='search_text' style='width: 20em'>\n";
 
     foreach ($SearchParams_choices as $param_name => $choices)
     {
         $value = (int) get_enumerated_param($_REQUEST, $param_name, '999', array_keys($choices));
         echo dropdown_select($param_name, $value, $choices);
     }
+
+    echo "<input type='submit' value='Search'></td>\n";
```
chrismiceli, line 364: Localization? I see we do it sometimes and not others, so probably something to address later.

chrismiceli (approved):

tasks.php
cpeel: Very little of this file has been localized, unfortunately. I did some in places I touched when I was thinking about it -- but clearly not all 😦 . It's probably worth a separate MR to clean all of these up (and a second separate MR to update all of the `mysqli_query()` and escape clauses.

srjfoo (approved):
Apologies ahead of time for the length of this comment. 🙂 --

In the task detail block, granted I'm deliberately pushing the limits, with a narrow window, the right-hand table overlaps the left-hand table:
image

Because the two blocks are independent tables, the height can vary from nearly (but not quite) the same height ...
Firefox:
image

Chrome:
image

to one quite a bit taller:

Chrome:
image

Firefox:
image

Because of the disparity in how browsers handle it, the position of the "Details" heading varies, depending. In the following screenshot, the "Details" heading is closer to the bottom of the task details table than it is to the details text block (and closer than the task heading).
image

Predictably, in Firefox, there's plenty of space above in that particular version, but when the tables are almost the same height, it's the same visual issue of the Details heading being closer to the block above it than the actual detail text. (And I really hope this is all making sense.)

For the comments -- this is purely an aesthetic reaction. It's pretty obvious from the link to the username and the bold text where each comment begins, but visually, I'd like a little more space between the bottom of one comment and the top of the next (instead of the same amount as between the heading for a comment and the related comment). Or a thin horizontal line as is used for the news items.

And finally, another purely aesthetic reaction: Is there a real need to use the theme colors in the task-details-block? That's a very large block of solid color.

All that being said, it's fully functional, and I can certainly live with it.

cpeel:
> In the task detail block, granted I'm deliberately pushing the limits, with a narrow window, the right-hand table overlaps the left-hand table:

At the risk of sounding too cavalier: yes, and? The current implementation does something similar with the rows, but because they shoved the whole thing in a single table with the same background color you can't tell as readily.

> Because of the disparity in how browsers handle it, the position of the "Details" heading varies, depending. In the following screenshot, the "Details" heading is closer to the bottom of the task details table than it is to the details text block (and closer than the task heading).

Yeah, that's annoying and I haven't figured out how to deal with it. I also haven't figured out if I really care.

> For the comments -- this is purely an aesthetic reaction. It's pretty obvious from the link to the username and the bold text where each comment begins, but visually, I'd like a little more space between the bottom of one comment and the top of the next (instead of the same amount as between the heading for a comment and the related comment). Or a thin horizontal line as is used for the news items.

I will futz with the padding some. I'd prefer not to have a horizontal line.

> And finally, another purely aesthetic reaction: Is there a real need to use the theme colors in the task-details-block? That's a very large block of solid color.

Based on my testing we need _something_ in the "label" side. Having it all white made it all mush up together. I'm open to other ideas.

srjfoo:
> > In the task detail block, granted I'm deliberately pushing the limits, with a narrow window, the right-hand table overlaps the left-hand table:
> 
> At the risk of sounding too cavalier: yes, and? The current implementation does something similar with the rows, but because they shoved the whole thing in a single table with the same background color you can't tell as readily.

That may be why they shoved the whole thing into a single table with the same background color. I think, too, that the  As I said, I can live with it.

Hm. Taking another look, and comparing it with the current task center, it's not just that -- the current task center has a lot more background color, and it's lighter, with a much lighter border color. Sorry I don't have any helpful suggestions. :(

...

> > And finally, another purely aesthetic reaction: Is there a real need to use the theme colors in the task-details-block? That's a very large block of solid color.
> 
> Based on my testing we need _something_ in the "label" side. Having it all white made it all mush up together. I'm open to other ideas.

Yeah, I can see that it would.

Perhaps it's just that I'm used to the project search form which does have a white background for the label columns. The difference is that the label columns are right-aligned, and it's clear which label is associated with what. That may not be appropriate here.

tasks.php
```diff
@@ -1086,11 +1046,11 @@ function select_and_list_tasks($sql_condition)
...
     $columns = array(
         'task_id'          => " style='text-align: center;'",
-        'task_summary'     => " class='wrap'",
-        'task_type'        => "",
+        'task_summary'     => "",
+        'task_type'        => " style='white-space: nowrap;'",
         'task_severity'    => "",
         'task_priority'    => "",
-        'date_edited'      => " style='text-align: center;'",
+        'date_edited'      => " style='white-space: nowrap; text-align: center;'",
```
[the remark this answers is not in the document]
cpeel: No reason - I will update them.

cpeel:
> Perhaps it's just that I'm used to the project search form which does have a white background for the label columns. The difference is that the label columns are right-aligned, and it's clear which label is associated with what. That may not be appropriate here.

<nod> I think the Project Search is the only place where we do that right-aligning bit. While we could do that here too, that's not a design that I want to encourage.

The sidebar color for the label fields is what is used on the User Preferences page, but there the rows have borders too. Team pages is another place where we have a similar label:value table and those have the background color for the whole table for both the label and values.

As usual: it appears that we are wildly inconsistent in this 😁

cpeel:
I've added an explicit row height so that the two tables at the top of an issue line up if there are the same number of rows. If there aren't the same number, however, they'll still be misaligned.

I've also added more spacing to pull the `Details` header down a bit and modified the comment spacing.

srjfoo:
> > Perhaps it's just that I'm used to the project search form which does have a white background for the label columns. The difference is that the label columns are right-aligned, and it's clear which label is associated with what. That may not be appropriate here.
> 
> I think the Project Search is the only place where we do that right-aligning bit. While we could do that here too, that's not a design that I want to encourage.
> 
> The sidebar color for the label fields is what is used on the User Preferences page, but there the rows have borders too. Team pages is another place where we have a similar label:value table and those have the background color for the whole table for both the label and values.
> 
> As usual: it appears that we are wildly inconsistent in this 😁

🤣  (you don't say!)

srjfoo:
> I've added an explicit row height so that the two tables at the top of an issue line up if there are the same number of rows. If there aren't the same number, however, they'll still be misaligned.
> 
> I've also added more spacing to pull the `Details` header down a bit and modified the comment spacing.

It all looks better, including the little bit of extra space between the comments. 🙂

PR #456, opened 2020-10-12
Add 1, 7, and 30 day active users to v1/stats/site

bpfoley:
(All publicly available information already.)

While here clean up default.php to use new PHP features.

### DIFF
--- a/api/dp-openapi.yaml
+++ b/api/dp-openapi.yaml
@@ -600,6 +600,15 @@ components:
         registered_users:
           type: integer
           readOnly: true
+        active_users_1_day:
+          type: integer
+          readOnly: true
+        active_users_7_day:
+          type: integer
+          readOnly: true
+        active_users_30_day:
+          type: integer
+          readOnly: true
       description: Statistics for the site
 
     round_stats:
--- a/api/v1_stats.inc
+++ b/api/v1_stats.inc
@@ -11,10 +11,23 @@ function api_v1_stats_site($method, $data, $query_params)
     $res = DPDatabase::query("SELECT COUNT(*) FROM users");
     list($registered_users) = mysqli_fetch_row($res);
 
-    return [
+    $stats = [
         "server_time" => date(DATE_ATOM),
         "registered_users" => $registered_users,
     ];
+
+    foreach ( array(1,7,30) as $days_back )
+    {
+        $res = DPDatabase::query(sprintf("
+            SELECT COUNT(*)
+            FROM users
+            WHERE t_last_activity > UNIX_TIMESTAMP() - %d * 24*60*60
+        ", $days_back));
+        list($num_users) = mysqli_fetch_row($res);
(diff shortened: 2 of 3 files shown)


api/v1_stats.inc
```diff
@@ -11,10 +11,22 @@ function api_v1_stats_site($method, $data, $query_params)
...
 
-    return [
+    $stats = [
         "server_time" => date(DATE_ATOM),
         "registered_users" => $registered_users,
     ];
+
+    foreach ( array(1,7,30) as $days_back )
+    {
+        $res = DPDatabase::query("
```
cpeel, line 21: As part of preventing SQL injections, we've been using `sprintf()` to insert string and integer variables, like `$days_back`, into SQL. Even here where it's not user input we'd like to use it because some developer down the road may make it user-selectable and then we have a vulnerability. eg:

```php
    $sql = sprintf("
        SELECT COUNT(*)
        FROM users
        WHERE t_last_activity > UNIX_TIMESTAMP() - %d * 24*60*60
    ", $days_back);
```

api/v1_stats.inc
```diff
@@ -11,10 +11,22 @@ function api_v1_stats_site($method, $data, $query_params)
...
+            SELECT COUNT(*)
+            FROM users
+            WHERE t_last_activity > UNIX_TIMESTAMP() - $days_back * 24*60*60
+        ");
+        list($num_users) = mysqli_fetch_row($res);
+        $stats[$days_back . "_day_active_users"] = $num_users;
```
cpeel, line 27: I would probably swap the order of the key so that it doesn't start with a number such that it would be a valid variable name should some consumer want to do so: `active_users_1_day`, etc.

cpeel (commented):
Minor comments but overall looks good!

api/v1_stats.inc
bpfoley: ACK. Done.

api/v1_stats.inc
bpfoley: The tyranny of C-style identifiers :-)

api/dp-openapi.yaml
```diff
@@ -600,6 +600,15 @@ components:
         registered_users:
           type: integer
           readOnly: true
+        1_day_active_users:
```
cpeel, line 603: Don't forget to change these to match!

bpfoley: Woops. Done.

default.php
```diff
@@ -137,13 +137,12 @@
 // Show the number of users that have been active over various recent timescales.
 foreach ( array(1,7,30) as $days_back )
 {
-    $res = mysqli_query(DPDatabase::get_connection(), "
+    $res = DPDatabase::query("
```
cpeel, line 140: `sprintf()` here too please -- since you're changing it.

PR #457, opened 2020-10-13
List destructuring

bpfoley:


cpeel:
I've thrown this into a [list-destructuring](https://www.pgdp.org/~cpeel/c.branch/list-destructuring/) sandbox for testing.

70ray (approved):

pinc/ProjectSearchResults.inc
```diff
@@ -607,8 +607,7 @@ class ProjectSearchResults
         $numrows = mysqli_num_rows($result);
 
         $res_found = DPDatabase::query("SELECT FOUND_ROWS()");
```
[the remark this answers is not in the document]
cpeel: That's really great to know Ray, thanks. Ubuntu 20.04 will be running 8.0.21, so while we don't need to do something about this immediately, we'll need to do something about it by at least 2024.

srjfoo (approved):
I saw no problems with the things I could figure out how to test.

bpfoley:
Since this is PHP and the compiler won't save us from ourselves, I'd be happy to split this up into separate MRs which we can merge one at a time. There's two completely mechanical changes (which really shouldn't break anything) and two slightly trickier ones (which might have odd corner cases).

Mechanical
1. Replace `mysqli_real_escape_string(DPDatabase::get_connection(), X)` with `DPDatabase::escape(X)`
2. Use list destructuring.

Slightly trickier:
1. Replace `mysqli_query(DPDatabase::get_connection(), $sql) or die(DPDatabase::log_error());` and cases of `mysqli_query(DPDatabase::get_connection(), $sql)` where we probably should have failures on errors with `DPDatabase::query($sql)`
2. Wrap many of the `::query("sql with $interpolation")` expressions in `sprintf`.

Or, if you think it's safe enough, we can merge what we've got, and I can do followup MRs for the rest of the codebase.

pinc/dpsession_via_php_sessions.inc
```diff
@@ -134,34 +138,38 @@ function mysql_session_select($sid)
...
     if ( $SessionExists == 0 )
     {
-        $query = "
+        $query = sprintf("
             INSERT INTO sessions
                 (sid, expiration, value)
             VALUES
-                ('$sid', '$expiration', '$value')
-            ";
+                ('%s', '%s', '%s')",
```
cpeel, line 156: Since `expiration` is an integer, we should use `%d` for it here and not escape it.

pinc/dpsession_via_php_sessions.inc
```diff
@@ -134,34 +138,38 @@ function mysql_session_select($sid)
...
     else
     {
-        $query = "
+        $query = sprintf("
             UPDATE sessions
             SET
-                expiration = '$expiration',
-                value = '$value'
-            WHERE sid = '$sid' AND expiration > ".time()
+                expiration = '%s',
```
cpeel, line 166: Since `expiration` is an integer, we should use `%d` for it here and not escape it.

cpeel (commented):
@chrismiceli and I have been working through the codebase making these changes, and we've seen both of us fat-finger & break something -- even in the mechanical changes -- so we've been testing things as we go. Mostly because, as you've seen, some of our SQL is sloppy and we're using strings where we should be using integers, etc.

We want to replace all of the `mysqli_query()` calls with `DPDatabase::query()`. We want most, but not all, of the `mysqli_query()` calls to die on failure. "Not all" is the tricky part.

We are also trying to standardize how we're doing our SQL escaping, including building the SQL separate from running the query so it's easier to debug later. See https://www.pgdp.net/wiki/DP_Code_Best_Practices#Building_SQL

Breaking this up into a couple of smaller, testable, MRs would be preferable.

pinc/dpsession_via_php_sessions.inc
bpfoley: OK.

pinc/dpsession_via_php_sessions.inc
bpfoley: OK.

bpfoley:
> Breaking this up into a couple of smaller, testable, MRs would be preferable.

OK. That would seem to argue towards doing a few files at a time rather than doing single mechanical search-and-replaces across the whole codebase. I'll abandon this MR and create a few others.

PR #458, opened 2020-10-13
Use R202009 as an upgrade checkpoint

cpeel:
Rather than continue to have all releases support upgrades from any release, use the R202009 release as a checkpoint. Users wanting to upgrade from an earlier version need to upgrade to R202009 first before upgrading past it. 20 years seems like a reasonable checkpoint.

This also means we can stop talking and providing data on phpBB 2.x.


srjfoo (approved):

SETUP/UPGRADE.md
```diff
@@ -80,108 +80,15 @@ you want to run and add the following line immediately after the opening
...
-* c/SETUP/upgrade/09/
-* c/SETUP/upgrade/10/
-* c/SETUP/upgrade/11/
-* c/SETUP/upgrade/12/
-* c/SETUP/upgrade/13/
-* c/SETUP/upgrade/14/
-
-### Upgrading from release 1.5
-Run the scripts in the following directories in order
+If you are upgrading a release before R202009, you must upgrade to R202009
```
chrismiceli, line 85: How are they supposed to do this? Should we mention a tag or something in git for them to get instructions for upgrading to R202009?

cpeel: They're suppose to download and install R202009 first. I will make that clearer and reference the Releases page.

cpeel: I've added a link to the releases page in the message. I didn't add more specific wording because if they are upgrading from a release, they should have either downloaded it from Github, or they are using git and it should be clear based on the tags.

chrismiceli (approved):

70ray (approved):

PR #460, opened 2020-10-14
Usertally

bpfoley:
Report number of users who've proofed/formatted one or more page in a round.

### DIFF
--- a/api/dp-openapi.yaml
+++ b/api/dp-openapi.yaml
@@ -647,6 +647,9 @@ components:
         month_actual:
           type: integer
           readOnly: true
+        positive_users:
+          type: integer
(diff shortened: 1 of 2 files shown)


bpfoley:
Yeah, that was my mistake. I managed to accidentally do a merge on git pull because I hadn't set up `~/.gitconfig` properly on my VM. See https://blog.sffc.xyz/post/185195398930/why-you-should-use-git-pull-ff-only

cpeel:
Ah. Yeah, I never use `git pull`, ever. I always `git fetch` and `git rebase`. It's wordier but it's explicit and always Does What I Want 😁

PR #462, opened 2020-10-15
remove-dotted-border: removed dotted bottom border

srjfoo:
Manage Image Sources (either to propose a new image source or edit an existing one) was the only place that a dotted border was used, and it was used simply as a spacer (Permissions row). Removed bottom border and let the drop-down menus be located next to the labels.

Project Managers and Project Facilitators can use this for proposing new image sources. Only Image Sources Managers, and site admins can edit.

Testable: [remove-dotted-border](https://www.pgdp.org/~srjfoo/c.branch/remove-dotted-border/tools/project_manager/projectmgr.php)

### DIFF
--- a/tools/project_manager/manage_image_sources.php
+++ b/tools/project_manager/manage_image_sources.php
@@ -359,7 +359,7 @@ class ImageSource
                 ? (empty($_REQUEST[$field]) ? '-1' : $_REQUEST[$field])
                 : $this->$field;
 
-            $editing .= "<div class='perms_wrapper'>$col[label]</div>";
+            $editing .= "$col[label]: ";
             $editing .= "<select name='$col[field]'>";
             foreach (array('1' => _('Yes'),'0' => _('No'),'-1' => _('Unknown')) as $val => $opt)
             {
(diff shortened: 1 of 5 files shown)


tools/project_manager/manage_image_sources.php
```diff
@@ -384,7 +384,9 @@ function _show_edit_permissions_row()
             ? (empty($_REQUEST[$field]) ? '2' : $_REQUEST[$field])
             : $this->$field;
 
-        $editing .= "<div class='perms_wrapper'>" . _("Visibility on Info Page") . "</div> <select name='$field'>";
+        $editing .= _("Visibility on Info Page: ");
+        $editing .= "<select name= '$field'>";
```
chrismiceli, line 388: space after = could be removed. Also, we could attribute encode field, but seems safe currently.

srjfoo: Thanks for catching that extra space.

I've added `attr_safe`, but please bear in mind that I will need to be told where and when to add it until I get a better handle on when and where it's appropriate. And, given that I've seen different patterns within the code, probably how, as well. 😁

chrismiceli: I think the idea is that if you don't control the content, you need to be prepared it may have invalid characters in regards to being used as an attribute (characters such as ' or " which could allow XSS attacks. In this case it appears we control it, but there is no harm in encoding something that won't change after encoding so it is best (in my opinion) to just always encode.

chrismiceli (approved):

PR #465, opened 2020-10-16
q3: Clean up queries using new query & escape fns, list destructuring.

bpfoley:
Several files with small changes and a larger set of changes to `ProjectSearchResults.inc`

### DIFF
--- a/SETUP/upgrade/06/update_projects_table.php
+++ b/SETUP/upgrade/06/update_projects_table.php
@@ -63,8 +63,7 @@ $res = mysqli_query(DPDatabase::get_connection(), "
     FROM projects
     WHERE LEFT(comments,8) = 'special:'
 ") or die(mysqli_error(DPDatabase::get_connection()));
-$row = mysqli_fetch_row($res);
-$num_left = $row[0];
+list($num_left) = mysqli_fetch_row($res);
 echo "There are $num_left project comments left that begin with 'SPECIAL:'.\n";
 
 echo "\n";
@@ -85,8 +84,7 @@ $res = mysqli_query(DPDatabase::get_connection(), "
     FROM queue_defns
     WHERE INSTR(project_selector,'special:')
 ") or die(mysqli_error(DPDatabase::get_connection()));
-$row = mysqli_fetch_row($res);
-$num_left = $row[0];
+list($num_left) = mysqli_fetch_row($res);
 echo "There are $num_left queue_defns left whose project_selector mentions 'SPECIAL:'.\n";
 
 
@@ -107,8 +105,7 @@ $res = mysqli_query(DPDatabase::get_connection(), "
     FROM projects
     WHERE INSTR(comments,'document.php')
 ") or die(mysqli_error(DPDatabase::get_connection()));
-$row = mysqli_fetch_row($res);
-$num_left = $row[0];
+list($num_left) = mysqli_fetch_row($res);
 echo "There are still $num_left project comments left that mention document.php in some way.\n";
 
 echo "\n";
--- a/stats/release_queue.php
+++ b/stats/release_queue.php
@@ -65,25 +65,32 @@ if (!isset($name))
         echo "</tr>\n";
     }
 
-    $q_res = mysqli_query(DPDatabase::get_connection(), "
+    $q_sql = sprintf("
         SELECT *
         FROM queue_defns
-        WHERE round_id='$round_id'
-        ORDER BY ordering
-    ") or die(DPDatabase::log_error());
+        WHERE round_id='%s'
+        ORDER BY ordering",
+        DPDatabase::escape($round_id)
+    );
+    $q_res = DPDatabase::query($q_sql);
+
     while ( $qd = mysqli_fetch_object($q_res) )
     {
         $cooked_project_selector = cook_project_selector($qd->project_selector);
-        $c_res = mysqli_query(DPDatabase::get_connection(), "
+        $state_clause = sprintf(
+            "state='%s'",
+            DPDatabase::escape($round->project_waiting_state)
+        );
+        $c_sql = "
             SELECT COUNT(*)
             FROM projects
             WHERE ($cooked_project_selector)
-                AND state='{$round->project_waiting_state}'
-        ");
+                AND $state_clause
+        ";
+        $c_res = DPDatabase::query($c_sql);
         if ($c_res)
         {
-            $row = mysqli_fetch_row($c_res);
-            $current_length = $row[0];
+            list($current_length) = mysqli_fetch_row($c_res);
         }
         else
         {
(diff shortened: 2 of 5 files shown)


SETUP/upgrade/06/update_projects_table.php
```diff
@@ -63,8 +63,7 @@
     FROM projects
     WHERE LEFT(comments,8) = 'special:'
 ") or die(mysqli_error(DPDatabase::get_connection()));
-$row = mysqli_fetch_row($res);
-$num_left = $row[0];
```
chrismiceli, line 67: do we in general want to update the upgrade scripts? It just increases test scope that I am not sure is entirely necessary.

stats/release_queue.php
```diff
@@ -152,23 +152,23 @@
     }
 
 
-    $comments_url1 = mysqli_real_escape_string(DPDatabase::get_connection(), "<a href='$code_url/project.php?id=");
-    $comments_url2 = mysqli_real_escape_string(DPDatabase::get_connection(), "'>");
-    $comments_url3 = mysqli_real_escape_string(DPDatabase::get_connection(), "</a>");
+    $comments_url1 = DPDatabase::escape("<a href='$code_url/project.php?id=");
```
chrismiceli, line 155: are these unused?

stats/release_queue.php
```diff
@@ -152,23 +152,23 @@
...
         SELECT
 
             concat('$comments_url1',projectID,'$comments_url2', nameofwork, '$comments_url3') as '" 
-                . mysqli_real_escape_string(DPDatabase::get_connection(), _("Name of Work")) . "',
-            authorsname as '" . mysqli_real_escape_string(DPDatabase::get_connection(), _("Author's Name")) . "',
-            language    as '" . mysqli_real_escape_string(DPDatabase::get_connection(), _("Language")) . "',
-            genre       as '" . mysqli_real_escape_string(DPDatabase::get_connection(), _("Genre")) . "',
-            difficulty  as '" . mysqli_real_escape_string(DPDatabase::get_connection(), _("Difficulty")) . "',
-            username    as '" . mysqli_real_escape_string(DPDatabase::get_connection(), _("Project Manager")) . "',
+                . DPDatabase::escape(_("Name of Work")) . "',
```
chrismiceli, line 163: we can't use sprintf here?

tools/project_manager/manage_image_sources.php
```diff
@@ -100,7 +99,7 @@
         else
         {
             output_header('', NO_STATSBAR);
-                if ($new)
+                if ($count > 0)
```
chrismiceli, line 102: Is this right? You replace `!$new` above with `$count > 0` and `$new` here with the same `$count > 0`.

SETUP/upgrade/06/update_projects_table.php
bpfoley: Perhaps not. I'm happy to leave them be if you prefer.

stats/release_queue.php
bpfoley: They're used in the SQL statement directly below this. HTML in the database! :)

stats/release_queue.php
bpfoley: True. There's a bunch of unwieldy concats here, but the main problem with `sprintf` is that it makes it harder on the reader to check which `%s` matches up with which `sprintf` argument. If only PHP format specifiers could take named arguments like Python/Rust can.

tools/project_manager/manage_image_sources.php
bpfoley: Semantically it's the same. $count is the result of a `SELECT COUNT(*)` statement which is always non-negative. `$new` was set if the count was 0, so if `$count > 0` then we have found an already existing image with the same ID.

SETUP/upgrade/06/update_projects_table.php
cpeel: This upgrade script in particular is going to get blown away in https://github.com/DistributedProofreaders/dproofreaders/pull/458 so I don't care either way and am happy to punt that decision down the road a bit.

stats/release_queue.php
cpeel: Well, HTML in DB queries because of the way some of the tables are output. It's horrific, don't look at the code.

stats/release_queue.php
cpeel: Yes, let's not try to fix this with `sprintf()` here since we need to just stop trying to squeeze direct table output. I aspire to rip all the `dpsql_dump_*()` functions out at some point.

What you have here is good, @bpfoley.

tools/project_manager/manage_image_sources.php
cpeel: No, Chris is right. You've changed the logic here such that the request for approval will happen for *existing* records (`$count > 0`) when it needs to happen for *new* records (`$count == 0`).

cpeel (commented):
Except for the comment in `manage_image_sources.php` all of this looks good and tests out.

tools/project_manager/manage_image_sources.php
bpfoley: Oh blast. I'm an idiot. Thanks both. Updated.

cpeel:
Looks like you have a linting failure:
```
./../stats/release_queue.php
PHP Parse error:  syntax error, unexpected ')' in ./../stats/release_queue.php on line 86
ERROR: PHP lint failure in ./../stats/release_queue.php
```

bpfoley:
> Looks like you have a linting failure:
> 
> ```
> ./../stats/release_queue.php
> PHP Parse error:  syntax error, unexpected ')' in ./../stats/release_queue.php on line 86
> ERROR: PHP lint failure in ./../stats/release_queue.php
> ```

Huh. How strange. I thought I ran it through a linter before pushing.

stats/release_queue.php
```diff
@@ -65,25 +65,29 @@
...
+    $q_res = DPDatabase::query($q_sql);
+
     while ( $qd = mysqli_fetch_object($q_res) )
     {
         $cooked_project_selector = cook_project_selector($qd->project_selector);
-        $c_res = mysqli_query(DPDatabase::get_connection(), "
+        $c_sql = sprintf("
             SELECT COUNT(*)
             FROM projects
             WHERE ($cooked_project_selector)
```
cpeel, line 83: `$cooked_project_selector` contains `%`s so we can't use `sprintf()` here (found by testing in my sandbox).

cpeel (commented):
It's stuff like this that makes these changes so freak'n challenging to both make and test :/

stats/release_queue.php
bpfoley: Oof. OK, I've really got to check out how to use the sandboxes again.

PR #466, opened 2020-10-16
Deduplicate DPDatabase code implemented by query()

bpfoley:


pinc/DPDatabase.inc
```diff
@@ -111,15 +107,11 @@ final class DPDatabase
...
+        $result = self::query(sprintf("
             SELECT TABLE_COLLATION
             FROM information_schema.tables
             WHERE table_schema='%s' AND table_name='%s'
-        ", self::$_db_name, $table_name);
-
-        $result = mysqli_query(DPDatabase::get_connection(), $sql);
-        if(!$result)
-            throw new DBQueryError(DPDatabase::log_error());
+        ", self::$_db_name, $table_name));
```
chrismiceli, line 114: Should we encode table name or do we trust every caller of this method?

cpeel: Yes, let's escape the table name. Let's also retain the separation between creating the `$sql` and running the query.

bpfoley: Done.

PR #467, opened 2020-10-16
Optional Latex math preview using MathJax

70ray:
This allows preview of Latex math markup.
It incorporates changes recommended by @cpeel in an earlier version.

Can be tested in https://www.pgdp.org/~rp31/c.branch/math_preview_3

### DIFF
--- a/SETUP/tests/jsTests/formatPreviewTests.js
+++ b/SETUP/tests/jsTests/formatPreviewTests.js
@@ -97,8 +97,8 @@ QUnit.module("Format preview test", function() {
     QUnit.test("unrecognised tag, u not enabled", function (assert) {
         text = "ab <u>cd</u>";
         issArray = analyse(text, configuration);
-        issueTest(assert, 1, 3, 1, "unRecTag", 0);
-        issueTest(assert, 0, 8, 1, "unRecTag", 0);
+        issueTest(assert, 1, 3, 3, "unRecTag", 0);
+        issueTest(assert, 0, 8, 4, "unRecTag", 0);
     });
 
     QUnit.test("u tag enabled", function (assert) {
--- a/tools/proofers/preview.js
+++ b/tools/proofers/preview.js
@@ -333,14 +333,29 @@ $(function () {
 
         // check for an unrecognised tag
         function unRecog() {
-            var re = new RegExp("<(?!(?:\\/?(?:" + ILTags + ")|tb)>)", "g");
-            var result;
+            let reMaybeTag = /<(\/?)(\w{1,3})>/g;
+            let reGoodTag = new RegExp("^(?:" + ILTags + ")$");
             while (true) {
-                result = re.exec(txt);
+                let result = reMaybeTag.exec(txt);
                 if (null === result) {
                     break;
                 }
-                reportIssue(result.index, 1, "unRecTag");
+                let tagString = result[2];
+                let tagLen = result[0].length;
+
+                if(tagString === "tb") {
+                    if(result[1]) {
+                        // found </tb>
+                        reportIssue(result.index, tagLen, "unRecTag");
+                    }
+                    // ignore thought break
+                    continue;
+                }
+
+                if(!reGoodTag.test(tagString)) {
+                    reportIssue(result.index, tagLen, "unRecTag");
+                    continue;
+                }
             }
         }
 
@@ -689,6 +704,49 @@ $(function () {
             }
         }
 
+        // check that math delimiters \[ \], \( \) are correctly matched
+        function checkMath() {
+            var mathRe = /\\\[|\\\]|\\\(|\\\)/g;
+            var result;
+            var tag, openTag = null;
+            var openStart = 0;
+            while (true) {
+                result = mathRe.exec(txt);
+                if (null === result) {
+                    // if there is a tag open mark it as an error
+                    if (openTag) {
+                        reportIssue(openStart, 2, "noEndTag");
+                    }
+                    break;
+                }
+                tag = result[0].charAt(1);
+                // if no open tag and ( or [ set open else error
+                if (!openTag) {
+                    if ((tag === '(') || (tag === '[')) {
+                        openTag = tag;
+                        openStart = result.index;
+                    } else {
+                        reportIssue(result.index, 2, "noStartTag");
+                    }
+                } else {
+                    if (((openTag === '(') && (tag === ')')) || ((openTag === '[') && (tag === ']'))) {
+                        // correctly matched
+                        openTag = null;
+                    } else if (((openTag === '(') && (tag === ']')) || ((openTag === '[') && (tag === ')'))) {
+                        // mismatched
+                        openTag = null;
+                        reportIssue(openStart, 2, "misMatchTag");
+                        reportIssue(result.index, 2, "misMatchTag");
+                    } else {
+                        // a start tag of either sort follows a start tag
+                        reportIssue(openStart, 2, "noEndTag");
+                        openTag = tag;
+                        openStart = result.index;
+                    }
+                }
+            }
+        }
+
         parseInLine();
         // if inline parse fails then checkSC might not work
         if (parseOK) {
@@ -704,6 +762,9 @@ $(function () {
         unRecog();
         checkTab();
         checkBlankLines();
+        if(config.allowMathPreview) {
+            checkMath();
+        }
 
         // we must avoid two issues giving overlapping markup
         // sort them first and mark from end towards beginning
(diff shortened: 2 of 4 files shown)


lhamilton1:
Nice to have this change!

tools/proofers/preview.js
```diff
@@ -689,6 +704,49 @@ $(function () {
             }
         }
 
+        // check that math delimiters \[ \], \( \) are correctly matched
+        function checkMath() {
+            var mathRe = /\\\[|\\\]|\\\(|\\\)/g;
+            var result;
+            var tag, openTag = null;
+            var openStart = 0;
+            while (true) {
```
chrismiceli, line 713: infinite loops scare me in case the break is never hit. Would it be possible to rework or somehow guarantee we will always break?

tools/proofers/preview.js
```diff
@@ -689,6 +704,49 @@ $(function () {
             }
         }
 
+        // check that math delimiters \[ \], \( \) are correctly matched
+        function checkMath() {
+            var mathRe = /\\\[|\\\]|\\\(|\\\)/g;
+            var result;
+            var tag, openTag = null;
```
chrismiceli, line 711: why set to null, I only see falsy checks, nothing explicitly to null.

tools/proofers/preview.js
```diff
@@ -689,6 +704,49 @@ $(function () {
...
+                    break;
+                }
+                tag = result[0].charAt(1);
+                // if no open tag and ( or [ set open else error
+                if (!openTag) {
+                    if ((tag === '(') || (tag === '[')) {
+                        openTag = tag;
+                        openStart = result.index;
+                    } else {
+                        reportIssue(result.index, 2, "noStartTag");
```
chrismiceli, line 729: It would be nice if we had an enum type for this. Something like:
`
var ISSUES = {
  /* ... */
  NO_START_TAG: 'noStartTag',
};
`

tools/proofers/preview.js
```diff
@@ -895,12 +956,42 @@ $(function () {
...
             }
-            // sub- and super-scripts
-            txt = txt.replace(/_\{([^\}]+)\}/g, '<span class="sub"' + colorString);
-            txt = txt.replace(/\^\{([^\}]+)\}/g, '<span class="sup"' + colorString);
-            // single char superscript -  any char except {
-            // do not allow < incase it's a tag which would screw up
-            txt = txt.replace(/\^([^\{<])/g, '<span class="sup"' + colorString);
+
+            // process sub- and super-scripts
+            function ssProcess(text) {
```
chrismiceli, line 961: is ss common to refer to sub and super scripts? I only think of some other historical entities.

tools/proofers/preview.js
```diff
@@ -895,12 +956,42 @@ $(function () {
...
+                text = text.replace(/_\{(.+?)\}/g, '<span class="sub"' + colorString);
+                text = text.replace(/\^\{(.+?)\}/g, '<span class="sup"' + colorString);
+                // single char superscript -  any char except {
+                // do not allow < incase it's a tag which would screw up
```
chrismiceli, line 965: should we just encode and cover all issues?

tools/proofers/preview.js
```diff
@@ -895,12 +956,42 @@ $(function () {
...
+            // processes the text by textFunction but in math mode
+            // only outside math markup
+            function processExMath(text, textFunction) {
+                if(!styler.allowMathPreview) {
+                    return textFunction(txt);
+                } else {
+                    // find whole math strings \[ ... \] or \( ... \)
+                    let txtOut = "";
+                    let mathRegex = /\\\[[^]*?\\\]|\\\([^]*?\\\)/g;
+                    for(;;) {
```
chrismiceli, line 978: same here for infinite loops. And if we do use the, we should probably be consistent on for(;;) vs while(true)

tools/proofers/preview.js
```diff
@@ -895,12 +956,42 @@ $(function () {
...
-            // do not allow < incase it's a tag which would screw up
-            txt = txt.replace(/\^([^\{<])/g, '<span class="sup"' + colorString);
+
+            // process sub- and super-scripts
+            function ssProcess(text) {
+                text = text.replace(/_\{(.+?)\}/g, '<span class="sub"' + colorString);
+                text = text.replace(/\^\{(.+?)\}/g, '<span class="sup"' + colorString);
+                // single char superscript -  any char except {
+                // do not allow < incase it's a tag which would screw up
+                return text.replace(/\^([^{<])/g, '<span class="sup"' + colorString);
```
chrismiceli, line 966: this function can't decide if it is immutable or not it seems. If we return the text, why modify the argument? I'd prefer to not modify arguments in general.

tools/proofers/preview.js
```diff
@@ -895,12 +956,42 @@ $(function () {
...
+            }
+
+            // processes the text by textFunction but in math mode
+            // only outside math markup
+            function processExMath(text, textFunction) {
```
chrismiceli, line 971: can we write unit tests for newer code?

SETUP/tests/jsTests/formatPreviewTests.js
```diff
@@ -97,8 +97,8 @@ QUnit.module("Format preview test", function() {
     QUnit.test("unrecognised tag, u not enabled", function (assert) {
         text = "ab <u>cd</u>";
         issArray = analyse(text, configuration);
-        issueTest(assert, 1, 3, 1, "unRecTag", 0);
-        issueTest(assert, 0, 8, 1, "unRecTag", 0);
+        issueTest(assert, 1, 3, 3, "unRecTag", 0);
```
chrismiceli, line 100: why did this change? I don't see the test being related to math at all.

tools/proofers/previewControl.js
```diff
@@ -180,6 +189,15 @@ $( function() {
         setViewColors(outerPrev);
         viewMode = previewStyles.initialViewMode;
         $("#" + viewMode).prop("checked", true);
+        // check if MathJax already loaded. Will break if load more than once
+        if(previewStyles.allowMathPreview && (typeof(MathJax) === 'undefined')) {
```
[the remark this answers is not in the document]
70ray: I did it like this to allow caching which is disabled by default with getScript(). Although it probably won't make much difference. Perhaps it would be better to use the simplest code rather an unnoticeable improvement in performance.

tools/proofers/preview.js
70ray: I suppose we could run the loop a large number of times (1000?) and if it doesn't "break" before this show a warning message.

tools/proofers/preview.js
70ray: I suppose you are right, I will change it.

tools/proofers/preview.js
70ray: Yes, I'll look into this and perhaps make another branch to address it.

tools/proofers/preview.js
70ray: I guess you mean this: https://en.wikipedia.org/wiki/SS_Jaguar_100.
I'll make a better name.

tools/proofers/preview.js
70ray: It's a long time ago, but I think this was meant to avoid two overlapping html markups which would produce a nonsensical result, but I'll have another look at it.

tools/proofers/preview.js
70ray: Yes I think eslint 'no constant condition' will catch while(true) but allow for(;;).
I'll make another branch to tidy this up.

tools/proofers/preview.js
70ray: This was written several months ago. Itdoes seem a bit strange. I'll look at it again.

tools/proofers/preview.js
70ray: I'll do that.

SETUP/tests/jsTests/formatPreviewTests.js
70ray: I'm sure there was a good reason. Several commits were merged together several months ago for this MR because of a problem with conflicts with some other changes which were happening at the same time so the original issue has got lost. Perhaps I should redo it all.

70ray:
Closing while investigating the unrelated change @chrismiceli found.

tools/proofers/preview.js
chrismiceli: If you are convinced that it isn't necessary I'm not opposed to infinite loops, I just can't convince myself by looking at this without investing some more time.

PR #469, opened 2020-10-16
Add a sample phpcs.xml configuration

bpfoley:
Works out-of-box with Visual Studio Code's phpcs extension, and quells
warnings most of the code misformatting issues we have.

We can incrementally fix these up and re-enable the warnings we decide
are useful.

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -1,0 +1,77 @@
+<?xml version="1.0"?>
+<ruleset>
+    <file>./</file>
+
+    <rule ref="Generic">
+        <exclude name="Generic.PHP.ClosingPHPTag.NotFound"/>
+	    <exclude name="Generic.WhiteSpace.DisallowSpaceIndent.SpacesUsed"/>
+
+        <!-- We should decide on one of these. I vote for short only -->
+        <exclude name="Generic.Arrays.DisallowLongArraySyntax.Found"/>
+        <exclude name="Generic.Arrays.DisallowShortArraySyntax.Found"/>
+
+        <exclude name="Generic.Files.EndFileNoNewline.Found"/>
+        
+        <!-- I vote for $x = "foo", one space before, one after only-->
+        <exclude name="Generic.Formatting.MultipleStatementAlignment.NotSameWarning"/>
+        <exclude name="Generic.Formatting.MultipleStatementAlignment.IncorrectWarning"/>
+
+        
+        <!-- God, no... -->
+        <exclude name="Generic.Formatting.SpaceAfterNot.Incorrect"/>
+	
+        <!-- We've a lot of multiline strings that use these. -->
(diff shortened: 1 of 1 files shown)


bpfoley:
After spending a couple of hours setting this up, I see we also have Casey's PR #124, but it's possible some of the settings in this might be worth discussing/using.

cpeel:
Yeah, our code is all over the place. @chrismiceli got us linting the JS files which was a huge step in the right direction, but the PHP code is ... artisanal? chaos?

I don't use VSCode but am happy to pull this in if it helps you and others. Not sure I'm up for the renewed discussion of "what should our style be" right now, although I need to renew the work on https://github.com/DistributedProofreaders/dproofreaders/pull/124 I suppose.

bpfoley:
I'm very much in favor of having *some* coding style, preferably something that's enforced by automation. Although I don't always agree 100% with the decisions they make, my experiences both with [rustfmt](https://github.com/rust-lang/rustfmt) and Google's internal auto-formatting tooling have been very positive.

They almost entirely remove one whole class of low-level decision-making. You can write code with as sloppy formatting as you like, and on save/commit it gets formatted per the style guide. Once you get used to it, it's hard to go back.

cpeel:
Yup, I _love_ having `black` for my python formatting (that's what we standardize on at ExtraHop). `gofmt` too, although I don't actually do golang coding.

srjfoo:
I don't know anything about any of the code formatting tools, but if you want an opinion from a very inexperienced person, I'm very much in favor.

chrismiceli:
I am in the same boat. I am not sure what the preferred formatting tools are in php land nor how to enforce it at build time (if that is what we desire). I will vote for any as long as @cpeel is on board.

bpfoley:
Yeah, I've no idea either. I added `phpcs` simply because it seemed to be the most popular of the plugins for vscode. And it's very configurable, so we can go with whatever style we happen to like, incrementally adding/removing rules to phpcs.xml as we go.

One of the things I've left it reporting at the moment is the several files use tabs for indentation. Personally, I fall into the spaces camp in this particular holy war, simply because everybody seems to set up their tabstops differently. Also I'd like to go with denser formatting style, such as

```
	while (cond) {
		wibblewible = 1;
		teapot = 2;
		if (foo) {
			bar;
		} else {
			baz;
		}
		if (boz) {
			always_be_bracing();
		}
	}
```
rather than
```
	while (cond)
	{
		wibblewible = 1;
        teapot      = 2;
		if (foo)
		{
			bar;
		}
		else
		{
			baz;
		}
		if (boz)
			embrace_the_ambiguity();
	}
```

cpeel:
We are 100% agreed on spaces instead of tabs and I've been slowing fixing those as I find them over the past 14 years.

The dense vs not-dense is going to be something we need to hash out when we finally settle on a formatter.

70ray:
I agree with Chris.

PR #470, opened 2020-10-17
Update ppv report

srjfoo:
The initial focus of this branch was to pull inline styles for colors out of the form and set it up to use site-wide styling. I've tried to keep that part of it in the first commit. This involved changing some content, as well.

The second commit updates more content and adds/corrects links from the form to the appropriate sections in the guidelines, and adds dummy `id`s to the PP FAQ and PPV Guidelines documents in the code for use as redirects to the wiki.

To test, you'll need to grant yourself PPV access. Probably easiest if you just use [this link](https://www.pgdp.org/~srjfoo/c.branch/update-ppv-report/tools/post_proofers/ppv_report.php?project=projectID5771f9753c0c7) to go directly to a project.

### DIFF
--- a/tools/post_proofers/ppv_report.php
+++ b/tools/post_proofers/ppv_report.php
@@ -17,15 +17,9 @@ require_login();
 
 // -------------------------------------
 
-$theme_args['css_data'] = "
-input[type='text'], textarea {
-     background-color: #E2F2E1;
-}
-";
-
 $page_title = _('Post-Processing Verification Reporting');
 
-output_header($page_title, SHOW_STATSBAR, $theme_args);
+output_header($page_title);
 
 echo "\n<h1>$page_title</h1>\n";
 
@@ -103,6 +97,11 @@ mysqli_free_result($result);
 //
 // So, the date we are looking for is the latest transition to PPV.avail
 // before the earliest transition from PPV.avail to PPV.checked out...
+//
+// Note that the following queries do not take into accout the case where
+// the PPer, after having the project returned to them, returns the project
+// to the PP pool, after which it is checked out by another PPer who has
+// to submit for PPV.
 
 $pp_date = "";
 
@@ -156,7 +155,7 @@ if ($action == SHOW_BLANK_ENTRY_FORM || $action == HANDLE_ENTRY_FORM_SUBMISSION)
         global $i4, $i5;
         return ""
             . "\n$i4<tr>"
-            . "\n$i5<th colspan='2' class='$class'>$content</th>"
+            . "\n$i5<th colspan='2' class='$class center-align'>$content</th>"
             . "\n$i4</tr>";
     }
 
@@ -176,7 +175,7 @@ if ($action == SHOW_BLANK_ENTRY_FORM || $action == HANDLE_ENTRY_FORM_SUBMISSION)
         global $i4, $i5;
         return ""
             . "\n$i4<tr>"
-            . "\n$i5<td style='background-color: #CCCCCC; width: 40%;'><b>$left_content</b></td>"
+            . "\n$i5<th class='label' style='width: 25%;'><b>$left_content</b></td>"
             . "\n$i5<td>"
             . $right_content
             . "\n$i5</td>"
@@ -387,7 +386,7 @@ if ($action == SHOW_BLANK_ENTRY_FORM || $action == HANDLE_ENTRY_FORM_SUBMISSION)
 
     $entry_form = "<br>
           <form action='{$code_url}/tools/post_proofers/ppv_report.php?project=$projectid&amp;confirm=1' name='ppvform' method='post'>
-          <table class='ppv_reportcard' id='report_card'>
+          <table class='basic ppv_reportcard' id='report_card'>
 "
         . tr_w_one_cell_centered('major_section', _("Project Information"))
         . tr_w_two_cells(
--- a/tools/post_proofers/ppv_report.php
+++ b/tools/post_proofers/ppv_report.php
@@ -98,10 +98,10 @@ mysqli_free_result($result);
 // So, the date we are looking for is the latest transition to PPV.avail
 // before the earliest transition from PPV.avail to PPV.checked out...
 //
-// Note that the following queries do not take into accout the case where
-// the PPer, after having the project returned to them, returns the project
-// to the PP pool, after which it is checked out by another PPer who has
-// to submit for PPV.
+// The following queries do not take into account the case where the PPer,
+// after having the project returned to them, returns the project to the
+// PP pool, after which it is checked out by another PPer who has to
+// submit for PPV.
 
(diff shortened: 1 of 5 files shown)


srjfoo:
It is one of the oddities of this form that any PPVer can fill out the form for any project -- which is useful for collaborative purposes.

There may be places still there that should have `html_safe` or `attr_safe` added: I would appreciate help in identifying them.

faq/post_proof.php
```diff
@@ -537,6 +539,12 @@
 once the book has been posted for years.
 </p>
 
+<h3><a name="covers" id="covers"></a>Cover images</h3>
```
cpeel, line 542: We should either not have these in here (they shouldn't be needed for the redirect to work) or they should have valid content. These don't provide value to non-pgdp.net consumers. Ditto `faq/ppv.php`.

tools/post_proofers/ppv_report.php
```diff
@@ -556,13 +546,13 @@ function is_decimal_digits($s)
         )
         . tr_w_one_cell_centered('major_section', _("Copies"))
         . tr_w_two_cells(
-            _("Send to"),
+            sprintf(_("<a href='%s'>Send to"), "$ppv_guidelines_url#summary"),
```
cpeel, line 549: Missing `</a>`?

cpeel (commented):
Not sure if this is related to your changes (I suspect not) but this looks weird.

Screen Shot 2020-10-17 at 4 29 52 PM

srjfoo:
> Not sure if this is related to your changes (I suspect not) but this looks weird.
> 
> Screen Shot 2020-10-17 at 4 29 52 PM

The checked checkbox with nothing after it? That's because there's no PPer's username there, since the project that's being used has no PPer and actually isn't in PPV. The form happily accepts any valid projectID.

faq/post_proof.php
srjfoo: Would it be less confusing for non-pgdp.net consumers to add valid contend from the existing FAQs in the wiki than to remove the content altogether? Or just label any entries in the ToC as dummy entries for redirects?

cpeel: Copying over data from the wiki FAQs would be workable. I would also be fine with not having the anchors in the code FAQ files at all. None of the edits in this file are necessary for the redirects to work correctly.

srjfoo: Thinking about it a bit more, probably the latter. Especially for those I put in `faq/ppv.php`, I didn't even try to position them where they are in the actual PPV guidelines -- there are too many differences. I don't think going back to trying to keep the documents in the code up to date is a place we want to go right now.

cpeel:
> The checked checkbox with nothing after it? That's because there's no PPer's username there, since the project that's being used has no PPer and actually isn't in PPV. The form happily accepts any valid projectID.

It might be worthwhile to not output that checkbox if the label for it is empty. But this is a rare and unlikely edgecase and unrelated to your changes, so I'm fine keeping this as-is.

faq/post_proof.php
srjfoo: Superfluous sections removed.

tools/post_proofers/ppv_report.php
srjfoo: Ayup, thanks for spotting that!

faq/ppv.php
```diff
@@ -20,13 +20,25 @@
...
-<a href="#gen">A) General Procedures</a><br>
-<a href="#rr">B) Requests to Revise and Resubmit</a><br>
+<a href="#general">A) General Procedures</a><br>
+<a href="#revise">B) Requests to Revise and Resubmit</a><br>
 <a href="#checks">C) Specific Checks to Make</a><br>
 <a href="#printers">D) Printers' Errors and Transcriber's Note</a><br>
 <a href="#reader">E) Checking E-reader versions</a><br>
 <a href="#rptcard">F) Submitting a PPV Summary</a><br>
 <a href="#late">G) Late PPV Summaries</a>
+/* The following are dummy IDs to provide a way to redirect to the
```
cpeel, line 30: You don't need any of these.

faq/post_proof.php
```diff
@@ -44,6 +44,7 @@
         <li><a href="#markup">Check the markup</a></li>
         <li><a href="#title">Straighten up the title page, table of contents,
             and list of illustrations</a></li>
+        <li><a href="#covers">Cover image</a></li> // dummy ID to provide path to wiki FAQ
```
cpeel, line 47: You don't need any of the changes in this file.

faq/ppv.php
srjfoo: Ohhhh ... I misunderstood, then.

srjfoo:
> > The checked checkbox with nothing after it? That's because there's no PPer's username there, since the project that's being used has no PPer and actually isn't in PPV. The form happily accepts any valid projectID.
> 
> It might be worthwhile to not output that checkbox if the label for it is empty. But this is a rare and unlikely edgecase and unrelated to your changes, so I'm fine keeping this as-is.

I'm okay with doing whatever, but yeah, an edge case, and most likely not to crop up except in testing.

faq/ppv.php
```diff
@@ -20,8 +20,8 @@
 <h2>Contents</h2>
 
 <p>
-<a href="#gen">A) General Procedures</a><br>
-<a href="#rr">B) Requests to Revise and Resubmit</a><br>
+<a href="#general">A) General Procedures</a><br>
+<a href="#revise">B) Requests to Revise and Resubmit</a><br>
```
chrismiceli, line 24: do we have links to these anywhere? Wiki? Would we expect people to have bookmarked these?

chrismiceli (approved):

faq/ppv.php
srjfoo: The `id`s have been added in the wiki documents. @cpeel recommended removing the dummy links that I added to the documents in the code -- which I have done, but just hadn't gotten the squashed commit pushed yet. Just about to do that.

srjfoo: Actually, those stayed, because I renamed existing IDs for more clarity. Looks like I took out an extra space in `faq/post_proof.php`, which winds up being the only change that persisted. Some days. <sigh>

`faq/post_proof.php` and `faq/ppv.php` are the old PP and PPV FAQ/Guideline documents, and are way out of date. They've been useful as a means to redirect to wiki content.

faq/ppv.php
srjfoo: All changes from both `faq/ppv.php` and `faq/post_proof.php` were unnecessary and have been reverted, and the commit they were in squashed. I didn have to change one link in `tools/post_proofers/ppv_report.php` to match the reverted link.

faq/post_proof.php
srjfoo: Fixed.

PR #471, opened 2020-10-18
editproject.php: Use modern DPDatabase::query, escaping, sprintf

bpfoley:


tools/project_manager/editproject.php
```diff
@@ -688,18 +691,20 @@ function save_to_db()
             }
 
             // Update the projects database with the updated info
-            mysqli_query(DPDatabase::get_connection(), "
+            $sql = sprintf("
                 UPDATE projects SET
                     $pm_setter
                     $tlcc_setter
                     $md_setter
                     $common_project_settings
```
cpeel, line 699: There is no guarantee that the contents of `$common_project_settings` doesn't contain a `%`, so we can't use `sprintf()` here like this. I really hate how we build queries up like this in so many places.

You could do:
```php
            $where = sprintf("projectid='%s'", DPDatabase::escape($this->projectid));
            $sql = "
                UPDATE projects SET
                    $pm_setter
                    $tlcc_setter
                    $md_setter
                    $common_project_settings
                WHERE $where";
```

tools/project_manager/editproject.php
```diff
@@ -742,23 +747,25 @@ function save_to_db()
...
 
             // Insert a new row into the projects table
-            mysqli_query(DPDatabase::get_connection(), "
+            $sql = sprintf("
                 INSERT INTO projects
                 SET
-                    projectid    = '{$this->projectid}',
+                    projectid    = '%s',
                     $pm_setter
                     state        = '".PROJ_NEW."',
```
cpeel, line 755: For consistency, we should move the state into the `sprintf()` replace too.

tools/project_manager/editproject.php
bpfoley: Oh that's hideous. Yeah, I'll keep that in mind in future.

tools/project_manager/editproject.php
bpfoley: PHP! :(

PR #473, opened 2020-10-19
Preview font fix

70ray:
This fixes two bugs which appeared after the change to font selection in format preview:
The fonts in the selector were duplicated each time the configuration was changed - now the selector is only set up once at start.
The example text in the configuration now uses the selected font.
Some comments have been deleted which referred to the previous way of selecting preview fonts.


chrismiceli (approved):

srjfoo (approved):
Works for me!

PR #474, opened 2020-10-20
RandomRule.inc: Use modern DPDatabase::query, escaping, sprintf

bpfoley:


pinc/RandomRule.inc
```diff
@@ -41,9 +41,8 @@ class RandomRule
             WHERE id = %d
         ", $id);
 
-        $result = mysqli_query(DPDatabase::get_connection(), $sql);
-        if(!$result)
-        {
+        $result = DPDatabase::query($sql, FALSE);
```
chrismiceli, line 44: why FALSE on the throwing error?

pinc/RandomRule.inc
```diff
@@ -58,14 +57,14 @@ class RandomRule
...
+                langcode = '%s'",
+            DPDatabase::escape($document),
+            DPDatabase::escape($anchor),
+            DPDatabase::escape($langcode)
+        );
 
-        $result = mysqli_query(DPDatabase::get_connection(), $sql);
-        if(!$result)
-        {
+        $result = DPDatabase::query($sql, FALSE);
```
chrismiceli, line 66: same here, we can remove the false and the throw below probably I would imagine.

pinc/RandomRule.inc
bpfoley: Because I wanted to preserve the type of exception thrown in case it was significant. This one throws a UnexpectedValueException, whereas DPDatabase::query throws a DBQueryError.

pinc/RandomRule.inc
bpfoley: Same as above: UnexpectedValueException vs DBQueryError

pinc/RandomRule.inc
bpfoley: If the exception type isn't significant (and it looks like it isn't because there are only 2 (unrelated) places in the codebase that `catch(UnexpectedValueException $e)`), I agree we could leave it up to ::query to throw the exception.

cpeel: The `UnexpectedValueException` is wrong too (and I wrote it -- blame me). The intent was probably to indicate that a value someone passed in was wrong -- but if that were actually the case the SQL would return zero rows so `$result` would still be valid. I would remove the `$result` test and let `DBDatabase::query()` raise.

bpfoley: Done!

cpeel:
Tested and confirmed that this works.

PR #475, opened 2020-10-20
Project.inc: Use modern DPDatabase::query, escaping, sprintf

bpfoley:
Tested by browsing a bunch of project-related pages on pgdp.org

### DIFF
--- a/pinc/Project.inc
+++ b/pinc/Project.inc
@@ -31,9 +31,10 @@ class Project
             $sql = sprintf("
                 SELECT *
                 FROM projects
-                WHERE projectid = '%s'
-            ", mysqli_real_escape_string(DPDatabase::get_connection(), $arg));
-            $res = mysqli_query(DPDatabase::get_connection(), $sql) or die(DPDatabase::log_error());
+                WHERE projectid = '%s'",
+                DPDatabase::escape($arg)
+            );
+            $res = DPDatabase::query($sql);
             $row = mysqli_fetch_assoc($res);
             if (!$row)
             {
@@ -119,11 +120,14 @@ class Project
             return;
         }
 
-        $imso_res = mysqli_fetch_assoc(mysqli_query(DPDatabase::get_connection(), "
+        $sql = sprintf("
             SELECT full_name, credit
             FROM image_sources
-            WHERE code_name = '$this->image_source'
-        "));
+            WHERE code_name = '%s'",
+            DPDatabase::escape($this->image_source)
+        );
+        $result = DPDatabase::query($sql);
+        $imso_res = mysqli_fetch_assoc($result);
         $this->image_source_name = $imso_res['full_name'];
         $this->image_source_credit = $imso_res['credit'];
     }
@@ -163,11 +167,11 @@ class Project
         {
             return false;
         }
-
+        validate_projectID($this->projectid);
         $sql = "
             ALTER TABLE $this->projectid CONVERT TO CHARACTER SET utf8mb4;
         ";
-        $result = mysqli_query(DPDatabase::get_connection(), $sql);
+        $result = DPDatabase::query($sql, False);
         if(!$result)
         {
             throw new UTF8ConversionException(
@@ -187,11 +191,12 @@ class Project
         $sql = sprintf("
             SELECT projectid
             FROM project_charsuites
-            WHERE charsuite_name='%s'
-        ", mysqli_real_escape_string(DPDatabase::get_connection(), $charsuite->name));
+            WHERE charsuite_name='%s'",
+            DPDatabase::escape($charsuite->name)
+        );
 
         $projectids = [];
-        $result = mysqli_query(DPDatabase::get_connection(), $sql);
+        $result = DPDatabase::query($sql);
         while($row = mysqli_fetch_assoc($result))
         {
             $projectids[] = $row['projectid'];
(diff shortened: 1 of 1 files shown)


pinc/Project.inc
```diff
@@ -534,11 +541,12 @@ class Project
 
         $projectid = $this->projectid;
         $page_image_names = array();
-        $res = mysqli_query(DPDatabase::get_connection(), "
+        $sql = "
```
cpeel, line 544: Somewhat frustratingly, we can't rely on `$this->projectid` to be a valid project id. There are object construction paths that could mean it is anything. To ensure this is a valid project name, and not an injection attack, you can put a `validate_projectID($projectid);` before building the SQL.

pinc/Project.inc
```diff
@@ -777,24 +787,30 @@ EOS;
             $values[] = "('$this->projectid', '$state')";
```
cpeel, line 787: We should escape these values as well.

pinc/Project.inc
```diff
@@ -777,24 +787,30 @@ EOS;
...
     {
         global $pguser;
 
-        $states_str = surround_and_join( $states, "'", "'", ", " );
-        mysqli_query(DPDatabase::get_connection(), "
+        $projectid_clause = sprintf(
+            "projectid='%s'",
+            DPDatabase::escape($this->projectid)
+        );
+        $states_str = surround_and_join($states, "'", "'", ", ");
```
cpeel, line 806: `surround_and_join()` doesn't escape any of the states. It would probably be easiest if we built this the same way `add_holds()` does.

pinc/Project.inc
bpfoley: OK. It feels like I've had to scatter a bunch of validate_projectIDs around the place, which won't be great for performance. That said, they're all near DB queries, so I'm sure it's lost in the noise.

I'm not sure how `class Project` is designed, but... ordinarily wouldn't we solve this by making the validity of projectid a class invariant that's enforced by allowing only the constructor (and specific methods) to modify (and validate) $this->projectid, and making $projectid private?

pinc/Project.inc
bpfoley: Done.

pinc/Project.inc
bpfoley: Yeah, it's better for 'symmetry' between the two implementations too. Done.

pinc/Project.inc
```diff
@@ -252,13 +257,14 @@ class Project
                 $charsuite->title
             ));
         }
-
+        validate_projectID($this->projectid);
```
cpeel, line 260: You don't need this because `$this->projectid` is not being used as a table, but as a field you are already escaping.

pinc/Project.inc
```diff
@@ -270,13 +276,14 @@ class Project
         {
             return;
         }
-
+        validate_projectID($this->projectid);
```
cpeel, line 279: Ditto - don't need here.

pinc/Project.inc
```diff
@@ -294,13 +301,14 @@ class Project
         }
 
         // load regular character suites
-        $sql = "
+        validate_projectID($this->projectid);
```
cpeel, line 304: Ditto - don't need here.

pinc/Project.inc
```diff
@@ -718,11 +728,15 @@ EOS;
         }
 
         // Save $topic_id in db and in $this.
-        $update_project = mysqli_query(DPDatabase::get_connection(), "
+        validate_projectID($this->projectid);
```
cpeel, line 731: Ditto - don't need here.

pinc/Project.inc
```diff
@@ -736,15 +750,15 @@ EOS;
     {
         global $site_name, $code_url;
 
-        $projectid = $this->projectid;
+        validate_projectID($this->projectid);
```
cpeel, line 753: Not sure why this one is here at all.

pinc/Project.inc
```diff
@@ -755,11 +769,14 @@ EOS;
     // Return an array containing the states for which this project has holds.
     {
         $hold_states = array();
-        $res = mysqli_query(DPDatabase::get_connection(), "
+        validate_projectID($this->projectid);
```
cpeel, line 772: Ditto - don't need here.

pinc/Project.inc
```diff
@@ -866,37 +896,45 @@ XML;
     {
         $original_marc_array = $marc_record->get_yaz_array();
 
-        mysqli_query(DPDatabase::get_connection(), "
+        validate_projectID($this->project_id);
```
cpeel, line 899: Ditto - don't need here.

pinc/Project.inc
cpeel: Well it's because you did it wrong. We only need them in places where it is then used as a table name in an SQL statement.

cpeel: https://www.pgdp.net/wiki/DP_Code_Best_Practices#Validating_project_tables

bpfoley: Ah, I see. Removed as requested.

pinc/Project.inc
```diff
@@ -866,37 +891,44 @@ XML;
...
+        );
     }
 
     // Update the updated_* columns for this project's MARC record
     public function save_marc_record($marc_record)
     {
         $updated_marc_array = $marc_record->get_yaz_array();
-
-        mysqli_query(DPDatabase::get_connection(), "
+        validate_projectID($this->project_id);
```
cpeel, line 908: You don't need this.

pinc/Project.inc
```diff
@@ -866,37 +891,44 @@ XML;
...
+        DPDatabase::query($sql);
     }
 
     // Load the updated MARC record for this project
     public function load_marc_record()
     {
         $updated_record = new MARCRecord();
-
-        $result = mysqli_query(DPDatabase::get_connection(), "
+        validate_projectID($this->project_id);
```
cpeel, line 924: Ditto.

pinc/Project.inc
```diff
@@ -991,11 +1023,15 @@ XML;
...
 {
-    $res = mysqli_query(DPDatabase::get_connection(), "
+    $sql = sprintf("
         SELECT *
         FROM project_holds
-        WHERE projectid='$projectid' and state='$state'
-    ") or die(DPDatabase::log_error());
+        WHERE projectid='%s'
+            AND state='%s'",
+            DPDatabase::escape($projectid),
```
cpeel, line 1031: Too much indent here and line below.

PR #476, opened 2020-10-21
Misc preview

70ray:
This re-writes some code in preview.js to remove eslint exceptions.

Also adds some padding to the example texts in format preview configure.

Can be seen in https://www.pgdp.org/~rp31/c.branch/misc_preview Only the padding change should have any visible effect.

### DIFF
--- a/styles/preview.css
+++ b/styles/preview.css
@@ -47,6 +47,10 @@
     text-align: left;
 }
 
+#color_test {
+    padding: 0.15em;
+}
+
 /* in-line styles */ 
 .i {
     font-style: italic;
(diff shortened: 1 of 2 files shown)


srjfoo (approved):
I'm not qualified to judge on the eslint changes, but the padding change looks good.

The reason for the addition of the padding is that where the sample text showing what each kind of markup looks like with the user's current configurations is displayed, depending on the window width and background color, some of the samples actually appeared to look like tabs:

image

With the padding, they clearly are not:

image

styles/preview.css
```diff
@@ -47,6 +47,10 @@
     text-align: left;
 }
 
+#color_test {
```
chrismiceli, line 50: should we move this to less one day?

chrismiceli (approved):

styles/preview.css
cpeel: @srjfoo and I were chatting about this too. At some point some of this should move into `page_interface.less`, but perhaps not all of it.

PR #478, opened 2020-10-22
stats/new_team: Use DPDatabase::query, more consistent arg escaping

bpfoley:
And make the extraction (and cleanup) of the args from @POST more
consistent, so that what's checked against and displayed and insterted
into the DB are all the same.

### DIFF
--- a/stats/teams/new_team.php
+++ b/stats/teams/new_team.php
@@ -12,80 +12,98 @@ $user = User::load_current();
 
 $theme_extra_args = array("js_data" => get_newHelpWin_javascript("$code_url/pophelp.php?category=teams&name=edit_"));
 
+$teamname = stripAllString(trim(array_get($_POST, "teamname", "")));
+$text_data = stripAllString(array_get($_POST, "text_data", ""));
+$teamwebpage = stripAllString(array_get($_POST, "teamwebpage", ""));
+$tavatar = array_get($_POST, "tavatar", "");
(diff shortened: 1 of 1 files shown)


stats/teams/new_team.php
```diff
@@ -35,12 +35,19 @@
 }
 else if (isset($_POST['mkMake']))
 {
-    $result = mysqli_query(DPDatabase::get_connection(), sprintf("
+    $teamname = stripAllString(trim($_POST['teamname']));
```
cpeel, line 38: If we want to futz with this (and we should -- the team code is horrific) we should move these to the top of the file and always populate them instead of pulling them in various places. We also need to use `array_get()` to pull the variables because your code is generating a notice as-is for lines 41 & 42 when an avatar and icon are not provided.

bpfoley: Hopefully this should be better... I don't love the way we accept and silently trim spaces, and strip a subset of markup and escape the rest. Do you think it might be better if we just drop the stripAllString, and validate each piece of input: disallowing any leading or trailing space in the team names, and refusing < > and & in teamname, text_data, teamwebpage?

cpeel: This is better, thanks.

The team code makes me want to claw my eyes out. Per-item validation would be better, as would be dropping our custom formatting (really: why?). But none of this is important enough to work on right now IMHO.

PR #480, opened 2020-10-22
GM requested updates to new user welcome email

srjfoo:
GM requested revisions to the Welcome email to new users. Can be tested at [revise-welcome-email](https://www.pgdp.org/~srjfoo/c.branch/revise-welcome-email).

### DIFF
--- a/pinc/new_user_mails.inc
+++ b/pinc/new_user_mails.inc
@@ -50,7 +50,7 @@ EOMAIL;
 
 function maybe_welcome_mail($email, $real_name, $username)
 {
-    global $code_url, $site_url, $forums_url, $wiki_url;
+    global $code_url, $site_url, $forums_url;
     global $site_name, $site_abbreviation, $ELR_round;
     global $general_help_email_addr;
 
@@ -62,7 +62,7 @@ function maybe_welcome_mail($email, $real_name, $username)
     // Much of the contents of the email below is DP site-specific, however
     // not so much that it would be horrible for non-DP sites to use it without
     // customizing it.
-    $dp_wiki_url = "http://www.pgdp.net/wiki";
+    $dp_wiki_url = "https://www.pgdp.net/wiki";
 
     // only include the feedback paragraph if we are running at pgdp.*
     if(stripos($code_url, '//www.pgdp.') !== FALSE)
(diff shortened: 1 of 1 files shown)


srjfoo:
Of note -- yes, these are not currently localizable. That's on my to-do list for some other time.

Test by creating a new account to generate the content of the email. There is no `dp-feedback` user on TEST, so that won't generate a usable link, but I've checked all the others and they go where they're supposed to.

chrismiceli (approved):

cpeel:
This LGTM, but the length of the intro email is, frankly, ridiculous. The longer the email the less likely anyone is going to read any of it. If I received this email today I would probably ignore it. We should be striving to make the new user flow easy enough that we can reduce the length of this email by at least an order of magnitude.

srjfoo:
> This LGTM, but the length of the intro email is, frankly, ridiculous. The longer the email the less likely anyone is going to read any of it. If I received this email today I would probably ignore it. We should be striving to make the new user flow easy enough that we can reduce the length of this email by at least an order of magnitude.

I agree. The longer the email, and the longer the paragraphs within the email, the less likely it is to be read. @lhamilton1 and I have discussed it, and her goal is to make sure that the information is provided so that we have something to point to.

cpeel:
So this is the new "small print legalease" we have to CYA?

srjfoo:
> So this is the new "small print legalease" we have to CYA?

That's my understanding, yes.

cpeel:
Basically, if this is our new user experience we're doing it wrong.

lhamilton1:
This was basically the content the forums agreed to a while back but that had incorrect info in it since our SR upgrade (and the mentors were eager to add a sentence about the fact that we don't change what the author wrote). It is long, but then the detail-oriented type of volunteer we keep tends to be the sort that may be more apt to read longer emails. We did include a link so they can read it in the DPOD or come back to it later. I'd rather we updated it for now so it's not incorrect and went back to the forums to work out a better way.

cpeel:
> It is long, but then the detail-oriented type of volunteer we keep tends to be the sort that may be more apt to read longer emails.

It's so long as to be inaccessible.

This is not the hill I want to die on and will get it in, but this is really bad.

lhamilton1:
I promise we'll bring it up for change in the forums before long.

PR #481, opened 2020-10-23
Create and use a func for loading a task

cpeel:
Create a function for loading a task instead of sprinkling that logic around. This also adds a migration script to correctly treat task_id as a primary key. This is a stepping stone to updating tasks to use proper escaping and the new DPDatabase functions.

Note that due to the 2-level de-indenting in `TaskDetails()` there are a lot of changes to that function that are mostly just whitespace changes. Consider adjusting the view to hide whitespace changes when reviewing it.

Testable in the [consolidate-task-queries](https://www.pgdp.org/~cpeel/c.branch/consolidate-task-queries/tasks.php) sandbox.

### DIFF
--- a/SETUP/db_schema.sql
+++ b/SETUP/db_schema.sql
@@ -552,7 +552,7 @@ CREATE TABLE `tasks` (
   `edited_by` mediumint(9) NOT NULL default '0',
   `percent_complete` tinyint(3) NOT NULL default '0',
   `related_postings` mediumtext NOT NULL,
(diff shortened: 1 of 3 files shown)


srjfoo (approved):

bpfoley:
LGTM! The de-duplication is certainly a big improvement.

bpfoley (approved):

70ray (approved):

PR #486, opened 2020-10-24
site_admin/copy_pages.php: Use modern DPDatabase::query, escape, sprintf

bpfoley:
Interestingly, several cases here were incorrectly escaped: they
were already validated projectIDs, and are used as table names.

### DIFF
--- a/tools/site_admin/copy_pages.php
+++ b/tools/site_admin/copy_pages.php
@@ -241,10 +241,8 @@ function do_stuff( $projectid_, $from_image_, $page_name_handling,
             die("Project table $projectid is not UTF-8.");
         }
 
-        $res= mysqli_query(DPDatabase::get_connection(),
-            sprintf("DESCRIBE %s",
-            mysqli_real_escape_string(DPDatabase::get_connection(), $projectid_[$which]))
-        ) or die(DPDatabase::log_error());
+        $sql = "DESCRIBE {$projectid_[$which]}";
+        $res = DPDatabase::query($sql);
 
         $column_names = array();
         while ( $row = mysqli_fetch_assoc($res) )
@@ -282,12 +280,11 @@ function do_stuff( $projectid_, $from_image_, $page_name_handling,
 
         // ----------------------
 
-        $res = mysqli_query(DPDatabase::get_connection(), sprintf("
+        $sql = "
             SELECT image, fileid
-            FROM %s
-            ORDER BY image",
-            mysqli_real_escape_string(DPDatabase::get_connection(), $projectid)
-        )) or die(DPDatabase::log_error());
+            FROM $projectid
+            ORDER BY image";
+        $res = DPDatabase::query($sql);
 
         $n_pages = mysqli_num_rows($res);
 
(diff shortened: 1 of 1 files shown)


bpfoley:
I've tested copying pages from one project to another, and the sandbox is at https://www.pgdp.org/~bfoley/c.branch/cleanup12

tools/site_admin/copy_pages.php
```diff
@@ -651,13 +658,14 @@ function do_stuff( $projectid_, $from_image_, $page_name_handling,
...
         {
             $query = sprintf("
-                      SELECT username FROM user_project_info
-                      WHERE projectid = '%s' AND
-                            iste_$event = 1",
-                    mysqli_real_escape_string(DPDatabase::get_connection(), $projectid_['from'])
+                SELECT username FROM user_project_info
+                WHERE projectid = '%s' AND
+                    iste_%s = 1",
+                DPDatabase::escape($projectid_['from']),
```
chrismiceli, line 664: didn't you say above this was validated already?

bpfoley: Good catch. Yes I did!

tools/site_admin/copy_pages.php
```diff
@@ -282,12 +282,13 @@ function do_stuff( $projectid_, $from_image_, $page_name_handling,
 
         // ----------------------
 
-        $res = mysqli_query(DPDatabase::get_connection(), sprintf("
+        $sql = sprintf("
             SELECT image, fileid
             FROM %s
```
cpeel, line 287: I would embed the `$projectid` directly in the string and drop the `sprintf()`. That way the sprintf is reserved for escaping values and not the table or column names. https://www.pgdp.net/w/index.php?title=DP_Code_Best_Practices#Column_names_as_variables

cpeel: Ditto other places where we build SQL in this file.

bpfoley: Done.

tools/site_admin/copy_pages.php
```diff
@@ -596,25 +593,31 @@ function do_stuff( $projectid_, $from_image_, $page_name_handling,
...
+                'image',
+                'fileid'
+            ],
+            [
+                sprintf("'%s'", DPDatabase::escape($c_dst_image)),
+                sprintf("'%s'", DPDatabase::escape($c_dst_fileid))
+            ],
+            $items_list_template);
+
+        $insert = "INSERT INTO {$projectid_['to']}
```
cpeel, line 606: Move the `INSERT INTO` onto the next line.

tools/site_admin/copy_pages.php
```diff
@@ -651,13 +654,14 @@ function do_stuff( $projectid_, $from_image_, $page_name_handling,
         foreach ( $subscribable_project_events as $event => $label )
         {
             $query = sprintf("
-                      SELECT username FROM user_project_info
-                      WHERE projectid = '%s' AND
-                            iste_$event = 1",
-                    mysqli_real_escape_string(DPDatabase::get_connection(), $projectid_['from'])
+                SELECT username FROM user_project_info
+                WHERE projectid = '%s' AND
+                    iste_%s = 1",
```
cpeel, line 659: Keep `$event` in the string since it is a column name. We aren't escaping it or anything anyway.

srjfoo:
I've also tested and didn't see any problems.

PR #487, opened 2020-10-25
forum_interface.inc: Use modern DPDatabase::query, escaping, sprintf

bpfoley:
Again, a few prefixes to table names were escaped when they shouldn't be.

Tested on pgdp.org and sandbox available at https://www.pgdp.org/~bfoley/c.branch/cleanup13

### DIFF
--- a/pinc/forum_interface.inc
+++ b/pinc/forum_interface.inc
@@ -38,6 +38,15 @@ the amount of DP code included in those pages.
 
 assert(PHPBB_VERSION == 3, "Only phpBB version 3 is supported");
 
+
+/**
+* return a table name with the phpBB table prefix prepended to it
+*/
+function phpbb_table($table)
+{
+    return PHPBB_TABLE_PREFIX . "_" . $table;
+}
+
 function phpbb_lang($locale=FALSE)
 // Return the language string phpbb uses for $langcode for languages
 // installed in the forum interface. Fall back to English if $langcode
@@ -52,17 +61,17 @@ function phpbb_lang($locale=FALSE)
         // We can only do so much, so try for an exact match in lowercase
         // first, fall back to just the two-letter language code, and then
         // just fall back to English (which phpBB comes with by default).
-        $phpbb_table_prefix = PHPBB_TABLE_PREFIX;
         $lang_attempts = array(strtolower($locale), substr($locale, 0, 2));
         foreach($lang_attempts as $lang)
         {
             $query = sprintf("
                 SELECT *
-                FROM ${phpbb_table_prefix}_lang
-                WHERE lang_iso='%s'
-                ", mysqli_real_escape_string(DPDatabase::get_connection(), $lang));
-            $res = mysqli_query(DPDatabase::get_connection(), $query) or
-                die(DPDatabase::log_error());
+                FROM %s
+                WHERE lang_iso='%s'",
+                phpbb_table("lang"),
+                DPDatabase::escape($lang)
+            );
+            $res = DPDatabase::query($query);
 
             if (mysqli_num_rows($res) == 1)
                 return $lang;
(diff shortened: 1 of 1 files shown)


pinc/forum_interface.inc
```diff
@@ -335,12 +341,14 @@ function get_forum_user_details($username)
...
-        ", $row["user_id"]);
+        FROM %s_profile_fields_data
+        WHERE user_id = %d",
+        $phpbb_table_prefix,
+        $row["user_id"]
+    );
 
     // This will fail on phpBB < 3.1, which is fine and expected
-    $res = mysqli_query(DPDatabase::get_connection(), $query);
+    $res = DPDatabase::query($query, FALSE);
```
chrismiceli, line 351: why FALSE?

pinc/forum_interface.inc
```diff
@@ -533,7 +545,7 @@ function get_number_of_unread_messages($username)
             ";
     }
 
-    $res = mysqli_query(DPDatabase::get_connection(), $query) or die(DPDatabase::log_error());
```
chrismiceli, line 536: should we update the query on line 541 now too to use sprintf? and %d?

pinc/forum_interface.inc
```diff
@@ -636,7 +652,7 @@ function get_topic_details($topic_id)
         INNER JOIN $phpbb_users ON $phpbb_topics.topic_poster = $phpbb_users.user_id
         WHERE $phpbb_topics.topic_id = $topic_id
```
chrismiceli, line 653: sprintf?

pinc/forum_interface.inc
bpfoley: I'm basing it on the comment above ("this will fail on phpBB <3.1"). It appears the column names are different in older versions of phpBB, so the query will fail, and we don't want this to throw an exception.

pinc/forum_interface.inc
bpfoley: Sure, I can do this, but it makes for a very difficult to read
format string. Given `phpbb_topics`, `phpbb_forums`, `php_users` are local variables, I'm not sure if it's worth it. What do you think?

pinc/forum_interface.inc
```diff
@@ -58,11 +58,12 @@ function phpbb_lang($locale=FALSE)
         {
             $query = sprintf("
                 SELECT *
-                FROM ${phpbb_table_prefix}_lang
-                WHERE lang_iso='%s'
-                ", mysqli_real_escape_string(DPDatabase::get_connection(), $lang));
-            $res = mysqli_query(DPDatabase::get_connection(), $query) or
-                die(DPDatabase::log_error());
+                FROM %s_lang
```
cpeel, line 61: Keep this directly in the string since it's a table name.

cpeel: Ditto everywhere else.

bpfoley: Hopefully phpbb_table(...) helps a lot.

pinc/forum_interface.inc
```diff
@@ -634,9 +640,12 @@ function get_topic_details($topic_id)
         FROM $phpbb_topics
         INNER JOIN $phpbb_forums ON $phpbb_topics.forum_id = $phpbb_forums.forum_id
         INNER JOIN $phpbb_users ON $phpbb_topics.topic_poster = $phpbb_users.user_id
-        WHERE $phpbb_topics.topic_id = $topic_id
-        ";
-    $res = mysqli_query(DPDatabase::get_connection(), $query) or die(DPDatabase::log_error());
+        WHERE $phpbb_topics.topic_id =";
```
cpeel, line 643: Instead, please use:
```php
WHERE $phpbb_topics.topic_id = " . sprintf("%d", $topic_id);
```

pinc/forum_interface.inc
cpeel: This is fine, but I'm not sold that it's better than what was there before. Our best practices keeps the table names inside the string itself so the SQL is more readable to humans who care a lot more about the column and table names (the structure of the query) rather than the values: https://www.pgdp.net/wiki/DP_Code_Best_Practices#Validating_project_tables

pinc/forum_interface.inc
bpfoley: Done.

PR #489, opened 2020-10-27
background-colors: standardize bg color names

srjfoo:
This commit standardizes the common background color names, more or less (there are still some oddballs), and tries to use names that will give some indication of what they should be used for.

Also included, an adjustment to the defined general purpose tables: for the striped tables, odd and even stripe backgrounds were flipped so that the even stripe uses the page background and the odd uses the stripe color. All striped tables now have a header row background color. This should not affect the round
listings.

Test in [background-colors](https://www.pgdp.org/~srjfoo/c.branch/background-colors/) sandbox.

### DIFF
--- a/styles/layout.less
+++ b/styles/layout.less
@@ -1,6 +1,30 @@
 @import "global.less";
 @import "page_interfaces.less";
 
+
+/* ------------------------------------------------------------------------ */
+/* Common color definitions
+   Where a particular color family has multiple shades available, a
+   suffix indicates relative contrast as compared with the default
+   @page-background color. */
+
+/* Shades based on @page-background. Various contrast levels, commonly
+   used for table rows or other containers that should stand out from
+   the page background */
+@table-cell-base-highc: #cccccc;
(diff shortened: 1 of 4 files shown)


srjfoo:
Most of the changes should be invisible. The ones that spring to mind immediately are the red background in the Page Detail table for pages that the user has done in a previous round, and that will be visible only if a user works on a project in multiple rounds (it's significantly lighter now, which makes the text in the cell somewhat easier to read). The previously extremely light grey background on the popup for the Actions column in project search results is now white.

And of course the tables. Compare [background-colors/styles/style_demo.php](https://www.pgdp.org/~srjfoo/c.branch/background-colors/styles/style_demo.php) with [c/styles/style_demo.php](https://www.pgdp.org/c/styles/style_demo.php)

chrismiceli (approved):

cpeel:
This is really great work, Sharon. Thanks for corralling the colors into something sensible!

PR #491, opened 2020-10-27
Project.inc: Fix bug introduced by daa207bd

bpfoley:
does_project_page_table_exist would throw an exception if the
project didn't exist.

Found by srjfoo.


srjfoo (approved):
Fixes the problem.

PR #492, opened 2020-10-28
Preview adjust

70ray:
This does the check for invalid characters before entering preview and removes the check for tab character in preview.

Also improves layout of the configuration panel so the color selection is less confusing (at least to me) and the other options have a more uniform style. The changes made the indenting of the html wrong so I re-indented it all. This makes it seem much more different than it really is. It perhaps would be better to remove all the html indenting.

Can be seen in sandbox https://www.pgdp.org/~rp31/c.branch/preview_adjust

### DIFF
--- a/scripts/text_validator.js
+++ b/scripts/text_validator.js
@@ -46,6 +46,10 @@ $(function () {
     }
 
     function _validateText() {
+        // IE HACK - xregexp causes problems in IE11, return true if IE
+        if(document.documentMode) {
+            return true;
+        }
         var text = textArea.value;
         // IE HACK - IE11 does not support string normalization
         if(String.prototype.normalize) {
--- a/scripts/text_validator.js
+++ b/scripts/text_validator.js
@@ -46,10 +46,6 @@ $(function () {
     }
 
     function _validateText() {
-        // IE HACK - xregexp causes problems in IE11, return true if IE
-        if(document.documentMode) {
-            return true;
-        }
         var text = textArea.value;
         // IE HACK - IE11 does not support string normalization
         if(String.prototype.normalize) {
--- a/SETUP/tests/jsTests/characterValidation.js
+++ b/SETUP/tests/jsTests/characterValidation.js
@@ -51,7 +51,6 @@ QUnit.module("Character validation test", function() {
     // (which IE doesn't do) in regex constructor in character_test.js
     /* QUnit.test("Astral plane", function (assert) {
         validCharRegex = makeValidCharRegex("^(?:[\n\r 𓀀-𓀂])$");
-        validCharRegex = makeValidCharRegex("[\n\r 𓀀𓀁𓀂]");
         assert.strictEqual(testText("𓀀"), true);
         assert.strictEqual(testText("\u{13000}"), true);
         assert.strictEqual(testText("𓀂"), true);
--- a/SETUP/tests/jsTests/formatPreviewTests.js
+++ b/SETUP/tests/jsTests/formatPreviewTests.js
@@ -112,12 +112,6 @@ QUnit.module("Format preview test", function() {
         noIssueTest(assert);
     });
 
-    QUnit.test("tabulate character", function (assert) {
-        text = "ab\tcd";
-        issArray = analyse(text, configuration);
-        issueTest(assert, 0, 2, 1, "tabChar", 1);
-    });
-
     QUnit.test("character before footnote", function (assert) {
         text = "xy[1]\n\nz[Footnote 1: ab]";
         issArray = analyse(text, configuration);
--- a/tools/proofers/preview.inc
+++ b/tools/proofers/preview.inc
@@ -2,18 +2,6 @@
 $relPath="./../../pinc/";
 include_once($relPath.'misc.inc');
 
-function get_style_strings()
-{
-    return array(
-        "plain" => _("Plain text"),
-        "italic" => _("Italic"),
-        "bold" => _("Bold"),
-        "gesperrt" => _("Gesperrt"),
-        "small_caps" => _("Small caps"),
-        "font_change" => _("Font change"),
-    );
-}
-
 function get_preview_messages()
 {
     return array(
@@ -62,13 +50,32 @@ function get_preview_js_vars()
 {
     $superscript = _("superscript");
     $subscript = _("subscript");
-    $styles = get_style_strings();
 
-    $demo_string = javascript_safe("{$styles['plain']} _{{$subscript}} <i>{$styles['italic']}</i> ^{{$superscript}} <b>{$styles['bold']}</b> <g>{$styles['gesperrt']}</g> <sc>{$styles['small_caps']}</sc> <f>{$styles['font_change']}</f>");
-    $ie_warn = javascript_safe(_("This function is not available for Internet Explorer versions less than 9"));
+    $tag_names = [
+        "t" => _("Plain text"),
+        "i" => _("Italic"),
+        "b" => _("Bold"),
+        "g" => _("Gesperrt"),
+        "sc" => _("Small caps"),
+        "f" => _("Font change"),
+        "etc" => _("Other tags"),
+        "err" => _("Issues"),
+        "hlt" => _("Poss. Issues")
+    ];
+
+    $demo_string = "{$tag_names['t']} _{{$subscript}} <i>{$tag_names['i']}</i> ^{{$superscript}} <b>{$tag_names['b']}</b> <g>{$tag_names['g']}</g> <sc>{$tag_names['sc']}</sc> <f>{$tag_names['f']}</f>";
+
+    $preview_strings = json_encode([
+        "background" => _("Background"),
+        "text" => _("Text"),
+        "previewDemo" => $demo_string,
+        "ieWarn" => _("This function is not available for Internet Explorer versions less than 9"),
+    ]);
+
     $preview_msg = json_encode(get_preview_messages());
+    $tag_names = json_encode($tag_names);
 
-    return "var previewMessages = $preview_msg; var previewDemo = '$demo_string'; var ieWarn = '$ie_warn';";
+    return "var previewMessages = $preview_msg; var tagNames = $tag_names; var previewStrings = $preview_strings; ";
 }
 
 function output_preview_div()
@@ -93,7 +100,6 @@ function output_preview_div()
     $issues = html_safe(_("Issues"));
     $poss_iss = html_safe(_("Poss. Issues"));
     $other_tags = html_safe(_("Other tags"));
-    $background = html_safe(_("Background"));
     $allow_underine = html_safe(_("Allow <u> for underline"));
     $math_preview = html_safe(_("Preview Math"));
     $suppress_warnings = html_safe(_('Suppress warnings'));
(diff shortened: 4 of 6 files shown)


cpeel:
> The changes made the indenting of the html wrong so I re-indented it all. This makes it seem much more different than it really is. It perhaps would be better to remove all the html indenting.

Viewing the diff by hiding whitespace helps when reviewing this.

cpeel:
I think the change to move the "Choose colors" label makes that control more confusing, not less confusing, because now the color selection spans two disparate fieldsets. Using a radio button to change a selection to select the color for that selection is not a standard widget and what make this interface unintuitive to me.

What might be better is removing the radio buttons entirely and just showing the colors for all of the items on one screen in a table format.
```
             Text    Default    Background   Default
Default     [color]               [color]
Italic      [color]    [x]        [color]      [ ]
Bold        [color]    [ ]        [color]      [x]
```

At least in Chrome, a disabled character selector doesn't look disabled which makes it also a bit confusing (i.e. "Why can't I click on the color? Oh, it's because default checkbox is checked"). Having the color selector hidden when default is checked would work if we have the default text and background color always shown on the screen though.

70ray:
That seems a reasonable idea. I'll try it.

70ray:
The failure doesn't seem to be related to this branch.

chrismiceli (approved):

70ray (commented):
Perhaps it would be better to construct the colour table in javascript so the construction and decoding of the 'id's were all in the same file.

cpeel:
> The failure doesn't seem to be related to this branch.

I don't know where it is, but its in the JS stuff. Maybe @chrismiceli can help find it?
```
> node ./SETUP/tests/jsUnit.js
382 assertions of 383 passed, 1 failed.
```

cpeel:
I think this UI is better, but I'm curious what @srjfoo thinks.

chrismiceli:
> > The failure doesn't seem to be related to this branch.
> 
> I don't know where it is, but its in the JS stuff. Maybe @chrismiceli can help find it?
> 
> ```
> > node ./SETUP/tests/jsUnit.js
> 382 assertions of 383 passed, 1 failed.
> ```

Looks like there is a tab character test that needs to be updated:
image

srjfoo:
> I think this UI is better, but I'm curious what @srjfoo thinks.

I think it's better, too (though all that empty space to the right of the color config seems so ... empty!). Chrome and Safari worked just fine. However, I found a couple of issues with FF.

1. I haven't been able to pin down exactly what causes it, but it may be caused by switching to a different window/application while the color picker is active. Somehow there's a problem even opening the color picker at times. Sometimes I could single-click on the box I wanted to change color for, sometimes a double-click would work. Sometimes nothing would work. I could uncheck and re-check the checkboxes (which would then reset the colors to what the user has for default settings), but couldn't actually get the picker to work.

2. Another unexpected thing in FF: unlike in Safari and Chrome, the only way to close the picker was to explicitly click on the close box. This was unexpected because usually when I see that particular configuration, just clicking outside the picker will close it (and that's how it operated for both Chrome and Safari).

srjfoo:
>     ... but it may be caused by switching to a different window/application while the color picker is active.

After some additional testing, I don't think this has anything to do with it. But there's definitely a problem with the color picker and FF. I don't remember that being an issue with the last version, but no clue what's different.

70ray:
> > > The failure doesn't seem to be related to this branch.
> > 
> > 
> > I don't know where it is, but its in the JS stuff. Maybe @chrismiceli can help find it?
> > ```
> > > node ./SETUP/tests/jsUnit.js
> > 382 assertions of 383 passed, 1 failed.
> > ```
> 
> Looks like there is a tab character test that needs to be updated:
> image

Thanks Chris,
When I looked in the log I missed the real problem but saw "/home/travis/build/DistributedProofreaders/dproofreaders/SETUP/tests/jsTests/srchrepTests.js

   22:55  warning  Identifier 'text_data' is not in camel case  camelcase"
 which I thought must be the problem but couldn't understand why it didn't cause other failures.

srjfoo:
> > ... but it may be caused by switching to a different window/application while the color picker is active.
> 
> After some additional testing, I don't think this has anything to do with it. But there's definitely a problem with the color picker and FF. I don't remember that being an issue with the last version, but no clue what's different.

I checked out a page on PROD and poked at the older version, and was able to get similar results. That is, I was able to break the color picker selection in FF on a Mac, so this is not something new to this branch. Testing on Windows would be a good idea, too.

chrismiceli:
> > > > The failure doesn't seem to be related to this branch.
> > > 
> > > 
> > > I don't know where it is, but its in the JS stuff. Maybe @chrismiceli can help find it?
> > > ```
> > > > node ./SETUP/tests/jsUnit.js
> > > 382 assertions of 383 passed, 1 failed.
> > > ```
> > 
> > 
> > Looks like there is a tab character test that needs to be updated:
> > image
> 
> Thanks Chris,
> When I looked in the log I missed the real problem but saw "/home/travis/build/DistributedProofreaders/dproofreaders/SETUP/tests/jsTests/srchrepTests.js
> 
> 22:55 warning Identifier 'text_data' is not in camel case camelcase"
> which I thought must be the problem but couldn't understand why it didn't cause other failures.

Yeah, I'll send a PR to get rid of those red herring warnings. To see the error, I had to open the test suite dproofreaders/SETUP/tests/qunit.html in a browser.

70ray:
> > > ... but it may be caused by switching to a different window/application while the color picker is active.
> > 
> > 
> > After some additional testing, I don't think this has anything to do with it. But there's definitely a problem with the color picker and FF. I don't remember that being an issue with the last version, but no clue what's different.
> 
> I checked out a page on PROD and poked at the older version, and was able to get similar results. That is, I was able to break the color picker selection in FF on a Mac, so this is not something new to this branch. Testing on Windows would be a good idea, too.

I tested it with FF and Edge on windows with no problems. Although on FF the colour picker has to be closed as you say.
On Internet explorer, preview now doesn't work at all. Probably because the character validator causes a javascript problem. This looks like the same problem that causes the image pane to disappear in wordcheck which exists now on PROD.

70ray:
> Perhaps it would be better to construct the colour table in javascript so the construction and decoding of the 'id's were all in the same file.

The latest commit does this. It uses event data to identify the controls rather than decoding id strings.

srjfoo:
> I tested it with FF and Edge on windows with no problems. Although on FF the colour picker has to be closed as you say.
Yes -- as long as the color picker is closed after choosing a color, it operates fine. But if it is not closed after each choice, eventually there are likely to be problems. I don't know if that needs to be documented or not. Safari and Chrome, it's not necessary to explicitly close the picker that I could tell.

scripts/text_validator.js
```diff
@@ -46,6 +46,10 @@ $(function () {
     }
 
     function _validateText() {
+        // IE HACK - xregexp causes problems in IE11, return true if IE
```
cpeel, line 49: I think if you take this out the MR is ready for merging.

70ray: I'd already undone this and rebased it so this last push should have had the required effect.

srjfoo (commented):
Has the sandbox been updated? I tried some testing just now and ...
* Nothing happens if I click on the `Preview` button
* If I click on WordCheck, the image is hidden once I leave WC.

I've gotten the same results in both FF and Safari.

70ray:
> Has the sandbox been updated? I tried some testing just now and ...
> 
>     * Nothing happens if I click on the `Preview` button
> 
>     * If I click on WordCheck, the image is hidden once I leave WC.
> 
> 
> I've gotten the same results in both FF and Safari.

It hasn't been updated. I'll let you know when I've updated it.

cpeel:
I went ahead and threw this into https://www.pgdp.org/~cpeel/c.branch/preview_adjust/ @srjfoo

srjfoo:
> It hasn't been updated. I've updated it now.

> I went ahead and threw this into https://www.pgdp.org/~cpeel/c.branch/preview_adjust/ @srjfoo

Thanks @70ray and @cpeel.

Somewhere along the line, my browsers seem to have gotten corrupted, so, initially, I was seeing the same issues I mentioned above. Checking it out on a different computer, and then again on this one after reboot, and everything looks good. My apologies for the delay.

srjfoo (approved):

70ray:
I tried testing it with Safari on Browserstack and the colour boxes started out all black. It seemed not possible to start the browser console on Browserstack (at least the icon at the top right didn't do anything). But if it works on a real computer that should be a good test.

srjfoo:
> I tried testing it with Safari on Browserstack and the colour boxes started out all black. It seemed not possible to start the browser console on Browserstack (at least the icon at the top right didn't do anything). But if it works on a real computer that should be a good test.

Now that you mention it, I noticed that, too. However, that was just in the picker menu. The colors I'd previously set were still showing in the sample display. I tried clicking on the color pickers, and as long as I didn't choose a new color, nothing changed.

So, I went in and did some further testing. If I change a color, the color in the box changes, and a new color is set. However, if I stop proofreading, and start an independent session, the color boxes are back to black again. I found [this](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/color) article, though I don't know if that's of any use or not.

cpeel:
> So, I went in and did some further testing. If I change a color, the color in the box changes, and a new color is set. However, if I stop proofreading, and start an independent session, the color boxes are back to black again. I found [this](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/color) article, though I don't know if that's of any use or not.

How does this behavior differ from what's already enabled on PROD?

70ray:
> Now that you mention it, I noticed that, too. However, that was just in the picker menu. The colors I'd previously set were still showing in the sample display. I tried clicking on the color pickers, and as long as I didn't choose a new color, nothing changed.
> 
> So, I went in and did some further testing. If I change a color, the color in the box changes, and a new color is set. However, if I stop proofreading, and start an independent session, the color boxes are back to black again. I found [this](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/color) article, though I don't know if that's of any use or not.
That could perhaps explain it. I'll look into it.

srjfoo:
> > So, I went in and did some further testing. If I change a color, the color in the box changes, and a new color is set. However, if I stop proofreading, and start an independent session, the color boxes are back to black again. I found [this](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/color) article, though I don't know if that's of any use or not.
> 
> How does this behavior differ from what's already enabled on PROD?

On PROD right now is the old picker style that just has the two color boxes. There are only ever two boxes, but the colors in them seem to be correctly mapped for the chosen function.

70ray:
I've made a fix so that the color boxes start out with the correct colours in Safari and updated the sandbox.

srjfoo:
> I've made a fix so that the color boxes start out with the correct colours in Safari and updated the sandbox. Tested on Browserstack.

Apologies for the slow response in testing the fix. Looks good to me.

srjfoo (approved):

cpeel:
@70ray - I'm not sure what caused the merge conflict, but please rebase this off of master and I'll get it merged in.

70ray:
> @70ray - I'm not sure what caused the merge conflict, but please rebase this off of master and I'll get it merged in.

Done this.

PR #493, opened 2020-10-29
Provide charsuite guidance for copy pages

cpeel:
While not a perfect check, provide some guidance to SAs about charsuite mismatches when copying pages between projects. @srjfoo identified this functional gap of the prior code in her testing of https://github.com/DistributedProofreaders/dproofreaders/pull/486.

### DIFF
--- a/tools/site_admin/copy_pages.php
+++ b/tools/site_admin/copy_pages.php
@@ -231,17 +231,21 @@ function do_stuff( $projectid_, $from_image_, $page_name_handling,
         die( "Error: no projectid data supplied" );
     }
 
+    $project_obj = [
+        "from" => new Project($projectid_["from"]),
+        "to" => new Project($projectid_["to"]),
+    ];
+
     foreach ( array( 'from', 'to' ) as $which )
     {
-        $projectid = $projectid_[$which];
-        $project = new Project($projectid);
+        $project = $project_obj[$which];
 
         if(!$project->is_utf8)
         {
-            die("Project table $projectid is not UTF-8.");
+            die("Project table {$project->projectid} is not UTF-8.");
         }
 
-        $sql = "DESCRIBE {$projectid_[$which]}";
+        $sql = "DESCRIBE {$project->projectid}";
         $res = DPDatabase::query($sql);
 
         $column_names = array();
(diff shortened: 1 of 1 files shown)


srjfoo (approved):
Tested. Looks good.

tools/site_admin/copy_pages.php
```diff
@@ -231,17 +231,21 @@ function do_stuff( $projectid_, $from_image_, $page_name_handling,
...
     foreach ( array( 'from', 'to' ) as $which )
     {
-        $projectid = $projectid_[$which];
-        $project = new Project($projectid);
+        $project = $project_obj[$which];
 
         if(!$project->is_utf8)
         {
-            die("Project table $projectid is not UTF-8.");
+            die("Project table {$project->projectid} is not UTF-8.");
```
chrismiceli, line 245: Does the project constructor not do this for us?

tools/site_admin/copy_pages.php
```diff
@@ -383,6 +386,33 @@ function do_stuff( $projectid_, $from_image_, $page_name_handling,
...
+        function ($a, $b) { return $a == $b ? 0 : 1; }
+    );
+    if ($charsuites_not_in_to)
+    {
+        echo "<p class='warning'>" . _("Character Suite mismatch") . "</p>";
+        echo "<p>" . _("The following character suites are in the source project but not in the destination project. Ensure the copied pages do not use these or adjust the destination project character suites accordingly.") . "</p>";
+        echo "<ul>";
+        foreach($charsuites_not_in_to as $charsuite)
+        {
+            echo "<li>" . $charsuite->name . "</li>";
```
chrismiceli, line 401: Html safe would be nice here just in case

chrismiceli (approved):

tools/site_admin/copy_pages.php
cpeel: No, it can't. We did not convert project tables to UTF-8 for posted or archived projects. If those projects get un-posted or un-archived (which is rare but does happen) they will need to be converted to UTF-8 before being used. We have a few places in the code that check for UTF-8-ness and a SA script that allows SAs to convert them.

tools/site_admin/copy_pages.php
cpeel: Character suite names are defined in the code and while unlikely to contain beyond alphanumeric, dashes, and underscores technically could. I'll get this in.

chrismiceli (approved):

PR #494, opened 2020-10-31
Name text and border colors

srjfoo:
In `layout.less`, name text and border colors. Changed a few more isolated instances of inline styles controlling color to use classes.

In a separate commit, fixed some improperly closed elements in `styles/style_demo.php`.

sandbox: [text-border-colors](https://www.pgdp.org/~srjfoo/c.branch/text-border-colors/)

### DIFF
--- a/pinc/mentorbanner.inc
+++ b/pinc/mentorbanner.inc
@@ -33,21 +33,17 @@ function mentor_banner($round)
         case 0:
         case 1:
         case 2:
-            $font_boost = "1em";
-            $font_col = '#339933';
+            $class = 'mentor-recent';
             break;
         case 3:
         case 4:
-            $font_boost = "large";
-            $font_col = "#FF6600";
+            $class = 'mentor-older';
             break;
         default:
-            $font_boost = "x-large";
-            $font_col = "#FF0000";
+            $class = 'mentor-oldest';
             break;
     }
-
-    echo "<p style='color: $font_col; font-weight: bold; font-size: $font_boost; text-align: center'>";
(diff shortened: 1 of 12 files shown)


pinc/mentorbanner.inc
```diff
@@ -33,21 +33,17 @@ function mentor_banner($round)
...
             break;
         default:
-            $font_boost = "x-large";
-            $font_col = "#FF0000";
+            $class = 'mentor-oldest';
             break;
     }
-
-    echo "<p style='color: $font_col; font-weight: bold; font-size: $font_boost; text-align: center'>";
+    echo "<p class = '$class'>";
```
cpeel, line 46: We don't include spaces between tag params and their values, so let's remove the extra spaces here.

styles/layout.less
```diff
@@ -694,19 +738,22 @@ table.snapshottable {
     }
 }
 
+@access-yes: @text-ok;
```
cpeel, line 741: I'm not certain we want to tie the access colors to the text OK/error/warning colors. Consider the case where a theme changes warning to orange -- I don't think an orange info icon is the right message for "eligible".

srjfoo: Fair comment -- and I felt uncomfortable using the error/warning colors there. Not sure why I left that, actually. I can either put the colors in directly, or define additional color names--do you have a preference?

srjfoo: Actually, never mind ... because that was already a level of indirection. I'll just assign the colors to those variables directly.

PR #495, opened 2020-10-31
Better param validation for page detail

cpeel:
Do more robust validation for page detail parameters and address an XSS vulnerability.

This change also allows any user to see the pages for a specific user by setting `select_by_user` to that username. This data is already visible to them in the table so why make it harder to see?

Testable in the [page-detail-params](https://www.pgdp.org/~cpeel/c.branch/page-detail-params/) sandbox.

### DIFF
--- a/tools/project_manager/page_detail.php
+++ b/tools/project_manager/page_detail.php
@@ -10,51 +10,46 @@ require_login();
 
 $projectid = get_projectID_param($_GET, 'project');
 $show_image_size = get_integer_param($_GET,'show_image_size',0,0,1);
+$round_for_page_selection = get_enumerated_param($_GET, 'select_by_round', NULL, array_keys($Round_for_round_id_), TRUE);
 
-$project = new Project( $projectid );
-
-if ( isset($_GET['select_by_user']) )
+// select_by_user can have three possible values:
+//    NULL = show all users
+//    '' = show current user
+//    <username> = show specific user
+$username_for_page_selection = array_get($_GET, 'select_by_user', NULL);
+if($username_for_page_selection === '')
 {
-    $sbu = $_GET['select_by_user'];
-    if ( empty($sbu) )
+    $username_for_page_selection = $pguser;
+}
+
+// Validate user selection
+if($username_for_page_selection)
+{
+    // Only SAs and PFs can use select_by_user. This is to prevent users from
+    // tweaking this value and determining what user proofread what page in a
+    // project. This is possible because while the table renderer will hide
+    // usernames for users not privileged to access them, select_by_user and
+    // select_by_round will still only present the requested rows allowing the
+    // requester to determine what pages a user did in which round.
+    if(! (user_is_a_sitemanager() || user_is_proj_facilitator()) )
     {
-        // Show just the current user's pages.
         $username_for_page_selection = $pguser;
     }
-    else
-    {
-        // Explicitly specify a particular user.
-        // This is only available to PFs & SAs.
-        // (Yes, even though it merely filters
-        // information that is available to all.)
-        if ( user_is_a_sitemanager() || user_is_proj_facilitator() )
-        {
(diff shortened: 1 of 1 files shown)


tools/project_manager/page_detail.php
```diff
@@ -10,51 +10,32 @@
...
 $projectid = get_projectID_param($_GET, 'project');
 $show_image_size = get_integer_param($_GET,'show_image_size',0,0,1);
-
-$project = new Project( $projectid );
-
-if ( isset($_GET['select_by_user']) )
+$round_for_page_selection = get_enumerated_param($_GET, 'select_by_round', NULL, array_keys($Round_for_round_id_), TRUE);
+
+// select_by_user can have three possible values:
+//    NULL = show all user
```
bpfoley, line 16: Nit: typo s/user/users/

tools/project_manager/page_detail.php
```diff
@@ -80,14 +61,14 @@
     {
         if (is_null($round_for_page_selection) )
         {
-            echo sprintf( _("Showing only the pages of user '%s'."), 
-                          $username_for_page_selection );
+            echo sprintf( _("Showing only the pages of user '%s'."),
+                          html_safe($username_for_page_selection) );
```
bpfoley, line 65: Are these actually necessary? Given we've validated `$username_for_page_selection` above (forcing it to be NULL or a valid user). `check_username` on registration only allows `/[^-a-zA-Z0-9@. ]/`, none of which are HTML escapes.

tools/project_manager/page_detail.php
```diff
@@ -80,14 +61,14 @@
...
         }
         else
         {
-            echo sprintf( _("Showing only the pages of user '%1\$s' in round %2\$s."), 
-                          $username_for_page_selection, 
-                          $round_for_page_selection );
+            echo sprintf( _("Showing only the pages of user '%1\$s' in round %2\$s."),
+                          html_safe($username_for_page_selection),
+                          html_safe($round_for_page_selection) );
```
bpfoley, line 71: Again, `$round_for_page_selection` is validated, and none of our round IDs include HTML escapes.

tools/project_manager/page_detail.php
cpeel: They aren't now that we're validating the username and round (they were before we did it). I can remove them to simplify the diff if you'd like.

bpfoley: Either works for me. I suppose escaping is a belt-and-braces thing. If the username/rounds are ever allowed to include escapes, then we'd be in trouble.

srjfoo (commented):
Shouldn't this show only pages proofed by the current user, assuming that the user is not also the PM or the PPer? That is, the PM or the PPer can see all usernames for all rounds, but an unprivileged user can see usernames only for the pages they've proofed. Shouldn't the `select_by_user` only work on those pages they've proofed, not for all pages in the project?

Screenshot using an unprivileged account:
image

I verified with my main account that the two pages listed without usernames were proofed in P1 by rouelle.

cpeel:
> Shouldn't this show only pages proofed by the current user, assuming that the user is not also the PM or the PPer? That is, the PM or the PPer can see all usernames for all rounds, but an unprivileged user can see usernames only for the pages they've proofed. Shouldn't the `select_by_user` only work on those pages they've proofed, not for all pages in the project?

The only non-validation functional change I made was to allow `select_by_user` to be used for non-PM/SAs to show pages for a specific user. It looks like this clashes with the "hide the proofer" logic when rendering the table allowing a user to figure out what pages in a project a user worked on by tweaking `select_by_user` and `select_by_round`. I'll add that enforcement back (with a comment this time so we know why it is there).

chrismiceli (approved):

srjfoo (approved):
Looks good now.

PR #496, opened 2020-11-01
manage_special_days.php: Use DPDatabase APIs, sprintf, set_col_str

bpfoley:
Move set_col_num, set_col_str to misc because it's generally useful.

### DIFF
--- a/pinc/misc.inc
+++ b/pinc/misc.inc
@@ -580,6 +580,30 @@ function surround_and_join( $strings, $L, $R, $joiner )
     return implode($joiner,$parts);
 }
 
+/**
+ * Return a SQL "colname='value'" clause, where `$colname` is
+ * not escaped, and `$s` is an escaped and quoted string.
+ */
+function set_col_str($colname, $s)
+{
+    return sprintf("%s='%s'", $colname, DPDatabase::escape($s));
+}
+
+/**
+ * Return a SQL "colname=value" clause, where `$colname` is
+ * not escaped, and `$i` is treated as an integer.
+ *
+ * If `$i` is non-numeric, an `InvalidArgumentException` is thrown.
+ */
+function set_col_num($colname, $i)
+{
+    if (!is_numeric($i))
+    {
+        throw new InvalidArgumentException("$i is not a number");
+    }
+    return sprintf("%s=%d", $colname, $i);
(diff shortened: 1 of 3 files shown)


bpfoley:
Available in https://www.pgdp.org/~bfoley/c.branch/cleanup15/

No obvious problems adding, removing, editing a special day.

More needs to be done here to clean up the input handling rather than using raw `$_POST`, but I'm trying to limit the scope of these changes.

pinc/misc.inc
```diff
@@ -596,21 +620,20 @@ function surround_and_join( $strings, $L, $R, $joiner )
...
 {
-    $update_fields = array();
+    $update_fields = [];
     foreach($source_array as $field => $value)
     {
-        if(in_array($field, $string_fields))
+        if (in_array($field, $string_fields))
         {
-            $value = sprintf("'%s'", mysqli_real_escape_string(DPDatabase::get_connection(), trim($value)));
+	    $update_fields[] = set_col_str($field, $value);
```
cpeel, line 628: We should retain the `trim()` or we should ensure that this is done (if it needs to be done) in all of the callers of this function.

srjfoo:
> No obvious problems adding, removing, editing a special day.

No problem adding or editing ... but the script doesn't remove special days, does it? (enable and disable, yes, but not actually delete)

bpfoley:
> > No obvious problems adding, removing, editing a special day.
> 
> No problem adding or editing ... but the script doesn't remove special days, does it? (enable and disable, yes, but not actually delete)

Yup, you're right. There's no actual delete (which is a little odd, when you think about it.)

srjfoo:
> > > No obvious problems adding, removing, editing a special day.
> > 
> > 
> > No problem adding or editing ... but the script doesn't remove special days, does it? (enable and disable, yes, but not actually delete)
> 
> Yup, you're right. There's no actual delete (which is a little odd, when you think about it.)

As far as I can tell, we don't delete special days. We may disable them, but if a project was ever created and made it all the way through with the special day still enabled, it's associated with that project forever. If we deleted special days, we'd have to go through and make sure that any project that used it had it removed.

srjfoo (approved):
Works as expected as far as I can tell.

PR #499, opened 2020-11-03
access_log.inc: Use DPDatabase, and set_col helper functions

bpfoley:
Changes to log_access_change and get_first_granted_date are both exercised by looking at a member detail page and granting/revoking permissions for a round. 

Sandbox at https://www.pgdp.org/~bfoley/c.branch/cleanup16/stats/members/mdetail.php?id=151

### DIFF
--- a/pinc/access_log.inc
+++ b/pinc/access_log.inc
@@ -1,22 +1,19 @@
 <?php
+include_once($relPath.'misc.inc');
 
 function log_access_change( $subject_username, $modifier_username, $activity_id, $action_type )
 {
-    $sql = sprintf("
+    $setters = join(", ", [
+        set_col_num("timestamp", time()),
+        set_col_str("subject_username", $subject_username),
+        set_col_str("modifier_username", $modifier_username),
+        set_col_str("action", $action_type),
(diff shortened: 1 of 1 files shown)


srjfoo (approved):
Didn't stumble over anything unusual in testing

pinc/access_log.inc
```diff
@@ -1,22 +1,19 @@
 <?php
+include_once($relPath.'misc.inc');
```
chrismiceli, line 2: Bringing this in has no unintended consequences (performance? side effects?)

bpfoley: So, the only side effects should be to bring a bunch of utility functions (and 2 empty Exception classes) into the global namespace.

If I understand things correctly, there shouldn't really be any noticeable performance issues (apart from possibly an extra `stat()` syscall to check `misc.inc` hasn't changed) because every `.php` file will end up being precompiled and stored in the opcode cache.

In any event, all 3 files that include `access_log.inc` already include `misc.inc`.

cpeel: <nod> This is very likely a no-op since other things include `misc.inc`.

PR #500, opened 2020-11-03
stats/misc_stats1.php: Use DPDatabase::escape

bpfoley:
This file really needs a more comprehensive cleanup,
but at least this is a step in the right direction.

Sandbox for testing at 
https://www.pgdp.org/~bfoley/c.branch/cleanup17/stats/misc_stats1.php?tally_name=P2

### DIFF
--- a/stats/misc_stats1.php
+++ b/stats/misc_stats1.php
@@ -9,7 +9,7 @@ include_once($relPath.'misc.inc'); // get_enumerated_param()
 require_login();
 
 $valid_tally_names = array_keys(get_page_tally_names());
-$tally_name   = get_enumerated_param($_GET, 'tally_name', null, $valid_tally_names);
+$tally_name = get_enumerated_param($_GET, 'tally_name', null, $valid_tally_names);
 
 // -----------------------------------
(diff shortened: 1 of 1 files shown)


stats/misc_stats1.php
```diff
@@ -84,10 +84,10 @@ function show_top_days( $n, $when )
         select_from_site_past_tallies_and_goals(
             $tally_name,
             "SELECT
-                {date} as '" . mysqli_real_escape_string(DPDatabase::get_connection(), _("Date")) . "',
-                tally_delta as '" . mysqli_real_escape_string(DPDatabase::get_connection(), _("Pages Proofread")) . "',
+                {date} as '" . DPDatabase::escape(_("Date")) . "',
```
chrismiceli, line 87: We localize the column aliases? This does seem odd. Do we raw dump these results somewhere to be shown in the UI?

cpeel: Yeah, these dump out to a table with the fields as column names. It's a long-standing and horrible design I don't want to solve here.

PR #501, opened 2020-11-03
wordcheck_engine: Use DPDatabase APIs and use PHP7 array syntax

bpfoley:
Sandbox in https://www.pgdp.org/~bfoley/c.branch/cleanup18/

Tested by going through various permutations of the wordcheck interface.

### DIFF
--- a/pinc/wordcheck_engine.inc
+++ b/pinc/wordcheck_engine.inc
@@ -83,13 +83,13 @@ function _get_bad_words_for_project_text($which_result_values, $text, $projectid
     list($langcode3s, $new_messages) = get_langcode3s_for_languages($languages);
     $messages += $new_messages;
 
-    $word_lists = array(
+    $word_lists = [
         "adhoc_good" => $adhoc_good_words,
         "site_good" => load_site_good_words($langcode3s),
         "site_bad" => load_site_bad_words($langcode3s),
         "project_good" => load_project_good_words($projectid),
         "project_bad" => load_project_bad_words($projectid)
-    );
+    ];
 
     list($input_words_w_freq, $bad_words, $new_messages) =
         get_bad_words_for_text($text, $languages, $word_lists);
@@ -118,7 +118,7 @@ function _get_bad_words_for_project_text($which_result_values, $text, $projectid
         assert(FALSE);
     }
 
-    return array($bad_words_in_text, $languages, $messages);
+    return [$bad_words_in_text, $languages, $messages];
 }
 
 
@@ -160,8 +160,8 @@ function get_bad_words_for_text($text, $languages, $word_lists=[])
 
     // The site
     {
-        $acc->remove_good_words(array_get($word_lists, "site_good", array()));
-        $acc->add_bad_words(array_get($word_lists, "site_bad", array()), WC_SITE);
+        $acc->remove_good_words(array_get($word_lists, "site_good", []));
+        $acc->add_bad_words(array_get($word_lists, "site_bad", []), WC_SITE);
 
         $acc->add_bad_words(
             get_bad_words_via_pattern($input_words_w_freq, $languages), WC_SITE);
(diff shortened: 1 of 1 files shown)


pinc/wordcheck_engine.inc
```diff
@@ -378,23 +378,23 @@ function get_bad_words_via_pattern($input_words_w_freq, $languages)
...
-    );
-    $exceptions = array();
+    ];
+    $exceptions = [];
     foreach($langcode3s as $langcode3)
     {
         $exceptions[] = array_get($exceptions_for_lang, $langcode3, "");
     }
-    $exceptions = implode('|', $exceptions);
+    $exceptions = join('|', $exceptions);
```
chrismiceli, line 397: why this change?

pinc/wordcheck_engine.inc
```diff
@@ -1103,15 +1100,7 @@ function get_langcode3s_for_languages($languages)
...
-    $natcaseresult = strnatcasecmp($a, $b);
-    if($natcaseresult != 0)
-    {
-        return $natcaseresult;
-    }
-    else
-    {
-        return strnatcmp($a, $b);
-    }
+    return strnatcasecmp($a, $b)?:strnatcmp($a, $b);
```
chrismiceli, line 1103: I like the explicit check to 0 that we had before instead of relying on it being falsy.

pinc/wordcheck_engine.inc
```diff
@@ -563,44 +565,36 @@ function copy_project_wordcheck_events($from_projectid, $to_projectid)
...
 function save_wordcheck_event($projectid, $round, $page, $proofer, $suggestions, $corrections)
 {
-    $suggestion_list = implode("\n", $suggestions);
-
-    $correction_list = array();
-
     // save the corrections in wdiff format
-    foreach($corrections as $correction) {
-        list($orig, $new) = $correction;
+    $correction_list = [];
```
chrismiceli, line 581: is this used anymore?

bpfoley: Well spotted. It looks like not. We're inserting and reading it from the DB, but as far as I can see the only user of `load_wordcheck_events` in `show_good_word_suggestions_detail.php` doesn't use it anymore. I'd prefer to do this as a followup MR so as not to change any behavior in this one though.

pinc/wordcheck_engine.inc
bpfoley: OK. Perhaps I'm too used to false-y 0 in C and Python.

pinc/wordcheck_engine.inc
bpfoley: No great reason. I just trip over `implode` every time I see it. Every other language I know that has a similar feature calls it `join` and the word feels less jargony and more pseudocode to me. I figured that since I was doing the cosmetic `array(...)` to `[...]` conversion I may as well do this too. (And of course, PHP being PHP, `split` had to be a different, now deprecated function that isn't the inverse of `join`, so we're stuck with `explode`. Sigh.)

pinc/wordcheck_engine.inc
```diff
@@ -626,12 +620,16 @@ function count_wordcheck_suggestion_events($projectid, $start_time=0)
...
+        SELECT * FROM wordcheck_events
+        WHERE projectid='%s',
+        AND timestamp>%d",
+        DPDatabase::escape($projectid),
+        $start_time
+    );
 
-    $res = mysqli_query(DPDatabase::get_connection(), $sql);
+    $eventArray = [];
+    $res = DPDatabase::query($sql, FALSE);
```
cpeel, line 632: While this is true to the original, the original is wrong and this should raise on error. If the query returns zero rows (which is possible) `$res` will still be not FALSE.

pinc/wordcheck_engine.inc
cpeel: I'm neutral on these changes. I'm happy to take them but I don't feel like we need to search/replace to change them.

pinc/wordcheck_engine.inc
```diff
@@ -614,7 +608,7 @@ function count_wordcheck_suggestion_events($projectid, $start_time=0)
             WHERE projectid='$projectid'
```
cpeel, line 608: We need to escape `$projectid` and ensure `$start_time` is a number via sprintf. Also we should raise an exception on error (the FALSE check is wrong in the original code).

bpfoley: Done.

PR #502, opened 2020-11-03
skip validate in Internet Explorer

70ray:


scripts/text_validator.js
```diff
@@ -46,6 +46,10 @@ $(function () {
     }
 
     function _validateText() {
+        // IE HACK - xregexp causes problems in IE11, return true if IE
+        if(document.documentMode) {
```
chrismiceli, line 50: is this the best way to determine IE? Here seems to imply this only works in dev tools:
https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/platform-apis/dn255470(v=vs.85)

70ray: It seems to work.
But on the other hand, the first error appearing in the IE console is about an uncaught exception in xregexp so perhaps we should try  to catch it which maybe we should do anyway since there may be other problems lying in wait.

cpeel: I think we should solve this differently. We already have a hack below for IE11 to skip normalization but we should still attempt to validate and this doesn't do that.

70ray:
made a different fix. The name is now wrong so closing this and making a new MR

PR #503, opened 2020-11-03
Cleanup language code

cpeel:
Our language code is a bit of a mess. We have some functions that deal with locales and some that deal with projects. This attempts to make a clear distinction between the two:
* iso_lang_list - project-related language arrays and functions
* languages.inc - locale-related language functions

This MR removes some duplicate mapping code and some unused functions. It also moves global arrays into functions.

The only place this should have a functional change is that it will better detect the available dictionaries and how they map to a project's languages. This doesn't solve the Ancient Greek WordCheck problem -- I have that solved in a subsequent commit but needed to get this mess cleaned up first.

Testable in the [cleanup-language-code](https://www.pgdp.org/~cpeel/c.branch/cleanup-language-code) sandbox.


srjfoo (approved):

pinc/iso_lang_list.inc
```diff
@@ -1,4 +1,61 @@
...
+
+// Returns ISO 639 code(s) for language(s) of a project.
+// $which could be "primary", "secondary" or "all":
+// - "primary" returns the code of primary language
+// - "secondary" returns the code of secondary language, or NULL if the
+//   secondary language doesn't exist
+// - "all" returns array with the code(s) of both language(s); if the secondary
+//   language doesn't exist, the array only has one element
+// If a language doesn't have ISO 639 code, NULL is returned; this might change
+function proj_lang_code($proj_lang, $which="all")
```
chrismiceli, line 15: an enum would be nice to specifically enumerate options

chrismiceli (approved):

pinc/iso_lang_list.inc
cpeel: PHP doesn't have enums unfortunately.

Interestingly, the only value that is ever used when calling this function is `primary` which tells me we might just want to remove the option entirely and simplify the function:
```
$ grep -R proj_lang_code *
pinc/wordcheck_engine.inc:        $langcode = proj_lang_code($language, "primary");
pinc/wordcheck_engine.inc:        $langcode = proj_lang_code($language, "primary");
pinc/LPage.inc:        $lang = proj_lang_code($project->language,"primary");
```

cpeel: Let's leave this as-is for this MR. I see how I want to rework project languages slightly differently but think that needs to happen in a follow-up.

PR #505, opened 2020-11-04
for_mentors.php: Use DPDatabase::escape()

bpfoley:
This needs a more comprehensive cleanup -- embedding UI values in the
SQL queries is horrible --  but baby steps.

### DIFF
--- a/tools/proofers/for_mentors.php
+++ b/tools/proofers/for_mentors.php
@@ -37,7 +37,7 @@ else
     // Consider the page they came from.
     $referer = @$_SERVER['HTTP_REFERER'];
 
-    // If they're coming to this page from a MENTORS ONLY book in X2, 
+    // If they're coming to this page from a MENTORS ONLY book in X2,
     // referrer should contain &expected_state=X2.proj_avail.
     foreach ( $Round_for_round_id_ as $round )
         {
@@ -208,16 +208,17 @@ function output_project_details($mentored_round, $projectid, $nameofwork, $autho
 
 function get_beginner_projects_in_state($state)
 {
-    $sql = "
+    $sql = sprintf("
         SELECT projectid, nameofwork, authorsname
         FROM projects
         WHERE
             difficulty = 'BEGINNER' AND
-            state='$state'
+            state='%s'
         ORDER BY
-            modifieddate ASC
-    ";
-    $result = mysqli_query(DPDatabase::get_connection(), $sql);
+            modifieddate ASC",
+        DPDatabase::escape($state)
+    );
+    $result = DPDatabase::query($sql);
     $projects = [];
     while($proj_obj = mysqli_fetch_object($result))
     {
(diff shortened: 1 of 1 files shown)


bpfoley:
Sandbox at: https://www.pgdp.org/~bfoley/c.branch/cleanup19

tools/proofers/for_mentors.php
```diff
@@ -208,16 +208,17 @@ function output_project_details($mentored_round, $projectid, $nameofwork, $autho
...
         WHERE
             difficulty = 'BEGINNER' AND
-            state='$state'
+            state='%s'
         ORDER BY
-            modifieddate ASC
-    ";
-    $result = mysqli_query(DPDatabase::get_connection(), $sql);
+            modifieddate ASC",
+        DPDatabase::escape($state)
```
chrismiceli, line 219: we didn't encode before, we're sure all callers are not encoding already and we won't introduce double encoding, right?

tools/proofers/for_mentors.php
```diff
@@ -243,14 +244,14 @@ function page_summary_sql($mentored_round, $projectid)
             ELSE CONCAT('<a href=\""
                 .$code_url . "/stats/members/mdetail.php?&id=',u.u_id,
                 '\">',u.username,'</a>')
-            END AS '" . mysqli_real_escape_string(DPDatabase::get_connection(), _("Proofreader")) . "',
-            COUNT(1) AS '" . mysqli_real_escape_string(DPDatabase::get_connection(), _("Pages this project")) . "',
+            END AS '" . DPDatabase::escape(_("Proofreader")) . "',
+            COUNT(1) AS '" . DPDatabase::escape(_("Pages this project")) . "',
```
chrismiceli, line 248: sprintf perhaps?

tools/proofers/for_mentors.php
```diff
@@ -273,10 +274,10 @@ function page_list_sql($mentored_round, $projectid)
         image
     ";
 
-    $page = mysqli_real_escape_string(DPDatabase::get_connection(), _("Page"));
-    $saved = mysqli_real_escape_string(DPDatabase::get_connection(), _("Saved"));
-    $proofreader = mysqli_real_escape_string(DPDatabase::get_connection(), _("Proofreader"));
-    $wc_events = mysqli_real_escape_string(DPDatabase::get_connection(), _("WordCheck Events"));
+    $page = DPDatabase::escape(_("Page"));
+    $saved = DPDatabase::escape(_("Saved"));
+    $proofreader = DPDatabase::escape(_("Proofreader"));
```
chrismiceli, line 279: do we need these as separate variables? I'd just sprintf below without creating these if possible.

tools/proofers/for_mentors.php
bpfoley: True, but there are only 2 callers and they're just using one of the `Round::foo_state` fields which will be something like `P3.avail`.

```$ ag get_beginner_projects_in_state
tools/proofers/for_mentors.php
123:$projects_available = get_beginner_projects_in_state($mentoring_round->project_available_state);
147:$projects_waiting = get_beginner_projects_in_state($mentoring_round->project_waiting_state);
209:function get_beginner_projects_in_state($state)
```

We don't really need to escape, but I figured I may as well do it for consistency. I can undo that if you like.

tools/proofers/for_mentors.php
bpfoley: I think in this case it would end up being confusing. There's a lot of interleaving of field names (using variable interpolation) which we can't directly `sprintf` because the strings might contain `%`s.

Ironically, I think the more readable solution in this case is to introduce some variables like the comment below so we don't need to have all the escaping and concatenation inline. Let me know if you dislike it.

tools/proofers/for_mentors.php
bpfoley: Technically no, but I think `sprintf` is a bit of a no-go here for exactly the same reason as above -- we've too many interpolated variables interleave with `DPDatabase::escape` values that we'd like to keep as interpolated variables.

PR #506, opened 2020-11-04
project.php: Use DPDatabase APIs, more robust escaping

bpfoley:
And a couple of minor code cleanups.

### DIFF
--- a/project.php
+++ b/project.php
@@ -23,6 +23,7 @@ include_once($relPath.'forum_interface.inc'); // get_last_post_time_in_topic & g
 include_once($relPath.'misc.inc'); // html_safe(), get_enumerated_param(), get_integer_param(), array_get(), humanize_bytes()
 include_once($relPath.'faq.inc');
 include_once($relPath.'daily_page_limit.inc'); // get_dpl_count_for_user_in_round
+include_once($relPath.'special_colors.inc'); // load_special_days
 
 // If the requestor is not logged in, we refer to them as a "guest".
 
@@ -317,13 +318,14 @@ function decide_blurbs()
 
         // Has the user saved a page of this project since the project comments were
         // last changed? If not, it's unlikely they've seen the revised comments.
-        $res = mysqli_query(DPDatabase::get_connection(), "
+        $sql = "
             SELECT {$round->time_column_name}
             FROM $projectid
-            WHERE state='{$round->page_save_state}' AND {$round->user_column_name}='$pguser'
+            WHERE state='{$round->page_save_state}' AND {$round->user_column_name}='" . DPDatabase::escape($pguser) . "'
             ORDER BY {$round->time_column_name} DESC
             LIMIT 1
-        ");
+        ";
+        $res = DPDatabase::query($sql);
         $row = mysqli_fetch_assoc($res);
         if (!$row)
         {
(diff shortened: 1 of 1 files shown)


bpfoley:
Sandbox at: https://www.pgdp.org/~bfoley/c.branch/cleanup20

project.php
```diff
@@ -317,13 +317,14 @@ function decide_blurbs()
 
         // Has the user saved a page of this project since the project comments were
         // last changed? If not, it's unlikely they've seen the revised comments.
-        $res = mysqli_query(DPDatabase::get_connection(), "
+        $sql = "
             SELECT {$round->time_column_name}
             FROM $projectid
-            WHERE state='{$round->page_save_state}' AND {$round->user_column_name}='$pguser'
+            WHERE state='{$round->page_save_state}' AND {$round->user_column_name}='" . DPDatabase::escape($pguser) . "'
```
chrismiceli, line 323: no sprintf?

project.php
```diff
@@ -899,11 +902,12 @@ function recentlyproofed( $wlist )
     $sql = "
         SELECT $projectid.image, $projectid.state, {$round->time_column_name}, $wordcheck_query
         FROM $projectid
-        WHERE {$round->user_column_name}='$pguser' AND $state_condition
+        WHERE {$round->user_column_name}='" . DPDatabase::escape($pguser) . "'
```
chrismiceli, line 905: same here, sprintf?

project.php
bpfoley: It's awkward to use` sprintf` in this case because most of the interpolated variables are field names and the style guide [says](https://www.pgdp.net/wiki/DP_Code_Best_Practices#Column_names_as_variables) to keep those inline rather than using `%s`. In other places with similar issues, Casey suggested using concatenation like this.

project.php
bpfoley: Same rationale as above.

project.php
cpeel: Yes, this is less bad than the alternatives.

project.php
```diff
@@ -478,9 +479,10 @@ function do_project_info_table()
             $sql = sprintf("
```
cpeel, line 479: Rather than fixing this SQL, let's replace it with:
```php
            $special_days = load_special_days();
            if(isset($special_days[$spec_code]))
            {
                $spec_display = $special_days[$spec_code]["display_name"];
            }
            else
            {
                $spec_display = "($spec_code)";
            }
```
from `pinc/special_colors.inc`.

bpfoley: Rather fascinatingly, PHP 7.0's null coalescing is more powerful than I thought. It can do far more than just replacing tedious `if isset($foo) { $foo } else { $bar }`.

You can use it with chains of array derefs and it does what you'd hope it would do. Eg
```
  error_reporting(-1);
  $aa = [ "foo" => [
            "bar" => "baz",
            "boz" => ""
          ]
        ];
  var_dump(NULL["foo"]["bar"]??"other1"); # => "other1"
  var_dump($aa["foo"]["xar"]??"other2");  # => "other2"
  var_dump($aa["xoo"]["bar"]??"other3");  # => "other3"
  var_dump($aa["foo"]["bar"]??"other4");  # => "baz"
  var_dump($aa["foo"]["boz"]??"other5");  # => ""
  var_dump($x??"I didn't define x!");     # => "I didn't define x!"
```

And we can use it in this code too...

cpeel: Holy cow that's awesome! Let's use more of this!

project.php
```diff
@@ -466,29 +467,16 @@ function do_project_info_table()
...
-            $spec_res = mysqli_fetch_assoc(mysqli_query(DPDatabase::get_connection(), $sql));
-            if ($spec_res)
-            {
-                $spec_display = $spec_res['display_name'];
-            }
-            else
-            {
-                $spec_display = "($spec_code)";
-            }
+            $special_days = load_special_days();
```
cpeel, line 478: You need to import `pinc/special_colors.inc` for this to work though.

project.php
bpfoley: I think what's going on here is that PHP suppresses all errors, squashes them to null and terminates early whenever it evaluates the left hand side of a `??` (which is why the`$x??"Undefined x"` case works).

So in one sense it's a modern version of `@`, with all the pitfalls that go along with that. But yeah, I like it!

project.php
bpfoley: Argh! Sorry!

PR #507, opened 2020-11-04
stats_central.php: Use DPDatabase, and factor out repeated code

bpfoley:
Sandbox at: https://www.pgdp.org/~bfoley/c.branch/cleanup21

### DIFF
--- a/stats/stats_central.php
+++ b/stats/stats_central.php
@@ -49,6 +49,20 @@ show_news_for_page("STATS");
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 //General site stats with links to view the queue's
 
+function count_books_in_state($state, $clauses="")
+{
+    $sql = sprintf("
+        SELECT COUNT(*) AS numbooks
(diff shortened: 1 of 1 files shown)


stats/stats_central.php
```diff
@@ -70,12 +84,8 @@
...
     );
 
-  //get total books posted  in the last 7 days
-
-    $books = mysqli_query(DPDatabase::get_connection(), "SELECT count(*) AS numbooks FROM projects
-                          WHERE modifieddate >= $begin_time AND state = '".PROJ_SUBMIT_PG_POSTED."'");
-    $row = mysqli_fetch_assoc($books);
-    $totalbooks = $row["numbooks"];
+    //get total books posted  in the last 7 days
+    $totalbooks = count_books_in_state(PROJ_SUBMIT_PG_POSTED, "AND modifieddate >= $begin_time");
```
chrismiceli, line 88: do we want to use sprintf here with %d?

chrismiceli (approved):

stats/stats_central.php
bpfoley: We could, but given that `$begin_time` doesn't depend on data under control of the user, I think we're pretty safe.

```$begin_time = time() - 604800; // in seconds```

PR #510, opened 2020-11-05
Support phpunit8 and phpunit9

chrismiceli:
change php versions in travis

### DIFF
--- a/SETUP/tests/README.md
+++ b/SETUP/tests/README.md
@@ -4,13 +4,7 @@ These are phpunit tests.
 
 ## Version requirements
 
-The phpunit tests, as-coded, will work with phpunit 5.7, 6.x, and 7.x. They
-will not work in 8.x or later because the setUp() and tearDown() functions
-do not specify void return types. (See the
-[phpunit8 release notes](https://phpunit.de/announcements/phpunit-8.html).)
-That capability isn't available until PHP 7.1, which the DP code will
-nominally work with, but isn't available on the TEST server or in many
-modern distros.
+The phpunit tests, as-coded, will work with phpunit 6.x, 7.x, 8.x, and 9.x.
 
 ## Running tests
 
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ pinc/udb_user.php
 messages.mo
 
 node_modules
+.phpunit.result.cache
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,9 @@ php:
     # Test on a wide array of PHP versions. 7.3 and later can't be added until
     # the phpunit tests are updated and a more recent version of phpunit is
     # specified below.
-    - 7.0
-    - 7.1
     - 7.2
+    - 7.3
+    - 7.4
 services:
     - mysql
 before_install:
@@ -23,10 +23,9 @@ install:
     - php install_db.php
     - cd ..
 before_script:
-    # our phpunit tests won't work with phpunit8 (see SETUP/tests/README.md)
-    # so specify a version that works with php 7.0, 7.1, 7.2 (see
+    # Specify a version that works with php 7.2, 7.3, 7.4 (see
     # https://phpunit.de/supported-versions.html)
-    - curl -sSfL -o ~/.phpenv/versions/$(phpenv version-name)/bin/phpunit https://phar.phpunit.de/phpunit-6.phar
+    - curl -sSfL -o ~/.phpenv/versions/$(phpenv version-name)/bin/phpunit https://phar.phpunit.de/phpunit-8.phar
     - phpunit --version
     # We need npm for eslint
     - nvm install 12 --lts
--- a/SETUP/tests/MARCRecordTest.php
+++ b/SETUP/tests/MARCRecordTest.php
@@ -8,7 +8,7 @@ class MARCRecordTest extends PHPUnit\Framework\TestCase
     private $YAZ_ARRAY = NULL;
     private $YAZ_ARRAY_STR = NULL;
 
-    protected function setUp()
+    protected function setUp(): void
     {
         // Load the yaz array from disk
         $yaz_array_b64 = file_get_contents("./data/yaz_array.b64");
--- a/SETUP/tests/NonactivatedUserTest.php
+++ b/SETUP/tests/NonactivatedUserTest.php
@@ -21,7 +21,7 @@ class NonactivatedUserTest extends PHPUnit\Framework\TestCase
         return $register;
     }
 
-    protected function tearDown()
+    protected function tearDown(): void
     {
         foreach($this->createdRecords as $username)
         {
@@ -73,37 +73,29 @@ class NonactivatedUserTest extends PHPUnit\Framework\TestCase
         $this->assertEquals($id, $user->id);
     }
 
-    /**
-     * @expectedException NonexistentNonactivatedUserException
-     */
     public function testLoadInvalidUser()
     {
+        $this->expectException(NonexistentNonactivatedUserException::class);
         new NonactivatedUser("blahblah");
     }
 
-    /**
-     * @expectedException DomainException
-     */
     public function testSetUnsettable()
     {
+        $this->expectException(DomainException::class);
         $user = new NonactivatedUser();
         $user->id = "";
     }
 
-    /**
-     * @expectedException DomainException
-     */
     public function testSetImmutable()
     {
+        $this->expectException(DomainException::class);
         $user = $this->createNonactivatedUser($this->TEST_USERNAME);
         $user->username = "";
     }
 
-    /**
-     * @expectedException NonexistentNonactivatedUserException
-     */
     public function testDeleteNonactivatedUser()
     {
+        $this->expectException(NonexistentNonactivatedUserException::class);
         # Creating and deleting a user show leave us with no user...
         $user = $this->createNonactivatedUser($this->TEST_USERNAME);
         $user->delete();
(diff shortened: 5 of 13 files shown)


chrismiceli:
There is some warnings in phpunit9 about class names not matching file names, but I didn't want to convolute this change any more. We can do that in a subsequent PR.

PHP 7.3 and 7.4 builds failed with
`
Validating charsuite basic-latin...
ERROR: pickerset ¶ has codepoints not in charsuite:
    ©®: U+00a9
Makefile:16: recipe for target 'lint' failed
`

I will need to look into that. Also, weirdly, locally I get this error running tests:
`
1) UserTest::testValidateValidUserNotStrictDifferWhitespace
Failed asserting that false is true.

/home/chrismiceli/dproofreaders/SETUP/tests/UserTest.php:102
`
but I think it is a mysql configuration issue.

cpeel:
> PHP 7.3 and 7.4 builds failed with
> `Validating charsuite basic-latin... ERROR: pickerset ¶ has codepoints not in charsuite: ©®: U+00a9 Makefile:16: recipe for target 'lint' failed`

This is because the code needs the mbstring PHP module to work correctly with 7.3 and later. PHP 7.3 changed from PCRE to PCRE2 and something in that messes up `portable-utf8` when it falls back to regexs.

SETUP/tests/README.md
```diff
@@ -4,13 +4,7 @@ These are phpunit tests.
...
 ## Version requirements
 
-The phpunit tests, as-coded, will work with phpunit 5.7, 6.x, and 7.x. They
-will not work in 8.x or later because the setUp() and tearDown() functions
-do not specify void return types. (See the
-[phpunit8 release notes](https://phpunit.de/announcements/phpunit-8.html).)
-That capability isn't available until PHP 7.1, which the DP code will
-nominally work with, but isn't available on the TEST server or in many
-modern distros.
+The phpunit tests, as-coded, will work with phpunit 6.x, 7.x, 8.x, and 9.x.
```
cpeel, line 7: These will still work with phpunit 6 and 7? I assumed they would only work with 8 and later.

cpeel:
> > PHP 7.3 and 7.4 builds failed with
> > `Validating charsuite basic-latin... ERROR: pickerset ¶ has codepoints not in charsuite: ©®: U+00a9 Makefile:16: recipe for target 'lint' failed`
> 
> This is because the code needs the mbstring PHP module to work correctly with 7.3 and later. PHP 7.3 changed from PCRE to PCRE2 and something in that messes up `portable-utf8` when it falls back to regexs.

@bpfoley points out that this doesn't solve it (and indeed the Travis PHP versions already have [mbstring and intl enabled already](https://github.com/travis-ci/travis-cookbooks/blob/precise-stable/ci_environment/phpbuild/templates/default/default_configure_options.erb)). I saw this on TEST at one point that I thought I solved by ensuring that the PHP cli 7.4 version had the right modules installed but perhaps that wasn't it.

I still feel confident that the reason it breaks on 7.3 and later is the PCRE2 change, but clearly I don't grok how to solve it.

SETUP/tests/README.md
chrismiceli: well for Php 7.4 on my machine at least:
phpunit 6:
image

phpunit 7:
image

ignore the failure, I think it is a db configuration issue.

chrismiceli:
> > > PHP 7.3 and 7.4 builds failed with
> > > `Validating charsuite basic-latin... ERROR: pickerset ¶ has codepoints not in charsuite: ©®: U+00a9 Makefile:16: recipe for target 'lint' failed`
> > 
> > 
> > This is because the code needs the mbstring PHP module to work correctly with 7.3 and later. PHP 7.3 changed from PCRE to PCRE2 and something in that messes up `portable-utf8` when it falls back to regexs.
> 
> @bpfoley points out that this doesn't solve it (and indeed the Travis PHP versions already have [mbstring and intl enabled already](https://github.com/travis-ci/travis-cookbooks/blob/precise-stable/ci_environment/phpbuild/templates/default/default_configure_options.erb)). I saw this on TEST at one point that I thought I solved by ensuring that the PHP cli 7.4 version had the right modules installed but perhaps that wasn't it.
> 
> I still feel confident that the reason it breaks on 7.3 and later is the PCRE2 change, but clearly I don't grok how to solve it.

This is the only one that fails. If you skip the failure, all other character suites pass. I boiled it down to this snippet:
```
echo preg_replace_callback("/\X/u", function ($matches)
  {
    $match = $matches[0];
    if(1 === preg_match("/^(?:[ ¡-¬®-ÿ])$/u", $match))
    {
       return "";
    }
    return $match;
  },
  "©®"
);
echo "\n";
```
In travis and my local machine with php 7.4, it fails returning `©®`. It should return an empty string. The weird thing is, if I var_dump $match, it seems to think '©®' is 1 character. I wonder if \X is treating the registered trademark code point as a combining character. Testing each character separately works! I'm not a php expert though, so not sure how to proceed from here.

cpeel:
> In travis and my local machine with php 7.4, it fails returning `©®`. It should return an empty string. The weird thing is, if I var_dump $match, it seems to think '©®' is 1 character. I wonder if \X is treating the registered trademark code point as a combining character. Testing each character separately works! I'm not a php expert though, so not sure how to proceed from here.

That code is one of the basic building blocks of our Unicode support so something we need to track down. @70ray might have some ideas since he figured out how to make combining characters work for us.

chrismiceli:
> > In travis and my local machine with php 7.4, it fails returning `©®`. It should return an empty string. The weird thing is, if I var_dump $match, it seems to think '©®' is 1 character. I wonder if \X is treating the registered trademark code point as a combining character. Testing each character separately works! I'm not a php expert though, so not sure how to proceed from here.
> 
> That code is one of the basic building blocks of our Unicode support so something we need to track down. @70ray might have some ideas since he figured out how to make combining characters work for us.

Ok, I've simplified the test case even more:
```
echo preg_replace_callback("/\X/u", function ($matches)
  {
    $match = $matches[0];
    var_dump($match);
    if(1 === preg_match("/^(?:[\u{00ae}\u{00a9}])$/u", $match))
    {
       return "";
    }
    return $match;
  },
  "\u{00ae}\u{00a9}"
);
```
In PHP >= 7.3, the var_dump returns the string "©®" which doesn't match the preg_match. In PHP < 7.3, it matches twice with "©" and "®" separately, which both pass the preg_match. The \X pcre in pcre2 (introduced in 7.3) seems to think that "©®" is a single grapheme cluster. The weird thing is, if I use grapheme_extract, it treats it as 2 grapheme clusters in 7.2 & 7.3/7.4.
```
$str = "\u{00ae}\u{00a9}";
$next = 0;
$maxbytes = strlen($str);

while ($next < $maxbytes) {
    $char = grapheme_extract($str, 1, GRAPHEME_EXTR_COUNT, $next, $next);
    if (empty($char)) {
        continue;
    }
    echo "{$char} - This utf8 character is " . strlen($char) . ' bytes long', PHP_EOL;
}
```
outputs:
```
® - This utf8 character is 2 bytes long
© - This utf8 character is 2 bytes long
```
I considered switching the logic in the unicode.inc file, but we use \X in many places so I'm not sure that is trivially achievable. I will keep looking, but any input would be helpful.

cpeel:
Nice investigation!

> I considered switching the logic in the unicode.inc file, but we use \X in many places so I'm not sure that is trivially achievable. I will keep looking, but any input would be helpful.

The only thing we use \X for is to separate graphemes and it looks like we only use that in `unicode.inc` -- so if we can replace all instances of that with `grapheme_extract()` I think that sounds like the right approach.

70ray:
> Nice investigation!
> 
> > I considered switching the logic in the unicode.inc file, but we use \X in many places so I'm not sure that is trivially achievable. I will keep looking, but any input would be helpful.
> 
> The only thing we use \X for is to separate graphemes and it looks like we only use that in `unicode.inc` -- so if we can replace all instances of that with `grapheme_extract()` I think that sounds like the right approach.

I agree. It looks like a bug in \X implementation

chrismiceli:
> > Nice investigation!
> > > I considered switching the logic in the unicode.inc file, but we use \X in many places so I'm not sure that is trivially achievable. I will keep looking, but any input would be helpful.
> > 
> > 
> > The only thing we use \X for is to separate graphemes and it looks like we only use that in `unicode.inc` -- so if we can replace all instances of that with `grapheme_extract()` I think that sounds like the right approach.
> 
> I agree. It looks like a bug in \X implementation

Ok, replaced with grapheme_extract implementation. Please examine it closely. The tests pass without modification though in all PHP versions.

chrismiceli:
Sandbox available here since I'm now modifying unicode.inc: https://www.pgdp.org/~chrismiceli/c.branch/phpunit

cpeel:
Let's get the `grapheme_extract()` commit into a separate MR so we can evaluate & test it independently.

Also, I think we should make `split_graphemes()` a generator so we don't have to build the (possibly very large) array in-memory. Thankfully, [PHP does indeed support them](https://www.php.net/manual/en/language.generators.overview.php).

chrismiceli:
> Let's get the `grapheme_extract()` commit into a separate MR so we can evaluate & test it independently.
> 
> Also, I think we should make `split_graphemes()` a generator so we don't have to build the (possibly very large) array in-memory. Thankfully, [PHP does indeed support them](https://www.php.net/manual/en/language.generators.overview.php).

I love generators in javascript. I didn't know they exist in php. Let me update that in the other MR.

PR #511, opened 2020-11-05
travis.yml: CREATE mysql user and GRANT privileges separately

bpfoley:
GRANT ... ON ... TO ... IDENTIFIED BY ...; is deprecated in MySQL 5.7
(and errors in MySQL 8 on Ubuntu 20.10)


chrismiceli (approved):

cpeel:
We need to update `SETUP/INSTALL.md` as well then and should do so as part of this commit since they're related.

bpfoley:
> We need to update `SETUP/INSTALL.md` as well then and should do so as part of this commit since they're related.

Good point. Done.

srjfoo (approved):

PR #512, opened 2020-11-05
travis.py: Typo, should invoke 'npm' not 'nvm'.

bpfoley:
I'm guessing this was working because npm is
already installed on the Travis workers.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ before_script:
     - curl -sSfL -o ~/.phpenv/versions/$(phpenv version-name)/bin/phpunit https://phar.phpunit.de/phpunit-6.phar
     - phpunit --version
     # We need npm for eslint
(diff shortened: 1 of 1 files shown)


chrismiceli (commented):
I think this is expected. I'll have to get back to you when I have access to a computer but this allows us to use specific node versions when running lint/tests.

chrismiceli (commented):
Ok, at a computer. The configuration is as it should be. We use nvm to install node version 12 for us. Once it is installed, we can use npm. You can see your checks vailed. If you're not familiar, nvm is a tool to download and manage node versions: https://github.com/nvm-sh/nvm. Usually in travis you could just set the language to node_js, but we use 2 languages in our travis file, php and node, so we have to do it this way last I checked.

bpfoley:
Ah, thankyou. My bad :)

PR #513, opened 2020-11-05
Use language script as part of spellcheck

cpeel:
When presented with text containing two different scripts, aspell behaves differently based on the dictionary. Words containing non-Latin characters are ignored completely with the English dictionary, yet words containing Latin characters are flagged as spelling errors with the Ancient Greek dictionary.

To fix this, restrict misspelled words (as returned from aspell) to only words that contain the script of the language we're checking. This ensures that only words with Latin characters are returned when checking against the English or French dictionary, and only Greek words are returned when using the Ancient Greek dictionary.

This is testable in the [wordcheck-script-support](https://www.pgdp.org/~cpeel/c.branch/wordcheck-script-support/) sandbox.

### DIFF
--- a/pinc/iso_lang_list.inc
+++ b/pinc/iso_lang_list.inc
@@ -56,6 +56,43 @@ function get_languages_with_dictionaries() {
     return $returnArray;
 }
 
+// Given a langcode, return the Unicode script that language is written in.
+// Will return "Latin" if the langcode is not known.
+function get_script_for_langcode($langcode)
+{
+    // This, decidedly incomplete, mapping was created by taking the list of
+    // available aspell 0.60 dictionaries and running the "Native Name" through
+    // utf8_string_scripts().
+    $langcode_script_map = [
+        "am" => "Ethiopic",
+        "ar" => "Arabic",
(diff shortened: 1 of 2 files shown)


chrismiceli (approved):

pinc/iso_lang_list.inc
```diff
@@ -56,6 +56,43 @@ function get_languages_with_dictionaries() {
     return $returnArray;
 }
 
+// Given a langcode, return the Unicode script that language is written in.
+// Will return "Latin" if the langcode is not known.
```
srjfoo, line 60: Would it be slightly clearer to say 'Will return "Latin" for any langcode that doesn't have an entry in `$langcode_script_map`'?

srjfoo (approved):

pinc/iso_lang_list.inc
cpeel: It would be clearer for this implementation but less useful in the future if the implementation changes and the comment isn't updated (which happens surprisingly often). The function is pretty simple to begin with so I don't think this is necessary here.

PR #515, opened 2020-11-06
Test validate

70ray:
The Astral test is omitted because it fails without the u flag which we can't use with Internet explorer. I raised thsi issue with XRegexp on gitHub.
slevithan replied:

> XRegExp requires and defers to native handling for flag u and the astral ranges that u enables within character classes. As you mentioned, IE doesn't support flag u natively, so this won't work there.
> 
> Would it be possible to approach this a different way, using Unicode scripts, categories, blocks, or properties, rather than an explicit astral range? E.g. XRegExp('^(?:\\p{EgyptianHieroglyphs}|[\n\r])$', 'A').test('𓀀');

but that won't fit in with our present procedure.

### DIFF
--- a/scripts/character_test.js
+++ b/scripts/character_test.js
@@ -8,23 +8,23 @@
 const charMatch = XRegExp("\\PM\\pM*", "Ag");
 
 // this regular expression (constructed below) matches individual good characters
-var goodChar;
+var validCharRegex;
 
-function makeGoodCharRegex(characterPattern) {
+function makeValidCharRegex(characterPattern) {
     // IE HACK - IE11 does not support u flag
     return XRegExp(characterPattern, "A");
     // need the u flag for Astral plane characters.
-    // return XRegExp(validCharacterPattern, "Au");
+    // return XRegExp(characterPattern, "Au");
 }
 
 $(function () {
     // need to define this after the page has loaded so validCharacterPattern
     // is available
-    goodChar = makeGoodCharRegex(validCharacterPattern);
+    validCharRegex = makeValidCharRegex(validCharacterPattern);
 });
 
 function testChar(character) {
-    return goodChar.test(character);
+    return validCharRegex.test(character);
 }
 
 // return false if text contains any bad characters
--- a/scripts/character_test.js
+++ b/scripts/character_test.js
@@ -10,17 +10,20 @@ const charMatch = XRegExp("\\PM\\pM*", "Ag");
 // this regular expression (constructed below) matches individual good characters
 var goodChar;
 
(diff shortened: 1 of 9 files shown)


scripts/character_test.js
```diff
@@ -1,26 +1,20 @@
...
 
-// this regular expression (constructed below) matches individual good characters
-var goodChar;
-
-$(function () {
-    // need to define this after the page has loaded so validCharacterPattern
-    // is available
-    goodChar = XRegExp(validCharacterPattern, "A");
-});
+function makeGoodCharRegex(validCharacterPattern) {
```
cpeel, line 10: `makeValidCharRegex()` would be a better name.

cpeel (commented):
Can we do this in some way that doesn't break out `good_char.js`? It's a terribly small JS file that does one specific thing that the browser has to do a request for.

scripts/good_char.js
```diff
@@ -0,0 +1,9 @@
+/* global $ makeGoodCharRegex validCharacterPattern */
+/* exported goodChar */
+
+var goodChar;
+$(function () {
+    // need to define this after the page has loaded so validCharacterPattern
```
chrismiceli, line 6: should this just be at the end of the body then? Does our php js data support that?

cpeel: > Does our php js data support that?

It does not. All JS data is inserted in the header.

70ray:
> Can we do this in some way that doesn't break out `good_char.js`? It's a terribly small JS file that does one specific thing that the browser has to do a request for.

We could put it in "onLoad" in the "body_attributes". I'll try it.

cpeel:
Why can't we keep it where it is?

cpeel:
> Why can't we keep it where it is?

To rephrase: moving this into an external file and then needing to touch every place we want to use the validator results in a lot of churn and code complexity solely to add tests. I have to believe there is a better way to have the tests work without all of the churn in all of the files that use the validator.

70ray:
> Can we do this in some way that doesn't break out `good_char.js`? It's a terribly small JS file that does one specific thing that the browser has to do a request for.

We could put it in "onLoad" in the "body_attributes". I'll try it.

> > Why can't we keep it where it is?
> 
> To rephrase: moving this into an external file and then needing to touch every place we want to use the validator results in a lot of churn and code complexity solely to add tests. I have to believe there is a better way to have the tests work without all of the churn in all of the files that use the validator.

You are right!

scripts/character_test.js
70ray: Done this.

chrismiceli (approved):

PR #518, opened 2020-11-06
pp_stats.php: Sprinkle DPDatabase::escape()

bpfoley:
More cleanup needed, but we should start somewhere.


bpfoley:
Sandbox at https://www.pgdp.org/~bfoley/c.branch/cleanup22/stats/pp_stats.php

chrismiceli (approved):

PR #521, opened 2020-11-10
avoid \X due to issues splitting graphemes correctly

chrismiceli:
Split into separate MR to increase confidence in its correctness.

### DIFF
--- a/pinc/unicode.inc
+++ b/pinc/unicode.inc
@@ -110,6 +110,7 @@ function split_graphemes($string)
 # Split a Unicode string into codepoints keeping combining characters together
 function utf8_codepoints_combining($string)
 {
+    $codepoints = [];
     foreach(split_graphemes($string) as $grapheme)
     {
         $codepoints[] = string_to_codepoints_string($grapheme, ">");
--- a/pinc/unicode.inc
+++ b/pinc/unicode.inc
@@ -96,14 +96,23 @@ function utf8_character_name($characters)
     return implode(" + ", $title_pieces);
 }
 
+function split_graphemes($string)
+{
+    $next = 0;
+    $maxbytes = strlen($string);
+
+    while ($next < $maxbytes)
+    {
+        yield grapheme_extract($string, 1, GRAPHEME_EXTR_COUNT, $next, $next);
+    }
+}
+
 # Split a Unicode string into codepoints keeping combining characters together
 function utf8_codepoints_combining($string)
 {
-    $codepoints = [];
-    preg_match_all("/(\X)/u", $string, $matches);
-    foreach($matches[1] as $character)
+    foreach(split_graphemes($string) as $grapheme)
     {
-        $codepoints[] = string_to_codepoints_string($character, ">");
+        $codepoints[] = string_to_codepoints_string($grapheme, ">");
     }
     return $codepoints;
 }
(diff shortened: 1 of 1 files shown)


chrismiceli:
Available in testing here: https://www.pgdp.org/~chrismiceli/c.branch/pcre-replace

cpeel:
Some context for testing: these functions are the cornerstone of our Unicode character enforcement, including support for combining characters. When testing, we should include combining, BMP, and Astral characters.

The easiest way to exercise these is probably going to be through `add_files.php` and ensuring that the expected invalid characters are flagged *and removed* if loaded. Testing through the proofreading UI isn't going to be effective since invalid characters will get flagged via JS before hitting these functions.

Another testing approach is to add invalid characters (like tabs) to a character suite and ensure `lint_charsuites.php` catches it.

cpeel:
> Testing through the proofreading UI isn't going to be effective since invalid characters will get flagged via JS before hitting these functions.

If it would ease testing, we can also temporarily update the testing sandbox to bypass the JS character validation to allow testing via the proofreading interface.

srjfoo:
> Some context for testing: these functions are the cornerstone of our Unicode character enforcement, including support for combining characters. When testing, we should include combining, BMP, and Astral characters.
> 
> The easiest way to exercise these is probably going to be through `add_files.php` and ensuring that the expected invalid characters are flagged _and removed_ if loaded. Testing through the proofreading UI isn't going to be effective since invalid characters will get flagged via JS before hitting these functions.
> 
> Another testing approach is to add invalid characters (like tabs) to a character suite and ensure `lint_charsuites.php` catches it.

I hacked [pcre-replace/tools/charsuites.php?charsuite=semitic-and-indic](https://www.pgdp.org/~chrismiceli/c.branch/pcre-replace/tools/charsuites.php?charsuite=semitic-and-indic) to include a couple of astral plane characters, some random characters in the BMP, a horizontal tab and a vertical tab, and a character with 3 combining diacritics, as well as a character with a single combining diacritic that should be normalized. (Should probably test one that will normalize that isn't in the Basic Latin charsuite, too.) Note that this will disappear when the sandbox is regenerated.

`add_files.php` had an unexpected result. Compare what was loaded via the [pcre-replace](https://www.pgdp.org/~chrismiceli/c.branch/pcre-replace/tools/project_manager/downloadproofed.php?project=projectID5136454928681&image=105.png&round_num=0) sandbox with what was loaded via [my sandbox](https://www.pgdp.org/c/tools/project_manager/downloadproofed.php?project=projectID5f97a65fc8225&image=105.png&round_num=0)

The whole purpose for having the JS character validation is to give the proofers and formatters the opportunity to remove any "bad" characters, themselves, instead of having characters disappear that they didn't realize were bad and won't realize were removed upon save -- it might be a good idea to disable that for the purpose of testing so that we can make sure what's supposed to happen is happening.

70ray (approved):

pinc/unicode.inc
```diff
@@ -96,14 +96,28 @@ function utf8_character_name($characters)
...
 
+function split_graphemes($string)
+{
+    $next = 0;
+    $maxbytes = strlen($string);
+
+    while ($next < $maxbytes)
+    {
+        $char = grapheme_extract($string, 1, GRAPHEME_EXTR_COUNT, $next, $next);
+        if (empty($char))
```
cpeel, line 107: This is the source of our missing-zero problem. From [PHP docs for empty()](https://www.php.net/manual/en/function.empty.php):
```
The following values are considered to be empty:
* "" (an empty string)
* 0 (0 as an integer)
* 0.0 (0 as a float)
* "0" (0 as a string)
* NULL
* FALSE
* array() (an empty array)
```

Are we certain this check is needed? I see it in all of the examples but I'm unclear on WHY it is needed it.

70ray: I don't know either.

chrismiceli: I just updated to allow "0".

cpeel: The more I look at the `empty()` check in the PHP doc user notes the more I think it's just wrong and isn't needed at all.

pinc/unicode.inc
```diff
@@ -96,14 +96,28 @@ function utf8_character_name($characters)
...
 
+function split_graphemes($string)
+{
+    $next = 0;
+    $maxbytes = strlen($string);
+
+    while ($next < $maxbytes)
+    {
+        $char = grapheme_extract($string, 1, GRAPHEME_EXTR_COUNT, $next, $next);
+        if (isset($char) && strlen($char) > 0)
```
70ray, line 107: shouldn't it be just

`yield grapheme_extract($string, 1, GRAPHEME_EXTR_COUNT, $next, $next);`

srjfoo:
I've tested again via `add_files.php`, and got the desired results (i.e. all zeroes were in place). Should I re-do the charsuite test, also?

cpeel:
> I've tested again via `add_files.php`, and got the desired results (i.e. all zeroes were in place). Should I re-do the charsuite test, also?

From a sense of paranoia given this change, yes please.

PR #523, opened 2020-11-14
More color tweaks

srjfoo:
In doing a series of checks to see if we were ready to actually start wider viewing of a new theme, I found a few more styling problems that needed to be resolved. Hopefully this is the last of them (at least for now).

Test in [color-tweaks](https://www.pgdp.org/~srjfoo/c.branch/color-tweaks/)

The only significant visual difference should be in the list of quiz default messages. Compare: [default-quiz-messages-before](https://www.pgdp.org/c/quiz/generic/wizard/default_messages.php) with [default-quiz-messages-after](https://www.pgdp.org/~srjfoo/c.branch/color-tweaks/quiz/generic/wizard/default_messages.php)

### DIFF
--- a/quiz/generic/wizard/default_messages.php
+++ b/quiz/generic/wizard/default_messages.php
@@ -4,12 +4,9 @@ include_once($relPath.'base.inc');
 include_once($relPath.'theme.inc');
 include_once('../quiz_defaults.inc'); // $messages
 
-$theme_args["css_data"] = "h2 {font-size:110%; margin: 0;}
-    p.message_id {margin: 0 0 .5em 1em; background-color: #dddddd;}";
-
 require_login();
 
-output_header(_('Quiz Messages'), SHOW_STATSBAR, $theme_args);
+output_header(_('Quiz Messages'));
 
 echo "<h1>" . _("Default Quiz Messages") . "</h1>";
 
@@ -18,7 +15,7 @@ echo "<p>" . _("(By the way, you should <b>not</b> name any of your custom messa
 
 foreach ($messages as $key => $message)
 {
-    echo "<hr><p class='message_id'>" . _("Message ID:") . " $key</p>\n";
+    echo "<hr><p>" . _("Message ID:") . " $key</p>\n";
     echo "<h2>" . $message['message_title'] . "</h2>\n";
     echo "<p>" . $message["message_body"] . "</p>\n";
     if (isset($message['wiki_ref']))
--- a/styles/page_interfaces.less
+++ b/styles/page_interfaces.less
@@ -20,9 +20,23 @@
 @control-pane-text: black;
 @control-pane-menu: Bisque;
 
+/* Links */
+@page-interface-link-color: #0000ff;
+@page-interface-visited-color: #990099;
+@page-interface-active-color: #ff0000;
+
 .page-interface {
     background-color: @page-interface-background;
     color: @page-interface-text;
+    a:link {
+        color: @page-interface-link-color;
+    }
+    a:visited {
+        color: @page-interface-visited-color;
+    }
+    a:active {
+        color: @page-interface-active-color;
+    }
 
     .text-pane {
         background-color: @text-pane-background;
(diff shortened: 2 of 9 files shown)


quiz/generic/wizard/default_messages.php
```diff
@@ -4,12 +4,9 @@
 include_once($relPath.'theme.inc');
 include_once('../quiz_defaults.inc'); // $messages
 
-$theme_args["css_data"] = "h2 {font-size:110%; margin: 0;}
-    p.message_id {margin: 0 0 .5em 1em; background-color: #dddddd;}";
```
cpeel, line 8: Let's remove the reference to this on line 18 as well.

srjfoo: Yes, let's! And thanks for pointing that out. I totally forgot to look for it.

Done!

chrismiceli (approved):

srjfoo:
I realized that I hadn't styled the text-box border for words that WC identifies as needing attention -- once any kind of style is forced (like the background and text color), one needs to make sure that the border style is as desired, also. The latest push incorporates that change, and also adds `.x-thin-border` as a defined mixin. The only place this is currently used is in the WC interface.

styles/page_interfaces.less
```diff
@@ -163,6 +186,22 @@
...
+                .x-thin-border(@border-color-base-lowc);
+            }
+        }
+    }
+    #wc_header {
+        input, select {
+            background-color: @page-background;
+            color: @page-text;
+            .solid-border;
+            .button;
```
cpeel, line 203: I don't think we want the button style on the drop-down -- just on the button. I also think we want to tweak the `.button` class so it looks more like our other buttons (now that we set styles for buttons). Consider:
```diff
diff --git a/styles/global.less b/styles/global.less
index 1fd54d2b1..58f8d253d 100644
--- a/styles/global.less
+++ b/styles/global.less
@@ -116,20 +116,6 @@
     visibility: hidden;
 }

-.button {
-    /*
-     * We can't use appearance: button here because Chrome has deprecated it:
-     * https://developers.google.com/web/updates/2019/12/chrome-80-deps-rems
-     * so we make a "button like object" that won't really resemble the other
-     * buttons rendered, but still looks like a button.
-     */
-    border: solid 1px lightgray;
-    border-radius: 5px;
-    font-size: 0.7rem;
-    padding: 2px 6px;
-    cursor: default;
-}
-
 .nodisp {
     display: none;
 }
diff --git a/styles/layout.less b/styles/layout.less
index d8db95956..0b9c1647e 100644
--- a/styles/layout.less
+++ b/styles/layout.less
@@ -98,7 +98,21 @@
 }

 .button {
+    /*
+     * We can't use appearance: button here because Chrome has deprecated it:
+     * https://developers.google.com/web/updates/2019/12/chrome-80-deps-rems
+     * so we make a "button like object" that won't really resemble the other
+     * buttons rendered, but still looks like a button.
+     */
+    border-radius: 3px;
+    cursor: default;
     .sans-serif;
+    .solid-border;
+}
+
+span.button {
+    font-size: 0.8em;
+    padding: 1px 4px;
 }

 /* ------------------------------------------------------------------------ */
diff --git a/styles/page_interfaces.less b/styles/page_interfaces.less
index aaf2e5a97..6d88d7252 100644
--- a/styles/page_interfaces.less
+++ b/styles/page_interfaces.less
@@ -200,6 +200,8 @@
             background-color: @page-background;
             color: @page-text;
             .solid-border;
+        }
+        button, input[type=submit], input[type=button] {
             .button;
         }
     }
```

srjfoo: Thanks, done.

Button styling moved from `global.less` to `layout.less`, and styling tweaked.

Fair point about not styling the dropdown as a button. That was sloppy of me. Turns out that FF, Chrome and Safari all have different default button looks, and I was trying to match what I was seeing in FF (both the drop-down and the button have a border radius). The dropdown no longer uses the button styling.

PR #524, opened 2020-11-17
1889: MRU picker can contain characters not in enabled character suite

chrismiceli:
filter out characters not present in character picker

### DIFF
--- a/tools/proofers/character_selector.js
+++ b/tools/proofers/character_selector.js
@@ -38,6 +38,16 @@ $(function () {
         mru = [];
     }
 
+    var pickerChars = {};
+    $('.picker').each(function pickerEach() {
+        pickerChars[$(this).text()] = true;
+    });
+
+    // filter out invalid characters
+    mru = mru.filter(function mruFilter(mruCharacter) {
+        return pickerChars[mruCharacter.character] === true;
+    });
+
     window.addEventListener('beforeunload', function() {
         localStorage.setItem(storageKey, JSON.stringify(mru));
     });
--- a/tools/proofers/character_selector.js
+++ b/tools/proofers/character_selector.js
@@ -39,12 +39,12 @@ $(function () {
     }
 
     var pickerChars = {};
-    $('.picker').each(function pickerEach() {
+    $('.picker').each(function() {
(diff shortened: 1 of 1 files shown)


chrismiceli:
Sandbox here https://www.pgdp.org/~chrismiceli/c.branch/issue/1889/

srjfoo (approved):
Looks good to me.

I tested by adding Polytonic Greek to a project, proofing a page and adding Greek using the character picker, verifying that the MRU had the Greek characters, saving, removing the Polytonic Greek charsuite and verifying that only the Greek characters were removed from the MRU.

If there's any other way it needs to be tested, let me know.

chrismiceli:
> Looks good to me.
> 
> I tested by adding Polytonic Greek to a project, proofing a page and adding Greek using the character picker, verifying that the MRU had the Greek characters, saving, removing the Polytonic Greek charsuite and verifying that only the Greek characters were removed from the MRU.
> 
> If there's any other way it needs to be tested, let me know.

You could use browser developer tools to modify localStorage but what you did is the real world scenario. Thanks for testing!

tools/proofers/character_selector.js
```diff
@@ -38,6 +38,16 @@ $(function () {
...
+    var pickerChars = {};
+    $('.picker').each(function pickerEach() {
+        pickerChars[$(this).text()] = true;
+    });
+
+    // filter out invalid characters
+    mru = mru.filter(function mruFilter(mruCharacter) {
+        return pickerChars[mruCharacter.character] === true;
+    });
+
```
70ray, line 50: The functions could be anonymous.

70ray (approved):

tools/proofers/character_selector.js
chrismiceli: Sure, I am used to naming them for better stack traces, but this code isn't too complex.

srjfoo:
Re-tested after the code update, with different pickersets but the same results: MRU loses the characters that are in a character suite once it's removed. Re-did test with custom characters and the same results. Once the Project custom characters are removed, any instances of them in the MRU are also removed.

PR #525, opened 2020-11-18
Make task details render better on narrow screens

cpeel:
For narrows screens, such a mobile devices, make the task details transition to rows instead of columns using a flex layout. Task 1960

Bonus: make sure that verylongstringswithoutspacesincomments break instead of forcing the window very (very) wide.

Testable in [task-center-flex-layout](https://www.pgdp.org/~cpeel/c.branch/task-center-flex-layout/) sandbox.

### DIFF
--- a/styles/layout.less
+++ b/styles/layout.less
@@ -1518,12 +1518,22 @@ div.task-comment {
     margin-bottom: 1em;
     .task-comment-body {
         margin-left: 1em;
+        word-break: break-word;
     }
+}
+
+div.task-detail {
+    display: flex;
+    align-items: flex-start;
+    width: 100%;
 }
 
 table.task-detail-block {
     .solid-border;
-    width: 100%;
+    margin: 0 0.1em;
+    width: 50%;
+    flex-grow: 1;
+    flex-basis: auto;
     tr {
         th, td {
             padding: 0.1em 0.5em 0.1em 0.5em;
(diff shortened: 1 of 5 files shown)


srjfoo:
Something isn't quite right in the computer browser: 
image

The left-hand section appears to be the full-width of the comment section. If I make the window narrow enough, it wraps below, but otherwise not.

cpeel:
> The left-hand section appears to be the full-width of the comment section. If I make the window narrow enough, it wraps below, but otherwise not.

Only seems to be in Firefox -- not happening in Chrome. I'll see about fixing it for FF and testing it in Safari.

srjfoo:
> > The left-hand section appears to be the full-width of the comment section. If I make the window narrow enough, it wraps below, but otherwise not.
> 
> Only seems to be in Firefox -- not happening in Chrome. I'll see about fixing it for FF and testing it in Safari.

Weird. And yes, I see now that it doesn't happen in either Safari or Chrome. (Test in all your browsers, Sharon!)

cpeel:
I think I have a version that now works in Chrome, Safari, Firefox, Edge, and IE11. Firefox is the real PITA here, oddly. Even IE11 played nice.

srjfoo (approved):
Looks good.

chrismiceli (approved):

PR #529, opened 2020-11-27
Prevent racing add_files processes

cpeel:
Periodically we see a slew of errors in `php_errors` like:
```
DPDatabase.inc - log_error from /data/htdocs/c/pinc/DPage.inc:149: Duplicate entry '323' for key 'fileid'
```
This is called upstream by `add_files.php` which is used to load pages into a project. The error is that there is already a row in the project table for that page, but `add_files.php` checks for that so how is this possible? I believe this happens when a user clicks on the "Proceed with load" link and then, when it doesn't look like anything is happening (because it takes a while), they click on it again causing two `add_files.php` processes to race.

To prevent this, make the "Proceed with load" a POST and disable the submit button when it has been clicked.

As part of figuring this out I styled some of the error messages and ensured we weren't leaking filesystem details in `_do_command()` but instead log the errors to `php_errors`.

The sandbox to test this is [add-pages-errors](https://www.pgdp.org/~cpeel/c.branch/add-pages-errors/) and the focus should be on ensuring that existing page loads work as-expected.

### DIFF
--- a/tools/project_manager/add_files.php
+++ b/tools/project_manager/add_files.php
@@ -64,7 +64,9 @@ else
     // Prevent sneaky parent-link tricks.
     if (str_contains($rel_source, ".."))
     {
-        echo "Source directory '$rel_source' is not acceptable.";
+        echo "<p class='error'>";
+        echo sprintf(_("Source directory '%s' is not acceptable."), html_safe($rel_source));
+        echo "</p>";
         echo "<hr>\n";
         echo "Return to <a href='$code_url/project.php?id=$projectid'>Project Page</a>.\n";
         return;
@@ -135,7 +137,9 @@ else
 $r = chdir($source_project_dir);
 if ( !$r )
 {
-    echo "Directory '$source_project_dir' does not exist, or is inaccessible.\n";
+    echo "<p class='error'>";
+    echo sprintf(_("Directory '%s' does not exist, or is inaccessible."), html_safe($source_project_dir));
+    echo "</p>";
     echo "<hr>\n";
     echo "Return to <a href='$code_url/project.php?id=$projectid'>Project Page</a>.\n";
     return;
@@ -1011,8 +1015,11 @@ class Loader
             system($cmd, $exit_status);
             if ( $exit_status != 0 )
             {
-                echo "$cmd:<br>";
-                echo "exit status was $exit_status<br>";
+                error_log("add_files.php - error running \"$cmd\"; exit code: $exit_status");
+
+                echo "<p class='error'>";
+                echo "an error occurred during a filesystem operation, this has been logged for a site administrator to view";
+                echo "</p>";
             }
         }
     }
--- a/tools/project_manager/add_files.php
+++ b/tools/project_manager/add_files.php
@@ -19,12 +19,22 @@ $extra_args = [
             font-family: ' . get_proofreading_font_family_fallback() . ';
         }
     ',
+    'js_data' => '
+        $(document).ready(function () {
+            $("#execute").submit(function (e) {
+                // disable the submit button
+                $("#submit").attr("disabled", true);
+                return true;
+            });
+        });
+    ',
 ];
 
 $title = _("Add files");
 output_header($title, NO_STATSBAR, $extra_args);
 
-$projectid    = get_projectID_param($_GET, 'project');
+$projectid    = get_projectID_param($_REQUEST, 'project');
+$rel_source   = array_get($_REQUEST, "rel_source", "");
 $loading_tpnv = (@$_GET['tpnv'] == '1');
 
 abort_if_cant_edit_project( $projectid );
@@ -54,13 +64,12 @@ if(!user_can_add_project_pages($projectid, $loading_tpnv == 1 ? "tp&v" : "normal
     exit();
 }
 
-if ( $_GET['rel_source'] == '' )
+if ( $rel_source == '' )
 {
     die('rel_source parameter is empty or unset');
 }
 else
 {
-    $rel_source = $_GET['rel_source'];
     // Prevent sneaky parent-link tricks.
     if (str_contains($rel_source, ".."))
     {
@@ -171,7 +180,7 @@ else
     $loader = new Loader( $source_project_dir, $dest_project_dir, $projectid );
     $loader->analyze();
 
-    if ( !@$_GET['confirmed'] )
+    if ( !@$_POST['confirmed'] )
     {
         $loader->display();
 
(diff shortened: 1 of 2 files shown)


tools/project_manager/add_files.php
```diff
@@ -19,12 +19,22 @@
             font-family: ' . get_proofreading_font_family_fallback() . ';
         }
     ',
+    'js_data' => '
+        $(document).ready(function () {
+            $("#execute").submit(function (e) {
+                // disable the submit button
+                $("#submit").attr("disabled", true);
+                return true;
```
chrismiceli, line 27: This may make sense for a ton of our submits. Should we abstract it some?

chrismiceli (approved):

tools/project_manager/add_files.php
cpeel: Abstract it such that all forms have their submit buttons disabled upon click or abstract it such that they _could_ be disabled without adding this to individual pages?

My gut says to not proactively abstract until we know of where else we want to use it so we know how best to abstract it. I'm open to ideas!

chrismiceli: Definitely the latter. Just so it is easy to enable for a given form if we think it is necessary, perhaps with some data- attribute. But we can cross that bridge perhaps when we encounter another place we think warrants this change. Especially if it needs to be reenabled under some circumstances such as client side errors.

srjfoo: It would be useful for Remote File Manager's `Upload` button to appear disabled once it's pressed. There can be a significant delay between pressing the button and getting the message showing the progress of validity checking and virus checking.

cpeel: I thought that was fixed? In Chrome after I it Upload and the file is uploaded it immediately gives me a Upload status page as it's doing the checking. Maybe it works differently in different browsers. Regardless, we wouldn't be able to use this as-is for the RFM as the Upload button isn't a form Submit button.

srjfoo:
I was able to deliberately generate a batch of error messages:
```for base=382, project_add_page raised a DB error``` for bases 382-431 (there were 432 text/png pairs). I'm no longer certain exactly how I did it, because I tried again and did not see the same results. Pretty sure it had to do with using the browser refresh, because the submit button was definitely disabled (I'm assuming that the first page to trigger an error message was 382 because the script was in the process of loading that page when I refreshed the page).

I don't know if that's worth worrying about -- hopefully the disabled button will be a sufficient deterrent. (I can try to repeat the problem tomorrow if you think it's worth doing.)

Of more concern, when I first tried to load the project with my initial upload (`test.zip`), I got the following message:
```source does not exist: /home/uploads/incoming/Users/srjfoo/test.zip```

The check to make sure that the file exists appears to now be being done *after* the zip has been unzipped into a directory and the original zip discarded. I tested just using the unzipped directory name, and the images/text loaded just fine with that, but it should also accept the zipped file, to cause the least confusion with the users, since that's the way it works now. I.e. first time, you specify the zip filename, subsequent project loads with the same source, the zip no longer exists, so you have to use the directory name. I suspect that most PMs rarely use anything but the zip name, because if they need to reload, they're most likely to be needing to do it from a new zip, and so most will never use the directory name for loading a project.

tools/project_manager/add_files.php
srjfoo: I'll check again tomorrow -- I thought I noticed a delay, but I may be wrong about that. That also may be the reason we didn't disable it at the time, come to think of it.

cpeel:
> Of more concern, when I first tried to load the project with my initial upload (`test.zip`), I got the following message:
> `source does not exist: /home/uploads/incoming/Users/srjfoo/test.zip`
> 
> The check to make sure that the file exists appears to now be being done _after_ the zip has been unzipped into a directory and the original zip discarded. I tested just using the unzipped directory name, and the images/text loaded just fine with that, but it should also accept the zipped file, to cause the least confusion with the users, since that's the way it works now. I.e. first time, you specify the zip filename, subsequent project loads with the same source, the zip no longer exists, so you have to use the directory name. I suspect that most PMs rarely use anything but the zip name, because if they need to reload, they're most likely to be needing to do it from a new zip, and so most will never use the directory name for loading a project.

None of that has changed and it works now just the same as it did before: the first time a `.zip` is used it extracts it and deletes the zip file, so trying to specify the zip file again will cause that error.

cpeel:
> I was able to deliberately generate a batch of error messages:
> `for base=382, project_add_page raised a DB error` for bases 382-431 (there were 432 text/png pairs). I'm no longer certain exactly how I did it, because I tried again and did not see the same results. Pretty sure it had to do with using the browser refresh, because the submit button was definitely disabled (I'm assuming that the first page to trigger an error message was 382 because the script was in the process of loading that page when I refreshed the page).
> 
> I don't know if that's worth worrying about -- hopefully the disabled button will be a sufficient deterrent. (I can try to repeat the problem tomorrow if you think it's worth doing.)

I think that's fine. It's still possible to cause this race condition to happen -- I'm just hoping the disabled button makes it happen less often.

tools/project_manager/add_files.php
srjfoo: I had a chance to look just now -- and since pressing the ```Upload``` button triggers a "Progress bar" showing the percentage, it shouldn't be an issue, so nevermind.

srjfoo:
> > Of more concern, when I first tried to load the project with my initial upload (`test.zip`), I got the following message:
> > `source does not exist: /home/uploads/incoming/Users/srjfoo/test.zip`
> > The check to make sure that the file exists appears to now be being done _after_ the zip has been unzipped into a directory and the original zip discarded. I tested just using the unzipped directory name, and the images/text loaded just fine with that, but it should also accept the zipped file, to cause the least confusion with the users, since that's the way it works now. I.e. first time, you specify the zip filename, subsequent project loads with the same source, the zip no longer exists, so you have to use the directory name. I suspect that most PMs rarely use anything but the zip name, because if they need to reload, they're most likely to be needing to do it from a new zip, and so most will never use the directory name for loading a project.
> 
> None of that has changed and it works now just the same as it did before: the first time a `.zip` is used it extracts it and deletes the zip file, so trying to specify the zip file again will cause that error.

I just checked again, and it's not working in the `add-pages-errors` sandbox the way it is on TEST. It gives the error on trying to load a zip the first time.

cpeel:
> I just checked again, and it's not working in the `add-pages-errors` sandbox the way it is on TEST. It gives the error on trying to load a zip the first time.

Yup! The problem is that I was re-using the original URL in the form submit, which doesn't work when the original `rel_source` has a zip in it because the `rel_source` changes for the subsequent page. I've updated the form to pass in that and the projectid instad.

tools/project_manager/add_files.php
cpeel: I'm going to defer abstracting this until we have an obvious second need for it. I like the idea but I'm wary of adding the complexity without a second use-case.

srjfoo:
> > I just checked again, and it's not working in the `add-pages-errors` sandbox the way it is on TEST. It gives the error on trying to load a zip the first time.
> 
> Yup! The problem is that I was re-using the original URL in the form submit, which doesn't work when the original `rel_source` has a zip in it because the `rel_source` changes for the subsequent page. I've updated the form to pass in that and the projectid instad.

Looks good! 👍

srjfoo (approved):

PR #530, opened 2020-12-05
Markdown POC

chrismiceli:


chrismiceli:
I can't seem to connect to sandbox to deploy these changes, but here is a screenshot of html being ignored and markdown being rendered:
image

srjfoo:
I played around with it in [this project](https://www.pgdp.org/~chrismiceli/c.branch/markdown/project.php?id=projectID5df912d3cae6b). I didn't completely convert everything, but ran up against what may be one of many complaints PMs will have. Native markdown does not allow for centering (and I just got lazy about converting a second link. :D).

Looks like a good start!

pinc/3rdparty/parsedown/README.md
```diff
@@ -0,0 +1,4 @@
+# Portable UTF-8
```
cpeel, line 1: s/Portable UTF-8/Parsedown/

cpeel (commented):
This is really great!

Looks like it's time to start a broader discussion with Linda and the rest of the SAs about this.

SETUP/db_schema.sql
```diff
@@ -360,6 +360,7 @@ CREATE TABLE `projects` (
   `language` varchar(255) NOT NULL default '',
   `username` varchar(25) NOT NULL default '',
   `comments` text NOT NULL,
+  `unsafe_comment` tinyint(1) NOT NULL default '0',
```
cpeel, line 363: It might be better to have this be `comment_format varchar(8)` and default to `markdown` with `html` being the grandfathered version. This gives us a bit more flexibility if we need to support a different type in the future and also is clearer about what the comment format is. "unsafe" to me is ambiguous and could mean that we should be `html_safe()`ing the comment field if true, which is actually the opposite of what we want.

pinc/comment_inclusions.inc
```diff
@@ -10,9 +11,11 @@ include_once($relPath.'authors.inc');
 // comments are displayed, zipped up for the
 // post-processor, etc.
 //
-function parse_project_comments($comments) {
+function parse_project_comments($project) {
 	global $relPath;
 
+  $comments = $project->comments;
```
cpeel, line 17: Spacing.

pinc/comment_inclusions.inc
```diff
@@ -41,7 +44,14 @@ function parse_project_comments($comments) {
 			$i++;
 		}
 	}
-	return $comments;
+
+	if ($project->unsafe_comment) {
+		return ($comments);
```
cpeel, line 49: Spacing (tab not spaces), also below.

SETUP/db_schema.sql
chrismiceli: completely agree. I knew I'd get feedback here. Thanks.

pinc/comment_inclusions.inc
chrismiceli: These are tabs, unless I'm mistaken as to what you refer.

SETUP/upgrade/15/20201216_alter_projects.php
```diff
@@ -0,0 +1,23 @@
...
+include_once($relPath.'base.inc');
+
+header('Content-type: text/plain');
+
+// ------------------------------------------------------------
+
+echo "Adding comment_format  column to projects table..\n";
+$sql = "
+    ALTER TABLE projects
+        ADD COLUMN comment_format varchar(8) NOT NULL DEFAULT 'html' AFTER comments
```
cpeel, line 12: This default column value here and the one in `db_schema.sql` should match, and I think `markdown` is the correct one. So we should also add an update query in this file to set all existing projects to `html`.

pinc/comment_inclusions.inc
```diff
@@ -41,7 +44,14 @@ function parse_project_comments($comments) {
 			$i++;
 		}
 	}
-	return $comments;
+
+	if ($project->comment_format == 'html') {
```
cpeel, line 48: OH, this file uses tabs and not spaces. Yes, being consistent here is the way to go and I'll follow up with a clean-up commit for this and any other tab-based files later. I thought we'd gotten all of them but clearly not. Sorry for the confusion earlier.

cpeel (commented):
This looks really good, thanks Chris!

srjfoo:
I just realized that before we roll out to PROD, we also need to have a plan for existing biography and template files. More details in Slack.

SETUP/upgrade/15/20201216_alter_projects.php
chrismiceli: ah yes, I confused myself with default value for these projects and default constraint. I tried to google around for a way to alter a table by adding a column with default constraint with different value, but 2 statements should be fine.

cpeel:
We should show the project comment format on the edit project page since some PMs will have some of each.
Perhaps a line right above the project comment `textarea` that says either "Comment format: HTML" or "Comment format: Markdown" where "Markdown" is a link to https://www.pgdp.net/wiki/Markdown. [I'm open to other ways of showing this, that was just what came to mind for me.]

And for the biographies we could keep the `(HTML)` for HTML comments and add a `(Markdown)` for the markdown comments with the link to the above again.

chrismiceli:
> 
> 
> We should show the project comment format on the edit project page since some PMs will have some of each.
> Perhaps a line right above the project comment `textarea` that says either "Comment format: HTML" or "Comment format: Markdown" where "Markdown" is a link to https://www.pgdp.net/wiki/Markdown. [I'm open to other ways of showing this, that was just what came to mind for me.]
> 
> And for the biographies we could keep the `(HTML)` for HTML comments and add a `(Markdown)` for the markdown comments with the link to the above again.

Great idea. We explain templates and biographies in the same place so it makes sense here too.

tools/authors/addbio.php
```diff
@@ -114,11 +117,19 @@
...
 <input type="hidden" name="author_id" value="<?php echo $author_id; ?>">
 <?php
-if (isset($bio_id))
+if (isset($bio_id)) {
     echo '<input type="hidden" name="bio_id" value="' . $bio_id . '">';
+    echo '<input type="hidden" name="bio_format" value="' . $bio_format . '">';
+}
+
+if (!isset($bio_id) || $bio_format === 'markdown') {
+    $comment_format_text = "<a href=''>" . _("Markdown") . "</a>";
```
cpeel, line 126: Let's include the link here: https://www.pgdp.net/wiki/Markdown

SETUP/upgrade/15/20201222_alter_biographies.php
```diff
@@ -0,0 +1,34 @@
...
+        ADD COLUMN bio_format varchar(8) NOT NULL DEFAULT 'markdown' AFTER last_modified
+";
+
+echo "$sql\n";
+
+mysqli_query(DPDatabase::get_connection(), $sql) or die( mysqli_error(DPDatabase::get_connection()) );
+
+// ------------------------------------------------------------4
+
+echo "Setting comment_format for existing biographies to html..\n";
```
cpeel, line 21: s/comment_format/bio_format/

tools/authors/addbio.php
```diff
@@ -22,13 +22,15 @@
...
-    $bio       = $row["bio"];
+    $author_id  = $row["author_id"];
+    $bio        = $row["bio"];
+    $bio_format = $row["bio_format"];
 }
 elseif (isset($_POST['author_id'])) {
     // preview / save
     $author_id = get_integer_param($_POST, 'author_id', null, null, null, TRUE);
     $bio = $_POST['bio'];
+    $bio_format = $_POST['bio_format'];
```
cpeel, line 33: We should enforce the values here:
```php
$bio_format = get_enumerated_param($_POST, 'bio_format', null, ['html', 'markdown']);
```

tools/authors/addbio.php
cpeel: Also, we should not wrap either `HTML` or `Markdown` in `_()` since those are proper nouns and should not be translated differently, so this can simply be:
```
    $comment_format_text = "<a href='https://www.pgdp.net/wiki/Markdown'>Markdown</a>";
```

tools/project_manager/edit_common.inc
```diff
@@ -316,10 +316,12 @@ function maybe_echo_placeholder_option($initial)
 
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 
-function proj_comments_field($comments)
+function proj_comments_field($comments, $field_name, $comment_format)
 {
     $enc_comments = html_safe($comments);
 
+    $comment_format_text = $comment_format === 'html' ? _("HTML") : "<a href='https://www.pgdp.net/wiki/Markdown'>" . _("Markdown") . "</a>";
```
cpeel, line 323: As with bios, no need for `HTML` or `Markdown` to be localized.

tools/project_manager/edit_common.inc
```diff
@@ -316,10 +316,12 @@ function maybe_echo_placeholder_option($initial)
...
+    echo _("Comment format: ") . $comment_format_text . "<br>";
```
cpeel, line 324: We usually try to separate out label strings from formatting in case we re-use the labels elsewhere without the colon, i.e.
```php
    echo _("Comment format") . ": $comment_format_text<br>";
```

tools/authors/addbio.php
chrismiceli: whoops, I left the href to grab the url later and insert. /facepalm.

tools/project_manager/edit_common.inc
```diff
@@ -316,12 +316,31 @@ function maybe_echo_placeholder_option($initial)
 
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 
-function proj_comments_field($comments, $field_name, $comment_format)
+function proj_comments_format($comment_format)
+{
+    $comment_formant_list = array(
```
cpeel, line 321: s/comment_formant_list/comment_format_list/ (also where it is used in the foreach below)

cpeel (commented):
Only had one minor typo comment. This looks and works great for me!

cpeel:
Oh, and you might want to rebase off of upstream/master as there is a merge conflict. Bonus: squash all of your commits together.

chrismiceli:
> Oh, and you might want to rebase off of upstream/master as there is a merge conflict. Bonus: squash all of your commits together.

Absolutely. I'll take care of it tonight.

srjfoo:
I hadn't yet seen the Markdown link as applied to biographies, so I went poking (@chrismiceli, I've added you as an authors_db_manager). The Markdown link appears where it's needed, but one comment: although the Markdown displays quite nicely once it's actually invoked for PCs (whether the PCs are Markdown or HTML), the Markdown preview doesn't work when you're adding/editing a biography. (To be fair, it doesn't preview HTML, either, but only a string of text with no line endings.)

cpeel:
I don't want to invest more effort into the authors/biographies code since it's largely unused. Right now all bios are "previewed" as `html_safe($bio)` which is safe for the site. Not ideal, but good enough for now.

srjfoo:
> I don't want to invest more effort into the authors/biographies code since it's largely unused. Right now all bios are "previewed" as `html_safe($bio)` which is safe for the site. Not ideal, but good enough for now.

I agree with not wanting to put a lot of effort into the authors/biographies code. I tested bio preview for both Markdown and HTML, and neither displayed as it would show in the PCs. Sorry I wasn't clearer. It's something that appeared not to work, and I thought I'd best make sure that was intentional.

I just checked the main test site, and previewing a biography appears to be broken, there, too (at least if preview is supposed to display what the code will look like), so I probably should have checked there before I commented initially!

PR #532, opened 2020-12-08
enable markdown in tasks

chrismiceli:
sandbox: https://www.pgdp.org/~chrismiceli/c.branch/task_markdown

### DIFF
--- a/tasks.php
+++ b/tasks.php
@@ -88,15 +88,6 @@ elseif ($request_method == 'POST')
         'search',
     );
     $action = get_enumerated_param($_POST, 'action', null, $valid_actions);
-    if ($action == 'add_comment') {
-        $add_comment_mode = 'add_comment';
-        if (array_get($_POST, 'preview_comment', NULL)) {
-            $add_comment_mode = 'preview_comment';
-        } elseif (array_get($_POST, 'edit_comment', NULL)) {
-            $add_comment_mode = 'edit_comment';
-            $action = 'show';
-        }
-    }
 }
 else
 {
@@ -654,7 +645,7 @@ function handle_action_on_a_specified_task()
 {
     global $pguser, $requester_u_id;
     global $now_sse;
-    global $action, $tasks_url, $add_comment_mode;
+    global $action, $tasks_url;
 
     // Default 'action' when a task is specified:
     if (is_null($action)) $action = 'show';
@@ -681,7 +672,8 @@ function handle_action_on_a_specified_task()
         $userSettings->remove_value('taskctr_notice', $task_id);
         metarefresh(0, "$tasks_url?task_id=$task_id");
     }
-    if ($action == 'show' || ($action == 'add_comment' && $add_comment_mode == 'preview_comment')) {
+
+    if ($action == 'show') {
         TaskHeader(title_string_for_task($pre_task));
         TaskDetails($task_id);
     }
@@ -821,38 +813,36 @@ function handle_action_on_a_specified_task()
         }
     }
     elseif ($action == 'add_comment') {
-        if ($add_comment_mode !== 'preview_comment') {
-            $comment = trim(array_get($_POST, 'task_comment', ''));
-            if ($comment) {
-                NotificationMail($task_id, "$pguser commented:\n\n$comment");
-                $sql = sprintf("
-                    INSERT INTO tasks_comments (task_id, u_id, comment_date, comment)
-                    VALUES (%d, %d, %d, '%s')",
-                    $task_id,
-                    $requester_u_id,
-                    $now_sse,
-                    DPDatabase::escape($comment)
-                );
-                DPDatabase::query($sql);
-    
-                $sql = sprintf("
-                    UPDATE tasks
-                    SET date_edited = %d, edited_by = %d
-                    WHERE task_id = %d",
-                    $now_sse,
-                    $requester_u_id,
-                    $task_id
-                );
-                DPDatabase::query($sql);
-    
-                // subscribe the user to the task for notifications
-                $userSettings =& Settings::get_Settings($pguser);
-                $userSettings->add_value('taskctr_notice', $task_id);
-    
-                // After posting the comment, we should reload as to clear POST data
-                //   and avoid comments being posted multiple times.
-                metarefresh(0, "$tasks_url?action=show&task_id=$task_id");
-            }
+        $comment = trim(array_get($_POST, 'task_comment', ''));
+        if ($comment) {
+            NotificationMail($task_id, "$pguser commented:\n\n$comment");
+            $sql = sprintf("
+                INSERT INTO tasks_comments (task_id, u_id, comment_date, comment)
+                VALUES (%d, %d, %d, '%s')",
+                $task_id,
+                $requester_u_id,
+                $now_sse,
+                DPDatabase::escape($comment)
+            );
+            DPDatabase::query($sql);
+
+            $sql = sprintf("
+                UPDATE tasks
+                SET date_edited = %d, edited_by = %d
+                WHERE task_id = %d",
+                $now_sse,
+                $requester_u_id,
+                $task_id
+            );
+            DPDatabase::query($sql);
+
+            // subscribe the user to the task for notifications
+            $userSettings =& Settings::get_Settings($pguser);
+            $userSettings->add_value('taskctr_notice', $task_id);
+
+            // After posting the comment, we should reload as to clear POST data
+            //   and avoid comments being posted multiple times.
+            metarefresh(0, "$tasks_url?action=show&task_id=$task_id");
         }
         else {
             ShowError(_("You must supply a comment before clicking Add Comment."), true);
@@ -1468,20 +1458,9 @@ function ShowError($message, $goback = false)
     echo "$message</p>\n";
 }
 
-function TaskComment($Parsedown, $row) {
-  echo "<div class='task-comment'>";
-  $comment_username_link = private_message_link_for_uid($row['u_id']);
-  echo "<b>$comment_username_link - " . date("l d M Y @ g:ia", $row['comment_date']) . "</b>";
-  echo "<br>\n";
-  echo "<div class='task-comment-body'>";
-  echo $Parsedown->text($row['comment']);
-  echo "</div>";
-  echo "</div>";
-}
-
 function TaskComments($tid)
 {
-    global $tasks_url, $add_comment_mode, $now_sse, $requester_u_id;
+    global $tasks_url;
     $sql = sprintf("
         SELECT *
         FROM tasks_comments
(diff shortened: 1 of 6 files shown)


tasks.php
```diff
@@ -1719,6 +1722,7 @@ function property_format_value($property_id, $task_a, $for_list_of_tasks)
     global $tasks_close_array;
     global $tasks_status_array;
     global $versions_array;
+    global $Parsedown;
```
cpeel, line 1725: Why is the global here? I don't think we need it.

cpeel:
This looks great to me!

tasks.php
chrismiceli: whoops, leftover from refactor. Will remove

srjfoo (approved):
Looks good to me.

70ray (commented):
Works well.
We should provide some guidance about markdown and perhaps a link to a site that explains it in detail.
I think we should really have an editing and possibly  preview capability since there is now more scope for making mistakes. I started on a task to implement editing some time ago but postponed doing any more work on it till after the previous style changes and then forgot about it. I'm happy to relinquish it if anyone wants to take it over or I can take it up again.

srjfoo:
> We should provide some guidance about markdown and perhaps a link to a site that explains it in detail.

I agree, and I've got a draft started in my personal space in the TEST wiki [here](https://www.pgdp.org/wiki/User:Srjfoo/drafts). It's very rough, but is a start, and I'm happy for y'all to add to it, change what's there, or tear it out and start all over -- I just wanted to get a start on it.

There are several flavors of Markdown, so any link to an outside source should point to the flavor we use (for instance, in my research, I learned that some flavors actually do allow HTML to do things that Markdown does not do, or that it does confusingly, like using `<br>` instead of two blank spaces at the end of a line to cause a forced line break. Or using inline styling to get small caps.

> I think we should really have an editing and possibly preview capability since there is now more scope for making mistakes.

I would suggest puting a limitation on the ability to edit comments -- say a 24-hour window during which the comment can be edited by the user, with an override for siteadmins to be able to edit a comment on request, if needed, after that window closes.

70ray:
> I would suggest puting a limitation on the ability to edit comments -- say a 24-hour window during which the comment can be edited by the user, with an override for siteadmins to be able to edit a comment on request, if needed, after that window closes.

I was thinking of adding a field to the task comments table in the database to indicate if a comment had been edited or deleted so a comment could be marked as edited (or deleted if all the text had been removed or it was explicitly deleted).

cpeel:
> > I would suggest puting a limitation on the ability to edit comments -- say a 24-hour window during which the comment can be edited by the user, with an override for siteadmins to be able to edit a comment on request, if needed, after that window closes.
> 
> I was thinking of adding a field to the task comments table in the database to indicate if a comment had been edited or deleted so a comment could be marked as edited (or deleted if all the text had been removed or it was explicitly deleted).

I think having a field that indicates it was edited makes sense, but I still want to impose a window that people can make edits and 24 hours makes sense. Tasks stay around for years and if someone goes in and edits their message in the middle of a discussion months later it could render the task useless. I think we should also allow users to simply delete their comment in that same window and simply remove the entry from the DB entirely rather than show that a message had been there before.

chrismiceli:
> 
> 
> Works well.
> We should provide some guidance about markdown and perhaps a link to a site that explains it in detail.
> I think we should really have an editing and possibly preview capability since there is now more scope for making mistakes. I started on a task to implement editing some time ago but postponed doing any more work on it till after the previous style changes and then forgot about it. I'm happy to relinquish it if anyone wants to take it over or I can take it up again.

I went ahead and added a preview capability. This allows you to preview the markdown, and edit it if necessary before adding the comment. I think this should mitigate concerns over formatting some.

chrismiceli:
Preview mode:
image

viewing actual preview:
image

70ray (commented):
Looks good.
Would it be possible to  make it scroll down to the comment when previewing? (and adding?) in case the new comment is at the end of a long series of comments.

cpeel:
> I went ahead and added a preview capability. This allows you to preview the markdown, and edit it if necessary before adding the comment. I think this should mitigate concerns over formatting some.

I think this adds a bit too much code complexity for the functionality it enables. If, instead, we enabled comment editing a user could add a comment, and if it didn't look like they wanted, they could then edit it. Ray's idea of each comment having an anchor and jumping to that on add is a great idea.

srjfoo:
> > I went ahead and added a preview capability. This allows you to preview the markdown, and edit it if necessary before adding the comment. I think this should mitigate concerns over formatting some.
> 
> I think this adds a bit too much code complexity for the functionality it enables. If, instead, we enabled comment editing a user could add a comment, and if it didn't look like they wanted, they could then edit it. Ray's idea of each comment having an anchor and jumping to that on add is a great idea.

I like the idea of jumping to an anchor, too. Right now, `tasks_comments` does not have a single field that identifies the comment (though `comment_date` is probably close enough, it can't be completely guaranteed that two comments on different tasks might not be the same) -- should a unique autoincrement field be added and that used as the anchor?

cpeel:
> I like the idea of jumping to an anchor, too. Right now, `tasks_comments` does not have a single field that identifies the comment (though `comment_date` is probably close enough, it can't be completely guaranteed that two comments on different tasks might not be the same) -- should a unique autoincrement field be added and that used as the anchor?

Anchors need only be unique per task, not globally. A hash of the submitter and time will be more than unique enough for what we need -- they don't need to be recreateable by a human as we can provide them the link if we want.

chrismiceli:
Thanks for the feedback. I'll remove the preview stuff, add edit with 24 hour limit, and add anchors with jumps for add comment and edit comment. Let me know if that isn't desired.

70ray:
> Thanks for the feedback. I'll remove the preview stuff, add edit with 24 hour limit, and add anchors with jumps for add comment and edit comment. Let me know if that isn't desired.

If we add a db field for edited then I think we should keep preview. Then preview is just so you can check the markdown does what you intended and there aren't any typos, whereas marking something edited would mean it was edited later. If we don't have an extra field then it's all the same.

cpeel:
> > Thanks for the feedback. I'll remove the preview stuff, add edit with 24 hour limit, and add anchors with jumps for add comment and edit comment. Let me know if that isn't desired.
> 
> If we add a db field for edited then I think we should keep preview. Then preview is just so you can check the markdown does what you intended and there aren't any typos, whereas marking something edited would mean it was edited later. If we don't have an extra field then it's all the same.

I'm not sure that nuance is that important. If someone edits a comment -- regardless if it's to fix formatting or to remove whole paragraphs -- it is still edited. I would defer the new DB field for an edited flag at the moment though.

srjfoo:
> > > Thanks for the feedback. I'll remove the preview stuff, add edit with 24 hour limit, and add anchors with jumps for add comment and edit comment. Let me know if that isn't desired.
> > 
> > 
> > If we add a db field for edited then I think we should keep preview. Then preview is just so you can check the markdown does what you intended and there aren't any typos, whereas marking something edited would mean it was edited later. If we don't have an extra field then it's all the same.
> 
> I'm not sure that nuance is that important. If someone edits a comment -- regardless if it's to fix formatting or to remove whole paragraphs -- it is still edited. I would defer the new DB field for an edited flag at the moment though.

It can be useful to know that something has been edited at times, but unless we also have a way of indicating why it was edited, probably largely irrelevant if most of the edits are to fix formatting problems. :)

chrismiceli:
I added the comment_id for jumping to comment. I also added edit ability.

Comment with privilege to edit:
image

edit screen:
image


sandbox updated

tasks.php
```diff
@@ -1457,9 +1481,14 @@ function ShowError($message, $goback = false)
     echo "$message</p>\n";
 }
 
-function TaskComments($tid)
+function create_anchor_for_comment($u_id, $comment_date)
 {
-    global $tasks_url;
+  return "$u_id" . '_' . "$comment_date";
```
cpeel, line 1486: Indentation is off here as well.

tasks.php
```diff
@@ -887,6 +891,24 @@ function handle_action_on_a_specified_task()
         // Redirect back to show task page to clear POST data
         metarefresh(0, "$tasks_url?action=show&task_id=$task_id");
     }
+    else if ($action == 'save_edit_comment') {
+      $comment_id = array_get($_POST, 'comment_id', "");
+      $comment = trim(array_get($_POST, 'task_comment', ''));
+      [$u_id, $comment_date] = explode('_', $comment_id, 2);
+      if ($u_id === $requester_u_id) {
```
cpeel, line 898: We should enable SAs to edit comments as well. While SAs do this very rarely, right now we need to go into the database to do it. That would change this to be:
```
if ($u_id === $requester_u_id || user_is_a_sitemanager()) {
```

tasks.php
```diff
@@ -1470,24 +1499,45 @@ function TaskComments($tid)
     $result = DPDatabase::query($sql);
 
     echo "<h2>" . _("Comments") . "</h2>";
+    $Parsedown = new Parsedown();
+    $Parsedown->setSafeMode(true);
     while ($row = mysqli_fetch_assoc($result)) {
-        echo "<div class='task-comment'>";
+        $comment_id = create_anchor_for_comment($row['u_id'], $row['comment_date']);
+        $is_owner = $requester_u_id == $row['u_id'];
```
cpeel, line 1506: As part of allowing SAs to edit comments, consider changing `$is_owner` to `$can_edit_comment` and then:
```
$can_edit_comment = $requester_u_id == $row['u_id'] || user_is_a_sitemanager();
```

cpeel (commented):
This is quite nice. A few small formatting things and a request to allow SAs to edit comments.

srjfoo (commented):
Weren't we going to put a 24-hour time limit on the comment edits (for the commenter, anyway)? (Might be better to allow unlimited edits for SAs so that if there's an overriding reason to edit, we can do it without having to go into the database.)

srjfoo:
Creating a new task, editing a task, and adding a comment will all send notifications. Currently, editing a comment will not send a notification -- do we want it to? (I think I would prefer to know when a comment is edited, but I'm also aware that a lot of editing will likely be to fix formatting issues, and one comment might be edited many times before the author gets it right.)

cpeel:
Ah, yes, we do want to limit users to only editing comments they have made in the last 24 hours, good catch Sharon.

I do not want edited comments to send emails. They don't in the forums, for instance.

srjfoo:
> I do not want edited comments to send emails. They don't in the forums, for instance.

That's reasonable -- I have mixed feelings. I mentioned it because editing a task does cause a new notification to be sent.

cpeel:
> > I do not want edited comments to send emails. They don't in the forums, for instance.
> 
> That's reasonable -- I have mixed feelings. I mentioned it because editing a task does cause a new notification to be sent.

That's fair, but perhaps it shouldn't. Right now getting an email that the task hasn't been edited without any idea on what was changed isn't very useful to me. Changing that (one way or the other) is outside this MR though.

srjfoo:
> > > I do not want edited comments to send emails. They don't in the forums, for instance.
> > 
> > 
> > That's reasonable -- I have mixed feelings. I mentioned it because editing a task does cause a new notification to be sent.
> 
> That's fair, but perhaps it shouldn't. Right now getting an email that the task hasn't been edited without any idea on what was changed isn't very useful to me. Changing that (one way or the other) is outside this MR though.

I agree (both that it's outside the scope of this MR, and that I'd like to know what changed when I get a notification that a task was edited, rather than no notification that something changed).

srjfoo (approved):
Looks good to me.

PR #533, opened 2020-12-15
darken-wc-border

srjfoo:
Changed the border around words identified by WC to a darker grey, as some proofreaders felt the need for a higher contrast border between the two background colors. Since that was the only place the .x-thin-border mixin was used, removed it.

Testable: [darken-wc-border/](https://www.pgdp.org/~srjfoo/c.branch/darken-wc-border/)

### DIFF
--- a/styles/layout.less
+++ b/styles/layout.less
@@ -52,6 +52,7 @@
  */
 
 @border-color-base: @page-text;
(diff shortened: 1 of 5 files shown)


srjfoo:
Before: 
image

and after: 
image

chrismiceli (approved):

PR #534, opened 2020-12-15
Load user by email

cpeel:
While email is not guaranteed to be unique, there are times being able to load a user by email is very useful for some admin scripts. This adds the necessary function and index. It also cleans up some indexes in the dbschema that were missed when the API landed earlier.


srjfoo (approved):

chrismiceli:
Are the admin scripts handling the non unique exception?

chrismiceli (approved):

cpeel:
> Are the admin scripts handling the non unique exception?

They are, yes.

PR #536, opened 2020-12-16
Better account activation experience

cpeel:
The account activation experience is one of the first entrypoints for new users and yet has some of the poorest user experience because it gets so little visibility for current users / developers.

This MR seeks to improve that by:
* Fixing the post-activation login form. Adding the CSRF token in e0c3ecd4a back in late August broke this so every time a new user is asked to sign in with it they get an error message.
* Improving the layout of the post-activation login form so it's more than a single input and looks like a login form.
* Improving the text layout of the activation page, adding a heading and some paragraph tags.
* Adding a header to the post-addproofer form submission page.

The only user workflow change is always redirecting logged-in users to the Activity Hub when they access the activation page. This should be an edge case anyway as, ideally, existing users should not be creating additional accounts.

These can be tested in the [better-activation-experience](https://www.pgdp.org/~cpeel/c.branch/better-activation-experience/) sandbox.

### DIFF
--- a/accounts/addproofer.php
+++ b/accounts/addproofer.php
@@ -80,15 +80,18 @@ if(count($_POST))
             setcookie("http_referer", '', time() - 1);
 
             // Page shown when account is successfully created
-            $header = sprintf(_("User %s Registered Successfully"), $username);
-            output_header($header);
+            $title = _("Registration complete");
+            output_header($title);
+            echo "<h1>$title</h1>";
 
             // Send them an activation e-mail
             maybe_activate_mail($email, $real_name, $register->id, $username, $intlang);
 
+            echo "<p>";
             echo sprintf(
                _("User %s registered successfully. Please check the e-mail being sent to you for further information about activating your account. This extra step is taken so that no-one can register you to the site without your knowledge."),
-               $username);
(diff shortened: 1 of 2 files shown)


srjfoo (approved):

accounts/addproofer.php
```diff
@@ -80,15 +80,18 @@
...
+            output_header($title);
+            echo "<h1>$title</h1>";
 
             // Send them an activation e-mail
             maybe_activate_mail($email, $real_name, $register->id, $username, $intlang);
 
+            echo "<p>";
             echo sprintf(
                _("User %s registered successfully. Please check the e-mail being sent to you for further information about activating your account. This extra step is taken so that no-one can register you to the site without your knowledge."),
                $username);
```
chrismiceli, line 93: should we html_safe this since it includes untrusted user input (username)? Though I'm pretty sure our regexp on username's prevents any html characters.

chrismiceli (approved):

accounts/addproofer.php
cpeel: The regex prevents invalid characters, but this is easy and we should do it anyway.

PR #537, opened 2020-12-17
New user experience improvements

cpeel:
This is a small set of changes to improve the new user experience:
* Change the proofreading interface default for new users to *not* open in a new window. This often interacts poorly with popup blockers and we should really phase this out altogether.
* Make the reset password link on the failed login page more prominent and remove the very outdated line about page caching.
* Copy some of the wording in the new user emails into the callouts for new users. This includes adding a new callout on the project page right at the very top for proofreaders with < 100 pages.
* Small update on the quiz pages to link to the rounds that the quizzes are pre-reqs for.

Testable in the [better-new-user-info](https://www.pgdp.org/~cpeel/c.branch/better-new-user-info) sandbox. To see the new user callouts, your user must have less than 100 pages proofread on TEST.

### DIFF
--- a/SETUP/upgrade/15/20201217_alter_user_profiles.php
+++ b/SETUP/upgrade/15/20201217_alter_user_profiles.php
@@ -1,0 +1,23 @@
+<?php
+$relPath='../../../pinc/';
+include_once($relPath.'base.inc');
+
+header('Content-type: text/plain');
+
+// ------------------------------------------------------------
+
+echo "Changing default on user_profiles.i_newwin..\n";
+$sql = "
+    ALTER TABLE user_profiles
+         ALTER i_newwin SET DEFAULT 0
+";
+
+echo "$sql\n";
+
+mysqli_query(DPDatabase::get_connection(), $sql) or die( mysqli_error(DPDatabase::get_connection()) );
+
+// ------------------------------------------------------------
+
+echo "\nDone!\n";
+
+// vim: sw=4 ts=4 expandtab
--- a/pinc/gradual.inc
+++ b/pinc/gradual.inc
@@ -145,7 +145,7 @@ function maybe_output_new_proofer_project_message()
         echo "</div>";
 
         echo "<p>";
-        echo _("You are a click away from proofreading a page. Take a moment to scroll down and read the Project Comments which contain instructions specific to this project. Once you're ready to start proofreading, click the \"Start Proofreading\" link.");
+        echo sprintf(_("You are a click away from proofreading a page. Take a moment to scroll down and read the Project Comments which contain instructions specific to this project. Once you're ready to start proofreading, click the '%s' link."), _("Start Proofreading"));
         echo "<p>";
 
         echo "<p>";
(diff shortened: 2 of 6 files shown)


SETUP/upgrade/15/20201217_alter_user_profiles.php
```diff
@@ -0,0 +1,23 @@
...
+$relPath='../../../pinc/';
+include_once($relPath.'base.inc');
+
+header('Content-type: text/plain');
+
+// ------------------------------------------------------------
+
+echo "Changing default on user_profiles.i_newwin..\n";
+$sql = "
+    ALTER TABLE user_profiles
```
chrismiceli, line 11: just apply this for everyone please :D.

pinc/gradual.inc
```diff
@@ -126,4 +131,36 @@ function maybe_output_new_proofer_message()
...
+    $pagesproofed = get_pages_proofed_maybe_simulated();
+    if($pagesproofed < 100)
+    {
+        echo "<div class='callout'>";
+        echo "<div class='calloutheader'>";
+        echo _("Welcome New Proofreader!");
+        echo "</div>";
+
+        echo "<p>";
+        echo _("You are a click away from proofreading a page. Take a moment to scroll down and read the Project Comments which contain instructions specific to this project. Once you're ready to start proofreading, click the \"Start Proofreading\" link.");
```
chrismiceli, line 148: I'm always wary of referencing other strings in a string in case one gets updated and this one is left behind, leaving a confusing user experience. I'm particularly referring to "Start Proofreading".

chrismiceli (approved):

lhamilton1:
Perhaps the link in the Quizzes should be "Passing this quiz is required prior to working in XX"

cpeel:
> Perhaps the link in the Quizzes should be "Passing this quiz is required prior to working in XX"

XX could be a comma-separated list of activities that require this quiz, i.e. `F2, PP`, but we could make it:
```
Passing this quiz is required prior to working in: XX
```

lhamilton1:
I like that, Casey! I'm thinking perhaps Also that the text on the last part of the quiz should also provide a link to P1 but that is something we would do from a quiz edit

cpeel:
> I like that, Casey! I'm thinking perhaps Also that the text on the last part of the quiz should also provide a link to P1 but that is something we would do from a quiz edit

I have links back to the rounds in a separate branch that handles P1 requiring quiz to access.

pinc/gradual.inc
cpeel: This is a fair fear. In this case I'm less worried about the string being changed in one place and not the other and more worried that translators will translate the string in two different ways -- which is a very real concern. I will pull this out into a separate string but keep it duplicated as we have no way of making this the same string without a random global variable which seems like the wrong approach.

srjfoo (approved):

pinc/gradual.inc
srjfoo: I agree about it being a legitimate concern, but didn't like the idea of forcing it to be a global variable. Pulling it out as a separate string sounds like a good idea. That way, it'll match the string that's already translated (assuming the capitalization matches), and won't need a second translation.

cpeel: See the last commit for this change.

SETUP/upgrade/15/20201217_alter_user_profiles.php
cpeel: Tempting, but I suspect some long-time users would object to someone changing their settings on them like this :D

srjfoo (approved):

PR #538, opened 2020-12-18
Require users to pass the basic proofreading quiz before working in P1

cpeel:
By GM request, this code update requires users to pass the basic proofreading quiz before working in P1. The change itself is a two-line change in `pinc/stages.inc` but ensuring that new users can figure out that they need to do this and how to do it is what everything else is for.

Broadly, this MR attempts to guide new users to the basic proofreading quiz using a new callout and then, once they successfully pass the quiz, a pointer to P1 from the quizzes. The code has been changed such that if we decide to remove the quiz requirement the site will go back to the way it is now with only removing the requirement from `pinc/stages.inc`.

Like many changes, this is a balancing act between making things work for pgdp.net without limiting how other sites use the code. It also needs to balance how activity pre-reqs are handled for P1 vs for other rounds since they use the same underlying enforcement mechanism.

Some notable changes:
* P1 (i.e. the Entry Level Round (ELR)) is always shown on the navbar regardless if the user has access to it or not.
* The Activity Hub and Round pages will show the updated callout pointing users to the basic proofreading quiz.
* If a quiz is required for a given round/stage and the user has successfully completed it, a callout will be shown on the quiz page to point the user to the round page (this is how we get users back to the P1 page from the quizzes).
* Before, to get access to rounds that were `REQ-AUTO`, the user needed to click the link in the Entrance Requirements section on the round page to gain access to the round. Now, if the user has satisfied all requirements, `REQ-AUTO` rounds will automatically grant user's access the first time they hit the round page. This removes the "find the link" problem from new proofreaders for P1. Importantly, `P2` and `F1` are also `REQ-AUTO` meaning that as soon as the user passes all of the requirements they will also be auto-granted into that round simply by accessing the round page.
* Finally, this rips out large parts of the New User Email that are redundant, changing it from a novel to a novella. Specifically:
  * Don't tell users how to log in -- users are not idiots and can figure that out.
  * The failed login page already provides troubleshooting steps, including how to reset the password and how to contact help. Don't include this in the email too.
  * Rely on the callouts to move users through how to start proofreading. We should solve UX issues with UX fixes, not with a 200+ line new user email.

This is testable in the [require-p1-quiz](https://www.pgdp.org/~cpeel/c.branch/require-p1-quiz/) sandbox. For best results, ensure the user you are testing with has not proofread more than 100 pages on the TEST server. I strongly encourage testing with a brand new user you create so you can see the entire new user workflow.

To help testing, these two scripts can be used to reset your P1 access for your user or delete your quiz results (they don't show output, just do the necessary DB updates for your username):
* https://www.pgdp.org/~cpeel/c.branch/require-p1-quiz/tools/reset-quizzes.php
* https://www.pgdp.org/~cpeel/c.branch/require-p1-quiz/tools/reset-p1-access.php

### DIFF
--- a/pinc/Activity.inc
+++ b/pinc/Activity.inc
@@ -10,9 +10,9 @@ include_once($relPath.'User.inc');
 
 $ACCESS_CRITERIA = array(
     'days since reg' => _('Days since registration'),
-    'quiz/p_mod1'    => sprintf(_('<a href="%1$s">moderate proofreading quiz %2$d</a> pass'),"$code_url/quiz/start.php?show_level=P_MOD", 1),
-    'quiz/p_mod2'    => sprintf(_('<a href="%1$s">moderate proofreading quiz %2$d</a> pass'),"$code_url/quiz/start.php?show_level=P_MOD", 2),
-    'quiz/f_only'    => sprintf(_("<a href='%s'>formatting quiz</a> pass"),"$code_url/quiz/start.php?show_only=format"),
+    'quiz/p_mod1'    => sprintf(_('Successfully complete the <a href="%1$s">moderate proofreading quiz %2$d</a>'),"$code_url/quiz/start.php?show_level=P_MOD", 1),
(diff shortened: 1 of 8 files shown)


chrismiceli (approved):
Code changes look good.

srjfoo (approved):

PR #540, opened 2020-12-24
theme-charcoal

srjfoo:
This just themes the basic site pages -- Front Page, Activity Hub, Round and Pool pages, Project Pages, and the like.

Items deferred:
1. Proofreading interface and related -- that's a separate process, and work has begun, but it's not ready for review.
1. Special Day colors. This MR shows the special day colors as they currently are on PROD. There are a couple of alternatives under discussion, but no clear decision on which to use.
1. Logo and graphs -- needs more investigation. After we drop support for IE-11, we should be able to use SVGs directly, but we'll still need to figure out how.
1. Where Project Managers force background colors in PCs without specifying text color, there may be some readability issues.

Some theming work was necessary in page_interfaces so that the colors for text areas buttons and menus would match the proofing interface instead of the theme (specifically in the Standard PI and WC). There are a couple of places (WordCheck context and quiz pages) where a page will be hybrid, because part of the page is charcoal themed, but part comes from the page interfaces part of the code, which is not yet themed.

Available for testing in the [theme-charcoal](https://www.pgdp.org/~srjfoo/c.branch/theme-charcoal/) sandbox. You'll need to go to your user prefs and change the theme to `Charcoal`.

### DIFF
--- a/styles/themes/charcoal.less
+++ b/styles/themes/charcoal.less
@@ -1,0 +1,218 @@
+/* ------------------------------------------------------------------------ */
+/* Theme coloring for Charcoal */
+
+@import "../layout.less";
+
+@theme-name: "charcoal";
+
+@page-background: #262626;
+@page-text: #e2e2e2;
+@header-background: #1a1a1a;
+@header-text: #d5d5d5;
+@navbar-background: black;
+@navbar-text: #cdcdcd;
+@sidebar-background: #1a1a1a;
+@sidebar-text: #d5d5d5;
+@footer-background: black;
+@footer-text: #cdcdcd;
+@heading-color: #bcbcbc;
+@link-color: #b3ccff;
+@visited-color: #dfb9b9;
+@active-color: #ff66aa;
+@sans-serif-font: Verdana, Helvetica, sans-serif;
+
+/* Do not underline links */
+a {
+    text-decoration: none;
+}
+
+/* 
+ * Image treatment. This is primarily aimed at proofing
+ * images. Leaving information here as a reference for
+ * working on theming page_interfaces.less. Probably use
+ * invert() for dark theme. and either none or opacity()
+ * if we want to tone down the bright white for the enhanced
+ * interface. Currently disabled.
+ */
+// @image-property: invert(85%);
+
+/* Form and menu colors */
+@form-text-background: #343434;
+@form-text-color: @page-text;
+
+/* input and select styling */
+input,
+button,
+select {
+    background-color: @form-text-background;
+    color: @form-text-color;
+    .solid-border;
+}
+
+/* ------------------------------------------------------------------------ */
+/* Common color definitions
+ * Where a particular color family has multiple shades available, a
+ * suffix indicates relative contrast as compared with the default
+ * @page-background color.
+ */
+
+/* Shades based on @page-background. Various contrast levels, commonly
+ * used for table rows or other containers that should stand out from
+ * the page background
+ */
+@table-cell-base-highc: #121212;
+@table-cell-base-mediumc: #3a3a3a;
+@table-cell-base-lowc: #4d4d4d;
+
+/* Background colors with various meanings, some needing two different
+ * contrast levels. The colors are roughly related to the default colors
+ * as defined in layout.less
+ */
+@table-cell-ok-highc: #004d00;
+@table-cell-ok-lowc: #009900;
+@table-cell-alert: #b34700;
+@table-cell-not-ok-highc: #66001a;
+@table-cell-not-ok-lowc: #b3002d;
+@table-cell-pending: #b35900;
+
+/* Special text colors that have a particular meaning that are not
+ * related to the theme colors but nonetheless may need to be
+ * adjusted depending on the theme. Also, needed variation on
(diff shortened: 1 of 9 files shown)


styles/themes/charcoal.less
```diff
@@ -0,0 +1,226 @@
...
+    color: @form-text-color;
+    .solid-border;
+}
+
+/* ------------------------------------------------------------------------ */
+/* Common color definitions
+ * Where a particular color family has multiple shades available, a
+ * suffix indicates relative contrast as compared with the default
+ * @page-background color.
+ * All of the color names defined in this file are the default colors
```
cpeel, line 50: I think we can remove this paragraph.

styles/themes/charcoal.less
```diff
@@ -0,0 +1,226 @@
...
+@table-cell-not-ok-highc: #66001a; // high contrast, red family
+@table-cell-not-ok-lowc: #b3002d;  // low contrast, red family
+@table-cell-pending: #b35900; //#e67300;      // orange family -- contrast iffy
+
+/* Special text colors that have a particular meaning that are not
+ * related to the theme colors but nonetheless may need to be
+ * adjusted depending on the theme. Also, needed variation on
+ * theme default page-text color
+ */
+@text-error: #ff8080;        // red family    -- too bright
```
cpeel, line 79: Is this still too bright and the one two down too dark, or were these left over from your development?

styles/themes/charcoal.less
```diff
@@ -0,0 +1,226 @@
...
+
+/* Background colors with various meanings, some needing two different
+ * contrast levels
+ */
+@table-cell-ok-highc: #004d00;     // high contrast, green family
+@table-cell-ok-lowc: #009900;      // low contrast, green family
+@table-cell-alert: #b34700;        // yellow/orange family
+@table-cell-not-ok-highc: #66001a; // high contrast, red family
+@table-cell-not-ok-lowc: #b3002d;  // low contrast, red family
+@table-cell-pending: #b35900; //#e67300;      // orange family -- contrast iffy
```
cpeel, line 72: If we don't want the #e67300 we should remove it. Also, is the contrast still iffy?

styles/themes/charcoal.less
```diff
@@ -0,0 +1,226 @@
...
+
+@news-border: #787878; // shade of grey
+@news-normal: #ababab; // shade of grey
+@news-announce-high: #df206c; // default maroon, shade of red
+// @news-announce-medium: // shade of orange
+@news-announce-low: #8080ff; // shade of purple
+@news-celebrate: #17cf17; // shade of green
+@news-maintain: #f53d3d; // shade of red
+
+/* Overrides: for now, a place to stash things that will
```
cpeel, line 173: Is this comment still true?

styles/themes/charcoal.less
```diff
@@ -0,0 +1,226 @@
...
+    .default-border;
+}
+
+a {
+    text-decoration: none;
+}
+
+#navbar {
+    .current-tab {
+        // border-bottom: none;
```
cpeel, line 188: If we don't want this, we should remove it.

cpeel (commented):
The theme itself looks great! Can you please review the comments in `charcoal.less` and make sure they make sense? We should also remove any commented-out styles that were part of the development process and are not needed.

Finally, your commit description is great, but please update the commit summary to be a bit more declarative, eg `Add new charcoal theme`

styles/themes/charcoal.less
srjfoo: Done.

styles/themes/charcoal.less
srjfoo: I'm not sure the colors are perfect, but the colors have been tweaked so that they're no longer too bright/dark.

styles/themes/charcoal.less
srjfoo: Removed -- and no, no longer iffy, I think. Just hadn't updated the comment. Thanks.

styles/themes/charcoal.less
srjfoo: I must have needed that at some point, but as far as I could tell, none of that definition was having any effect. I also moved the removal of the underline for links up just under the main color definitions.

styles/themes/charcoal.less
srjfoo: Not really. Removed. Actually, the whole `charcoal.less` file is an override, isn't it?

srjfoo:
> The theme itself looks great! Can you please review the comments in `charcoal.less` and make sure they make sense? We should also remove any commented-out styles that were part of the development process and are not needed.
> 
> Finally, your commit description is great, but please update the commit summary to be a bit more declarative, eg `Add new charcoal theme`

Commit summary updated. I've reviewed the comments, removed some, clarified others. I'm sure there are still rough spots remaining. If you see anything else that needs updating, let me know :)

cpeel:
I've added a commit that inserts the theme into the DB and a corresponding `db_schema.sql` change.

This LGTM!

srjfoo:
> I've added a commit that inserts the theme into the DB and a corresponding `db_schema.sql` change.
> 
> This LGTM!

Great, thanks!

PR #541, opened 2020-12-24
Check headers before resetting profile cookie

cpeel:
Cookies can only be set before headers have been sent to the browser, otherwise PHP generates a warning. Check that headers haven't been sent before resetting invalid profile cookies. Failing to reset it is fine since only valid profile cookies are ever used.

Seen in the `php_errors` log on PROD.

While I don't know for certain why we're seeing this, it could happen by a user deleting a profile in one browser that a second web browser had been using.

### DIFF
--- a/pinc/User.inc
+++ b/pinc/User.inc
@@ -112,7 +112,15 @@ class User
                     catch(NonexistentUserProfileException $exception)
                     {
                         // reset the cookie since the value was invalid
-                        setcookie("profile");
+
+                        // We can only reset the cookie if the headers haven't
+                        // yet been sent. Resetting the cookie is a nicety
+                        // since it will only be used if it is valid, so it's
+                        // OK if we can't.
+                        if(!header_sent())
+                        {
+                            setcookie("profile");
+                        }
 
                         // fall through to getting the profile from the user record
                     }


pinc/User.inc
```diff
@@ -112,7 +112,15 @@ class User
...
                         // reset the cookie since the value was invalid
-                        setcookie("profile");
+
+                        // We can only reset the cookie if the headers haven't
+                        // yet been sent. Resetting the cookie is a nicety
+                        // since it will only be used if it is valid, so it's
+                        // OK if we can't.
+                        if(!header_sent())
+                        {
+                            setcookie("profile");
```
chrismiceli, line 122: I'd be really curious why the cookies are already sent. Wouldn't that mean some meta refresh or something? Do we need to return after those perhaps?

chrismiceli (approved):

pinc/User.inc
cpeel: It isn't specifically that cookies have already been sent, just that headers have already been sent (which may or may not involve cookies). This codepath happens when a user profile is loaded, which could happen potentially anywhere in the code.

The codepath to set the cookie to a valid value (up on line 79) _must_ happen before headers or sent or it will fail. Currently we only do that in `userprefs.php` when the user selects a profile to use -- explicitly before we send other output.

srjfoo (approved):

70ray (approved):

PR #544, opened 2020-12-30
Remove project comment templates

cpeel:
This removes project comment templates. It includes an upgrade script to replace the project comment template reference with the contents of the template itself. Because the template files must exist for this upgrade script to work we can't remove the contents of `pinc/templates/` until _after_ we cut a code release in 2020-02.

The UI aspects of this can be viewed in [remove-templates](https://www.pgdp.org/~cpeel/c.branch/remove-templates/). I have validated the upgrade script works by testing it against a single project and confirming it works.


project.php
```diff
@@ -761,16 +761,6 @@ function do_project_info_table()
     {
         $comments = $project->comments;
 
-        // Automatically prepend second round comments template for Beginners Only
-        // The r2 in the template name is an artifact of the old two-round structure
-        if ($project->difficulty == "beginner")
-        {
```
chrismiceli, line 767: How will this happen now?

chrismiceli (approved):

SETUP/upgrade/15/20201229_replace_project_templates.php
```diff
@@ -0,0 +1,48 @@
...
+while(list($projectid, $comments) = mysqli_fetch_row($result))
+{
+    // handle all templates in the comment
+    while(preg_match("/\[template=([^\]]+)\]/", $comments, $matches))
+    {
+        $template = $matches[1];
+        $template_path = "$relPath/templates/comment_files/$template";
+        if(!is_file($template_path))
+        {
+            die("Template '$template' referenced in $projectid but does not exist on the filesystem.\n");
```
chrismiceli, line 28: Maybe skip instead?

project.php
cpeel: We will have this code as a PROD hack (or something much like it):
```php

        // Automatically prepend second round comments template for Beginners Only
        // The r2 in the template name is an artifact of the old two-round structure
        if ($project->difficulty == "beginner")
        {
            if ($round && $round->is_a_mentor_round() )
            {
                $comments = "
                <p>This book has changed from a BEGINNERS ONLY project to a MENTORS project.</p>

                <p><b>NOTE:</b> You may find a list of which beginner has done which page, and  the number of pages proofread by each, at the <a href='https://www.pgdp.net/c/tools/proofers/for_mentors.php'>For Mentors page</a>.</p>

                <p>Also, should you be issued a page that shows any of the Project Facilitators as the P1 proofreader, you may safely skip the feedback--unless, of course, we NEED it! That page will have been reclaimed, then proofread by one of us in order to move the project on to P2.</p>
                " . $comments;
            }
        }
```

srjfoo: I was actually looking at that bit of code today and thought to myself -- "Aaah, that's where it would go!". One other comment to make about what is currently BGr2 -- because all BEGINs are held in P3 Waiting until after the newest part has been out of P2 for at least 100 days, and merged before being released into P3, you can probably not worry about removing that bit of comments after P2. Only the destination part for the merge will need to have PCs updated, and the squirrel doing the merge can as easily remove/comment out that as the rest of the BEGIN PCs, while putting in the new PCs for P3+.

srjfoo (approved):

project.php
cpeel: > you can probably not worry about removing that bit of comments after P2

The PROD hack won't actually alter the PCs, it is only added when rendered, so there is nothing for the PM to remove.

SETUP/upgrade/15/20201229_replace_project_templates.php
cpeel: Yes, that would be a better idea. The script could be run again after the problem is fixed and if the admin decides not to fix a problem it wouldn't break the system. Done.

PR #545, opened 2021-01-02
Add a row cell for special days indicator

cpeel:
Rather than coloring the whole table row for special days, just do a single column that contains a special day indicator. This improves accessibility and overall UX.

Show special day information in round tables and show the legend to all users.

This is testable in the [special-color-improvements](https://www.pgdp.org/~cpeel/c.branch/special-color-improvements/) sandbox. See also the [discussion in the forums](https://www.pgdp.net/phpBB3/viewtopic.php?f=4&t=72729).

### DIFF
--- a/SETUP/db_schema.sql
+++ b/SETUP/db_schema.sql
@@ -522,6 +522,7 @@ CREATE TABLE `special_days` (
   `date_changes` varchar(100) default NULL,
   `info_url` varchar(255) default NULL,
   `image_url` varchar(255) default NULL,
+  `symbol` varchar(2) default '',
   UNIQUE KEY `spec_code` (`spec_code`)
 ) COMMENT='definitions of SPECIAL days';
 # --------------------------------------------------------
--- a/pinc/special_colors.inc
+++ b/pinc/special_colors.inc
@@ -6,16 +6,19 @@ function load_common_special_days()
 {
     return [
         "_BIRTHDAY_RECENT" => [
-            'color' => "CCFFFF",
+            'color' => "#CCFFFF",
             'display_name' => "Authors with recent birthdays",
+            'symbol' => '🍰',
         ],
         "_BIRTHDAY_TODAY" => [
-            'color' => "33CCFF",
+            'color' => "#33CCFF",
             'display_name' => "Authors with birthdays today",
+            'symbol' => '🎂',
         ],
         "_OTHER_SPECIAL" => [
-            'color' => "FFFF66",
+            'color' => "#FFFF66",
             'display_name' => "Other Special",
+            'symbol' => '🎉',
         ],
     ];
 }
@@ -25,7 +28,7 @@ function load_special_days()
     $specials_array = load_common_special_days();
 
     $sql = "
-        SELECT spec_code, color, display_name
+        SELECT spec_code, color, display_name, symbol
         FROM special_days
         ORDER BY display_name
     ";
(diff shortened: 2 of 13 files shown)


SETUP/db_schema.sql
```diff
@@ -522,6 +522,7 @@ CREATE TABLE `special_days` (
   `date_changes` varchar(100) default NULL,
   `info_url` varchar(255) default NULL,
   `image_url` varchar(255) default NULL,
+  `symbol` varchar(2) default '',
```
chrismiceli, line 525: I'm not going to pretend to know how mysql handles unicode but if it uses utf-16 like javascript with surrogate pairs and compex emoji you may need more space.

pinc/special_colors.inc
```diff
@@ -126,27 +123,70 @@ function echo_special_legend( $projects_where_clause)
     $day_array = array_merge($day_array, array_values(load_common_special_days()));
 
     $specs_count = count($day_array);
-    echo "<h3><a href='$code_url/tools/project_manager/show_specials.php'>$legend_text</a></h3>";
-    echo "<table class='basic'>\n";
-    $day_index = 0;
-    $table_columns = min($specs_count, 4);
-    while($day_index < $specs_count)
+    echo "<h3>$legend_text</h3>";
```
chrismiceli, line 126: Html safe?

pinc/special_colors.inc
```diff
@@ -126,27 +123,70 @@ function echo_special_legend( $projects_where_clause)
...
+    for( ; $column_index > 0; $column_index--)
+    {
+        echo "<td class='top-align'>";
+        echo "<table class='basic'>\n";
+        $column = array_shift($columns);
+        foreach($column as $day_entry)
+        {
+            list($style, $contents) = get_special_day_cell_parts($day_entry, FALSE);
+            echo "<tr>";
+            echo "<td style='$style'>$contents</td>";
```
chrismiceli, line 152: If special days are user controlled html safe? Are we sure the style will be ok too or should we atr safe too?

pinc/special_colors.inc
```diff
@@ -126,27 +123,70 @@ function echo_special_legend( $projects_where_clause)
...
+        "text-align: center; width: 1em;";
+
+    // If the symbol is empty, return an em-space so that the browser
+    // will show a tooltip for the cell.
+    $contents = $special_day['symbol'] == '' ? '&emsp;' : $special_day['symbol'];
+    if($include_title)
+    {
+        $title = attr_safe($special_day['display_name']);
+        $contents = "<span title='$title'>$contents</span>";
+    }
```
chrismiceli, line 178: Same here. It'd be hard to do some xss with 2 chars but still.

chrismiceli (approved):

srjfoo (approved):

SETUP/db_schema.sql
cpeel: In MySQL 5.5 and later, a single character is a single Unicode character regardless of how wide it is. So `varchar(2)` will allow any two Unicode characters.

pinc/special_colors.inc
cpeel: Special Days are SA controlled. The only user input factored into the style is the color and it's reasonable to expect SAs to put in valid colors I think.

pinc/special_colors.inc
cpeel: I'm not worried about SAs introducing XSS here -- there are numerous other, easier ways they could do this if they really wanted to.

pinc/special_colors.inc
```diff
@@ -126,27 +123,70 @@ function echo_special_legend( $projects_where_clause)
...
+// Select a reasonable text color given a specific background color.
+// From https://stackoverflow.com/questions/3942878/how-to-decide-font-color-in-white-or-black-depending-on-background-color
+function pick_text_color_given_bgcolor($bg_color, $light_color, $dark_color)
+{
+    $color = str_replace("#", "", $bg_color);
+
+    $r = hexdec(substr($color, 0, 2)); // hexToR
+    $g = hexdec(substr($color, 2, 4)); // hexToG
+    $b = hexdec(substr($color, 4, 6)); // hexToB
+    return ((($r * 0.299) + ($g * 0.587) + ($b * 0.114)) > 186) ? $dark_color : $light_color;
```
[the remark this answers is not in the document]
cpeel: This is largely a fallback for systems that do not have an emoji and that fall back to some text form (e.g. Windows 10 does not have a China flag emoji and shows `CN` instead). I'm hoping we will be able to use an emoji (or nothing) for most Special Days. This algorithm is not perfect but it's better than nothing and I don't want to have to set a text color for each special day too. We might have to revisit this later but I'm hoping this is good enough for v1.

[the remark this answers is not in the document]
srjfoo: I think so, yes.

pinc/special_colors.inc
cpeel: Good call and done. 👍

chrismiceli (approved):

70ray (approved):

PR #548, opened 2021-01-06
Birthday and otherday Special days

70ray:
This implements task 1961. https://www.pgdp.net/c/tasks.php?task_id=1961.
This is the easy option leaving the api function uncahnged.

Running in sandbox https://www.pgdp.org/~rp31/c.branch/spec_days_1

### DIFF
--- a/pinc/ProjectSearchForm.inc
+++ b/pinc/ProjectSearchForm.inc
@@ -202,14 +202,17 @@ class ProjectSearchForm
         $this->define_form_widgets();
     }
 
-    static public function special_day_options()
+    function all_special_day_options()
     {
-        $special_day_options = array(
+        return array(
             '' => _('Any day'),
             'Birthday' => _('Birthday'),
             'Otherday' => _('Otherday'),
-        );
+        ) + $this->special_day_options();
+    }
 
+    static public function special_day_options()
+    {
         $sql = "
             SELECT
                 spec_code,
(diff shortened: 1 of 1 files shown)


pinc/ProjectSearchForm.inc
```diff
@@ -142,12 +131,51 @@ class ProjectSearchWidget
...
             {
                 $value = normalize_whitespace($value);
-
                 $contribution = "$column_name LIKE '%$value%'";
             }
         }
         return $contribution;
     }
+
+    function sql_from_multi($values, $column_name, $comparator)
```
chrismiceli, line 140: I would be explicit in the variable name at least that you expect $values to be encoded already. Something like ```$encodedValues``` so people know if you need to encode them or not before calling this method.

cpeel: I would rather move the encoding to this function.

pinc/ProjectSearchForm.inc
```diff
@@ -174,10 +202,17 @@ class ProjectSearchForm
...
+        return array(
+            '' => _('Any day'),
+            'Birthday' => _('Birthday'),
+            'Otherday' => _('Otherday'),
+        ) + $this->special_day_options();
+    }
+
     static public function special_day_options()
     {
-        $special_day_options = array();
```
cpeel, line 179: You need to keep this so that `$special_day_options` is pre-defined as an array before you append to it in the while loop further down.

pinc/ProjectSearchForm.inc
70ray: Done this.

pinc/ProjectSearchForm.inc
70ray: Yes. Changed this now.

cpeel:
This LGTM although I would appreciate @srjfoo's keen validation eyes on this if she has time.

srjfoo:
> This LGTM although I would appreciate @srjfoo's keen validation eyes on this if she has time.

It appears to work as expected.

srjfoo (approved):

PR #549, opened 2021-01-08
Add volunteer cert info to new user email

srjfoo:
At the request of the GM: Add a paragraph that provides a pointer for new users to information on certification of volunteer work. A larger percentage than before of new volunteers seem to be students needing to do volunteer work.

If you need to, you can test it in the [update-welcome-email](https://www.pgdp.org/~srjfoo/c.branch/update-welcome-email) sandbox.

### DIFF
--- a/pinc/new_user_mails.inc
+++ b/pinc/new_user_mails.inc
@@ -161,6 +161,15 @@ and browse the resources in our Information for New and
 Returning Volunteers area at
 $dp_wiki_url/DP_Official_Documentation:General#Information_For_New_and_Returning_Volunteers
 
+
+Certification of Volunteer Work
+-------------------------------
+
+$site_name tracks only the number of pages proofread. We
+does not track the amount of time our volunteers spend working on those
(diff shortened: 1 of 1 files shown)


chrismiceli (approved):

pinc/new_user_mails.inc
```diff
@@ -161,6 +161,15 @@ and browse the resources in our Information for New and
 Returning Volunteers area at
 $dp_wiki_url/DP_Official_Documentation:General#Information_For_New_and_Returning_Volunteers
 
+
+Certification of Volunteer Work
+-------------------------------
+
+Distributed Proofreaders tracks only the number of pages proofread. It
```
cpeel, line 168: `Distributed Proofreaders` should be `$site_name`. Also should it be "We do not track" vs "It does not track"?

srjfoo: Done, and thanks -- I feel kind of silly for missing `$site_name`. And "We" sounds like a good update.

PR #550, opened 2021-01-10
Don't use absolute for split

chrismiceli:
I was experimenting replacing some of the <frameset>s with splitControl, but noticed it used absolute positioning. I figured we could do this with flex.

I really want @70ray's input here and not trying to step on your toes. Just a minor tweak to your major contribution.

I deployed here and you can check it out here:
https://www.pgdp.org/~chrismiceli/c.branch/flexSplit/tools/page_browser.php?project=projectID5e23ae7ff0a81&mode=imageText&imagefile=016.png
or in the manual test pages:
https://www.pgdp.org/~chrismiceli/c.branch/flexSplit/SETUP/tests/manual_web/split_test/index.html

If we do accept this PR, we'll need to update the text here: https://github.com/DistributedProofreaders/dproofreaders/blob/master/SETUP/tests/manual_web/split_test/index.html#L19

### DIFF
--- a/SETUP/tests/manual_web/split_test/index.html
+++ b/SETUP/tests/manual_web/split_test/index.html
@@ -13,14 +13,5 @@
   <li><a href="switchable_split.php">Switchable between horizontal and vertical split with indicator for split percentage changes</a></li>
   <li><a href="vertical_horizontal_split_flex.php">Vertical and Horizontal Split with text in a textarea and different splitter bar colours and widths</a></li>
 </ul>
-<h2>Implementation notes</h2>
-<p>
-If window resize can cause the size of the splitter area to change then the body must have style 'overflow: hidden'. This is necessary because
-the panes are absolutely positioned with sizes calculated from the size of their container.
-When the container becomes smaller as a result of resizing the window we recalculate the size of the panes.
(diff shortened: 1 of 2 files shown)


cpeel:
I suspect, but do not know, that the original version didn't use flex to avoid some of the problems in IE11 (which we no longer have to worry about).

70ray (approved):
Works well and is considerably simpler. I never thought of doing like this.

srjfoo (approved):
It looks good to me.

Bear in mind that the JS is way above my head, so I'm looking just at how it operates. This change would be strictly invisible to the users? (I.e. where the sliders exist now, they would remain; where they don't exist now, they would not be added?)

chrismiceli:
> 
> 
> I suspect, but do not know, that the original version didn't use flex to avoid some of the problems in IE11 (which we no longer have to worry about).

For what that is worth, I also tested in Internet explorer and all seemed well.

chrismiceli:
> 
> 
> It looks good to me.
> 
> Bear in mind that the JS is way above my head, so I'm looking just at how it operates. This change would be strictly invisible to the users? (I.e. where the sliders exist now, they would remain; where they don't exist now, they would not be added?)

exactly, the user will see no difference, but it will make my conversion of some framesets to split control easier and the code lets the browser align the 2nd panel, which should be more accurate and hopefully faster (not the goal though).

PR #552, opened 2021-01-11
New character suite: Symbols collection

srjfoo:
Character suite `symbols-collection`, contains astronomical, zodiac, apothecary and music symbols, grouped into two picker sets. The Zodiac symbols are combined with the VS15 variation selector to try
to force them to appear as normal type instead of as emoji, though this does not work in all cases.

Character suite can be viewed at [symbols-collection](https://www.pgdp.org/~srjfoo/c.branch/symbols-collection/tools/charsuites.php?charsuite=symbols-collection); used in a project at [Elegantní Tanecník - WordCheck Testing](https://www.pgdp.org/~srjfoo/c.branch/symbols-collection/project.php?id=projectID47127a26a1f9b&expected_state=P1.proj_avail).

### DIFF
--- a/SETUP/UNICODE.md
+++ b/SETUP/UNICODE.md
@@ -42,7 +42,7 @@ that are shown to the users in the proofreading interface character picker.
 Picker sets are optional and, if they are used, need not cover the entire set
 of codepoints in the character suite.
 
-The code currently ships with 4 character suites:
+The code currently ships with the following character suites:
 
 * Basic Latin - characters contained within ISO-8859-1 and CP-1252
 * Extended European Latin A - Esperanto, Welsh, Sami, Romanian, Hungarian
@@ -52,6 +52,7 @@ The code currently ships with 4 character suites:
 * Polytonic Greek
 * Medievalist supplement - for transcription/analysis of medieval writings
 * Semitic and Indic transcriptions - Romanized forms of Arabic, Hebrew, Sanskrit
+* Symbols collection - Astronomical, Music, Zodiac and Apothecary symbols
 
 All character suites can be viewed using the [All Character Suites](../tools/charsuites.php)
 page.
--- a/pinc/charsuite-symbols-collection.inc
+++ b/pinc/charsuite-symbols-collection.inc
@@ -1,0 +1,53 @@
+<?php
+include_once($relPath."CharSuites.inc");
+
+$charsuite = new CharSuite("symbols-collection", _("Symbols collection"));
+$charsuite->description = _("Astronomical, zodiac, and a limited number of apothecary and music symbols. The zodiac symbols are provided with the variation selector, VS15, to try to force them to display as ordinary charcters instead of as emoji. This may not work for all fonts.");
+
+$charsuite->codepoints = [
+    'U+0292',
+    'U+2108',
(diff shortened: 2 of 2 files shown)


SETUP/UNICODE.md
```diff
@@ -42,7 +42,7 @@ that are shown to the users in the proofreading interface character picker.
 Picker sets are optional and, if they are used, need not cover the entire set
 of codepoints in the character suite.
 
-The code currently ships with 4 character suites:
+The code currently ships with 9 character suites:
```
chrismiceli, line 45: This will probably get out of date like before. Maybe reword to just ```The code currently ships with the following character suites:```?

pinc/charsuite-symbols-collection.inc
```diff
@@ -0,0 +1,53 @@
+<?php
+include_once($relPath."CharSuites.inc");
+
+$charsuite = new CharSuite("symbols-collection", _("Symbols collection"));
+$charsuite->description = _("Astronomical, zodiac and a limited number of apothecary and music symbols. The zodiac symbols are provided with the variation selector, VS15, to try to force them to display as ordinary charcters instead of as emoji. This may not work for all fonts.");
```
chrismiceli, line 5: what's our opinion on serial commas? :D

chrismiceli, line 5: should we link to some wiki about variation selectors?

chrismiceli (approved):

SETUP/UNICODE.md
cpeel: This is a great idea.

pinc/charsuite-symbols-collection.inc
cpeel: Yes - Andrew, Sharon and I chatted about this and one is forthcoming.

SETUP/UNICODE.md
srjfoo: I agree. I'm waiting on a wiki page so I can update everything at the same time.

pinc/charsuite-symbols-collection.inc
srjfoo: That they're used inconsistently across the site. What would you suggest here? \<grin\>

chrismiceli: I have no suggestion. I just usually use it so it stuck out to me. I approved because this shouldn't hold anything up. Just pointing it out in case there was some standard or you had a preference.

SETUP/UNICODE.md
srjfoo: Done.

pinc/charsuite-symbols-collection.inc
srjfoo: After reviewing the wikipedia page on serial commas and re-reading the blurb, I added it here (but neglected to add it in the `.md` file, which I've chosen to leave as-is unless there are mutterings on inconsistency :D).

srjfoo: The wiki page has been started, but not completed as of this moment. However, I went ahead and added a link to it -- it will hopefully be ready by the time we need it.

PR #553, opened 2021-01-12
Hiero

chrismiceli:
I started out trying to just remove the ```<frameset>```, but it turns out it really wasn't needed in the first place. I combined the 2 files and that is pretty much it. I had to do a form submit on change instead of refreshing the other frame but that was all the major changes.

While I was there though, the formatting between the 2 files was different, so I reconciled them. I then also did html_safe and attr_safe for html template.

I then split the 1 javascript function into its own file to lint and unit test it.

You can verify the changes at https://www.pgdp.org/~chrismiceli/c.branch/hiero/tools/proofers/hiero/index.php or access it through the proofing interface.

Visually almost identical (hr instead of frame, also not resizable anymore):
image


tools/proofers/hiero/index.php
```diff
@@ -2,15 +2,201 @@
...
+<form name="hieroform" method="POST">
+<table>
+<tr><td rowspan="2">
+<textarea name="hierobox" rows="4" cols="30">
+<?php $hierobox=@$_POST['hierobox']; echo html_safe($hierobox); ?>
+</textarea>
+</td><td colspan="2">
+<select onChange="hieroform.submit()" name="table">
+<?php
+echo "<option>".html_safe(WH_Text("Tables"))."</option>";
```
cpeel, line 65: Disclaimer: I know this is simply a lift from what was there before.

Rather than have this interlaced php -> html -> php -> html -> php, consider building the select options into a string first, then just echo'ing the value. Even better would be to use a heredoc for the HTML so the entire files is php:

```php
echo <<<HTML
<form name="hieroform" method="POST">
<table>
<tr><td rowspan="2">
<textarea name="hierobox" rows="4" cols="30">
$escaped_hierobox
</textarea>
</td><td colspan="2">
<select onChange="hieroform.submit()" name="table">
$table_options
</select>
...
HTML;
```

Ideally, shove the select-generating logic (and the `$tables` and `$syntax` arrays) into two functions after the HTML blob, but I'm OK leaving that logic at the top-level of the file too.

cpeel (commented):
This is a huge improvement! My comments are at further improving the prior code that was lifted into `index.php`.

tools/proofers/hiero/index.php
```diff
@@ -2,15 +2,222 @@
...
+        if (strlen($v) <= 2)
+            $table_options .= html_safe($v) . " &mdash; ";
+        $table_options .= html_safe(WH_Text($v));
+        $table_options .= "</option>";
+    }
+
+    return $table_options;
+}
+
+function syntax_options()
```
cpeel, line 201: `syntax_options()` includes the `<select>` tags but `table_options()` doesn't. They should behave similarly. From the naming of them I would only expect them to include the `<option>`s.

srjfoo (approved):
I don't see any differences in operation between the sandbox and PROD.

tools/proofers/hiero/index.php
```diff
@@ -2,15 +2,222 @@
...
+    "X",
+    "Y",
+    "Z",
+    "Aa",
+    "All"
+);
+
+$syntax=array("-",":","*","!");
+$escaped_hierobox = html_safe($hierobox);
+$table_options = table_options();
```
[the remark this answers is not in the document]
chrismiceli: whoops, my brain just turned off it seems.

PR #554, opened 2021-01-13
Update Proofing Interface Help

srjfoo:
The initial purpose of this MR was to remove an unused graphic from prooffacehelp. It turned into a major update, because there were other things that were just flat wrong. I don't claim to have fixed everything that needs to be fixed (layout leaves something to be desired), but I think I've got the worst of the wording problems fixed.

In other commits: added an ID to the Project Comments so that they can be linked to directly instead of to the top of the project page. Deleted several unused graphics. Removed an unused button from `pinc/button_defs.inc`

To best see what's changed in `prooffacehelp.php`, compare [update-prooffacehelp--enhanced](https://www.pgdp.org/~srjfoo/c.branch/update-prooffacehelp/faq/prooffacehelp.php?i_type=1) with [PROD--enhanced](https://www.pgdp.net/c/faq/prooffacehelp.php?i_type=1) for the enhanced interface, and [update-prooffacehelp--standard](https://www.pgdp.org/~srjfoo/c.branch/update-prooffacehelp/faq/prooffacehelp.php?i_type=0) with [PROD--standard](https://www.pgdp.net/c/faq/prooffacehelp.php?i_type=0) for the standard interface.


srjfoo:
Further notes:

This file will need a major upgrade when the new proofing interface is rolled out. Until then, it's questionable how much effort we should put into fixing this one, other than to make sure that it's not providing outdated information. Among the things I did not do was fix outdated HTML other than adding closing paragraph tags where they were not closed, unless I needed to fix it for another reason, so there's still a lot of outdated code in there.

faq/prooffacehelp.php
```diff
@@ -613,19 +514,17 @@ function echo_row( $name, $tooltip, $button_image_base, $accelerator )
...
 from the dropdown menu to change the font size of the proofreading text.
 </TD></TR>
 
 <TR><TD VALIGN="TOP"><B>Font Face</B>
-<BR>Selection Menu</TD><TD>Select a named font from the dropdown menu.  If the font is
-installed on your system, the proofreading text will change to the selected font.  If the
-font is not installed on your system, your system will either automatically select a
-font in the same family or leave the current font unchanged.
+<BR>Selection Menu</TD><TD class='top-align'>Select a font from the dropdown menu. The choices are
+Browser Default, DP Sans Mono, DejaVu Sans Mono. If you choose a font installed
```
cpeel, line 523: I would lean towards not mentioning specific fonts, because then we need to keep them in sync if we add/remove some later. I don't feel super strong about it, however.

cpeel (commented):
This is a huge improvement - thank you for working on it!

pinc/button_defs.inc
```diff
@@ -181,15 +181,6 @@ function echo_button( $button_id, $which_interface )
                 'src'       => "gfx/bt16.png",
             );
             break;
-
-        case CONVERT_UTF8:
```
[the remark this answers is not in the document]
srjfoo: Thanks for the reminder -- I totally forgot about that. Done

faq/prooffacehelp.php
srjfoo: I have no strong objection to removing the list of fonts. Honestly, hopefully that part of the enhanced interface is reasonably self-explanatory. Done.

PR #564, opened 2021-01-17
Encode Astral characters in phpBB 3.3.x subjects

cpeel:
phpBB 3.3.x supports Astral characters (like emojis) in posts, but they need to be encoded correctly for topic subjects. This is most easily seen in the phpBB's [posting.php](https://github.com/phpbb/phpbb/blob/master/phpBB/posting.php#L762) file.

The `utf8_encode_ucr()` function doesn't exist in phpBB 3.2.x so this should be safe for those versions (which is also what is running on PROD right now).

Testable in the [astral-forum-subjects](https://www.pgdp.org/~cpeel/c.branch/astral-forum-subjects/) sandbox.

[Task 1886](https://www.pgdp.net/c/tasks.php?task_id=1886)

### DIFF
--- a/pinc/phpbb3.inc
+++ b/pinc/phpbb3.inc
@@ -242,6 +242,12 @@ function _insert_post(
     //       $type_cast_helper->set_var() in phpBB3/phpbb/request/type_cast_helper.php
     $subject = htmlspecialchars($subject, ENT_COMPAT, 'UTF-8', false);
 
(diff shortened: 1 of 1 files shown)


srjfoo (approved):
I already had some projects set up with astral plane characters in the project title. This fixed the problem (both for author and title).

chrismiceli (approved):

Issue #562, opened 2021-01-17
Remove frameset from PM WordCheck tools


cpeel:
Three of the PM WordCheck tools use framesets:
* `tools/project_manager/show_word_context.php`
* `tools/project_manager/show_all_good_word_suggestions.php`
* `tools/project_manager/show_good_word_suggestions_detail.php`

These need to be updated to use the new `scripts/splitControl.js` instead.

chrismiceli:
Should we close this not that it is merged, or wait until deployed? Not sure our process for github issues.

cpeel:
This ticket is for 3 different PM WordCheck interfaces and our MR was just for one of them, right?

chrismiceli:
Ah yes.

cpeel:
The merge of the `page_browser_uses_api` branch resolved this issue.

PR #566, opened 2021-01-20
562 Remove frameset from PM WordCheck tools

chrismiceli:
replace frameset with splitcontrol
This is step 1. There is at least one more nested frame in this path, then I can replace the parent frameset. Available in sandbox: https://www.pgdp.org/~chrismiceli/c.branch/issue/562 and can be seen at https://www.pgdp.org/c/tools/project_manager/show_all_good_word_suggestions.php

### DIFF
--- a/tools/project_manager/show_good_word_suggestions_detail.js
+++ b/tools/project_manager/show_good_word_suggestions_detail.js
@@ -1,0 +1,59 @@
+/*global $ pageBrowserData imageControl pageChanger projectSelector pageTitle */
+
+$(function () {
+    let STORAGE_KEY = 'show_good_word_suggestions_details';
+    var hiddenLayout = function() {
+        return $("<input>", {type: 'hidden', name: 'layout', value: show_good_word_suggestions_details.layout});
+    };
+    var hiddenCutoff = function() {
+        return $("<input>", {type: 'hidden', name: 'timeCutoff', value: show_good_word_suggestions_details.timeCutoff});
+    };
+    var hiddenWord = function() {
+        return $("<input>", {type: 'hidden', name: 'word', value: show_good_word_suggestions_details.word});
+    };
+        var hiddenProjectId = function() {
+        return $("<input>", {type: 'hidden', name: 'projectid', value: show_good_word_suggestions_details.projectid});
+    };
+    let splitPercent = parseFloat(localStorage.getItem(STORAGE_KEY));
+    if(!splitPercent) {
+        // if not defined or 0, set the value to 100 to the text is shown
+        splitPercent = 30;
+    }
+    let mainSplit = splitControl("#show_good_word_suggestions_detail_container", {
+        splitVertical: show_good_word_suggestions_details.layout === 'vertical',
+        splitPercent: splitPercent,
+    });
+    mainSplit.reLayout();
+    mainSplit.dragEnd.add(function (percent) {
+        localStorage.setItem(STORAGE_KEY, percent);
+    });
+    let topDiv = $("#page-browser");
+    // the non-scrolling area which will contain any error message and the form with controls
+    let fixHead = $("<div>", {class: 'fixed-box control-pane'});
+    topDiv.append(fixHead);
+
+    if(pageBrowserData.errorMessage) {
+        fixHead.append($("<p>", {class: 'error'}).append(pageBrowserData.errorMessage));
+    }
+
+    let pageControlForm = $("<form>", {method: "get"});
+    pageControlForm.append(hiddenProjectId());
+    pageControlForm.append(hiddenLayout());
+    pageControlForm.append(hiddenCutoff());
+    pageControlForm.append(hiddenWord());
+    pageControlForm.prop("class", "inline");
+    pageControlForm.append(pageChanger(pageControlForm));
+    fixHead.append(pageControlForm);
+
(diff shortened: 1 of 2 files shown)


chrismiceli:
It may help to compare the changes with split view ignoring whitespace changes as move of the changes are just indenting level change.

tools/project_manager/show_good_word_suggestions_detail.php
```diff
@@ -45,109 +46,100 @@
...
-<?php
-    exit;
+$header_args = [
+  "js_files" => [
+    "$code_url/scripts/splitControl.js",
+    "$code_url/scripts/page_browse.js",
+    "./show_good_word_suggestions_detail.js",
+  ],
+  "js_data" => get_page_data_js($projectid, $imagefile, 'OCR', $error_message) . get_proofreading_interface_data_js() . "
+        var show_good_word_suggestions_details = {
```
cpeel, line 56: It would be better to just `json_encode()` a PHP associative array so that we don't have to worry about any escaping here.

tools/project_manager/show_good_word_suggestions_detail.php
```diff
@@ -45,109 +46,100 @@
...
-<frame name="worddetailframe" src="show_good_word_suggestions_detail.php?projectid=<?php echo $projectid; ?>&amp;word=<?php echo $encWord; ?>&amp;timeCutoff=<?php echo $timeCutoff; ?>&amp;frame=left">
-<frame name="imageframe" src="show_good_word_suggestions_detail.php?projectid=<?php echo $projectid; ?>&amp;word=<?php echo $encWord; ?>&amp;timeCutoff=<?php echo $timeCutoff; ?>&amp;frame=right">
-</frameset>
-<noframes>
-<?php echo _("Your browser currently does not display frames!"); ?>
-</noframes>
-<?php
-    exit;
+$header_args = [
+  "js_files" => [
```
cpeel, line 50: Please use 4-space indents in this block.

tools/project_manager/show_good_word_suggestions_detail.php
```diff
@@ -45,109 +46,100 @@
...
+    sprintf(_("Suggestion context for '%1\$s' in %2\$s"),
+        $word, $project_name),
+    "</h2>";
+
+echo "<p>";
+echo "<a href='show_word_context.php?projectid=$projectid&amp;word=$encWord'>" .
+      _("Show full context set for this word") . "</a>";
+
+echo " | ";
+echo "<a href='" . attr_safe($_SERVER['PHP_SELF']) . "?projectid=$projectid&amp;word=$encWord&amp;timeCutoff=$timeCutoff&amp;";
```
cpeel, line 95: I'm pretty sure you can drop the `attr_safe()` bit altogether and just use:
```html
<a href='?projectid=$projectid...>
```

tools/project_manager/show_good_word_suggestions_detail.php
```diff
@@ -45,109 +46,100 @@
...
-                " &nbsp; | &nbsp; ";
-            echo "<b>" . _("Context") . "</b>:<br><span class='mono'>$context_string</span><br>";
-        }
-        echo "</p>";
-        echo "<hr>";
-
+    echo "<p><b>" . _("Date") . "</b>: " . strftime($datetime_format,$time) . "<br>";
+    echo "<b>" . _("Round") . "</b>: $round &nbsp; | &nbsp; ";
+    echo "<b>" . _("Proofreader") . "</b>: " . private_message_link($proofer) . "<br>";
+    echo "<b>" . _("Page") . "</b>: <a href='$code_url/tools/project_manager/show_good_word_suggestions_detail.php?projectid=$projectid&amp;imagefile=$page&amp;word=$encWord&amp;timeCutoff=$timeCutoff'>$page</a><br>";
```
cpeel, line 119: Since this is self-referencing I think `<a href='?projectid=$projectid...>` should work here too.

cpeel (commented):
I'm not able to test your sandbox -- I'm guessing the sandbox-building script got confused with the slash in your branch name.

tools/project_manager/show_good_word_suggestions_detail.php
chrismiceli: I'll fix these and the eslint errors in build tonight.

tools/project_manager/show_good_word_suggestions_detail.js
```diff
@@ -0,0 +1,60 @@
...
+    };
+    var hiddenWord = function() {
+        return $("<input>", {type: 'hidden', name: 'word', value: showGoodWordSuggestionsDetail.word});
+    };
+    var hiddenProjectId = function() {
+        return $("<input>", {type: 'hidden', name: 'projectid', value: showGoodWordSuggestionsDetail.projectid});
+    };
+    let splitPercent = parseFloat(localStorage.getItem(STORAGE_KEY));
+    if(!splitPercent) {
+        // if not defined or 0, set the value to 100 to the text is shown
```
cpeel, line 19: This comment doesn't seem to match what it's actually doing.

srjfoo (approved):
Scrolling and pane content look good now!

cpeel:
This fixes https://github.com/DistributedProofreaders/dproofreaders/issues/562

Issue #557, opened 2021-01-16
Use composer for PHP dependencies


cpeel:
Update the code to use [composer](https://getcomposer.org/) for PHP dependencies. This will open up a wide range of libraries we can use and remove the need to bundle some dependencies with the code.

The first package to move to composer -- and prove out the move -- is Parsedown and remove it from `3rdparty/parsedown`.

This needs to include updates to:
* `credits.php`
* `SETUP/INSTALL.md`
* deployment tooling
* documentation and information for developers on how to keep their sandboxes updated

PR #568, opened 2021-01-23
Use composer as a packaging manager

cpeel:
*This MR is only for merging after the R202102 DP release.*

Rather than trying to bundle PHP dependencies ourselves, move to the [composer](https://getcomposer.org/) PHP dependency manager. This will open up a much wider range of libraries for us to use rather than rolling our own or forcing admins to install dependencies manually. Initially this is for Parsedown with others to come.

Most of this is behind-the-scenes stuff, but you can see the new credits page in the [use-composer](https://www.pgdp.org/~cpeel/c.branch/use-composer/credits.php) sandbox.

I have updated the TEST-server support scripts (`~cpeel/publish-branch.sh`, etc) to include updating composer packages and added a new `~cpeel/configure-sandbox.sh` script that can be used to update primary sandboxes (it does a `./SETUP/configure` in addition to the composer install). I will update the dev wiki with these changes once this is merged in.

Fixes https://github.com/DistributedProofreaders/dproofreaders/issues/557


srjfoo (approved):
Credits look good; I took a quick look at `markdown` in tasks comments and PCs and both looked good.

SETUP/INSTALL.md
```diff
@@ -28,6 +28,11 @@ Ubuntu system package names.
 * MySQL - php-mysql
 * zip - php-zip
 
+### Composer
+PHP package dependencies are managed using [Composer](https://getcomposer.org/).
+You need a recent version of Composer installed to manage the relevant PHP
```
chrismiceli, line 33: are there any known limits on which versions are allowed?

composer.json
```diff
@@ -0,0 +1,11 @@
+{
+    "name": "distributedproofreaders/dproofreaders",
+    "description": "Distributed Proofreaders is a web application intended to ease the process of converting public domain books into e-texts.",
+    "type": "project",
+    "license": "GPL-2.0-or-later",
+    "minimum-stability": "stable",
+    "require": {
+        "erusev/parsedown": "1.8.0-beta-7",
```
chrismiceli, line 8: beta version? Which version did we have before?

credits.php
```diff
@@ -16,24 +16,41 @@
...
+    {
+        echo "<li>";
+        echo "<b><a href='" . $detail["url"] . "'>" . $detail["name"] . "</a></b><br>";
+        if(isset($detail["license_url"]))
+        {
+            echo _("License") . ": <a href='" . $detail["license_url"] . "'>" . $detail["license"] . "</a>";
+        }
+        else
+        {
+            echo _("License") . ": " . $detail["license"];
```
chrismiceli, line 46: should we encode throughout or assume that this is trusted properly encoded data?

pinc/bootstrap.inc
```diff
@@ -27,6 +27,9 @@ include_once($relPath.'site_vars.php');
 // Register autoloader
 spl_autoload_register('dp_class_autoloader');
 
+// Load composer autoloader
+require_once($relPath.'../vendor/autoload.php');
```
chrismiceli, line 31: this is ok on all pages even if they don't use composer deps?

composer.lock
```diff
@@ -0,0 +1,117 @@
...
+                "reference": "fe7a50eceb4a3c867cc9fa9c0aa906b1067d1955"
+            },
+            "dist": {
+                "type": "zip",
+                "url": "https://api.github.com/repos/erusev/parsedown/zipball/fe7a50eceb4a3c867cc9fa9c0aa906b1067d1955",
+                "reference": "fe7a50eceb4a3c867cc9fa9c0aa906b1067d1955",
+                "shasum": ""
+            },
+            "require": {
+                "ext-mbstring": "*",
```
chrismiceli, line 24: parsedown itself seems to depend on mbstring some, this won't conflict with our version (until we switch to composer mbstring)?

chrismiceli (approved):

composer.json
cpeel: This is actually what we are using now. I don't know why it's labeled as beta `<shrug>`.

SETUP/INSTALL.md
cpeel: Not that I know of. TEST has the version included with Ubuntu 20.04, but any 1.x and 2.x version should work. We aren't using any of the new 2.x features (and if we did, the JSON file will still work with 1.x).

pinc/bootstrap.inc
cpeel: Yup. This just loads the autoloader so that when classes are used the autoloader finds the right files to load. The docs on the [autoloader](https://www.php.net/manual/en/language.oop5.autoload.php) for more information on how this works..

composer.lock
cpeel: The dproofreaders code already [requires the mbstring extension](https://github.com/DistributedProofreaders/dproofreaders/blob/master/SETUP/INSTALL.md#php) which is what this line specifies in the composer file, so we're good here.

pinc/bootstrap.inc
chrismiceli: Nice, i learn something every day

credits.php
cpeel: That's a good catch. I will add the appropriate escaping.

cpeel: And fixed.

chrismiceli (approved):

srjfoo (approved):

PR #571, opened 2021-01-31
Move button menu function into enhanced frame

cpeel:
Move the enhanced interface's button output function out of it's own file and into the calling file. More importantly, make the function entirely PHP and not a horrid mismatch of HTML + PHP.

This is in prep for some code styling standardization work. The mismash of HTML + PHP did not play nice, hence this MR.

Testable in [phpize-button-menu](https://www.pgdp.org/~cpeel/c.branch/phpize-button-menu/) sandbox. This only impacts the enhanced interface. There should be no functional changes.


srjfoo (commented):
Buttons all seem to work the same as they do in the current version.

srjfoo (approved):

chrismiceli (approved):

PR #572, opened 2021-02-01
1494: Modify the behaviour of the [** ] button

chrismiceli:
modify behavior
add test

### DIFF
--- a/tools/proofers/dp_proof.js
+++ b/tools/proofers/dp_proof.js
@@ -202,6 +202,8 @@ function insertTags(tagOpen, tagClose, replace, retainSelection) {
         if (selection === '') {
             tagOpen = tagOpen.replace(': ', '');
         }
+    } else if (tagOpen === '[** ' && selection) {
+        tagOpen = selection + tagOpen;
     }
 
     var subst = tagOpen + selection + tagClose;
--- a/SETUP/tests/jsTests/dp_proofTests.js
+++ b/SETUP/tests/jsTests/dp_proofTests.js
@@ -1,0 +1,89 @@
+/* global QUnit surroundSelection */
+
+QUnit.module("dp_proof tests", function(hooks) {
+    let setTestText = (text, start, end) => {
+        window.docRef = {
+            editform: {
+                // eslint-disable-next-line camelcase
+                text_data: document.createElement('textarea'),
+            }
+        };
+        window.docRef.editform.text_data.value = text;
+        window.docRef.editform.text_data.setSelectionRange(start, end, 'forward');
+    };
+
+    hooks.beforeEach(function() {
+        window.inFace = 0;
+    });
+
+    hooks.afterEach(function() {
+        delete window.docRef;
+        delete window.inFace;
+    });
+
+    QUnit.test("surroundSelection surrounds selection", function (assert) {
+        setTestText('hello', 0, 5);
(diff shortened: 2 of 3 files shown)


chrismiceli:
I'm not sure if this is something we want, and if so I can revert the changes and just commit the tests, which I think add value. Also, something seems up with the sandbox script @cpeel so I couldn't get the sandbox up. Since it is a trivial change, I did not try to set up the branch sandbox myself in the interest of saving time. I will of course do the work if desired.

tools/proofers/dp_proof.js
```diff
@@ -202,6 +202,8 @@ function insertTags(tagOpen, tagClose, replace, retainSelection) {
         if (selection === '') {
             tagOpen = tagOpen.replace(': ', '');
         }
+    } else if (tagOpen === '[** ' && selection) {
+        tagOpen = selection + tagOpen;
```
srjfoo, line 206: Stupid question, since I can't see it in operation ... does the assignment order matter? As shown below in line 209, the actual order should be tagOpen + selection + tagClose.

chrismiceli: The tagClose gets added later. I'm just here modifying the tagOpen to be [** selection. TagClose of ] gets added below.

srjfoo: That was my assumption -- what confused me was the order of `tagOpen = selection + tagOpen;` in line 206 being `selection + tagOpen` instead of `tagOpen + selection`. (Very literal-minded, here, out of ignorance.)

cpeel:
> Also, something seems up with the sandbox script @cpeel so I couldn't get the sandbox up.

I have the publish script fixed. My fix to address `/`s in the branch name was incomplete.

tools/proofers/dp_proof.js
chrismiceli: Let's say selection was "Hello". on 206, we convert tagOpen from "[** " to "Hellow[** ". Then later when we update the text area, we'll change "Hello" to tagOpen + selection + tagClose which turns into "Hello[** Hello]". I'll update sandbox with the code once I get off later and can log into and run the update script.

srjfoo: Okay, now I understand -- thanks!

chrismiceli:
> 
> 
> > Also, something seems up with the sandbox script @cpeel so I couldn't get the sandbox up.
> 
> I have the publish script fixed. My fix to address `/`s in the branch name was incomplete.

It seems to work:
https://www.pgdp.org/~chrismiceli/c.branch/issue_1494/
It did tell me this:
```
Composer could not find a composer.json file in /home/chrismiceli/public_html/c.branch/issue_1494
To initialize a project, please create a composer.json file as described in the https://getcomposer.org/ "Getting Started" section
```
but I just ignored it and all seems well.

cpeel:
> It seems to work:
> https://www.pgdp.org/~chrismiceli/c.branch/issue_1494/
> It did tell me this:
> 
> ```
> Composer could not find a composer.json file in /home/chrismiceli/public_html/c.branch/issue_1494
> To initialize a project, please create a composer.json file as described in the https://getcomposer.org/ "Getting Started" section
> ```

Yup, the script is prep'd for the composer-based branches now and will report that error for pre-composer branches.

srjfoo (approved):
Looks good to me.

Issue #563, opened 2021-01-17
Add htmlpurifier for sanitizing HTML user input


cpeel:
Add [htmlpurifier](http://htmlpurifier.org/) for sanitizing HTML user input. This is available via composer and will replace the PROD and TEST hacks that use the same tool.

PR #573, opened 2021-02-01
Sanitize HTML with htmlpurifier

cpeel:
Sanitize HTML-formatted content from users with htmlpurifier before presenting it to the user.

This is the same HTML sanitizer already hacked into TEST and PROD, this just pulls it into the codebase officially via composer.

Testable in the [use-htmlpurifier](https://www.pgdp.org/~cpeel/c.branch/use-htmlpurifier) sandbox.

Fixes https://github.com/DistributedProofreaders/dproofreaders/issues/563

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,2 @@
 pinc/3rdparty
+vendor
--- a/SETUP/check_require_login.php
+++ b/SETUP/check_require_login.php
@@ -109,18 +109,18 @@ $ok_files = [
 $files = get_all_php_files("../");
 foreach($files as $file)
 {
-    echo "$file\n";
-
-    # If it requires authentication, skip it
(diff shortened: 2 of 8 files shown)


chrismiceli:
We're not going to currently do anything for existing projects? Or did you run a script against PROD?

chrismiceli (approved):

cpeel:
> We're not going to currently do anything for existing projects? Or did you run a script against PROD?

We sanitize the comments on output, not on save, so this is live for all existing and new projects.

srjfoo:
> > We're not going to currently do anything for existing projects? Or did you run a script against PROD?
> 
> We sanitize the comments on output, not on save, so this is live for all existing and new projects.

And PMs have been notified that it's happening.

srjfoo (approved):

Issue #558, opened 2021-01-16
Replace portable-utf8 with composer-supported version


cpeel:
To support Unicode, we pulled in [portable-utf8](http://pageconfig.com/post/portable-utf8) as it provided Unicode tooling in a single file. The file is not being kept up to date for new PHP versions, however, and after we move to use composer (https://github.com/DistributedProofreaders/dproofreaders/issues/557) we can leverage this composer-based portable-utf8 library: https://github.com/voku/portable-utf8

cpeel:
These are the functions we currently use from `pinc/3rdparty/portable-utf8/portable-utf8.php`:

```
is_utf8
utf8_access
utf8_bom
utf8_chr
utf8_chr_to_hex
utf8_codepoints
utf8_ord
utf8_range
utf8_split
utf8_str_replace
utf8_string
utf8_string_has_bom
utf8_strlen
utf8_substr
utf8_url_slug
```

cpeel:
Because the dproofreaders code explicitly requires mbstring and intl and (at the time we do this) a minimum PHP version of 7.4, we should exclude the symfony polyfills in `composer.json`:
```json
{
    "replace": {
        "symfony/polyfill-php72": "*",
        "symfony/polyfill-php74": "*",
        "symfony/polyfill-mbstring": "*",
        "symfony/polyfill-intl-grapheme": "*",
        "symfony/polyfill-intl-normalizer": "*",
    }
}
```
See https://github.com/voku/portable-utf8#install-portable-utf-8-via-composer-require and https://github.com/symfony/polyfill#design.

PR #574, opened 2021-02-05
Move to voku/portable-utf8

cpeel:
Replace our single-file [portable-utf8](http://pageconfig.com/post/portable-utf8) with [voku/portable-utf8](https://github.com/voku/portable-utf8) which is better supported.

While this is mostly a straightforward change, it does alter one of our fundamental Unicode building blocks so it's rather intrusive. It will be important for us to do a thorough touch of all the major Unicode areas (adding files to project with invalid characters for the project, proofreading, WordCheck, charsuites rendering, etc).

This is testable in the [composer-portable-utf8](https://www.pgdp.org/~cpeel/c.branch/composer-portable-utf8/) sandbox.

This shouldn't be merged in until after PROD is upgraded to PHP 7.4 (hopefully this month).

Fixes https://github.com/DistributedProofreaders/dproofreaders/issues/558


SETUP/lint_charsuites.php
```diff
@@ -1,9 +1,12 @@
 #!/usr/bin/env php
 <?php
 $relPath="../pinc/";
+include_once($relPath."base.inc");
 include_once($relPath."CharSuites.inc");
 include_once($relPath."unicode.inc");
 
+use voku\helper\UTF8;
```
chrismiceli, line 8: if you wanted, you could do ```use voku\helper\UTF8 as utf8;``` to avoid updating the rest of the file, but I think I prefer this.

chrismiceli (approved):

SETUP/lint_charsuites.php
cpeel: I don't think so - `UTF8::` is a specifier to access a static class function whereas `utf8_chr_to_hex()` is a function name.

srjfoo (approved):
I've worked over loading projects with a variety of the test files I used for earlier testing (adding invalid characters for one project and adding the same files to another project for which the charsuite is valid, adding characters that are rejected for all projects), poked at WC and used the character pickers in the proofing interface, as well as looked at the character suite pages. Everything seems to be working as far as I can tell. Nudge me if there are other checks you can think of that I should make.

PR #575, opened 2021-02-09
Update messages for better translation

srjfoo:
This is a collection of mostly small tweaks to the phrasing of a few
of the user-visible messages, to make them easier to localize, or
to standardize them with similar messages. Fixed a couple of typos.

Viewable in the [message-tweaks](https://www.pgdp.org/~srjfoo/c.branch/message-tweaks/) sandbox.

### DIFF
--- a/faq/pophelp/prefs/prefs_pophelp.inc
+++ b/faq/pophelp/prefs/prefs_pophelp.inc
@@ -116,8 +116,8 @@ so be sure that your current email address is on record.</p>")
     'content' =>
 _("<p>This setting controls who is able to see your statistics.</p>
 <ul>
-<li><b>Private:</b> Only registered users.</li>
-<li><b>Anonymous:</b> Only you.</li>
+<li><b>Private:</b> Only logged-in users.</li>
+<li><b>Anonymous:</b> Only you and administrators.</li>
 </ul>")
     ),
 
--- a/pinc/charsuite-symbols-collection.inc
+++ b/pinc/charsuite-symbols-collection.inc
@@ -2,7 +2,7 @@
 include_once($relPath."CharSuites.inc");
 
 $charsuite = new CharSuite("symbols-collection", _("Symbols collection"));
-$charsuite->description = _("Astronomical, zodiac, and a limited number of apothecary and music symbols. The zodiac symbols are provided with the variation selector, VS15, to try to force them to display as ordinary charcters instead of as emoji. This may not work for all fonts.");
+$charsuite->description = _("Astronomical, zodiac, apothecary and music symbols.");
 
 $charsuite->codepoints = [
     'U+0292',
--- a/pinc/project_edit.inc
+++ b/pinc/project_edit.inc
@@ -110,7 +110,7 @@ function abort_if_cant_edit_project( $projectid )
     }
     else
     {
-        echo "unexpected return value from user_can_edit_project: '$result'";
+        echo _("unexpected return value from user_can_edit_project") . ": '$result'";
         exit;
     }
 }
--- a/project.php
+++ b/project.php
@@ -1003,7 +1003,7 @@ function do_early_uploads()
         echo "<input type='hidden' name='tpnv' value='1'>\n";
         echo "<b>", _("Add title page and verso from directory or zip file:"), "</b>";
         echo "<br>\n";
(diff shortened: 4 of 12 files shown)


chrismiceli (approved):

pinc/project_edit.inc
```diff
@@ -110,7 +110,7 @@ function abort_if_cant_edit_project( $projectid )
     }
     else
     {
-        echo "unexpected return value from user_can_edit_project: '$result'";
+        echo _("unexpected return value from user_can_edit_project") . ": '$result'";
```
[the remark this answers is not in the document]
srjfoo: I asked Casey about that before I made this change -- he recommended just making it look like the other cases where `unexpected return value from user_can_edit_project` is used (the others were already `_()`ed).

project.php
```diff
@@ -1003,7 +1003,7 @@ function do_early_uploads()
         echo "<input type='hidden' name='tpnv' value='1'>\n";
         echo "<b>", _("Add title page and verso from directory or zip file:"), "</b>";
         echo "<br>\n";
-        $initial_rel_source = "$user_dir/";
+        $initial_rel_source = "Users/$user_dir/";
```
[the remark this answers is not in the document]
srjfoo: Perhaps so, but I don't know if it's worth doing. I think the only two places it's used are in this file. The only reason for updating this was so that it wouldn't be flat-outright wrong if a site chooses to enable metadata. The old version of the upload script was much less complex, and did not have separate `Users` and `Commons` directories; I suspect this just got overlooked when the script was enhanced.

faq/pophelp/prefs/prefs_pophelp.inc
```diff
@@ -116,8 +116,8 @@ so be sure that your current email address is on record.</p>")
     'content' =>
 _("<p>This setting controls who is able to see your statistics.</p>
 <ul>
-<li><b>Private:</b> Only registered users.</li>
-<li><b>Anonymous:</b> Only you.</li>
+<li><b>Private:</b> Only logged-in registered users.</li>
```
cpeel, line 119: `registered` seems redundant here -- they can't be logged in if they're not registered.

pinc/charsuite-symbols-collection.inc
```diff
@@ -2,7 +2,7 @@
 include_once($relPath."CharSuites.inc");
 
 $charsuite = new CharSuite("symbols-collection", _("Symbols collection"));
-$charsuite->description = _("Astronomical, zodiac, and a limited number of apothecary and music symbols. The zodiac symbols are provided with the variation selector, VS15, to try to force them to display as ordinary charcters instead of as emoji. This may not work for all fonts.");
+$charsuite->description = _("Astronomical, zodiac, and a limited number of apothecary and music symbols. The zodiac symbols are provided with the variation selector, VS15, to try to force them to display as ordinary characters instead of as emoji. This may not work for all fonts.");
```
cpeel, line 5: While we're here did we want to do the update to this recommended by bunny-crunch?

pinc/project_edit.inc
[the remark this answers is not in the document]
cpeel: Yeah, better to be consistent with the other strings at this point. As I was telling Sharon, we do a poor job of separating dev-level messages vs user-level messages -- both on what to display and what to log.

faq/pophelp/prefs/prefs_pophelp.inc
srjfoo: Good point -- and we switched from `registered` to `logged in` elsewhere. I'll get that change in when I get the charsuite one done.

pinc/charsuite-symbols-collection.inc
srjfoo: Good idea, thanks. I would like to verify with him my understanding of what he thinks it should say, so probably tomorrow before I get the change pushed.

faq/pophelp/prefs/prefs_pophelp.inc
srjfoo: Done.

pinc/charsuite-symbols-collection.inc
srjfoo: Done.

Issue #569, opened 2021-01-24
Duplicate key error when a user commits to SR something twice


cpeel:
If a user commits to SR a project twice, the DB query to log this fails with a duplicate key error:

```
DPDatabase.inc - log_error from /data/htdocs/c/pinc/smoothread.inc:80: Duplicate entry '<KEYVAL>' for key 'PRIMARY'
```

This is because `sr_commit()` in `pinc/smoothread.inc` blindly does an insert. This query needs to be updated to use `ON DUPLICATE KEY UPDATE` to update the entry if it already exists.

PR #576, opened 2021-02-09
Fix smoothreading SQL error on dup commit

cpeel:
This fixes https://github.com/DistributedProofreaders/dproofreaders/issues/569 (in one commit) and updates the SQL in `pinc/smoothread.inc` to use the `DPDatabase` escaping and query functions (in the second commit).

Testable in the [smoothread-sql](https://www.pgdp.org/~cpeel/c.branch/smoothread-sql/) sandbox.

### DIFF
--- a/pinc/smoothread.inc
+++ b/pinc/smoothread.inc
@@ -28,7 +28,7 @@ include_once($relPath.'user_project_info.inc'); // notify_project_event_subscrib
 *
 * Remarks:
 *
-* Currently only a check for existence of record and field being <> 0 is done. The latter
+* Currently only a check for existence of record and field being > 0 is done. The latter
 * would allow the extension for a timestamp and the resetting to 0 if revoked in order
 * to still show users that have once committed but then revoked that commitment.
 *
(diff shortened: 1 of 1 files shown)


pinc/smoothread.inc
```diff
@@ -28,31 +28,25 @@ include_once($relPath.'user_project_info.inc'); // notify_project_event_subscrib
...
-
-    if ($row['committed'] <> 0)
-    {
-        return true;
-    }
-
-    return false;
+        WHERE projectid = '%s'
+            AND user = '%s'
+            AND committed > 0
```
chrismiceli, line 44: it's never negative right? Before we had <> 0 and now just > 0

chrismiceli (approved):

pinc/smoothread.inc
cpeel: Fair question. The field is a `tinyint` and right now we only store 1s in it (we delete rows instead of setting it to 0) so `> 0` is valid for how we're currently using it.

70ray (approved):

srjfoo (approved):
Looks good to me.

PR #577, opened 2021-02-10
use api for page browser

70ray:
This uses the new api for page browser but should appear much as before.
It changes the url to reflect the current situation.
There is now only one javascript file. The top level functions could be put in another file as before or moved inside the $( function() - document ready)
There was a slight problem with the splitter which now changes the style of the container which can have undesirable consequences if you re-use the contaner. This is avoided by not re-using the container.
We should perhaps further consider the simpleHeader option which was intended to be used in image-only mode but can also be used otherwise by manipulating the url.
This should be running in https://www.pgdp.org/~rp31/c.branch/browse_api_2


chrismiceli:
This appears to break https://www.pgdp.org/~rp31/c.branch/browse_api_2/tools/project_manager/show_all_good_word_suggestions.php which uses the page browser.

cpeel:
> This appears to break https://www.pgdp.org/~rp31/c.branch/browse_api_2/tools/project_manager/show_all_good_word_suggestions.php which uses the page browser. Namely the frame show_good_word_suggestions_detail.php

Thinking about these embedded cases, we probably need a flag to not manipulate the URL when the page browser doesn't have the whole window.

cpeel:
This is quite cool!

We do need to retain the current control pane layout, however. There were A Lot Of Opinions about that and this should be a mostly behind-the-scenes change.

chrismiceli:
> 
> 
> > This appears to break https://www.pgdp.org/~rp31/c.branch/browse_api_2/tools/project_manager/show_all_good_word_suggestions.php which uses the page browser. Namely the frame show_good_word_suggestions_detail.php
> 
> Thinking about these embedded cases, we probably need a flag to not manipulate the URL when the page browser doesn't have the whole window.

Yeah, I was thinking about this when I was consuming this page browser component. It could be refactored a little to make page interactions configurable more easily. A lot of the code is shared between the 2 users of it.

70ray:
I changed the file "page_controls.inc" without realising it was used anywhere else. I'll restore the missing function.

70ray:
Missing pieces are now restored so show_all_good_word_suggestions.php should work now.

70ray:
There are now several duplicate named javascript functions which I should do something about, although it all seems to work.

cpeel:
I'm not certain restoring those was the right answer. I think you were on the right track before and we need the Page Browser to support an embedded and non-embedded mode and update the embedded usecases accordingly.

70ray:
show_all_good_word_suggestions is still using the old functions. It wasn't working because I'd removed the files it was looking for, not realising they were used elsewhere than in page_browser. What we really need to do is to incorporate the new functions into it. I'll try doing that.

cpeel:
> show_all_good_word_suggestions is still using the old functions. It wasn't working because I'd removed the files it was looking for, not realising they were used elsewhere than in page_browser. What we really need to do is to incorporate the new functions into it. I'll try doing that.

Right, which is why I'm confused that the fix was adding those files back in rather than updating `show_all_good_word_suggestions.php`.

70ray:
> We do need to retain the current control pane layout, however. There were A Lot Of Opinions about that and this should be a mostly behind-the-scenes change.
I'll change it so it looks like the old one.

tools/project_manager/show_good_word_suggestions_detail.php
```diff
@@ -136,11 +131,8 @@
 echo "</div>";
 
 echo "<div class='overflow-hidden display-flex' style='flex-direction: column;'>";
-if (isset($imagefile)) {
-    echo "<div id='page-browser' class='overflow-hidden'></div>";
-} else {
-    echo "<p style='margin: 0.5em'>" . _("Select one of the page links to view the page image (scan).") . "</p>";
-}
+echo "<iframe name='imageframe' style='border: none;' title='Image' width='100%' height='100%' srcdoc='<p>" . _("Select one of the page links to view the page image (scan).") . "</p>'></iframe>";
```
cpeel, line 134: We shouldn't have to use an iframe here -- we need to be able to embed the PageBrowser in other pages as a reusable widget.

70ray: I'll see what I can do.

cpeel:
The `show_good_word_suggestions_detail` looks much better, thanks.

> > We do need to retain the current control pane layout, however. There were A Lot Of Opinions about that and this should be a mostly behind-the-scenes change.
> 
> I'll change it so it looks like the old one.

Still not quite there yet, compare:
* [your sandbox](https://www.pgdp.org/~rp31/c.branch/browse_api_2/tools/page_browser.php?project=projectID45c225f598e32&imagefile=004.png&mode=imageText&round_id=OCR)
* [curren TEST](https://www.pgdp.org/c/tools/page_browser.php?project=projectID45c225f598e32&mode=imageText&imagefile=004.png)

70ray:
I'm making some more changes.

70ray:
Control layout should now match the previous version.

scripts/page_browse.js
```diff
@@ -470,10 +480,12 @@ function pageBrowse(params, replaceUrl) {
...
                 selectAProject();
             });
-            fixHead.append($("<p>").append(projectData.title, resetButton));
+            const projectRef = new URL(proofIntData.projectFile);
+            projectRef.searchParams.append("id", projectId);
+            fixHead.append($("<p>").append($("<a>", {href: projectRef}).append(projectData.title)), resetButton);
         }
         // get pages
-        $.ajax(makeApiAjaxSettings("v1/projects/" + projectId + "/pages")).done(displayPages)
+        $.ajax(makeApiAjaxSettings(`v1/projects/${projectId}/pages`)).done(displayPages)
```
chrismiceli, line 488: we're ok since no IE support, right?

chrismiceli (approved):
Do we need to add/alter unit tests?

scripts/page_browse.js
cpeel: We have officially dropped IE support. 👍

srjfoo:
Focusing on WordCheck for the moment, compare this page that was brought up from the [Words that WordCheck would currently flag](https://www.pgdp.org/~rp31/c.branch/browse_api_2/tools/project_manager/show_word_context.php?projectid=projectID5e2699b12eb09&word=496c6c) with this from [Suggestions from proofreaders](https://www.pgdp.org/~rp31/c.branch/browse_api_2/tools/project_manager/show_good_word_suggestions_detail.php?projectid=projectID5e2699b12eb09&word=747269) (It's especially dramatic if you look at it in Charcoal, but there's a difference, also, in the light themes).

Is there a reason for going to a button for selecting a page instead of the link? Perhaps it's a flaw in our approach of making the buttons unobtrusive, but the link is much more visible than the button is. Do we need to re-think the button styling?

When I bring up a page in the browser (any version of it, as far as I can tell), those in the `browse_api_2` sandbox do not allow navigation via `option`/`alt` up- and down-arrows, even when I explicitly click on the text/image to make sure they have focus. Is it possible to restore that?

70ray:
> Focusing on WordCheck for the moment, compare this page that was brought up from the [Words that WordCheck would currently flag](https://www.pgdp.org/~rp31/c.branch/browse_api_2/tools/project_manager/show_word_context.php?projectid=projectID5e2699b12eb09&word=496c6c) with this from [Suggestions from proofreaders](https://www.pgdp.org/~rp31/c.branch/browse_api_2/tools/project_manager/show_good_word_suggestions_detail.php?projectid=projectID5e2699b12eb09&word=747269) (It's especially dramatic if you look at it in Charcoal, but there's a difference, also, in the light themes).
> 
Perhaps we should look at re-working show_word_context.php so it also does not use frames. It looks like it could use some code in common.
> Is there a reason for going to a button for selecting a page instead of the link? Perhaps it's a flaw in our approach of making the buttons unobtrusive, but the link is much more visible than the button is. Do we need to re-think the button styling?
> 
Using a button seems like a more natural way of doing it when the action doesn't involve navigating to a new page. These buttons do not have a style applied by us, in Firefox on Linux they look much larger than in Firefox on Windows. Perhaps we should style them so they look somilar in different browsers and OSs.
> When I bring up a page in the browser (any version of it, as far as I can tell), those in the `browse_api_2` sandbox do not allow navigation via `option`/`alt` up- and down-arrows, even when I explicitly click on the text/image to make sure they have focus. Is it possible to restore that?

I removed that because it was interacting with a built-in function of the select control that responds to Alt up- and down-arrows and thought no-one would notice :) but that function does not seem to do anything useful now. I'll have another look at it.

tools/project_manager/show_good_word_suggestions_detail.php
```diff
@@ -49,8 +49,6 @@
 $details = json_encode([
     "layout" => $layout,
```
70ray, line 50: It would be nice to be able to switch layout without reloading the page. But this code saves layout changes in user settings. To do this we would have to implement an api for GET and PUT to user settings. Do we want to do this or would it be better to save changes locally instead?

70ray, line 50: It would be nice to be able to switch layout without reloading the page. But this code saves layout changes in user settings. To do this we would have to implement an api for GET and PUT to user settings. Do we want to do this or would it be better to save changes locally instead?

cpeel: It would be better if we just kept this locally.

cpeel:
> > Focusing on WordCheck for the moment, compare this page that was brought up from the [Words that WordCheck would currently flag](https://www.pgdp.org/~rp31/c.branch/browse_api_2/tools/project_manager/show_word_context.php?projectid=projectID5e2699b12eb09&word=496c6c) with this from [Suggestions from proofreaders](https://www.pgdp.org/~rp31/c.branch/browse_api_2/tools/project_manager/show_good_word_suggestions_detail.php?projectid=projectID5e2699b12eb09&word=747269) (It's especially dramatic if you look at it in Charcoal, but there's a difference, also, in the light themes).
> 
> Perhaps we should look at re-working show_word_context.php so it also does not use frames. It looks like it could use some code in common.

Yes, @chrismiceli did one of the three tools mentioned in https://github.com/DistributedProofreaders/dproofreaders/issues/562. We shouldn't try to boil the ocean in this one MR, however, as it makes it very hard to review. It would be better to make this MR as small as possible (while also being complete) and have a subsequent one that makes iterative changes to these tools.

70ray:
More detail on the page selector:
How it works now is the default: If the page selector has focus, then up arrow moves to the page higher up in the list, i.e. the previous page. Contrariwise for down arrow. Similarly if the zoom control has focus, the up and down arrows change the zoom number.
Previously Alt up or down arrow changes page in the same way if the page selector has focus or not. If the zoom control has focus Alt up or down changes both zoom and page on the first press, then only page. If the page control has focus, arrow without alt changes page on the first press then does nothing.

srjfoo:
> > Is there a reason for going to a button for selecting a page instead of the link? Perhaps it's a flaw in our approach of making the buttons unobtrusive, but the link is much more visible than the button is. Do we need to re-think the button styling?
> 
> Using a button seems like a more natural way of doing it when the action doesn't involve navigating to a new page. These buttons do not have a style applied by us, in Firefox on Linux they look much larger than in Firefox on Windows. Perhaps we should style them so they look similar in different browsers and OSs.

Good point about the styling on the buttons. I think we do actually style at least some of the buttons in Charcoal, but that's more in the nature of an override.

I don't know if a button is more intuitive or not, but I certainly noticed the difference. The button blends in with the text. The link, because of link coloring (at least in the themes I tried it out in (Project Gutenberg and Charcoal)) stands out much more, and if the PM is skimming for the place to click, will be easier to see.

I'm not sure of the wisdom of making this particular change without asking for feedback. PMs may be perfectly okay with, and prefer the button blending in, but I wouldn't count on it.

> 
> > When I bring up a page in the browser (any version of it, as far as I can tell), those in the `browse_api_2` sandbox do not allow navigation via `option`/`alt` up- and down-arrows, even when I explicitly click on the text/image to make sure they have focus. Is it possible to restore that?
> 
> I removed that because it was interacting with a built-in function of the select control that responds to Alt up- and down-arrows and thought no-one would notice :) but that function does not seem to do anything useful now. I'll have another look at it.

:grin: -- I certainly noticed! I loved the ability to navigate between pages with `Alt` and an arrow. Maybe I was the only one, but probably not. Mind you, making sure the page selector has focus and navigating that way is only slightly awkward, and I can definitely get used to that method too.

What's super-nice is that there no longer appears to be a flicker when changing pages!

For what it's worth, in this setup (and remembering I'm on a Mac, not Windows), if the page selector has focus, I don't need the `Alt` key -- just the down-arrow.

srjfoo:
> > > Focusing on WordCheck for the moment, compare this page that was brought up from the [Words that WordCheck would currently flag](https://www.pgdp.org/~rp31/c.branch/browse_api_2/tools/project_manager/show_word_context.php?projectid=projectID5e2699b12eb09&word=496c6c) with this from [Suggestions from proofreaders](https://www.pgdp.org/~rp31/c.branch/browse_api_2/tools/project_manager/show_good_word_suggestions_detail.php?projectid=projectID5e2699b12eb09&word=747269) (It's especially dramatic if you look at it in Charcoal, but there's a difference, also, in the light themes).
> > 
> > 
> > Perhaps we should look at re-working show_word_context.php so it also does not use frames. It looks like it could use some code in common.
> 
> Yes, @chrismiceli did one of the three tools mentioned in #562. We shouldn't try to boil the ocean in this one MR, however, as it makes it very hard to review. It would be better to make this MR as small as possible (while also being complete) and have a subsequent one that makes iterative changes to these tools.

I totally agree about not trying to do everything at once! My comment was based on me not knowing exactly what I should be looking at, and documenting the differences I noticed between the sandbox and what's active on TEST.

To clarify -- ignore differences in the styling, and just look at the operation of the page browser function wherever it appears?

70ray:
Re: Buttons, yes they are a different (better) size in charcoal theme! We should adjust the styling so this size is applied to other themes also.
This is Classic grey in Linux 
image

This is charcoal in Linux:
image


 Another point about the links/buttons is that previously you could open the link in a new window or tab. I don't know if anyone ever did this but these options are not possible in the widget version.

I can restore the Alt arrow functions to how they were before.

tools/project_manager/show_good_word_suggestions_detail.php
```diff
@@ -119,7 +118,7 @@
     echo "<p><b>" . _("Date") . "</b>: " . strftime($datetime_format,$time) . "<br>";
     echo "<b>" . _("Round") . "</b>: $round &nbsp; | &nbsp; ";
     echo "<b>" . _("Proofreader") . "</b>: " . private_message_link($proofer) . "<br>";
-    echo "<b>" . _("Page") . "</b>: <a href='?projectid=$projectid&amp;imagefile=$page&amp;word=$encWord&amp;timeCutoff=$timeCutoff'>$page</a><br>";
+    echo "<b>" . _("Page") . "</b>: <input type='button' class='page-select' value='$page'><br>";
```
cpeel, line 121: Does the `.click()` handler not get run on links? I feel like we should be able to keep this a link and avoid all of the fallout of making this a button.

cpeel:
> Re: Buttons, yes they are a different (better) size in charcoal theme! We should adjust the styling so this size is applied to other themes also.

That's not a good idea for a variety of reasons, mostly because it will either change buttons everywhere or we will have a weird one-off button here. I have to believe there is a way to do this keeping it a link possibly even with the `.click()` event like you have it working now.

srjfoo:
> Another point about the links/buttons is that previously you could open the link in a new window or tab. I don't know if anyone ever did this but these options are not possible in the widget version.

Re: opening the link in a different window -- sometimes I did and sometimes I didn't. Is it just the button that means that you can't open the image in a new window or tab? Or would you not be able to even with a link?

I'm certainly guilty of making assumptions on what the users will be willing to accept and not complain about, and having to adjust my thinking, but this strikes me as something that we should not make assumptions about unless you're prepared to get it torn to pieces in the forums after rollout, and have your justifications ready to trot out if you do get complaints. (I'm a bit gunshy at the moment, so may be over-sensitive. UI/UX design is difficult and often painful, I think.)

That is ... is this the only way it can be done?

> 
> I can restore the Alt arrow functions to how they were before.

👍

cpeel:
> > Another point about the links/buttons is that previously you could open the link in a new window or tab. I don't know if anyone ever did this but these options are not possible in the widget version.
> 
> Re: opening the link in a different window -- sometimes I did and sometimes I didn't. Is it just the button that means that you can't open the image in a new window or tab? Or would you not be able to even with a link?

The interfaces were designed to have the link clicked on and the context frame updated. We should continue with that design and not support having the link open in a new window with a right-click. If there is fallout from that the better approach would be to add a second link next to it that would open the full page browser in its own window.

tools/project_manager/show_good_word_suggestions_detail.php
70ray: No doubt we can make it a link somehow.

srjfoo:
Whoops -- I just realized that I got this partially composed, but not sent.

> > > Another point about the links/buttons is that previously you could open the link in a new window or tab. I don't know if anyone ever did this but these options are not possible in the widget version.
> > 
> > 
> > Re: opening the link in a different window -- sometimes I did and sometimes I didn't. Is it just the button that means that you can't open the image in a new window or tab? Or would you not be able to even with a link?
> 
> The interfaces were designed to have the link clicked on and the context frame updated. We should continue with that design and not support having the link open in a new window with a right-click. If there is fallout from that the better approach would be to add a second link next to it that would open the full page browser in its own window.

It's a thorny problem. If it can't be opened in a new window because of the interface design, then maybe it shouldn't be shown as a link -- people expect to be able to right-click on links and open them in a new window or tab.

srjfoo:
Quoting from a couple of conversations, since this discussion kind of got split ...

> > > > Another point about the links/buttons is that previously you could open the link in a new window or tab. I don't know if anyone ever did this but these options are not possible in the widget version.
> > > 
> > > 
> > > Re: opening the link in a different window -- sometimes I did and sometimes I didn't. Is it just the button that means that you can't open the image in a new window or tab? Or would you not be able to even with a link?
> > 
> > 
> > The interfaces were designed to have the link clicked on and the context frame updated. We should continue with that design and not support having the link open in a new window with a right-click. If there is fallout from that the better approach would be to add a second link next to it that would open the full page browser in its own window.
> 
> It's a thorny problem. If it can't be opened in a new window because of the interface design, then maybe it shouldn't be shown as a link -- people expect to be able to right-click on links and open them in a new window or tab.

And ...

> > Re: Buttons, yes they are a different (better) size in charcoal theme! We should adjust the styling so this size is applied to other themes also.
> 
> That's not a good idea for a variety of reasons, mostly because it will either change buttons everywhere or we will have a weird one-off button here. I have to believe there is a way to do this keeping it a link possibly even with the `.click()` event like you have it working now.

I've thought about this for several hours, and even though I'm concerned about having a link not behave as I would ordinarily expect a link to behave, I think that it's better to leave it visually a link if possible, for visibility.

70ray:
> > Re: Buttons, yes they are a different (better) size in charcoal theme! We should adjust the styling so this size is applied to other themes also.

This is no longer relevant but the unusually large buttons only seem to appear in Firefox on Linux. All other browser/OS combinations have more reasonably sized buttons. So it looks like a bug/feature of Firefox/Linux and not anything we need to do anything about.

70ray:
The links are now restored and Alt/Option arrow functions (improved to avoid multiple actions). After the first page, the view of subsequent pages is changed without redrawing the whole area so flicker is reduced.

srjfoo:
> The links are now restored and Alt/Option arrow functions (improved to avoid multiple actions). After the first page, the view of subsequent pages is changed without redrawing the whole area so flicker is reduced.

I definitely appreciate the ability to navigate using `Alt`/`Option` + the up and down arrows, thank you.

In my testing just now, though, while the page that I brought up for context from "Candidates for Good Words List from WordCheck" page works fine bringing up the context image, the corresponding operation from the link brought up from "Candidates for Good Words List from Proofreaders" did not. For example, [this page](https://www.pgdp.org/~rp31/c.branch/browse_api_2/tools/project_manager/show_good_word_suggestions_detail.php?projectid=projectID5e23800f9f15c&word=666f7265746f70) shows "Please select a page" in the pull-down in the image pane, and when I click on the link in the list to the left, I get a message that says "There is no page undefined in this project".

Also, if I go back to the main "Candidates for Good Words List from Proofreaders" page and choose another word without closing the existing context window, another page does not open, it just changes in the previous window without bringing the window to the front -- I have a feeling that's going to be confusing.

This behavior is different from the current behavior, and I don't think we should return to a click on a link overwriting the contents of an existing open window.

pinc/page_controls.inc
```diff
@@ -103,12 +22,15 @@ function get_proofreading_interface_data_js()
...
         "reset" => _("Select a different project"),
         "projectid" => _("Project ID"),
         "selectAPage" => _("Please select a page"),
         "browserDefault" => BROWSER_DEFAULT_STR,
-        "displayImageX" => _("Display Image: %s"),
+        "browsePages" => _("Browse pages"),
+        "displayPageX" => _("Display Page: %s"),
+        "noPages" => _("There are no pages in this project"),
+        "enterID" => _("Please enter a project ID"),
+        "noPageX" => _("There is no page %s in this project"),
```
srjfoo, line 33: In testing, I ran into this as "There is no page undefined in this project", which didn't make sense to me. Looking at the code, it looks like `%s` is supposed to be an image filename, but for some reason, the image filename is not being picked up when I click on the link, so it comes through as the very confusing "undefined"?

70ray: > In my testing just now, though, while the page that I brought up for context from "Candidates for Good Words List from WordCheck" page works fine bringing up the context image, the corresponding operation from the link brought up from "Candidates for Good Words List from Proofreaders" did not. For example, [this page](https://www.pgdp.org/~rp31/c.branch/browse_api_2/tools/project_manager/show_good_word_suggestions_detail.php?projectid=projectID5e23800f9f15c&word=666f7265746f70) shows "Please select a page" in the pull-down in the image pane, and when I click on the link in the list to the left, I get a message that says "There is no page undefined in this project".

I can't reproduce these problems. Can you have a look at the browser console to see if it shows any error messages?

70ray: I saw the "There is no page undefined in this project" message when first starting up my windows computer. I think it was on a tab that was already open in the browser. On refreshing the page it went away so it looks like a transient caching problem.

70ray:
Perhaps we should not proceed with this branch but take a different route so the different pages work consistently:
First de-framify the three files in a simple way using iframes:
project_manager/show_good_word_suggestions_detail.php,
project_manager/show_word_context.php
project_manager/show_all_good_word_suggestions.php

Then convert the first two to use the api widget (instead of "page-browser.php" in an iframe) and leave "show_all_good_word_suggestions.php" with an iframe (which will contain one of the other pages) for the time being.

cpeel:
What I'd like to do is this: let me create a branch off of master called `page-browser-uses-api`. We can then merge this code into that branch as-is. We can then update the three tools in question to use the updated page and get that merged into the `page-browser-uses-api`. Then we can get that out on TEST and get some testing time on it. This is a big enough change (in the right direction!) that getting some end-user testing on seems to make sense.

Assuming this sounds good, I'm going to carve out the `api/v1_projects.inc` change into a separate commit and get that also pushed into `master` since that's a useful bugfix as-is.

70ray:
> What I'd like to do is this: let me create a branch off of master called `page-browser-uses-api`. We can then merge this code into that branch as-is. We can then update the three tools in question to use the updated page and get that merged into the `page-browser-uses-api`. Then we can get that out on TEST and get some testing time on it. This is a big enough change (in the right direction!) that getting some end-user testing on seems to make sense.
> 
> Assuming this sounds good, I'm going to carve out the `api/v1_projects.inc` change into a separate commit and get that also pushed into `master` since that's a useful bugfix as-is.

That sounds like a reasonable plan.

srjfoo:
> > What I'd like to do is this: let me create a branch off of master called `page-browser-uses-api`. We can then merge this code into that branch as-is. We can then update the three tools in question to use the updated page and get that merged into the `page-browser-uses-api`. Then we can get that out on TEST and get some testing time on it. This is a big enough change (in the right direction!) that getting some end-user testing on seems to make sense.
> > Assuming this sounds good, I'm going to carve out the `api/v1_projects.inc` change into a separate commit and get that also pushed into `master` since that's a useful bugfix as-is.
> 
> That sounds like a reasonable plan.

I like the idea, too!

PR #578, opened 2021-02-14
Centralize project functions

cpeel:
Move two project-related functions into the Project class. Also delete some unused code.

Testable in the [centralize-project-functions](https://www.pgdp.org/~cpeel/c.branch/centralize-project-functions/) sandbox.

I tested automodify, project page, project search, and edit project functionality. The metadata pages are untested (and also unused at pgdp.net/org and probably broken in other ways).


pinc/Project.inc
```diff
@@ -554,6 +554,32 @@ class Project
         return $this->can_be_managed_by_current_user || $this->PPer_is_current_user;
     }
 
+    function get_num_pages_in_state($state=NULL, $counter="*")
```
chrismiceli, line 557: might want to explicitly call out that this is sql and won't be encoded in a query by naming it something like $counter_sql or something perhaps?

chrismiceli (approved):

pinc/Project.inc
cpeel: Good call - done.

70ray (approved):

srjfoo (approved):
In addition, I tested the bad project stuff.

PR #580, opened 2021-02-15
Notify PMs about holds preventing projects from transitioning

cpeel:
Per GM / Squirrel request: This adds a notification to the PM when automodify attempts to transition a project but is unable to do so because there is a hold on the project. This is for all project holds, not just holds in Available.

A PM will only get one email unless the hold is removed and re-added, or if a proofreader later proofreads a new page. The latter is to address the scenario where:
1. The project has a hold for Available
2. The project is prevented from transitioning and the PM is notified
3. The PM clears some pages but does not remove the hold
4. Proofreaders proofread more pages
5. The project is (again) prevented from transitioning.

This is testable in the [notify-held-projects](https://www.pgdp.org/~cpeel/c.branch/notify-held-projects/) sandbox. We might need to add some special logging since emails aren't sent out from TEST.

This doesn't address the request for the projects to be moved to Unavailable after a certain period of time. That work will build on top of this.

### DIFF
--- a/SETUP/CHANGELOG.md
+++ b/SETUP/CHANGELOG.md
@@ -3,6 +3,10 @@
 Major changes to this project are documented here. For minor changes,
 see the git history.
 
+## R??????
+Scripts supporting this upgrade are in `SETUP/upgrade/16`
+
+
 ## R202102
 Scripts supporting this upgrade are in `SETUP/upgrade/15`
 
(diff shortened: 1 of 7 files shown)


cpeel:
I mistakenly updated `automodify.php` as if https://github.com/DistributedProofreaders/dproofreaders/pull/578 had been merged in (which it hasn't). Please hold on testing this until I get that merged in and this branch rebased.

chrismiceli (approved):

srjfoo (approved):

PR #583, opened 2021-02-21
Fix various small errors found in php_errors

cpeel:
This addresses various small warnings and notices seen in `php_errors`. It also hides the Delete button for illustrations if the user does not have permissions to manage the project (this was reported in the forums).

Available in the [fix-errors](https://www.pgdp.org/~cpeel/c.branch/fix-errors/) sandbox (not a very creative branch title, I know).

### DIFF
--- a/tools/proofers/images_index.php
+++ b/tools/proofers/images_index.php
@@ -110,7 +110,7 @@ function list_images( $project, $image_names, $these_are_page_images )
     echo "<h4 class='center-align'>$header</h4>";
 
     $show_replace_links = $project->can_be_managed_by_current_user;
-    $show_delete_links = !$these_are_page_images && ($project->state == PROJ_NEW || $project->state == PROJ_P1_UNAVAILABLE);
+    $show_delete_links = $project->can_be_managed_by_current_user && !$these_are_page_images && ($project->state == PROJ_NEW || $project->state == PROJ_P1_UNAVAILABLE);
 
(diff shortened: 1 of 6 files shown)


srjfoo (approved):

tools/proofers/images_index.php
```diff
@@ -214,7 +214,7 @@ function show_delete_all_link( $project, $image_names )
 {
     global $code_url;
 
-    if(($project->state == PROJ_NEW || $project->state == PROJ_P1_UNAVAILABLE) && !empty($image_names))
+    if($project->can_be_managed_by_current_user && ($project->state == PROJ_NEW || $project->state == PROJ_P1_UNAVAILABLE) && !empty($image_names))
```
chrismiceli, line 217: I assume we have a similar check on the actual action, right?

chrismiceli (approved):

tools/proofers/images_index.php
srjfoo: I verified by testing that an unprivileged use can see and press the button, but are told they are not authorized to manage the project. I haven't verified that by looking at the code, though.

srjfoo: I shoul have been more specific. The above comment refers to the existing code on TEST and PROD. The code in @cpeel's fix-errors sandbox correctly hides the button.

70ray (approved):

PR #584, opened 2021-02-21
Message updates and corrections

srjfoo:
In the `ppv_report.php`, updated one of the items to be clearer about the intent. Also deleted old obsolete `ppv_report.txt` file.

One typo correction.

In `for_mentors.php`, updated informational text for mentors.

Changes viewable in [more-message-tweaks](https://www.pgdp.org/~srjfoo/c.branch/more-message-tweaks).


tools/proofers/for_mentors.php
```diff
@@ -117,7 +117,7 @@
 echo "<p>";
 echo sprintf(_("Projects with pages available to Mentors in round %s."), "<b>$mentoring_round->id</b>");
 echo " ";
-echo _("Oldest project listed first.");
+echo sprintf(_("Please check the Saved column in the 'Which proofreader did each page...' section for each project listed below, to make sure you first work on the project with the oldest pages saved in %s."), "<b>$mentored_round->id</b>");
```
chrismiceli, line 120: Are we still encouraging use of <b>? I see we use it above so not a priority.

chrismiceli (approved):

tools/proofers/for_mentors.php
srjfoo: I'm confused -- in what context? This MR was focused only on updating the text.

windymilla (approved):
All looks good - the Englifh change should mean PPers get mor credit for tackling Fraktur, for example.

tools/proofers/for_mentors.php
chrismiceli: Ugh, formatting got all messed up. I meant the b tag. In other MRs we changed tags in favor of styles like I, center, etc.

srjfoo: Ah, I see. I think you have to escape HTML tags -- \\\<b\\\> will produce \<b\>.

Yes, I noticed that, and can change those to \<span class='bold'\> if it's appropriate, but was, in this case, just following the existing pattern. There are other places in the code where those tags are used. It's not important enough to search them all out and change them all at once, I think, but is it appropriate to change them as we find them?

chrismiceli: I'm not super adamant about forcing it now but yeah as we go through it will help if we ever bite the bullet one day and replace all.

cpeel: In this case I would keep the `<b>`s to align with other strings in the file.

PR #586, opened 2021-02-25
Enable the API by default

cpeel:
Now that Page Browser requires the API, enable it by default and
provide additional documentation on why. Also provide more docs
around the use of PHP sessions to access the API.

### DIFF
--- a/SETUP/API.md
+++ b/SETUP/API.md
@@ -5,14 +5,18 @@ such as projects, word lists, and statistics.
 
 ## Configuring
 
-To enable the API, update the following API-related values in your
-`configuration.sh` file and re-configuring your site.
+The API is enabled by default because some UI features, like the Page Browser,
+require it.
 
-* `API_ENABLE` - if set to `TRUE` the API is enabled.
+While disabling the API will break these features, you can still do so by
+updating the `API_ENABLE` setting in your `configuration.sh` file and
+re-configuring your site.
 
-It is also highly recommended that you update your web browser to rewrite
-requests for `/api/` to `api/index.php?url=`. This will enable access to
-URLs like:
+* `API_ENABLE` - if set to `FALSE` the API is disabled.
+
+While not required for the DP UI, for external users it is also highly
+recommended that you update your web server to rewrite requests for `/api/`
+to `api/index.php?url=`. This will enable access to URLs like:
 ```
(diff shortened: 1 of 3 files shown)


SETUP/API.md
```diff
@@ -5,14 +5,18 @@ such as projects, word lists, and statistics.
...
+re-configuring your site.
 
-It is also highly recommended that you update your web browser to rewrite
-requests for `/api/` to `api/index.php?url=`. This will enable access to
-URLs like:
+* `API_ENABLE` - if set to `FALSE` the API is disabled.
+
+While not required for the DP UI, for external users it is also highly
+recommended that you update your web browser to rewrite requests for `/api/`
+to `api/index.php?url=`. This will enable access to URLs like:
```
srjfoo, line 19: I find "web browser" confusing -- is this instruction meant for individual users, or for the installer of the code.

Other than that confusion, looks good.

srjfoo (approved):

SETUP/API.md
cpeel: That should say `web server` - thanks for catching that. I'll get it fixed.

70ray (approved):

SETUP/API.md
```diff
@@ -31,12 +35,18 @@ RewriteRule ^/api/(.*)$ %{DOCUMENT_ROOT}/c/api/index.php?url=$1 [L,QSA]
 
 ## Access Control
 
-API access is limited to individuals with API keys. API keys are associated
-with users and the API calls are restricted to the same information as the
-user who owns it.
+API access is limited to individuals with API keys or valid PHP session keys.
```
chrismiceli, line 38: You think there is value in allowing us to disable api access from api keys while retaining php session keys? This would allow us to disable it say if some security issue is discovered with the key access or some pgdp implementation decides they don't want it?

cpeel: I had considered that and decided that there probably isn't value in splitting them apart. If a pgdp deployment doesn't want to enable session keys they can simply not give any out (since they are manually opt-in) or empty any that have been. Since the API codepath is identical for the two authentication methods, if there is a security issue found the site will want to disable both of them (particularly since all logged-in users have defacto access to the API with the right HTTP request and session cookies).

PR #587, opened 2021-03-01
Deframe show_all_good_word_suggestions and show_word_context

70ray:
show_word_context uses same pattern as show_good_word_suggestions_detail, using a common js file renamed as show_word_context.js since this name seems appropriate for both cases.
show_all_good_word_suggestions uses an iframe
Running in https://www.pgdp.org/~rp31/c.branch/all_suggestions

### DIFF
--- a/tools/project_manager/show_all_good_word_suggestions.php
+++ b/tools/project_manager/show_all_good_word_suggestions.php
@@ -7,7 +7,7 @@ include_once($relPath.'misc.inc'); // attr_safe()
 include_once($relPath.'Stopwatch.inc');
 include_once($relPath.'misc.inc'); // array_get(), get_integer_param(), surround_and_join()
 include_once('./post_files.inc');
-include_once("./word_freq_table.inc");
+include_once("./word_freq_table.inc"); // echo_cutoff_text()
 
 require_login();
 
@@ -102,20 +102,19 @@ if($frame=="left") {
         echo _("View projects for user:") . " <input type='text' name='pm' value='" . attr_safe($pm) . "' size='10'><br>";
     }
 
-echo _("Show") . ": ";
-echo "<select name='timeCutoff'>";
-echo "<option value='0'"; if($timeCutoff==0) echo "selected"; echo ">" . _("All suggestions") . "</option>";
-echo "<option value='-1'"; if($timeCutoff==-1) echo "selected"; echo ">" . _("Suggestions since Good Words List was saved") . "</option>";
-$timeCutoffOptions=array(1,2,3,4,5,6,7,14,21);
-foreach($timeCutoffOptions as $timeCutoffOption) {
-    $timeCutoffValue = ceil((time() - 24*60*60*$timeCutoffOption)/100)*100;
-    echo "<option value='$timeCutoffValue'";
-    if($timeCutoff==$timeCutoffValue) echo "selected";
-    echo ">" . sprintf(_("Suggestions made in the past %d days"),$timeCutoffOption) . "</option>";
-}
-echo "</select>";
-echo "<br>";
-
+    echo _("Show") . ": ";
+    echo "<select name='timeCutoff'>";
+    echo "<option value='0'"; if($timeCutoff==0) echo "selected"; echo ">" . _("All suggestions") . "</option>";
+    echo "<option value='-1'"; if($timeCutoff==-1) echo "selected"; echo ">" . _("Suggestions since Good Words List was saved") . "</option>";
+    $timeCutoffOptions=array(1,2,3,4,5,6,7,14,21);
+    foreach($timeCutoffOptions as $timeCutoffOption) {
+        $timeCutoffValue = ceil((time() - 24*60*60*$timeCutoffOption)/100)*100;
+        echo "<option value='$timeCutoffValue'";
+        if($timeCutoff==$timeCutoffValue) echo "selected";
+        echo ">" . sprintf(_("Suggestions made in the past %d days"),$timeCutoffOption) . "</option>";
+    }
+    echo "</select>";
+    echo "<br>";
 
     echo "<input type='submit' value='Submit'></p>";
     echo "</form>";
--- a/tools/project_manager/show_all_good_word_suggestions.php
+++ b/tools/project_manager/show_all_good_word_suggestions.php
@@ -6,9 +6,9 @@ include_once($relPath.'slim_header.inc');
 include_once($relPath.'misc.inc'); // attr_safe()
 include_once($relPath.'Stopwatch.inc');
 include_once($relPath.'misc.inc'); // array_get(), get_integer_param(), surround_and_join()
-include_once('./post_files.inc');
-include_once("./word_freq_table.inc"); // echo_cutoff_text()
-include_once($relPath.'page_controls.inc');
+include_once($relPath.'page_controls.inc'); // get_proofreading_interface_data_js()
+include_once('./post_files.inc'); // page_info_query()
+include_once("./word_freq_table.inc"); // echo_cutoff_text(), printTableFrequencies(), decode_word()
 
 require_login();
 
@@ -76,7 +76,7 @@ $header_args = [
 
 slim_header(_("Manage Suggestions"), $header_args);
 
-echo "<div id='suggestions_container' style='flex: auto;width: 100%; height: 100%'>";
+echo "<div id='suggestions_container' style='flex: auto; width: 100%; height: 100%'>";
 echo "<div class='overflow-auto'>";
 echo "<div style='padding: 0.5em'>";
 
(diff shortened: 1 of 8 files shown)


tools/project_manager/show_word_context.php
```diff
@@ -40,103 +41,94 @@
...
+];
+
+slim_header(_("Word Context"), $header_args);
+echo "<div id='show_word_context_container' style='flex: auto;width: 100%;height: 100%'>";
+echo "<div class='overflow-auto'>";
+echo "<div style='padding: 0.5em'>";
+
+$project_name = get_project_name($projectid);
+// TRANSLATORS: %1$s is a word, %2$s is the project name.
+echo "<h2>", sprintf(_("Context for '%1\$s' in %2\$s"), $word, $project_name), "</h2>";
```
chrismiceli, line 68: would it make sense to use those multiline strings in php? https://phpf1.com/tutorial/php-heredoc-syntax.html

tools/project_manager/show_word_context.php
```diff
@@ -40,103 +41,94 @@
...
+if($layout == LAYOUT_HORIZ)
+    echo "layout=" . LAYOUT_VERT . "'>" . _("Change to vertical layout");
+else
+    echo "layout=" . LAYOUT_HORIZ . "'>" . _("Change to horizontal layout");
+echo "</a>";
+echo "</p>";
+
+echo "<form method='GET' id='wordInstancesForm'>";
+echo "<input type='hidden' name='projectid' value='$projectid'>";
+echo "<input type='hidden' name='word' value='$encWord'>";
```
chrismiceli, line 82: I think encWord can't be invalid html attribute so we don't need to encode it, right?

chrismiceli (commented):
Looks good to me mostly. Though I'll have to play with it in the sandbox.

tools/project_manager/show_word_context.php
srjfoo: I haven't had time to look at the code yet, but be aware that (unless something has changed in the last year or so) HEREDOC strings are not translatable.

tools/project_manager/show_all_good_word_suggestions.php
```diff
@@ -7,7 +7,8 @@
 include_once($relPath.'Stopwatch.inc');
 include_once($relPath.'misc.inc'); // array_get(), get_integer_param(), surround_and_join()
 include_once('./post_files.inc');
-include_once("./word_freq_table.inc");
+include_once("./word_freq_table.inc"); // echo_cutoff_text()
+include_once($relPath.'page_controls.inc');
```
cpeel, line 11: Please move this up above line 9 so that all of the `$relPath` includes are grouped together

tools/project_manager/show_word_context.php
```diff
@@ -11,6 +11,7 @@
 include_once($relPath.'misc.inc'); // array_get(), get_integer_param()
 include_once('./post_files.inc');
 include_once("./word_freq_table.inc");
+include_once($relPath.'page_controls.inc');
```
cpeel, line 14: Move this up next to the other `$relPath` includes.

tools/project_manager/show_word_context.php
cpeel: Correct - `$encWord` is guaranteed to be safe.

cpeel (commented):
I haven't yet tested this, but sans the two minor things I found this looks good to me in broad strokes.

srjfoo (approved):
Testing looks really good! I especially like how much more smoothly the images are displayed -- no flicker when you change images! And I appreciate being able to use the `alt`/`option` plus up and down arrows to navigate.

Note that in further testing, whether clicking on subsequent `context` links overwrites the previous displayed page depends on whether the page was opened by clicking or explicitly opened in a new window by right-clicking on the `context` link. This is not new behavior, and the overwriting probably goes back to the original design (probably about the time the same thing was done in the proofreading interface, which has since been redesigned). Apologies for not figuring this out before. (I still don't like that behavior, but that's outside the scope of this MR, and as far as I know, PMs have not complained about it loudly.)

I noted one spacing question, but that's it.

tools/project_manager/show_all_good_word_suggestions.php
70ray: Done this.

tools/project_manager/show_word_context.php
70ray: Done this.

tools/project_manager/show_all_good_word_suggestions.php
```diff
@@ -52,55 +48,48 @@
...
-    slim_header(_("Manage Suggestions"), array("js_data" => get_cutoff_script($cutoffOptions,$instances)));
+    "body_attributes" => 'class="no-margin overflow-hidden" style="height: 100vh; width: 100vw"',
+];
 
-    echo "<h1>" . _("Manage Suggestions") . "</h1>";
+slim_header(_("Manage Suggestions"), $header_args);
 
-    // TRANSLATORS: PM = project manager
-    echo "<p><a href='$code_url/tools/project_manager/projectmgr.php' target='_TOP'>" . _("Return to the PM page") . "</a></p>";
+echo "<div id='suggestions_container' style='flex: auto;width: 100%; height: 100%'>";
```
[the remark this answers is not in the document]
70ray: It would be better, yes.

70ray:
> Note that in further testing, whether clicking on subsequent `context` links overwrites the previous displayed page depends on whether the page was opened by clicking or explicitly opened in a new window by right-clicking on the `context` link. This is not new behavior, and the overwriting probably goes back to the original design (probably about the time the same thing was done in the proofreading interface, which has since been redesigned). Apologies for not figuring this out before. (I still don't like that behavior, but that's outside the scope of this MR, and as far as I know, PMs have not complained about it loudly.)

I hadn't realised that there are some pages that look similar but are really different. In show_good_word_suggestions.php left-click opens the links in a new tab (or window) and reuses it but in show_all_good_word_suggestions.php left-click opens the link in a re-usable frame. Right-click always opens in a new tab or window and doesn't re-use it.

70ray:
Now changed the horizontal/vertical switch to use javascript so it is much quicker. It initially uses the value saved in user settings but any changes are saved locally. This will mean that changes will be specific to each browser and device.

tools/project_manager/show_word_context.php
```diff
@@ -28,115 +27,93 @@
...
-$default_layout =  $userSettings->get_value("show_word_context_layout", LAYOUT_HORIZ);
-
-$layout_choices = array(LAYOUT_HORIZ, LAYOUT_VERT);
-$layout = get_enumerated_param($_GET, 'layout', $default_layout, $layout_choices);
-if($layout != $default_layout)
-{
-    $userSettings->set_value("show_word_context_layout", $layout);
-}
+// if not set gives "horizontal"
+$default_layout =  $userSettings->get_value("show_word_context_layout", "horizontal");
```
cpeel, line 31: If we're never setting in the DB (which is fine) we should ditch using this so that we're only using local storage and not a mix of the two. Here and in `tools/project_manager/show_good_word_suggestions_detail.php`.

tools/project_manager/show_word_context.php
```diff
@@ -28,115 +27,93 @@
...
+echo "<div id='show_word_context_container' style='flex: auto;width: 100%;height: 100%'>";
+echo "<div class='overflow-auto'>";
+echo "<div style='padding: 0.5em'>";
+
+$project_name = get_project_name($projectid);
+// TRANSLATORS: %1$s is a word, %2$s is the project name.
+echo "<h2>", sprintf(_("Context for '%1\$s' in %2\$s"), $word, $project_name), "</h2>";
+
+echo "<p>";
+echo "<a href='javascript:void(0)' id='h_v_switch'></a>";
```
cpeel, line 63: I think we should use `href='#'` for these links-that-aren't-a-link, as that appears to be more standard-compliant: https://ourcodeworld.com/articles/read/209/should-i-use-hash-or-javascript-void-0-on-the-href-attribute-of-my-empty-links

tools/project_manager/show_word_context.php
```diff
@@ -28,115 +27,93 @@
...
+        "./show_word_context.js",
+    ],
+    "js_data" => get_proofreading_interface_data_js() . "
+        var showWordContext = $details;",
+
+    "body_attributes" => 'class="no-margin overflow-hidden" style="height: 100vh; width: 100vw"',
+];
+
+slim_header(_("Word Context"), $header_args);
+echo "<div id='show_word_context_container' style='flex: auto;width: 100%;height: 100%'>";
```
cpeel, line 54: nit: spaces after the `;` in the style definition.

cpeel (commented):
One last review from me at the code level.

chrismiceli (approved):
I played with the sandbox and it looks really good. The iframe really simplifies trying to handle all 3 "pages" of form data.

cpeel:
This is the remainder of fixing https://github.com/DistributedProofreaders/dproofreaders/issues/562

PR #588, opened 2021-03-02
Move phpunit tests to GitHub Actions

cpeel:
This moves the phpunit tests over to GitHub actions. The linting and JS tests will be moved over later into separate jobs inside the `static-checks` workflow.

https://github.com/cpeel/dproofreaders/actions/runs/612436083 shows what this will look like after merging.


srjfoo (approved):

.github/workflows/static-checks.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Configure mysql
+        run: |
+          mysql -uroot -proot -e "CREATE DATABASE dp_db CHARACTER SET utf8mb4;"
+          mysql -uroot -proot -e "CREATE USER dp_user@localhost IDENTIFIED BY 'dp_password';"
+          mysql -uroot -proot -e "GRANT ALL ON dp_db.* TO dp_user@localhost;"
+          ./SETUP/configure ./SETUP/tests/travis-ci_configuration.sh .
+      - name: Install DB schema
+        run: |
+          cd SETUP
+          php install_db.php
```
chrismiceli, line 26: we can't run SETUP/install_db.php to avoid cd?

.travis.yml
```diff
@@ -2,32 +2,11 @@ language: php
...
-    - cd SETUP
-    - php install_db.php
-    - cd ..
 before_script:
-    # Specify a version that works with php 7.2, 7.3, 7.4 (see
-    # https://phpunit.de/supported-versions.html)
-    - curl -sSfL -o ~/.phpenv/versions/$(phpenv version-name)/bin/phpunit https://phar.phpunit.de/phpunit-8.phar
-    - phpunit --version
     # We need npm for eslint
     - nvm install 12 --lts
```
chrismiceli, line 8: want me to look at moving node stuff to github actions?

chrismiceli (approved):

70ray (approved):

.github/workflows/static-checks.yml
cpeel: Right now, no, because `install_db.php` assumes `db_schema.sql` is in the current working directory. That could be changed (and I'd welcome an MR to do so) but not in this MR I think.

.travis.yml
cpeel: If you want to move the JS tests, that would be great. I want to handle all language linting in a separate job and it looks like the super-linter people are amenable to fixing the bug I found in their code.

PR #592, opened 2021-03-10
Fix scNoCap error criterion in format preview

LegionMammal978:
This change downgrades the scNoCap error into a warning when the text within the tags begins with an asterisk. This allows for cases where a word typed in small caps is fragmented over multiple pages or footnotes. I stumbled across this while formatting the other day, and I figured it would be a relatively trivial fix.

### DIFF
--- a/tools/proofers/preview.js
+++ b/tools/proofers/preview.js
@@ -384,12 +384,12 @@ $(function () {
                     continue;
                 }
                 res1 = result[1];
-                if (res1 === res1.toLowerCase()) { //no upper case found - definite
+                if (res1 === res1.toLowerCase() && res1.charAt(0) !== "*") { // no upper case found - definite
                     reportIssue(result.index, 4, "scNoCap", 1);
                     continue;
                 }
                 res1 = removeComments(res1);
-                if (res1 === res1.toLowerCase()) { //upper case was in a note - poss
+                if (res1 === res1.toLowerCase()) { // either a lowercase fragment, or upper case was in a note - poss
                 // mark only a text char incase no tags shown
                     reportIssue(result.index + 4, 1, "scNoCap", 0);
                 }
--- a/SETUP/tests/jsTests/formatPreviewTests.js
+++ b/SETUP/tests/jsTests/formatPreviewTests.js
@@ -257,6 +257,12 @@ QUnit.module("Format preview test", function() {
         issueTest(assert, 0, 0, 4, "scNoCap", 1);
     });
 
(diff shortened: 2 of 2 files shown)


tools/proofers/preview.js
```diff
@@ -384,12 +384,12 @@ $(function () {
                     continue;
                 }
                 res1 = result[1];
-                if (res1 === res1.toLowerCase()) { //no upper case found - definite
+                if (res1 === res1.toLowerCase() && res1.charAt(0) !== "*") { // no upper case found - definite
```
chrismiceli, line 387: I believe we have some unit tests for this area. Could you add one for this case perhaps: https://github.com/DistributedProofreaders/dproofreaders/blob/master/SETUP/tests/jsTests/formatPreviewTests.js

LegionMammal978: I have added a relevant unit test.

chrismiceli (approved):
Looks good to me. I would feel more comfortable with @70ray's review though.

srjfoo:
> Looks good to me. I would feel more comfortable with @70ray's review though.

Ditto.

For those who want to test, it's available in the [sc-nocap](https://www.pgdp.org/~srjfoo/c.branch/sc-nocap/) sandbox.

srjfoo (approved):

tools/proofers/preview.js
```diff
@@ -384,12 +384,12 @@ $(function () {
...
                     reportIssue(result.index, 4, "scNoCap", 1);
                     continue;
                 }
                 res1 = removeComments(res1);
-                if (res1 === res1.toLowerCase()) { //upper case was in a note - poss
+                if (res1 === res1.toLowerCase()) { // either a lowercase fragment, or upper case was in a note - poss
```
70ray, line 392: It makes the logic slightly confusing to put both cases in here. It could be better to make a test for the first character in another condition, but this would make it more long-winded.

70ray (commented):
I should mention that "I-am-back" sent me a PM regarding the same issue some time ago but doing anything about it had slipped to the bottom of my list.
He mentioned a more general case of a phrase such as "<sc>Hurley burley</sc>" being split over two pages and suggested that we could check if the page starts with "<sc> or <sc>*". This would not catch the case of a split footnote mentioned by LegionMammal978 but it would be possible to check if it was a continuation footnote and if the text in the footnote starts with "<sc>".
Alternatively we could downgrade all cases of no capitals in smallcaps to "possible issue".

70ray (approved):

cpeel:
This will be rolled out to pgdp.net in our April deployment.

PR #593, opened 2021-03-11
Increase minimum middleware versions

cpeel:
This resets sets our minimum middleware versions to what is currently running on PROD. These are also the middleware versions that are still getting security updates.


pinc/misc.inc
```diff
@@ -581,9 +581,12 @@ function endswith( $subject, $suffix )
     return ( substr( $subject, -strlen($suffix) ) == $suffix );
 }
 
-function str_contains( $haystack, $needle )
-{
-    return ( strpos( $haystack, $needle ) !== FALSE );
+if(!function_exists('str_contains')) {
+    // str_contains() now exists in PHP8
+    function str_contains( $haystack, $needle )
```
chrismiceli, line 586: this is available outside the if scope since if doesn't have scope in php, right?

chrismiceli (approved):

pinc/misc.inc
cpeel: Yes, PHP functions are always global regardless of where they are defined unless they're in classes.

srjfoo (approved):

70ray (approved):

PR #594, opened 2021-03-11
Use the DP API in the Page Browser

cpeel:
This is a perfunctory MR for tracking purposes - this code has already been reviewed and tested. Go ahead and approve it if we think this is ready to merge for the April deployment (I think it is).

I want to use this MR to capture the developer draft for the Recent Additions topic announcement of this new feature when it is deployed in April. @70ray, @chrismiceli please use Slack to collaborate on the draft with @srjfoo and post the final version in this MR.

[The MR may get merged before the draft is finished, but still include the draft when it is ready.]


srjfoo (approved):

70ray (approved):

srjfoo:
First draft:

> The WordCheck tools that Project Managers use are no longer using old-style frames, and are now HTML5-compliant.
>
> The Page Browser, including the WordCheck tools for PMs that show word context, now use the API to display images, and where possible, the associated text. This is an important milestone for the development of future proofreading interfaces.
>
> There may be some small style changes, but no functionality has been removed, and page display should be fater with less flickering.

chrismiceli (approved):
Draft announcement text looks good to me.

70ray:
> First draft:
> 
> > The WordCheck tools that Project Managers use are no longer using old-style frames, and are now HTML5-compliant.
> > The Page Browser, including the WordCheck tools for PMs that show word context, now uses the API to display images, and where possible, the associated text. This is an important milestone for the development of future proofreading interfaces.
> > There may be some small style changes, but no functionality has been removed, and page display should be fater with less flickering.

there is a typo: "should be faster"
Do we need to explain what API is? Perhaps something like "it enables the text to be fetched from the server without having to reload the whole page".

cpeel:
I don't think we need to explain what an API is (although we can field questions if it comes up). I'm not sure most people will actually care and those that do will know what it is.

srjfoo:
> > First draft:
...
> > > There may be some small style changes, but no functionality has been removed, and page display should be fater with less flickering.
> 
> there is a typo: "should be faster"

Typo fixed; thanks for spotting that!

PR #595, opened 2021-03-13
Update user_password field length in non_activated_users

cpeel:
phpBB increased the size of the maximum user_password so we need to increase ours in non_activated_users to match.

Note: This is actually a phpBB-hashed version of the password -- we do not store passwords in the clear.


srjfoo (approved):

chrismiceli (approved):
No UI checks we need to change?

cpeel:
> No UI checks we need to change?

No - we aren't actually changing the length of the passwords we accept, phpBB 3.3.3 appears to use a password hash that results in a larger string.

PR #597, opened 2021-03-16
Work around portable-utf8 bug by using escape sequences in regex

cpeel:
This addresses two different problems:
1. For some reason `grapheme_extract()` treats "\r\n" as a single grapheme but all of our code assumes they are two separate characters (which they are).
2. Work around the portable-utf8 bug which is not correctly converting some Unicode strings to characters by using escape values for regexs instead.

Testable in the [fix-grapheme-split](https://www.pgdp.org/~cpeel/c.branch/fix-grapheme-split/) sandbox.

### DIFF
--- a/SETUP/lint_charsuites.php
+++ b/SETUP/lint_charsuites.php
@@ -15,13 +15,21 @@ foreach(CharSuites::get_all() as $charsuite)
     echo "Validating charsuite $charsuite->name...\n";
 
     // Validate that codepoint specifiers don't include both a range and a
-    // combining character.
+    // combining character. And that all ranges are low to high.
     foreach($charsuite->codepoints as $codepoint)
     {
         if(stripos($codepoint, "-") !== FALSE && stripos($codepoint, ">") !== FALSE)
         {
             echo sprintf("ERROR: %s codepoint has both a range and a combining character\n", $codepoint);
             exit(1);
+        }
+
+        if (stripos($codepoint, "-") !== FALSE) {
+            list($start, $end) = explode('-', $codepoint);
+            if ($start > $end) {
+                echo sprintf("ERROR: %s codepoint range start is greater than range end\n", $codepoint);
+                exit(1);
+            }
         }
     }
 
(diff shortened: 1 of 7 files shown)


chrismiceli (approved):
It's all my fault!

cpeel:
This fixes new lines / carriage returns by my suspicion is that this is just masking some other problem. I want to do more testing because I suspect this isn't really the fix that we want.

srjfoo (approved):

srjfoo:
> This fixes new lines / carriage returns by my suspicion is that this is just masking some other problem. I want to do more testing because I suspect this isn't really the fix that we want.

I went back and tested project loads again on the main test site, and realized that the disallowed character I was seeing on project load was `U+00ad`, not `U+000d`, so I wasn't actually seeing problems on project load, just on page saves.

Do you have suggestions for additional testing? (In addition to page saves and project loads)

cpeel:
Some initial "type random ASCII into a page" testing showed we are doing something horrible wrong with numbers as well, so this is much deeper than just newline+carriage returns. I need to spend more time on it later.

cpeel:
This has been updated to work around the portable-utf8 bug. We only need portable-utf8 to escape the broken characters when building regexs, so let's just use Unicode escape sequences for the regexs instead.

This impacts the JS validation in the UI as well as on the backend. We should test Astral plane characters as well.

70ray (approved):
That seems a good way of doing it.

70ray (approved):
We can re-instate the commented-out test for astral characters in characterValidation.js now that you have removed the IE hack.

cpeel:
> We can re-instate the commented-out test for astral characters in characterValidation.js now that you have removed the IE hack.

Thanks for the reminder - I've uncommented that test.

chrismiceli (approved):

srjfoo (approved):

cpeel:
There was some original scope creep on making this work for Astral characters but that was rolled back and this MR focused on achieving parity with what is on PROD now. We'll follow up with full Astral character support in a separate MR.

PR #599, opened 2021-03-20
set samesite cookies

chrismiceli:
Set all cookies to SameSite=Lax which is the default on newer browsers when not specified to None anyways. Set CSRFtoken cookie to SameSite=Strict as it is only sent on form POST/PUT which cannot be done from 3rd party context, even top level navigation.

Testable in sandbox here:
https://www.pgdp.org/~chrismiceli/c.branch/samesite

Scenarios to verify:

1. Login
2. Logout
3. Saving user preferences
4. Navigate to page with ?lang=<lang> and verify language cookie is set and respected throughout application as usual
5. Registering a new proofer. Also verify that navigating to the activate link works from another domain, such as an email web page like gmail.
6. Forums and wiki still accessible

You can read more about samesite cookie properties here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite

This should get rid of warnings you can see in the firefox console in pgdp.net like
```
Some cookies are misusing the recommended “SameSite“ attribute 2
Cookie “DP_Session” will be soon rejected because it has the “SameSite” attribute set to “None” or an invalid value, without the “secure” attribute. To know more about the “SameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite
```

### DIFF
--- a/SETUP/Makefile
+++ b/SETUP/Makefile
@@ -22,7 +22,7 @@ lint_code:
 	$(SELF_DIR)lint_php_files.sh $(SELF_DIR)..
 	$(SELF_DIR)lint_json_files.sh $(SELF_DIR)..
 
-lint: lint_charsuites lint_code
+lint: lint_charsuites lint_css lint_code
 
 #----------------------------------------------------------------------------
 # CSS compilation
--- a/SETUP/Makefile
+++ b/SETUP/Makefile
@@ -22,7 +22,7 @@ lint_code:
 	$(SELF_DIR)lint_php_files.sh $(SELF_DIR)..
 	$(SELF_DIR)lint_json_files.sh $(SELF_DIR)..
 
-lint: lint_charsuites lint_less lint_code
+lint: lint_charsuites lint_code
 
 #----------------------------------------------------------------------------
 # CSS compilation
--- a/pinc/User.inc
+++ b/pinc/User.inc
@@ -76,7 +76,7 @@ class User
                 // now set the profile cookie if this is for the current user
                 if($this->username == User::current_username())
                 {
-                    setcookie("profile", $this->current_profile->id);
+                    dp_setcookie("profile", $this->current_profile->id);
                 }
                 break;
             default:
@@ -119,7 +119,7 @@ class User
                         // OK if we can't.
                         if(!header_sent())
                         {
-                            setcookie("profile");
+                            dp_setcookie("profile");
                         }
 
                         // fall through to getting the profile from the user record
--- a/pinc/bootstrap.inc
+++ b/pinc/bootstrap.inc
@@ -73,4 +73,18 @@ function dp_class_autoloader($class)
     }
 }
 
+function dp_setcookie($name, $value, $expires=0, $options=[]) {
+    global $use_secure_cookie;
+
+    $defaults = [
+        "expires" => $expires,
+        "path" => "/",
+        "secure" => $use_secure_cookie,
+        "samesite" => "Lax",
+    ];
+
+    $options = array_merge($defaults, $options);
+    return setcookie($name, $value, $options);    
+}
+
 // vim: sw=4 ts=4 expandtab
--- a/pinc/bootstrap.inc
+++ b/pinc/bootstrap.inc
@@ -73,7 +73,7 @@ function dp_class_autoloader($class)
     }
 }
 
-function dp_setcookie($name, $value, $expires=0, $options=[]) {
+function dp_setcookie($name, $value="", $expires=0, $options=[]) {
     global $use_secure_cookies;
 
     $defaults = [
--- a/pinc/bootstrap.inc
+++ b/pinc/bootstrap.inc
@@ -74,12 +74,12 @@ function dp_class_autoloader($class)
 }
 
 function dp_setcookie($name, $value, $expires=0, $options=[]) {
-    global $use_secure_cookie;
+    global $use_secure_cookies;
 
     $defaults = [
         "expires" => $expires,
         "path" => "/",
-        "secure" => $use_secure_cookie,
+        "secure" => $use_secure_cookies,
         "samesite" => "Lax",
     ];
 
--- a/pinc/gettext_setup.inc
+++ b/pinc/gettext_setup.inc
@@ -33,13 +33,7 @@ function get_desired_language()
     if (isset($_GET['lang']) && lang_code($_GET['lang'])) {
         // User requested a specific language for this page
         $locale = lang_code($_GET['lang']);
-        setcookie("language", $locale, [
-          "expires" => time()+31536000,
-          "path" => "/",
-          "domain" => "",
-          "secure" => $use_secure_cookies,
-          "samesite" => "Lax"
-        ]);
+        dp_setcookie("language", $locale, time()+31536000);
     } else if ($user && $user->u_intlang){
         // User is logged in and has set their language preference
         $locale = $user->u_intlang;
--- a/pinc/gettext_setup.inc
+++ b/pinc/gettext_setup.inc
@@ -20,8 +20,6 @@ include_once($relPath.'languages.inc');
 // * If everything else fails, default to en_US.
 function get_desired_language()
 {
-    global $use_secure_cookies;
-
     // This function maybe called multiple times within a page load.
     // Only do the calculations to determine the locale once.
     static $locale = NULL;
--- a/pinc/gettext_setup.inc
+++ b/pinc/gettext_setup.inc
@@ -33,7 +33,13 @@ function get_desired_language()
     if (isset($_GET['lang']) && lang_code($_GET['lang'])) {
         // User requested a specific language for this page
         $locale = lang_code($_GET['lang']);
-        setcookie("language", $locale, time()+31536000, "/", "", $use_secure_cookies);
+        setcookie("language", $locale, [
+          "expires" => time()+31536000,
+          "path" => "/",
+          "domain" => "",
+          "secure" => $use_secure_cookies,
+          "samesite" => "Lax"
+        ]);
     } else if ($user && $user->u_intlang){
         // User is logged in and has set their language preference
         $locale = $user->u_intlang;
--- a/pinc/misc.inc
+++ b/pinc/misc.inc
@@ -1183,11 +1183,7 @@ function set_csrf_token($force_set = TRUE)
         return;
 
     $token = bin2hex(random_bytes(32));
-    setcookie("CSRFtoken", $token, [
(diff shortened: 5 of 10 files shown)


SETUP/Makefile
```diff
@@ -22,7 +22,7 @@ lint_code:
 	$(SELF_DIR)lint_php_files.sh $(SELF_DIR)..
 	$(SELF_DIR)lint_json_files.sh $(SELF_DIR)..
 
-lint: lint_charsuites lint_less lint_code
```
chrismiceli, line 25: lint_less is not a target so make complained when I ran ```make lint```. The github automation call the targets individually so this doesn't show up in checks.

tools/setlangcookie.php
```diff
@@ -12,7 +12,12 @@
 if($language)
 {
     // set the cookie
-    setcookie("language", $language, time()+31536000, "/");
+    setcookie("language", $language, [
```
chrismiceli, line 15: I am not sure of this scenario, but it looks like you can set the language with a "lang" query param which results in a cookie being set.

SETUP/Makefile
cpeel: Oops. This should be `lint_css` instead though.

cpeel:
Thanks for both finding this and putting up the code! I appreciate you getting ahead of future FF changes.

Instead of smearing these across all of the uses of `setcookie()` (and you missed a few based on my grepping), consider creating a `dp_setcookie($name, $value, $expires=0, $options=[])` function in `pinc/bootstrap.inc` that will provide defaults that the caller can optionally override and use it across the entire codebase. This would also centralize our use of `$use_secure_cookie` as well. Something like this (totally untested):
```php
function dp_setcookie($name, $value, $expires=0, $options=[]) {
    global $use_secure_cookie;

    $defaults = [
        "expires" => $expires,
        "path" => "/",
        "secure" => $use_secure_cookie,
        "samesite" => "Lax",
    ];

    $options = array_merge($defaults, $options);
    return setcookie($name, $value, $options);    
}
```

For now, I would let `domain` just default to `""` and rip out the bits in `pinc/dpsession_via_cookies.inc` that checks for it (that code should probably be removed entirely -- I have no idea if it works).

Let's be sure and update all of the files that use `setcookie()`:
```
accounts/addproofer.php
pinc/User.inc
pinc/dpsession.inc
pinc/dpsession_via_cookies.inc
pinc/dpsession_via_php_sessions.inc
pinc/forum_interface.inc
pinc/gettext_setup.inc
pinc/misc.inc
tools/setlangcookie.php
```

cpeel:
> ... and rip out the bits in `pinc/dpsession_via_cookies.inc` that checks for it (that code should probably be removed entirely -- I have no idea if it works).

On second thought, let's leave the code in `pinc/dpsession_via_cookies.inc` that sets `COOKIE_DOMAIN` and just pass that into the new `dp_setcookie()` function for domain. Less code code churn I think.

pinc/gettext_setup.inc
```diff
@@ -33,7 +33,7 @@ function get_desired_language()
     if (isset($_GET['lang']) && lang_code($_GET['lang'])) {
         // User requested a specific language for this page
         $locale = lang_code($_GET['lang']);
-        setcookie("language", $locale, time()+31536000, "/", "", $use_secure_cookies);
+        dp_setcookie("language", $locale, time()+31536000);
```
cpeel, line 36: Don't forget to remove the `global $use_secure_cookies` on line 23.

pinc/misc.inc
```diff
@@ -1183,15 +1183,18 @@ function set_csrf_token($force_set = TRUE)
         return;
 
     $token = bin2hex(random_bytes(32));
-    setcookie("CSRFtoken", $token, time() + 60 * 60 * 24, '/', '', $use_secure_cookies, TRUE);
+    dp_setcookie("CSRFtoken", $token, time() + 60 * 60 * 24, [
```
cpeel, line 1186: Ditto removing `global $use_secure_cookies;`.

pinc/misc.inc
```diff
@@ -1183,15 +1183,18 @@ function set_csrf_token($force_set = TRUE)
...
+      "httponly" => TRUE,
+      "samesite" => "Strict",
+    ]);
     $_COOKIE["CSRFtoken"] = $token;
 }
 
 function destroy_csrf_token()
 {
     global $use_secure_cookies;
```
cpeel, line 1195: Here too.

pinc/bootstrap.inc
```diff
@@ -73,4 +73,18 @@ function dp_class_autoloader($class)
     }
 }
 
+function dp_setcookie($name, $value, $expires=0, $options=[]) {
+    global $use_secure_cookie;
```
cpeel, line 77: Whoops, this is suppose to be `$use_secure_cookies` (ditto below). My bad.

cpeel (commented):
Make sure that the only places `$use_secure_cookies` are used after your changes are `pinc/site_vars.php.template` and `pinc/bootstrap.inc`.

pinc/gettext_setup.inc
chrismiceli: This is what happens when you rush a PR while your child takes a nap :).

chrismiceli:
> `use_secure_cookies`

Looks like we are good from my evaluation.

pinc/User.inc
```diff
@@ -119,7 +119,7 @@ class User
                         // OK if we can't.
                         if(!header_sent())
                         {
-                            setcookie("profile");
+                            dp_setcookie("profile");
```
cpeel, line 122: This is going to fail because the `dp_setcookie()` definition requires a value. I guess in that function declaration we should default `$value=''`.

cpeel (commented):
My one comment notwithstanding, this looks and works great!

Thank you for helping making the site more secure and supported on modern browsers!

srjfoo:
> Scenarios to verify:
> 
> 1. Login
> 
> 1. Logout
> 
> 1. Saving user preferences
> 
> 1. Navigate to page with ?lang= and verify language cookie is set and respected throughout application as usual
> 
> 1. Registering a new proofer. Also verify that navigating to the activate link works from another domain, such as an email web page like gmail.
> 
> 1. Forums and wiki still accessible

I've verified everything except "Navigate to page with ?lang= ...", so it's time for me to learn something new. I tried setting the site to French while logged out, but that didn't add `?lang=fr` to the URL, so I obviously don't understand what I'm supposed to be doing. Also, how do I verify that the language cookie is set.

> This should get rid of warnings you can see in the firefox console in pgdp.net like
> 
> `Some cookies are misusing the recommended “SameSite“ attribute 2`
>
> `Cookie “DP_Session” will be soon rejected because it has the “SameSite” attribute set to “None” or an invalid value, without the “secure” attribute. To know more about the “SameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite`

And another thing to learn -- I poked at the console a little bit, but did not see that warning. But I plead ignorance on how to use the full extent of the console's capabilities, so I may be missing something.

srjfoo:
> I've verified everything except "Navigate to page with ?lang= ...", so it's time for me to learn something new. I tried setting the site to French while logged out, but that didn't add `?lang=fr` to the URL, so I obviously don't understand what I'm supposed to be doing. Also, how do I verify that the language cookie is set.

To clarify, when I added `?lang=fr` to `default.php`, I did not get the site in French.

cpeel:
> To clarify, when I added `?lang=fr` to `default.php`, I did not get the site in French.

`lang` is misnamed, it's actually a locale: https://www.pgdp.org/c/default.php?lang=fr_FR

srjfoo:
> > To clarify, when I added `?lang=fr` to `default.php`, I did not get the site in French.
> 
> `lang` is misnamed, it's actually a locale: https://www.pgdp.org/c/default.php?lang=fr_FR

That did it, thanks! Didn't even occur to me last night to try that. I'd still like to know how to verify that the cookie is set.

Also, referring back to:

> Navigate to page with ?lang= and verify language cookie is set and respected throughout application as usual

Initially, I interpreted that as meaning that as I navigated away from the page where I manually set `?lang=fr_FR`, I should continue to see the site in French. That is not the case, so I'm thinking I misinterpreted. So I need clarification on "and respected throughout application as usual".

srjfoo:
> Initially, I interpreted that as meaning that as I navigated away from the page where I manually set `?lang=fr_FR`, I should continue to see the site in French. That is not the case, so I'm thinking I misinterpreted. So I need clarification on "and respected throughout application as usual".

After thinking about it, I just went through multiple pages adding `?lang=fr_FR` or `&lang=fr_FR` (whichever was appropriate). Round and pool pages look good; project pages look good, My Projects looks good, and the Page Browser, also.

Proofreading interface doesn't seem to handle it.

chrismiceli:
> And another thing to learn -- I poked at the console a little bit, but did not see that warning. But I plead ignorance on how to use the full extent of the console's capabilities, so I may be missing something.

Weird, I can only sometimes get the message to show. Perhaps it is due to my privacy settings. If I open pgdp.net in an in-private window, I see the message:
image

srjfoo:
> > And another thing to learn -- I poked at the console a little bit, but did not see that warning. But I plead ignorance on how to use the full extent of the console's capabilities, so I may be missing something.
> 
> Weird, I can only sometimes get the message to show. Perhaps it is due to my privacy settings. If I open pgdp.net in an in-private window, ...

Private window allowed me to see the message (only before logging in, and even when your sandbox was specified. I didn't try it for PROD). Thanks!

cpeel:
> Initially, I interpreted that as meaning that as I navigated away from the page where I manually set `?lang=fr_FR`, I should continue to see the site in French. That is not the case, so I'm thinking I misinterpreted.

That should work, but only if you're not logged in. After you are logged in your user preferences take precedence.

srjfoo:
> > Initially, I interpreted that as meaning that as I navigated away from the page where I manually set `?lang=fr_FR`, I should continue to see the site in French. That is not the case, so I'm thinking I misinterpreted.
> 
> That should work, but only if you're not logged in. After you are logged in your user preferences take precedence.

Ahah! That was it. And yes, the language change was retained across all pages I was able to link to while logged out.

chrismiceli:
Sounds like there aren't any more changes requested from me, right?

srjfoo:
> Sounds like there aren't any more changes requested from me, right?

I haven't verified tracking cookies (I need instruction), but other than that, I've poked at everything you suggested and it looks good.

srjfoo (approved):

PR #601, opened 2021-03-21
1488: Clickable words in Good Words Candidate List

chrismiceli:
surround input with label to ease clicking

 https://www.pgdp.org/~chrismiceli/c.branch/issue_1488

### DIFF
--- a/tools/project_manager/word_freq_table.inc
+++ b/tools/project_manager/word_freq_table.inc
@@ -198,7 +198,7 @@ function printTableFrequencies($initialFreq,$cutoffOptions,$wordCount,$instance,
         $line = "<tr>";
 
         if($checkbox) {
-            $line .= "<td class='mono' style='white-space: nowrap;'>" . $checkbox[$word] . " " . rtrim($word) . "</td>";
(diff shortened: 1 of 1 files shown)


chrismiceli:
The clickable zone now includes the words
image

srjfoo (approved):

70ray (approved):

cpeel:
If we want to announce this change (I'm on the fence) please include an announcement in this MR.

srjfoo:
> If we want to announce this change (I'm on the fence) please include an announcement in this MR.

I wouldn't give it its own announcement, I don't think. It was an old task, and a small change. If we have a collection of small stuff we might add it? (I'm willing to be overruled if anyone has strong feelings about it.

chrismiceli:
> 
> 
> > If we want to announce this change (I'm on the fence) please include an announcement in this MR.
> 
> I wouldn't give it its own announcement, I don't think. It was an old task, and a small change. If we have a collection of small stuff we might add it? (I'm willing to be overruled if anyone has strong feelings about it.

Works for me.

PR #603, opened 2021-03-21
Properly escape language regex filters

cpeel:
Properly escape language regex filters. This not only filters the regex special characters but the values themselves. This was found in `php_errors` from the every-two-weeks Detectify run.

Testable in [escape-lang-regex](https://www.pgdp.org/~cpeel/c.branch/escape-lang-regex/) sandbox.


srjfoo (approved):
Once I remembered that the round page filters have different configs, everything worked fine 😁

chrismiceli (approved):

PR #605, opened 2021-03-28
Update descriptive text for Bronze list

srjfoo:
The primary goal of this MR was to update the descriptive text for the Bronze list so that it matches what's actually displayed. See [Task 1163](https://www.pgdp.net/c/tasks.php?action=show&task_id=1163). To more correctly describe each star metal, the common boilerplate paragraph was removed, and similar text was customized for each one.

The original task and comments suggested changing the text as a possible option. Possibly in the future, the other suggestion will be implemented, but meanwhile, the text will at least correctly describe what's displayed.

Also added a maximum number of items to display per page.
Testable: [~srjfoo/c.branch/update-bronze-blurb](https://www.pgdp.org/~srjfoo/c.branch/update-bronze-blurb/list_etexts.php?x=b&sort=5).

### DIFF
--- a/list_etexts.php
+++ b/list_etexts.php
@@ -9,7 +9,7 @@ include_once($relPath.'misc.inc'); // get_enumerated_param(), get_integer_param(
 
 $x        = get_enumerated_param($_GET, 'x', 'g', array('g', 's', 'b'));
 $sort     = get_integer_param($_GET, 'sort',      0, 0, 5);
-$per_page = get_integer_param($_GET, 'per_page', 20, 1, NULL);
+$per_page = get_integer_param($_GET, 'per_page', 20, 1, 100);
 $offset   = get_integer_param($_GET, 'offset',    0, 0, NULL);
 
 if($x == "g") {
--- a/list_etexts.php
+++ b/list_etexts.php
@@ -4,32 +4,41 @@ include_once($relPath.'base.inc');
 include_once($relPath.'theme.inc');
 include_once($relPath.'project_states.inc');
 include_once($relPath.'list_projects.inc');
-include_once($relPath.'misc.inc'); // get_enumerated_param(), get_integer_param()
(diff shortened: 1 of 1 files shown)


chrismiceli (approved):
Change looks good to me technically.

list_etexts.php
```diff
@@ -8,28 +8,26 @@
...
-
 if($x == "g") {
     $type = "gold";
     $title = _("Completed Gold E-Texts");
     $state = SQL_CONDITION_GOLD;
-    $info = _("Below is the list of Gold e-texts that have been produced on this site. Gold e-texts are books that have passed through all phases of proofreading, formatting, and post-processing. They have been submitted to Project Gutenberg and are now available for your enjoyment and download.");
+    $info = _("Below is the list of Gold e-texts that have been produced on this site. Gold e-texts are books that have passed through all phases of proofreading, formatting, and post-processing. They have been submitted to Project Gutenberg and are now available for your enjoyment and download.") . "<p></p>" . _("These e-texts are the product of hundreds of hours of labor donated by all of our volunteers. The list is sorted with the most recently submitted e-texts at the top. You can sort them based upon your own preferences by clicking below. Enjoy!!");
```
cpeel, line 18: It's not at all intuitive why we place `<p></p>` tags between the sentences (actually, it should be `</p><p>` but I digress).

It would be better if we made `$info` into an array that we surround with paragraph tags when we output them. This would also allow us to put the two sentences on separate rows for shorter lines.

e.g.
```php
$info = [
    _("Para 1"),
    _("Para 2"),
];
```

Then when we output them using this convenience function in `misc.inc`:
```php
echo surround_and_join($info, "<p>", "</p>", "");
```

srjfoo: Thanks, I'll do that. And ... that `<p></p>` was an error on my part. I thought I'd carefully done `</p><p>` for all instances.

srjfoo: And done.

PR #610, opened 2021-04-02
d3.js implementation of cumulative_total_proj_summary_graph

chrismiceli:
TEST has some data to make the chart render:
https://www.pgdp.org/~chrismiceli/c.branch/graphs/stats/stats_central.php
image


chrismiceli:
We plan on merging this into a branch, not master. I'll defer to @cpeel to create that branch if we decide to merge these changes.

srjfoo:
I really like the way it looks on the Charcoal theme. but realize now that the background color for the graph container has always been the sidebar color for the theme -- all this time, I never noticed that the background color was hidden by the graphs. See [the corresponding graph on PROD](https://www.pgdp.net/c/stats/stats_central.php).

I think we still need a legend, don't we? And some notion of exactly what the date means -- is that 2008 January 1 of 2008, or is that just an arbitrary point some time in that year. The other thing missing is the timestamp for when the graph was generated.

cpeel:
> The other thing missing is the timestamp for when the graph was generated.

The only reason that date is there is because the current graphs are generated and cached for some period of time. Since these graphs are going to be live I wouldn't put a date on them.

srjfoo:
> > The other thing missing is the timestamp for when the graph was generated.
> 
> The only reason that date is there is because the current graphs are generated and cached for some period of time. Since these graphs are going to be live I wouldn't put a date on them.

Thanks -- that hadn't sunk in. I agree -- if the graphs are on live data, we don't need the date and time the graph was generated.

chrismiceli:
> 
> 
> I really like the way it looks on the Charcoal theme. but realize now that the background color for the graph container has always been the sidebar color for the theme -- all this time, I never noticed that the background color was hidden by the graphs. See [the corresponding graph on PROD](https://www.pgdp.net/c/stats/stats_central.php).
> 
> I think we still need a legend, don't we? And some notion of exactly what the date means -- is that 2008 January 1 of 2008, or is that just an arbitrary point some time in that year. The other thing missing is the timestamp for when the graph was generated.

I'm working on the legend now as well as moving the yaxis to the right hand side. As for the dates, I believe it is January 1st, but let me see if I can configure it more.

chrismiceli:
Updated screenshot with right y axis, date in x axis, and legend:
image

srjfoo:
I like the way it looks! (Though maybe shift the legend down a little bit -- the line spacing looks wrong between the top of the legend and the title of the graph.

A few things we should probably discuss.

The graph doesn't resize for narrow screens. Take a look at it on a phone.

In the past (and currently on PROD), the background color of the graph container has not mattered. In fact, I don't think I ever realized until you started doing this @chrismiceli, that the background color of the containers looks like it's the same as the sidebar color. On a light-colored theme, it seemed fine for the background of the graphs to be white. Do we want to consider changing the background color of the containers to be the default page background (i.e. white for the light themes, and a different color of grey for Charcoal), or leave it as-is? Here's a composite screenshot for reference.

image

And, finally, a question -- there is no border to the graph other than the axes left and bottom. That doesn't bother me -- should it be something we worry about?

chrismiceli:
> 
> 
> I like the way it looks! (Though maybe shift the legend down a little bit -- the line spacing looks wrong between the top of the legend and the title of the graph.
> 
> A few things we should probably discuss.
> 
> The graph doesn't resize for narrow screens. Take a look at it on a phone.
> 
> In the past (and currently on PROD), the background color of the graph container has not mattered. In fact, I don't think I ever realized until you started doing this @chrismiceli, that the background color of the containers looks like it's the same as the sidebar color. On a light-colored theme, it seemed fine for the background of the graphs to be white. Do we want to consider changing the background color of the containers to be the default page background (i.e. white for the light themes, and a different color of grey for Charcoal), or leave it as-is? Here's a composite screenshot for reference.
> 
> image
> 
> And, finally, a question -- there is no border to the graph other than the axes left and bottom. That doesn't bother me -- should it be something we worry about?

I've pushed a change for the narrow view and legend. I like the idea about the background color, just not sure how best to implement it since we're just using a themed table that we use in many other places. I can override those styles, but then there is some stray colors I am not sure what to do with. See below for an example:
image

srjfoo:
> I've pushed a change for the narrow view and legend. I like the idea about the background color, just not sure how best to implement it since we're just using a themed table that we use in many other places. I can override those styles, but then there is some stray colors I am not sure what to do with. (snipped screenshot)

Thanks -- it's resizing on the phone now, and that and the position of the legend looks good.

See the Table formats at the bottom of the [Style Demos](https://www.pgdp.org/c/styles/style_demo.php) page. Currently the table style used for the graphs is the "Themed table". If we want to continue displaying graphs inside tables, and want the themed color for the \<th\> element and a border, but don't want the themed background color, we may need to define a new table style.

I haven't studied the graphs, and don't really have an opinion, but -- they're not really tabular data, are they? And I don't know the answer to these questions, but thought I'd toss them out -- is there an overarching reason for keeping them in tables? If we do keep them in tables, is there a reason for the table header to repeat the graph title, since the title is shown on the graph, itself?

There may be other questions we should be asking that I haven't yet thought of.

Ignoring the question of whether or not the graphs should be being displayed inside tables, and addressing the questions you asked in your marked-up screenshot above, if I'm understanding the graphs correctly, the background of the graph is transparent? Which means that it will take on the color of the background it's up against. The places you have circled on your screenshot would take on the background color of the table rows in the table (under the current setup), just as they do now.

If we create a new table style class where the header row is the themed color and the rows take on the default page background, the circled areas will be that color, too -- the same color as the background that shows through the transparent graph background.

Both your circled areas are the same color, because they're separate themed tables. It doesn't show in the graphs currently being used because the graphs have a white background. If you change the table class, the background will change.

chrismiceli:
> 
> 
> > I've pushed a change for the narrow view and legend. I like the idea about the background color, just not sure how best to implement it since we're just using a themed table that we use in many other places. I can override those styles, but then there is some stray colors I am not sure what to do with. (snipped screenshot)
> 
> Thanks -- it's resizing on the phone now, and that and the position of the legend looks good.
> 
> See the Table formats at the bottom of the [Style Demos](https://www.pgdp.org/c/styles/style_demo.php) page. Currently the table style used for the graphs is the "Themed table". If we want to continue displaying graphs inside tables, and want the themed color for the <th> element and a border, but don't want the themed background color, we may need to define a new table style.
> 
> I haven't studied the graphs, and don't really have an opinion, but -- they're not really tabular data, are they? And I don't know the answer to these questions, but thought I'd toss them out -- is there an overarching reason for keeping them in tables? If we do keep them in tables, is there a reason for the table header to repeat the graph title, since the title is shown on the graph, itself?
> 
> There may be other questions we should be asking that I haven't yet thought of.
> 
> Ignoring the question of whether or not the graphs should be being displayed inside tables, and addressing the questions you asked in your marked-up screenshot above, if I'm understanding the graphs correctly, the background of the graph is transparent? Which means that it will take on the color of the background it's up against. The places you have circled on your screenshot would take on the background color of the table rows in the table (under the current setup), just as they do now.
> 
> If we create a new table style class where the header row is the themed color and the rows take on the default page background, the circled areas will be that color, too -- the same color as the background that shows through the transparent graph background.
> 
> Both your circled areas are the same color, because they're separate themed tables. It doesn't show in the graphs currently being used because the graphs have a white background. If you change the table class, the background will change.

You are right, currently the background is transparent. We can make it any other color too though, independent of table background coloring. It may look weird though if it doesn't take the full width of the table (as my example screenshot above). I too don't see why these are tables at all and am open to pulling them out of the table if we feel so inclined.

cpeel:
I've created the `svg-graph` branch that we can do this iterative development in.

chrismiceli:
> 
> 
> I've created the `svg-graph` branch that we can do this iterative development in.

Ok, rebased against svg-graphs and resolved conflicts due to format. I'll begin in earnest chart migration with your go ahead @cpeel

stats/cumulative_total_proj_summary_graph.inc
```diff
@@ -0,0 +1,42 @@
+<?php
```
cpeel, line 1: Please move this file to `pinc/graph_data.inc` or some other similarly-named, more-generic file we can use for more graph data functions.

stats/stats_central.php
```diff
@@ -6,11 +6,22 @@
...
+include_once('./cumulative_total_proj_summary_graph.inc');
 
 require_login();
 
 $title = _("Statistics Central");
-output_header($title);
+
+$cumulative_total_proj_summary_graph_data = cumulative_total_proj_summary_graph();
+output_header($title, SHOW_STATSBAR, [
+    "js_files" => [
```
cpeel, line 17: Let's create a `get_graph_js_files()` function in the now-common PHP include file that returns these two JS files. That way we have a central place we reference the D3 files and return various chart-type JS files. This means we need to update the `pinc/3rdpart/d3/README.md` on the filename to update if it changes.

stats/cumulative_total_proj_summary_graph.inc
```diff
@@ -0,0 +1,42 @@
...
+
+include_once($relPath.'base.inc');
```
cpeel, line 3: Only `.php` files need to include `base.inc`.

cpeel (commented):
Some minor structural updates to this MR then we will get this merged into the new `svg-graphs` branch!

chrismiceli:
I applied the changes suggested by @cpeel and merged in master to the svg branch is up to date.

srjfoo:
> I applied the changes suggested by @cpeel and merged in master to the svg branch is up to date. Sandbox also updated:
> https://www.pgdp.org/~chrismiceli/c.branch/graphs/stats/stats_central.php

Just confirming that there should be no functional changes?

srjfoo:
When poking around just now, I realized something -- it looks like the only graph that's inside a table is the one on the `stats_central.php`. I didn't look at all of the graphs, but the ones in news items certainly aren't within a table, nor are the others that I poked at. Does it sound reasonable to remove that graph from a table?

cpeel:
> When poking around just now, I realized something -- it looks like the only graph that's inside a table is the one on the `stats_central.php`. I didn't look at all of the graphs, but the ones in news items certainly aren't within a table, nor are the others that I poked at. Does it sound reasonable to remove that graph from a table?

We should leave that one in the table to match the rest of the (way over-tabled) page for now.

chrismiceli:
> should

Correct, no functional changes.

PR #611, opened 2021-04-03
Revert "change href to #"

70ray:
This reverts commit 0c1d3b32f1cfe0942915b9f26d1a8195b08013f7.
'#' allows 'open in new tab' which opens the same page instead of the intended page in the new tab.
With the javascript void version, the options for opening in a new tab or window are not shown.
Perhaps a better fix would be to use an actual button but style it to look like a link in order to avoid changing the user experience.

published to https://www.pgdp.org/~rp31/c.branch/new_tab_fix

### DIFF
--- a/tools/project_manager/show_good_word_suggestions_detail.php
+++ b/tools/project_manager/show_good_word_suggestions_detail.php
@@ -80,7 +80,7 @@ echo "<a href='show_word_context.php?projectid=$projectid&amp;word=$encWord'>" .
       _("Show full context set for this word") . "</a>";
 
 echo " | ";
-echo "<a href='#' id='h_v_switch'></a>";
+echo "<a href='javascript:void(0)' id='h_v_switch'></a>";
 echo "</p>";
 
 foreach($word_suggestions as $suggestion) {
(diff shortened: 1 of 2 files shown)


chrismiceli:
I would even consider just styling it like a button and maybe add a link to pagebrowser from the detailframe iframe? That way if they want a new tab, they can do it still, but not from this link.

srjfoo:
> I would even consider just styling it like a button and maybe add a link to pagebrowser from the detailframe iframe? That way if they want a new tab, they can do it still, but not from this link.

Ray tested that recently, and I'm afraid I was the loudest voice against the change. It's not that I dislike buttons, but the change to a button made the link blend into the background way too much. Showing it as a link keeps the visibility as it is now. (Note that I'm also not against changing the UI if it improves the UX. This wouldn't have.)

chrismiceli:
> 
> 
> > I would even consider just styling it like a button and maybe add a link to pagebrowser from the detailframe iframe? That way if they want a new tab, they can do it still, but not from this link.
> 
> Ray tested that recently, and I'm afraid I was the loudest voice against the change. It's not that I dislike buttons, but the change to a button made the link blend into the background way too much. Showing it as a link keeps the visibility as it is now. (Note that I'm also not against changing the UI if it improves the UX. This wouldn't have.)

Then this I suppose is the best solution without redoing the design too much. Though having a link to the page browser on the detailframe wouldn't hurt.

chrismiceli (approved):

70ray:
using the javascrpt:void(0) is quite commonly used so I don't think we need to take any trouble to avoid it. I don't know about adding another link.

srjfoo (approved):

PR #613, opened 2021-04-09
page.image widget implemented

70ray:
similar to cpeel's proposal.
now in page browser.
not persistent yet

in sandbox https://www.pgdp.org/~rp31/c.branch/control_box_2

Here's a [direct link to a page](https://www.pgdp.org/~rp31/c.branch/control_box_2/tools/page_browser.php?project=projectID45c225f598e32&imagefile=019.png&mode=image).


cpeel:
This is really quite amazing, very nicely done.

Architecturally within the code we should pull out the image widget from the page browser such that it can stand completely alone for re-use elsewhere, including within the page browser. That is, I would like to be able to create an imageWidget -- by passing in a div that it controls -- that is entirely self-contained such that it can be plugged into the current Standard and Enhanced interfaces. The imageWidget would need a control to set the page when it changes (either within the page browser or in the proofreading interface).

cpeel:
The rotate controls are nice but don't do quite what I would expect. In the horizontal view I can rotate the image and end up not seeing any of the image on the screen. I suspect that making this work intuitively is going to require some amount of user input and effort and I'm not certain it's worth investing in right now.

I would like to see Zoom in / Zoom out icons that zoom in/out by a certain percentage though.

srjfoo:
> The rotate controls are nice but don't do quite what I would expect. In the horizontal view I can rotate the image and end up not seeing any of the image on the screen. I suspect that making this work intuitively is going to require some amount of user input and effort and I'm not certain it's worth investing in right now.

And before we show it to the general testing users, we should probably put some thought into designing how we think it should work. In normal mode (presumably whether the image is in portrait or landscape orientation), it appears that the image is anchored to the top of the display area, even if the magnification is set so that the image is shorter than the hight of the display area. Possibly that ought to be the case if the image is rotated, also?

Either way, I mentioned this in Slack to make sure that it was acknowledged as something that had been requested. I agree with Casey that it might be better to put off implementing it. (It might be a good idea to separate the rotation controls from the zoom controls, since it isn't, strictly speaking, a zoom control)

> 
> I would like to see Zoom in / Zoom out icons that zoom in/out by a certain percentage though.

That would be nice, too.

70ray:
> The rotate controls are nice but don't do quite what I would expect. In the horizontal view I can rotate the image and end up not seeing any of the image on the screen. I suspect that making this work intuitively is going to require some amount of user input and effort and I'm not certain it's worth investing in right now.

You are probably right. This was something I'd made earlier. It was quite difficult to make it work as well as it does now so there may not be much chance of improving it.
> 
> I would like to see Zoom in / Zoom out icons that zoom in/out by a certain percentage though.
Rhat should be easy to do.

scripts/control_pane.js
```diff
@@ -0,0 +1,310 @@
...
+    let controlPane = $("<div>", {class: 'page-interface control-panel'});
+    // control2 contains the controls, control1 & control3 adjust the layout
+    let control1 = $("<div>");
+    let control2 = $("<div>").css({flex: '0 1 auto'});
+    let control3 = $("<div>");
+    controlPane.append(control1, control2, control3);
+
+    let menu = $("<div>", {class: "nav-menu"});
+
+    let menuButton = $("<input>", {type: 'button', value: '⌘'}).click(function () {
```
chrismiceli, line 115: I had no clue what this character meant. I don't have a mac and am not used to seeing it. Unicode calls it "PLACE OF INTEREST SIGN" but using the sandbox, I had no idea what to expect when clicking it. Same for the buttons about sizing and moving the controls. Could we have some alt text at least?

scripts/control_pane.js
```diff
@@ -0,0 +1,310 @@
...
+        controlPane.css({"text-align": "", "flex-direction": "row"});
+    }
+
+    function controlVert() {
+        container.css({flexDirection: 'row'});
+        $(".condiv", control2).css({display: "block", padding: "0.1em 0"});
+        controlPane.css({"text-align": "center", "flex-direction": "column"});
+    }
+
+    let leftButton = $("<input>", {type: 'button', class: 'navbutton', value: '⇦'});
```
chrismiceli, line 148: here, is where I think some alt text or text instead of icons could help.

70ray:
I've just checked for browser support and we can't use 'grid' yet. It should be fairly easy to use a table instead. Flex is ok.

scripts/control_pane.js
srjfoo: I think of it as a cloverleaf, but it's most familiar to Mac users as the label for the command key (where Windows uses `ctrl-c`/`v` for copy and paste, Macs use `cmd-c`/`v` for copy and paste, as an example).

I did a half-hearted search of Unicode characters to see if I could find a unicode for a `+` with up/down/left/right arrows on it like Casey's mockup has, but I didn't find anything (there are so many versions of arrows that you'd think there would be, but I didn't find one). Ditto for a compass rose. There is a compass, but I'm not sure that's a good choice. Maybe a diamond? How would something like `❖` or `⟡` or `⟐` work?

Alt text is a good idea.

cpeel: I think using ⌘ is actually rather brilliant, as it _is_ a place of interest. Much of this interface is going to be "try it and see what happens". Agreed that alt text will really help for discovery.

70ray: Yes, that would be good.

cpeel:
Thinking about the size of the `[_____] %` field when the control is docked to the sides: it's really nice having the input be of type number because it gives us validation and use of the keyboard for up/down spinner, but I wonder if we should try applying a (possibly-global) style to remove the spinner UI itself and shrink the field size?

I found https://www.w3schools.com/howto/howto_css_hide_arrow_number.asp

srjfoo:
> Thinking about the size of the `[_____] %` field when the control is docked to the sides: it's really nice having the input be of type number because it gives us validation and use of the keyboard for up/down spinner, but I wonder if we should try applying a (possibly-global) style to remove the spinner UI itself and shrink the field size?

Looks like that would still allow increasing/decreasing by using the arrow keys, even without the visible controls. My only concern would be that some people are very mouse-oriented, and will likely not realize that it's still scrollable using the keyboard. On the other hand, they can type an exact percentage. Could we narrow the input box? I tried typing in huge numbers, and anything greater than 999 will get you a red outline for the box -- ***but*** it'll cheerfully zoom in to 999899% if you type that kind of number in. If the spinner won't go any larger than 999%, should that field size be 3 digits instead of 4? That could narrow the input a bit. For that matter, can we reject anything larger than 999%, or just have it map to 999%? (I just checked and it'll cheerfully accept a negative number, too, which it probably shouldn't. Same red box, but if someone does type in a negative number, perhaps force it to 10%?).

> 
> I found https://www.w3schools.com/howto/howto_css_hide_arrow_number.asp

Nice!

Looking at the `[+]` and `[-]` buttons, would it make more sense for them to be +/- 10% instead of 1%?

70ray:
I've changed the percent input as suggested by @cpeel and added some javascript validation for the cases found by @srjfoo.
Added tooltip titles. The use of definite articles in the titles is inconsistent. I need a second opinion whether to use more or less. Or any improvements to the wording.

cpeel:
Consider enforcing the size of the control buttons with:
```diff
diff --git a/styles/page_interfaces.less b/styles/page_interfaces.less
index 3f5fba865..8ad5bd1f4 100644
--- a/styles/page_interfaces.less
+++ b/styles/page_interfaces.less
@@ -494,6 +494,8 @@
     button, input[type=submit], input[type=button] {
         box-shadow: 1px 1px grey;
         border-radius: 3px;
+        width: 25px;
+        height: 25px;
     }
 }
```

cpeel:
The tooltips and wording look great to me! 👍 

Some other possibly-useful UX improvements:
* When the menu is up, having ⌘ close it -- I keep expecting that to work and it doesn't.
* Styling the ⌘ button as depressed (we do this for the selected pickerset menu) while the menu is up, although maybe not do this if we do the one above?

"control pane", both in the code and filenames, isn't very specific -- what is it controlling? The page? The text? Something else?. "image_control_pane" would be more specific. Actually, is it a requirement that this be a "pane" at all? Could we have some other consumer that uses this for a not-full-pane `<div>` in a page? In that case something like "image_widget" might even be more accurate.

cpeel:
And can I just say again how awesome I think this is!?

cpeel:
Pulling in my comment from slack:
> The arrow icons in the menus look good to me in Chrome but in FF they are small and awkward -- I assume because they are using different fonts. I may try to pull in FontAwesome to see if there are some better icons. (FontAwesome 4.7 is the version we would want to use as after that their licensing changed: https://fontawesome.com/v4.7.0/icons/). These are the icons we use for the social media links on PROD via a hack.

I've got a POC with the FontAwesome icons available at https://www.pgdp.org/~cpeel/c.branch/control_box_2/tools/page_browser.php?project=projectID45c225f598e32&imagefile=014.png&mode=image

If we like that I can push the changes up to this MR if @70ray wants.

srjfoo:
> If we like that I can push the changes up to this MR if @70ray wants.

I like it.

And I agree, it's a terrific enhancement (though I would argue that what we have now is also a great enhancement over what we had before).

srjfoo:
> In that case something like "image_widget" might even be more accurate.

How about "image_widget_controls"? I thought that the image widget was the whole thing, not just the control pane?

srjfoo:
> I've changed the percent input as suggested by @cpeel and added some javascript validation for the cases found by @srjfoo.

Looks good!

> Added tooltip titles. The use of definite articles in the titles is inconsistent. I need a second opinion whether to use more or less. Or any improvements to the wording.

I think the only place where the use of the definite article surprised me was in the alignment -- would it make sense to use "`Left-align controls`", "`Center controls`" and "`Right-align controls`", when the control bar is at the top or bottom (and top/middle/bottom when it's left or right) for consistency, instead of the wordier description? The only reason to change it would be for consistency -- the current descriptions are fine.

cpeel:
> > In that case something like "image_widget" might even be more accurate.
> 
> How about "image_widget_controls"? I thought that the image widget was the whole thing, not just the control pane?

This _is_ the entire widget though.

Looking at it closer the construction of the widget is a bit odd. Currently you pass in the `` element and the code wraps that in a div and attaches everything to it and the caller still "holds onto" the image element to change the src attribute. It would probably be better if widget was passed a `<div>` element that it would control the contents of including adding the `` tag and setting its value. The widget could then export a `setImage()` method that the caller could use to change the image. Functionally I think it would be the same but I think that would be better encapsulation.

70ray:
I've changed the percent input as suggested by @cpeel and added some javascript validation for the cases found by @srjfoo.
Added tooltip titles. The use of definite articles in the titles is inconsistent. I need a second opinion whether to use more or less. Or any improvements to the wording.

> Consider enforcing the size of the control buttons with:
> 
> ```diff
> diff --git a/styles/page_interfaces.less b/styles/page_interfaces.less
> index 3f5fba865..8ad5bd1f4 100644
> --- a/styles/page_interfaces.less
> +++ b/styles/page_interfaces.less
> @@ -494,6 +494,8 @@
>      button, input[type=submit], input[type=button] {
>          box-shadow: 1px 1px grey;
>          border-radius: 3px;
> +        width: 25px;
> +        height: 25px;
>      }
>  }
> ```

Would we want to do this in conjunction with the font change? On its own the characters are no longer in the middle of the button which looks a bit odd.

70ray:
> The tooltips and wording look great to me! +1
> 
> Some other possibly-useful UX improvements:
> 
>     * When the menu is up, having ⌘ close it -- I keep expecting that to work and it doesn't.
> 
>     * Styling the ⌘ button as depressed (we do this for the selected pickerset menu) while the menu is up, although maybe not do this if we do the one above?

Changed to gear symbol and made it toggle the menu.

> 
> 
> "control pane", both in the code and filenames, isn't very specific -- what is it controlling? The page? The text? Something else?. "image_control_pane" would be more specific. Actually, is it a requirement that this be a "pane" at all? Could we have some other consumer that uses this for a not-full-pane `<div>` in a page? In that case something like "image_widget" might even be more accurate.

It would be possible to use the same control bar for the text controls in the page browser, although having it at the side would perhaps not be so good because the font names can be long; it could be made configurable so the "side" options were suppressed. Anything with 'image' would then not be approprate.
I'm quite happy to change it to something else -- I'm not very inventive with names.

70ray:
> I think the only place where the use of the definite article surprised me was in the alignment -- would it make sense to use "`Left-align controls`", "`Center controls`" and "`Right-align controls`", when the control bar is at the top or bottom (and top/middle/bottom when it's left or right) for consistency, instead of the wordier description? The only reason to change it would be for consistency -- the current descriptions are fine.

I think the shorter descriptions are better!

70ray:
> Looking at it closer the construction of the widget is a bit odd. Currently you pass in the `` element and the code wraps that in a div and attaches everything to it and the caller still "holds onto" the image element to change the src attribute. It would probably be better if widget was passed a `<div>` element that it would control the contents of including adding the `` tag and setting its value. The widget could then export a `setImage()` method that the caller could use to change the image. Functionally I think it would be the same but I think that would be better encapsulation.

That seems a reasonable way to do it.

cpeel:
> Would we want to do this in conjunction with the font change? On its own the characters are no longer in the middle of the button which looks a bit odd.

I looked into this and it seems that FF has some left padding on the button which makes it not-centered (Chrome doesn't have this padding) -- this seemed to help:
```diff
diff --git a/styles/page_interfaces.less b/styles/page_interfaces.less
index 8ad5bd1f4..672b63d02 100644
--- a/styles/page_interfaces.less
+++ b/styles/page_interfaces.less
@@ -496,6 +496,7 @@
         border-radius: 3px;
         width: 25px;
         height: 25px;
+        padding: 0;
     }
 }

```

srjfoo:
> It would be possible to use the same control bar for the text controls in the page browser, although having it at the side would perhaps not be so good because the font names can be long; it could be made configurable so the "side" options were suppressed. Anything with 'image' would then not be approprate.

Just for consideration, I was envisioning the controls for the text being completely separate from those for the image, so that if the proofer/formatter chose side, they could choose the outside. Also, when the text-only mode is being used, I assume all four options would be available? Though you're right about the font names.

I would want text controls to be with the text pane.

cpeel:
@70ray, I've merged in Font Awesome into master so we can use its icons here. If you rebase off of master and pull down my fork you can cherry-pick https://github.com/cpeel/dproofreaders/commit/52f92ad2ebb0fa6aabae94e54078aa1383a0d8d5 to use the icons.

70ray:
> @70ray, I've merged in Font Awesome into master so we can use its icons here. If you rebase off of master and pull down my fork you can cherry-pick [cpeel@52f92ad](https://github.com/cpeel/dproofreaders/commit/52f92ad2ebb0fa6aabae94e54078aa1383a0d8d5) to use the icons.

I've done this but something is not right. The gear icon looks different and the fit width and height arrows are missing.

cpeel:
> I've done this but something is not right. The gear icon looks different and the fit width and height arrows are missing.

The gear is the one from Font Awesome, so that part is OK. When I "upgraded" my use of Font Awesome from 4 to 5 I got the names of the arrows wrong, sorry about that:
```diff
diff --git a/scripts/control_pane.js b/scripts/control_pane.js
index 78e88ed53..b0398615d 100644
--- a/scripts/control_pane.js
+++ b/scripts/control_pane.js
@@ -54,14 +54,14 @@ var imageControl = function(imageElement, data, saveData) {
         imageElement.width('100%');
         unPersist();
     })
-        .append($("<i>", {class: 'fas fa-arrows-h'}));
+        .append($("<i>", {class: 'fas fa-arrows-alt-h'}));

     let fitHeight = $("<button>", {title: texts.fitHeight}).click(function () {
         imageElement.height('100%');
         imageElement.width("auto");
         unPersist();
     })
-        .append($("<i>", {class: 'fas fa-arrows-v'}));
+        .append($("<i>", {class: 'fas fa-arrows-alt-v'}));

     let zoomIn = $("<button>", {title: texts.zoomIn}).click(function () {
         percent *= 1.1;
```

scripts/control_pane.js
```diff
@@ -0,0 +1,327 @@
...
+    navStyle();
+
+    return {
+        container: container,
+        contentPane: contentPane,
+        addControls: addControls
+    };
+}
+
+function makeImagePane(imagePaneData, saveData) {
```
cpeel, line 316: Naming-wise, this doesn't have to be a _pane_, it could just be a random `<div>` in the page that the caller wants an image widget in.

cpeel (commented):
Please rename the `control_pane.js/inc` files to `image_widget` or `image_viewer` or similar and use that name consistently in the files to refer to this new widget. We should also update those files and `page_interface.less` to not use "control pane" or "pane" at all since it can be used outside of a pane. So `makeImagePane` could become `makeImageViewer` or `makeImageWidget`.

scripts/control_pane.js
70ray: The same control area could in principle be used with a text area, (though I haven't managed to make it work properly with the sub-splitter in the mentor mode yet).

cpeel: If you want to make this a generic control pane that could be used for both an image and text it needs some restructuring to be more modular. A generic control pane shouldn't include controls other than managing the pane positions. The caller would construct the controls to insert into the control pane (such as the fit-to-size, zoom percent, zoom in/out) and the key to persist the position for that.

I'm not against that work, but if that's your intent it needs some rearchitecting.

70ray: That should be easy to do. I'm also now inclined to give the divs to the function rather than return them as you suggested earlier.
How about "control bar" for the more generic name?

70ray: I just noticed your original proposal allowed for persistent fit width or height. I'll make it work like that. We can always change it if people prefer it not to persist.

70ray: Your way of storing toolbar position is really neat. I'll make it like that.

cpeel: "Control bar" is a great name for the generic dockable control bar.

> That should be easy to do. I'm also now inclined to give the divs to the function rather than return them as you suggested earlier.

Actually, my suggestion for the image widget was to pass the container `<div>`s to the function and have your code consider it the parent bounding box and add the necessary contents inside it. I was unclear about that, my apologies. For the control bar I'm not exactly certain what makes the most sense to pass in such that it can easily be reused.

70ray: > I just noticed your original proposal allowed for persistent fit width or height. I'll make it work like that. We can always change it if people prefer it not to persist.
This seems to lead to excessive complexity, perhaps leave it to later.

cpeel: That makes sense to me.

70ray:
Text widget also implemented.

cpeel:
The new code layout looks good!

Something that I suspect users will complain about and we should decide if we care about (we probably do) is the horizontal alignment of the image in the horizontal view. In the currently-live Page Browser the image is left-aligned. In the new model it is centered. In the past, users have been vocal about wanting to have the image left-aligned such that it is right above the text when using the horizontal orientation.

We are inconsistent in how this currently works in the proofreading interfaces, though:
* Standard
  * Image: centered
  * Text: centered
* Enhanced
  * Image: left-aligned
  * Text: centered

So ironically the new image viewer would be better for the enhanced interface with the new centered model, but I suspect people will complain about the change of image alignment in the Page Browser.

Some possible solutions:
* Left-align the image in the horizontal orientation but center-align it in the vertical orientation
* Add a `<`, `|`, `>` buttons (using icons that match the dock control) to the image toolbar to allow the user to change the alignment.

The second one is more flexible and more forward-looking but it adds additional complexity that I'm not certain is needed right now. Thoughts @srjfoo?

srjfoo (commented):
> The second one is more flexible and more forward-looking but it adds additional complexity that I'm not certain is needed right now. Thoughts @srjfoo?

I thought I remembered a task that requested that the image be centered in the Enhanced Interface -- actually, I thought I remembered more than one, but I only found one when I went digging: [Task 378](https://www.pgdp.net/c/tasks.php?action=show&task_id=378).

This is where the question of design comes in, I think. In both of the current proofreading interfaces, the textarea is an object in the text div/frame, and is not congruent with the edges of said max borders for the container. So, I think the first question is, when we move on to implementing the new proofreading interface, will we be following the same design as before (i.e. the text widget will be contained within, but not congruent with the text div)? (Incidentally, when we get to that point, that may be a good thing to get public input on, to give us enough of a feel for which is the better option.) Note that I don't expect that question to be answered now, but it does tend to indicate that maximum flexibility might be appropriate.

Right now, in the page browser, the text is always left-aligned, and congruent (at least, I hope I'm using "congruent" correctly 😁 ) with the edges of the container div. Given that, the image should probably be left-aligned, also. Users can, and probably will continue to, narrow or widen the window to get what they want to see -- we can't really align things perfectly. Take this for example: Because of the large left margin on the image, there's no way you can get the left edges of the text in the image to automatically align with the text in the horizontal text display. If the user wants the text to actually align, they will have to make sure that the image can scroll, and do the alignment manually.
image
like so 
image

For my purposes for the second example above, it doesn't really matter whether the image is centered or left-aligned, because in order to make that work, I have to have the horizontal scroll for the image.

Anyway, back to the original question. Should it be user selectable? Probably, unless it makes the code too complex. If it does, we should choose one and be able to defend our choice in public, and treat text and image the same. And presumably be willing to change from one to the other if the bulk of the evidence goes against us. But, again, treat text and image the same.

I suspect that if we're going to add that flexibility, it probably ought to be added now, but I can be convinced otherwise.

---

Other things ... 

1. This is just a niggle, and probably not worth mentioning, but ... in the code, `centre` vs. `center`. Should we be standardized on the U.S. spelling? I searched the code and found no instances of `centre` in the active code, but lots of `center` (note that I was using the lazy OS search, so it's entirely possible that I missed something). I'm sure there are other instances of inconsistency between U.S. and British English spellings, so it's probably not worth worrying about, but I thought I'd ask.
2. I'm not sure there's anything we can do about this one, but is there a way to clarify what is meant by wrap? As an example, in (Format) Preview, `wrap` means paragraph wrap. In the proofreading interface and the page browser, it means `line wrap`. Would it help to make a tooltip for it that says "line wrap"? I'm not actually sure that it's worth worrying about, because "line wrap" is what's been meant for the proofreading interfaces from the beginning -- it's only since we've had Ray's Format Preview that we've had the option to view the paragraph wrap.
3. One of the other things that has been mentioned is the ability to flip the position of text and image (i.e. so that image is on the right or the bottom). On thinking about it some more, I realize that, like being able to rotate or invert an image, this is perhaps better left for future enhancements, so perhaps what we need to do is look at what we want in the core functionality?

My questions above are just things that I thought were worth keeping in mind; the bottom line is that I'm loving the way it works!

70ray:
Perhaps we should just make the alignment match how it is now.

70ray:
>     1. This is just a niggle, and probably not worth mentioning, but ... in the code, `centre` vs. `center`. Should we be standardized on the U.S. spelling? I searched the code and found no instances of `centre` in the active code, but lots of `center` (note that I was using the lazy OS search, so it's entirely possible that I missed something). I'm sure there are other instances of inconsistency between U.S. and British English spellings, so it's probably not worth worrying about, but I thought I'd ask.

I guess we should standardize on US English.

srjfoo:
> We can make the alignment match how it is now.

How are you defining "how it is now"? How it currently works in the page browser? How it currently works in the proofreading interfaces? If the latter, remember, there are two proofreading interfaces. One, the image is left-aligned, and the other, it's centered.

70ray:
> > We can make the alignment match how it is now.
> 
> How are you defining "how it is now"? How it currently works in the page browser? How it currently works in the proofreading interfaces? If the latter, remember, there are two proofreading interfaces. One, the image is left-aligned, and the other, it's centered.

I meant varying with the context to match how it was.

70ray:
The image widget is now in the proofreadinf interfaces. This is just to show it working and would need a lot of tidying up.

cpeel:
> The image widget is now in the proofreading interfaces. This is just to show it working and would need a lot of tidying up.

Great! Looks like we needed to make some changes to some of our standard controls for that to work, so it's good to figure that out now.

What I'd like to do from here is to:
1. Reduce this MR to only the changes for the new image widget in the Page Browser (which I think are all the commits before 8f1c5cf?)
2. Get this merged into a new image-widget branch (which I will create in just a moment and update the target for this MR)
3. Publish the upstream image-widget branch to a sandbox (I can do this if you'd like)
4. Start soliciting feedback from our focus group

After we have this merged in, if you want to open a separate work-in-progress MR for the proofreading interface changes so we don't lose them for when we eventually want to do that work that would be grand.

srjfoo:
> > The image widget is now in the proofreading interfaces. This is just to show it working and would need a lot of tidying up.
> 
> Great! Looks like we needed to make some changes to some of our standard controls for that to work, so it's good to figure that out now.
> 
> What I'd like to do from here is to:
> 
>    1. Reduce this MR to only the changes for the new image widget in the Page Browser (which I think are all the commits before [8f1c5cf](https://github.com/DistributedProofreaders/dproofreaders/commit/8f1c5cf7459cc332358bc363912cb3d1d3406c45)?)
>    2. Get this merged into a new image-widget branch (which I will create in just a moment and update the target for this MR)
>    3. Publish the upstream image-widget branch to a sandbox (I can do this if you'd like)
>    4. Start soliciting feedback from our focus group
> 
> 
> After we have this merged in, if you want to open a separate work-in-progress MR for the proofreading interface changes so we don't lose them for when we eventually want to do that work that would be grand.

That sounds like a good idea. I liked seeing how the image widget would fit into the proofreading interfaces, but I don't think we're quite ready for that to be finalized yet -- and definitely not ready to fit the text widget in, yet.

70ray:
> > The image widget is now in the proofreading interfaces. This is just to show it working and would need a lot of tidying up.
> 
> Great! Looks like we needed to make some changes to some of our standard controls for that to work, so it's good to figure that out now.
> 
> What I'd like to do from here is to:
> 
>     1. Reduce this MR to only the changes for the new image widget in the Page Browser (which I think are all the commits before [8f1c5cf](https://github.com/DistributedProofreaders/dproofreaders/commit/8f1c5cf7459cc332358bc363912cb3d1d3406c45)?)
> 
>     2. Get this merged into a new image-widget branch (which I will create in just a moment and update the target for this MR)
> 
>     3. Publish the upstream image-widget branch to a sandbox (I can do this if you'd like)
> 
>     4. Start soliciting feedback from our focus group
> 
> 
> After we have this merged in, if you want to open a separate work-in-progress MR for the proofreading interface changes so we don't lose them for when we eventually want to do that work that would be grand.

It might be easiest to re-make the image widget commits. The way the local storage works is better after the last commit.
The changes to the split control are probably not needed for the image widget alone but it could be best to put them in so they can be tested. I'd like to look at that area again. The previous splitter had some strange effects when using it with the text widget with a splitter to make a restricted view.

70ray:
I've made a new branch and removed the commits for the proofreading interface as you suggested.

cpeel:
Should we close this now that https://github.com/DistributedProofreaders/dproofreaders/pull/623 is merged into a new upstream branch?

70ray:
> Should we close this now that #623 is merged into a new upstream branch?

Yes I'll do that.

PR #614, opened 2021-04-10
Pre-coding-standards fixes

cpeel:
After testing PHP-CS-Fixer against our codebase for standardizing the style, there were a few changes that it made that weren't ideal mostly around comments and some of our array definitions. This addresses those that I found and also removes all the `// vim:` lines.

There should be _no functional changes_ in this MR, but I put it up in a sandbox just in case someone wanted to poke at it: [pre-coding-standards-fixes](https://www.pgdp.org/~cpeel/c.branch/pre-coding-standards-fixes/).


cpeel:
To make this MR (a bit) easier to review I've removed the `// vim:` updates commit. I'll put that in the coding-standards MR since it'll touch everything anyway.

SETUP/check_require_login.php
```diff
@@ -64,7 +64,6 @@
     "pinc/udb_user.php",
     "pinc/site_registration_protection.php",
     # 3rd party libraries
-    "pinc/functions_insert_post.php",
```
chrismiceli, line 67: we don't need this file anymore?When I try to access on PROD I get:
image

faq/pt/proofreading_guidelines.php
```diff
@@ -8,7 +8,7 @@
 include_once($relPath.'pg.inc');
 include_once($relPath.'theme.inc');
 
-#maybe_redirect_to_external_faq("pt");
+maybe_redirect_to_external_faq("pt");
```
chrismiceli, line 11: this is a functional change, no?

pinc/showavailablebooks.inc
```diff
@@ -243,14 +243,14 @@ function show_projects_for_pool( $pool, $checkedout_or_available )
...
-        "genre"               => array( 'Genre',  _("Genre") ),
-        "n_pages"             => array( 'PgTot',  _("Pages") ),
-        $pool->foo_field_name => array( 'Person', $pool->foo_Header ),
-    );
+    $columns = [
+        "nameofwork" => ['Title', _("Title")],
+        "authorsname" => ['Author', _("Author")],
+        "language" => ['Lang', _("Language")],
+        "genre" => ['Genre', _("Genre")],
+        "n_pages" => ['PgTot', _("Total Pages")],
```
chrismiceli, line 251: another functional change, right? Just making sure they're intentional.

chrismiceli (approved):

faq/pt/proofreading_guidelines.php
srjfoo: But will only affect the Portuguese Proofreading Guidelines. The other languages already have active redirects to the wiki.

SETUP/check_require_login.php
cpeel: Nope. It was for phpBB2 support which we dropped in 2017. Removing the file is part of this MR.

faq/pt/proofreading_guidelines.php
cpeel: Yes, this is a functional change (so my statement about there not being one is incorrect). This is intentional -- I don't know why was commented out prior -- I assume it was a mistake somewhere.

pinc/showavailablebooks.inc
cpeel: This was a copy/paste mistake - thank you for catching it!

accounts/addproofer.php
```diff
@@ -175,8 +175,16 @@
...
+echo "    <input type='radio' name='email_updates' value='1' ";
+if ($email_updates) {
+    echo "checked";
+}
+echo "    >" . _("Yes") . "&nbsp;&nbsp;";
+echo "    <input type='radio' name='email_updates' value='0' ";
+if (!$email_updates) {
+    echo "checked";
+}
+echo "    >" . _("No");
```
srjfoo, line 187: The code obviously works the way it is, so this is just curiosity -- when should a label (like the "Yes" and "No") actually be wrapped in a `<label>` element?

pinc/stages.inc
```diff
@@ -15,40 +15,34 @@ include_once($relPath.'access_change_emails.inc'); // pp_access_change_email(),
...
+if ($site_supports_metadata) {
+    new Stage(
+        'CA',
+        _('Clearance Approvals'),
+        array(),
+        'REQ-HUMAN',
+        '',
+        NULL, /* access_change_callback */
+        _('A qualified person must confirm that the work is in the public domain.'),
+        NULL,
```
srjfoo, line 27: `NULL` to lower case?

pinc/stages.inc
```diff
@@ -15,40 +15,34 @@ include_once($relPath.'access_change_emails.inc'); // pp_access_change_email(),
...
+if ($site_supports_metadata) {
+    new Stage(
+        'MD',
+        _('Metadata Collection'),
+        array(),
+        'REQ-HUMAN',
+        '',
+        NULL, /* access_change_callback */
+        _('We collect various bits of metadata for each page in a project.'),
+        NULL,
```
srjfoo, line 43: ditto above, and others below?

srjfoo (approved):

pinc/stages.inc
cpeel: I'm actually going to let the code formatter take care of those things (as well as a few more in this stanza). I just wanted to get them indented since the formatter didn't do what I wanted it to.

pinc/stages.inc
cpeel: ditto my reply above :)

accounts/addproofer.php
cpeel: `<label>` elements are optional. They allow users to click on the text to access the associated checkbox/radio buttons in addition to the button themselves, but they are not required. We don't use them consistently in the code (indeed, we use them in relatively few places).

srjfoo: Got it! Thanks for the clarification!

70ray (approved):

cpeel:
I will wait to merge this in after the deployment next week (and probably a few days of bake-time thereafter).

cpeel:
The rotate controls are nice but don't do quite what I would expect. In the horizontal view I can rotate the image and end up not seeing any of the image on the screen. I suspect that making this work intuitively is going to require some amount of user input and effort and I'm not certain it's worth investing in right now.

I would like to see Zoom in / Zoom out icons that zoom in/out by a certain percentage though.

accounts/addproofer.php
chrismiceli: I can take a look at enhancing these checkboxes once the formatting MR is resolved.

chrismiceli (approved):

PR #615, opened 2021-04-12
Add Font Awesome Free 5.15.3

cpeel:
This pulls in [Font Awesome](https://fontawesome.com/) Free, an icon pack via web font. This gives us access to a wide number of icons we can use across the site, such as in the in-development image viewer widget.


srjfoo (approved):

pinc/html_page_common.inc
```diff
@@ -63,6 +63,7 @@ function output_html_header($nameofpage, $extra_args = array(), $show_statsbar =
     $theme_name = $user ? $user->i_theme : "project_gutenberg";
     $css_files = array(
         "$code_url/styles/themes/$theme_name.css",
+        "$code_url/pinc/3rdparty/fontawesome/css/all.min.css",
```
chrismiceli, line 66: Do we need all the different font styles? I'm sure the browser probably only requests the font files for the used style sheets, but may prefetch if it gets bored and we could avoid that if we used just the stylesheets for the fonts we intend to use. I'm specifically wonder about brands:
https://fontawesome.com/how-to-use/on-the-web/setup/hosting-font-awesome-yourself#using-certain-styles

chrismiceli (commented):
Per https://fontawesome.com/how-to-use/on-the-web/setup/hosting-font-awesome-yourself#download-contents, it seems like there are other files included in font awesome not included here. Is this some subset of a release?

chrismiceli:
Does this incur a font download on a page that doesn't have any icon fonts? Are we worried about the size of all.min.css (~60kb) on pages that don't have icon fonts? If so, we can opt into it? I suppose there is browser caching so it is only loaded once hopefully.

pinc/html_page_common.inc
cpeel: I thought it easiest if we include all of the styles so if a developer goes to the Font Awesome website and sees an icon in the Free pack they want to use then it's available. I actually want the brands for pgdp.net so I can remove the PROD hack for the social media icons :D

chrismiceli:
Also, for what it is worth. It seems the industry is moving towards svg for icons and font-awesome does seem to support svg sprites. https://www.lambdatest.com/blog/its-2019-lets-end-the-debate-on-icon-fonts-vs-svg-icons/

cpeel:
> Per https://fontawesome.com/how-to-use/on-the-web/setup/hosting-font-awesome-yourself#download-contents, it seems like there are other files included in font awesome not included here. Is this some subset of a release?

Yes, per the `pinc/3rdparty/fontawesome/README.md` this is just the webfont subset -- not the original SVGs or the JS stuff.

chrismiceli:
> 
> 
> > Per https://fontawesome.com/how-to-use/on-the-web/setup/hosting-font-awesome-yourself#download-contents, it seems like there are other files included in font awesome not included here. Is this some subset of a release?
> 
> Yes, per the `pinc/3rdparty/fontawesome/README.md` this is just the webfont subset -- not the original SVGs or the JS stuff.

I'm expected to read that file called README? :D

cpeel:
> Does this incur a font download on a page that doesn't have any icon fonts? Are we worried about the size of all.min.css (~60kb) on pages that don't have icon fonts? If so, we can opt into it? I suppose there is browser caching so it is only loaded once hopefully.

The CSS will be downloaded with every page -- just like our theme CSS -- but:
1. we go through a lot of effort to make efficient use of the browser cache for CSS / JS via Apache configs
2. we serve up deflated content

The font files themselves (woff, woff2, etc) are only downloaded by the browser if the font is actually used on the page (and then it is cached aggressively via Apache config settings).

> Also, for what it is worth. It seems the industry is moving towards svg for icons and font-awesome does seem to support svg sprites. https://www.lambdatest.com/blog/its-2019-lets-end-the-debate-on-icon-fonts-vs-svg-icons/

Oh, *interesting*! I wasn't aware of this! Although seeing how complicated it is to style SVG icons based on that article I don't think I want to blaze down that path quite yet.

chrismiceli (approved):

PR #616, opened 2021-04-14
Ensure valid memcache keys for API rate limiting

cpeel:
We use usernames as memcache keys when tracking API access for rate limiting. Memcache keys can't contain whitespaces or control characters -- but usernames _can_ contain whitespaces -- which causes API access to fail for users with spaces in their names. This results in a server error which breaks the Page Browser for these users.

We can't merely replace spaces with another character as that might potentially conflict with a different user. To fix this, simply hash the username. Also log the error server-side so we are aware of it.

Testable in the [fix-api-rate-key-names](https://www.pgdp.org/~cpeel/c.branch/fix-api-rate-key-names) sandbox.

### DIFF
--- a/api/index.php
+++ b/api/index.php
@@ -83,6 +83,11 @@ function api_rate_limit($key)
         return;
     }
 
+    // memcached keys can't contain whitespaces or control characters:
+    // https://github.com/memcached/memcached/blob/master/doc/protocol.txt#L41
+    // To enforce that, hash the key.
+    $key = md5($key);
+
     $memcache = new Memcached();
     $memcache->addServer('localhost', 11211);
 
@@ -110,7 +115,8 @@ function api_rate_limit($key)
     if($memcache->set("$key:count", $count) === FALSE)
     {
         // if we can't set the value, memcached probably isn't running
-        // regardless, we can't enforce rate limiting so return an error
+        // regardless, we can't enforce rate limiting so log it and return an error
+        error_log("api/index.php - Error setting $key:count=$count in memcache");
         throw new ServerError();
     }
 


srjfoo (approved):

api/index.php
```diff
@@ -83,6 +83,11 @@ function api_rate_limit($key)
         return;
     }
 
+    // memcached keys can't contain whitespaces or control characters:
+    // https://github.com/memcached/memcached/blob/master/doc/protocol.txt#L41
+    // To enforce that, hash the key.
+    $key = md5($key);
```
chrismiceli, line 89: probably not a huge concern, but md5 is susceptible to [length extension attacks](https://en.wikipedia.org/wiki/Length_extension_attack) so in theory someone could create a username to cause a hash conflict on the memcache entry, causing a user to be able to block another user by making them hit their rate limit. Probably not a huge concern, but should we use a different hash algorithm? Or maybe some dumb algorithm like:
```
'user name' => 'user|name' // replace ' ' with '|'
'user|name' => 'user||name' // replace '|' with '||'
```
that should prevent any conflict on the replace unless I'm mistaken.

chrismiceli (approved):

api/index.php
cpeel: Usernames are limited to 25 characters, so I'm not worried about people trying to generate hash collisions just to -- at most -- double their rate limit. I also want this to be fast since it's in the path of every API call.

Usernames are restricted to `a-zA-Z0-9@._ ` so I thought about doing a simple search/replace with some other character, but:
1. that assumes that someone wouldn't edit the list of valid characters in `pinc/User.inc` to include the now-special character
2. that the caller of this function doesn't decide to use some other key besides the username which might also contain invalid characters for memcache keys

Hashing the value seemed like the best way to make this work now and also prevent surprises by future code changes that might run into problems with other encoding solutions.

srjfoo: Actually, I thought we specifically excluded the underscore character because of the wiki. Have we also now excluded things like `@` (we have quite a few usernames with that, also)?

cpeel: New registrations cannot use the underscore, but existing accounts still have them.

cpeel: See https://github.com/DistributedProofreaders/dproofreaders/blob/master/pinc/User.inc#L459 for the gory details :D

70ray (approved):

PR #617, opened 2021-04-15
scroll text and image to top when page changes

70ray:
This should make the behaviour like it was before.

in sandbox https://www.pgdp.org/~rp31/c.branch/change_page_scroll

### DIFF
--- a/scripts/page_browse.js
+++ b/scripts/page_browse.js
@@ -332,7 +332,8 @@ function pageBrowse(params, replaceUrl, mentorMode = false) {
                 document.title = proofIntData.strings.displayPageX.replace("%s", imageFileName);
                 if(displayMode !== "text") {
                     imageElement.attr("src", page.image_url).parent()
-                        .scrollTop(0).scrollLeft(0);
+                        .scrollTop(0)
+                        .scrollLeft(0);
                 }
                 if(displayMode !== "image") {
                     // if the supplied round_id was invalid it will be replaced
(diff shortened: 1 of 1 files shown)


cpeel:
I think this is a good change. Are there other differences we should be thinking about before someone else does? We retained the zoom level before, so that one is fine.

scripts/page_browse.js
```diff
@@ -331,7 +331,8 @@ function pageBrowse(params, replaceUrl, mentorMode = false) {
                 params.set("imagefile", imageFileName);
                 document.title = proofIntData.strings.displayPageX.replace("%s", imageFileName);
                 if(displayMode !== "text") {
-                    imageElement.attr("src", page.image_url);
+                    imageElement.attr("src", page.image_url).parent()
+                        .scrollTop(0);
```
chrismiceli, line 335: this is only vertical position, right? Do we want to reset horizontal scroll as well?

cpeel: Oh, good point. The prior behavior is "scroll to the top left", but is that what people would want/expect? People may want to zoom + horiz scroll to remove the margins and have that persist across pages.

Although changing it so that it scrolls top and left is probably the path of least resistance at this point in time.

70ray: Good point, I'll look at it also.

srjfoo: Top left as a default makes the most sense to me. We're unlikely to please everyone, but I would think that's what most people would expect, if not for it to stay in the same position as the previous page.

70ray: Good point, I'll look at it also.

70ray: Implemented now in the sandbox.

chrismiceli:
Looks like there are 2 lint warnings we should address.

70ray:
> Looks like there are 2 lint warnings we should address.

Sorry, I knew that was going to happen but forgot to fix it.

chrismiceli (approved):

cpeel:
> > Looks like there are 2 lint warnings we should address.
> 
> Sorry, I knew that was going to happen but forgot to fix it. Stangely the warnings don't show up on my view of GitHub, has something changed?

It looks like Warnings don't fail the linter -- just Errors: https://github.com/DistributedProofreaders/dproofreaders/runs/2360080048?check_suite_focus=true#step:7:10

That's something we might want to fix.

PR #618, opened 2021-04-19
Fail on warnings

chrismiceli:
This is to address a comment from @cpeel that our js linter should fail on warnings. I also took the liberty to update dependencies to the latest versions.

### DIFF
--- a/SETUP/check_require_login.php
+++ b/SETUP/check_require_login.php
@@ -115,6 +115,10 @@ foreach($files as $file)
     if(startswith($file, "vendor/"))
         continue;
 
+    # If it's in the node_modules directory, skip it
+    if(startswith($file, "node_modules/"))
+    continue;
+
     echo "$file\n";
 
     # If it requires authentication, skip it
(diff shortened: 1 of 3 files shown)


SETUP/check_require_login.php
```diff
@@ -115,6 +115,10 @@
     if(startswith($file, "vendor/"))
         continue;
 
+    # If it's in the node_modules directory, skip it
+    if(startswith($file, "node_modules/"))
```
chrismiceli, line 119: Apparently the flatted npm dependency required by has a php files in it:
https://github.com/WebReflection/flatted
Strange.
It is used by flat-cache which is required by eslint itself.

cpeel: Please go ahead and use the new if/then styling format even though it doesn't (yet) match the rest of the file:
```php
    // If it's in the node_modules directory, skip it
    if (startswith($file, "node_modules/")) {
        continue;
    }
```

That'll be 3 fewer lines that PHP-CS-Fixer needs to change 😁

PR #620, opened 2021-04-21
skip node_modules

chrismiceli:
I noticed it scanned and found 2 php files in node_modules. This is a waste of processing as we don't want to touch these files. The linter also made these other changes so I thought I'd include them here.


cpeel:
Thanks for finding this! Weird that the fixer didn't fix those two files in my pass.

Issue #619, opened 2021-04-20
Add PHP-CS-Fixer checks to Github Actions CI


cpeel:
Now that we have PHP coding standards we should enforce them in a Github Actions CI workflow. The workflow should output a diff of any changes it detects to assist the user.

PR #622, opened 2021-04-22
Add PHP-CS-Fixer to GitHub linting

cpeel:
Add PHP style enforcement to our CI pipeline. Note that the pipeline will (correctly) fail right now until https://github.com/DistributedProofreaders/dproofreaders/pull/621 merges in. I intentionally wanted to get this in prior so we can see what it looks like when it fails.

This fixes https://github.com/DistributedProofreaders/dproofreaders/issues/619

### DIFF
--- a/.github/workflows/static-checks.yml
+++ b/.github/workflows/static-checks.yml
@@ -26,7 +26,7 @@ jobs:
           php install_db.php
           cd ..
       - name: Install composer dependencies
-        run: composer install
+        run: composer install --no-dev
       - name: Install aspell dependencies for WordCheck tests
         run: sudo apt install aspell aspell-en
       - name: Run phpunit tests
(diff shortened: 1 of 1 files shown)


srjfoo (approved):

.github/workflows/static-checks.yml
```diff
@@ -49,20 +49,28 @@ jobs:
...
-        run: cd SETUP && make lint_code && cd ..
-      - name: Run js lint
+      - name: Run JS lint
         run: npm run lint
+      - name: Install composer dependencies
+        run: composer install
+      - name: Run PHP-CS-Fixer and confirm no changes
+        run: |
+          ./vendor/bin/php-cs-fixer fix
+          git diff --exit-code
```
chrismiceli, line 73: this seems kind of iffy to me. Could we not use some cli arg to php-cs-fixer, like --stop-on-violation
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.18/doc/usage.rst

cpeel: `--stop-on-violation` will stop after finding/fixing the first one but returns a zero exit code so the pipeline won't show a failure.

We could use `--dry-run --diff` which will return a non-zero exit code and still show the diff. I can change it to that.

chrismiceli (approved):

PR #623, opened 2021-04-23
New Image Widget

70ray:
This uses imageWidget in page  browse and word suggestions tbut not for proofreading. Also textWidget only in page browse.


styles/page_interfaces.less
```diff
@@ -458,3 +458,59 @@
         min-height: 10px;
     }
 }
+
+/* ------------------------------------------------------------------------ */
+/* Control Pane */
+
+.control-panel {
```
cpeel, line 465: Please rename this `control-bar` to reflect the JS naming, also the comment two lines up.

styles/page_interfaces.less
```diff
@@ -458,3 +458,59 @@
...
+        height: 25px;
+        padding: 0;
+    }
+
+    input[type=checkbox] {
+        vertical-align: bottom;
+    }
+}
+
+.nav-menu {
```
cpeel, line 507: Perhaps `control-bar-menu`?

cpeel:
I've changed this MR to merge into the new `image-widget` branch. My hope is that this branch contains mergeable as-is commits into `master` when we're ready, but this way we can do iterative design work.

srjfoo (approved):
Some quick testing, and it appears to work as expected.

70ray:
Should we change "wrap" to "line wrap" as @srjfoo  commented earlier? The longer text would stil be shorter than the font selector.

srjfoo:
> Should we change "wrap" to "line wrap" as @srjfoo commented earlier? The longer text would stil be shorter than the font selector.

I just wanted to mention that I'm fine with whatever choice is made. I mainly brought it up because it wasn't clear to me in the testing which kind of wrap I should expect until after I played with it. However, the userprefs aren't clear on that for the proofreading interfaces, either.

cpeel:
> Should we change "wrap" to "line wrap" as @srjfoo commented earlier? The longer text would stil be shorter than the font selector.

My suggestion is that we leave it "wrap" as that's what's currently in the Page Browser.

cpeel:
I consider this ready to squash-and-merge, any last thoughts before I do?

srjfoo:
> I consider this ready to squash-and-merge, any last thoughts before I do?

Go for it!

PR #624, opened 2021-04-23
prevent scroll caret into view on selecting a picker

70ray:
Task https://www.pgdp.net/c/tasks.php?action=show&task_id=1985
According to https://caniuse.com/?search=focus%3ApreventScroll it won't work on Safari or some older browsers.

### DIFF
--- a/tools/proofers/character_selector.js
+++ b/tools/proofers/character_selector.js
@@ -15,7 +15,7 @@ $(function () {
         $(".selected-tab", charSelector).removeClass("selected-tab");
         $("#" + newCode, charSelector).addClass("selected-tab");
         largeChar.value = ""; // remove old character
-        top.focusText();
+        top.focusText(true);
     }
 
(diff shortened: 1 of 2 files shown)


chrismiceli (approved):

cpeel:
This LGTM. I assume (but have not tested) that Safari just ignores the don't-scroll request and doesn't raise an error but am hoping @srjfoo can test and confirm.

srjfoo:
> This LGTM. I assume (but have not tested) that Safari just ignores the don't-scroll request and doesn't raise an error but am hoping @srjfoo can test and confirm.

I tested it on Safari, and (I've rechecked it twice) what it looks like to me is that it isn't scrolling at all. I.e. when I tested it on the current code as implemented on the main TEST sandbox, it didn't scroll. And when I tested it in Ray's sandbox, it still didn't scroll, so I think we can safely ignore Safari for now.

srjfoo (approved):

PR #626, opened 2021-04-27
Keep browse history

70ray:


tools/page_browser.js
```diff
@@ -1,13 +1,21 @@
-/*global $ pageBrowse mentorMode */
+/*global pageBrowse mentorMode */
 
-$(function () {
+window.onload = function () {
+    // this function is not called when fwd/back button is pressed
     let url = new URL(window.location.href);
-    let params = new URLSearchParams(url.search);
 
     function replaceUrl() {
```
chrismiceli, line 7: Should we name this function pushUrl now that we are using push?

chrismiceli (approved):

PR #628, opened 2021-04-28
Add marking dashes to drag bar and split always

70ray:
change default to 80% and read only except in mentor mode

### DIFF
--- a/scripts/page_browse.js
+++ b/scripts/page_browse.js
@@ -93,32 +93,13 @@ var maketextControl = function(textArea, storageKey) {
     return [fontFaceSelector, fontSizeSelector, wrapControl];
 };
 
-function makeTextWidget(container, storageKey, splitter = false, reLayout = null, readOnly = true) {
+function makeTextWidget(container, storageKey, readOnly = true) {
     let textWidgetKey = storageKey + "-textwidget";
     let textArea = $("<textarea>", {class: "text-pane"});
     textArea.prop("readonly", readOnly);
     let controls = maketextControl(textArea, textWidgetKey);
-    let controlDiv = makeControlDiv(container, controls, textWidgetKey, reLayout);
-    if(splitter) {
-        let splitterKey = textWidgetKey + "-split";
-        let textSplitData = JSON.parse(localStorage.getItem(splitterKey));
-        if(!$.isPlainObject(textSplitData)) {
-            textSplitData = {
-                splitPercent: 80
-            };
-        }
-        let topTextDiv = $("<div>").append(textArea);
-        let bottomTextDiv = $("<div>");
-        controlDiv.content.append(topTextDiv, bottomTextDiv);
-
-        let subSplitter = splitControl(controlDiv.content, {splitVertical: false, splitPercent: textSplitData.splitPercent, reDraw: reLayout});
-        subSplitter.dragEnd.add(function (percent) {
-            textSplitData.splitPercent = percent;
-            localStorage.setItem(splitterKey, JSON.stringify(textSplitData));
-        });
-    } else {
-        controlDiv.content.append(textArea);
-    }
+    let controlDiv = makeControlDiv(container, controls, textWidgetKey);
+    controlDiv.content.append(textArea);
     return {
         setText: function (text) {
             textArea.val(text)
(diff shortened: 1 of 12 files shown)


chrismiceli (approved):

srjfoo (approved):

cpeel:
I think we're going to get questions and a lot of confused users on what will be seen as the "empty" bottom panel in the text pane. If we want to make the text subsplitter (which I think we shouldn't have to begin with as it's a very confusing UX) generally available I think we need to keep the initial size to 100% so it's available but not obvious to decrease the confusion.

I will defer to @srjfoo on the final say however.

srjfoo:
> I think we're going to get questions and a lot of confused users on what will be seen as the "empty" bottom panel in the text pane. If we want to make the text subsplitter (which I think we shouldn't have to begin with as it's a very confusing UX) generally available I think we need to keep the initial size to 100% so it's available but not obvious to decrease the confusion.
> 
> I will defer to @srjfoo on the final say however.

My main concern has nothing to do with the page browser, actually, but looking to the future to the point when the widgets are incorporated into the proofreading interfaces (I am assuming that the text widget, like the image widget, is ultimately destined for the proofreading interfaces).

Right now, both proofreading interfaces have the ability to adjust the size of the textarea. If the text widget is not going to replace the textarea (or the text frame or whatever -- I'm having trouble visualizing it), then I think that adding the splitter for everyone is irrelevant, and we don't need to do it.

I can guarantee you, though, that if we're going to shoehorn the text widget into the proofreading interfaces, we really need to retain the proofreaders' ability to adjust at least the height (in number of lines viewed). If we don't, we _**will**_ see complaints.

In other words, when I have made a point of asking about it in our previous discussions, I've been thinking of the future, not of the present.

So, I guess my questions are: What is the goal of the text widget? If the horizontal control is not part of the widget, what will it be part of?

srjfoo:
Further comment -- I have no problem with leaving this un-merged while we consider it further. Things like "even if we will need it when it comes to fitting the text widget into the proofreading interfaces, it's better to wait until then and not confuse users of the page browser at this point in time". Or anything else that's relevant that I haven't thought of. I'm definitely not omniscient!

70ray:
I must admit the the splitter in the text widget does look a bit odd, especially if the control bar is at the bottom. Perhaps if the control bar was inside, it would look like a resizeable window, similar to how the enhanced interface looks. I'll look into this.

70ray:
I put the text widget inside the sub-splitter. It looks better, but it doesn't do anything useful that you couldn't do by reducing the window height so I think we should remove the sub-splitter altogether.
If anyone comes up with a good reason why we need it then we can put it back.

70ray:
This is running in https://www.pgdp.org/~rp31/c.branch/widget_split

The splitter in the text area is always enabled when image and text are shown, not just for mentors, and defaults to being right at the bottom of the page.

scripts/page_browse.js
```diff
@@ -400,12 +381,23 @@ function pageBrowse(params, storageKey, replaceUrl, mentorMode = false) {
                         imageWidget = makeImageWidget(imageDiv, storageKey);
                         stretchDiv.append(imageDiv, textDiv);
                         let theSplitter = viewSplitter(stretchDiv, storageKey);
-                        if(mentorMode) {
-                            // make a text widget with splitter
-                            textWidget = makeTextWidget(textDiv, storageKey, true, theSplitter.mainSplit.reSize);
-                        } else {
-                            textWidget = makeTextWidget(textDiv, storageKey);
+                        let topTextDiv = $("<div>");//.append(textArea);
```
chrismiceli, line 384: remove the commented code if not needed to save space (I know, not much, but can't hurt).

scripts/splitControl.js
```diff
@@ -53,7 +53,10 @@ var splitControl = function(container, config) {
     let pane1 = $(children[0]);
     let pane2 = $(children[1]);
 
-    let dragBar = $("<div>").css({"background-color": theConfig.dragBarColor, flex: `0 0 ${theConfig.dragBarSize}px`});
+    let dragBar = $("<div>").css({"background-color": theConfig.dragBarColor, flex: `0 0 ${theConfig.dragBarSize}px`, display: 'flex', 'justify-content': 'center'})
+        // this div puts a row of dashes in the centre of the drag bar
+        .append($("<div>").css({border: `${theConfig.dragBarSize / 3}px dashed DimGray`, margin: `${theConfig.dragBarSize / 6}px`, flex: '0 0 6em'}));
```
chrismiceli, line 58: does DimGray make sense for all our themes or should we use some css that is themeable here?

chrismiceli (approved):

scripts/splitControl.js
srjfoo: Currently the proofing interfaces and page browser and related are not themed, but I think should be themable. Anything that specifies a color in the (non-css/less) code will have to have it pulled out at some point.

It's true that the colors in the Format Preview display can be customised by the individual user, but the container cannot be.

Mind you, it works for now, and can be changed when work starts on the theming, but it would be better if it was located in the `page_interfaces.less` file, not in the `js` file.

scripts/page_browse.js
70ray: Thanks, done this.

scripts/splitControl.js
70ray: All of the drag bar styling could be moved to css, even the width which would avoid the new complexity of adjusting the marking to fit the drag bar.

chrismiceli (approved):

srjfoo:
What's the status of this MR?

70ray:
> 
> 
> What's the status of this MR?

I'll close it now. I'll keep my branch so we can perhaps put the drag bar markers in later.

PR #629, opened 2021-05-02
Widget browse

70ray:
This makes a "browser widget" so the position of the controls can be configured.


srjfoo:
This isn't really directly related to the focus of this MR, but I think the "Select a different project" function probably needs a "cancel" option so that if the user changes their mind (or doesn't have the required projectID handy), they can get back to the page browser without having to go back to the page that the page browser was invoked from (mentioned by @chrismiceli in a previous MR). Currently, there's no way off of that page other than selecting a different project, using the browser back button, using the browser history menu or closing the page.

70ray:
> This isn't really directly related to the focus of this MR, but I think the "Select a different project" function probably needs a "cancel" option so that if the user changes their mind (or doesn't have the required projectID handy), they can get back to the page browser without having to go back to the page that the page browser was invoked from (mentioned by @chrismiceli in a previous MR). Currently, there's no way off of that page other than selecting a different project, using the browser back button, using the browser history menu or closing the page.

In this MR, selecting a different project is outside the 'browse widget' so it may be possible to push the history in this case so you could use the back button. I'll look at it.

srjfoo:
Bear in mind that the javascript is over my head, so my comments here are general. As far as I can tell, it works well, and as intended. I'm wondering, though, why separate out the link to the project page and the "Select a different project" form from the rest of the page controls.

chrismiceli:
Is this in a sandbox somewhere?

chrismiceli (approved):

chrismiceli:
> 
> 
> Bear in mind that the javascript is over my head, so my comments here are general. As far as I can tell, it works well, and as intended. I'm wondering, though, why separate out the link to the project page and the "Select a different project" form from the rest of the page controls.

Javascript and all looks good to me

70ray:
> I'm wondering, though, why separate out the link to the project page and the "Select a different project" form from the rest of the page controls.

It seemed like this was at a 'higher level' than the other controls, and is not used at all in the use for looking at word context.

More generally I was wondering what the use case for 'select a different project' is? In actual proofreading or PMing I have only linked to the page I wanted to look at in the page-browser from another page. If it is used, there must be a more convenient way than typing or pasting in a project id: perhaps we could make a link from the project search results?

70ray:
> This isn't really directly related to the focus of this MR, but I think the "Select a different project" function probably needs a "cancel" option so that if the user changes their mind (or doesn't have the required projectID handy), they can get back to the page browser without having to go back to the page that the page browser was invoked from (mentioned by @chrismiceli in a previous MR). Currently, there's no way off of that page other than selecting a different project, using the browser back button, using the browser history menu or closing the page.

I think adding a cancel button would not be appropriate because if you were starting from here then cancel would have no meaning.
If I were to suggest an alert "Are you really sure you want to select adifferent project?" then Casey would be horrified.
Recording history just for this case would allow you to go back but it would seem a bit odd to record history in such an arbitrary way.
Recording all history would perhaps be the easiest option.
Removing the ability to select a different project would avoid the issue but might not be universally popular.

70ray:
Added 2 old commits that were left out because I accidentally started from the wrong place.

cpeel:
I do not think having the bar in the browse interface being dockable is a good user experience and will only add both complexity and confusion.
Screen Shot 2021-05-03 at 8 31 11 AM

I'm also confused on why this change is targeted to the `image-widget` branch as it seems to have nothing to do with the image widget. It his the result of user feedback on the image widget changes?

@70ray, when you open MRs please add more detail in the description of what the change is and why we are making it as well as a sandbox to test the changes.

srjfoo:
> > This isn't really directly related to the focus of this MR, but I think the "Select a different project" function probably needs a "cancel" option so that if the user changes their mind (or doesn't have the required projectID handy), they can get back to the page browser without having to go back to the page that the page browser was invoked from (mentioned by @chrismiceli in a previous MR). Currently, there's no way off of that page other than selecting a different project, using the browser back button, using the browser history menu or closing the page.
> 
> I think adding a cancel button would not be appropriate because if you were starting from here then cancel would have no meaning.

Good point.

> If I were to suggest an alert "Are you really sure you want to select adifferent project?" then Casey would be horrified.

That does seem a bit excessive, yes (an "Are you really sure ..." alert, that is).

> Recording history just for this case would allow you to go back but it would seem a bit odd to record history in such an arbitrary way.
> Recording all history would perhaps be the easiest option.
> Removing the ability to select a different project would avoid the issue but might not be universally popular.

I don't remember the reason for adding the ability to select a different project in the first place -- why was it? I wouldn't miss it, but defer to others.

If we keep it, we keep less-than-optimal behavior on the hopefully rare occasion someone actually uses the button and changes their mind.

srjfoo:
> I do not think having the bar in the browse interface being dockable is a good user experience and will only add both complexity and confusion.

I admit that I don't see much need to be able to change where the page information is docked, either.

It's also important, when testing, not to throw too much at the people you're asking to test for you. Should we be focusing on asking for feedback just on the image widget, itself? We're pretty close to opening this up for general testing. What do we want people looking at?

70ray:
> I do not think having the bar in the browse interface being dockable is a good user experience and will only add both complexity and confusion.

You could well be right. It does look confusing. But I think some people (one in particular comes to mind) would appreciate being able to put the controls at the side or bottom of the page. I thought of perhaps colouring the different control areas differently but this could make it look even more confusing.

> I'm also confused on why this change is targeted to the `image-widget` branch as it seems to have nothing to do with the image widget. It his the result of user feedback on the image widget changes?

You are right, it has nothing to do with the image widget changes. It occurred to me that the same configurable control bar could be used  for the page controls in the browser and perhaps also in the future for the proofreading controls. This seemed like a good place to try it out since the infrastructure is already there. And extending its use could reveal desirable changes which would make it more generally useful. I should have commented that this is only a test run and not intended to be merged into the branch. At least not now.
> 
> @70ray, when you open MRs please add more detail in the description of what the change is and why we are making it as well as a sandbox to test the changes.

Sorry I forgot to mention the sandbox here. It is https://www.pgdp.org/~rp31/c.branch/widget_browse.

70ray:
> I don't remember the reason for adding the ability to select a different project in the first place -- why was it? I wouldn't miss it, but defer to others.

It has been in the browse page (formerly known as /tools/mentors/view_page_text_image.php) for a long time when it was known as "Reset". That's the only reason to keep it.

70ray:
> It's also important, when testing, not to throw too much at the people you're asking to test for you. Should we be focusing on asking for feedback just on the image widget, itself? We're pretty close to opening this up for general testing. What do we want people looking at?

I agree. I couldn't resist trying out maximum configurability when the possibility occured to me but best not to do it now.

70ray:
I'll close this MR, but the comments regarding "select a different project" are still relevant to the image_widget branch.

srjfoo:
> I'll close this MR, but the comments regarding "select a different project" are still relevant to the image_widget branch.

Yes, they're still something we need to think about. If canceling and going back to the previous image is not possible, what are the options? (And should we move that discussion elsewhere?)

It would also be a good idea to change the sandbox link in the image-viewer-wg channel to point to a different sandbox.

PR #630, opened 2021-05-02
Fix sort order of mentor projects

srjfoo:
Change sort order of lists of projects at the top of the for_mentors script to display projects available and waiting based on when they left the mentored round, and include the date and time that each project left the mentored round so that mentors can see for themseves when each project left the mentored round.

Existing sort is based on `projects.modifieddate`, which may or may not give the correct information on which project slice left the mentored round soonest; because of the duplicate author block, if there are a lot of different titles, slices can be released out of order. Changed sort to look at `project_events.timestamp` for the round complete event for the mentored round.

Compare [fix-mentor-sort](https://www.pgdp.org/~srjfoo/c.branch/fix-mentor-sort/tools/proofers/for_mentors.php) with the [current version](https://www.pgdp.org/c/tools/proofers/for_mentors.php).

### DIFF
--- a/tools/proofers/for_mentors.php
+++ b/tools/proofers/for_mentors.php
@@ -105,12 +105,12 @@ echo " ";
 echo sprintf(_("Please check the Saved column in the 'Which proofreader did each page...' section for each project listed below, to make sure you first work on the project with the oldest pages saved in %s."), "<b>$mentored_round->id</b>");
 echo "</p>";
 
-$projects_available = get_beginner_projects_in_state($mentoring_round->project_available_state);
+$projects_available = get_beginner_projects_in_state($mentoring_round->project_available_state, $mentored_round->id);
 if ($projects_available) {
     echo "<ol>";
     foreach ($projects_available as $proj_obj) {
         echo "<li><a href='#$proj_obj->projectid'>";
-        echo output_project_label($proj_obj->nameofwork, $proj_obj->authorsname);
+        echo output_project_label($proj_obj->nameofwork, $proj_obj->authorsname, $proj_obj->t_left_mentee_round);
         echo "</a></li>";
     }
     echo "</ol>";
@@ -125,13 +125,13 @@ echo " ";
 echo _("Oldest project listed first.");
 echo "</p>";
 
-$projects_waiting = get_beginner_projects_in_state($mentoring_round->project_waiting_state);
+$projects_waiting = get_beginner_projects_in_state($mentoring_round->project_waiting_state, $mentored_round->id);
 if ($projects_waiting) {
     echo "<ol>";
     foreach ($projects_waiting as $proj_obj) {
         $project = new Project($proj_obj->projectid);
         echo "<li>";
-        echo output_project_label($proj_obj->nameofwork, $proj_obj->authorsname);
+        echo output_project_label($proj_obj->nameofwork, $proj_obj->authorsname, $proj_obj->t_left_mentee_round);
         if (in_array($mentoring_round->project_waiting_state, $project->get_hold_states())) {
             // TRANSLATORS: string indicates that the project is "on hold"
             echo " <b>[" . _("On hold") . "]</b>";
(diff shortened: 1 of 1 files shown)


srjfoo:
The reworked script does what I want it to do, but I'm sure that the code can be improved -- please make suggestions for improvements!

In particular, though because of the limited audience it probably doesn't matter much, I'm pretty sure that the date for the lists at the top of the page is not localizable, but given the large number of approaches to displaying the date, I just picked one (I was trying to keep it simple).

Also concerning the date -- and this may need to be fixed, but I'm unsure how the time zone is done for that particular script, it should match the date as shown in the project history for the corresponding event. So, it should be server time. Server time is UTC for TEST, but US Eastern time for PROD. And I'm unsure from reading the documentation how that sorts out.

tools/proofers/for_mentors.php
```diff
@@ -150,10 +150,16 @@
 
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 
-function output_project_label($nameofwork, $authorsname)
+function output_project_label($nameofwork, $authorsname, $date = null)
 {
-    // TRANSLATORS: format is <title> by <author>.
-    echo sprintf("%1\$s by %2\$s", $nameofwork, $authorsname);
+    if (!is_null($date)) {
+        $date = sprintf("(" . ("%s"), strftime("%Y-%m-%d %R %Z", $date) . ")");
```
70ray, line 156: Could you write this as

```php
    $date = strftime("%Y-%m-%d %R %Z", $date);
    // TRANSLATORS: format is <title> by <author> (<date>).
    echo sprintf(_("%1\$s by %2\$s (%3\$s)"), $nameofwork, $authorsname, $date);
```

srjfoo: Hm ... yes, I probably should have done it that way. Thinking back, I think I was initially trying to figure out how to build the date so that I wouldn't need the conditional if the date is `null`, but decided I was better off just including it. And then forgot. Thanks!

srjfoo: Done.

tools/proofers/for_mentors.php
```diff
@@ -181,16 +187,26 @@ function output_project_details($mentored_round, $projectid, $nameofwork, $autho
...
+            projects.authorsname,
+            MAX(project_events.timestamp) AS t_left_mentee_round
+        FROM projects LEFT OUTER JOIN project_events USING (projectid)
         WHERE
-            difficulty = 'BEGINNER' AND
-            state='%s'
+            difficulty = 'BEGINNER'
+            AND project_events.event_type = 'transition'
+            AND project_events.details2 = '%s'
+            AND projects.state='%s'
```
cpeel, line 205: Consider improving the prior version and put spaces around `=`.

tools/proofers/for_mentors.php
```diff
@@ -181,16 +187,26 @@ function output_project_details($mentored_round, $projectid, $nameofwork, $autho
...
+        GROUP BY projects.projectid
         ORDER BY
-            modifieddate ASC",
+            MAX(project_events.timestamp) ASC",
```
cpeel, line 208: I've never seen a function in an order by like this so I don't know what it does. Is it necessary?

tools/proofers/for_mentors.php
cpeel: In this case I would just append the date to the same string and have one fewer strings the translators need to translate:
```php
    // TRANSLATORS: format is <title> by <author>.
    echo sprintf("%1\$s by %2\$s", $nameofwork, $authorsname);
    if (!is_null($date)) {
        $date = strftime("%Y-%m-%d %R %Z", $date);
        echo " ($date)";
    }
```

srjfoo: Thanks, that simplifies it quite a bit! One quick question, though -- doesn't the `"%1\$s by %2\$s"` need to be wrapped in `_()`, or am I missing something?

cpeel: It sure should!

tools/proofers/for_mentors.php
srjfoo: Speaking of the `MAX` being part of it? I think I tried just ordering by `project_events.timestamp`, and mysql didn't like it. I'll go back and double-check from the mysql command line, though. Maybe it was something else that was the problem at the time, and not that.

srjfoo: Hm. Didn't even need to go to the command line -- just took it out in the code, and the code still works, so it must've been something else when I was testing the query.

tools/proofers/for_mentors.php
srjfoo: Done. I also took the liberty of looking for other instances and added spaces in the couple I found.

PR #634, opened 2021-05-10
Add the ability to disable special days

srjfoo:
Until now once a special day was created and enabled so that Project Managers could assign the special day to a project, it was permanent.

This fixes it so that a special day can be disabled for assigning to projects, but can remain for projects that used it prior to it being disabled. It also allows disabled special days to be selected in the project search form, and adds an indicator to the `show_specials` list to indicate which special days have been disabled.

Sandbox here [add-disable-special-day](https://www.pgdp.org/~srjfoo/c.branch/add-disable-special-day/).


cpeel (approved):
This looks good, nicely done!

PR #637, opened 2021-05-19
User logon stats

chrismiceli:
Replaces 4 of the 5 graphs on https://www.pgdp.org/~chrismiceli/c.branch/user_logon_stats/stats/user_logon_stats.php
The bar graph will be replaced in a subsequent MR.

image
image
image
image


chrismiceli:
Note, only the last commit is new. The rest is just updating svg-graphs. Not sure the best way to do this to make the review easier.

cpeel:
> Note, only the last commit is new. The rest is just updating svg-graphs. Not sure the best way to do this to make the review easier.

Why are you updating the branch? If they are strictly required I can rebase the branch, but otherwise you should be basing your changes off the branch not off of master.

chrismiceli:
> > Note, only the last commit is new. The rest is just updating svg-graphs. Not sure the best way to do this to make the review easier.
> 
> Why are you updating the branch? If they are strictly required I can rebase the branch, but otherwise you should be basing your changes off the branch not off of master.

I think I merged in master just to reduce drift. If you want to rebase the branch I could avoid it next time to reduce your burden.

cpeel:
> > Why are you updating the branch? If they are strictly required I can rebase the branch, but otherwise you should be basing your changes off the branch not off of master.
> 
> I think I merged in master just to reduce drift. If you want to rebase the branch I could avoid it next time to reduce your burden.

I'm not worried about the branches diverging a little bit nor do I want to keep rebasing the branch. Please just rebase your changes off of svg-graph.

chrismiceli:
> 
> 
> > > Why are you updating the branch? If they are strictly required I can rebase the branch, but otherwise you should be basing your changes off the branch not off of master.
> > 
> > 
> > I think I merged in master just to reduce drift. If you want to rebase the branch I could avoid it next time to reduce your burden.
> 
> I'm not worried about the branches diverging a little bit nor do I want to keep rebasing the branch. Please just rebase your changes off of svg-graph.

Ok, done. Sorry for the disruption.

srjfoo:
I'm not much use in reviewing the js code, so my comments are strictly concerning the presentational aspects. While I think some tweaking needs to be done, it's really nice to see progress!

1. I'm assuming that the height is adjustable? As of right now, the new graphs seem to be about 50%-ish taller (I think it varies slightly from graph to graph?) than the existing ones.
2. Is "Number of users newly logged in each hour" using the same data as the corresponding [current graph](https://www.pgdp.org/c/stats/user_logon_stats.php)?  When I try to compare the two, the height of the bars doesn't seem to correspond. For example, somewhere around Dec. 20, 2020, there's a spike to 7. I don't see a corresponding spike on the new version of the graph. (The other graphs' shapes look right, but the dates look off? But the first one, the shapes of the graphs aren't even the same, I think?)
3. Do we need fractional logons on that first graph?
4. Should/can the graphs have alternating bars as the old graphs do?
5. X-axis labels -- I suspect we need to work out a way to avoid dates like Nov and Feb 31 -- consider what the reaction of our users would be! 😁 
6. X-axis labels again -- and this is an aesthetic consideration, I guess -- it looks like the old graphs have vertical labels for the x-axis so that they can be labeled every 28 days? Is this not possible with the new graphing software? (I'm actually fine with it being every 3 months, it's just a difference, so I thought I'd ask)
7. I think the graphs need some separation. For the old graphs, the vertical X-axis labels and the "graph generated at ..." line at the bottom right of each graph gives a visual breakpoint, though, even for those, it might be better to have a bit more separation. The new graphs are so close together that the graph title looks more like it belongs to the graph above it than the graph below it. Perhaps a horizontal rule between them?

chrismiceli:
> 
> 
> I'm not much use in reviewing the js code, so my comments are strictly concerning the presentational aspects. While I think some tweaking needs to be done, it's really nice to see progress!
> 
>     1. I'm assuming that the height is adjustable? As of right now, the new graphs seem to be about 50%-ish taller (I think it varies slightly from graph to graph?) than the existing ones.
> 
>     2. Is "Number of users newly logged in each hour" using the same data as the corresponding [current graph](https://www.pgdp.org/c/stats/user_logon_stats.php)?  When I try to compare the two, the height of the bars doesn't seem to correspond. For example, somewhere around Dec. 20, 2020, there's a spike to 7. I don't see a corresponding spike on the new version of the graph. (The other graphs' shapes look right, but the dates look off? But the first one, the shapes of the graphs aren't even the same, I think?)
> 
>     3. Do we need fractional logons on that first graph?
> 
>     4. Should/can the graphs have alternating bars as the old graphs do?
> 
>     5. X-axis labels -- I suspect we need to work out a way to avoid dates like Nov and Feb 31 -- consider what the reaction of our users would be! 😁
> 
>     6. X-axis labels again -- and this is an aesthetic consideration, I guess -- it looks like the old graphs have vertical labels for the x-axis so that they can be labeled every 28 days? Is this not possible with the new graphing software? (I'm actually fine with it being every 3 months, it's just a difference, so I thought I'd ask)
> 
>     7. I think the graphs need some separation. For the old graphs, the vertical X-axis labels and the "graph generated at ..." line at the bottom right of each graph gives a visual breakpoint, though, even for those, it might be better to have a bit more separation. The new graphs are so close together that the graph title looks more like it belongs to the graph above it than the graph below it. Perhaps a horizontal rule between them?

Great feedback.
1. Yes, the height was 100px too much. This actually applies to the original chart I migrated as well. I will fix to match old charts.
2. The new charts use an "area" chart, so instead of drawing a vertical rectangle for each point, it draws a polygon. That peak of 7 is too extreme it seems for the number of data points to render clearly. I'll see if I can improve it. Keep in mind there are a ton of data points in these charts (11,477 points each!). As for the incorrect data, I think I figured it out. The months are off by one. This also is why we have Feb 31 :).
3. No, that doesn't make sense. Let me see if I can correct that.
4. I can try to implement that.
5. The months are off by one. I will correct.
6. The x-axis tick count can be suggested but the library kind of figures it out for us. I can look at seeing if I can influence more appropriately.
7. I'm not opposed to a horizontal rule here. I'll add it for review.

I've pushed fixes for the height, months, and horizontal rule for early feedback.

srjfoo:
> > I'm not much use in reviewing the js code, so my comments are strictly concerning the presentational aspects. While I think some tweaking needs to be done, it's really nice to see progress!
> > 
> > 1. I'm assuming that the height is adjustable? As of right now, the new graphs seem to be about 50%-ish taller (I think it varies slightly from graph to graph?) than the existing ones.
> > 
> > 2. Is "Number of users newly logged in each hour" using the same data as the corresponding [current graph](https://www.pgdp.org/c/stats/user_logon_stats.php)?  When I try to compare the two, the height of the bars doesn't seem to correspond. For example, somewhere around Dec. 20, 2020, there's a spike to 7. I don't see a corresponding spike on the new version of the graph. (The other graphs' shapes look right, but the dates look off? But the first one, the shapes of the graphs aren't even the same, I think?)
> > 
> > 3. Do we need fractional logons on that first graph?
> > 
> > 4. Should/can the graphs have alternating bars as the old graphs do?
> > 
> > 5. X-axis labels -- I suspect we need to work out a way to avoid dates like Nov and Feb 31 -- consider what the reaction of our users would be! 😁
> > 
> > 6. X-axis labels again -- and this is an aesthetic consideration, I guess -- it looks like the old graphs have vertical labels for the x-axis so that they can be labeled every 28 days? Is this not possible with the new graphing software? (I'm actually fine with it being every 3 months, it's just a difference, so I thought I'd ask)
> > 
> > 7. I think the graphs need some separation. For the old graphs, the vertical X-axis labels and the "graph generated at ..." line at the bottom right of each graph gives a visual breakpoint, though, even for those, it might be better to have a bit more separation. The new graphs are so close together that the graph title looks more like it belongs to the graph above it than the graph below it. Perhaps a horizontal rule between them?
> > 
> 
> Great feedback.
> 
> 1. Yes, the height was 100px too much. This actually applies to the original chart I migrated as well. I will fix to match old charts.

It looks like the dimensions of the old graphs include the axis labels, whereas the new graphs do not. The graphs themselves look much better, but are still taller than the corresponding graph when compared side-by-side. This is just a comment, not a judgment on whether or not they should be.

> 2. The new charts use an "area" chart, so instead of drawing a vertical rectangle for each point, it draws a polygon. That peak of 7 is too extreme it seems for the number of data points to render clearly. I'll see if I can improve it. Keep in mind there are a ton of data points in these charts (11,477 points each!). As for the incorrect data, I think I figured it out. The months are off by one. This also is why we have Feb 31 :).

That first chart still doesn't seem to match up very well, though it may just be because the tops of the bars are so severely attenuated.

> 3. No, that doesn't make sense. Let me see if I can correct that.

👍 

> 4. I can try to implement that.

I don't think it's a must, but it does give something for the eyes to follow.

> 5. The months are off by one. I will correct.

👍 

> 6. The x-axis tick count can be suggested but the library kind of figures it out for us. I can look at seeing if I can influence more appropriately.

Thanks -- it's interesting seeing what's possible.

> 7. I'm not opposed to a horizontal rule here. I'll add it for review.
> 
> 
> I've pushed fixes for the height, months, and horizontal rule for early feedback.

The horizontal rule makes it ***much*** easier to tell what label goes where, thanks! Other comments interleaved above.

srjfoo:
I should have done a screenshot earlier -- this is what I meant when I said that the height of the new graphs **_without_** the x-axis labels is the height of the old graphs ***with*** the axis label.
image

I think I said before that I don't know how much it matters; I'm mentioning it mostly because I don't know that the data needs the extra height, so ... whatever you think is best.

stats/user_logon_stats.php
```diff
@@ -2,21 +2,26 @@
...
-    "jpgraph_files/users_logging_on.php?past=year&amp;preceding=day",
-    "jpgraph_files/users_logging_on.php?past=year&amp;preceding=week",
-    "jpgraph_files/users_logging_on.php?past=year&amp;preceding=fourweek",
-];
-
-foreach ($images as $image) {
-    echo "<img style='max-width: 100%' src='$image'><br>\n";
-}
+echo "<div style='max-width: 640px'>";
+// echo "<div id='past_day_preceding_hour' style='max-height: 400px'></div><hr>";
```
cpeel, line 22: Is this intentionally commented out?

cpeel: Instead of echo'ing these out individually, consider using an array+loop (like was there before) so that if we decide to tweak the style or `<hr>` we can do so by editing one line instead of 5.

cpeel (commented):
`chart.js` as a name is a bit ambiguous if you aren't aware of what it's doing and what it's for. Perhaps `svgCharts.js` for at least a hint at how it should be used?

A reminder to remove `stats/jpgraph_files/users_logging_on.php` from `SETUP/check_require_login.php` as well.

Overall I quite like the new style & look! 👍

pinc/graph_data.inc
```diff
@@ -48,3 +48,75 @@ function cumulative_total_proj_summary_graph()
...
+
+function user_logging_on($past, $preceding)
+{
+    // For each hour in the $past interval,
+    // show the number of (distinct) users who had logged in
+    // (at least once) during the $preceding interval.
+
+    $seconds_per_day = 24 * 60 * 60;
+
+    switch ($past) {
```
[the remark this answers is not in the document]
chrismiceli: weird, I fixed it up manually and it kept the change, but previous runs didn't make any changes either, even after clearing cache.

stats/user_logon_stats.php
chrismiceli: It was since this one chart was a bar chat. I'll uncomment and add the bar chart in.

cpeel (approved):

srjfoo (approved):

srjfoo (approved):

PR #638, opened 2021-05-20
Prevent editing word lists for archived projects

cpeel:
Word lists are stored in the project directory and thus if the directory has been deleted (because the project was archived or deleted) the project's word lists cannot be managed. Prevent the link from showing up on the project page and do a more robust job of checking for these edgecases in the edit word lists page and failing early. This fixes an error seen in `php_errors` from PROD.

This consolidates some of the validation logic in `edit_project_word_lists.php`. It's still a hot mess (which I can say with both authority and shame since I wrote it all 14 years ago).

Testable in the [fix-archived-wordlist-edits](https://www.pgdp.org/~cpeel/c.branch/fix-archived-wordlist-edits/) sandbox.

Please help validate that this behaves like we would expect during project create. It should but that's the only real problematic edgecase that comes to mind.

### DIFF
--- a/pinc/ProjectSearchResults.inc
+++ b/pinc/ProjectSearchResults.inc
@@ -315,23 +315,24 @@ class OptionsColumn extends Column
 {
     public function print_cell_data($project)
     {
-        global $pguser, $projects_url, $code_url;
+        global $pguser, $code_url;
         $projectid = $project->projectid;
-        $actions = '';
+        $actions = [];
 
         $topic_id = $project->topic_id;
         if (!empty($topic_id)) {
             $url = get_url_to_view_topic($topic_id) . "&amp;view=unread#unread";
-            $actions .= "<a href='$url' target='_blank'>" . _("View Forum") . "</a><br>\n";
+            $actions[] = "<a href='$url' target='_blank'>" . _("View Forum") . "</a>";
         }
         if ($project->user_can_do_quick_check()) {
-            $actions .= "<a href='$code_url/tools/project_manager/project_quick_check.php?projectid=$projectid' target='_blank'>"._("Project Quick Check")."</a><br>";
+            $actions[] = "<a href='$code_url/tools/project_manager/project_quick_check.php?projectid=$projectid' target='_blank'>" . _("Project Quick Check") . "</a>";
         }
         if ($project->can_be_managed_by_user($pguser)) {
-            $actions .= "<a href='$code_url/tools/project_manager/editproject.php?action=edit&amp;project=$projectid' target='_blank'>" . _("Edit Info") . "</a><br>
-            <a href='$code_url/tools/project_manager/edit_project_word_lists.php?projectid=$projectid' target='_blank'>" . _("Edit Word Lists") . "</a><br>
-            <a href='$code_url/../noncvs/send_email.php?to=db-req&amp;subject=$projectid' target='_blank'>"._("Email db-req")."</a><br>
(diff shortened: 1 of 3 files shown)


tools/project_manager/edit_project_word_lists.php
```diff
@@ -9,24 +9,22 @@
...
 include_once($relPath.'misc.inc');  // attr_safe(), html_safe()
-include_once($relPath.'faq.inc');  // attr_safe()
+include_once($relPath.'faq.inc');
 
 require_login();
 
 $return = array_get($_REQUEST, "return", "$code_url/tools/project_manager/projectmgr.php");
+$projectid = get_projectID_param($_REQUEST, 'projectid');
 
-if (!user_is_PM()) {
```
chrismiceli, line 18: This is now covered by user_can_edit_project now right?

chrismiceli (approved):

srjfoo:
Everything looks good from the project page (pre-existing archived project and newly created empty project), but the link to edit word lists should probably be removed from the links in project search results See [pinc/ProjectSearchResults.inc#L332](https://github.com/DistributedProofreaders/dproofreaders/blob/43833ad204d81ec12d705737c50a6e42d4993559/pinc/ProjectSearchResults.inc#L332)

cpeel:
> Everything looks good from the project page (pre-existing archived project and newly created empty project), but the link to edit word lists should probably be removed from the links in project search results See [pinc/ProjectSearchResults.inc#L332](https://github.com/DistributedProofreaders/dproofreaders/blob/43833ad204d81ec12d705737c50a6e42d4993559/pinc/ProjectSearchResults.inc#L332)

Ah, good catch. I've added a second commit that addresses it. The updated code removes the noncvs links which we'll add back in a PROD hack instead.

pinc/ProjectSearchResults.inc
```diff
@@ -350,14 +351,15 @@ class OptionsColumn extends Column
             if ($project->state == PROJ_POST_FIRST_UNAVAILABLE ||
                 $project->state == PROJ_POST_FIRST_AVAILABLE ||
                 $project->state == PROJ_POST_FIRST_CHECKED_OUT) {
-                $actions .= "<a href=\"$projects_url/$projectid/$projectid.zip\">" . _("Download") . "</a><br>\n";
+                $actions[] = "<a href=\"$project->url/$projectid.zip\">" . _("Download") . "</a>";
             }
             if ($project->state == PROJ_POST_SECOND_CHECKED_OUT ||
                 $project->state == PROJ_POST_COMPLETE) {
-                $actions .= "<a href=\"$projects_url/$projectid/".$projectid."_second.zip\">" . _("Download") . "</a><br>\n";
+                $actions[] = "<a href=\"$project->url/".$projectid."_second.zip\">" . _("Download") . "</a>";
```
srjfoo, line 358: *Very* picky, I know, but, spaces around the dots?

srjfoo (approved):
Looks good.

In the project search results Actions column, are the links to db-req and dp-format PROD hacks? If so, since you changed the structure of the code for the links, do those need to be updated, too?

srjfoo:
> Looks good.
> 
> In the project search results Actions column, are the links to db-req and dp-format PROD hacks? If so, since you changed the structure of the code for the links, do those need to be updated, too?

<grin> -- you beat me to it. I didn't read your comment here until after I had already made that comment!

pinc/ProjectSearchResults.inc
cpeel: Fixed by integrating the variable into the string itself.

srjfoo (approved):

PR #639, opened 2021-05-27
Update links to PG

srjfoo:
PG now redirects to `https`; this updates links within the DP code to default to `https`, and updates `crontab/import_pg_catalog.php`, `stats/requested_books.php` and `tools/post_proofers/ppv_report.php` to reference `pg.inc`. `SETUP/MediaWiki_extensions/dpExtensions.php` was also updated to use `https`.

Removed link to deprecated PG Greek transliteration FAQ from transliteration tool. Updated `faq.inc` to point to the correct page for the Greek page in the DP wiki.

Also removed several definitions within pg.inc that were used only in the FAQs in the code that now redirect to the DP Official Documentation in the wiki.

In a separate commit, deleted the two `.sit` files. The files are so old that the current version of Stuffit Expander (2021) won't open them.

Partially testable: [pg-link-updates](https://www.pgdp.org/~srjfoo/c.branch/pg-link-updates/). The two `faq/` files that were changed have redirects to the wiki and should be invisible to the users. The cron, I need instructions on testing, myself. For the rest, anywhere that a link to a PG ebook is posted should now point to `https`.


srjfoo:
Additional informational notes, in a separate comment because I don't seem to be able to be terse: The plan is, after the next code release, to remove the outdated text of all the FAQs, leaving only the redirects. For now, the only two I actually made changes to are `faq/ProoferFAQ.php` and `faq/cp.php`, because they were the only two that contained links to PG other than the ones I deemed it best to keep. However, the FAQs that are in the code, redirected to the DP Official Documentation, are generally quite out of date, and would need to be totally rewritten if moved back into the code.

Perhaps a bit over-zealously, I added notes at the top of these two, to contact a siteadmin if the FAQ in question was visible. I'm perfectly happy taking those out if it would be better to do so. In `faq/ProoferFAQ.php`, instead of just substituting in the correct donate URL, I removed the whole form, because upon reading it, and checking the history, this part of the FAQ was so old that it was pointing to PG for making donations to DP, and it didn't seem worth fixing that.

faq/ProoferFAQ.php
```diff
@@ -13,6 +12,10 @@
 <h1>Beginning Proofreaders' <br> Frequently Asked Questions (FAQ)</h1>
 <h3 align="right">Version 1.7, released May 27, 2004 </h3>
 
+<hr>
+<p class='warning'>The content for this FAQ document has been moved into the wiki and is no longer maintained in the code. If you see this message, please notify a site admin.</p>
```
chrismiceli, line 16: Maybe put the email or mechanism to contact site admin.

faq/ProoferFAQ.php
```diff
@@ -55,7 +58,7 @@
 
 <h3><a name="1">1. What is Distributed Proofreaders? </a></h3>
 Distributed Proofreaders is an effort to support <a href="
-<?php echo $PG_home_url; ?>">Project Gutenberg</a>, and a recognized affiliated site of
+<?php echo 'https://www.gutenberg.org/'; ?>">Project Gutenberg</a>, and a recognized affiliated site of
```
chrismiceli, line 61: Can we reuse PG_home_url?

faq/ProoferFAQ.php
```diff
@@ -102,7 +105,7 @@
     only legal to do this with older books (in general, copyrighted before 1923). As a
     result, Project Gutenberg is mostly comprised of the "Classics." </p>
 <p> You can read more about the history of Project Gutenberg <a
-    href="<?php echo $PG_history_url; ?>">here</a></p>
+    href="<?php echo 'https://www.gutenberg.org/about/'; ?>">here</a></p>
```
chrismiceli, line 108: same here

pinc/pg.inc
```diff
@@ -3,26 +3,16 @@
...
-$PG_html_faq_url = "https://www.gutenberg.org/wiki/Gutenberg:HTML_FAQ";
-$PG_donation_url = "https://www.gutenberg.org/wiki/Gutenberg:Project_Gutenberg_Needs_Your_Donation";
-$PG_copyright_howto_url = "https://www.gutenberg.org/wiki/Category:How-To";
-$PG_greek_howto_url = "https://www.gutenberg.org/wiki/Gutenberg:Greek_How-To";
-$PG_mailing_lists_url = "https://www.gutenberg.org/wiki/Gutenberg:Subscribe_How-To";
-
-$PG_in_progress_url = "http://www.dprice48.freeserve.co.uk/GutIP.html";
-
-$PG_paypal_business_id = "donate@gutenberg.org";
+$PG_catalog_url = "https://www.gutenberg.org/ebooks/";
```
chrismiceli, line 6: Could be redefined as ```$PG_catalog_url = $PG_home_url . "ebooks/";``` that way we only have to change one place if we needed to update again, but may be overkill.

stats/requested_books.php
```diff
@@ -72,10 +73,10 @@
 echo "<br>\n";
 echo "<h2>" . _("Most Requested Books Posted to Project Gutenberg") . "</h2>\n";
 
-$pg_url1 = DPDatabase::escape("<a href='http://www.gutenberg.org/ebooks/");
+$pg_url1 = DPDatabase::escape(sprintf("<a href='%s", get_pg_catalog_url_for_etext('')));
```
chrismiceli, line 76: do we need a migration on the table since existing entries will not have https?

pinc/faq.inc
```diff
@@ -71,8 +71,8 @@ $external_faq_overrides = [
     "request-support" => [
         "en" => "https://www.pgdp.net/wiki/DP_Official_Documentation:Main/Request_Support",
     ],
-    "transliterating-greek" => [
-        "en" => "https://www.pgdp.net/wiki/DP_Official_Documentation:Proofreading/Transliterating_Greek",
+    "greek" => [
+        "en" => "$wiki_url/Greek",
```
cpeel, line 75: We need to hard-code the value as `https://www.pgdp.net/wiki/` so that this works for TEST as well as PROD. We can't rely on `$wiki_url` for this.

stats/requested_books.php
cpeel: This string isn't stored in the DB so there's nothing that needs to be migrated.

pinc/pg.inc
```diff
@@ -3,26 +3,16 @@
 // Variables, functions, etc. (mostly URL-related) related to Project Gutenberg.
 
 $PG_home_url = "https://www.gutenberg.org/";
```
cpeel, line 5: It's more intuitive for URLs in variables to not have the training `/`. Consider:
```php
$url = "{$PG_home_url}ebooks";
$url = "$PG_home_url/ebooks";
```

I would suggest removing the trailing slash here and in `$PG_catalog_url`. It doesn't look like there would need to be any other code changes given how that variable is used -- except in `tools/project_manager/show_image_sources.php` which should be using `get_pg_catalog_url_for_etext()` anyway.

pinc/pg.inc
cpeel: `$PG_catalog_url` doesn't appear to be used anywhere now.

pinc/pg.inc
```diff
@@ -3,26 +3,16 @@
...
+$PG_catalog_url = "https://www.gutenberg.org/ebooks/";
+$PG_faq_url = "https://www.gutenberg.org/help/faq.html";
+$PGLAF_upload_url = "https://upload.pglaf.org/";
 
 // -----------------------------------------------------------------------------
 
 function get_pg_catalog_url_for_etext($etext_number)
 // Given a PG etext number, return a URL for the PG catalog page for that text.
 {
-    return "http://www.gutenberg.org/ebooks/$etext_number";
```
cpeel, line 25: Consider:
```php
{
    global $PG_home_url;
    return "$PG_home_url/ebooks/$etext_number";
}
```

faq/ProoferFAQ.php
srjfoo: Done.

pinc/pg.inc
srjfoo: Thanks for confirming that -- I suspected that it wasn't but I wasn't 100% certain.

pinc/pg.inc
srjfoo: Done, and thanks. I'd wondered if I should be doing it that way.

faq/ProoferFAQ.php
srjfoo: see above

pinc/faq.inc
srjfoo: Done.

pinc/pg.inc
srjfoo: Actually, there's also `get_pg_catalog_link_for_etext()`, which seems to work there.

crontab/import_pg_catalog.php
```diff
@@ -3,6 +3,7 @@
 include_once($relPath.'base.inc');
 include_once($relPath.'misc.inc');
 include_once($relPath.'Stopwatch.inc');
+include_once($relPath.'pg.inc');
 
 require_localhost_request(true /* deny_cli */);
```
cpeel, line 8: I feel comfortable with your changes to this file as-is, but if you _wanted_ to test it, you could do so by commenting this out and accessing the page via your web browser.

cpeel (approved):

crontab/import_pg_catalog.php
srjfoo: Thanks -- yes, I like being able to test my code changes. 🙂

PR #641, opened 2021-06-02
Added PM link to character suites

70ray:
This implements task 1989.
Is this the best wording and position in the list?

Sandbox at: https://www.pgdp.org/~rp31/c.branch/link_char_suite

It is apparent that the file  tools/project_manager/projectmgr.inc contains some redundant logic and is itself redundant in that the functions it implements are now each used in only one (different) php file and could be moved there. This could be addressed in another branch.

### DIFF
--- a/tools/project_manager/projectmgr.inc
+++ b/tools/project_manager/projectmgr.inc
@@ -38,6 +38,7 @@ function echo_manager_links()
         }
         echo "<li><a href='$code_url/tools/project_manager/project_quick_check.php'>"._("Project Quick Check")."</a></li>\n";
         echo "<li><a href='$code_url/tools/project_manager/clearance_check.php'>"._("Clearance Check")."</a></li>\n";
+        echo "<li><a href='$code_url/tools/charsuites.php'>"._("Character Suites")."</a></li>\n";
         // SITE-SPECIFIC
         echo "<li>"._("Email").": ";
         echo "<a href='$code_url/../noncvs/send_email.php?to=db-req'>db-req</a> | ";


chrismiceli (approved):
Looks good to me. The issue also mentions the wiki. Should we add a link there on the character suites page perhaps?

srjfoo (approved):
I just added Task 1781 as a related task (as it's a noncvs link, though, it would probably need to be added as a site-specific hack).

I also just added the requested link on the [Recommended custom characters list](https://www.pgdp.net/wiki/Recommended_custom_characters_list) page in the wiki.

`echo_manager_links()` is there because the PM page needs to have links to some related scripts on the PM page. They used to be displayed in centered paragraph form at the top of the page. This form is something of a compromise to avoid adding a full sidebar on a page where horizontal space is at a premium, while stylistically making the page layout more similar to other site pages.

One problem with adding links to the "PM Links" box is that it's already pretty long and adding more just makes it longer.

All of the above being said, I agree that a link to the All Character Suites page is certainly appropriate for that list of links.

70ray:
IIRC the manager links list used to be displayed also on the project page (if the user was a PM) which is why it is in an .inc file.

cpeel:
This LGTM and I'll get it merged in.

We should pull the two noncvs links out of the code and into a PROD hack at some point, but that is something that Sharon or I need to do.

srjfoo:
> This LGTM and I'll get it merged in.
> 
> We should pull the two noncvs links out of the code and into a PROD hack at some point, but that is something that Sharon or I need to do.

I'm happy to do it, but feel like I should bring up the question of some other tasks that have requested additions to either the PM links or the action links in the list of projects on the PM page. Oldest to newest:

[#1334: noncvs pm tools available via a link from the PM page](https://www.pgdp.net/c/tasks.php?action=show&task_id=1334) (we do actually now have a noncvs [Control Panel](https://www.pgdp.net/noncvs/control_panel.php).

[#1699: Add Project Retread/Skip Recommendations link to PM page](https://www.pgdp.net/c/tasks.php?action=show&task_id=1699) (related task [#1781: Add retread/skip link to PM "links"](https://www.pgdp.net/c/tasks.php?action=show&task_id=1781))

Regarding 1699/1781, my initial thought was that one was asking for a link in the list of links at the top of the PM page, and the other was asking for the link in the actions dropdown, but I'm not completely clear on that, having re-read them.

Before pulling the current noncvs links out into site-specific hacks, it might be a good idea to figure out what to do about those requests.

PR #642, opened 2021-06-08
Allow different settings for horizontal/vertical split

70ray:
Add setSplitPercent() function to imageControl for this.
imageControl, textControl, controlBar, imageWidget & textWidget
use different storage keys for h or v layout and return setup
functions to use the key to find and apply settings.
Switching h/v fires a callback to do this.
control bar names improved to compassPoint etc.

### DIFF
--- a/scripts/control_bar.js
+++ b/scripts/control_bar.js
@@ -8,7 +8,6 @@ var makeImageControl = function(imageElement) {
     const percentInput = $("<input>", {type: 'number', min: '1', max: '999', value: percent, title: texts.zoomPercent});
 
     function setZoom() {
-        imageElement.css({"vertical-align": "middle"});
         imageElement.width(10 * percent);
         imageElement.height("auto");
     }
@@ -342,7 +341,7 @@ function makeImageWidget(container, align = "C") {
         C: "center",
         R: "right"
     };
-    const imageElement = $("<img>").css("cursor", "grab");
+    const imageElement = $("<img>", {class: "middle-align"}).css("cursor", "grab");
     const imageControl = makeImageControl(imageElement);
     const controlDiv = makeControlDiv(container, imageControl.controls);
 
--- a/scripts/control_bar.js
+++ b/scripts/control_bar.js
@@ -2,24 +2,19 @@
 /* exported makeImageWidget */
 
 // Construct the image sizing controls.
-var imageControl = function(imageElement, storageKey) {
-    let imageKey = storageKey + "-image";
-    let imageData = JSON.parse(localStorage.getItem(imageKey));
-    if(!$.isPlainObject(imageData)) {
-        imageData = {zoom: 100};
-    }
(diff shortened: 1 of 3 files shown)


srjfoo:
@70ray, is there a sandbox available for this?

70ray:
Sorry forgot to mention Sandbox at: https://www.pgdp.org/~rp31/c.branch/hv_persist

70ray:
Using the Promise shouldn't make any functional difference, just hopefully makes it easier to see how the code works.

srjfoo:
> Using the Promise shouldn't make any functional difference, just hopefully makes it easier to see how the code works.

I think what I need for helping me see how the code works (by reading the code, that is, instead of just testing) is a javascript tutorial that reminds me of what JS I used to know (which wasn't much), and then carries on into the advanced stuff that you're using. That being said, I've exercised the persistence and tried to also exercise the other controls a bit, and other than image size persisting, I don't notice any change in operation, so that's good.

70ray:
@srjfoo It should be possible to have different settings for everything in horizonal and vertical layouts; so you could have the control bars at the side in horizontal mode and at the top in vertical mode, and different fonts and font sizes as well as different zoom factors.
(I often feel like you described when looking at Chris's code)

srjfoo:
> @srjfoo It should be possible to have different settings for everything in horizonal and vertical layouts; so you could have the control bars at the side in horizontal mode and at the top in vertical mode, and different fonts and font sizes as well as different zoom factors.

And according to my testing, that's what we seem to have.

> (I often feel like you described when looking at Chris's code)

I'm not alone! :grin:

srjfoo (approved):

scripts/control_bar.js
```diff
@@ -361,6 +371,12 @@ function makeImageWidget(container, storageKey, align = "C") {
     });
 
     return {
+        setup: function (storageKey) {
+            let imageWidgetKey = storageKey + "-imagewidget";
```
chrismiceli, line 375: you use const above, so we could use it here too, though this is a minor comment.

scripts/page_browse.js
```diff
@@ -76,42 +55,60 @@ var maketextControl = function(textArea, storageKey) {
...
     let wrapCheck = $("<input>", {type: 'checkbox'});
 
-    // stored value is "W" or "N", if not set textWrap will be false
-    let textWrap = ("W" === textData.textWrap);
-    wrapCheck.prop("checked", textWrap);
-    setWrap(textWrap);
-
     wrapCheck.change(function () {
-        textWrap = wrapCheck.prop("checked");
+        let textWrap = wrapCheck.prop("checked");
```
chrismiceli, line 59: same with const since we don't modify textData.

chrismiceli (approved):

scripts/control_bar.js
70ray: Done this.

PR #643, opened 2021-06-09
Graph flags per page

chrismiceli:
I started this before being tasked with solving news items. I will investigate that next @cpeel. This is just a one off bar chart. You an experience it in sandbox here:
https://www.pgdp.org/~chrismiceli/c.branch/graph_flags_per_page/tools/project_manager/show_project_wordcheck_stats.php?projectid=projectID5e23b336176b9
image

Note, this alters the bar chart in the user_logon_stats.php:
https://www.pgdp.org/~chrismiceli/c.branch/graph_flags_per_page/stats/user_logon_stats.php
It makes the xaxis vertical and moves the legend to the left which is in line with the existing jpgraph solution.


70ray:
I tried it on another project (projectID5c13a64918b30). The top graph looks ok but on the bottom graph the horizontal axis labels become very bunched up and overlapping.

srjfoo:
> projectID5c13a64918b30

The way that's handled in the old-style graphs is to only show some of the x-axis labels. See [the top graph in the main sandbox version](https://www.pgdp.org/c/tools/project_manager/show_project_wordcheck_stats.php?projectid=projectID5c13a64918b30).

Mentioning graph height relative to the old graph again for this one -- mostly because it's much taller. Any thoughts on whether it's better in that respect than the old one?

cpeel (approved):
This LGTM. I'm happy to merge it in after @srjfoo gives the 👍

srjfoo:
> This LGTM. I'm happy to merge it in after @srjfoo gives the 👍

I definitely like the way things are looking ... I was waiting to see a solution for the case @70ray brought up regarding the labels of the horzontal axis; for instance, [projectID5c13a64918b30](https://www.pgdp.org/~chrismiceli/c.branch/graph_flags_per_page/tools/project_manager/show_project_wordcheck_stats.php?projectid=projectID5c13a64918b30).

One other comment about that -- one of the pngs has a longer filename. While that particular filename pattern is unlikely, many PMs do use filenames longer than 3 digits. Can those labels default to three digit, but grow if the filename is > 3 digits?

chrismiceli:
> 
> 
> > This LGTM. I'm happy to merge it in after @srjfoo gives the 👍
> 
> I definitely like the way things are looking ... I was waiting to see a solution for the case @70ray brought up regarding the labels of the horzontal axis; for instance, [projectID5c13a64918b30](https://www.pgdp.org/~chrismiceli/c.branch/graph_flags_per_page/tools/project_manager/show_project_wordcheck_stats.php?projectid=projectID5c13a64918b30).
> 
> One other comment about that -- one of the pngs has a longer filename. While that particular filename pattern is unlikely, many PMs do use filenames longer than 3 digits. Can those labels default to three digit, but grow if the filename is > 3 digits?

I added a fix to filter out some x values if there are too many. As for the page names, if the length is too long, it is simply truncated. Probably not the best. I can look at this in another iteration. What does jpgraph currently do?

srjfoo:
> > > This LGTM. I'm happy to merge it in after @srjfoo gives the 👍
> > 
> > 
> > I definitely like the way things are looking ... I was waiting to see a solution for the case @70ray brought up regarding the labels of the horzontal axis; for instance, [projectID5c13a64918b30](https://www.pgdp.org/~chrismiceli/c.branch/graph_flags_per_page/tools/project_manager/show_project_wordcheck_stats.php?projectid=projectID5c13a64918b30).
> > One other comment about that -- one of the pngs has a longer filename. While that particular filename pattern is unlikely, many PMs do use filenames longer than 3 digits. Can those labels default to three digit, but grow if the filename is > 3 digits?
> 
> I added a fix to filter out some x values if there are too many. As for the page names, if the length is too long, it is simply truncated. Probably not the best. I can look at this in another iteration. What does jpgraph currently do?

Compare [projectID5c13a64918b30](https://www.pgdp.org/~chrismiceli/c.branch/graph_flags_per_page/tools/project_manager/show_project_wordcheck_stats.php?projectid=projectID5c13a64918b30) with [projectID5e23b336176b9](https://www.pgdp.org/~chrismiceli/c.branch/graph_flags_per_page/tools/project_manager/show_project_wordcheck_stats.php?projectid=projectID5e23b336176b9) ... um, nevermind. projectID5c13a64918b30 was one of the test projects where we were seeing what happened if the PM put unterminated inline markup in the author and title fields and it spilled over into the graphs. I've corrected it, so the text on the graph page is no longer in italics. :grin:

Looking at the jpgraph graph, it grows to fit the filename. 
image

as compared with

image

I'm good with it as-is, but we will probably need to do something about the filenames -- quite a few PMs like to number the proofing images so that they can give some indication of real page number, which can make for some rather exotic png numbers `a064a.png` is not an uncommon construct pattern.

srjfoo (approved):

70ray (approved):

PR #644, opened 2021-06-21
Fix issue referencing table before created

chrismiceli:
Noticed this when trying to setup a local dp from scratch.


cpeel:
Good catch - thank you!

PR #645, opened 2021-06-23
convert graph_pages_per_number_of_flags

chrismiceli:
Convert the 2nd bar chart on the statistics page. Example here:
https://www.pgdp.org/~chrismiceli/c.branch/graph_pages_per_number_of_flags/tools/project_manager/show_project_wordcheck_stats.php?projectid=projectID5e23ae7ff0a81


cpeel (approved):

srjfoo (approved):
The graphs look good.

I keep mentioning the greater height of the new graphs relative to that of the old ones. I'd be interested to know what everyone else thinks.

Also, where there are multiple graphs on a page, I recommend an `<hr>` between them, to separate them visually; otherwise, the title of any graph other than the top one looks closer to the graph above than to the graph it's the title of.

70ray:
On Firefox in Linux the title text is slightly clipped at the top as in the screenshot:
image

srjfoo:
> On Firefox in Linux the title text is slightly clipped \<snipped image\> ...

Thanks for the reminder to check other browser @70ray -- FF and Safari seem to be okay on my Mac, but Chrome shows the clipping, as well: 
image

chrismiceli:
> 
> 
> > On Firefox in Linux the title text is slightly clipped <snipped image> ...
> 
> Thanks for the reminder to check other browser @70ray -- FF and Safari seem to be okay on my Mac, but Chrome shows the clipping, as well (edited image to crop out superfluous material):
> image

Thanks for finding these @70ray and @srjfoo. I pushed a change to give 2 more pixels to the title. If we keep running into these, I may need to try a different approach.

srjfoo:
> Thanks for finding these @70ray and @srjfoo. I pushed a change to give 2 more pixels to the title. If we keep running into these, I may need to try a different approach.

Is it possible to specify ems instead of pixels? Mind you, I haven't a clue whether that's appropriate or not.

70ray:
Latest commit doesn't quite fix it on my system. I had a search for this sort of problem and the only solution seemed to be to draw a piece of dummy text and measure its height so you can the set the top margin and the text position appropriately.

srjfoo:
> Latest commit doesn't quite fix it on my system. I had a search for this sort of problem and the only solution seemed to be to draw a piece of dummy text and measure its height so you can the set the top margin and the text position appropriately.

Since it fixed it for me, for normal circumstances, I didn't think to try oddball fonts. But that's a good point. I decided to test it out in FF and Chrome (one has to jump through a lot of hoops to change fonts on Safari, so I didn't bother, there). Here's a screenshot of what it looks like with a relatively wild font (Zapfino). Hopefully, we don't need to take fonts like this into account, but it does rather illustrate @70ray's point : 

Chrome:

image

Firefox:

image

I found it interesting that Chrome and FF use different fonts for the axis labels.

chrismiceli:
> 
> 
> > Latest commit doesn't quite fix it on my system. I had a search for this sort of problem and the only solution seemed to be to draw a piece of dummy text and measure its height so you can the set the top margin and the text position appropriately.
> 
> Since it fixed it for me, for normal circumstances, I didn't think to try oddball fonts. But that's a good point. I decided to test it out in FF and Chrome (one has to jump through a lot of hoops to change fonts on Safari, so I didn't bother, there). Here's a screenshot of what it looks like with a relatively wild font (Zapfino). Hopefully, we don't need to take fonts like this into account, but it does rather illustrate @70ray's point :
> 
> Chrome:
> 
> image
> 
> Firefox:
> 
> image
> 
> I found it interesting that Chrome and FF use different fonts for the axis labels.

I think we'll have to cross that custom font bridge when it comes to it. I believe @70ray is right though. Based on my research as well, we'd have to measure, which I suppose we could do, but we don't have any more room anyways for the title due to the top margin, unless we wanted to start increasing the heights of our charts. I'm going to push a change to add 1 more pixel. I tested in FF for linux and it seems all basic letters are ok
image

srjfoo:
> I think we'll have to cross that custom font bridge when it comes to it. I believe @70ray is right though. Based on my research as well, we'd have to measure, which I suppose we could do, but we don't have any more room anyways for the title due to the top margin, unless we wanted to start increasing the heights of our charts. I'm going to push a change to add 1 more pixel. I tested in FF for linux and it seems all basic letters are ok
> image

The font in my example was actually something I set as my default browser font. I think we can get away with just saying "if you use a crazy font, of course it's going to look bad!", but then it occurred to me that I ought to also check font size.

It looks like you've sized that space to correspond to "a hair larger than size 16" (as shown in FF and Chrome prefs). I confirmed this by seeing what it looked like at 17 point/px (not sure at this point what the correct terminology is), which was okay, and then 18 and up, which were clipped.

Is there a way to determine what the user's default browser font size is? If not, then we could probably look at the userprefs and go with the user's font size for their proofreading font, maybe, and stick with 16 as the minimum? Dunno -- just trying to figure it out from the perspective of someone who doesn't understand the graphing software yet.

chrismiceli:
> 
> 
> > I think we'll have to cross that custom font bridge when it comes to it. I believe @70ray is right though. Based on my research as well, we'd have to measure, which I suppose we could do, but we don't have any more room anyways for the title due to the top margin, unless we wanted to start increasing the heights of our charts. I'm going to push a change to add 1 more pixel. I tested in FF for linux and it seems all basic letters are ok
> > image
> 
> The font in my example was actually something I set as my default browser font. I think we can get away with just saying "if you use a crazy font, of course it's going to look bad!", but then it occurred to me that I ought to also check font size.
> 
> It looks like you've sized that space to correspond to "a hair larger than size 16" (as shown in FF and Chrome prefs). I confirmed this by seeing what it looked like at 17 point/px (not sure at this point what the correct terminology is), which was okay, and then 18 and up, which were clipped.
> 
> Is there a way to determine what the user's default browser font size is? If not, then we could probably look at the userprefs and go with the user's font size for their proofreading font, maybe, and stick with 16 as the minimum? Dunno -- just trying to figure it out from the perspective of someone who doesn't understand the graphing software yet.

Actually, let me just explicitly set the font-size on the title and other text elements to avoid this problem altogether. Does that seem acceptable?

srjfoo:
> > > I think we'll have to cross that custom font bridge when it comes to it. I believe @70ray is right though. Based on my research as well, we'd have to measure, which I suppose we could do, but we don't have any more room anyways for the title due to the top margin, unless we wanted to start increasing the heights of our charts. I'm going to push a change to add 1 more pixel. I tested in FF for linux and it seems all basic letters are ok
> > > image
> > 
> > 
> > The font in my example was actually something I set as my default browser font. I think we can get away with just saying "if you use a crazy font, of course it's going to look bad!", but then it occurred to me that I ought to also check font size.
> > It looks like you've sized that space to correspond to "a hair larger than size 16" (as shown in FF and Chrome prefs). I confirmed this by seeing what it looked like at 17 point/px (not sure at this point what the correct terminology is), which was okay, and then 18 and up, which were clipped.
> > Is there a way to determine what the user's default browser font size is? If not, then we could probably look at the userprefs and go with the user's font size for their proofreading font, maybe, and stick with 16 as the minimum? Dunno -- just trying to figure it out from the perspective of someone who doesn't understand the graphing software yet.
> 
> Actually, let me just explicitly set the font-size on the title and other text elements to avoid this problem altogether. Does that seem acceptable?

Absolutely -- but I didn't think about suggesting it until we were on our way to the grocery store. Good call!

cpeel:
What are the next steps on this branch? Not sure if this is waiting on additional code updates or if this is ready -- apologies for not poking sooner.

Since the last merge it looks like it needs a rebase to resolve merge conflicts.

chrismiceli:
> 
> 
> What are the next steps on this branch? Not sure if this is waiting on additional code updates or if this is ready -- apologies for not poking sooner.
> 
> Since the last merge it looks like it needs a rebase to resolve merge conflicts.

I will fix the conflict and we can merge afterwards. It is a simple chart for word list statistics that also fixes the title truncation issues Ray found.

chrismiceli:
> 
> 
> > What are the next steps on this branch? Not sure if this is waiting on additional code updates or if this is ready -- apologies for not poking sooner.
> > Since the last merge it looks like it needs a rebase to resolve merge conflicts.
> 
> I will fix the conflict and we can merge afterwards. It is a simple chart for word list statistics that also fixes the title truncation issues Ray found.

Resolved merge conflict and ready for merge. Sandbox updated as well:
https://www.pgdp.org/~chrismiceli/c.branch/graph_pages_per_number_of_flags/tools/project_manager/show_project_wordcheck_stats.php?projectid=projectID5e23ae7ff0a81

PR #646, opened 2021-06-30
Convert round_backlog to an iframed full html page

chrismiceli:
https://www.pgdp.org/~chrismiceli/c.branch/round_backlog/tools/proofers/round.php?round_id=P2

image

Note, this requires manually updating news items that reference the round_days.php img.

### DIFF
--- a/stats/jpgraph_files/round_backlog.php
+++ b/stats/jpgraph_files/round_backlog.php
@@ -2,9 +2,9 @@
 $relPath = "./../../pinc/";
 include_once($relPath.'base.inc');
 include_once($relPath.'stages.inc');
-include_once('common.inc');
 include_once($relPath.'graph_data.inc');
 include_once($relPath.'slim_header.inc');
+include_once('common.inc');
 
 
 // This image shows the total number of pages remaining in each round.
@@ -59,7 +59,7 @@ $x_title = _("Help is most needed in the red rounds");
 
 // If this is a new system there won't be any stats so don't divide by zero
 if ($stats_total == 0) {
-  $title = _("No pages found.");
+    $title = _("No pages found.");
 }
 
 $js_data = '$(function(){barChart("round_backlog",' . json_encode([
@@ -75,16 +75,15 @@ $js_data = '$(function(){barChart("round_backlog",' . json_encode([
     "width" => $width,
     "height" => $height,
     "barBorder" => true,
-    "bottomLegend" => $x_title
+    "bottomLegend" => $x_title,
 ]) . ');});';
 
 
 slim_header($title, [
-  "body_attributes" => "style='margin: 0'",
-  "js_files" => get_graph_js_files(),
-  "js_data" => $js_data,
+    "body_attributes" => "style='margin: 0'",
+    "js_files" => get_graph_js_files(),
+    "js_data" => $js_data,
 
 ]);
 
 echo "<div id='round_backlog' style='width:" . $width . "px;height:" . $height . "px;'></div>";
-
--- a/stats/jpgraph_files/round_backlog.php
+++ b/stats/jpgraph_files/round_backlog.php
@@ -3,6 +3,9 @@ $relPath = "./../../pinc/";
 include_once($relPath.'base.inc');
 include_once($relPath.'stages.inc');
 include_once('common.inc');
+include_once($relPath.'graph_data.inc');
+include_once($relPath.'slim_header.inc');
+
 
 // This image shows the total number of pages remaining in each round.
 // The red bars show the rounds that have more than the average number
@@ -13,12 +16,8 @@ include_once('common.inc');
 //
 // see also: http://www.pgdp.net/wiki/Round_backlog_graphs
 
-// Start with creating the Graph, this enables the use of the cache
-// where possisble
 $width = 300;
 $height = 200;
-$cache_timeout = 59; // in minutes
-$graph = new Graph($width, $height, get_image_cache_filename(), $cache_timeout);
 
 // Pull all interested phases, primarily all the rounds and PP
 $interested_phases = array_keys($Round_for_round_id_);
@@ -29,11 +28,6 @@ $stats = get_round_backlog_stats($interested_phases);
 
 // get the total of all phases
 $stats_total = array_sum($stats);
-
-// If this is a new system there won't be any stats so don't divide by zero
-if ($stats_total == 0) {
-    dpgraph_error(_("No pages found."), $width, $height);
-}
 
 // calculate the goal percent as 100 / number_of_phases
 $goal_percent = ceil(100 / count($stats));
@@ -63,46 +57,34 @@ $title = _("Pages remaining in Rounds");
 //              Try to keep the translation the same length as the English text.
 $x_title = _("Help is most needed in the red rounds");
 
-// Why not scale from zero to max value?
-$graph->SetScale("textlin", 0, max($datay) * 1.1);
-$graph->graph_theme = null;
-$graph->img->SetAntiAliasing();
+// If this is a new system there won't be any stats so don't divide by zero
+if ($stats_total == 0) {
+  $title = _("No pages found.");
+}
 
-// Set background to white
-$graph->SetMarginColor('white');
+$js_data = '$(function(){barChart("round_backlog",' . json_encode([
+    "title" => $title,
+    "axisLeft" => true,
+    "barColors" => $barColors,
+    "data" => [
+        $x_title => [
+            "x" => $datax,
+            "y" => $datay,
+        ],
+    ],
+    "width" => $width,
+    "height" => $height,
+    "barBorder" => true,
+    "bottomLegend" => $x_title
+]) . ');});';
 
-// Add a drop shadow
-$graph->SetShadow();
 
-// Adjust the margin a bit to make more room for titles
-// left, right, top, bottom
-$graph->img->SetMargin(50, 20, 30, 60);
+slim_header($title, [
+  "body_attributes" => "style='margin: 0'",
+  "js_files" => get_graph_js_files(),
+  "js_data" => $js_data,
 
-// Set title
-$graph->title->Set($title);
-$graph->title->SetFont($jpgraph_FF, FS_BOLD, 10);
+]);
 
-// Set X axis
-$graph->xaxis->SetTickLabels($datax);
-$graph->xaxis->title->Set($x_title);
-$graph->xaxis->title->SetFont($jpgraph_FF, $jpgraph_FS);
-$graph->xaxis->title->SetMargin(10);
-$graph->xaxis->HideTicks();
+echo "<div id='round_backlog' style='width:" . $width . "px;height:" . $height . "px;'></div>";
 
-// Set Y axis
-$graph->yaxis->HideTicks();
-
-// Create a bar plot
-$plot = new BarPlot($datay);
-$plot->SetFillColor($barColors);
-$graph->Add($plot);
-
-// Add display of values at top of bars
-$plot->value->Show();
-$plot->value->SetColor("black");
-$plot->value->SetFont($jpgraph_FF, $jpgraph_FS, 8);
-$plot->value->SetAngle(0);
-$plot->value->SetFormat("%d");
-
-// Display the graph
-$graph->Stroke();
--- a/stats/jpgraph_files/round_backlog.php
+++ b/stats/jpgraph_files/round_backlog.php
@@ -76,6 +76,7 @@ $js_data = '$(function(){barChart("round_backlog",' . json_encode([
     "height" => $height,
     "barBorder" => true,
     "bottomLegend" => $x_title,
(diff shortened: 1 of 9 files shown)


chrismiceli:
I see the same title issue @70ray noticed on the the other MR. Those later commits will resolve the issue here as well.

styles/layout.less
```diff
@@ -528,6 +528,11 @@ div.newsitem {
     img {
         max-width: 100%;
     }
+    iframe {
+      width: 320px;
+      height: 205px;
+      border: 0;
+    }
```
srjfoo, line 535: I've unarchived the news item for the PP page that has graphs in it. Not all of the graphs are the same size, and in the case of the PP one (and the AH one on PROD), the second graph really shouldn't be the same size as the other one, so if this has to be specified here, it should probably have some specificity added..

image

Based on other places where there are multiple graphs, there should probably be some king of visual separation, also.

stats/jpgraph_files/round_backlog.php
```diff
@@ -3,6 +3,9 @@
 include_once($relPath.'base.inc');
 include_once($relPath.'stages.inc');
 include_once('common.inc');
+include_once($relPath.'graph_data.inc');
```
cpeel, line 6: Place these before the `common.inc` include.

stats/jpgraph_files/round_backlog.php
```diff
@@ -63,46 +57,34 @@
...
-$plot->value->SetColor("black");
-$plot->value->SetFont($jpgraph_FF, $jpgraph_FS, 8);
-$plot->value->SetAngle(0);
-$plot->value->SetFormat("%d");
-
-// Display the graph
-$graph->Stroke();
+// If this is a new system there won't be any stats so don't divide by zero
+if ($stats_total == 0) {
+  $title = _("No pages found.");
```
cpeel, line 62: This should have 4 spaces -- did the formatter not catch this? Here and below in the `slim_header()` call.

styles/layout.less
cpeel: I think this is something that the news item needs to define based on the graph they want to include. I wouldn't add it to the global CSS.

stats/jpgraph_files/round_backlog.php
chrismiceli: It can't catch it if I don't run it :/. I thought the code formatter checked in the checks on github, but not for this branch since it forked before all that work was finished.

chrismiceli:
Just pushed round_backlog_days changes too which you can see in the PP news item:
https://www.pgdp.org/~chrismiceli/c.branch/round_backlog/tools/pool.php?pool_id=PP

srjfoo:
> Just pushed round_backlog_days changes too which you can see in the PP news item:
> https://www.pgdp.org/~chrismiceli/c.branch/round_backlog/tools/pool.php?pool_id=PP

I was initially confused, because the new graph doesn't match either one of the existing ones -- and finally realized that the TEST news item graphs don't necessarily match the PROD ones in where they're located. It probably would be a good idea to align them for testing purposes? (I can do that if you'd like)

The two so far, I feel like the y-axis labels are awfully cramped, which looks unbalanced to me. In the case of the jpgraphs, there are fewer labels, and the graphs have alternating horizontal bars for contrast, plus the bar, itself, is labeled with the graphed number. Are either of those latter two possible?

chrismiceli:
> 
> 
> > Just pushed round_backlog_days changes too which you can see in the PP news item:
> > https://www.pgdp.org/~chrismiceli/c.branch/round_backlog/tools/pool.php?pool_id=PP
> 
> I was initially confused, because the new graph doesn't match either one of the existing ones -- and finally realized that the TEST news item graphs don't necessarily match the PROD ones in where they're located. It probably would be a good idea to align them for testing purposes? (I can do that if you'd like)
> 
> The two so far, I feel like the y-axis labels are awfully cramped, which looks unbalanced to me. In the case of the jpgraphs, there are fewer labels, and the graphs have alternating horizontal bars for contrast, plus the bar, itself, is labeled with the graphed number. Are either of those latter two possible?

Yes, feel free to update the news items as you wish. Just make sure the relative url to the iframe pages point to my branch or else you'll get the old img versions.

As for enhancing the charts, let me see if I can reduce # of entries on y axis as well as add totals above the bars for these charts. It seems unique to these bar charts as opposed to bar chart here for instance: https://www.pgdp.net/c/stats/user_logon_stats.php

chrismiceli:
> 
> 
> > > Just pushed round_backlog_days changes too which you can see in the PP news item:
> > > https://www.pgdp.org/~chrismiceli/c.branch/round_backlog/tools/pool.php?pool_id=PP
> > 
> > 
> > I was initially confused, because the new graph doesn't match either one of the existing ones -- and finally realized that the TEST news item graphs don't necessarily match the PROD ones in where they're located. It probably would be a good idea to align them for testing purposes? (I can do that if you'd like)
> > The two so far, I feel like the y-axis labels are awfully cramped, which looks unbalanced to me. In the case of the jpgraphs, there are fewer labels, and the graphs have alternating horizontal bars for contrast, plus the bar, itself, is labeled with the graphed number. Are either of those latter two possible?
> 
> Yes, feel free to update the news items as you wish. Just make sure the relative url to the iframe pages point to my branch or else you'll get the old img versions.
> 
> As for enhancing the charts, let me see if I can reduce # of entries on y axis as well as add totals above the bars for these charts. It seems unique to these bar charts as opposed to bar chart here for instance: https://www.pgdp.net/c/stats/user_logon_stats.php

Ok, I went ahead and limited the yaxis to be at most 5 labels. This roughly lines up with prod. I think adding the number on the top is a little strange to be honest. It isn't present on the other bar charts, and can lead to weird rendering issues like the 444 in the current prod round backlog days chart: https://www.pgdp.net/c/stats/jpgraph_files/round_backlog_days.php

srjfoo:
\<snip\>
> > Yes, feel free to update the news items as you wish. Just make sure the relative url to the iframe pages point to my branch or else you'll get the old img versions.
> > As for enhancing the charts, let me see if I can reduce # of entries on y axis as well as add totals above the bars for these charts. It seems unique to these bar charts as opposed to bar chart here for instance: https://www.pgdp.net/c/stats/user_logon_stats.php
> 
> Ok, I went ahead and limited the yaxis to be at most 5 labels. This roughly lines up with prod. I think adding the number on the top is a little strange to be honest. It isn't present on the other bar charts, and can lead to weird rendering issues like the 444 in the current prod round backlog days chart: https://www.pgdp.net/c/stats/jpgraph_files/round_backlog_days.php

I like the y-axis labels on those quite a lot better, thanks!

As for the labels on the bars for `round_backlog.php` and `round_backlog_days.php` ...
* yes, there's apparently no way to put the text layer above the graph border layer in jpgraph; I'm pretty sure the attempt was made to do so.
* For [user_logon_stats.php](https://www.pgdp.net/c/stats/user_logon_stats.php), I suspect that the bar label isn't much needed, given the volume and that it's relatively easy to guess what number each bar represents. For `round_backlog.php` and `round_backlog_days.php`, there's no way to guess the exact number by projecting the height of the bar on the y-axis.
 
So the question then becomes, "Does the exact number of pages or days matter?" -- I think that's a question that the GM needs to answer. I don't have any strong feelings one way or another, but I suspect that the absence will be noticed, at least by some.

chrismiceli:
> <snip>

> So the question then becomes, "Does the exact number of pages or days matter?" -- I think that's a question that the GM needs to answer. I don't have any strong feelings one way or another, but I suspect that the absence will be noticed, at least by some.

I think I can add a tooltip on hover. I could even make it work on the other bar charts. Do you think this would be sufficient?

lhamilton1:
I think the "on hover" would do fine to show the numbers. We definitely need the numbers since DPers are so detail oriented but the on hover should cover it. It wouldn't be viewable from an iPad but most other platforms would show it.

srjfoo:
> I think the "on hover" would do fine to show the numbers. We definitely need the numbers since DPers are so detail oriented but the on hover should cover it. It wouldn't be viewable from an iPad but most other platforms would show it.

I like the tooltip idea, too.

chrismiceli:
> 
> 
> > I think the "on hover" would do fine to show the numbers. We definitely need the numbers since DPers are so detail oriented but the on hover should cover it. It wouldn't be viewable from an iPad but most other platforms would show it.
> 
> I like the tooltip idea, too.

I just did a push for basic tooltips. Let me know what you think. Sandbox updated.

srjfoo:
> > > I think the "on hover" would do fine to show the numbers. We definitely need the numbers since DPers are so detail oriented but the on hover should cover it. It wouldn't be viewable from an iPad but most other platforms would show it.
> > 
> > 
> > I like the tooltip idea, too.
> 
> I just did a push for basic tooltips. Let me know what you think. Sandbox updated.

The tooltips look fine to me, both light and dark themes (except that for 0 days to finish all pages, since there's not a bar to hover over, you presumably just have to take it on faith that there's nothing). However, there's something really weird going on that I don't understand. At least on Firefox, Chrome and Safari on a Mac. When I first open the page, both of the new graphs show with both horizontal and vertical scrollbars, thus:
image

As soon as I hover over a bar on a graph, the scrollbars go away for that graph, but if I refresh the page, they come back.

Also, if you look at the label for the graph at the top, we're back to the discussion of it getting clipped under some circumstances. On FF, it's flush with the top edge -- or possibly there's a pixel to spare. On Chrome, I must have my default font size set slightly larger, because it's clipped at the top. On Safari, I just roll my eyes because I have absolutely no control over fonts in Safari.

image

A second screenshot showing what it looks like in FF after I've hovered over a bar on each graph. The other purpose for this screenshot is to give an idea of what two graphs side-by-side might look like.

Regarding the graph title and caption text -- notice in the jpgraphs, the graph title may be slightly larger than the axis label text, but it's not as large as the text of the news item (of course, it's a different font, too) -- **_AND_** it's bolded. The caption isn't bold, but it's still notiiceably smaller than the body text in the news item.

It looks like there's some automatic padding at the bottom of the graphs, but not at the top.

Translation of all my blather ... I think that the title and caption text needs to be smaller, and the title needs to be bold. Not sure whether we need to force padding, but we probably need to be paying attention to how the graphs fit in.

lhamilton1:
I tried the tooltip on the iPad by simply putting my finger on any bar and it showed the number the bar represented.

chrismiceli:
I fixed scrollbars. I say if everyone agrees, merge this and the other open MR for charts, and we can tweak title size and padding in next MR?

srjfoo:
> I fixed scrollbars. I say if everyone agrees, merge this and the other open MR for charts, and we can tweak title size and padding in next MR?

(And graph height, discussion of, at some point.)

In the Edit interface for news items, the new graphs still have scrollbars -- is that expected behavior? It's only visible to news editors, so not a big deal, but I thought I'd mention it.

Dealing with title size and padding in a separate MR sounds like the right approach, since it will need to be applied to all graphs, uniformly, I think.

---

This is totally unrelated, but, looking at the news item graphs just now, "Days to finish all pages in Rounds" is showing no bars at all. I don't think it's a problem with your graph (the jpgraph version shows the same), and I haven't looked at the algorithm that's used to calculate those numbers, so I don't know if there's a problem with it that we need to fix. Any ideas?

chrismiceli:
> 
> 
> > I fixed scrollbars. I say if everyone agrees, merge this and the other open MR for charts, and we can tweak title size and padding in next MR?
> 
> (And graph height, discussion of, at some point.)
> 
> In the Edit interface for news items, the new graphs still have scrollbars -- is that expected behavior? It's only visible to news editors, so not a big deal, but I thought I'd mention it.
> 
> Dealing with title size and padding in a separate MR sounds like the right approach, since it will need to be applied to all graphs, uniformly, I think.
> 
> This is totally unrelated, but, looking at the news item graphs just now, "Days to finish all pages in Rounds" is showing no bars at all. I don't think it's a problem with your graph (the jpgraph version shows the same), and I haven't looked at the algorithm that's used to calculate those numbers, so I don't know if there's a problem with it that we need to fix. Any ideas?

This is because the style that removes the border is for the selector ```div.newsitem iframe``` but when the news item is shown in the edit mode it is ```div.newsedit iframe```. this should apply to the other similar styles like https://github.com/DistributedProofreaders/dproofreaders/blob/master/styles/layout.less#L518

On the unrelated note, if the data matches between my sandbox and test, there should be nothing to worry about, though it is weird it changes so much so recently.

In any case, I am not anticipating making any more changes to this MR unless there are more comments.

srjfoo:
> > In the Edit interface for news items, the new graphs still have scrollbars -- is that expected behavior? It's only visible to news editors, so not a big deal, but I thought I'd mention it.
> 
> This is because the style that removes the border is for the selector `div.newsitem iframe` but when the news item is shown in the edit mode it is `div.newsedit iframe`. this should apply to the other similar styles like https://github.com/DistributedProofreaders/dproofreaders/blob/master/styles/layout.less#L518

Thanks for the clarification!

PR #647, opened 2021-07-01
Rem proof note

70ray:
This first checks that proofers' notes are properly closed then removes them before format checking. This allows multi-line notes as in task 1977 and avoids various complications which can arise such as tags inside proofers' notes and notes inside markup.


70ray:
Sandbox at: https://www.pgdp.org/~rp31/c.branch/rem_proof_note.

70ray:
Still doesn't handle latest https://www.pgdp.net/phpBB3/viewtopic.php?p=1250240#p1250240.

70ray:
Modified to handle nested pairs of brackest inside proofers note.

PR #651, opened 2021-07-29
Use image widget in proofreading interfaces

70ray:
Both standard and enhanced.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/enh_proof_widget


srjfoo (commented):
I've taken a look in both proofreading interfaces, moved the control bar around, played with the zoom a bit, and checked out WC and Preview. It all looks good!

Is there anything else I should be looking at?

faq/prooffacehelp.php
```diff
@@ -491,7 +491,6 @@ function echo_row($name, $tooltip, $button_image_base, $accelerator)
     echo_row('Show All Text', 'Show All Text', 'bt9', '');
     echo_row('Undo Revert', 'Undo Revert', 'bt7', '');
     echo_row('Revert to Original Document', 'Revert to Original Document', 'bt8', '');
-    echo_row('Refresh Image', 'Refresh Image', 'bt6', '');
     echo_row('Set Image Zoom Percent', 'Set Image Zoom Percent', 'bt10', ''); ?>
```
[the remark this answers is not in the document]
cpeel: I would defer updating `prooffacehelp`. I'm dubious how useful that is anyway and it's unclear how we would offer links to it within the Image Viewer Widget.

We do need to pull out the image zoom from userprefs, the initial DB schema, and a DB migration script as part of this. I can help with that if we want to do merge this into a temporary branch instead of `master`.

cpeel:
In the enhanced interface, the Image Widget doesn't seem to take up the full pane at some sizes. This is on Chrome on MacOS.
Screen Shot 2021-07-29 at 4 30 07 PM

srjfoo:
> In the enhanced interface, the Image Widget doesn't seem to take up the full pane at some sizes. This is on Chrome on MacOS.
> Screen Shot 2021-07-29 at 4 30 07 PM

For what it's worth, Firefox on macOS: 
image

The text pane has the same problem in the Enhanced proofing interface, though neither does in WC: 
image

Maybe the problem isn't with the image pane, but with the way the text block is displayed within the text page?

faq/prooffacehelp.php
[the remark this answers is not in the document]
srjfoo: > I would defer updating `prooffacehelp`. I'm dubious how useful that is anyway and it's unclear how we would offer links to it within the Image Viewer Widget.

I think that the image-related stuff that will no longer be available does need to be pulled out, because those buttons/links won't be in the interfaces anymore. Other than that, I agree -- I think at a minimum, we can wait.

Even if we do decide to update `prooffacehelp` to describe the Image Viewer Widget, though, I don't think we'd need to provide links to it within the widget. As far as I know, the only place on a proofing page that links to it is the `Help` link in the documentation section of the Toolbox. All that would be needed is updating the descriptions of the functionality. But I agree that I don't know how useful it would be. All of the buttons in the widget have tooltips, and are pretty self-explanatory.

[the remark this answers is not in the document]
cpeel: I was unclear, sorry about that. I agree we should remove references to the old controls but do not think we should add docs on new controls.

tools/proofers/image_block_enh.inc
```diff
@@ -47,143 +29,38 @@ function ibe_get_styles()
...
-                top:0px;
-                z-index:1;
-            }
+                #controlframe {
+                    position:absolute;
+                    left:$textLeft;
+                    top:$textTop;
+                    width:$textWidth%;
+                    height:$textHeight%;
+                    /* clip:rect(0px, 100%, 100%, 0px); -- broken in MS Edge */
```
chrismiceli, line 41: maybe just remove if we don't want it.

tools/proofers/image_frame_std.php
```diff
@@ -3,28 +3,37 @@
...
+
+$header_args = [
+    "js_files" => $js_files,
+    "js_data" => get_control_bar_texts() . "
+            var imageData = $image_data;
+        ",
+    "body_attributes" => 'id="standard_interface_image"',
+];
+
+slim_header("Image Frame", $header_args);
```
chrismiceli, line 35: i18n

tools/proofers/proof_frame_enh.inc
```diff
@@ -26,8 +36,12 @@ function echo_proof_frame_enh($ppage)
...
             "$code_url/tools/proofers/process_diacritcal_markup.js",
+            "$code_url/scripts/control_bar.js",
+            "$code_url/tools/proofers/proof_image.js",
         ],
-        "js_data" => ibe_get_js() . get_preview_js_vars() . get_preview_font_data_js() . "
+        "js_data" => get_preview_js_vars() . get_preview_font_data_js() .
+        get_control_bar_texts() . "
+        var imageData = $image_data;
         function ldAll() {
             top.initializeStuff(1);
```
chrismiceli, line 46: already present, but could probably use a more informative name :).

faq/prooffacehelp.php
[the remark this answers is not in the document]
70ray: > `Set Image Zoom Percent` needs to go, too. -- that icon and function no longer exist where they are.
> 
I've removed this and also the texts that go with the removed lines.

> I believe that the corresponding image manipulation function for the standard interface needs to come out, too (approximagely lines 426-438).
> 
Done this also.

> `<tb>`
> 
> Related, but not in this MR -- should the image zoom come out of the userprefs, or is that still used to set initial zoom settings for initial sessions in a new project? A really quick test (not exhaustive) seems to indicate that it goes by the existing local setting, not by what's in the userprefs, so should probably come out of the userprefs?

The setting is not used in the image widget but it is still used in md_phase2.php. I don't know what to do about it. Possible options:
1. enable metadata temporarily and consider whether to use the image widget in md_phase2.php.
2. remove all the metadata code.
3. something else.

70ray:
> In the enhanced interface, the Image Widget doesn't seem to take up the full pane at some sizes. This is on Chrome on MacOS.
> Screen Shot 2021-07-29 at 4 30 07 PM

The proportions of the panes are similar in the old interface but it looks odd here. I've added a border to the pane which makes it look more uniform with the text pane.

tools/proofers/image_block_enh.inc
70ray: Done this.

tools/proofers/image_frame_std.php
70ray: OK in principle but this is in a frame so won't ever be seen and we want to replace it all anyway soon.

tools/proofers/proof_frame_enh.inc
70ray: Good idea. But unless we dig into it to discover exactly what it does any new name could be misleading.

chrismiceli: absolutely

tools/proofers/image_frame_std.php
chrismiceli: Expected as such. Thanks

chrismiceli (approved):

faq/prooffacehelp.php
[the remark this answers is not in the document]
cpeel: In a perfect world we would do option 1. However, I have no confidence that the metadata code actually works and I don't think it's worthwhile to spend time figuring out if it does just to do that option.

My suggestion is that we update `md_phase2.php` to set the initial zoom (`$iWidth`) to the default values for those fields and be done with it.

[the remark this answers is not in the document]
srjfoo: My impression was that the `md_phase2`-related code was incomplete, and certainly hasn't been maintained over the years. -- I like @cpeel's suggestion.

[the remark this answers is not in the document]
70ray: Done this.
Should we also remove the database entries?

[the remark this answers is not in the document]
chrismiceli: I would think it would be nice to clear up database, but that's my not very strong opinion.

tools/proofers/image_block_enh.inc
```diff
@@ -47,143 +29,38 @@ function ibe_get_styles()
...
-                text-align:center;
-                overflow:auto;
-            }
-            #imagedisplay {
-                position:absolute;
-                left:0px;
-                top:0px;
-                z-index:1;
-            }
+                #controlframe {
```
srjfoo, line 35: This is very picky, and I don't know that I have a strong opinion one way or the other, but is using "frame" in the id (here and for `#newimageframe`) appropriate since we're moving away from coding using frames?

srjfoo (commented):
For some reason, I don't seem to be able to quote some of the comments, so, constructing the quotes myself:

> @70ray: Should we also remove the database entries?

> @chrismiceli: I would think it would be nice to clear up database, but that's my not very strong opinion.

In https://github.com/DistributedProofreaders/dproofreaders/pull/651#discussion_r679546329 @cpeel said

> We do need to pull out the image zoom from userprefs, the initial DB schema, and a DB migration script as part of this. I can help with that if we want to do merge this into a temporary branch instead of master.

If that's the database cleanup you mean, then yes. (If there's another db cleanup you mean, I'm confused -- can you clarify?)

faq/prooffacehelp.php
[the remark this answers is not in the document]
cpeel: Yes, we need to update `db_schema.sql` and have an upgrade script that drops the columns from the DB.

tools/proofers/image_block_enh.inc
70ray: You are right. These aren't frames but divs.

faq/prooffacehelp.php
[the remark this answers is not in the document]
70ray: I've updated `db_schema.sql` and made an upgrade script by modifying existing models.

SETUP/upgrade/16/20210802_clean_user_profiles.php
```diff
@@ -0,0 +1,22 @@
+<?php
```
cpeel, line 1: Please rename this script `20210802_alter_user_profiles.php` since we're altering the table and not removing rows from it. The script's contents look good to me.

srjfoo (approved):
The testing I've done looks good (I leave the fine-tuning of the code to y'all).

Do we have enough time before the next PROD rollout to get enough public testing done before rolling it out to PROD, or do we want to give it a good long time for testing and wait until September?

cpeel:
> Do we have enough time before the next PROD rollout to get enough public testing done before rolling it out to PROD, or do we want to give it a good long time for testing and wait until September?

I would strongly suggest we do not try to rush this in the next week and rather get it in immediately after and wait until September's release to give lots of testing / complaining time.

srjfoo:
> > Do we have enough time before the next PROD rollout to get enough public testing done before rolling it out to PROD, or do we want to give it a good long time for testing and wait until September?
> 
> I would strongly suggest we do not try to rush this in the next week and rather get it in immediately after and wait until September's release to give lots of testing / complaining time.

That's my strong preference, too, thanks.

SETUP/upgrade/16/20210802_clean_user_profiles.php
70ray: Done this.

cpeel:
It looks like there are still two references to these variables in `faq/pophelp/prefs/prefs_pophelp.inc` we need to remove.

70ray:
> It looks like there are still two references to these variables in `faq/pophelp/prefs/prefs_pophelp.inc` we need to remove.

Thanks, I've removed them now.

cpeel:
This looks good to me. I will wait until after the August deployment to PROD on Tuesday before merging this in. That way we can give people a month to play with it on TEST before we roll it out.

PR #652, opened 2021-07-30
Don't include URLs in credits for PG uploads

srjfoo:
GM request; PG has asked that URLs (including email addresses) not be included in uploads to PG. Investigation showed that some of the image sources' credit lines included links. This MR updates the edit form for PMs to edit the project pages, and the Image Sources management form and display page to explicitly emphasize that the Credits line should not include URLs. Sandbox: [image-credits-no-link](https://www.pgdp.org/~srjfoo/c.branch/image-credits-no-link/).

### DIFF
--- a/tools/project_manager/editproject.php
+++ b/tools/project_manager/editproject.php
@@ -454,12 +454,11 @@ class ProjectInfoHolder
                 $errors .= sprintf(
                     _("Posted Number \"%s\" is not of the correct format."),
                     html_safe($this->postednum)) . "<br>";
-                // You'll sometimes see PG etext numbers with a 'C' appended.
-                // The 'C' is not part of the etext number
-                // (e.g., it does not appear in PG's RDF catalog),
-                // rather it's a bit of information about the identified text,
-                // namely that it's still under (US) copyright.
-                // Anyhow, the 'C' should not be included here.
+                // Occasionally, there will be a PG ebook that is still
+                // under U.S. copyright. This is indicated in their system
+                // by appending a 'C' to the etext number. The link to
+                // the etext, however, does not include the 'C', nor should
+                // the DP link. If this changes, update the pattern here.
             }
         }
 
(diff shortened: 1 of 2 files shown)


tools/project_manager/editproject.php
```diff
@@ -454,12 +454,6 @@ public function set_from_post()
                 $errors .= sprintf(
                     _("Posted Number \"%s\" is not of the correct format."),
                     html_safe($this->postednum)) . "<br>";
-                // You'll sometimes see PG etext numbers with a 'C' appended.
-                // The 'C' is not part of the etext number
-                // (e.g., it does not appear in PG's RDF catalog),
-                // rather it's a bit of information about the identified text,
-                // namely that it's still under (US) copyright.
-                // Anyhow, the 'C' should not be included here.
```
srjfoo, line 462: Just a note regarding this comment: I removed it because it's irrelevant as far as the code is concerned. It's the sort of thing that the squirrel or squirrels marking posted need to be aware of, but since the code allows only numeric postednums, it doesn't really belong here.

If there's a strong preference for putting it back in, I will -- I just don't think it's needed.

cpeel (approved):

chrismiceli (approved):

tools/project_manager/editproject.php
jmdyck: The fact that the code allows only numeric postednums doesn't make the comment irrelevant. Rather, the comment explains why it's valid for the code to impose that restriction, despite the existence of PG etext numbers that appear to violate the restriction.

I think what happened was, someone asked me to modify the code + db to allow for the 'C'. I found that it wasn't necessary, so I added the comment to inform any future dev that it wasn't necessary. (This point in the code would be one place they'd have to change, so they'd have to see the comment.)

I think PG is still using this practice, but they don't often post copyrighted texts, so the examples are sparse. The latest one I found in a new posting was 63849C (2020-11-22). They also show up reposts of older texts. So it's still possible that the question could come up again. Of course, if the question is addressed elsewhere, then it would make sense to link to that.

(Tangent: There's the possibility that another site using the DP code might be posting to somewhere that uses "posted nums" that don't conform to our restrictions. We ignore this possibility, but I don't think anyone's ever complained.)

srjfoo: Thanks for the pointer to 63849C. I looked it up and the ebook, itself does not have the "C" appended anywhere in the document. The only place I saw it was in the posted notice. I suspect that's why the note confused me.

I did confirm with Greg that it was still the practice, but he said that hopefully, that ebook is the last one.

I've added a note back in that I hope makes sense. Good point about it being good to flag it to future devs that this is where any pattern matching changes should be made to enforce a particular pattern.

PR #654, opened 2021-07-31
convert bar charts on misc_user_stats

chrismiceli:
https://www.pgdp.org/~chrismiceli/c.branch/misc_user_stats/stats/misc_user_stats.php

I didn't change the pie charts. They are only used on this page and
https://www.pgdp.net/c/stats/jpgraph_files/equilibria.php
which both are not linked from anywhere in pgdp but used in general reports. Perhaps they could be bar charts? Because I didn't touch them, the spacing between the pie chart and the svg chart are slightly too close, but I'll fix it when I address the pie chart.


pinc/graph_data.inc
```diff
@@ -154,3 +154,73 @@ function tallyboard_deltas($tally_name, $holder_type, $holder_id, $days_back)
...
+        'year' => [
+            'format' => '%Y',
+            'title' => _("Number of users registered per year"),
+        ],
+    ];
+
+    $date_format = $time_interval_options[$time_interval]['format'];
+    $title = $time_interval_options[$time_interval]['title'];
+
+    $res = mysqli_query(DPDatabase::get_connection(), "
```
cpeel, line 182: I know this is merely a lift of what was there before, but please replace these `mysqli_query()` calls with `DPDatabase::query()` (without the `die()`).

chrismiceli: Great idea. I was trying to keep diffs minimal, but this makes sense.

srjfoo:
Looking at the equilibria charts -- I'm not convinced that converting them to bar charts would give the same picture of the data. I see the purpose of the pie charts as showing the numbers as a percentage of the whole.

Mind you, if what people want instead, is a bar chart showing absolute numbers, that's a different matter (but then the equilibria charts would need a new name).

I'm not sure how useful the two pie charts in Miscellaneous User Statistics are (I'm willing to be convinced otherwise):
* For "Number of users per user interface language", nothing but English, Browser default and French are even available (on PROD), so the others are just taking up space in the legend, and aren't useful
* "Number of users per country" -- `com`, `net`, and `org` kind of muddy the waters. It would be more correct to call that "Number of users per Top Level Domain".

pinc/graph_data.inc
```diff
@@ -101,11 +100,14 @@ function user_logging_on($past, $preceding)
...
+    $sql = sprintf("SELECT DATE_FORMAT(FROM_UNIXTIME(time_stamp),'%s'), %s
       FROM user_active_log
-      WHERE time_stamp >= $min_timestamp
-      ORDER BY time_stamp");
+      WHERE time_stamp >= %d
+      ORDER BY time_stamp",
+      DPDatabase::escape($date_format),
+      DPDatabase::escape($column_name),
+      $min_timestamp);
+    $result = DPDatabase::query($sql);
```
srjfoo, line 111: Column spacing? Also, add space before the `%s` in `SELECT DATE_FORMAT(FROM_UNIXTIME(time_stamp),'%s')`?

pinc/graph_data.inc
```diff
@@ -154,3 +156,73 @@ function tallyboard_deltas($tally_name, $holder_type, $holder_id, $days_back)
...
+    ];
+
+    $date_format = $time_interval_options[$time_interval]['format'];
+    $title = $time_interval_options[$time_interval]['title'];
+
+    $sql = sprintf("SELECT FROM_UNIXTIME(date_created,'%s'), COUNT(*)
+      FROM users
+      GROUP BY 1
+      ORDER BY date_created", DPDatabase::escape($date_format));
+    $res = DPDatabase::query($sql);
```
srjfoo, line 188: ditto column spacing, here and lines 210-214?

srjfoo (commented):
The bar charts look great!

lhamilton1:
I'd like to keep the equilibria charts as pie charts since they are showing a balance of a whole and look good and are easy to read because they have big slices. But the miscellaneous user stats should really go to bar. They have too many small amounts and the slice labels blur all together if they're pies.

chrismiceli:
> 
> 
> I'd like to keep the equilibria charts as pie charts since they are showing a balance of a whole and look good and are easy to read because they have big slices. But the miscellaneous user stats should really go to bar. They have too many small amounts and the slice labels blur all together if they're pies.

Ok, thanks for the input. Do you think we should use absolute values in the bar charts, or the percent value as the yaxis height?

lhamilton1:
I'd prefer absolute values for the bars

pinc/graph_data.inc
```diff
@@ -101,11 +100,14 @@ function user_logging_on($past, $preceding)
     ///////////////////////////////////////////////////
     //query db and put results into arrays
 
-    $result = mysqli_query(DPDatabase::get_connection(), "
-      SELECT DATE_FORMAT(FROM_UNIXTIME(time_stamp),'$date_format'), $column_name
+    $sql = sprintf("SELECT DATE_FORMAT(FROM_UNIXTIME(time_stamp),'%s'), %s
```
cpeel: Column names can't be escaped this way. Please leave `$column_name` inline in the SQL per the [best practices](https://www.pgdp.net/wiki/DP_Code_Best_Practices#Column_names_as_variables).

chrismiceli: oh yes, thanks for reminding me.

cpeel (approved):

srjfoo (approved):

70ray (approved):

PR #655, opened 2021-08-04
convert projects_Xed_graphs

chrismiceli:
This one is a little different. Each state had their own color. I ignored those for this iteration so we can follow themes, but let me know if the colors should come back, it is trivial to add. Also, the 3rd chart on each page had negative values, which doesn't make sense to me, so I didn't do it that way. If someone can explain the negatives to me, I'll apply the logic to the new graphs, however overriding the xAxis looks quite strange to me so I'd rather avoid it.

Sandbox links:
https://www.pgdp.org/~chrismiceli/c.branch/projects_Xed_graphs/stats/projects_Xed_graphs.php?which=created
https://www.pgdp.org/~chrismiceli/c.branch/projects_Xed_graphs/stats/projects_Xed_graphs.php?which=proofed
https://www.pgdp.org/~chrismiceli/c.branch/projects_Xed_graphs/stats/projects_Xed_graphs.php?which=PPd
https://www.pgdp.org/~chrismiceli/c.branch/projects_Xed_graphs/stats/projects_Xed_graphs.php?which=posted


pinc/graph_data.inc
```diff
@@ -289,3 +289,221 @@ function users_by_country()
...
+    $datay1 = array_subtract_first_from_each($y_num_projects);
+    array_shift($datay1);
+
+    // Pad out the rest of the month
+    for ($i = count($datay1); $i < $maxday; $i++) {
+        $datax[$i] = $i + 1;
+        $datay1[$i] = "";
+    }
+
+    //draw_projects_graph(
```
cpeel, line 421: Leftover debugging?

stats/projects_Xed_graphs.php
```diff
@@ -4,24 +4,36 @@
...
-output_header($psd->graphs_title);
+$charts = [
+    ["id" => "curr_month_proj", "config" => curr_month_proj($which)],
+    ["id" => "cumulative_month_proj", "config" => cumulative_month_proj($which)],
+    ["id" => "total_proj_graph", "config" => total_proj_graph($which)],
+    ["id" => "cumulative_total_proj_graph", "config" => cumulative_total_proj_graph($which)],
+];
 
-echo "<h1>$psd->graphs_title</h1>";
+$js_data = '$(function(){';
```
cpeel, line 22: Not for this MR, but perhaps we need a function to create this JS and reuse it in these various files?

srjfoo:
> If someone can explain the negatives to me, I'll apply the logic to the new graphs, however overriding the xAxis looks quite strange to me so I'd rather avoid it.

@chrismiceli and I discussed this a bit in Slack -- I'm reluctant to ignore the negative numbers altogether, because they're visible on the graphs on PROD, too. I looked at the `project_state_stats` table, though, and none of those numbers are negative, so those negative bars are being calculated somehow. And I'm not to the point where I grok what the graphs are supposed to mean. So until we get to the point where we understand what's being plotted, I think we should keep the negative numbers (but I'm also willing to be convinced otherwise).

Fwiw, it's not just the third chart -- if you look at the top chart on the [Proofread Projects Graphs](https://www.pgdp.net/c/stats/projects_Xed_graphs.php?which=proofed) page on PROD, you'll see that the second bar is -1. So, the first and third charts, at least, have the potential to go negative.

> Each state had their own color. I ignored those for this iteration so we can follow themes, but let me know if the colors should come back, it is trivial to add.

The different colors make no sense to me. I think that the color should follow the theme.

stats/projects_Xed_graphs.php
chrismiceli: Yeah, once we have a clear pattern after going through all the graphs, this is something we should definitely look into.

chrismiceli:
> 
> 
> > If someone can explain the negatives to me, I'll apply the logic to the new graphs, however overriding the xAxis looks quite strange to me so I'd rather avoid it.
> 
> @chrismiceli and I discussed this a bit in Slack -- I'm reluctant to ignore the negative numbers altogether, because they're visible on the graphs on PROD, too. I looked at the `project_state_stats` table, though, and none of those numbers are negative, so those negative bars are being calculated somehow. And I'm not to the point where I grok what the graphs are supposed to mean. So until we get to the point where we understand what's being plotted, I think we should keep the negative numbers (but I'm also willing to be convinced otherwise).
> 
> Fwiw, it's not just the third chart -- if you look at the top chart on the [Proofread Projects Graphs](https://www.pgdp.net/c/stats/projects_Xed_graphs.php?which=proofed) page on PROD, you'll see that the second bar is -1. So, the first and third charts, at least, have the potential to go negative.
> 
> > Each state had their own color. I ignored those for this iteration so we can follow themes, but let me know if the colors should come back, it is trivial to add.
> 
> The different colors make no sense to me. I think that the color should follow the theme.

I removed the colors and added support for negatives, though it looks strange until we address min width on the bars. Perhaps we should have a 0 line too, but I know that was another common complaint. No yaxis bars for ranges. I say we address these 2 before merge to master, but after converting all the charts so we can style them all at once.

srjfoo:
> I removed the colors and added support for negatives, though it looks strange until we address min width on the bars. Perhaps we should have a 0 line too, but I know that was another common complaint. No yaxis bars for ranges. I say we address these 2 before merge to master, but after converting all the charts so we can style them all at once.

That sounds reasonable to me.

I've looked through the graphs a couple more times, and the things that I would mention are the same things that apply to a lot of the graphs, and should be resolved by across-the-board style changes. The graphs match between old version and new version, and if we're going to question what data is being plotted and why, that should be done separately.

srjfoo (approved):

70ray:
I'm seeing something rather odd:
image

chrismiceli:
> 
> 
> I'm seeing something rather odd:
> image

Are you referring to 3rd graph which has negative numbers without any yAxis lines? It does look odd, especially given the thin lines. We'll fix all that after the graphs are using the svg code. Unless you're referring to something else.

70ray:
> > I'm seeing something rather odd:
> 
> Are you referring to 3rd graph which has negative numbers without any yAxis lines? It does look odd, especially given the thin lines. We'll fix all that after the graphs are using the svg code. Unless you're referring to something else.
No, sorry I should have been more clear. I meant the fainter bar sticking up towards the end of Cumulative graph. I would have thought the graph should be monotonically increasing without any peaks. I thought it was perhaps it is a bug in the browser but it appears in Firefox and chromium.

srjfoo:
> > > I'm seeing something rather odd:
> > 
> > 
> > Are you referring to 3rd graph which has negative numbers without any yAxis lines? It does look odd, especially given the thin lines. We'll fix all that after the graphs are using the svg code. Unless you're referring to something else.
> 
> No, sorry I should have been more clear. I meant the fainter bar sticking up towards the end of Cumulative graph. I would have thought the graph should be monotonically increasing without any peaks. I thought it was perhaps it is a bug in the browser but it appears in Firefox and chromium.

There are some anomalies in the `project_state_stats` table in the TEST database -- the double-height bars are for those days where there are two entries for each state between 2020-05-20 and 2020-09-12 (inclusive). There are other weirdnesses (to me) in how the data is plotted, but that's irrelevant for now, I think, and would involve digging into just how the data that's plotted is determined. Compare to the same two graphs as implemented in jpgraph, and the shapes of those two graphs are the same: 
image

70ray:
> > > > I'm seeing something rather odd:
> > > 
> > > 
> > > Are you referring to 3rd graph which has negative numbers without any yAxis lines? It does look odd, especially given the thin lines. We'll fix all that after the graphs are using the svg code. Unless you're referring to something else.
> > 
> > 
> > No, sorry I should have been more clear. I meant the fainter bar sticking up towards the end of Cumulative graph. I would have thought the graph should be monotonically increasing without any peaks. I thought it was perhaps it is a bug in the browser but it appears in Firefox and chromium.
> 
> There are some anomalies in the `project_state_stats` table in the TEST database -- the double-height bars are for those days where there are two entries for each state between 2020-05-20 and 2020-09-12 (inclusive). There are other weirdnesses (to me) in how the data is plotted, but that's irrelevant for now, I think, and would involve digging into just how the data that's plotted is determined. Compare to the same two graphs as implemented in jpgraph, and the shapes of those two graphs are the same:

Thanks, I see it's the negative numbers causing it.

70ray (approved):

pinc/graph_data.inc
```diff
@@ -289,3 +289,212 @@ function users_by_country()
...
+    $year = date("Y", $todaysTimeStamp);
+    $month = date("m", $todaysTimeStamp);
+    $monthVar = _(date("F", $todaysTimeStamp));
+    $timeframe = "$monthVar $year";
+
+    $maxday = get_number_of_days_in_current_month();
+
+    //query db and put results into arrays
+    $sql = sprintf("SELECT DAYOFMONTH(date) as day, SUM(num_projects)
+      FROM project_state_stats
```
cpeel, line 335: I guess the formatter isn't enforcing the indents of these strings? Please indent these 4 spaces. Ditto the other SQL in the subsequent functions (some of which are indented 1 space from the parent, and at least one with 0 indents).

srjfoo: Is it supposed to enforce code indenting inside strings? If not, then we'll need to be more conscious about remembering to do it manually for this situation.

pinc/graph_data.inc
```diff
@@ -289,3 +289,212 @@ function users_by_country()
...
+
+    $timeframe = _('since stats began');
+
+
+    //query db and put results into arrays
+    $result = DPDatabase::query("SELECT date, SUM(num_projects)
+      FROM project_state_stats
+      WHERE $psd->state_selector
+      GROUP BY date
+      ORDER BY date");
```
srjfoo, line 447: Should this and the same style at line 477 be updated to use the updated querying styling?

chrismiceli: Yup, my search was faulty and only looked at $sql lines

srjfoo (approved):

PR #658, opened 2021-08-08
Rename SF genre to SF&F

srjfoo:
GM request: Rename the `Science Fiction` genre to `Science Fiction & Fantasy`. This acknowledges explicitly how the genre is actually used.

Branch sandbox at [rename-sf-genre](https://www.pgdp.org/~srjfoo/c.branch/rename-sf-genre/).

### DIFF
--- a/SETUP/CHANGELOG.md
+++ b/SETUP/CHANGELOG.md
@@ -6,6 +6,10 @@ see the git history.
 ## R??????
 Scripts supporting this upgrade are in `SETUP/upgrade/16`
 
+* Renamed the 'Science Fiction' genre to 'Science Fiction & Fantasy'.
+    There is no upgrade script for the `queue_defns` table; since
+    release queues do not come pre-defined, updating those entries
+    should be done manually by those who have genre queues enabled.
 
 ## R202102
 Scripts supporting this upgrade are in `SETUP/upgrade/15`
--- a/SETUP/upgrade/16/20210807_update_sf_genre.php
+++ b/SETUP/upgrade/16/20210807_update_sf_genre.php
@@ -1,0 +1,22 @@
+<?php
+$relPath = '../../../pinc/';
+include_once($relPath.'base.inc');
+
+header('Content-type: text/plain');
+
+// ------------------------------------------------------------
+
+echo "Updating genre to 'Science Fiction & Fantasy' where it's 'Science Fiction' ...\n";
+$sql = "
+    UPDATE projects
+        SET genre = 'Science Fiction & Fantasy'
+        WHERE genre = 'Science Fiction'
+";
+
+echo "$sql\n";
+
+mysqli_query(DPDatabase::get_connection(), $sql) or die(mysqli_error(DPDatabase::get_connection()));
+
+// ------------------------------------------------------------
+
+echo "\nDone!\n";
--- a/pinc/ProjectSearchResults.inc
+++ b/pinc/ProjectSearchResults.inc
@@ -218,7 +218,7 @@ class GenreColumn extends Column
         if ($prefix != "") {
             $genre = $prefix." ".$genre;
         }
-        echo $genre;
+        echo html_safe($genre);
     }
 }
 
--- a/pinc/genres.inc
+++ b/pinc/genres.inc
@@ -74,7 +74,7 @@ function load_genre_translation_array()
         'Romance' => _('Romance'),
         'Satire' => _('Satire'),
         'Science' => _('Science'),
-        'Science Fiction' => _('Science Fiction'),
+        'Science Fiction & Fantasy' => _('Science Fiction & Fantasy'),
         'Short Story' => _('Short Story'),
         'Sociology' => _('Sociology'),
         'Speech' => _('Speech'),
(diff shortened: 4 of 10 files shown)


cpeel:
As this is our first genre with `&` we need to make sure that it is escaped correctly everywhere in the HTML. Some places that I've found:
```diff
diff --git a/pinc/ProjectSearchResults.inc b/pinc/ProjectSearchResults.inc
index 507f690da..259a0b515 100644
--- a/pinc/ProjectSearchResults.inc
+++ b/pinc/ProjectSearchResults.inc
@@ -218,7 +218,7 @@ class GenreColumn extends Column
         if ($prefix != "") {
             $genre = $prefix." ".$genre;
         }
-        echo $genre;
+        echo html_safe($genre);
     }
 }

diff --git a/project.php b/project.php
index befa14d13..a90ba8e22 100644
--- a/project.php
+++ b/project.php
@@ -424,7 +424,7 @@ function do_project_info_table()
     echo_row_a(_("Title"), $project->nameofwork, true);
     echo_row_a(_("Author"), $project->authorsname, true);
     echo_row_a(_("Language"), $project->language);
-    echo_row_a(_("Genre"), _($project->genre));
+    echo_row_a(_("Genre"), _($project->genre), true);

     echo_row_a(_("Difficulty"), _($project->difficulty));

```

srjfoo:
> As this is our first genre with `&` we need to make sure that it is escaped correctly everywhere in the HTML. Some places that I've found:
> 
> ```diff
> diff --git a/pinc/ProjectSearchResults.inc b/pinc/ProjectSearchResults.inc
> index 507f690da..259a0b515 100644
> --- a/pinc/ProjectSearchResults.inc
> +++ b/pinc/ProjectSearchResults.inc
> @@ -218,7 +218,7 @@ class GenreColumn extends Column
>          if ($prefix != "") {
>              $genre = $prefix." ".$genre;
>          }
> -        echo $genre;
> +        echo html_safe($genre);
>      }
>  }
> 
> diff --git a/project.php b/project.php
> index befa14d13..a90ba8e22 100644
> --- a/project.php
> +++ b/project.php
> @@ -424,7 +424,7 @@ function do_project_info_table()
>      echo_row_a(_("Title"), $project->nameofwork, true);
>      echo_row_a(_("Author"), $project->authorsname, true);
>      echo_row_a(_("Language"), $project->language);
> -    echo_row_a(_("Genre"), _($project->genre));
> +    echo_row_a(_("Genre"), _($project->genre), true);
> 
>      echo_row_a(_("Difficulty"), _($project->difficulty));
> ```

Thanks -- I think the next question is whether we should avoid that altogether and have me replace the ampersand with 'and'. I used the ampersand primarily because of string length.

cpeel:
> Thanks -- I think the next question is whether we should avoid that altogether and have me replace the ampersand with 'and'. I used the ampersand primarily because of string length.

Well we should be doing that escaping anyway, but I don't feel strongly if you want to punt and use `and` instead.

srjfoo:
> > Thanks -- I think the next question is whether we should avoid that altogether and have me replace the ampersand with 'and'. I used the ampersand primarily because of string length.
> 
> Well we should be doing that escaping anyway, but I don't feel strongly if you want to punt and use `and` instead.

👍  -- no, I have no objection to making the change, especially if we should be doing the escaping, anyway.

chrismiceli:
> 
> 
> > > Thanks -- I think the next question is whether we should avoid that altogether and have me replace the ampersand with 'and'. I used the ampersand primarily because of string length.
> > 
> > 
> > Well we should be doing that escaping anyway, but I don't feel strongly if you want to punt and use `and` instead.
> 
> 👍 -- no, I have no objection to making the change, especially if we should be doing the escaping, anyway.

I don't mind taking a look at the escaping separate of this ticket for genre and other elated fields perhaps.

srjfoo:
> > > > Thanks -- I think the next question is whether we should avoid that altogether and have me replace the ampersand with 'and'. I used the ampersand primarily because of string length.
> > > 
> > > 
> > > Well we should be doing that escaping anyway, but I don't feel strongly if you want to punt and use `and` instead.
> > 
> > 
> > 👍 -- no, I have no objection to making the change, especially if we should be doing the escaping, anyway.
> 
> I don't mind taking a look at the escaping separate of this ticket for genre and other elated fields perhaps.

That might be a good idea, anyway. I haven't updated the MR yet, but I found some of them -- however, I don't know enough to know what I'm missing. I'll get a revised MR up in a bit, and we can look at it then.

srjfoo:
> > > > > Thanks -- I think the next question is whether we should avoid that altogether and have me replace the ampersand with 'and'. I used the ampersand primarily because of string length.
> > > > 
> > > > 
> > > > Well we should be doing that escaping anyway, but I don't feel strongly if you want to punt and use `and` instead.
> > > 
> > > 
> > > 👍 -- no, I have no objection to making the change, especially if we should be doing the escaping, anyway.
> > 
> > 
> > I don't mind taking a look at the escaping separate of this ticket for genre and other elated fields perhaps.
> 
> That might be a good idea, anyway. I haven't updated the MR yet, but I found some of them -- however, I don't know enough to know what I'm missing. I'll get a revised MR up in a bit, and we can look at it then.

I don't know if I got everything, or if what I did get was correct. I did update the sandbox, and what I checked seems to work there, but just the display looking okay doesn't mean the code is necessarily correct.

chrismiceli (approved):
I'll gladly resolve any conflicts after this merges with my MR.

cpeel:
Thanks @chrismiceli - I'll get this merged in and if you can rebase your MR we can get it in too.

PR #659, opened 2021-08-10
html_safe project properties more cautiously

chrismiceli:
sandbox:
https://www.pgdp.org/~chrismiceli/c.branch/genre_safe/
There should be no visible changes unless you are aware of some persisted XSS :).


cpeel (approved):
Code-wise this looks good to me. I don't *think* we're now escaping things that could contain HTML (and thus shouldn't be escaped).

srjfoo (approved):
Looks good to me, too.

PR #661, opened 2021-08-13
update npm versions for security fixes

chrismiceli:
Based on recent dependabot I'm updating major versions to fix npm security issues that don't really impact us since we only use this for development.


cpeel:
Awesome, thanks Chris!

PR #662, opened 2021-08-14
fix render issue on project

chrismiceli:
https://www.pgdp.org/~chrismiceli/c.branch/html_safe/

### DIFF
--- a/project.php
+++ b/project.php
@@ -492,7 +492,7 @@ function do_project_info_table()
 
     $formatted_now = strftime($time_format, time());
     $ct = _("Current Time");
-    $current_time_addition = "&nbsp;&nbsp;&nbsp;(" . html_safe($ct) . ": " . html_safe($formatted_now) . ")";
+    $current_time_addition = "&nbsp;&nbsp;&nbsp;" . html_safe("($ct: $formatted_now)");
 
     echo_row_a(
         _("Last Edit of Project Info"),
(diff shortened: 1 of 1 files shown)


srjfoo (approved):

project.php
```diff
@@ -492,13 +492,12 @@ function do_project_info_table()
 
     $formatted_now = strftime($time_format, time());
     $ct = _("Current Time");
-    $current_time_addition = "&nbsp;&nbsp;&nbsp;($ct: $formatted_now)";
+    $current_time_addition = "&nbsp;&nbsp;&nbsp;(" . html_safe($ct) . ": " . html_safe($formatted_now) . ")";
```
cpeel, line 495: I think this would be a bit easier to read if we just pulled out the spaces-for-formatting from the strings:
```php
$current_time_addition = "&nbsp;&nbsp;&nbsp;" . html_safe("($ct: $formatted_now)");
```

cpeel:
BTW, for PROD I would just recommend hot-fixing this by removing the `true` and not go through the hassle of rebasing and rolling out a new update for this one fix.

srjfoo:
> BTW, for PROD I would just recommend hot-fixing this by removing the `true` and not go through the hassle of rebasing and rolling out a new update for this one fix.

Done!

project.php
chrismiceli: Updated here and in sandbox. You're suggestion is much better :).

srjfoo (approved):

PR #663, opened 2021-08-17
Hide overflow from toolbar

70ray:
This avoids the effect seen when the control bar is at the side and the image view is very short so the buttons will not fit in the available space.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/widget_adjust_1


chrismiceli (approved):

srjfoo (approved):
The purpose of this is to keep the image pane from scrolling below the image if the pane is very short (as in horizontal mode) with the controls to the side. This takes care of it, but we might want to put a warning somewhere (perhaps in the announcement) that if the image pane is too short, the controls will be clipped.

PR #665, opened 2021-08-18
Avoid scrolling vertical toolbar in page browse

70ray:
set overflow hidden in pageBrowse and do not set pane overflow in splitControl.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/widget_adjust_2

### DIFF
--- a/scripts/page_browse.js
+++ b/scripts/page_browse.js
@@ -438,7 +438,7 @@ function pageBrowse(params, storageKey, replaceUrl, mentorMode = false) {
                             stretchDiv.append(textDiv);
                             break;
                         case "imageText": {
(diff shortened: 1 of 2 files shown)


chrismiceli (approved):

srjfoo (approved):
I've approved this, because it works as advertised, and matches the corresponding change in #663, but in retrospect, I'm not sure this is the way to handle the issue. See [my comment](https://github.com/DistributedProofreaders/dproofreaders/pull/666#issuecomment-901673603) in #666 (the second number 1).

70ray:
Another MR allows the control bar to scroll instead.

PR #666, opened 2021-08-18
Widget rotate 2

70ray:
This enables the image to be rotated using a canvas rather than the previous attempt using css transforms in which it proved difficult to prevent the image from being scrolled out of view.

This new method has the unforeseen effect of avoiding the iPad fit-height issue.

If we do not want to implement the rotation controls immediately we can disable them and test the rest of it first.

It also has the following differences:
1. The image size is relative to the natural width rather than 1000pixels. This could be changed.
2. The alignment is always to the left when the image does not fill the window. This could be implemented to be left, centre or right.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/widget_rotate_2

### DIFF
--- a/scripts/control_bar.js
+++ b/scripts/control_bar.js
@@ -132,7 +132,7 @@ var makeImageControl = function(canvas, reSize) {
     })
         .append($("<i>", {class: 'fas fa-arrows-alt-v'}));
 
-    let clockRotateInput = $("<button>")
+    let clockRotateInput = $("<button>", {title: texts.clockRotate})
         .append($("<i>", {class: 'fas fa-redo-alt'}))
         .click( function () {
             let newCos = sine;
@@ -141,7 +141,7 @@ var makeImageControl = function(canvas, reSize) {
             reDraw();
         });
 
-    let anticlockRotateInput = $("<button>")
+    let anticlockRotateInput = $("<button>", {title: texts.anticlockRotate})
         .append($("<i>", {class: 'fas fa-undo-alt'}))
         .click( function () {
             let newCos = -sine;
--- a/scripts/control_bar.js
+++ b/scripts/control_bar.js
@@ -223,7 +223,6 @@ var makeImageControl = function(canvas, align, reSize) {
         });
 
     const zoomIn = $("<button>", {title: texts.zoomIn}).click(function () {
-        setPercent();
         scale *= 1.1;
         setPercent();
         drawImage();
--- a/scripts/control_bar.js
+++ b/scripts/control_bar.js
@@ -37,7 +37,7 @@ var makeImageControl = function(canvas, align, reSize) {
     // of view. So redraw when pane size changes.
     // This would be easy if we could use ResizeObserver(). In the meantime
     // use a resize callback which gets fired when pane size changes.
-    function reDraw() {
+    function drawImage() {
         getScrollbarWidth();
         // clearRect acts through transform
         ctx.resetTransform();
@@ -109,7 +109,7 @@ var makeImageControl = function(canvas, align, reSize) {
     }
 
     if(reSize) {
-        reSize.add(reDraw);
+        reSize.add(drawImage);
     }
 
     function clearCanvas() {
@@ -121,7 +121,7 @@ var makeImageControl = function(canvas, align, reSize) {
         clearCanvas();
         imageWidth = image.width;
         imageHeight = image.height;
-        reDraw();
+        drawImage();
     };
 
     function saveZoom(percent) {
@@ -143,7 +143,7 @@ var makeImageControl = function(canvas, align, reSize) {
         // in case above has changed it
         this.value = percent;
         scale = percent / 100;
-        reDraw();
+        drawImage();
         saveZoom(percent);
     });
 
@@ -181,7 +181,7 @@ var makeImageControl = function(canvas, align, reSize) {
             }
         }
         setPercent();
-        reDraw();
+        drawImage();
     })
         .append($("<i>", {class: 'fas fa-arrows-alt-h'}));
 
@@ -200,7 +200,7 @@ var makeImageControl = function(canvas, align, reSize) {
             }
         }
         setPercent();
-        reDraw();
+        drawImage();
     })
         .append($("<i>", {class: 'fas fa-arrows-alt-v'}));
 
@@ -210,7 +210,7 @@ var makeImageControl = function(canvas, align, reSize) {
             let newCos = sine;
             sine = -cosine;
             cosine = newCos;
-            reDraw();
+            drawImage();
         });
 
     let anticlockRotateInput = $("<button>", {title: texts.anticlockRotate})
@@ -219,21 +219,21 @@ var makeImageControl = function(canvas, align, reSize) {
             let newCos = -sine;
             sine = cosine;
             cosine = newCos;
-            reDraw();
+            drawImage();
         });
 
     const zoomIn = $("<button>", {title: texts.zoomIn}).click(function () {
         setPercent();
         scale *= 1.1;
         setPercent();
-        reDraw();
+        drawImage();
     })
         .append($("<i>", {class: 'fas fa-search-plus'}));
 
     const zoomOut = $("<button>", {title: texts.zoomOut}).click(function () {
         scale /= 1.1;
         setPercent();
-        reDraw();
+        drawImage();
     })
         .append($("<i>", {class: 'fas fa-search-minus'}));
 
@@ -258,7 +258,7 @@ var makeImageControl = function(canvas, align, reSize) {
             let percent = imageData.zoom;
             percentInput.val(percent);
             scale = percent / 100;
-            reDraw();
+            drawImage();
         },
         setImage: function(src) {
             // reset to normal orientation
--- a/scripts/control_bar.js
+++ b/scripts/control_bar.js
@@ -4,6 +4,7 @@
 // Construct the image sizing controls.
 var makeImageControl = function(canvas, align, reSize) {
     let imageKey;
+    let percent;
     let scale = 1;
     let imageWidth, imageHeight;
     const image = document.createElement("img");
@@ -117,10 +118,22 @@ var makeImageControl = function(canvas, align, reSize) {
         ctx.clearRect( 0, 0, canvas.width, canvas.height);
     }
 
+    let relative1000 = true;
+    function percentToScale() {
+        if(relative1000) {
+            // 100% means image scaled to be 1000px wide
+            scale = percent * 10 / imageWidth;
+        } else {
+            // 100% means natural size
+            scale = percent / 100;
+        }
+    }
+
     image.onload = function() {
         clearCanvas();
         imageWidth = image.width;
         imageHeight = image.height;
+        percentToScale();
         drawImage();
     };
 
@@ -129,7 +142,6 @@ var makeImageControl = function(canvas, align, reSize) {
     }
 
     percentInput.change(function() {
-        let percent;
         const value = parseInt(this.value);
         if(isNaN(value)) {
             percent = 100;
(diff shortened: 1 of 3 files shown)


cpeel:
> The alignment is always to the left when the image does not fill the window. This could be implemented to be left, centre or right.

If we switch to canvas I think it should work to have no user-visible changes this this is already live in the Page Browser and we have the updates to the proofreading interfaces already out for testing.

srjfoo:
> > The alignment is always to the left when the image does not fill the window. This could be implemented to be left, centre or right.
> 
> If we switch to canvas I think it should work to have no user-visible changes this this is already live in the Page Browser and we have the updates to the proofreading interfaces already out for testing.

I was thinking along the same lines. Currently, the Page Browser and Standard Interface have the image centered. Only the Enhanced Interface has the image left-aligned. I would prefer, also, to see no user-visible changes in the alignment.

70ray:
> > > The alignment is always to the left when the image does not fill the window. This could be implemented to be left, centre or right.
> > 
> > 
> > If we switch to canvas I think it should work to have no user-visible changes this this is already live in the Page Browser and we have the updates to the proofreading interfaces already out for testing.
> 
> I was thinking along the same lines. Currently, the Page Browser and Standard Interface have the image centered. Only the Enhanced Interface has the image left-aligned. I would prefer, also, to see no user-visible changes in the alignment.

I will look into implementing this. It shouldn't be difficult.
How do y'all feel about the percentage width being relative to natural size or 1000px?

srjfoo:
> > > > The alignment is always to the left when the image does not fill the window. This could be implemented to be left, centre or right.
> > > 
> > > 
> > > If we switch to canvas I think it should work to have no user-visible changes this this is already live in the Page Browser and we have the updates to the proofreading interfaces already out for testing.
> > 
> > 
> > I was thinking along the same lines. Currently, the Page Browser and Standard Interface have the image centered. Only the Enhanced Interface has the image left-aligned. I would prefer, also, to see no user-visible changes in the alignment.
> 
> I will look into implementing this. It shouldn't be difficult.
> How do y'all feel about the percentage width being relative to natural size or 1000px?

Thanks for the reminder. I think that's an issue that should be discussed separately and not changed without discussion. I would stick with 1000px for now, but raise it as a separate question that can be implemented easily.

However ... I would like to better understand the differences between the two modes, so that when we do start the discussion, I can do a pro and con list. Since you have it implemented in the sandbox, let me play with it for a bit.

As for the question about removing the rotation controls for now -- I want to play with it a bit before I offer an opinion, and I haven't had a chance to do that yet. I'll get back to you on it.

scripts/control_bar.js
```diff
@@ -2,21 +2,121 @@
...
+    // This introduces a new problem: if the image is smaller than the pane
+    // and the pane size decreases it is possible to scroll the image out
+    // of view. So redraw when pane size changes.
+    // This would be easy if we could use ResizeObserver(). In the meantime
+    // use a resize callback which gets fired when pane size changes.
+    function reDraw() {
+        getScrollbarWidth();
+        // clearRect acts through transform
+        ctx.resetTransform();
+        ctx.clearRect( 0, 0, canvas.width, canvas.height);
```
chrismiceli, line 48: minor, but space before 0 feels odd. I'm surprised the linter doesn't complain.

scripts/control_bar.js
```diff
@@ -2,21 +2,121 @@
...
+        let scaleWidth = imageWidth * scale;
+        let scaleHeight = imageHeight * scale;
+        let newWidth, newHeight;
+        if(cosine != 0) {
```
chrismiceli, line 52: same here with the if.

scripts/control_bar.js
```diff
@@ -2,21 +2,121 @@
...
+    // If we always make the canvas fit image then when reducing the image size
+    // rapidly with spinner the canvas does not get cleared even when
+    // redrawing with setTimer(0).
+    // So, make canvas dimension fit image dimension only if it is larger
+    // than the pane. A scroll bar will then appear. Otherwise make canvas
+    // dimension fill pane. Canvas now always clears.
+    // This introduces a new problem: if the image is smaller than the pane
+    // and the pane size decreases it is possible to scroll the image out
+    // of view. So redraw when pane size changes.
+    // This would be easy if we could use ResizeObserver(). In the meantime
```
chrismiceli, line 42: we don't support IE do we? If not, we can use ResizeObserver.

scripts/control_bar.js
```diff
@@ -2,21 +2,121 @@
...
+    var getScrollbarWidth = function() {
+        if (scrollbarWidth === false) {
+            let testDiv = document.createElement('div');
+            testDiv.innerHTML = '<div style="width:50px;height:5px;position:absolute; overflow:scroll;"><div style="width:100px;height:5px;"></div></div>';
+            let innerDiv = testDiv.firstChild;
+            document.body.appendChild(testDiv);
+            scrollbarWidth = innerDiv.offsetWidth - innerDiv.clientWidth;
+            document.body.removeChild(testDiv);
+        }
+        return scrollbarWidth;
```
chrismiceli, line 26: no one uses this return value. I'd prefer that they do and avoid having to be aware of the local variable perhaps?

scripts/control_bar.js
```diff
@@ -2,21 +2,121 @@
...
+
+    const percentInput = $("<input>", {type: 'number', min: '1', max: '999', value: 100, title: texts.zoomPercent});
+
+    let sine = 0, cosine = 1;
+
+    let scrollbarWidth = false;
+    var getScrollbarWidth = function() {
+        if (scrollbarWidth === false) {
+            let testDiv = document.createElement('div');
+            testDiv.innerHTML = '<div style="width:50px;height:5px;position:absolute; overflow:scroll;"><div style="width:100px;height:5px;"></div></div>';
```
chrismiceli, line 20: Will this cause a flash of a scroll bar on slow browsers? I wonder if there is a solution where we don't need to know scrollbar width.

chrismiceli (commented):
Another weird thing I noticed is when you right click the "image" and open in new tab, you don't get the raw image, but the canvas image, with the background. I prefer if I could get to the raw image, which I've done when proofing before, though I'm not sure why I do it. I imagine other people would have similar feedback. Otherwise, how do you get the raw image from a project?

srjfoo:
I haven't finished all I want to test, but wanted to leave some notes before I quit for the night ...

Results of my iPad testing:

1. Yes, fit height seems to work on the iPad now -- excellent! (I'd be happier if I understood why it wasn't working before -- was it an Apple bug, or something obscure that wasn't being terminated properly that other browsers/tablets default to?).

2. Regarding the continued disappearance of the tools in the upper right: it seems to happen only if the image is larger than the image pane (in Chrome and Safari. FF seems to be okay). I.e. if you fit height or make it even smaller, it doesn't disappear. If you switch to top from either side, it doesn't disappear. It only disappears if you switch bottom to top. (That's not a function of the current MR, though -- I just checked the main TEST sandbox, and that's true in the current implementation.)

3. The thing that I find most bothersome in my current testing is that in the Page Browser, in image only mode, I have no image controls at all in Chrome and Safari. Firefox seems to be okay. If I switch to image and text, I have image controls, but the settings don't, and probably shouldn't, persist if I switch back to image-only.


Some more general comments from testing:

1. If the toolbar is top or bottom, and the image pane is narrower than the width that the controls take up, the buttons wrap. Can we do this for the buttons on the side? I can't guarantee that we'll get that question if we don't do it, but there's a good chance of it. More people than just me prefer horizontal mode with short image and textareas, so as to filter out too many distractions. (I'm fine with having the controls at the top, so they might be, too, but treating the controls differently if they're to the side than if they're top or bottom seems like the wrong way to go about it, the more I think about it.)

2. I agree with Chris that the result of context-clicking on the image to bring it up in a new tab is not what I would expect. (@chrismiceli -- you can actually get the raw image from going to the "View Images Online" link that's available at detail level 3 or 4.)

I need to do more testing on all three browsers on the iPad, but not tonight, comparing against the main TEST sandbox, but I've run out of steam for the night.

70ray:
> Another weird thing I noticed is when you right click the "image" and open in new tab, you don't get the raw image, but the canvas image, with the background. I prefer if I could get to the raw image, which I've done when proofing before, though I'm not sure why I do it. I imagine other people would have similar feedback. Otherwise, how do you get the raw image from a project?

I am not sure I undestand what you are seeing. With FF I get much the same view, the url says something like "blob:null/811f472e-a0f2-438f-a01d-925ee3b66566" rather than something like "https://www.pgdp.org/projects/projectID4c49f39aa829c/140.png". With chromium or edge the option to open image in new tab doesn't appear at all with the canvas image.

scripts/control_bar.js
70ray: fixed

scripts/control_bar.js
70ray: I agree it looks better with a space but they are all ike this

scripts/control_bar.js
70ray: We only claim to support "Chrome 50 * Firefox 50 * Microsoft Edge * Opera 40 * Safari 10" so we can't use it yet.

scripts/control_bar.js
70ray: I copied this from somewhere. I've simplified it.

scripts/control_bar.js
70ray: I haven't managed to find such a solution.

70ray:
>     1. Yes, fit height seems to work on the iPad now -- excellent! (I'd be happier if I understood why it wasn't working before -- was it an Apple bug, or something obscure that wasn't being terminated properly that other browsers/tablets default to?).
> 

It seems to be an iPad feature.
> 
>     3. The thing that I find most bothersome in my current testing is that in the Page Browser, in image only mode, I have no image controls at all in Chrome and Safari. Firefox seems to be okay. If I switch to image and text, I have image controls, but the settings don't, and probably shouldn't, persist if I switch back to image-only.
> 
This looks like another ipad feature - vertical height vh problem. It seems this has caused a lot of problems but should be fixable.
> 

I'll look at your other comments soon.

srjfoo:
> > Another weird thing I noticed is when you right click the "image" and open in new tab, you don't get the raw image, but the canvas image, with the background. I prefer if I could get to the raw image, which I've done when proofing before, though I'm not sure why I do it. I imagine other people would have similar feedback. Otherwise, how do you get the raw image from a project?
> 
> I am not sure I undestand what you are seeing. With FF I get much the same view, the url says something like "blob:null/811f472e-a0f2-438f-a01d-925ee3b66566" rather than something like "https://www.pgdp.org/projects/projectID4c49f39aa829c/140.png". With chromium or edge the option to open image in new tab doesn't appear at all with the canvas image.

I'm seeing what you describe -- the "blob:....." string. What is live now is the 'something like "https://www.pgdp.org/projects/projectID4c49f39aa829c/140.png"', however. For what it's worth, this is something that I've done more than once -- not so much from the page browser, but from the proofreading interface, and I'm pretty sure that others do it, too.

srjfoo:
> >
> >1. Yes, fit height seems to work on the iPad now -- excellent! (I'd be happier if I understood why it wasn't working before -- was it an Apple bug, or something obscure that wasn't being terminated properly that other browsers/tablets default to?).
> >
> 
> It seems to be an iPad feature.
> 
> >
> > 3. The thing that I find most bothersome in my current testing is that in the Page Browser, in image only mode, I have no image controls at all in Chrome and Safari. Firefox seems to be okay. If I switch to image and text, I have image controls, but the settings don't, and probably shouldn't, persist if I switch back to image-only.

\<sigh\> -- "feature", huh? (Not a "bug", we did this on purpose, honest!)

> This looks like another ipad feature - vertical height vh problem. It seems this has caused a lot of problems but should be fixable. The toolbar is at the bottom but off the screen.

\<sigh\> again. I've been wondering if it was that, but there's literally no way to get it back on the screen right now.

> I'll look at your other comments soon.

No big hurry. I need to do some more testing, too.

70ray:
Fit-height problem withiPad fixed a different way. Don't implement image rotate yet.

PR #667, opened 2021-08-20
pages_daily to svg graphs

chrismiceli:
https://www.pgdp.org/~chrismiceli/c.branch/pages_daily/stats/pages_proofed_graphs.php?tally_name=P1


srjfoo:
The blue lines in graph 4 don't appear to have the same weight as the green and orange lines -- I can see them if I squint, and because, comparing with the jpgraph currently on TEST, I know the lines should be there, but if I didn't know something was supposed to be there, I wouldn't notice.

Graph 5, is the blue partially transparent?Maybe it's just because there is so little data, but I also didn't notice that right at first.

Everything else looks great!

chrismiceli:
> The blue lines in graph 4 don't appear to have the same weight as the green and orange lines -- I can see them if I squint, and because, comparing with the jpgraph currently on TEST, I know the lines should be there, but if I didn't know something was supposed to be there, I wouldn't notice.
> 
> Graph 5, is the blue partially transparent?Maybe it's just because there is so little data, but I also didn't notice that right at first.
> 
> Everything else looks great!

Yeah that's the same as the other graphs. I think I'll solve it by making a minimum width of bars but will play with that after the conversions so I can see how changes imlact all the graphs at once.

srjfoo:
> > The blue lines in graph 4 don't appear to have the same weight as the green and orange lines -- I can see them if I squint, and because, comparing with the jpgraph currently on TEST, I know the lines should be there, but if I didn't know something was supposed to be there, I wouldn't notice.
> > Graph 5, is the blue partially transparent?Maybe it's just because there is so little data, but I also didn't notice that right at first.
> > Everything else looks great!
> 
> Yeah that's the same as the other graphs. I think I'll solve it by making a minimum width of bars but will play with that after the conversions so I can see how changes imlact all the graphs at once.

I think what confused me was that the green and orange lines looked thicker. But this is also the first time we've had multi-color graphs, I guess?

chrismiceli:
> > > The blue lines in graph 4 don't appear to have the same weight as the green and orange lines -- I can see them if I squint, and because, comparing with the jpgraph currently on TEST, I know the lines should be there, but if I didn't know something was supposed to be there, I wouldn't notice.
> > > Graph 5, is the blue partially transparent?Maybe it's just because there is so little data, but I also didn't notice that right at first.
> > > Everything else looks great!
> > 
> > 
> > Yeah that's the same as the other graphs. I think I'll solve it by making a minimum width of bars but will play with that after the conversions so I can see how changes imlact all the graphs at once.
> 
> I think what confused me was that the green and orange lines looked thicker. But this is also the first time we've had multi-color graphs, I guess?

Besides the stacked one on stats_central yeah this is the first.

srjfoo (approved):

PR #670, opened 2021-08-22
Bar scroll

70ray:
The vertical control bar now scrolls if its contents will not fit. The scroll bar covers the "%" sign. The bar can also be scrolled horizontally (by arrow keys) to reveal it but the bar is so narrow that the browser does not display the horizontal scroll bar.
The way this is implemented is that the bar overflow is now set to "auto". Unfortunately this prevents the menu from appearing. This is apparently because it is in a relatively positioned div. This div was removed and the menu position set by javascript functions instead. Thanks to "https://css-tricks.com/popping-hidden-overflow/" and other sites.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/bar_scroll

### DIFF
--- a/scripts/control_bar.js
+++ b/scripts/control_bar.js
@@ -121,12 +121,33 @@ function makeControlDiv(container, controls, onChange) {
     controlBar.append(control1, control2, control3);
 
     const menu = $("<div>", {class: "control-bar-menu"});
-
     const menuButton = $("<button>", {title: texts.adjustPanel})
         .append($("<i>", {class: 'fas fa-cog'}))
         .click(function () {
+            if(menu.is(":hidden")) {
+                // find the position of the menu button and set position of
+                // menu relative to it
+                const menuButtonWidth = menuButton.outerWidth();
+                const menuWidth = menu.outerWidth();
+                const buttonPos = menuButton.offset();
+                switch(compassPoint) {
+                case "N":
+                    menu.css({top: buttonPos.top + 1.5 * menuButtonWidth, left: buttonPos.left});
+                    break;
+                case "W":
+                    menu.css({top: buttonPos.top, left: buttonPos.left + 1.5 * menuButtonWidth});
+                    break;
+                case "S":
+                    menu.css({top: buttonPos.top - menuWidth - 0.5 * menuButtonWidth, left: buttonPos.left});
+                    break;
+                default: // E
+                    menu.css({top: buttonPos.top, left: buttonPos.left - menuWidth - 0.5 * menuButtonWidth});
+                    break;
+                }
+            }
             menu.toggle();
         });
+
 
     // build navBox
     const navBox = $("<table>");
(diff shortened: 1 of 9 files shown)


chrismiceli (approved):

srjfoo:
> The vertical control bar now scrolls if its contents will not fit. The scroll bar covers the "%" sign. The bar can also be scrolled horizontally (by arrow keys) to reveal it but the bar is so narrow that the browser does not display the horizontal scroll bar.

For what it's worth, I do see a horizontal scrollbar on the vertical toolbar in FF and Safari, macOS 11.5.2 (I gave in and upgraded my alternate machine to Big Sur). For some reason, though, Chrome leaves the toolbar it's regular size and adds the scrollbar to the outside of the toolbar.

> The way this is implemented is that the bar overflow is now set to "auto". Unfortunately this prevents the menu from appearing. This is apparently because it is in a relatively positioned div. This div was removed and the menu position set by javascript functions instead. Thanks to "https://css-tricks.com/popping-hidden-overflow/" and other sites.

Which menu does it prevent appearing?

\<tb\>

Something to think about: while this solves the problem of having the buttons become inaccessible when the toolbar is very short and to the right or left, it's implemented differently from the way it's implemented for the horizontal toolbar.

I'm not an ordinary user, and the difference doesn't bother me, but it's worth considering what ordinary users will think of it.

70ray:
> > The vertical control bar now scrolls if its contents will not fit. The scroll bar covers the "%" sign. The bar can also be scrolled horizontally (by arrow keys) to reveal it but the bar is so narrow that the browser does not display the horizontal scroll bar.
> 
> For what it's worth, I do see a horizontal scrollbar on the vertical toolbar in FF and Safari, macOS 11.5.2 (I gave in and upgraded my alternate machine to Big Sur). For some reason, though, Chrome leaves the toolbar it's regular size and adds the scrollbar to the outside of the toolbar.
> 
Yes, different browsers seem to do scrollbars differently. I just noticed that Safari puts the scrollbar on top of the content and then hides it. This would avoid the difficulty I mentioned yesterday where the scrollbar changes the size of the viewport.

> > The way this is implemented is that the bar overflow is now set to "auto". Unfortunately this prevents the menu from appearing. This is apparently because it is in a relatively positioned div. This div was removed and the menu position set by javascript functions instead. Thanks to "https://css-tricks.com/popping-hidden-overflow/" and other sites.
> 
> Which menu does it prevent appearing?
> 
I meant the popup menu with the arrows to set the  positions of the toolbars.
> <tb>
> 
> Something to think about: while this solves the problem of having the buttons become inaccessible when the toolbar is very short and to the right or left, it's implemented differently from the way it's implemented for the horizontal toolbar.
> 
I think html is unsymmetrical regarding vertical and horizontal. When you make a window narrower text and boxes can wrap onto the next line and the viewport they are in can become taller to accomodate them. But it doesn't work like that the other way round (unless perhaps if it set up for a top to bottom writing system - I don't know). 
It would though be possible to stop the tools wrapping when the bar is horizontal so it scrolls instead.

> I'm not an ordinary user, and the difference doesn't bother me, but it's worth considering what ordinary users will think of it.
Yes. I tend to assume everyone uses a large screen where these situations will never arise but this is probably not the case.

70ray:
Sandbox updated with fix for disappearing toolbar controls.

chrismiceli (approved):

srjfoo (approved):

70ray:
The latest changes allow the page browser textarea to appear correctly on iPad.

chrismiceli (approved):
Code looks good.

PR #671, opened 2021-08-23
Pp stage goal

chrismiceli:
https://www.pgdp.org/~chrismiceli/c.branch/pp_stage_goal/stats/pp_stage_goal.php

This MR has 2 commits. One simply commits pp_stage_goal from noncvs, and then the 2nd converts it to svg.
This also covers steps 3 and 4 of our plan and removes unused graphs.


srjfoo:
I marked a project posted and formatted another couple of F2 pages so we could see both bars (ironically, the "project" I "marked posted" had the same number of pages in it that had until this point, been formatted in F2 this month :grin:), and put it in the [news item](https://www.pgdp.org/~chrismiceli/c.branch/pp_stage_goal/activity_hub.php) for the AH.

It's not obvious from bringing up the graph standalone, but once it's in the  news item, you can see that the graph object takes up the same amount of space the others do, even though the graph does not. It might also be a good idea to make the graph slightly wider and put a small amount more space between the bars and a bit more space between the title and the top of the graph.

stats/pp_stage_goal.php
```diff
@@ -0,0 +1,81 @@
...
+
+// Get start of current month as unixtime
+$m_start = mktime(0, 0, 0, date('m'), 1, date('Y'));
+
+// We want the "PP pages goal" to be equal to the current month's F2 actuals
+$page_offset = 0;
+$site_stats = get_site_page_tally_summary("F2");
+$pp_page_goal = $site_stats->curr_month_actual + $page_offset;
+
+// Start with creating the Graph, this enables the use of the cache
```
cpeel, line 17: This comment is no longer true and can be removed. We could probably move the width/height definitions down to just before we define `$js_data`.

stats/pp_stage_goal.php
```diff
@@ -0,0 +1,81 @@
...
+
+// calculate the goal percent as long as $pp_page_goal isn't zero
+if ($pp_page_goal != 0) {
+    $goal_percent = ceil(($pp_pages_total / $pp_page_goal) * 100);
+} else {
+    $goal_percent = 0;
+}
+
+// Some graph variables
+$datax = [_("F2"), _("PP")];
```
cpeel, line 42: This is why the graph was in noncvs -- other sites in theory might not have a F2 if they've changed the stages to only have 1 (or 3!) formatting arounds. We can mitigate this somewhat by ensuring that we are comparing against the last formatting round:
```php
# not exactly elegant, but quick and clear:
$last_formatting_round = Null;
foreach($Round_for_round_id_ as $id => $round) {
    if(is_formatting_round($round)) {
        $last_formatting_round = $id;
    }
}
```

Also: F2 and PP should not be translated as they are effectively language-agnostic keys.

srjfoo: Does it have to be the last formatting round, or just the last **_round_** before the project goes into PP?

cpeel: Probably the latter, good point. So perhaps:
```php
$round_before_PP = Null;
foreach($Round_for_round_id_ as $id => $round) {
    if($round->id == "PP") {
        break;
    }
    $round_before_PP = $round->id;
}
```

chrismiceli:
> 
> 
> I marked a project posted and formatted another couple of F2 pages so we could see both bars (ironically, the "project" I "marked posted" had the same number of pages in it that had until this point, been formatted in F2 this month 😁), and put it in the [news item](https://www.pgdp.org/~chrismiceli/c.branch/pp_stage_goal/activity_hub.php) for the AH.
> 
> It's not obvious from bringing up the graph standalone, but once it's in the news item, you can see that the graph object takes up the same amount of space the others do, even though the graph does not. It might also be a good idea to make the graph slightly wider and put a small amount more space between the bars and a bit more space between the title and the top of the graph.

I adjusted the news item to set an appropriate width after making the graph a little wider and adjusted space between title and axis.

srjfoo:
> I adjusted the news item to set an appropriate width after making the graph a little wider and adjusted space between title and axis.

Well, duh! Very obvious in retrospect, but I didn't even think to check the container size!

srjfoo (approved):

PR #675, opened 2021-08-28
equilibria

chrismiceli:
https://www.pgdp.org/~chrismiceli/c.branch/equilibria/stats/equilibria.php


srjfoo (commented):
Do we need two significant digits after the decimal point? The current equilibria charts just have one, and also display the percent sign, which maybe we ought to keep?

I proofed a handful of pages so that there would be something in the first chart, and one thing that I notice is that the "0.00" percent for F1 is sitting above the division between P3 and F2. That does not correspond to the current graph, where the zero does appear halfway in between. I'm wondering a couple of things:
1. Is it likely to be misleading?
2. For 0%, do we even need the number there? I know that we have it on the current graphs, but I think it adds more clutter than value, and the number of pages is listed in the legend, anyway, so I don't think it's really needed.

The only other comments I can think of are about things that are slated for consideration on a global level -- these look really spiffy!

srjfoo:
I realized this morning that I should have included screenshots, because what I'm seeing won't necessarily persist: 

image

image

I don't know how easy it is to do, but it might also make more sense (if it's possible) to put the percentages around the periphery if possible: readability on some colors is more of a problem than on others. (If it's not possible, it's going to be even more important on these charts to be able to theme the chart colors.)

stats/equilibria.php
```diff
@@ -59,8 +41,28 @@ function display_graph($d)
     }
 
     if ($total_pages == 0) {
-        dpgraph_error(_("No pages saved in specified range"), 660, 400);
+        $title = _("No pages saved in specified range");
```
cpeel, line 44: I feels weird to have this as the page title. Perhaps have the title be `_("Graph error")` and have an `$error = _("No pages saved in specified range");` that gets output in an `<p class='error'>` after the title if defined?

chrismiceli: good points. I have implemented where the graph title shows and the error show below, that way the user is aware of which chart has an error.

chrismiceli:
> 
> 
> I realized this morning that I should have included screenshots, because what I'm seeing won't necessarily persist:
> 
> image
> 
> image
> 
> I don't know how easy it is to do, but it might also make more sense (if it's possible) to put the percentages around the periphery if possible: readability on some colors is more of a problem than on others. (If it's not possible, it's going to be even more important on these charts to be able to theme the chart colors.)

Good points. I filtered out 0 labels and moved the labels outside the pie. The placement of the labels is actually quite tricky as finding the center of a text label to place on a radius without overlapping the pie is not trivial. We can always fine tune too during display updates we make to any graphs before merging.

srjfoo (approved):

PR #677, opened 2021-09-01
Add SR to the snapshot table on the AH

srjfoo:
All activity stages except Smooth Reading show information in the site progress snapshot table on the Activity Hub. This adds the SR counts and implements the spirit of task #728.

Sandbox at [ah-sr](https://www.pgdp.org/~srjfoo/c.branch/ah-sr/activity_hub.php)

Testers: note that you must have proofed at least 2 pages in order to see the filters (I decreased it just especially for this sandbox).

### DIFF
--- a/activity_hub.php
+++ b/activity_hub.php
@@ -95,6 +95,18 @@ while ([$project_state, $count] = mysqli_fetch_row($result)) {
     $n_projects_in_state_[$project_state] = $count;
 }
 
+// SR projects have an additional qualifier beyond just the state,
+// so we need an additional query to pull the SR count, and to add
+// it to $n_projects_in_state_
+$sql = "
+    SELECT COUNT(*)
+    FROM projects
+    WHERE state = '" . PROJ_POST_FIRST_CHECKED_OUT . "' AND
+        smoothread_deadline > UNIX_TIMESTAMP()
+";
+$result = DPDatabase::query($sql);
+[$n_projects_in_state_['SR']] = mysqli_fetch_row($result);
+
 // Users can elect to show project stats for all projects, or just for
 // projects that match their project filter (ie: the filter used on
 // the corresponding Round or Pool pages to limit the projects shown on those
@@ -121,7 +133,6 @@ if ($pagesproofed <= 20) {
     $show_filtered_projects = false;
     $show_filtering_links = false;
 }
-
 
 progress_snapshot_table($show_filtered_projects, $show_filtering_links, ($pagesproofed < 300));
 
@@ -243,12 +254,36 @@ function progress_snapshot_table($show_filtered_projects, $show_filtering_links,
         if (is_a($stage, 'Pool') || is_a($stage, 'Round')) {
             continue;
         }
+        if ($stage->id == "SR") {
+            echo "\n<tr>";
+            echo "<th class='activity-icon-header'></th>";
+            echo "<th colspan='3'>" . _("Projects") . " = ";
+            if ($show_filtered_projects) {
+                if ($show_filtering_links) {
+                    echo "<a href='?show_filtered=0#progress_snapshot'>" . pgettext("all projects", "All") . "</a> | ";
+                }
+                echo "<b>" . _("Filtered") . "</b>";
+            } else {
+                echo "<b>" . pgettext("all projects", "All") . "</b>";
+                if ($show_filtering_links) {
+                    echo " | <a href='?show_filtered=1#progress_snapshot'>" . _("Filtered") . "</a>";
+                }
+            }
+            echo "</tr>\n";
 
-        $desired_states = [];
+            //There is no real 'SR' state, but we hacked a row into the
+            // n_projects_in_state array to hold the SR count.
+            $desired_states = ['SR'];
 
-        summarize_stage($stage, $desired_states);
+            summarize_stage($stage, $desired_states, $show_filtered_projects, $stage->id);
+        } else {
+            $desired_states = [];
+
+            summarize_stage($stage, $desired_states);
+        }
     }
     echo "\n</table>\n";
+
     echo "<a href='faq/site_progress_snapshot_legend.php' target='_blank'>" . _("Information about this table") . "</a>";
 }
 
@@ -337,10 +372,16 @@ function summarize_stage($stage, $desired_states, $show_filtered_projects = fals
             }
         }
 
+        if ($stage->id == 'SR') {
+            $where_clause = "state = '" . PROJ_POST_FIRST_CHECKED_OUT . "' AND smoothread_deadline > UNIX_TIMESTAMP()";
+        } else {
+            $where_clause = "state IN ($states_list)";
+        }
+
         $sql = "
             SELECT state, COUNT(*)
             FROM projects
-            WHERE state IN ($states_list) $project_filter
+            WHERE $where_clause $project_filter
             GROUP BY state
         ";
         $result = DPDatabase::query($sql);
@@ -393,6 +434,8 @@ function summarize_stage($stage, $desired_states, $show_filtered_projects = fals
             }
             echo "</td>";
         }
+    } elseif ($stage->id == 'SR') {
+        echo "<td colspan='3'>$total_projects</td>";
     } else {
         echo "<td colspan='3' class='nocell'></td>";
     }


srjfoo:
I'd especially like to thank @cpeel for helping -- I'd been beating my head against this for long enough that I was making no progress on getting the filtering sorted out.

I'm planning to run this past at least the SR coordinators for review, as well.

srjfoo:
One other thing to be aware of -- because the expected state for both "in PP" and "in SR" is `project_post_first_checked_out`, the "Total" and "In Progress" numbers for the PP stage include those projects that are in SR. That's always been the case, though I'm not sure how many people realize that. I did not add that explanation in this MR. I do plan to do so in a later one, but can move it into this one if it seems wiser to do so.

activity_hub.php
```diff
@@ -95,6 +95,23 @@
...
 }
 
+// SR projects have an additional qualifier beyond just the state
+// -- It can't be added to $n_projects_in_state_ because that
+// messes with the totals for the projects in PP, so create another
+// array just for SR.
+$sql = "
+    SELECT state, COUNT(*)
+    FROM projects
+    WHERE state = 'proj_post_first_checked_out' AND
```
chrismiceli, line 105: should we reference https://github.com/DistributedProofreaders/dproofreaders/blob/e29261c82e30c88fe38e7db9740dff8155c02f6a/pinc/project_states.inc#L282 instead just in case it ever changes or someone wonders where these states come from?

activity_hub.php
```diff
@@ -244,19 +261,36 @@ function progress_snapshot_table($show_filtered_projects, $show_filtering_links,
...
+            echo "<b>" . _("Filtered") . "</b>";
+        } else {
+            echo "<b>" . pgettext("all projects", "All") . "</b>";
+            if ($show_filtering_links) {
+                echo " | <a href='?show_filtered=1#progress_snapshot'>" . _("Filtered") . "</a>";
+            }
+        }
+        echo "</tr>\n";
+
+        $desired_states = ["proj_post_first_checked_out"];
```
chrismiceli, line 280: same here on reference to defined states

activity_hub.php
```diff
@@ -244,19 +261,36 @@ function progress_snapshot_table($show_filtered_projects, $show_filtering_links,
             continue;
         }
 
-        $desired_states = [];
+        echo "\n<tr>";
+        echo "<th class='activity-icon-header'>&nbsp;</th>";
```
chrismiceli, line 265: do we need the empty space for rendering?

activity_hub.php
```diff
@@ -298,8 +332,14 @@ function summarize_stage($stage, $desired_states, $show_filtered_projects = fals
         // already included in the following round's numbers.
         // (Use '@' to suppress "Undefined property" notice:
         // not every stage has a 'project_complete_state'.)
+        // SR is the odd-man-out, and the count needs to be pulled
+        // separately from the others since $stage_state is the same
+        // as for "in PP", but needs the smoothread_deadline to limit it.
         if ($stage_state == @$stage->project_complete_state) {
             $count = array_get($n_projects_transitioned_to_state_, $stage_state, 0);
+        } elseif ($stage->id == 'SR') {
```
chrismiceli, line 340: is there a way we could reference https://github.com/DistributedProofreaders/dproofreaders/blob/master/pinc/stages.inc#L584 in same light as stage references?

activity_hub.php
```diff
@@ -337,10 +377,16 @@ function summarize_stage($stage, $desired_states, $show_filtered_projects = fals
             }
         }
 
+        if ($stage->id == 'SR') {
+            $where_clause = "state = 'proj_post_first_checked_out' AND smoothread_deadline > UNIX_TIMESTAMP()";
```
chrismiceli, line 381: same here and on line above

chrismiceli:
> 
> 
> All activity stages except Smooth Reading show information in the site progress snapshot table on the Activity Hub. This adds the SR counts and implements the spirit of task #728.
> 
> Sandbox at [ah-sr](https://www.pgdp.org/~srjfoo/c/activity_hub.php)
> 
> Testers: note that you must have proofed at least 2 pages in order to see the filters (I decreased it just especially for this sandbox).

Is the sandbox link correct? The hub looked the same to me as TEST. Unless I don't have permission or something.

activity_hub.php
srjfoo: I'm not entirely sure what you're asking. Are you asking, "Do you have to refer to it by name?" If so, the answer is "I don't know, but ..."

SR is an odd bird. It was tacked[**hacked] on as a new stage, but wasn't given a separate state as the proofing rounds and the PP/PPV rounds were. To be in SR, a project must be in `PROJ_POST_FIRST_CHECKED_OUT` and must have a `smoothread_deadline` in the `projects` table that has not expired. That's the only way to uniquely identify it as being available for smooth reading.

I'd love for you to poke around and see if you can suggest something less hacky! :grin:

activity_hub.php
srjfoo: see previous comment

srjfoo:
> > All activity stages except Smooth Reading show information in the site progress snapshot table on the Activity Hub. This adds the SR counts and implements the spirit of task #728.
> > Sandbox at [ah-sr](https://www.pgdp.org/~srjfoo/c.branch/ah-sr/activity_hub.php)
> > Testers: note that you must have proofed at least 2 pages in order to see the filters (I decreased it just especially for this sandbox).
> 
> Is the sandbox link correct? The hub looked the same to me as TEST. Unless I don't have permission or something.

\<sigh\> ... I thought I got that fixed. I noticed it right after I opened the MR, but must not have gotten the edit saved. Try now. (I've corrected it in my quote here, too)

activity_hub.php
chrismiceli: I tried removing it using the DOM tools in the browser and notice no difference without it (the &nbsp;)

activity_hub.php
chrismiceli: Reading though, it actually doesn't appear we have some global define or constant for the id 'SR', so this seems fine.

srjfoo: Aaah, okay, that's what you meant. You know, I'm so used to the awareness that SR is _**not**_ a distinct state, that it never occurred to me to look, and it's always a good idea to check for things like that.

activity_hub.php
cpeel: I don't know but we probably don't. We presumably did waaaay back when it was developed in 2009.

activity_hub.php
```diff
@@ -95,6 +95,23 @@
...
+// array just for SR.
+$sql = "
+    SELECT state, COUNT(*)
+    FROM projects
+    WHERE state = 'proj_post_first_checked_out' AND
+        smoothread_deadline > UNIX_TIMESTAMP()
+";
+$result = DPDatabase::query($sql);
+
+$n_projects_in_sr_ = [];
```
cpeel, line 110: Rather than create a separate array (and I understand why you need them separate) consider using a special value for the "state" key in the `$n_projects_in_state_` array.

Something like:
```php
$sql = "
    SELECT COUNT(*)
    FROM projects
    WHERE state = 'proj_post_first_checked_out' AND
        smoothread_deadline > UNIX_TIMESTAMP()
";
[$n_projects_in_state_["SR"]] = mysqli_fetch_row($result);
```

activity_hub.php
srjfoo: Finally done

activity_hub.php
srjfoo: No, and upon looking at the other places in the code that had that, I have no clue why I stuck it in there in the first place. Fixed

activity_hub.php
srjfoo: Fixed

activity_hub.php
srjfoo: Finally done, and thanks for the suggestion. Though it took longer than it should because I went off in the wrong direction for a while. Still, a learning experience.

I've put it in as a second commit instead of squashing, in case anyone wants to look at before and after, but I'm also happy to squash the commits into one.

cpeel: This looks good to me if you want to squash it!

srjfoo: Done!

cpeel (approved):

chrismiceli (approved):

activity_hub.php
```diff
@@ -244,11 +255,30 @@ function progress_snapshot_table($show_filtered_projects, $show_filtering_links,
             continue;
         }
 
-        $desired_states = [];
+        echo "\n<tr>";
```
cpeel, line 258: After thinking about this some, we need to put this in an `if ($stage->id == "SR")` block to prevent breaking other sites that might have a stage that isn't SR. That way we retain the old functionality for non-SR stages and we use this new one for SR. Your updates to `summarize_stage()` already handle this case, we just need to update this spot.

srjfoo: Good point, and done.

PR #683, opened 2021-09-08
Update wording on AH and snapshot legend

srjfoo:
Both the tooltip for ⓘ on the Activity Hub and the description in site_progress_snapshot_legend.php say "are" eligible. This commit changes that to "may be" eligible (for example, most people are not eligible to work as PPVers, yet everyone sees that icon for the PPV activity. Add to the Activity description for SR an explanation that the SR count is included in the PP count (should this be with PP, instead, or in addition?).

[ah-cleanup](https://www.pgdp.org/~srjfoo/c.branch/ah-cleanup/activity_hub.php)

### DIFF
--- a/activity_hub.php
+++ b/activity_hub.php
@@ -306,7 +306,7 @@ function summarize_stage($stage, $desired_states, $show_filtered_projects = fals
     } elseif ($uao->all_minima_satisfied) {
         $access_class = "access-eligible";
         $access_icon = "ⓘ";
-        $access_text = _("You are eligible to work in this activity");
+        $access_text = _("You may be eligible to work in this activity");
         $access_link = "{$stage->relative_url}#Entrance_Requirements";
     } else {
         $access_class = "access-no";
@@ -496,7 +496,13 @@ function activity_descriptions()
 
     foreach ($Stage_for_id_ as $stage) {
         echo "<dt><b>$stage->id</b>: <a href='$code_url/{$stage->relative_url}'>{$stage->name}</a></dt>";
-        echo "<dd>{$stage->description}</dd>\n";
+        echo "<dd>{$stage->description}";
+        if (($stage->id == "PP") || ($stage->id == "SR")) {
+            echo " " .  _("The counts in the Snapshot table, above, for projects
+                available for SR are also included in the numbers of projects in
+                PP, and the PP total.");
+        }
+        echo "</dd>\n";
     }
 
     echo "</dl>";
--- a/faq/site_progress_snapshot_legend.php
+++ b/faq/site_progress_snapshot_legend.php
@@ -31,14 +31,14 @@ output_header(_('Site Progress Snapshot'), NO_STATSBAR);
 
 <p>In addition each activity has an icon that represents your current ability to work in that stage:
 <ul>
-  <li><div class='access-yes'>✓</div> - You have access to, and are able to work in, this stage.</li>
-  <li><div class='access-eligible'>ⓘ</div> - You satisfy the basic criteria and are eligible to work in this stage. Clicking on the icon will provide more information on how to obtain access.</li>
-  <li><div class='access-no'>✗</div> - You are not yet eligible to work in this stage. Clicking on the icon will show you the entrance requirements for that stage.</li>
+  <li><span class='access-yes' style='float: none;'>✓</span> - You have access to, and are able to work in, this stage.</li>
+  <li><span class='access-eligible' style='float: none;'>ⓘ</span> - You satisfy the basic criteria and may be eligible to work in this stage. Clicking on the icon in the Snapshot table will provide more information on how to obtain access.</li>
+  <li><span class='access-no' style='float: none;'>✗</span> - You are not yet eligible to work in this stage. Clicking on the icon in the Snapshot table will show you the entrance requirements for that stage.</li>
 </ul>
 </p>
 
 <h2>Project information</h2>
-<p>For each activity, information about projects in that activity is included. For projects in the proofreading rounds, this includes the Total, Waiting, and Available projects as well as a count of how many projects were Completed Today. For non-proofreading rounds, this includes the Total, Available, and In Progress projects.</p>
+<p>For each activity, information about projects in that activity is included. For projects in the proofreading rounds, this includes the Total, Waiting, and Available projects as well as a count of how many projects were Completed Today. For post-processing activities, this includes the Total, Available, and In Progress projects. For smooth reading, it displays the number of projects Available for SR.</p>
 
 <p>Project numbers can optionally be filtered using the filter criteria specified on corresponding pages. You can toggle between viewing project numbers for All projects or Filtered projects that match your filter.</p>
 


srjfoo:
Question: On the Activity Hub, in the Activity Description section, I chose to add the comment about SR numbers being included in the PP count. Should that comment be in both the PP and the SR sections?

activity_hub.php
```diff
@@ -496,7 +496,13 @@ function activity_descriptions()
 
     foreach ($Stage_for_id_ as $stage) {
         echo "<dt><b>$stage->id</b>: <a href='$code_url/{$stage->relative_url}'>{$stage->name}</a></dt>";
-        echo "<dd>{$stage->description}</dd>\n";
+        echo "<dd>{$stage->description}";
+        if ($stage->id == "SR") {
+            echo " The counts in the Snapshot table, above, for projects available
```
cpeel, line 501: Wrap this in `_()`.

cpeel (commented):
> Question: On the Activity Hub, in the Activity Description section, I chose to add the comment about SR numbers being included in the PP count. Should that comment be in both the PP and the SR sections?

I don't feel strongly either way.

activity_hub.php
srjfoo: Done, and thanks for spotting that!

faq/site_progress_snapshot_legend.php
```diff
@@ -31,14 +31,14 @@
 
 <p>In addition each activity has an icon that represents your current ability to work in that stage:
 <ul>
-  <li><div class='access-yes'>✓</div> - You have access to, and are able to work in, this stage.</li>
-  <li><div class='access-eligible'>ⓘ</div> - You satisfy the basic criteria and are eligible to work in this stage. Clicking on the icon will provide more information on how to obtain access.</li>
-  <li><div class='access-no'>✗</div> - You are not yet eligible to work in this stage. Clicking on the icon will show you the entrance requirements for that stage.</li>
+  <li><span class='access-yes' style='float: none;'>✓</span> - You have access to, and are able to work in, this stage.</li>
```
chrismiceli, line 34: Looks like access-* classes are only used in this faq and in the activity_hub:
https://github.com/DistributedProofreaders/dproofreaders/search?q=access-eligible
Should we instead just create a separate class to apply to the icons in activity_hub and avoid inline style here?

chrismiceli (approved):

faq/site_progress_snapshot_legend.php
srjfoo: The inline style, is, quite frankly, a bandaid. It's to override the "float right" that's used specifically for the snapshot table, so that the icons appear in their proper places in the list. My original plan was to update the text in this file, and the beginner help text on the Activity Hub, but the more I dug into it, the more I realized that would be overwhelming to beginners. After consulting with @lhamilton1 and @cpeel, I think that we've decided to create a new wiki page in the DP official documentation that describes the Snapshot Table on the Activity Hub and to remove this file altogether (and the beginner help text on the AH).

But the removal needs to wait until the wiki page is created -- so, another MR. I could have just left the file alone altogether, but am picky enough that I wanted the text to be correct and the icons to display properly, even if it's slated for the chopping block. 😕 

So, at this point, no, I don't think we need to create a separate class at this point (that may change, though).

lhamilton1:
It would make sense to have the info in the PP section too. Also -- " it can be optionally uploaded" might be shorter as "may be uploaded"

srjfoo:
> It would make sense to have the info in the PP section too. Also -- " it can be optionally uploaded" might be shorter as "may be uploaded"

I'll add it to the PP section, too. As for "it can be optionally uploaded", that's in a totally different part of the code, and is the same text that is displayed just above the list of projects available for [Smooth Reading on the SR pool page](https://www.pgdp.net/c/tools/post_proofers/smooth_reading.php#SR). I can make the change, I just wanted you to be aware that the change would appear in more than one place.

lhamilton1:
We can live with the optionally uploaded :)

srjfoo:
> > It would make sense to have the info in the PP section too. Also -- " it can be optionally uploaded" might be shorter as "may be uploaded"
> 
> I'll add it to the PP section, too.

And done.

activity_hub.php
```diff
@@ -496,7 +496,13 @@ function activity_descriptions()
 
     foreach ($Stage_for_id_ as $stage) {
         echo "<dt><b>$stage->id</b>: <a href='$code_url/{$stage->relative_url}'>{$stage->name}</a></dt>";
-        echo "<dd>{$stage->description}</dd>\n";
+        echo "<dd>{$stage->description}";
+        if (($stage->id == "PP") || ($stage->id == "SR")) {
+            echo _(" The counts in the Snapshot table, above, for projects available
```
cpeel, line 501: The leading space shouldn't be in the `_()` as it runs the risk of a translator removing it and it running together with the prior sentence. You probably want `echo " " . _("...");`

cpeel: I also can't remember where we are with multi-line translated strings. IIRC most of them are all in one line to avoid this?

faq/site_progress_snapshot_legend.php
cpeel: Agreed that doing this inline is probably the best bandaid right now given the future direction.

activity_hub.php
srjfoo: The problem with trying to put this part of the string in the same line (or paragraph) is that the base descriptions for all of these is in `stages.inc`, and it's not really appropriate (at least for PP) to add that string to the description. Conceivably, it might be added to the SR blurb, but I doubt the SRers care, and it would just add bloat.

We could also just ignore the issue altogether and take the additional text out. If nobody has noticed in the 16 years since the SR pool first saw light of day that projects "in SR" are also "in PP", does it really matter? Anyone who questions it, we can describe how things work.

I have no problem with taking it back out again -- I was probably over-thinking the issue to begin with :grin: -- but if we leave it in, it should stay a separate string from the rest of the blurb (and yes, if the decision is to leave it in, I'll move the leading space out of the translated string -- thanks!).

We might also consider, when we get to the point of dealing with the beginner help and the snapshot legend, whether we even need the activity descriptions on the AH, and if so, whether they should be unique instead of drawing from `stages.inc`.

cpeel: I think you misunderstood my comment. I wasn't suggesting combining it into the line above, but putting all of the translated string on a single line to be translated:
```php
echo " " . _("The counts in the Snapshot table, above, for projects available for SR are also included in the numbers of projects in PP, and the PP total.");
```

I forget out xgettext and our translation tooling handles multi-line strings.

srjfoo: Ah, okay. Yes, according to the [String localization Best practices](https://www.pgdp.net/wiki/DP_Code_Best_Practices#String_localization), it should handle multi-line strings:

* You may break the string over multiple lines, so as to avoid too long lines. But if you do so, keep it a single string literal (otherwise xgettext will not pick up the correct string).

Mind you, I haven't recently experimentally verified that, but I can easily do so.

srjfoo: I've now verified that it seems to work fine. It doesn't look pretty (especially in the `.po` file as viewed in a text editor) even in poedit:

image

But renders fine on the AH:
image

\<cough\> ... now I should probably go back and generate a clean POT, huh?

cpeel: As long as translators know they don't need to bother with newlines or whitespaces, that's fine.

PR #684, opened 2021-09-09
Update translate.php for UTF-8

srjfoo:
The `faq/translate.php` document still assumed Latin-1. Update it for UTF-8.

[update-translate-doc/faq/translate.php](https://www.pgdp.org/~srjfoo/c.branch/update-translate-doc/faq/translate.php)


cpeel:
Thanks for finding and fixing these!

PR #685, opened 2021-09-11
Document graphing library

chrismiceli:
remove chart references in favor of graph

https://www.pgdp.org/~chrismiceli/c.branch/documentation/stats/stats_central.php

### DIFF
--- a/SETUP/GRAPH.md
+++ b/SETUP/GRAPH.md
@@ -1,0 +1,138 @@
+# Graphs
+
+This file contains a description of the way graphs are created in dproofreaders.
+
+In case of problems, please post to the [DP Site Code](http://www.pgdp.net/phpBB3/viewforum.php?f=32)
+forum at the [primary DP site](https://www.pgdp.net/c).
+
+## Overview
+Graphs are drawn on the browser using a set of functions from [svgGraphs.js](../scripts/svgGraphs.js) that interact with the [d3js](https://d3js.org/) library to render [Scalable Vector Graphics (svg)](https://en.wikipedia.org/wiki/Scalable_Vector_Graphics) elements. The 4 types of graphs that are supported are
+1. Bar
+1. Line
+1. Pie
+1. Stacked Area
+
+These graphs are configured using a plain old javascript object. To facilitate creating the javascript to render a graph from PHP, a php utility method [build_svg_graph_inits](../pinc/graph_data.inc) was created. It is a function that takes an array of arrays representing the graphs to render. Each graph to render is a PHP array where the entries are
+1. type - type of graph to be rendered. Either `barLineGraph`, `pieGraph`, or `stackedAreaGraph`
+1. id - id of the div on the page to render the graph into
+1. configuration - PHP array configuring the graph. Expanded on below.
+Example Usage:
+
+```
+build_svg_graph_inits([["barLineGraph", "myGraphId", $graphConfig]]);
+```
+
+This will give you a string you can use to pass to js_data in the [output_header](../pinc/theme.inc) function. Something like
+
+```
+"$(function(){barLineGraph('myGraphId', graphConfigEncodedAsJSON)});"
+```
+
+# Configuration
+Each graph type supports a configuration object that allows control of the appearance of the graph as well as the data to the graph. Each graph type respects different configuration properties.
+## Bar/Line Graph
+```
+{
+  // Width of the graph in pixels. Defaults to 640 if not provided.
+  width?: number;
+  
+  // Height of the graph in pixels. Defaults to 400 if not provided.
+  height?: number;
+  
+  // Width of y axis in pixels. Defaults to 50 if not provided.
+  yAxisWidth?: number;
+
+  // Y axis label. Defaults to first series label or legend box for multiple series.
+  yAxisLabel?: string,
+  
+  // Height of x axis in pixels. Defaults to 50 if not provided.
+  xAxisHeight?: number;
+  
+  // Maximum bumber of y axis ticks to show. Defaults to d3js implementation.
+  yAxisTickCount?: number;
+  
+  // Should be set to true when multiple bar series are set in data that share.
+  // the same x values;
+  groupBars?: boolean;
+
+  // An array of color values for the series to be colored.
+  barColors?: string[];
+
+  // Whether a border should be drawn on the bars in the series.
+  barBorder?: boolean;
+
+  // Title of the graph, shown at the top center.
+  title?: string,
+
+  // Offset to legend box drawn when there are multiple series.
+  legendAdjustment?: {
+    x: number; // x offset for legend box
+    y: number; // y offset for legend box
+  }
+
+  // A value indicating whether the y axis label should be drawn at the bottom.
+  // Useful for error scenarios.
+  bottomLegend?: boolean;
+
+  // Data to render
+  data: {
+    [seriesTitle: string]: {
+      x: string[]; // X axis labels
+      y: string[]; // Y axis values, converted to a number
+    }
+  };
+}
+```
+
+## Pie Graph
+```
+  // Width of the graph in pixels. Defaults to 640 if not provided.
+  width?: number;
+  
+  // Height of the graph in pixels. Defaults to 400 if not provided.
+  height?: number;
+  
+  // Title of the graph, shown at the top center.
+  title?: string,
+
+  // Offset to legend box drawn when there are multiple series.
+  legendAdjustment?: {
+    x: number; // x offset for legend box
+    y: number; // y offset for legend box
+  }
+
+  // Message to show in the case there isn't data to render.
+  error?: string;
+
+  // Labels of pie wedges
+  labels: string[];
+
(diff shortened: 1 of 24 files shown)


SETUP/GRAPH.md
```diff
@@ -0,0 +1,138 @@
...
+  title?: string,
+
+  // Offset to legend box drawn when there are multiple series.
+  legendAdjustment?: {
+    x: number; // x offset for legend box
+    y: number; // y offset for legend box
+  }
+
+  // Message to show in the case there isn't data to render.
+  error?: string;
```
srjfoo, line 105: Am I missing something in the others? Many of the graphs may wind up with no data, particularly on the first day of the month. Or are those all handled in other ways?

SETUP/GRAPH.md
```diff
@@ -0,0 +1,138 @@
+# Graphs
```
cpeel, line 1: Please line-wrap this file at 80 columns to match our other markdown files.

cpeel: I would also call it `GRAPHS.md` (plurals).

SETUP/GRAPH.md
```diff
@@ -0,0 +1,138 @@
...
+
+This file contains a description of the way graphs are created in dproofreaders.
+
+In case of problems, please post to the [DP Site Code](http://www.pgdp.net/phpBB3/viewforum.php?f=32)
+forum at the [primary DP site](https://www.pgdp.net/c).
+
+## Overview
+Graphs are drawn on the browser using a set of functions from [svgGraphs.js](../scripts/svgGraphs.js) that interact with the [d3js](https://d3js.org/) library to render [Scalable Vector Graphics (svg)](https://en.wikipedia.org/wiki/Scalable_Vector_Graphics) elements. The 4 types of graphs that are supported are
+1. Bar
+1. Line
```
cpeel, line 11: For humans reading this in source code form, not just rendered form, please increment the numbers. Or, since the order doesn't matter, ideally change them to a bulleted list `*`.

cpeel: Ditto below.

SETUP/GRAPH.md
```diff
@@ -0,0 +1,138 @@
...
+1. Pie
+1. Stacked Area
+
+These graphs are configured using a plain old javascript object. To facilitate creating the javascript to render a graph from PHP, a php utility method [build_svg_graph_inits](../pinc/graph_data.inc) was created. It is a function that takes an array of arrays representing the graphs to render. Each graph to render is a PHP array where the entries are
+1. type - type of graph to be rendered. Either `barLineGraph`, `pieGraph`, or `stackedAreaGraph`
+1. id - id of the div on the page to render the graph into
+1. configuration - PHP array configuring the graph. Expanded on below.
+Example Usage:
+
+```
```
cpeel, line 21: Consider using ` ```php ` for styling these blocks.

SETUP/GRAPH.md
```diff
@@ -0,0 +1,138 @@
...
+This will give you a string you can use to pass to js_data in the [output_header](../pinc/theme.inc) function. Something like
+
+```
+"$(function(){barLineGraph('myGraphId', graphConfigEncodedAsJSON)});"
+```
+
+# Configuration
+Each graph type supports a configuration object that allows control of the appearance of the graph as well as the data to the graph. Each graph type respects different configuration properties.
+## Bar/Line Graph
+```
```
cpeel, line 34: Consider ` ```js ` (at least it looks like valid JS to me?)

SETUP/GRAPH.md
cpeel: Let's also add a reference to this fine in `SETUP/README.md`.

SETUP/GRAPH.md
```diff
@@ -0,0 +1,138 @@
...
+forum at the [primary DP site](https://www.pgdp.net/c).
+
+## Overview
+Graphs are drawn on the browser using a set of functions from [svgGraphs.js](../scripts/svgGraphs.js) that interact with the [d3js](https://d3js.org/) library to render [Scalable Vector Graphics (svg)](https://en.wikipedia.org/wiki/Scalable_Vector_Graphics) elements. The 4 types of graphs that are supported are
+1. Bar
+1. Line
+1. Pie
+1. Stacked Area
+
+These graphs are configured using a plain old javascript object. To facilitate creating the javascript to render a graph from PHP, a php utility method [build_svg_graph_inits](../pinc/graph_data.inc) was created. It is a function that takes an array of arrays representing the graphs to render. Each graph to render is a PHP array where the entries are
```
cpeel, line 15: s/php/PHP/ to match the second instance in the para.

SETUP/GRAPH.md
```diff
@@ -0,0 +1,138 @@
+# Graphs
+
+This file contains a description of the way graphs are created in dproofreaders.
+
+In case of problems, please post to the [DP Site Code](http://www.pgdp.net/phpBB3/viewforum.php?f=32)
+forum at the [primary DP site](https://www.pgdp.net/c).
+
+## Overview
+Graphs are drawn on the browser using a set of functions from [svgGraphs.js](../scripts/svgGraphs.js) that interact with the [d3js](https://d3js.org/) library to render [Scalable Vector Graphics (svg)](https://en.wikipedia.org/wiki/Scalable_Vector_Graphics) elements. The 4 types of graphs that are supported are
```
cpeel, line 9: Somewhere, probably in the Overview, we should talk a bit about embedded vs iframe graphs, at least in the sense that there are PHP pages that only contain the graph that can be used as iframes within a page whereas most of the graphs are rendered within a larger page.

My thinking is that in the future a developer might come in and say "I want to make this graph accessible as if it were an image" and being able to easily see where we've done this prior via iframe would be useful.

SETUP/GRAPH.md
chrismiceli: If there is no data, the graphing code I wrote is ideally capable of just not rendering anything but empty axis and title, which is better than jpgraph today. The reason pie graphs have this is they handled this case before, so I ported the logic.

SETUP/GRAPH.md
chrismiceli: That's amazing

SETUP/GRAPH.md
chrismiceli: It is essentially typescript so I can define the shapes of the objects. I'll annotate it as such.

SETUP/GRAPHS.md
```diff
@@ -0,0 +1,173 @@
...
+creating the javascript to render a graph from PHP, a PHP utility method
+[build_svg_graph_inits](../pinc/graph_data.inc) was created. It is a function
+that takes an array of arrays representing the graphs to render. Each graph to
+render is a PHP array where the entries are
+
+* type - type of graph to be rendered. Either `barLineGraph`, `pieGraph`, or
+`stackedAreaGraph`
+* id - id of the div on the page to render the graph into
+* configuration - PHP array configuring the graph. Expanded on below.
+Example Usage:
```
cpeel, line 32: Please add a blank line above this.

cpeel (approved):

cpeel:
Looks great!

SETUP/GRAPHS.md
```diff
@@ -0,0 +1,174 @@
...
+standalone page that renders just the graph, and then embed that page as an
+iframe. You may want to style the iFrame to remove any iFrame borders to have a
+seamless image like experience.
+
+One such page for example of a page designed for being embedded as an iFrame is
+[round_backlog_days.php](../stats/round_backlog_days.php). It can be embedded
+in a news item with the following html.
+
+```html
+<iframe src="/stats/round_backlog_days.php" style="width:320px;height:205px;border:0;">
```
srjfoo, line 60: I'm going to be picky again, but you're welcome to ignore me, because I know that the code works either way.

How should attributes be styled in the code? I.e. spaced for easier reading, or all squeezed together. I didn't notice this in the code before seeing it smushed together here, but in looking at the code, I see that's how it would be written out. At the same time, some of the other attributes have a space both between the attribute and the value, and following the semicolon.

I didn't notice anything talking about that in the style guide, so I guess partly I'm asking if there's a particular format we should aim for.

That being said, everything looks good to me other than that inconsistency.

srjfoo (approved):

SETUP/GRAPHS.md
chrismiceli: I think it depends on what your goal is. Removing spaces makes the HTML smaller (barely) but reduces legibility. Personally I think the spaces should probably be there since we don't aggressively attempt to make our HTML, JS, nor CSS minimal.

chrismiceli: That being said, I can update this in a subsequent MR when I fix the min width of bars.

srjfoo: I prefer legibility, since it's not a significant reduction in size.

chrismiceli: went ahead and implemented the change since this wasn't merged yet.

PR #686, opened 2021-09-19
change to using canvas to draw image, add rotation

70ray:
This is simpler than the previous version but there are two issues:

Using Firefox on Linux, when reducing the size of the image, if the image is entirely contained in the window (i.e. there are no scroll bars) part of the old, larger, image is left behind. Any action which causes the screen to be redrawn, such as pressing the alt key or changing the window size redraws correctly. This appears to be a bug in Firefox on Linux since I have not seen it on chromium on linux or Firefox, chrome, edge on windows.
I previously did not realize this was a localized bug so added the complexity of the previous version to work round it.
I have reported the bug to Mozilla.

Scrollbar issue: This does not apply to Safari where it appears that temporary scrollbars are drawn on top of the window contents. On other browsers the scrollbars take space from the window so when fitting the image to width, if a vertical scrollbar appears the client width is reduced so a horizontal scrollbar will also appear. Similarly for fitting to height. This is a minor problem. Although it could be addressed by measuring the scrollbar width and doing some calculations, the extra complexity may not be justified. 

Sandbox at: https://www.pgdp.org/~rp31/c.branch/image_rotate_1

### DIFF
--- a/scripts/page_browse.js
+++ b/scripts/page_browse.js
@@ -420,7 +420,7 @@ function pageBrowse(params, storageKey, replaceUrl, mentorMode = false) {
                     }
 
                     const imageDiv = $("<div>");
-                    imageWidget = makeImageWidget(imageDiv);
+                    imageWidget = makeImageWidget(imageDiv, "C");
                     imageWidget.setup(storageKey);
                     stretchDiv.append(imageDiv);
                     showImageText();
@@ -440,9 +440,9 @@ function pageBrowse(params, storageKey, replaceUrl, mentorMode = false) {
                             break;
                         case "imageText": {
                             const imageDiv = $("<div>");
-                            imageWidget = makeImageWidget(imageDiv);
                             stretchDiv.append(imageDiv, textDiv);
                             const theSplitter = viewSplitter(stretchDiv, storageKey);
+                            imageWidget = makeImageWidget(imageDiv, "C");
                             if(mentorMode) {
                                 // make a text widget with splitter
                                 textWidget = makeTextWidget(textDiv, true, theSplitter.mainSplit.reSize);
--- a/scripts/control_bar.js
+++ b/scripts/control_bar.js
@@ -2,7 +2,7 @@
 /* exported makeImageWidget */
 
 // Construct the image sizing controls.
-var makeImageControl = function(imageElement) {
+var makeImageControl = function(content) {
     let imageKey;
     // percent need not be an integer but is rounded for display and save
     // it will typically not be an integer after fit height or width or + or -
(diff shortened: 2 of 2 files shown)


scripts/page_browse.js
```diff
@@ -420,7 +420,7 @@ function pageBrowse(params, storageKey, replaceUrl, mentorMode = false) {
                     }
 
                     const imageDiv = $("<div>");
-                    imageWidget = makeImageWidget(imageDiv);
+                    imageWidget = makeImageWidget(imageDiv, "C");
```
chrismiceli, line 423: isn't "C" the default?

chrismiceli (commented):
I'd feel more comfortable if we could avoid the <canvas based on the issue in slack and the previous issue with open image in new tab url. Can we not achieve the rotate with https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotate()

70ray:
Closing this because of downsizing quality problem.

70ray:
> I'd feel more comfortable if we could avoid the <canvas based on the issue in slack and the previous issue with open image in new tab url. Can we not achieve the rotate with https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotate()

There was another problem with the css approach but I'll look at it again.
I never understood what the "open image in new tab url" problem was but if we use the css method it will presumably go away.

chrismiceli:
> 
> 
> > I'd feel more comfortable if we could avoid the <canvas based on the issue in slack and the previous issue with open image in new tab url. Can we not achieve the rotate with https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotate()
> 
> There was another problem with the css approach but I'll look at it again.
> I never understood what the "open image in new tab url" problem was but if we use the css method it will presumably go away.

I think the problem was mainly the url in the new tab was blob:// instead of link to raw image.

srjfoo:
> > > I'd feel more comfortable if we could avoid the <canvas based on the issue in slack and the previous issue with open image in new tab url. Can we not achieve the rotate with https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotate()
> > 
> > 
> > There was another problem with the css approach but I'll look at it again.
> > I never understood what the "open image in new tab url" problem was but if we use the css method it will presumably go away.
> 
> I think the problem was mainly the url in the new tab was blob:// instead of link to raw image.

Yes. I double-checked in FF in Ray's sandbox just now, and "open image in new tab" is available, **_BUT_** what it opens is a copy of the current image, at the size it's being displayed in the page browser. What's implemented on PROD and the main TEST sandbox opens the image in a new tab at the image's native size.

I had remembered that there were problems with the canvas approach, but didn't remember what all the issues were. We certainly didn't uncover the image degradition bit last time we were testing code that used it.

scripts/page_browse.js
70ray: Yes, at an intermediate stage I added another parameter so needed to specify this one, then removed it again and later sqaushed all the commits, but it got left behind.

70ray:
> ... "open image in new tab" is available, **_BUT_** what it opens is a copy of the current image, at the size it's being displayed in the page browser. What's implemented on PROD and the main TEST sandbox opens the image in a new tab at the image's native size.

Thanks.

PR #687, opened 2021-09-20
Graph theme

chrismiceli:
https://www.pgdp.org/~chrismiceli/c.branch/graph-theme/

Improve area charts with lots of points by using step function instead of line
support themeing (thought keeping the colors for now the same for all themes)
update documentation
prevent scrollbars on iframe when tooltip hovers near edge.

### DIFF
--- a/styles/themes/charcoal.less
+++ b/styles/themes/charcoal.less
@@ -21,6 +21,8 @@
 @active-color: #ff66aa;
 @sans-serif-font: Verdana, Helvetica, sans-serif;
 
+@graph-series-colors: #1f77b4, #ff7f0e, #2ca02c, #d62728, #9467bd, #8c564b, #e377c2, #7f7f7f, #bcbd22, #17becf;
+
 /* Do not underline links */
 a {
     text-decoration: none;
--- a/SETUP/GRAPHS.md
+++ b/SETUP/GRAPHS.md
@@ -166,7 +166,7 @@ respects different configuration properties.
   // Data to render
   data: {
     [seriesTitle: string]: {
-      x: string[]; // X axis labels
+      x: string[]; // X axis labels, converted to a date
       y: string[]; // Y axis values, converted to a number
     };
   };
--- a/scripts/svgGraphs.js
+++ b/scripts/svgGraphs.js
@@ -122,6 +122,7 @@ const {barLineGraph, stackedAreaGraph, pieGraph} = (function () {
             .range(d3.schemeCategory10);
 
         const area = d3.area()
+            .curve(d3.curveStep)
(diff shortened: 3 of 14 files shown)


srjfoo:
> Improve area charts with lots of points by using step function instead of line

Have these just not been adapted yet? The blue lines are still all but invisible.
image

-- also for the user's stats:
image

chrismiceli:
> 
> 
> > Improve area charts with lots of points by using step function instead of line
> 
> Have these just not been adapted yet? The blue lines are still all but invisible.
> image
> 
> -- also for the user's stats:
> image

It is improved, but not finished. You can see the improvement more on https://www.pgdp.org/~chrismiceli/c.branch/graph-theme/stats/user_logon_stats.php where the points aren't triangular anymore as I'm used the step line algorithm (https://github.com/d3/d3-shape/blob/v3.0.1/README.md#curveStep) instead of a linear algorithm (https://github.com/d3/d3-shape/blob/v3.0.1/README.md#curveLinear).

srjfoo:
> It is improved, but not finished. You can see the improvement more on https://www.pgdp.org/~chrismiceli/c.branch/graph-theme/stats/user_logon_stats.php where the points aren't triangular anymore as I'm used the step line algorithm (https://github.com/d3/d3-shape/blob/v3.0.1/README.md#curveStep) instead of a linear algorithm (https://github.com/d3/d3-shape/blob/v3.0.1/README.md#curveLinear).

👍 -- Aha! Much better, yes.

cpeel (approved):
This looks good to me code-wise!

styles/themes/charcoal.less
```diff
@@ -21,6 +21,8 @@
 @active-color: #ff66aa;
 @sans-serif-font: Verdana, Helvetica, sans-serif;
 
+@graph-series-colors: #1f77b4, #ff7f0e, #2ca02c, #d62728, #9467bd, #8c564b, #e377c2, #7f7f7f, #bcbd22, #17becf;
```
srjfoo, line 24: It's entirely appropriate for `@graph-series-colors` to be her for `charcoal.less`, but in general, for the standard light themes, we've kept the defaults in `layout.less` (and `page_interfaces.less`), and put overrides in any non-standard themes. Would it be appropriate at some point to move this piece of code into `layout.less` at some point for the other three themes?

srjfoo (approved):

styles/themes/charcoal.less
chrismiceli: that makes sense! I was wondering what the rationale was.

srjfoo (approved):

PR #688, opened 2021-09-22
Image rotate css

70ray:
This should avoid the image degradtion of the canvas method.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/image_rotate_css

### DIFF
--- a/scripts/control_bar.js
+++ b/scripts/control_bar.js
@@ -3,8 +3,13 @@
 
 // Construct the image sizing controls.
 var makeImageControl = function(content) {
-    const imageElement = $("<img>", {class: "middle-align"}).css("cursor", "grab");
-    content.append(imageElement);
+    // use plain js image so width or style.width is clear
+    const image = document.createElement("img");
+    image.classList.add("middle-align");
+    image.style.cursor = "grab";
+    const imageDiv = $("<div>", {class: "middle-align"}).css({overflow: "hidden", display: "inline-block"});
+    content.append(imageDiv);
+    imageDiv.append(image);
 
     let scrollDiffX = 0;
     let scrollDiffY = 0;
@@ -15,12 +20,12 @@ var makeImageControl = function(content) {
 
     function mouseup() {
         $(document).unbind("mousemove mouseup");
-        imageElement.css("cursor", "grab");
+        image.style.cursor = "grab";
     }
 
-    imageElement.mousedown( function(event) {
+    $(image).mousedown( function(event) {
         event.preventDefault();
-        imageElement.css("cursor", "grabbing");
+        image.style.cursor = "grabbing";
         scrollDiffX = event.pageX + content.scrollLeft();
         scrollDiffY = event.pageY + content.scrollTop();
         $(document).on("mousemove", mousemove)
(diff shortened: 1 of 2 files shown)


chrismiceli (approved):
Looks good to me. My only concern is anticlockwise vs counterclosewise, but that doesn't really matter :).

chrismiceli:
Can we squash some of the commits though?

scripts/control_bar.js
```diff
@@ -2,74 +2,149 @@
...
+    });
+
     let imageKey;
     // percent need not be an integer but is rounded for display and save
     // it will typically not be an integer after fit height or width or + or -
     let percent;
     const minPercent = 10;
     const maxPercent = 999;
     const defaultPercent = 100;
+    let sine = 0;
```
cpeel, line 49: While I understand these are ultimately the values used in the CSS, it would be much more intuitive in the code if we stored a clockwise rotation angle (0, 90, 180, 270) and converted that into sine/cosine in `setImageStyle()` via a lookup table.

scripts/control_bar.js
```diff
@@ -78,7 +153,9 @@ var makeImageControl = function(imageElement) {
             $("<span>", {class: "nowrap"}).append(percentInput, "%"),
             " ",
             zoomIn,
-            zoomOut
+            zoomOut,
+            clockRotateInput,
```
cpeel, line 157: Consider putting in a spacer (`" "`) so there's some room between the zoom and the rotation control pairs.

cpeel (commented):
No need to worry about the commits, I'll squash them all when this goes in.

70ray:
> Looks good to me. My only concern is anticlockwise vs counterclosewise, but that doesn't really matter :).

Is this British English vs. US English? If it is I suppose we should use US English and one day provide a translation to British English.

srjfoo:
> > Looks good to me. My only concern is anticlockwise vs counterclosewise, but that doesn't really matter :).
> 
> Is this British English vs. US English? If it is I suppose we should use US English and one day provide a translation to British English.

Yes, I believe counterclockwise is US English. At least, like Chris, that's what I grew up using. (Except where I saw 'widdershins' being used -- but that also comes from the British Isles. 😁 )

chrismiceli:
> widdershins

Yeah, definitely a brittish US vs British thing. I am not sure the voice of pgdp, but I assumed US since that's the copyright laws we follow. As for widdershins, I learn something each day :)

scripts/control_bar.js
70ray: We could do that but I think this method is simpler and avoids having to reset the index if the user rotates through more than 360.

scripts/control_bar.js
70ray: Done this.

70ray:
> > widdershins
> 
> Yeah, definitely a brittish US vs British thing. I am not sure the voice of pgdp, but I assumed US since that's the copyright laws we follow. As for widdershins, I learn something each day :)

Changed to counterclockwise.

srjfoo:
> > > widdershins
> > 
> > 
> > Yeah, definitely a brittish US vs British thing. I am not sure the voice of pgdp, but I assumed US since that's the copyright laws we follow. As for widdershins, I learn something each day :)
> 
> Changed to counterclockwise.

``` php
"anticlockRotate" => _("Rotate image counterclockwise 90°"),
```
Should probably change index name, also?

70ray:
> > > > widdershins
> > > 
> > > 
> > > Yeah, definitely a brittish US vs British thing. I am not sure the voice of pgdp, but I assumed US since that's the copyright laws we follow. As for widdershins, I learn something each day :)
> > 
> > 
> > Changed to counterclockwise.
> 
> ```
> "anticlockRotate" => _("Rotate image counterclockwise 90°"),
> ```
> 
> Should probably change index name, also?

That would seem reasonable.

scripts/control_bar.js
cpeel: That's fine. Please add a block comment here that maps the sine/cosine values to their rotation angle to help the next person who reads the code. Something like:
```js
// sine | cosine | clockwise rotation
// ----------------------------------
//   x  |   y    | z degrees
//   x  |   y    | z degrees
//   x  |   y    | z degrees
//   x  |   y    | z degrees
```

70ray: Good idea, I'll do that.

srjfoo (approved):
I've tested in the page browser and in the standard and enhanced proofreading interfaces. Image quality looks good, and opening the plain image in a separate tab works as expected, (and as expected), whether from the page browser or one of the proofreading interfaces.

Is there anything else that needs to be tested?

chrismiceli (approved):

PR #692, opened 2021-09-24
enable image move with arrow keys

70ray:
After clicking on the image it can be moved with the arrow keys as well as moved by dragging.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/image_drag_focus

### DIFF
--- a/scripts/control_bar.js
+++ b/scripts/control_bar.js
@@ -32,6 +32,9 @@ var makeImageControl = function(content) {
 
     $(image).mousedown( function(event) {
         event.preventDefault();
+
+        // so image can be moved with arrow keys
+        content[0].focus();
         image.style.cursor = "grabbing";
         scrollDiffX = event.pageX + content.scrollLeft();
         scrollDiffY = event.pageY + content.scrollTop();


srjfoo (approved):
Works for me.

scripts/control_bar.js
```diff
@@ -32,6 +32,9 @@ var makeImageControl = function(content) {
 
     $(image).mousedown( function(event) {
```
chrismiceli, line 33: would it be possible to set the mousedown on the content[0] div containing the img? That way we don't need to explicitly set focus as it would be on that component perhaps? Maybe not.

srjfoo: Bear in mind that I'm a total ignoramus as far as js goes -- I'm not entirely sure what this means? Does this mean default to image having focus when the window opens, or something else, entirely?

70ray: I'll try it.

70ray: I tried Chris's suggestion which I believe is just intended to simplify the code without changing the behaviour. The cursor drag works but the line `content[0].focus();` is still needed to make the arrow keys work. I don't understand why, peraps something to do with tabindex.

chrismiceli (approved):

scripts/control_bar.js
chrismiceli: worth a shot. thanks for trying it out.

PR #693, opened 2021-09-26
mozilla observer changes

chrismiceli:
Random recommendations from a somewhat strict mozilla tool for website security
https://observatory.mozilla.org/analyze/www.pgdp.org

### DIFF
--- a/pinc/base.inc
+++ b/pinc/base.inc
@@ -23,6 +23,10 @@ if (!headers_sent()) {
     // Disallow other sites from embedding pages in frames/iframes
     header("Content-Security-Policy: frame-ancestors 'self'", false);
     header("X-Frame-Options: SAMEORIGIN");
+    header("X-Content-Type-Options: nosniff");
+    if ($use_secure_cookies) {
+      header("Strict-Transport-Security: max-age=63072000; includeSubDomains; preload");
+    }
 }
 
 // If we don't have a database connection, don't try to resume the session.
--- a/pinc/base.inc
+++ b/pinc/base.inc
@@ -23,10 +23,6 @@ if (!headers_sent()) {
     // Disallow other sites from embedding pages in frames/iframes
     header("Content-Security-Policy: frame-ancestors 'self'", false);
     header("X-Frame-Options: SAMEORIGIN");
-    header("X-Content-Type-Options: nosniff");
-    if ($use_secure_cookies) {
-      header("Strict-Transport-Security: max-age=63072000; includeSubDomains; preload");
-    }
 }
 
 // If we don't have a database connection, don't try to resume the session.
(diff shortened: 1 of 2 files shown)


pinc/dpsession_via_php_sessions.inc
```diff
@@ -5,6 +5,8 @@ ini_set('session.name', "{$site_abbreviation}_Session");
 ini_set('session.use_cookies', 1);
 ini_set('session.cookie_domain', ''); // A sensible value will be used.
 ini_set('session.cookie_lifetime', 86400 * 31); // 31 days ~= 1 month
+ini_set('session.cookie_secure', $use_secure_cookies);
+ini_set('session.cookie_samesite', 'Lax');
```
chrismiceli, line 9: this is for cookies set before you login by php.

pinc/base.inc
```diff
@@ -23,6 +23,10 @@ if (!headers_sent()) {
     // Disallow other sites from embedding pages in frames/iframes
     header("Content-Security-Policy: frame-ancestors 'self'", false);
     header("X-Frame-Options: SAMEORIGIN");
+    header("X-Content-Type-Options: nosniff");
+    if ($use_secure_cookies) {
+      header("Strict-Transport-Security: max-age=63072000; includeSubDomains; preload");
```
cpeel, line 28: While pgdp.net does HTTP to HTTPS redirect, not all sites that use the code may. This would be better served in the `pgdp-production` branch that is just for pgdp.net.

pinc/base.inc
```diff
@@ -23,6 +23,10 @@ if (!headers_sent()) {
     // Disallow other sites from embedding pages in frames/iframes
     header("Content-Security-Policy: frame-ancestors 'self'", false);
     header("X-Frame-Options: SAMEORIGIN");
+    header("X-Content-Type-Options: nosniff");
```
cpeel, line 26: This appears to enable [CORB](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/services/network/cross_origin_read_blocking_explainer.md) which may, or may not, be problematic for pgdp.net or other sites. We should do a bit of investigation and testing to confirm this won't break anything.

pinc/base.inc
cpeel: This also needs 2 more spaces indent.

chrismiceli: I removed these. We can create a separate PR for prod hacks and investigate nosniff separately.

pinc/base.inc
chrismiceli: I'll do that one in a separate MR later once I get some testing on it.

cpeel (approved):

srjfoo (approved):

PR #694, opened 2021-10-01
Cache complex graph data

cpeel:
Cache graph data that takes a long time to generate in memcached. If memcached is not reachable (not installed, etc) this happily falls through.

Notably, this speeds up the most complex round stats graph, but could be used for any graph data. The cache is set to 1 hour by default which seems reasonable. On TEST the compressed cache data is 33kb so pretty small.

Testable in [svg-graph-cache](https://www.pgdp.org/~cpeel/c.branch/svg-graph-cache/stats/pages_proofed_graphs.php?tally_name=P1) sandbox. It will be important to test with different languages to ensure that the cached page includes strings in the user's desired language.


srjfoo (approved):

chrismiceli (approved):

srjfoo:
> ... It will be important to test with different languages to ensure that the cached page includes strings in the user's desired language.

Other than French, the localizations aren't really terribly up-to-date, and other than French and Icelandic, none of the non-English languages were enabled (I think that they weren't enabled long ago because when we were working on the UTF-8 testing, there was a time when we didn't have all the locales installed.

I do have them enabled now, though, and have verified that I can see localized strings in at least a subset of the graphs.

PR #695, opened 2021-10-03
Latex preview

70ray:
This is first draft and not intended to be merged but to invite criticism of the way it works so far.
It is proposed that a "preamble" text field be added to the project to contain project-specific stuff but this is not yet implemented. It could be edited in the same way as the other project details.
The Latex preview intercepts the format preview for a Latex book. At present this is detected by "LaTeX" in the title but would be better to detect a non-empty preamble field in the project.
Since the files are in a subdirectory it should be secure if we use the paranoid settings for reading and writing.

### DIFF
--- a/api/v1_projects.inc
+++ b/api/v1_projects.inc
@@ -327,31 +327,48 @@ function api_v1_project_pdflatex($method, $data, $query_params)
     if(!is_dir($temp_path)) {
         mkdir($temp_path);
     }
+    chdir($temp_path);
+
+/*  $files = glob("*");
+            foreach ($files as $file) {
+                unlink($file);
+            }*/
+
     $page_text = array_get($query_params, "text", "no text");
-    // TODO incorporate preamble and postamble
-    // TODO incorporate project-specific preamble
 
     $tmp_file_path = tempnam($temp_path, "");
-    $pdf_base_name = basename($tmp_file_path) . ".pdf";
-    $tex_file_path = $tmp_file_path . '.tex';
-    rename($tmp_file_path, $tex_file_path);
-    file_put_contents($tex_file_path, $page_text);
+    $base_name = basename($tmp_file_path);
+    $pdf_base_name = $base_name . ".pdf";
+    $tex_base_name = $base_name . '.tex';
+    rename($base_name, $tex_base_name);
+
+    // incorporate preamble and postamble
+    $preamble =
+'\\documentclass[12pt,oneside]{book}
+\\usepackage{amssymb}
+\\usepackage{amsmath}
+\\begin{document}
+';
+    $postamble = '\\end{document}
+';
+    // TODO incorporate project-specific preamble
+    file_put_contents($tex_base_name, $preamble . $page_text . $postamble);
 
     $result = [];
     $retval = -1;
-    $cmd = "pdflatex -no-shell-escape -halt-on-error -output-directory=$temp_path $tex_file_path";
+    $cmd = "pdflatex -no-shell-escape -halt-on-error $tex_base_name";
     exec($cmd, $result, $retval);
 
     // delete files no longer needed
-    unlink($tex_file_path);
-    unlink($tmp_file_path . '.log');
-    unlink($tmp_file_path . '.aux');
+    unlink($tex_base_name);
+    unlink($base_name . '.log');
+    unlink($base_name . '.aux');
     // pdf file will get deleted by cron
 
     // $retval means
     // 0: no serious problems: pdf file generated
     // 1: pdflatex problem: $result will contains necy. info.
-    // The log file is much longer and not needed
+    // (The log file is much longer and not needed)
     // >1: problem with exec, e.g. couldn't find pdflatex
 
     return [
--- a/api/v1_projects.inc
+++ b/api/v1_projects.inc
@@ -316,3 +316,11 @@ function api_v1_projects_pagerounds($method, $data, $query_params)
 
     return array_merge(["OCR"], array_keys($Round_for_round_id_));
 }
+
+function api_v1_project_pdflatex($method, $data, $query_params)
+{
+    $page_text = array_get($query_params, "text", "no text");
+    return [
+        "pdf" => $page_text,
+    ];
+}
--- a/api/v1_projects.inc
+++ b/api/v1_projects.inc
@@ -328,14 +328,35 @@ function api_v1_project_pdflatex($method, $data, $query_params)
         mkdir($temp_path);
     }
     $page_text = array_get($query_params, "text", "no text");
-    // temporary files will get deleted by cron
+    // TODO incorporate preamble and postamble
+    // TODO incorporate project-specific preamble
+
     $tmp_file_path = tempnam($temp_path, "");
-    rename($tmp_file_path, $tmp_file_path .= '.txt');
-//    $tmp_file_path .= ".txt";
-    file_put_contents($tmp_file_path, $page_text);
-    $file_base_name = basename($tmp_file_path);
+    $pdf_base_name = basename($tmp_file_path) . ".pdf";
+    $tex_file_path = $tmp_file_path . '.tex';
+    rename($tmp_file_path, $tex_file_path);
+    file_put_contents($tex_file_path, $page_text);
+
+    $result = [];
+    $retval = -1;
+    $cmd = "pdflatex -no-shell-escape -halt-on-error -output-directory=$temp_path $tex_file_path";
+    exec($cmd, $result, $retval);
+
+    // delete files no longer needed
+    unlink($tex_file_path);
+    unlink($tmp_file_path . '.log');
+    unlink($tmp_file_path . '.aux');
+    // pdf file will get deleted by cron
+
+    // $retval means
+    // 0: no serious problems: pdf file generated
+    // 1: pdflatex problem: $result will contains necy. info.
+    // The log file is much longer and not needed
+    // >1: problem with exec, e.g. couldn't find pdflatex
 
     return [
-        "url" => "$projects_url/$temp_dir_name/$file_base_name",
+        "retval" => $retval,
+        "result" => $result,
+        "url" => "$projects_url/$temp_dir_name/$pdf_base_name",
     ];
 }
--- a/api/v1_projects.inc
+++ b/api/v1_projects.inc
@@ -319,8 +319,23 @@ function api_v1_projects_pagerounds($method, $data, $query_params)
 
 function api_v1_project_pdflatex($method, $data, $query_params)
 {
+    global $projects_dir, $projects_url;
+
+    $temp_dir_name = "temp";
+
+    $temp_path = "$projects_dir/$temp_dir_name";
+    if(!is_dir($temp_path)) {
+        mkdir($temp_path);
+    }
     $page_text = array_get($query_params, "text", "no text");
+    // temporary files will get deleted by cron
+    $tmp_file_path = tempnam($temp_path, "");
+    rename($tmp_file_path, $tmp_file_path .= '.txt');
+//    $tmp_file_path .= ".txt";
+    file_put_contents($tmp_file_path, $page_text);
+    $file_base_name = basename($tmp_file_path);
+
     return [
-        "pdf" => $page_text,
+        "url" => "$projects_url/$temp_dir_name/$file_base_name",
     ];
 }
--- a/api/index.php
+++ b/api/index.php
@@ -30,7 +30,7 @@ function api()
 
     api_rate_limit($username);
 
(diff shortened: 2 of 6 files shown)


cpeel:
I'm still not sold that it is ever safe to compile arbitrary user-provided LaTeX code. https://tex.stackexchange.com/questions/528262/are-there-any-dangerous-commands-in-latex has some great reading.

For instance, LaTeX can include commands to write out files to the filesystem using the [`/write`](https://latexref.xyz/_005cwrite.html) command. In theory the file will be restricted to only being written in the top-level directory of the `.tex` file, but that in itself is problematic as it could write a URL-accessible virus that we are then hosting.

Please very very carefully review and test possible security issues.

srjfoo:
> Please very very carefully review and test possible security issues.

I will be happy to test for anything I know how to test for, but I need instruction in how to test for possible security issues. Probably because I don't understand what and where the possible security issues arise, I defer to @cpeel on when/if they are adequately covered.

For now, I've done a bit of testing of the preview function by copying page text from an [existing LaTeX project](https://www.pgdp.net/c/project.php?id=projectID401129f53953c) and pasting it into a page an [F2 project on TEST](https://www.pgdp.org/~rp31/c.branch/latex_preview/project.php?id=projectID45447792b4a5a) that I artificially labeled with the `{LaTeX}` tag in the project title.

A few comments/questions:

* What is the output meant for? I.e. is the intent that it will be printed out on 8.5 x 11 (or A-sized) sheets of paper? Is it intended to approximate the page size of the book? (I find the huge white margins to be distracting, but it would be worse to keep the same sized "page" and run the text from within a couple of ems of the edges.)

    I've looked at some of the older PDFs at PG, and at least early on, page size wasn't treated consistently ... see [PG #7825](https://www.gutenberg.org/files/7825/7825-pdf.pdf) and [PG #31076](https://www.gutenberg.org/files/31076/31076-pdf.pdf), one of which was and one of which was not DP, both of which take the minimal margin approach and [PG #36941](https://www.gutenberg.org/files/36941/36941-pdf.pdf), which is also DP, but takes a totally different layout approach. I would think that for previewing, cutting the page size and the margins would be a good idea.

* An example equation displays nicely as:
image

And is copyable as text, but if I copy and paste, I get:
```
w = log A
B + ∑
m
pmlog (z −am) −∑
n
qnlog (z −bn).
```
which isn't terribly useful. Is it more useful than an image in an HTML file?

* On several pages, I get no preview output, but a page with a bunch of tex markup, terminated by
```
!  ==> Fatal error occurred, no output PDF file produced!
Transcript written on ??????.log.
```
where the question marks appear to be a randomly generated alphanumeric string.

70ray:
>     * What is the output meant for? I.e. is the intent that it will be printed out on 8.5 x 11 (or A-sized) sheets of paper? Is it intended to approximate the page size of the book? (I find the huge white margins to be distracting, but it would be worse to keep the same sized "page" and run the text from within a couple of ems of the edges.)
>       I've looked at some of the older PDFs at PG, and at least early on, page size wasn't treated consistently ... see [PG #7825](https://www.gutenberg.org/files/7825/7825-pdf.pdf) and [PG #31076](https://www.gutenberg.org/files/31076/31076-pdf.pdf), one of which was and one of which was not DP, both of which take the minimal margin approach and [PG #36941](https://www.gutenberg.org/files/36941/36941-pdf.pdf), which is also DP, but takes a totally different layout approach. I would think that for previewing, cutting the page size and the margins would be a good idea.

Good point. Perhaps we should consult some more experienced Latex users about this.
> 
>     * An example equation displays nicely as:
> 
> 
> image
> 
> And is copyable as text, but if I copy and paste, I get:
> 
> ```
> w = log A
> B + ∑
> m
> pmlog (z −am) −∑
> n
> qnlog (z −bn).
> ```
> 
> which isn't terribly useful. Is it more useful than an image in an HTML file?

Probably not.
> 
>     * On several pages, I get no preview output, but a page with a bunch of tex markup, terminated by
> 
> 
> ```
> !  ==> Fatal error occurred, no output PDF file produced!
> Transcript written on ??????.log.
> ```
> 
> where the question marks appear to be a randomly generated alphanumeric string.

The randomly generated alphanumeric string is the base name of the tex file. The way it works is that the text is uploaded to the server and put in a file which is given a new unique name. The latex compiler uses this same name (with different extensions) for the output pdf and log files. I haven't used the log file so far because it doesn't seem to be any more helpful for finding problems than the output generated from the comand which is shown in the window when an error occurs.

Most Latex projects need to have several additional packages included in the preamble. If you look at the "tex" file of a book these are the lines starting with "\usepackage". I've only included two packages in the predefined preamble (amsmath and amssymb). This is why some pages get compiled correctly and others do not.
Additional packages needed would be specified in the project-specific preamble which would be made by the PM and could be updated when necessary during the formatting rounds.
Often the output generated includes a more helpful error message with the line number where it occured but sometimes it seems the compiler gets so confused that the output is not very helpful.

70ray:
> I'm still not sold that it is ever safe to compile arbitrary user-provided LaTeX code. https://tex.stackexchange.com/questions/528262/are-there-any-dangerous-commands-in-latex has some great reading.
> 
> For instance, LaTeX can include commands to write out files to the filesystem using the [`/write`](https://latexref.xyz/_005cwrite.html) command. In theory the file will be restricted to only being written in the top-level directory of the `.tex` file, but that in itself is problematic as it could write a URL-accessible virus that we are then hosting.
> 
> Please very very carefully review and test possible security issues.



> I'm still not sold that it is ever safe to compile arbitrary user-provided LaTeX code. https://tex.stackexchange.com/questions/528262/are-there-any-dangerous-commands-in-latex has some great reading.
> 
> For instance, LaTeX can include commands to write out files to the filesystem using the [`/write`](https://latexref.xyz/_005cwrite.html) command. In theory the file will be restricted to only being written in the top-level directory of the `.tex` file, but that in itself is problematic as it could write a URL-accessible virus that we are then hosting.
> 
> Please very very carefully review and test possible security issues.

We could run a virus check on the file with the uploaded text or the pdf file.

chrismiceli:
> 
> 
> > I'm still not sold that it is ever safe to compile arbitrary user-provided LaTeX code. https://tex.stackexchange.com/questions/528262/are-there-any-dangerous-commands-in-latex has some great reading.
> > For instance, LaTeX can include commands to write out files to the filesystem using the [`/write`](https://latexref.xyz/_005cwrite.html) command. In theory the file will be restricted to only being written in the top-level directory of the `.tex` file, but that in itself is problematic as it could write a URL-accessible virus that we are then hosting.
> > Please very very carefully review and test possible security issues.
> 
> > I'm still not sold that it is ever safe to compile arbitrary user-provided LaTeX code. https://tex.stackexchange.com/questions/528262/are-there-any-dangerous-commands-in-latex has some great reading.
> > For instance, LaTeX can include commands to write out files to the filesystem using the [`/write`](https://latexref.xyz/_005cwrite.html) command. In theory the file will be restricted to only being written in the top-level directory of the `.tex` file, but that in itself is problematic as it could write a URL-accessible virus that we are then hosting.
> > Please very very carefully review and test possible security issues.
> 
> We could run a virus check on the file with the uploaded text or the pdf file.

I think we'd have to worry about other issues too like denial of service type issues with infinite loops. We could run it in some apparmor or se linux sandbox with an extremely limited profile and set some ulimit cpu/memory resources perhaps, but I'm not an expert in that area.

70ray:
> 
> For instance, LaTeX can include commands to write out files to the filesystem using the [`/write`](https://latexref.xyz/_005cwrite.html) command. In theory the file will be restricted to only being written in the top-level directory of the `.tex` file, but that in itself is problematic as it could write a URL-accessible virus that we are then hosting.
> 
> Please very very carefully review and test possible security issues.

It is possible to download the (encoded) pdf file in the ajax response to avoid leaving any files on the server and not have them URL-accessible. I'm working on this.

lhamilton1:
Thanks for the information and reviews. 

I do not like the idea of having active code that can be input in our proofreading screens. A virus-checker would not catch problems and, even if we wrote a "catcher" routine to prevent known-to-be-dangerous codes, we might not catch everything. 

Also, operationally, this adds real complexity to our formatting environment. Foofers already have to remember a ton of rules etc. -- adding this would be asking a lot. 

I'm happy to have user-installed external tools that our PPers can use to convert the proofread/formatted text to LaTeX but not to have the LaTeX script in our proofreading environments.

70ray:
> I think we'd have to worry about other issues too like denial of service type issues with infinite loops.

This is a similar situation to the virus check operation which can take a long time. Presumably the pre-emptive multitasking environment prevents it affecting other users. It would be unlikely anyone would write an infinite loop in Latex accidentaly.

70ray:
The latest commit does not leave any files on the server and should overcome all security objections.

api/v1_projects.inc
```diff
@@ -316,3 +316,60 @@ function api_v1_projects_pagerounds($method, $data, $query_params)
...
+
+    $result = [];
+    $retval = -1;
+    $cmd = "pdflatex -no-shell-escape -halt-on-error $tex_base_name";
+    exec($cmd, $result, $retval);
+
+    $enc_pdf = (0 == $retval) ? base64_encode(file_get_contents($pdf_base_name)) : "";
+
+    // delete files no longer needed
+    $files = glob("*");
```
chrismiceli, line 357: We can't enumerate the outputted files ahead to avoid deleting anything by accident?

chrismiceli:
> 
> 
> The latest commit does not leave any files on the server and should overcome all security objections.

I'm still concerned about the latex script itself causing high memory usage, infinite loops, or accessing filesystem or network capabilities.

70ray:
> > The latest commit does not leave any files on the server and should overcome all security objections.
> 
> I'm still concerned about the latex script itself causing high memory usage, infinite loops, or accessing filesystem or network capabilities.

Perhaps we could devise some tests to check this out.

api/v1_projects.inc
70ray: Yes, we could do that. There will always be basename.tex, and, if the compile succeeded, .pdf and .aux. If there was anything else then removing the directory would fail but it would get cleaned up by the system (in 10 days?) since it is in the tmp directory.

chrismiceli:
> 
> 
> > > The latest commit does not leave any files on the server and should overcome all security objections.
> > 
> > 
> > I'm still concerned about the latex script itself causing high memory usage, infinite loops, or accessing filesystem or network capabilities.
> 
> Perhaps we could devise some tests to check this out.

Can you try:
https://tex.stackexchange.com/questions/148414/infinite-loop-with-algorithm2e
or
https://tex.stackexchange.com/questions/336425/infinite-loops-in-tex

and we could try some of the previous examples from Casey.

70ray:
> Can you try:  https://tex.stackexchange.com/questions/336425/infinite-loops-in-tex
> 

I tried this on my machine. It does illustrate a problem. I stopped the run after a few seconds but it had already written a pdf file of 42MB. If we run it non-manually we would have to have a way to stop it filling up the filing system.

srjfoo:
@cpeel and @lhamilton1 both brought up the potential security issues, but thus far one thing that nobody has answered to my satisfaction is the fact that the LaTeX code that is added to a page is stored in the database -- so if malicious code is added (intentionally or not), even if the result is not stored on the server, the code, itself, is stored in the database.

> It is proposed that a "preamble" text field be added to the project to contain project-specific stuff but this is not yet implemented. It could be edited in the same way as the other project details.

I have some concerns about this, also, as it assumes that a PM running a LaTeX project is LaTeX-literate. (Also, to the the `projects` table, or to any projectid table that is created as or modified to become a LaTeX project. I am assuming `projects` table as the fields in the projectid tables apply to all pages in the project, which would mean a copy of the preamble for each page.)

70ray:
> Thanks for the information and reviews. ...
> 

If we were to develop a proofreading application rather than have everything browser-based it would be possible to have a better editor with desirable features such as improved search/replace functionality and the ability to show line endings. It would then also be possible to integrate a preview for Latex markup using a Latex compiler on the user's own machine.
This may be some way off.

70ray:
> > Can you try:  https://tex.stackexchange.com/questions/336425/infinite-loops-in-tex
> 
> I tried this on my machine. It does illustrate a problem. I stopped the run after a few seconds but it had already written a pdf file of 42MB. If we run it non-manually we would have to have a way to stop it filling up the filing system.

It seems it is possible to have a size limited directory by creating a virtual filesystyem [https://www.linuxquestions.org/questions/linux-server-73/directory-quota-601140/](https://www.linuxquestions.org/questions/linux-server-73/directory-quota-601140/).
We would only need one such which could contain a few normal-sized page-pdf files.

PR #696, opened 2021-10-06
refinements based on feedback

chrismiceli:
fix empty series issue
adjust stroke width
formatting

https://www.pgdp.org/~chrismiceli/c.branch/refinements/

### DIFF
--- a/scripts/svgGraphs.js
+++ b/scripts/svgGraphs.js
@@ -3,7 +3,7 @@
 
 const {barLineGraph, stackedAreaGraph, pieGraph} = (function () {
     const margin = ({
-        top: 20,
+        top: 30,
         right: 45,
         bottom: 30,
         left: 45
@@ -12,7 +12,7 @@ const {barLineGraph, stackedAreaGraph, pieGraph} = (function () {
     function addTitle(svg, config, width) {
         svg.append("text")
             .attr("x", (config.width || width) / 2)
-            .attr("y", (margin.top / 2) + 3)
+            .attr("y", "13")
             .attr("font-size", "16px")
             .attr("text-anchor", "middle")
             .attr("fill", "currentColor")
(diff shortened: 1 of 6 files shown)


scripts/svgGraphs.js
```diff
@@ -154,140 +154,142 @@ const {barLineGraph, stackedAreaGraph, pieGraph} = (function () {
         const data = Object.entries(config.data).filter(([,{x}]) => x && x.length > 0);
         const svg = d3.select("#" + id).append("svg")
             .attr("viewBox", [0, 0, width, height]);
-        const tooltip = d3.select("#" + id)
```
chrismiceli, line 157: this diff is silly, I just added an if around a huge chunk of code to avoid error when there is no data. Probably on a dev issue, but just to avoid errors on first or month or some edge case.

scripts/svgGraphs.js
```diff
@@ -154,140 +154,142 @@ const {barLineGraph, stackedAreaGraph, pieGraph} = (function () {
...
-            } else if (seriesToRender === "bar" && (seriesData.type === undefined || seriesData.type === "bar")) {
+        if (data.length > 0) {
+            const tooltip = d3.select("#" + id)
+                .append("div")
+                .attr("class", "graph-tooltip")
+                .style("display", "none");
+
+            const mouseAction = (event, {value}) => {
+                tooltip.style("display", "")
+                    .text(d3.format('d')(value))
```
chrismiceli, line 165: I also changed format of tooltip to avoid .666666666666666666 causing tooltip to go off screen some or get clipped in iframe.

scripts/svgGraphs.js
```diff
@@ -154,140 +154,142 @@ const {barLineGraph, stackedAreaGraph, pieGraph} = (function () {
...
+                if (seriesToRender === "line" && seriesData.type === "line") {
+                    const line = d3.line()
+                        .x((d, i) => x(i))
+                        .y(({value}) => value < 0 ? y(0) : y(value));
+
+                    svg.append("path")
+                        .datum(data)
+                        .attr("fill", "none")
+                        .attr("class", `graph-series-stroke-${seriesIndex + 1}`)
+                        .attr("stroke-width", 2.5)
```
chrismiceli, line 252: this increases stroke width from 1.5 to 2.5 for line graphs.

scripts/svgGraphs.js
cpeel: <nod> Viewing the diff while ignoring whitespace is really useful for this MR!

PR #704, opened 2021-10-24
put commas in formatted numbers

chrismiceli:
put comma in y axis on stacked area graph
put comma in tooltip
https://www.pgdp.org/~chrismiceli/c.branch/groupedNumbers/stats/stats_central.php

Didn't noticed this in sandbox since less instances where commas would've shown up.


chrismiceli:
Thanks @vermiceli for this suggestion.

PR #707, opened 2021-11-02
stacked area y axis width adjustment

chrismiceli:
Fix yAxis for area charts as seen on user logon stats page
https://www.pgdp.org/~chrismiceli/c.branch/areaYAxisWidth/stats/user_logon_stats.php

also add 2 px between yAxis and label for bar line charts to give a little more room per request form @srjfoo

### DIFF
--- a/scripts/svgGraphs.js
+++ b/scripts/svgGraphs.js
@@ -94,6 +94,9 @@ const {barLineGraph, stackedAreaGraph, pieGraph} = (function () {
         data.y = "date";
         const series = d3.stack().keys(data.columns.slice(1))(data);
 
+        const svg = d3.select("#" + id).append("svg")
+            .attr("viewBox", [0, 0, width, height]);
+
         const y = d3.scaleLinear()
             .domain([0, d3.max(series, d => d3.max(d, d => d[1]))])
             .nice()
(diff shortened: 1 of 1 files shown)


cpeel (approved):

srjfoo:
This is irrelevant to the purpose of the current MR, but I realized when looking at things that the link to that graph is under "Miscellaneous statistics". Since that is a page of graphs, not statistics,
1. Should the page title and link be changed to "User Logon Graphs", and
2. Should the link to it be moved to "Miscellaneous Graphs", since we now have such a section? 

(That can either be done in this MR or held for a different one -- just tossing the idea out.)

srjfoo (approved):

chrismiceli:
> 
> 
> This is irrelevant to the purpose of the current MR, but I realized when looking at things that the link to that graph is under "Miscellaneous statistics". Since that is a page of graphs, not statistics,
> 
>     1. Should the page title and link be changed to "User Logon Graphs", and
> 
>     2. Should the link to it be moved to "Miscellaneous Graphs", since we now have such a section?
> 
> 
> (That can either be done in this MR or held for a different one -- just tossing the idea out.)

I'm not opposed to that in another MR as long as @cpeel is ok with it.

cpeel:
> I'm not opposed to that in another MR as long as @cpeel is ok with it.

In a separate MR would be preferable, please.

PR #710, opened 2021-11-10
Remove French projects from newcomer reserve

srjfoo:
At the request of the GM and the French language coordinator, French projects have been removed from the newcomer reserve.

Testable: [remove-french-reserve](https://www.pgdp.org/~srjfoo/c.branch/remove-french-reserve), compare against [test-vanilla](https://www.pgdp.org/~srjfoo/c.branch/test-vanilla) (which still has the newcomer reserve for French). The sandboxes have had the `page_tally_threshold` set to 5 pages so that the change can be verified without having to proof 500 pages.

### DIFF
--- a/pinc/stages.inc
+++ b/pinc/stages.inc
@@ -356,8 +356,8 @@ function can_user_get_pages_in_project($username, $project, $round)
         }
     }
 
-    // In P1, English and French projects, for a while after their release,
-    // are reserved for inexperienced users.
+    // Allow projects to be reserved for newcomers based on
+    // criteria defined in get_reserve_length().
     $n_days_of_reserve = get_reserve_length($project, $round);
     if ($n_days_of_reserve > 0) {
         $page_tally_threshold = 500;
@@ -446,7 +446,12 @@ function get_reserve_length($project, $round)
         return 0;
     }
 
-    if ($project->language == 'English' || $project->language == 'French') {
+    // If it is deemed appropriate to add a reserve for languages
+    // other than English, this is the block of code to customize.
+    // If the reserve will be the same as the English reserve, just
+    // add "|| $project->language == '<language>'" to the end of the
+    // condition. Otherwise, a most complex block will be needed.
+    if ($project->language == 'English') {
         // All 'special day' projects are exempted.
         if ($project->special_code != '') {
             return 0;
@@ -457,7 +462,7 @@ function get_reserve_length($project, $round)
     }
 
     // Anything not covered above
-    // (i.e., language other than 'English' or 'French')
+    // (i.e., language other than 'English')
     // is not reserved.
     return 0;
 }


chrismiceli:
I don't know how this works, but is this specific to pgdp and if so should it be in the prod branch?

srjfoo:
> I don't know how this works, but is this specific to pgdp and if so should it be in the prod branch?

Possibly it should be in the `pgdp-production` branch, but when I researched that for #700, I found it in `master`, not as a hack, so I kept it there. It's true that specific newcomer reserves will probably be different per site using the code, but the whole `get_reserve_length` function is there just for that purpose. So, I think the question is, where to draw the line. I'm open to other ideas.

chrismiceli (approved):
I'll defer to @cpeel to which branch, but since it is already in master I'm not opposed here.

pinc/stages.inc
```diff
@@ -356,8 +356,10 @@ function can_user_get_pages_in_project($username, $project, $round)
         }
     }
 
-    // In P1, English and French projects, for a while after their release,
-    // are reserved for inexperienced users.
+    // In P1, English projects, for a while after their release,
```
cpeel, line 359: I would make this comment generic and not include any of the possible reasons and defer that entirely to `get_reserve_length()` so we don't need to keep them in sync.

cpeel (approved):
I think this should go in `master` for now since it's in there already.

What should probably happen in clean-up after is that we:
1. change the `P1` references in `get_reserve_length()` to be the ELR round ID so that isn't hard-coded.
2. Move the non-ELR criteria to this function within the `pgdp-production` branch.

None of that is urgent and if we don't get around to it, that's fine too.

pinc/stages.inc
srjfoo: Good idea, and done.

srjfoo:
\<snip\>
> None of that is urgent and if we don't get around to it, that's fine too.

This is a good example of massive potential for scope creep. To do it right, wouldn't we want to fix it so that it would be easy to provide distinct reserves for multiple languages? Which also brought to mind that we have (I think) at least two tasks requesting that proofers be allowed to filter out projects that they can't proof in (which would need to include the "For Mentors" projects in P2).

cpeel:
> > None of that is urgent and if we don't get around to it, that's fine too.
> 
> This is a good example of massive potential for scope creep. To do it right, wouldn't we want to fix it so that it would be easy to provide distinct reserves for multiple languages? Which also brought to mind that we have (I think) at least two tasks requesting that proofers be allowed to filter out projects that they can't proof in (which would need to include the "For Mentors" projects in P2).

I think you're over thinking it. I'm not suggesting we provide an abstraction layer or any new functionality, more just moving the language-specific contents of the function out of `master` and into `pgdp-production`.

srjfoo:
> > > None of that is urgent and if we don't get around to it, that's fine too.
> > 
> > 
> > This is a good example of massive potential for scope creep. To do it right, wouldn't we want to fix it so that it would be easy to provide distinct reserves for multiple languages? Which also brought to mind that we have (I think) at least two tasks requesting that proofers be allowed to filter out projects that they can't proof in (which would need to include the "For Mentors" projects in P2).
> 
> I think you're over thinking it. I'm not suggesting we provide an abstraction layer or any new functionality, more just moving the language-specific contents of the function out of `master` and into `pgdp-production`.

Undoubtedly I'm overthinking it 🤣

PR #713, opened 2021-11-16
Multi note

70ray:
This enables multi-line proofers' notes and allows nested pairs of brackets inside proofers' notes. Tasks https://www.pgdp.net/c/tasks.php?action=show&task_id=2001 and https://www.pgdp.net/c/tasks.php?action=show&task_id=1977.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/multi_note

### DIFF
--- a/tools/proofers/preview.js
+++ b/tools/proofers/preview.js
@@ -1097,13 +1097,16 @@ $(function () {
         // merge issue arrays with notes so that if both start at same
         // index the issueStart appears after the note but the issueEnd
         // appears before the note.
-        // array will be reversed, high indexes will appear first
-        let allInserts = issueStarts.concat(noteArray).concat(issueEnds);
+        // there cannot be > one issue at the same index (if there were, one
+        // will have been discarded) but there can be more than one note, so
+        // ensure order of notes is unchanged
+        let allInserts = issueEnds.concat(noteArray).concat(issueStarts);
         allInserts.sort(function(a, b) {
             // if starts are same return 0, order unchanged
-            return b.start - a.start;
+            return a.start - b.start;
         });
-
+        // reverse the array so splice last elements first
+        allInserts.reverse();
         // splice issue marking and notes into text
         allInserts.forEach(function (insert) {
             tArray.splice(insert.start, 0, insert.text);
--- a/tools/proofers/preview.js
+++ b/tools/proofers/preview.js
@@ -846,7 +846,7 @@ $(function () {
         }
 
 
-        function addMarkUp(issArray, noteArray) {
+        function addMarkUp(text, issArray, noteArray) {
 
             function htmlEncode(s) {
                 return s.replace(/&/g, "&amp;")
@@ -854,14 +854,21 @@ $(function () {
                     .replace(/>/g, "&gt;");
             }
 
+            // temporarily encode < to placeholder █ so notes do not get styled
+            function modifiedEncode(s) {
+                return s.replace(/&/g, "&amp;")
+                    .replace(/</g, "█")
+                    .replace(/>/g, "&gt;");
+            }
+
             // split up the string into an array of characters
-            var tArray = txt.split("");
+            var tArray = text.split("");
             tArray = tArray.map(htmlEncode);
 //            tArray.forEach(htmlEncodeChar);
             let issueInserts = makeIssueInserts(issArray);
 
             noteArray.forEach(function(note) {
-                note.text = htmlEncode(note.text);
+                note.text = modifiedEncode(note.text);
             });
             // merge with notes so that if both start at same index the issue appears first
             // array will be reversed, high indexes will appear first
@@ -878,7 +885,12 @@ $(function () {
                 tArray.splice(insert.start, 0, insert.text);
             });
 
-            txt = tArray.join("");  // join it back into a string
+            return tArray.join("");  // join it back into a string
+        }
+
+        function decodeNoteTags() {
+            txt = txt.replace(/█/g, "&lt;");
+            // replace placeholder with &lt;
         }
 
         // add style and optional colouring for marked-up text
@@ -1094,9 +1106,13 @@ $(function () {
 
         let analysis = analyse(txt, styler);
         let issArray = analysis.issues;
-        txt = analysis.text;
-        addMarkUp(issArray, analysis.noteArray);
-
+//        txt = analysis.text;
+        if (wrapMode) {
+            // leave out notes
+            txt = addMarkUp(analysis.text, issArray, []);
+        } else {
+            txt = addMarkUp(analysis.text, issArray, analysis.noteArray);
+        }
         let issues = 0;
         let possIss = 0;
         issArray.forEach(function(issue) {
@@ -1111,6 +1127,7 @@ $(function () {
         let ok = (issues === 0);
         if (ok) {
             showStyle();
+            decodeNoteTags();
             if (wrapMode) {
                 reWrap();
             }
--- a/tools/proofers/preview.js
+++ b/tools/proofers/preview.js
@@ -183,11 +183,11 @@ $(function () {
             return result.index;
         }
 
-        function chkCharAfter(start, len, type, str1) {
+        function chkCharAfter(start, len, type, descriptor) {
             const ix = start + len;
             const end = findEnd(ix);
             if (/\S/.test(txt.slice(ix, end))) {
-                reportIssue(start, len, "charAfter", type, str1);
+                reportIssue(start, len, "charAfter", type, descriptor);
                 return true;
             }
             return false;
@@ -225,8 +225,8 @@ $(function () {
             }
 
             // check that no other characters are on the same line
-            function chkAlone(start, len, str1) {
-                if(chkCharAfter(start, len, 1, str1)) {
+            function chkAlone(start, len, descriptor) {
+                if(chkCharAfter(start, len, 1, descriptor)) {
                     return;
                 }
 
@@ -612,7 +612,7 @@ $(function () {
             }
 
             // check no chars follow on same line and next line is blank
-            function chkAfter(start, len, str1, type, checkBlank) {
+            function chkAfter(start, len, descriptor, type, checkBlank) {
                 // true if find */ or \n or eot
                 function endNWorBlank(pc) {
                     if (txt.slice(pc, pc + 2) === "*/") {
@@ -621,13 +621,13 @@ $(function () {
                     return !(/./).test(txt.charAt(pc));
                 }
 
-                if(chkCharAfter(start, len, type, str1)) {
+                if(chkCharAfter(start, len, type, descriptor)) {
                     return;
                 }
 
                 const end = findEnd(start + len);
                 if (checkBlank && !endNWorBlank(end + 1)) {
-                    reportIssue(start, len, "blankAfter", type, str1);
+                    reportIssue(start, len, "blankAfter", type, descriptor);
                 }
             }
 
--- a/tools/proofers/preview.js
+++ b/tools/proofers/preview.js
@@ -1,6 +1,21 @@
 /* exported makePreview analyse processExMath */
 /* global $ previewMessages XRegExp */
 
+// the formatting rules are applied as if proofers' notes were
+// invisible so remove them first and save for later. But first check if they
+// are malformed, if so mark the problem and do nothing else.
+
+// then analyse the text without notes and make an array of issues.
+// the issues and the notes are then merged back into the text.
+// The text could have <, > and & characters which should be html encoded,
+// do this before merging since the issue markup will contain < and >.
+// Thus the tags in the text will also be encoded.
+
+// The last stage is to interpret any tags to show the styling.
+// To avoid styling notes change these characters to something else before
+// merging and then convert back to html codes afterwards.
+
+
 var makePreview;
 var analyse;
 var getILTags;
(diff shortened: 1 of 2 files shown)


tools/proofers/preview.js
```diff
@@ -26,16 +26,26 @@ function processExMath(text, textFunction, allowMath) {
...
-    function removeComments(textLine) {
-        return textLine.replace(/\[\*\*[^\]]*\]/g, '');
-    }
-
-    // true if txtLine contains anything except only comments or spaces
-    function nonComment(textLine) {
-        return (/\S/.test(removeComments(textLine)));
+// find index of next unmatched ] return 0 if none found
+function findClose(txt, index) {
+    var result;
```
chrismiceli, line 31: Should we use let and const like our other js files?

tools/proofers/preview.js
```diff
@@ -26,16 +26,26 @@ function processExMath(text, textFunction, allowMath) {
...
+    var nestLevel = 0;
+    var re = /\[|\]/g;  // [ or ]
```
chrismiceli, line 33: Not sure if it matters, but I'd put the regex out of the function, so it won't be created every function invocation. Compiling regex can be expensive and I've seen this trick cause an app go from 8 min to instant.

tools/proofers/preview.js
```diff
@@ -107,6 +117,65 @@ $(function () {
...
+            // find start of note
+            while (null != (result = reNoteStart.exec(txt))) {
+                noteStartIndex = result.index;
+                noteEndIndex = findClose(txt, reNoteStart.lastIndex);
+                // copy text to start of note
+                txtOut += txt.slice(beginIndex, noteStartIndex);
+                outNoteStart += noteStartIndex - beginIndex;
+                // if note starts at beginning of a line and ends at the end of a line
+                // remove any following nl character also so doesn't count as a blank line and include the nl in the note
+                if(((noteStartIndex === 0) || (txt.charAt(noteStartIndex - 1) === "\n")) && ((noteEndIndex === maxIndex) || (txt.charAt(noteEndIndex + 1) === "\n"))) {
```
chrismiceli, line 160: maybe add some newlines to group the conditions more clearly:
```
if(((noteStartIndex === 0) || (txt.charAt(noteStartIndex - 1) === "\n")) &&
   ((noteEndIndex === maxIndex) || (txt.charAt(noteEndIndex + 1) === "\n"))) {
```

tools/proofers/preview.js
```diff
@@ -1125,6 +1065,57 @@ $(function () {
...
+        let issueStarts = [];
+        let issueEnds = [];
+        let errorString;
+        issArray.forEach(function(issue) {
+            if (issue.type === 0) {
+                errorString = makeErrStr("hlt");
+            } else {
+                errorString = makeErrStr("err");
+            }
+            let message = previewMessages[issue.code].replace("%s", issue.subText);
```
chrismiceli, line 1079: do we need to make sure issue.subText doesn't have any of the patterns used by replace (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_string_as_a_parameter). We ran into an issue like this in the find replace dialog, so you can see my solution there if necessary https://github.com/DistributedProofreaders/dproofreaders/blob/master/tools/proofers/srchrep.js#L27

srjfoo: May not be an adequate test, but I used 

```
[** [386]][** [** [386]]$1\n\n\\n\\n$$2][** A][** B]
```
as proofers' notes, and what was shown in Preview was

```
[** B][** A][** [** [386]]$1\n\n\n\n$$2][** [386]]
```

I see two issues immediately 
* while the `$$` was accurately displayed in the preview, the `\\` was not. A single backslash appears to be echoed just fine, but if it's being used preceding another backslash or a `$`, it's treated as an escape character.
* There are four consecutive notes there -- they're displayed in the preview in reverse order.

tools/proofers/preview.js
70ray: Yes.

tools/proofers/preview.js
70ray: Interesting. Done as you suggest.

tools/proofers/preview.js
70ray: good idea.

tools/proofers/preview.js
70ray: @chrismiceli The string is always predefined so it shouldn't be a problem. i.e. '/*', '/#', '*/', '#/', 'Footnote', 'Illustration', 'Sidenote', '&lt;tb&gt;'.

@srjfoo Interesting. I'll have to look into this.

70ray: @srjfoo I've revised the sorting code so the notes appear in the correct order.
It looks like the backslash issue is caused by MathJax. If I disable the Math preview the backslashes in your example appear correctly. Does it work like that for you?

srjfoo: Yep, the sorting looks good now.

And yes, turning off Math preview takes care of the backslash issue -- I didn't think of that. Is it fair to assume that that's something that will have to be lived with? (Bear in mind that I was using test strings that are unlikely to be used in reality to test per Chris' question).

Would it be appropriate to move the Math checkbox out of the "Configure" block and into the main Preview Toolbox? I realize we need to keep that small, but I had consciously decided "Well, there's no harm in leaving the Math checkbox checked permanently" and gone off and forgotten about it. If the checkbox is in the main toolbar, it's a reminder that "this is something you may want immediate on-off access to".

70ray: > Yep, the sorting looks good now.
> 
> And yes, turning off Math preview takes care of the backslash issue -- I didn't think of that. Is it fair to assume that that's something that will have to be lived with? (Bear in mind that I was using test strings that are unlikely to be used in reality to test per Chris' question).
> 
> Would it be appropriate to move the Math checkbox out of the "Configure" block and into the main Preview Toolbox? I realize we need to keep that small, but I had consciously decided "Well, there's no harm in leaving the Math checkbox checked permanently" and gone off and forgotten about it. If the checkbox is in the main toolbar, it's a reminder that "this is something you may want immediate on-off access to".

I don't see why MathJax should need to change the backslash characters in this way. I'll enquire why it happens.
In the meantime we can live with it as we have been without noticing the problem. Moving the checkbox is something we could consider.

srjfoo (approved):
This looks to me like it solves the problems mentioned in both tasks. I'm happy to put off dealing with the disappearing-double-backslashes-when-math-mode-is-turned-on, and considering moving the checkbox to enable math mode off to a future MR (or to just footnote it as a hopefully rare side-effect of enabling math mode, if that's the better option).

I would recommend giving the formatting community a week or two to review it before merging (that still gives us time to get it in for December's code deployment, and US Thanksgiving week is next week).

cpeel:
@70ray - is this ready for final testing and review?

@chrismiceli - can you please re-review when you have a moment?

srjfoo:
I'm not sure whether or not this issue was part of the original MR, because I didn't think to test `<tb>`s at that time.

In the sandbox, with color markup turned on:
image

and turned off:
image

On both PROD and the main TEST sandbox, the `<tb>` tag stays there either way.

70ray:
> @70ray - is this ready for final testing and review?
> 
No there is still an issue with the <tb> tag as @srjfoo found.

70ray:
It's still not quite right.

70ray:
This should be ready for final testing and review now.

srjfoo (approved):
> This should be ready for final testing and review now.

I've re-tested, I think, and also did some spot checking in BrowserStack. Looks good to me.

I've also requested additional testing in the [New version of Format Preview available for testing.](https://www.pgdp.net/phpBB3/viewtopic.php?t=75636) topic.

chrismiceli (approved):

tools/proofers/preview.js
```diff
@@ -26,16 +26,27 @@ function processExMath(text, textFunction, allowMath) {
...
-    function removeComments(textLine) {
-        return textLine.replace(/\[\*\*[^\]]*\]/g, '');
-    }
 
-    // true if txtLine contains anything except only comments or spaces
-    function nonComment(textLine) {
-        return (/\S/.test(removeComments(textLine)));
+// find index of next unmatched ] return 0 if none found
+const re = /\[|\]/g;  // [ or ]
+function findClose(txt, index) {
```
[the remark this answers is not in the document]
70ray: It is also tested in formatPreviewTests.js. Not sure if this is a good reason.

tools/proofers/preview.js
```diff
@@ -92,21 +103,75 @@ $(function () {
...
+            txtOut += txt.slice(beginIndex);
+
+            // remove trailing whitespace on each line and trailing blank lines
+            txtOut = txtOut.replace(/ *$/mg, "");
+            return txtOut.replace(/\s*$/, "");
+        }
+
         // find end of line (or eot) following ix
         function findEnd(ix) {
             var re = /\n|$/g;
```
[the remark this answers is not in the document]
70ray: Yes, perhaps later.

70ray:
Davide Cervone of MathJax has replied to my query about the backslashes: "In order to allow you to include a literal $, \(, \), \[ or \] in your text, MathJax processes double backslashes as single backslashes.  So you can use \\( to get a literal \( in your page.  There is a configuration parameter that controls whether that processing is done or not, so you can turn that off if you wish."

srjfoo:
> Davide Cervone of MathJax has replied to my query about the backslashes: "In order to allow you to include a literal $, (, ), [ or ] in your text, MathJax processes double backslashes as single backslashes. So you can use \( to get a literal ( in your page. There is a configuration parameter that controls whether that processing is done or not, so you can turn that off if you wish."

So, I guess the first question is, what do we want Preview to show? I guess at least for the "Tags" version, I expect everything to show in Preview. The thing is, "No Tags" really isn't fully "No Tags"; it's more properly "No inline tags", because all of the block level tags remain.

The second question that comes to mind is, can we proceed with this the way it is, with the caveat that backslashes have a special meaning in certain cases in MathJax, and that for now, if Math is turned on, backslashes used to escape special characters in the math markup will not be shown?

It's been a little over a week. I assume at least someone took the time to re-test, but we've heard 🦗 🦗 🦗 from the intended audience.

If there's some additional testing you'd like me to do, let me know.

70ray:
> > Davide Cervone of MathJax has replied to my query about the backslashes: "In order to allow you to include a literal $, (, ), [ or ] in your text, MathJax processes double backslashes as single backslashes. So you can use ( to get a literal ( in your page. There is a configuration parameter that controls whether that processing is done or not, so you can turn that off if you wish."
> 
> So, I guess the first question is, what do we want Preview to show? I guess at least for the "Tags" version, I expect everything to show in Preview. The thing is, "No Tags" really isn't fully "No Tags"; it's more properly "No inline tags", because all of the block level tags remain.
> 
Yes, "tags" means in-line tags in this context. `<tb>` although it looks like an in-line tag, is treated like an out-of-line tag for this purpose. We could make a new task to clarify this issue. 

> The second question that comes to mind is, can we proceed with this the way it is, with the caveat that backslashes have a special meaning in certain cases in MathJax, and that for now, if Math is turned on, backslashes used to escape special characters in the math markup will not be shown?
> 
I think that we can leave it like it is. I presume backslashes are unlikely to appear in any DP books. The single $ sign is by default not enabled as a delimiter for math so $ can appear without any problem. `\$` will also just produce a $ sign.

> It's been a little over a week. I assume at least someone took the time to re-test, but we've heard cricket cricket cricket from the intended audience.
> 
> If there's some additional testing you'd like me to do, let me know.

It is difficult to say. There always seems to be something that I hadn't thought of that catches us out. I came across one thing that could be considered a bug while testing but it is not new so not really in the scope of this.

srjfoo:
> > It's been a little over a week. I assume at least someone took the time to re-test, but we've heard cricket cricket cricket from the intended audience.
> > If there's some additional testing you'd like me to do, let me know.
> 
> It is difficult to say. There always seems to be something that I hadn't thought of that catches us out. I came across one thing that could be considered a bug while testing but it is not new so not really in the scope of this.

I think we want to keep the scope focused on the two tasks that you addressed in this MR, plus any associated problems we uncovered along the way that are deemed critical to their operation, so I'm in favor of proceeding, given that we've had no complaints to do with it in over a week.

My collection of tests is on [this page in your sandbox](https://www.pgdp.org/~rp31/c.branch/multi_note/tools/page_browser.php?project=projectID44dda6ce1d66d&imagefile=141.png&mode=imageText&round_id=F1), if you want to take a look at it and see if you think there's anything I haven't covered. It passes in your sandbox, but fails on the main test site (yes, it contains multiple tests, which, for troubleshooting errors is not good, but when the problems are fixed, it's convenient to have them all in one place).

70ray:
> 
> My collection of tests is on [this page in your sandbox](https://www.pgdp.org/~rp31/c.branch/multi_note/tools/page_browser.php?project=projectID44dda6ce1d66d&imagefile=141.png&mode=imageText&round_id=F1), if you want to take a look at it and see if you think there's anything I haven't covered. It passes in your sandbox, but fails on the main test site (yes, it contains multiple tests, which, for troubleshooting errors is not good, but when the problems are fixed, it's convenient to have them all in one place).

It looks like a good test to me.

chrismiceli (approved):

PR #714, opened 2021-11-19
#2004: DP logo text in dark mode is not visible

chrismiceli:
Add image for dark mode themes

https://www.pgdp.org/~chrismiceli/c.branch/issue_2004

Note, the slim seems to be only used in quiz pages.


chrismiceli:
The png was generated by taking the dp-logo.svg from https://www.pgdp.net/dp_branding/, changing the color to white, and then exporting as png to make the size small (~15kb instead of 112kb, which lines up with ~11kb for existing png).

quiz/small_theme.inc
```diff
@@ -3,6 +3,7 @@ include_once($relPath.'html_page_common.inc'); // output_html_header()
 include_once($relPath.'theme.inc'); // headerbar_text_array()
 include_once($relPath.'faq.inc');
 include_once($relPath.'quizzes.inc'); // get_quiz_id_param get_Quiz_with_id
+include_once($relPath.'Users.inc'); // get_quiz_id_param get_Quiz_with_id
```
srjfoo, line 6: Should be `User.inc`. (Took me a while -- it was staring me right in the face and I still didn'nt see it!)

quiz/small_theme.inc
```diff
@@ -11,10 +12,12 @@ function output_small_header($quiz)
     output_html_header('');
 
     $quiz_name = $quiz->short_name;
+    $user = User::load_current();
+    $logo = $user && $user->i_theme == "charcoal" ? "dp-logo-dark-360x68px.png" : "dp-logo-360x68px.png";
```
cpeel, line 16: I appreciate you working on this but I explicitly do not want to include or special-case things based on themes in the code. If we can figure out a way to do this purely in CSS I'm game, but we shouldn't do it in the code.

cpeel (changes requested):

quiz/small_theme.inc
srjfoo: @chrismiceli asked me to take a look at how the image looked, which I just kind of skimmed over when I first looked, but when I really started looking at it, the fuzziness of the white text really jumped out at me.

White on charcoal grey shows up flaws like that a lot more than black on light green/grey/blue does, though when I started looking at the PG theme, for contrast, I realized that even that looks fuzzy, especially when compared to the sharpness of the surrounding text.

All of this is leading up to -- I realize that the SVG file is significantly larger than the pngs, but can we get away with using an SVG logo?

cpeel: The SVG will be served up as gzip compressed so it'll be just 19k on the wire, and the browser will cache it, so the size shouldn't be an issue.

And just to be clear: I'm not suggesting that we must **theme** the SVG via CSS, but rather use CSS to serve up the appropriate file to begin with (as Chris mentioned in the task) rather than it being in PHP.

chrismiceli: using css to select the url is harder than I imagined since we don't know $code_url to generate an appropriate url

srjfoo:
It looks really lovely and incredibly sharp ... in Firefox on computers and at a minimum on my Pixel 3 on FF (and Chrome, but with a blue dropshadow). But the first place I saw it was on my iPad, and it doesn't look so good on the iPad (iPadOS 14.something). So that sent me off to BrowserStack, and the news isn't very good. Firefox is literally the only browser that I looked at that dependably, over the range of browsers we support, displayed it as it should be displayed.

Safari doesn't like it under any circumstances, as things are at the moment -- it shows the black or white tagline starting with Safari 12.1 on Mojave (the one before that, it doesn't even show the tagline), Chrome seems to want to show the dropshadow the same color as the logo.

Note that I did not do exhaustive testing, but just tested the top, middle and bottom, sort of. It's mostly enough to just show the login page unless you just want to see how it looks on Charcoal. And I even tried the most recent iPadOS on BrowserStack, with results that match my not-yet-upgraded-to-the-most-recent iPad.

I honestly don't know what the problem is -- we tested the graphs on BrowserStack, and they appear to be okay, so it's not that they won't display SVGs -- they just don't like this particular one.

quiz/small_theme.inc
cpeel: You should be able to specify a relative URL against the final place of the CSS -- you don't need to know the absolute URL.

chrismiceli:
> 

Ok, I'll work a different fix that doesn't rely on cloning svg and styling it with css. Thanks for testing.

quiz/small_theme.inc
chrismiceli: Yes, you're right @cpeel, I was confused because I got the relative path wrong and it wasn't working /facepalm.

chrismiceli:
> It looks really lovely and incredibly sharp ... in Firefox on computers and at a minimum on my Pixel 3 on FF (and Chrome, but with a blue dropshadow). But the first place I saw it was on my iPad, and it doesn't look so good on the iPad (iPadOS 14.something). So that sent me off to BrowserStack, and the news isn't very good. Firefox is literally the only browser that I looked at that dependably, over the range of browsers we support, displayed it as it should be displayed.
> 
> Safari doesn't like it under any circumstances, as things are at the moment -- it shows the black or white tagline starting with Safari 12.1 on Mojave (the one before that, it doesn't even show the tagline), Chrome seems to want to show the dropshadow the same color as the logo.
> 
> Note that I did not do exhaustive testing, but just tested the top, middle and bottom, sort of. It's mostly enough to just show the login page unless you just want to see how it looks on Charcoal. And I even tried the most recent iPadOS on BrowserStack, with results that match my not-yet-upgraded-to-the-most-recent iPad.
> 
> I honestly don't know what the problem is -- we tested the graphs on BrowserStack, and they appear to be okay, so it's not that they won't display SVGs -- they just don't like this particular one.

That svg syntax with inline reference seems sort of new, but still disappointing the differences between the browsers.

cpeel:
I *love* the direction this is going!!

It would probably be marginally better if we skipped the `` tag altogether and just went with 100% CSS:
```diff
diff --git a/pinc/theme.inc b/pinc/theme.inc
index 09d343e4f..10db44cb6 100644
--- a/pinc/theme.inc
+++ b/pinc/theme.inc
@@ -123,9 +123,10 @@ function html_logobar()
     global $maintenance, $testing, $alert_message;

     echo "\n<div id='header'>\n";
-    echo "<div id='logo'>\n";
-    echo "<a href='$code_url'><img width='360' height='68' alt='$site_name'></a>\n";
+    echo "<a href='$code_url'>\n";
+    echo "<div id='logo' role='img' alt='$site_name'>\n";
     echo "</div>\n";
+    echo "</a>\n";

     echo "<div id='titles-preserved'><p>\n";

diff --git a/styles/layout.less b/styles/layout.less
index 1d4d9dd1c..2665d03fe 100644
--- a/styles/layout.less
+++ b/styles/layout.less
@@ -208,12 +208,12 @@ header, #header {
     color: @header-text;
     #logo, #titles-preserved {
         display: table-cell;
-        img {
-          content: url('../../graphics/dp-logo-light.png');
-        }
     }
     #logo {
-        padding: 0.5em 0.5em 0.1em;
+        width: 360px;
+        height: 68px;
+        margin: 0.5em;
+        content: url('../../graphics/dp-logo-light.png');
     }
     #titles-preserved {
         .center-align;
```

I would also prefer if we could keep the original `dp-logo-360x68px.png` filename and make the new one `dp-logo-360x68px-dark.png` as that would align with the naming convention we use on the branding page.

Also: it would be really fantastic if we had a `dp-logo-500px-dark.png` to include on the branding page as well.

pinc/theme.inc
```diff
@@ -124,7 +124,7 @@ function html_logobar()
 
     echo "\n<div id='header'>\n";
     echo "<div id='logo'>\n";
-    echo "<a href='$code_url'><img src='$code_url/graphics/dp-logo-360x68px.png' width='360' height='68' alt='$site_name'></a>\n";
+    echo "<a href='$code_url' />\n";
```
cpeel, line 127: Done this way the link does not work for me in Chrome. In my proposal I had the `<a>` wrap the `<div>`.

styles/layout.less
```diff
@@ -211,6 +211,10 @@ header, #header {
     }
     #logo {
         padding: 0.5em 0.5em 0.1em;
```
cpeel, line 213: Now that we're setting the margin we need to remove the padding.

styles/themes/charcoal.css
```diff
@@ -1174,6 +1174,10 @@ header #titles-preserved,
 header #logo,
 #header #logo {
   padding: 0.5em 0.5em 0.1em;
+  width: 360px;
+  height: 68px;
+  margin: 0.5em;
+  content: url('../../graphics/dp-logo-550x96px-light.png');
```
cpeel, line 1180: Having the 500x96 for the branding page is great, but these files should remain 360x68 so we are not relying on the browser to resize them -- they'll look better across more browsers that way.

cpeel (changes requested):

styles/themes/charcoal.css
srjfoo: For what it's worth, reverting to the 360x68 version of the logo looks *_much_* worse on Charcoal. Perhaps because it's "white" on "black", the fuzziness shows much more clearly than it does on the light themes.

cpeel: To clarify, is it that the larger image downsized in the browser was clearer, or was it that the larger image at actual size was clearer? In a prior rev the image was notably bigger than the current 360x68 and I don't think we want it to take up more room in the header. If it's truly the former then I'm amenable to having the larger images.

cpeel (approved):
This works & looks great - thank you for working on it and your patience as we iterated over the details!

styles/themes/charcoal.css
srjfoo: Hm. I'm remembering that it was the larger image downsized in the browser was clearer. I had a browser window open with it using the 500px image inside the 360px container. Once this push came through, I refreshed the page. The image stayed the same size, but the quality degraded significantly. 

There won't be a difference on the light themes, but to me, the fuzziness is very noticeable on Charcoal.

srjfoo: Erh ... to clarify, there won't be a difference on the light themes between this and the current live version. There was still a definite improvement with the 500px image in the 360px container on the light theme as well.

chrismiceli:
I'm going to play with compression settings in Inkscape for the PNG from SVG export

styles/themes/charcoal.css
srjfoo: Ohhhh. Bitten by the difference-in-browsers issues once again. I finally thought to look in Chrome (I had the page open from before), and indeed, when I refreshed the page, the image got smaller. I was just looking at it in Firefox. <sigh> And yes, it also looks clearer in the larger version, but is definitely too large.

chrismiceli:
> I'm going to play with compression settings in Inkscape for the PNG from SVG export

Ok, so I tried playing with it, but 360px is not really enough pixels for that amount of text even with anti aliasing and no compression. Even the current one for light themes is pretty poor:
image
The exported dark version is about the same but contrast probably has something to do with visibility on the page:
image

I'm guessing the browser can do better subpixel rendering during sampling for the reduced size that can't be achieved on the png export when going from the 500px version, but images are not my expertise (what is?). So I think we can live with the current one, or use a larger one and let browsers shrink the image for us, but I'll defer to @cpeel .

srjfoo:
> So I think we can live with the current one, or use a larger one and let browsers shrink the image for us, but I'll defer to @cpeel .

I did see in Chrome what @cpeel was ralking about. I.e. Chrome not fitting the image into the container.

If it's not possible to force the browsers to uniformly fit the larger image into the smaller container, the only other thing I can think of is to possibly not use stark white for the text, but a very light grey, to reduce the contrast somewhat.

chrismiceli:
> 

Hmm, that was with the larger image? Perhaps that was to my fault during export of png or something. We could revisit that and see what is causing the issue if we agree fundamentally that 500px being shrunk by browser is acceptable.

srjfoo:
> Hmm, that was with the larger image? Perhaps that was to my fault during export of png or something. We could revisit that and see what is causing the issue if we agree fundamentally that 500px being shrunk by browser is acceptable.

I will defer to @cpeel, but if it can be made to work across browsers, I wouldn't think there would be a problem. The slightly larger file size shouldn't be a problem, should it?

cpeel:
> I'm guessing the browser can do better subpixel rendering during sampling for the reduced size that can't be achieved on the png export when going from the 500px version, but images are not my expertise (what is?). So I think we can live with the current one, or use a larger one and let browsers shrink the image for us, but I'll defer to @cpeel .

I'm amenable to having the higher resolution version scaled down in the browser for better quality if we can make it "look right" in all browsers. 👍

chrismiceli:
> > I'm guessing the browser can do better subpixel rendering during sampling for the reduced size that can't be achieved on the png export when going from the 500px version, but images are not my expertise (what is?). So I think we can live with the current one, or use a larger one and let browsers shrink the image for us, but I'll defer to @cpeel .
> 
> I'm amenable to having the higher resolution version scaled down in the browser for better quality if we can make it "look right" in all browsers. 👍

Ok, updated with larger image for browsers to scale down. It looks ok to me, let me know if I'm missing the issue you encountered last time.

srjfoo (approved):

srjfoo:
> > > I'm guessing the browser can do better subpixel rendering during sampling for the reduced size that can't be achieved on the png export when going from the 500px version, but images are not my expertise (what is?). So I think we can live with the current one, or use a larger one and let browsers shrink the image for us, but I'll defer to @cpeel .
> > 
> > 
> > I'm amenable to having the higher resolution version scaled down in the browser for better quality if we can make it "look right" in all browsers. 👍
> 
> Ok, updated with larger image for browsers to scale down. It looks ok to me, let me know if I'm missing the issue you encountered last time.

Here's what I see in Chrome:
image

The top one is with the update, the bottom one is from last night, with me not having refreshed the page. Seems significantly clearer to me, and it looks like it takes up the same space to me. I've tested only FF, Chrome and Safari on macOS Big Sur -- do I need to do some BrowserStack testing?

cpeel:
In Chrome on my (ancient) 11" MBA the new one looks decidedly blurry:
Screen Shot 2021-11-28 at 2 58 56 PM

compared to the old one:
Screen Shot 2021-11-28 at 2 58 59 PM

cpeel:
The tagline has always been a problem -- the image is just too small for the tagline to be rendered clearly.

cpeel:
What's fascinating to me is that it's just Chrome on the 11" MBA. The images are visually identical in Firefox.

cpeel:
I've taken Chris' branch and made the following changes:
* Created a `dp-logo-dark.svg` which only changes the tagline text from black to white
* Replaced the PNGs with SVGs
* Makes the entire logo every so slightly bigger (68px to 75 px high) while not making the header bar taller.

Code is in https://www.pgdp.org/~cpeel/c.branch/issue_2004/

I've tested that it looks good on:
* 15" MBP in Chrome, Firefox, Safari
* 11" MBA in Chrome, Firefox, Safari
* Pixel 4a in Chrome, Firefox

Indeed, on the MBP the SVG looks much better than the original PNG while on the MBA they look almost identical to the original.

The kicker is going to be testing this with BrowserStack against the oldest supported browsers to make sure they render the SVG correctly.

srjfoo:
> The kicker is going to be testing this with BrowserStack against the oldest supported browsers to make sure they render the SVG correctly.

I can certainly do some of that testing this evening.

srjfoo:
> The kicker is going to be testing this with BrowserStack against the oldest supported browsers to make sure they render the SVG correctly.

I tested with BrowserStack against Win7: Edge (oldest version), FF, Chrome and Opera; macOS High Sierra: Safari, FF, Chrome and Safari.

All of the browsers on both OS platforms did fine _**except**_ Firefox. I had to go all the way up to FF 63 to be able to see the logos at al. However, I don't think that it's the image that's the problem, though they fail in different ways. Neither the PNG nor the SVG was visible until FF 63, and then they both were, and matched the other browsers. My guess is that there's something that the earlier versions of FF didn't like about how the container was coded.

Screenshots, typical of FF 52 (oldest supported version) - 62:
[https://www.pgdp.org/~cpeel/c.branch/issue_2004/](https://www.pgdp.org/~cpeel/c.branch/issue_2004/): Logo container space was not reserved at all.

image

[https://www.pgdp.org/~chrismiceli/c.branch/issue_2004](https://www.pgdp.org/~chrismiceli/c.branch/issue_2004): Logo container space was reserved, and link worked.

image

cpeel:
> All of the browsers on both OS platforms did fine except Firefox. I had to go all the way up to FF 63 to be able to see the logos at al.

That's weird and unfortunate. FF 63 was released just back in 2018, so that's pretty recent. The feature we are using `content: url()` reportedly works all the way back to FF 2: https://caniuse.com/mdn-css_properties_content_url

 Thank you for the testing. I'll poke a bit at it on BrowserStack and see if I can't figure out what part it doesn't like.

srjfoo:
> I'll poke a bit at it on BrowserStack and see if I can't figure out what part it doesn't like.

Thanks! It's entirely possible that I was doing something wrong, but if I did, someone else is likely to, so it's best to figure out what.

cpeel:
I suspect [this is the feature](https://caniuse.com/mdn-css_properties_content_element_replacement) we are actually using and it isn't supported in FF until 63. Let me see if I can finagle this into a slightly different mechanism.

cpeel:
I hacked together a method using `::before` with `content: url()` which works in FF62 but images inserted in that format can't be resized and come in at their natural size. That's not great but I resized the SVGs to match but this breaks our mobile support. I'll muddle over this some more.

cpeel:
Ok, I've tried something different and went back to using an `` tag and using `content: url()` replacement only for the charcoal theme. What this means is -- assuming it works in all the other browser/versions -- is that the logo will show correctly for the non-dark theme in all our supported browsers. The dark theme logo will work in all supported browsers that aren't FF62 and earlier.

The sandbox has been updated and, assuming it initially looks good, will need another round of testing everywhere.

srjfoo:
> Ok, I've tried something different and went back to using an `` tag and using `content: url()` replacement only for the charcoal theme. What this means is -- assuming it works in all the other browser/versions -- is that the logo will show correctly for the non-dark theme in all our supported browsers. The dark theme logo will work in all supported browsers that aren't FF62 and earlier.
> 
> The sandbox has been updated and, assuming it initially looks good, will need another round of testing everywhere.

Here's hoping. I'm not too worried about charcoal. We'll just have to add that caveat to the announcement. I'll get back to the testing tomorrow.

srjfoo:
Many apologies for the delay, but I wanted to take a little time between testing sessions.

[https://www.pgdp.org/~cpeel/c.branch/issue_2004/](https://www.pgdp.org/~cpeel/c.branch/issue_2004/) -- testing on BrowserStack ...

* iOS on an iPhone 6S; FF considers TEST not a trusted site.
* iOS on an iPhone 7 -- only Safari and Chrome as options, and it's FF giving the trouble on the logo, but I went ahead and double-checked Safari anyway, and it looks fine.
* 2016 iPad Pro 9.7 running iOS11, Chrome looks fine. None of the iPads has FF as an option.
* Pixel 3 running FF (don't know what version) looks fine.
* Nexus 7 (the oldest google phone they have) logo looks fine.
* Win7, Win10, FF52, logo with black text.
* Win7, Win10, FF63, logo with white text.
* Win7, Win10, FF94, logo with white text.
* macOS El Capitan, FF52, logo with black text.
* macOS El Capitan, FF63, logo with white text.
* macOS El Capitan, FF78, logo with white text. (latest available FF on El Capitan)
* macOS Monterey, FF52, logo with black text.
* macOS Monterey, FF63, Something seriously wrong with FF 60s -- TEST doesn't render at all in BrowserStack (not just this sandbox, but `pgdp.org`). I didn't test the full range, but just a couple. On the other hand, who would choose to use FF63 on Monterey **_except_** for testing?
* macOS Monterey, FF94, logo with white text.

And of course on FF94 on Big Sur (real instead of BrowserStack emulation), the logo with the white tagline looks awesome.

I can do additional testing in other browsers later, but opted (except on the phones and iPad) to focus just on FF. Early indicators from the phones, though, encourage me to be hopeful.

cpeel:
That's all a good sign!! Unless @chrismiceli objects I'll get my changes into this branch in this PR so y'all can review my changes to his code.

chrismiceli:
> That's all a good sign!! Unless @chrismiceli objects I'll get my changes into this branch in this PR so y'all can review my changes to his code.

please do @cpeel! Thanks for helping this come to a solution.

cpeel:
MR is updated with what is in my `issue/2004` sandbox.

cpeel:
Something to note is that the forum styles use a copy of (not a reference) to the `graphics/dp-logo-360x68px.png` file we are removing. So it will continue to work just fine, but it will be visually different from the new, clearer version. We will probably want to coordinate with @wf49670 on updating the forum version to use the SVG when this is rolled out.

srjfoo:
> MR is updated with what is in my `issue/2004` sandbox.

My followup testing (black/white = black tagline for light themes, white tagline for charcoal) based on your `issue/2004` sandbox:

* macOS Monterey and High Sierra (emulated)
    * Safari -- black/white
    * Chrome -- 96: black/white; 54: black/white
    * Opera -- 81: black/white; 41: black/white

* macOS High Sierra (emulated), I also tried Edge 96 and 80: black/white

* macOS Big Sur (actual, most recent) -- Safari, Chrome, Opera: black/white. I was trying these on my current machines, not BrowserStack, so did not test the low-end.

* Win7, win10, win11
    * Edge 96: black/white; 80: black/white
    * Chrome 96: black/white; 54: black/white
    * Opera 81: black/white; 41: black/white

* win10, only, I also tried Edge 15 and 18 (win10 is the only version I saw versions of Edge earlier than 80 on). They look the same as FF < 63, i.e. black/black.

srjfoo (approved):

PR #715, opened 2021-11-26
Import dp_branding into the pgdp_production branch

srjfoo:
The `dp_branding` directory has lived outside the `/c` directory structure. This MR pulls it into the pgdp-production branch, so that it is available on both PROD and TEST and can be updated in both places, when necessary, as part of the code deployment process, and not depend on a human to remember to update it in both places when a change is made. First commit just pulls the directory in; second commit pulls the CSS out of the HEREDOC blocks in the index page and into `layout.less` and `charcoal.less`. Also added require_login().

Testable: [dp_branding](https://www.pgdp.org/~srjfoo/c.branch/import-dp_branding/dp_branding/)


srjfoo:
Further comments: For the background color behind the transparent color logos and marks, I went with the neutral `@table-cell-base-highc` in `layout.less`. This didn't work on the black and white logo on Charcoal, so I hacked just that one place in `charcoal.less`. So, while on the lighter themes, the backgrounds are the same behind all of the transparent. The lighter grey used on the light themes was too light for Charcoal, and the almost-black of the high-contrast cell background color on Charcoal was too dark for the black logo to be visible.

I toyed around with the idea of using the more mid-range grey for all of the logos/marks with transparent backgrounds on the charcoal theme, but it made more sense to me to make it relevant to the viewer's theme. Which brought me to wonder if it makes more sense for all of the images with transparent backgrounds should have the logobar background color instead of a neutral grey. (It wouldn't make a difference for the b&w logo -- that still needs the medium grey on charcoal to be visible.) So, that's the first question I'd like opinions on.

The second question -- In the interest of saving time, everything that was in the HEREDOC blocks except the CSS was left alone. All user-visible text outside the HEREDOC blocks is marked up for localization. Should I go ahead and take the time now to pull the remaining code out of the HEREDOC blocks so that it can be marked up? I don't feel like it's terribly critical to get it done quickly, but will defer to wiser heads. 😁 

Probably a good idea to test both light and dark themes, and to compare against [this version](https://www.pgdp.org/dp_branding/). Oh yeah, I removed the statsbar, too.

dp_branding/index.php
```diff
@@ -0,0 +1,158 @@
...
+$relPath = '../pinc/';
+include_once($relPath.'base.inc');
+include_once($relPath.'theme.inc');
+
+/*
+Author: cpeel
+Created: 2016-05-03
+*/
+
+require_login();
```
cpeel, line 11: This page needs to remain accessible to non-logged-in-users who want to use branding assets to link to our site.

cpeel (commented):
This looks great to me.

My biggest worry is that having changes in the generated `*.css` files is going to cause no end of headache with merge conflicts when we need to update this branch. I don't see any effective way around that, however, since the whole point is to make it work with the dark theme. We may be able to ameliorate that by putting the changes at the very bottom of the `*.less` files which would be my recommendation to start. If we find that resolving the conflicts becomes painful we can revisit this later.

70ray (approved):
Looks good to me.

dp_branding/index.php
srjfoo: Fixed.

srjfoo:
> This looks great to me.
> 
> My biggest worry is that having changes in the generated `*.css` files is going to cause no end of headache with merge conflicts when we need to update this branch. I don't see any effective way around that, however, since the whole point is to make it work with the dark theme. We may be able to ameliorate that by putting the changes at the very bottom of the `*.less` files which would be my recommendation to start. If we find that resolving the conflicts becomes painful we can revisit this later.

I've moved it to the end of the `*.less` files. I can squash the last two commits if you want me to, but left them separate for now.

cpeel:
This looks good. Please squash the last two commits.

srjfoo:
> This looks good. Please squash the last two commits.

Done.

lhamilton1:
Looks good to me too

PR #716, opened 2021-12-09
Add SVG logo with white tagline for dark themes

srjfoo:
Added SVG logo with a white tagline for dark themes. Added it to the logos on the `dp_branding/index.php` page, updated the page to display the SVG files, and verified that everything is visible whether viewed in a light theme or a dark theme.

Sandbox at [dp_branding/](https://www.pgdp.org/~srjfoo/c.branch/update-branding/dp_branding/)

Related: #715, merged into `pgdp-production`, but not yet deployed to PROD.


dp_branding/index.php
```diff
@@ -36,121 +36,202 @@
...
 
 echo "<h3 id='mark'>" . _("Mark") . "</h3>";
 
-echo "<p><b>" . _("SVG file") . ":</b> <a href='dp-mark.svg'>dp-mark.svg</a></p>";
-
-echo "<p><b>" . _("PNG files") . ":</b></p>";
+echo "<h4 id='mark'>" . _("SVG file") . "</h4>";
+
+echo <<<SVGMARK
+        <table class='branding'>
```
cpeel, line 146: Thought: should we make this SVG table like the Logo tables above where the left column has the "Transparent background" label and the right is the mark with the link?

srjfoo: Good idea. I've got a meeting coming up shortly, but I'll do that afterwards.

dp_branding/index.php
```diff
@@ -36,121 +36,202 @@
...
+        <tr>
+            <td></td>
+            <td>
+                <a href='dp-logo.svg'>dp-logo.svg</a>
+            </td>
+        </tr>
+        <tr>
+            <td>Transparent<br>Background, dark themes</td>
+        <td style='background-color:black;'>
+                <img src='dp-logo-dark.svg'>
```
chrismiceli, line 58: instead of linking to the branding/dp-logo-dark.svg, does it make sense to link to the graphics one and only have 1 copy of the file until we need to diff for the pgdp production branch?

chrismiceli: Same for dp-logo.svg now since it is in master brnach.

srjfoo: Since the logo and mark are specific to DP, and we actually would (I assume) prefer those using the code to have their own logo and mark, I would have thought that it should be the other way around, and have the logo and mark only in the `pgdp-production` branch, but that is beyond the scope of this MR and would involve changes in both `master` and `pgdp-production`. However, I'll defer to @cpeel on what the best way to do it is.

dp_branding/index.php
srjfoo: Done.

dp_branding/index.php
srjfoo: Some diffs can be ignored by setting the diff view to ignore white space.

cpeel: In this case I would leave the duplicate files. My rationale is that the `dp-logo.svg` file in PROD has existed at https://www.pgdp.net/dp_branding/dp-logo.svg since its inception and we might inadvertently break something external to the site were we to remove it. I also see value in keeping the branding page separate from the location of files in the main codebase since they could change independently (it would be easy to change code in `master` without realizing it would impact `pgdp-production`.

cpeel:
The page looks good to me.

cpeel (approved):

lhamilton1:
I fully agree!

chrismiceli (commented):
I don't think we're super consistent localizing the strings on this page, but it is an existing issue I believe so I'm not worried about it.

srjfoo:
> 

I'm willing to pull the tables out of the HEREDOC blocks, but as far as I know, that's the only way to make the user-facing text fully localizable.

cpeel:
Making the strings translatable should be a separate PR if we opt to do it. There wasn't a lot of value in doing that before it was pulled into part of the code because the strings wouldn't have been pulled out to have someone translate them.

PR #717, opened 2021-12-12
Use load_special_days() in more places

cpeel:
Use the `load_special_days()` code when populating the page to view them and when loading them when editing projects. This also adds a helper function for sorting them by the most commonly used methods. The only user-visible change is that special days that happen on the same month/day pair are now sub-sorted alphabetically by the special day display name. This is probably only relevant on TEST I think.

I did not change the queries used in `tools/site_admin/manage_special_days.php` since that is doing updates as well as queries. I think that's the last remaining place that queries the `special_days` table directly though.

This is some infra prep work as I putz on [Task 2011](https://www.pgdp.net/c/tasks.php?action=show&task_id=2011).

Testable in [centralize-special-day-query](https://www.pgdp.org/~cpeel/c.branch/centralize-special-day-query/) sandbox.

### DIFF
--- a/pinc/ProjectSearchForm.inc
+++ b/pinc/ProjectSearchForm.inc
@@ -189,20 +189,12 @@ class ProjectSearchForm
 
     public static function special_day_options()
     {
-        $special_day_options = [];
-        $sql = "
-            SELECT
-                spec_code,
-                display_name,
-                DATE_FORMAT(concat('2000-',open_month,'-',open_day),'%e %b')
-            FROM special_days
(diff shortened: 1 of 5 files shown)


chrismiceli (approved):

srjfoo (approved):
> The only user-visible change is that special days that happen on the same month/day pair are now sub-sorted alphabetically by the special day display name. This is probably only relevant on TEST I think.

👍 -- it's a nice change!

70ray (approved):

PR #719, opened 2021-12-18
Add API to get list of charsuites, special days, and image sources

cpeel:
This PR extends the API to allow users to get a list of character suites, special days, and image sources. This will be necessary for users to create projects (eventually) as part of [Task 2011](https://www.pgdp.net/c/tasks.php?action=show&task_id=2011).

The API intentionally does not use the exact same field names as those stored in the database. Instead, we try to present a more unified view to the outside world. Like other parts of the API, these only return data that the user can already see in the UI. This is most relevant to image sources which have scoped visibility -- some can only be viewed by PMs, or Image Source Managers / SAs. The PR includes update to the Swagger schema as well.

This code is live in the sandbox at https://www.pgdp.org/~cpeel/c.branch/api-proj-gets/ but there are no UI changes. Instead, testing can be done via `curl`:

```bash
# export the user's API key
# this one is for an unprivileged proofreader on TEST:
export API_KEY=review

# charsuites
curl -i -g -X GET "https://www.pgdp.org/~cpeel/c.branch/api-proj-gets/api/index.php?url=v1/projects/charsuites" \
    -H "accept: application/json" \
    -H "X-API-KEY: $API_KEY"

# special days
curl -i -g -X GET "https://www.pgdp.org/~cpeel/c.branch/api-proj-gets/api/index.php?url=v1/projects/specialdays" \
    -H "accept: application/json" \
    -H "X-API-KEY: $API_KEY"

# image sources
curl -i -g -X GET "https://www.pgdp.org/~cpeel/c.branch/api-proj-gets/api/index.php?url=v1/projects/imagesources" \
    -H "accept: application/json" \
    -H "X-API-KEY: $API_KEY"
```

All of these support an `enabled` parameter that, if present, will only return the entities that are enabled.
```bash
# only enabled image sources
curl -i -g -X GET "https://www.pgdp.org/~cpeel/c.branch/api-proj-gets/api/index.php?url=v1/projects/imagesources&enabled" \
    -H "accept: application/json" \
    -H "X-API-KEY: $API_KEY"
```

### DIFF
--- a/api/dp-openapi.yaml
+++ b/api/dp-openapi.yaml
@@ -173,6 +173,36 @@ paths:
         default:
           $ref: '#/components/responses/UnexpectedError'
 
+  /projects/imagesources:
+    get:
+      tags:
+      - project
+      description: Returns a list of valid project image sources
+      parameters:
+      - name: enabled
+        in: query
+        description: Only return enabled image sources
+        required: false
+        schema:
+          type: boolean
+      responses:
+        200:
+          description: List of valid image sources
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/image_source'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '429':
+          $ref: '#/components/responses/RateLimitExceeded'
+        default:
+          $ref: '#/components/responses/UnexpectedError'
+
   /projects/languages:
     get:
       tags:
@@ -678,6 +708,25 @@ components:
           type: array
           items:
             type: string
+        enabled:
+          type: boolean
+
+    image_source:
+      required:
+      - id
+      type: object
+      properties:
+        id:
+          type: string
+        name:
+          type: string
+        name_full:
+          type: string
+        url:
+          type: string
+          format: url
+        credit:
+          type: string
         enabled:
           type: boolean
 
--- a/api/dp-openapi.yaml
+++ b/api/dp-openapi.yaml
@@ -219,6 +219,36 @@ paths:
         default:
           $ref: '#/components/responses/UnexpectedError'
 
+  /projects/specialdays:
+    get:
+      tags:
+      - project
+      description: Returns a list of valid special days
+      parameters:
+      - name: enabled
+        in: query
+        description: Only return enabled special days
+        required: false
+        schema:
+          type: boolean
+      responses:
+        200:
+          description: List of special days
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/special_day'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '429':
+          $ref: '#/components/responses/RateLimitExceeded'
+        default:
+          $ref: '#/components/responses/UnexpectedError'
+
   /projects/states:
     get:
       tags:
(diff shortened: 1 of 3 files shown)


api/v1.inc
```diff
@@ -25,6 +25,9 @@ $router->add_route("GET", "v1/projects/genres", "api_v1_projects_genres");
 $router->add_route("GET", "v1/projects/languages", "api_v1_projects_languages");
 $router->add_route("GET", "v1/projects/states", "api_v1_projects_states");
 $router->add_route("GET", "v1/projects/pagerounds", "api_v1_projects_pagerounds");
+$router->add_route("GET", "v1/projects/charsuites", "api_v1_projects_charsuites");
```
chrismiceli, line 28: I know our project ids cannot resemble words like "states", or "charsuites", but for routing purposes, does order matter? That way an ambiguous route like "v1/projects/states" that can match both won't ever hit the "v1/projects/:projectid" route with :projectid set to "states"? I guess if it is working, it doesn't matter.

api/v1_projects.inc
```diff
@@ -316,3 +317,86 @@ function api_v1_projects_pagerounds($method, $data, $query_params)
...
+            continue;
+        }
+        $return_data[] = [
+            "id" => $special_day["spec_code"],
+            "name" => $special_day["display_name"],
+            "comment" => $special_day["comment"],
+            "url" => $special_day["info_url"],
+            "color" => $special_day["color"],
+            "symbol" => $special_day["symbol"],
+            "date_open" => sprintf("%02d-%02d", $special_day["open_month"], $special_day["open_day"]),
```
chrismiceli, line 364: do we ever care about time zones or are these days always UTC and no concern with time zones in the "future"

api/dp-openapi.yaml
```diff
@@ -189,6 +249,36 @@ paths:
...
+      parameters:
+      - name: enabled
+        in: query
+        description: Only return enabled special days
+        required: false
+        schema:
+          type: boolean
+      responses:
+        200:
+          description: List of valid character suites
```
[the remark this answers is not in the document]
cpeel: Fixed it - thanks for catching that copy/pasta error.

chrismiceli (approved):

api/v1.inc
cpeel: The ApiRouter evaluates static routes over parameter-based routes so the order isn't important. Although to be fair I had to look back at the code to confirm that for certain.

api/v1_projects.inc
cpeel: DP time is all "server time", so it isn't UTC and we also don't need to worry about timezones. It's an interesting artifact from the initial developers.

srjfoo: See the comment [here](https://github.com/DistributedProofreaders/dproofreaders/blob/d01225fed2fddcad30ab069a46d0a9d355293175/crontab/onoff_special_event_queues.php#L24). That applies only to defined special days, though, not birthdays and otherdays.

Today's been busier than I expected -- I haven't done any reviewing or testing, yet, but wanted to go ahead and point to the time zone info. I don't know if that's relevant to the API, though.

api/dp-openapi.yaml
```diff
@@ -626,6 +751,33 @@ components:
...
+        comment:
+          type: string
+        url:
+          type: string
+          format: url
+        color:
+          type: string
+        symbol:
+          type: string
+        date_open:
```
70ray, line 772: Why not keep month and date as separate fields? then the user could more easily format the date however they wanted to.

cpeel: Yeah, that's probably better. How does `date_open_month` and `date_open_day` sound as field names?

70ray: Sounds alright to me.

srjfoo: Works for me, too.

api/v1_projects.inc
```diff
@@ -316,3 +317,86 @@ function api_v1_projects_pagerounds($method, $data, $query_params)
...
+            continue;
+        }
+
+        $return_data[] = [
+            "id" => $image_source["code_name"],
+            "name" => $image_source["display_name"],
+            "name_full" => $image_source["full_name"],
+            "url" => $image_source["url"],
+            "credit" => $image_source["credit"],
+            "enabled" => $image_source["is_active"] == 1,
```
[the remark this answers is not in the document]
cpeel: I don't think there is, no. An image source is only usable / enabled if it's not pending and disabled which is what this is representing. The primary use of this API, at least initially, is going to be getting a list of special days to use within a project create / update call and those can only be used if they are enabled.

srjfoo (approved):

api/dp-openapi.yaml
cpeel: Updated.

srjfoo (approved):

PR #720, opened 2021-12-18
Centralize project difficulty definition

cpeel:
Define the project difficulty options once instead of *four* times. This is in prep work for project validation as part of Task 2011.

Testable in the [centralize-difficulty-list](https://www.pgdp.org/~cpeel/c.branch/centralize-difficulty-list/) sandbox.

### DIFF
--- a/pinc/filter_project_list.inc
+++ b/pinc/filter_project_list.inc
@@ -1,6 +1,7 @@
 <?php
 include_once($relPath.'misc.inc'); // attr_safe()
 include_once($relPath.'iso_lang_list.inc');
+include_once($relPath.'Project.inc'); // get_project_difficulties()
 
 // This file includes functions to handle the creation, update, and display of
 // the project filters used on the Round, Project, and Smooth Reading pages.
@@ -157,9 +158,9 @@ class DifficultyElement extends ProjectFilterElement
             echo " checked";
         }
         echo ">\n";
-        $difficulty_options = ['beginner' => _("Beginners Only"), 'easy' => _("Easy"), 'average' => _("Average"), 'hard' => _("Hard")];
-        foreach ($difficulty_options as $value => $option) {
-            echo "<label for='$value'>&nbsp;$option</label><input type='checkbox' name='difficulty[]' class='diff-opt' id='$value' value='$value'";
+        foreach (get_project_difficulties() as $value => $option) {
+            $escaped_option = html_safe($option);
+            echo "&nbsp;<label for='$value'>$escaped_option</label><input type='checkbox' name='difficulty[]' class='diff-opt' id='$value' value='$value'";
             if (is_array($difficulty) && in_array($value, $difficulty)) {
                 echo " checked";
                 $this->selected_options[] = $option;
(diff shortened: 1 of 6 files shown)


project.php
```diff
@@ -407,20 +407,6 @@ function do_project_info_table()
     // -------------------------------------------------------------------------
     // Information about the work itself (independent of DP)
 
-    // the array below should guarantee that the strings 'beginner',
```
cpeel, line 410: Actually, I need to update this to use the new function to get the translated values. Will do that soon and update this PR.

cpeel: Fixed.

pinc/filter_project_list.inc
```diff
@@ -157,8 +158,7 @@ class DifficultyElement extends ProjectFilterElement
             echo " checked";
         }
         echo ">\n";
-        $difficulty_options = ['beginner' => _("Beginners Only"), 'easy' => _("Easy"), 'average' => _("Average"), 'hard' => _("Hard")];
-        foreach ($difficulty_options as $value => $option) {
+        foreach (get_project_difficulties() as $value => $option) {
             echo "<label for='$value'>&nbsp;$option</label><input type='checkbox' name='difficulty[]' class='diff-opt' id='$value' value='$value'";
```
chrismiceli, line 162: ideally we'd encode in case french or something introduced some ' characters.

pinc/filter_project_list.inc
```diff
@@ -157,8 +158,7 @@ class DifficultyElement extends ProjectFilterElement
             echo " checked";
         }
         echo ">\n";
-        $difficulty_options = ['beginner' => _("Beginners Only"), 'easy' => _("Easy"), 'average' => _("Average"), 'hard' => _("Hard")];
```
chrismiceli, line 160: Do we care that "Beginners Only" is different than "Beginner"?

chrismiceli (approved):

pinc/filter_project_list.inc
cpeel: Good catch! I've escaped `$option`.

pinc/filter_project_list.inc
cpeel: I think consistency throughout the site is better, but am interested in @srjfoo's thoughts.

srjfoo: I'd go for consistency throughout the site. We don't really need to care that "Beginners Only" is different from "Beginner".

Though it's actually an interesting question. I think the only round where anything except the filter block adds "Only" is P1, where the difficulty level is prepended to the genre in the project listing. In P2, it's "MENTORS ONLY" and in P3-F2, there isn't a prefix at all. An argument could be made that the filter having "Only" in P2-F2 is misleading, and doesn't actually need to be there in P1. However that's not particularly relevant here. I still think that for now, it's probably best to be consistent.

srjfoo: > However that's not particularly relevant here.

🤦‍♀️ -- okay. I hadn't looked at the testing yet. Yes, I totally agree (following my reasoning above) that it's appropriate to take the "Only" out of the filter block.

srjfoo (approved):

70ray (approved):

PR #721, opened 2021-12-19
Bring project_jump.php from noncvs into the code

cpeel:
This brings the `noncvs/project_jump.php` script into the codebase. I modeled the code off of the other SA scripts -- and like those it isn't up to the same high standards as the rest of the codebase (translatable, etc). It is now generic and applies to whatever rounds are defined in the code. We'll need to augment this in `pgdp-production` to add the PP-specific stages we're interested in.

This script also now checks that the project's pages table exists before allowing the transition.

Testable in the [project-jump](https://www.pgdp.org/~cpeel/c.branch/project-jump/tools/site_admin/index.php) sandbox.

### DIFF
--- a/tools/site_admin/project_jump.php
+++ b/tools/site_admin/project_jump.php
@@ -1,0 +1,137 @@
+<?php
+$relPath = '../../pinc/';
+include_once($relPath.'base.inc');
+include_once($relPath.'misc.inc');
+include_once($relPath.'user_is.inc');
+include_once($relPath.'theme.inc');
+include_once($relPath.'Project.inc');
+include_once($relPath.'project_events.inc');
+
+require_login();
+
+if (!user_is_a_sitemanager()) {
+    die("You are not authorized to invoke this script.");
+}
+
+$action = get_enumerated_param($_POST, 'action', 'showform', ['showform', 'check', 'dojump']);
+$projectid = get_projectID_param($_POST, 'projectid', true);
+$valid_new_states = [];
+foreach ($Round_for_round_id_ as $id => $round) {
+    $valid_new_states[] = $round->project_unavailable_state;
+}
+$new_state = get_enumerated_param($_POST, 'new_state', null, $valid_new_states, true);
+
+
+$title = _("Project Jump");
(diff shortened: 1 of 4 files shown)


srjfoo (approved):
Looks good to me.

tools/site_admin/project_jump.php
```diff
@@ -0,0 +1,137 @@
...
+    echo "</form>";
+}
+
+function do_stuff($projectid, $new_state, $just_checking)
+{
+    global $pguser;
+    $project = new Project($projectid);
+    echo "<pre>";
+    echo "    projectid : $projectid\n";
+    echo "    title     : $project->nameofwork\n";
```
chrismiceli, line 96: should we encode the name of work in case there is a </pre> in it? I know, I know...

chrismiceli (approved):

tools/site_admin/project_jump.php
cpeel: We should, but all of the `site_admin/` scripts need TLC to not be using `<pre>` tags at all. So in this case I'd rather have the consistency. The SAs are use to them at this point :/

srjfoo: Speaking as an SA, I assume that the reason the `<pre>` wrappers were used in the first place was to make the text stand out, and nobody thought things through enough to use CSS when setting the scripts up. Most of them pre-date the update of the site-wide CSS.

I was assuming that in this case, @chrismiceli's comment was aimed at avoiding a situation like this (I added a `</pre>` to the project title):
image

where the PM included a `</pre>` in the nameofwork field. Which kind of breaks the display of the title. The title is actually "Book clone for deletion".

chrismiceli: That was my point, but I think I agree with @cpeel and we can do this all at once in another MR if we want since project creation is limited to ideally non malicious actors who won't embed XSS type attacks.

srjfoo: 👍  -- except for those of us doing the testing and going out of our way to break things. 😁

PR #722, opened 2021-12-20
2004: DP logo text in dark mode is not visible

chrismiceli:
fix logo in quiz layout

https://www.pgdp.org/~chrismiceli/c.branch/issue_2004/quiz/generic/main.php?quiz_page_id=p_basic_1

I think the better fix would probably be to adjust the output_small_header function to wrap the content in a page-container div similar to output_header does but it is riskier than this change I think.

I verified firefox only downloads the image once.

### DIFF
--- a/quiz/small_theme.inc
+++ b/quiz/small_theme.inc
@@ -16,6 +16,7 @@ function output_small_header($quiz)
     echo "<a href='$code_url'>\n";
     echo "<div id='logo'>\n";
     echo "<img src='$code_url/graphics/dp-logo.svg' alt='$site_name'>\n";
+    echo "</div>\n";
     echo "</a>";
     echo "</div>\n";
 
--- a/styles/layout.less
+++ b/styles/layout.less
@@ -212,7 +212,6 @@ header, #header {
     #logo {
         height: 75px;
         margin: 0.25em 0.5em;
-        content: url('../../graphics/dp-logo.svg');
         img {
             height: 100%;
             margin: 0.25em 0.5em;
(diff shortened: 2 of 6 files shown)


cpeel:
I think this is actually malformed HTML and is missing a closing div:
```diff
diff --git a/quiz/small_theme.inc b/quiz/small_theme.inc
index e86478213..89d3a8c0f 100644
--- a/quiz/small_theme.inc
+++ b/quiz/small_theme.inc
@@ -18,6 +18,7 @@ function output_small_header($quiz)
     echo "<img src='$code_url/graphics/dp-logo.svg' alt='$site_name'>\n";
     echo "</a>";
     echo "</div>\n";
+    echo "</div>\n";

     // XXX localization bug:
     // $guide_url should point to the locale-appropriate version of the guidelines.
```
Looks like my changes to your SVG branch took it out 😦

chrismiceli:
> I think this is actually malformed HTML and is missing a closing div:
> 
> ```diff
> diff --git a/quiz/small_theme.inc b/quiz/small_theme.inc
> index e86478213..89d3a8c0f 100644
> --- a/quiz/small_theme.inc
> +++ b/quiz/small_theme.inc
> @@ -18,6 +18,7 @@ function output_small_header($quiz)
>      echo "$site_name\n";
>      echo "</a>";
>      echo "</div>\n";
> +    echo "</div>\n";
> 
>      // XXX localization bug:
>      // $guide_url should point to the locale-appropriate version of the guidelines.
> ```
> 
> Looks like my changes to your SVG branch took it out 😦

I noticed the layout was wrong, but didn't check to see if you introduced it so I fixed it the same way charcoal worked. I didn't think to check if the layout error was recently introduced :). Thanks for pointing it out. This is a better fix.

PR #724, opened 2021-12-22
Add PP states to project_jump.php

cpeel:
_This is going into the `pgdp-production` branch._

This updates the new `project_jump.php` script in the code to include PP-specific states that were not easily abstracted out into a generic form for released code. This should now match what is currently live in `noncvs`.

Testable in the [project-jump](https://www.pgdp.org/~cpeel/c.branch/project-jump/) sandbox.


chrismiceli (approved):

srjfoo (approved):
Looks good to me.

I've pointed the db-req squirrels at the sandbox, in case they want to take it for a test-drive.

PR #725, opened 2021-12-27
Do not encode tags before show style

70ray:
but encode as part of showStyle function.
If any serious issues showStyle is not called so
encode tags instead.
This avoids the encoded ampersand causing all-cap text
to be interpreted as mixed case. Addresses https://www.pgdp.net/c/tasks.php?action=show&task_id=2014

Sandbox at: https://www.pgdp.org/~rp31/c.branch/small_cap_ampersand


srjfoo (approved):
I can't comment on the code, as I don't know enough javascript to spit at, but the fix works as advertised when compared against the current live version.

chrismiceli (approved):

PR #726, opened 2021-12-30
Speed up select WordCheck functions

cpeel:
After some profiling using xdebug on "Manage proofreaders' Suggestions" in response to [Task 2010](https://www.pgdp.net/c/tasks.php?action=show&task_id=2010) I made some surgical fixes to two WordCheck functions that improved the processing time by 25% according to the profile. I don't know how much this will improve okrick's performance problems on PROD, as I think the overriding factors for him are the total number of projects (70!) & text therein.

Testing should focus on ensuring that the results shown on the "Manage proofreaders' Suggestions" in the sandbox match TEST for various users at the "1" cutoff. Other WordCheck PM tools use this underlying code as well, but that's probably the most robust method.

Testable in the [faster-wordcheck-funcs](https://www.pgdp.org/~cpeel/c.branch/faster-wordcheck-funcs/) sandbox.

### DIFF
--- a/pinc/wordcheck_engine.inc
+++ b/pinc/wordcheck_engine.inc
@@ -971,12 +971,12 @@ function get_distinct_words_in_text($text)
     // So if $text is project-sized (say, 1.5Mb), the resulting array will be
     // rather large (about 30 Mb).)
     //
-    // Somewhat arbitrarily, we say that an intermediate array of about 1Mb is
-    // the maximum acceptable, which roughly works out to a $text of 50kb.
-    $break_point = 50 * 1024;
+    // Somewhat arbitrarily, we say that an intermediate array of about 2.5Mb is
+    // the maximum acceptable, which roughly works out to a $text of 128kb.
+    $break_point = 128 * 1024;
     // In practice, $text is either a page (usually under 5kb) or a project
-    // (usually at least 100kb), so any value for $break_point in the range of
-    // 5kb to 100kb would have the same effect on branching.
+    // (close to the 1Mb mark), so any value for $break_point in the range of
+    // 5kb to 1Mb would have the same effect on branching.
 
     if (is_string($text)) {
         if (strlen($text) < $break_point) {
--- a/pinc/wordcheck_engine.inc
+++ b/pinc/wordcheck_engine.inc
@@ -785,19 +785,24 @@ function load_word_list($path)
         return [];
     }
 
-    // Note: we're unable to use the file() function's FILE_IGNORE_NEW_LINES
-    // parameter since it wasn't added until PHP 5.0.0 *and* because
-    // it only removes \n, not \r\n, so we'd still need to rtrim later
-    $words = file($path);
-    if ($words === false) {
+    // It's faster to load the entire file into memory, remove any trailing
+    // spaces, and then explode it into an array than use file() to load
+    // it into an array and then rtrim() / rtrim_walk() the entries. Word lists
+    // should certainly be small enough that we can load the contents into
+    // memory for a brief period.
+
+    // Note that this function needs to work on files that contain \r\n line
+    // endings since that's what we write out in save_word_list() -- which
+    // precludes us from using file() with the FILE_IGNORE_NEW_LINES flag.
+    $file_contents = file_get_contents($path);
+    if ($file_contents === false) {
         return sprintf(_("Error reading file %s"), $path);
     }
 
-    // trim word list to remove trailing spaces
-    // needed in case site word list has trailing spaces
-    // And if/since the file has \r\n line-ends,
-    // this will also remove the trailing \r.
-    array_walk($words, 'rtrim_walk');
+    // remove any trailing spaces, although there shouldn't be any for
+    // lists saved with save_word_list(), and \r
+    $file_contents = preg_replace("/\s*\r*\n/", "\n", $file_contents);
+    $words = explode("\n", $file_contents);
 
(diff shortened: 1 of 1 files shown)


chrismiceli (approved):
I thought there was a way to limit suggestions to a given project? Do people want to review all their 70+ projects at once? Or is that an extreme outlier?

cpeel:
> I thought there was a way to limit suggestions to a given project? Do people want to review all their 70+ projects at once? Or is that an extreme outlier?

There are two ways to manage suggestions - by project and for all projects in rounds. The latter was added soon after WordCheck was rolled out in 2006 IIRC so that PMs didn't have to visit every project to review suggestions.

chrismiceli:
> 

That's what I thought. Thanks for clarifying. Should we add a round filter too maybe to the manage suggestions page? Just thinking out loud but not sure what PM workflows look like to be honest.

cpeel:
> Should we add a round filter too maybe to the manage suggestions page? Just thinking out loud but not sure what PM workflows look like to be honest.

I don't know if that would help since they'd then need to manually iterate over all the rounds to see all of their projects, but I don't fully grok the PM workflows either.

pinc/wordcheck_engine.inc
```diff
@@ -785,19 +785,24 @@ function load_word_list($path)
...
     }
 
-    // Note: we're unable to use the file() function's FILE_IGNORE_NEW_LINES
-    // parameter since it wasn't added until PHP 5.0.0 *and* because
-    // it only removes \n, not \r\n, so we'd still need to rtrim later
-    $words = file($path);
-    if ($words === false) {
+    // It's faster to load the entire file into memory, remove any trailing
+    // spaces, and then explode it into an array than use file() to load
+    // it into an array and then rtrim() / rtrim_walk() the entries. Word lists
```
cpeel, line 790: A quick check says that we actually have a couple of word lists that are 512K. I need to confirm this is still OK with word lists of that size.

pinc/wordcheck_engine.inc
```diff
@@ -785,19 +785,24 @@ function load_word_list($path)
...
     }
 
-    // trim word list to remove trailing spaces
-    // needed in case site word list has trailing spaces
-    // And if/since the file has \r\n line-ends,
-    // this will also remove the trailing \r.
-    array_walk($words, 'rtrim_walk');
+    // remove any trailing spaces, although there shouldn't be any for
+    // lists saved with save_word_list().
+    $file_contents = preg_replace("/\s+\r*\n/", "\n", $file_contents);
```
70ray, line 804: Isn't this equivalent to "/\s+/" since \s matches space or \r or \n ?

srjfoo:
> > > > I thought there was a way to limit suggestions to a given project? Do people want to review all their 70+ projects at once? Or is that an extreme outlier?
> > >
> > > There are two ways to manage suggestions - by project and for all projects in rounds. The latter was added soon after WordCheck was rolled out in 2006 IIRC so that PMs didn't have to visit every project to review suggestions.

> > That's what I thought. Thanks for clarifying. Should we add a round filter too maybe to the manage suggestions page? Just thinking out loud but not sure what PM workflows look like to be honest.

> I don't know if that would help since they'd then need to manually iterate over all the rounds to see all of their projects, but I don't fully grok the PM workflows either.

I think each PM tries to come up with a workflow that suits them. If a PM keeps up with everything (and depending on how high-volume they are, and how quickly their projects move, that could mean checking in multiple times a day), then the number of words needing review at any given time may be small, and being able to pull them all at once might be more useful. If there are huge numbers of words to be reviewed, though, a PM might not have time in any give sitting to review everything.

I've made a third suggestion in the task -- to use the fact that projects having proofer suggestions to be reviewed can be identified by enabling and checking the "actions" column in the PM search results page. The WC interface for the project can be brought up via that menu. Whether or not the suggestion will be useful, I don't know.

In any case, I've done some testing over several PMs on TEST, and both TEST and the sandbox show the same results, so all that looks good. However, there isn't the volume of data to be able to tell whether there's a speed advantage of one over the other.

srjfoo (approved):

pinc/wordcheck_engine.inc
cpeel: Good question. In this case I only want to trim **trailing** whitespaces, so I want to anchor the pattern with a newline. If I just removed `/\s+/` it would remove any duplicate whitespace which might also include whitespace elsewhere. There shouldn't be whitespace elsewhere, but I want to replicate what was happening before.

cpeel:
> In any case, I've done some testing over several PMs on TEST, and both TEST and the sandbox show the same results, so all that looks good. However, there isn't the volume of data to be able to tell whether there's a speed advantage of one over the other.

Yes, that is a continued problem. In this case at least I feel confident that we are at least no worse off based on the number of CPU cycles thanks to the profiler.

pinc/wordcheck_engine.inc
cpeel: I'm comfortable with this being OK and the change worth the memory cost for these larger files.

70ray (approved):

pinc/wordcheck_engine.inc
cpeel: Although your comment just now made me realize that I ought to use `/\s*\r*\n/` so it's clearer that we're correctly handling the case of `\r`s. Right now those are working because they're being consumed by the `\s` which might be confusing later. I'll get that updated now.

PR #727, opened 2021-12-30
Reduce quiz overhead for non-quiz pages

cpeel:
I found a few code improvements using the xdebug profiler, this is the first one and focuses on decreasing unnecessary overhead of every page load.

`pinc/Activity.inc` -- which virtually every page load uses -- includes `pinc/quizzes.inc` to determine the user's access privileges. `pinc/quizzes.inc` includes `quiz/generic/quiz_defaults.inc` which defines a very large global `$messages` array that is only used in the quizzes. This breaks apart that dependency which reduces the overhead for every page load.

This also improves the query and logic used to determine if a user has passed a quiz.

Testable in the [reduce-quiz-overhead](https://www.pgdp.org/~cpeel/c.branch/reduce-quiz-overhead/) sandbox. Primary thing to test is that the logic for determining if a user has passed a quiz (including aging out of older quizzes) still works as it did before.

### DIFF
--- a/pinc/Quiz.inc
+++ b/pinc/Quiz.inc
@@ -69,27 +69,17 @@ class Quiz
             SELECT *
             FROM quiz_passes
             WHERE username = '%s'
-        ", DPDatabase::escape($username));
+                AND quiz_page IN (%s)
+                AND result = 'pass'
+        ", DPDatabase::escape($username),
+            surround_and_join(array_keys($this->pages), '"', '"', ","));
         $result = DPDatabase::query($sql);
-        while ($attempt = mysqli_fetch_object($result)) {
-            if ($attempt->result != "pass") {
-                continue;
-            }
-            if (array_key_exists($attempt->quiz_page, $pages_required_results)) {
-                $pages_required_results[$attempt->quiz_page] = $attempt->date;
-            } else {
-                //This quiz page isn't relevant to this quiz.
-                continue;
-            }
-        }
 
-        if (isset($this->pass_requirements['maximum_age'])) {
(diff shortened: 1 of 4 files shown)


pinc/Quiz.inc
```diff
@@ -68,27 +69,17 @@ class Quiz
...
+            $pages_required_results[$attempt->quiz_page] = $attempt->date;
 
-        if (isset($this->pass_requirements['maximum_age'])) {
-            foreach ($pages_required_results as $quiz_page_id => $value) {
-                if ($value == "no") {
-                    // The user hasn't passed the quiz anyway.
-                    continue;
-                }
-                if ((time() - $value) > $this->pass_requirements['maximum_age']) {
+            if (isset($this->pass_requirements['maximum_age'])) {
```
chrismiceli, line 81: I'd combine the 2 ifs into one with && but not important.

chrismiceli (approved):
Love these improvements!

srjfoo (approved):
> Testable in the [reduce-quiz-overhead](https://www.pgdp.org/~cpeel/c.branch/reduce-quiz-overhead/) sandbox. Primary thing to test is that the logic for determining if a user has passed a quiz (including aging out of older quizzes) still works as it did before.

From all I could tell by looking at multiple userIDs and actually "passing" quizzes that had timed out, said logic appears to work just fine.

pinc/Quiz.inc
cpeel: I'd like to leave these separate for a bit more clarity in the code.

PR #728, opened 2021-12-30
More code optimizations for page loads

cpeel:
Another optimization found via xdebug profiling.
 
Determining a user's access to a round is done by computing a score based on the activity's (stage, round, etc) criteria. Only determine the score based on the necessary activity's criteria, rather than doing it based on all criteria and only using the ones for the activity. This is oddly expensive because we need to know a user's access for every activity to create the navbar links and some, but not all, activities require quiz passes. Only do the work that is actually necessary to determine a user's access.

Testable in the [more-code-optimizations](https://www.pgdp.org/~cpeel/c.branch/more-code-optimizations/) sandbox. Easiest way to test this is to confirm that the activity links in the navbar are consistent with the user's access and that the user's sat/unsat/etc page (available to SAs) matches what's live on TEST.

### DIFF
--- a/pinc/Activity.inc
+++ b/pinc/Activity.inc
@@ -246,7 +246,7 @@ function get_user_scores($username, $criteria)
         return [];
     }
 
-    $user_obj = new User($username);
+    $user_obj = $username == User::current_username() ? User::load_current() : new User($username);
 
     $user_scores = [];
     foreach ($criteria as $criterion_code => $criterion_descr) {
--- a/pinc/Activity.inc
+++ b/pinc/Activity.inc
@@ -113,7 +113,7 @@ class Activity
             return $uao;
         }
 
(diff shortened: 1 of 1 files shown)


chrismiceli (approved):

srjfoo:
> Testable in the [more-code-optimizations](https://www.pgdp.org/~cpeel/c.branch/more-code-optimizations/) sandbox. Easiest way to test this is to confirm that the activity links in the navbar are consistent with the user's access and that the user's sat/unsat/etc page (available to SAs) matches what's live on TEST.

Caveat -- P1 access is different between the code deployed to TEST and the sandbox, because the sandbox does not have the hacks in the `pgdp-production` branch.

srjfoo (approved):

PR #730, opened 2021-12-31
My Projects  & Search page optimizations

cpeel:
Last one, promise.

This has a minor improvement to the My Projects page to only update the user's My Project view settings in the DB if they've changed. This reduces some thrashing on the database in the most common case.

It has a major improvement to the PM and Search pages for users who are authorized to change page states. The aggregate time to determine what states the user can transition the projects to was almost the same as the SQL search for the list of projects. Caching viable transitions for a "from" state can help subsequent calls for later projects on the PM and Search pages.

Testable in the [myproject-search-optimizations](https://www.pgdp.org/~cpeel/c.branch/myproject-search-optimizations/) sandbox. The areas to watch out for are making sure the My Projects page correctly "remembers" the table sort settings when changed and that the states available in the drop-downs on the PM and Search pages for PMs & SAs are the same as they were prior.

### DIFF
--- a/tools/proofers/my_projects.php
+++ b/tools/proofers/my_projects.php
@@ -51,12 +51,12 @@ $pool_sort = get_enumerated_param(
     $pool_sort_options
 );
 
-// Update saved view and sort settings
-$userSettings->set_value("my_projects:round_view", $round_view);
-$userSettings->set_value("my_projects:pool_view", $pool_view);
-$userSettings->set_value("my_projects:round_sort", $round_sort);
-$userSettings->set_value("my_projects:pool_sort", $pool_sort);
-
+// Update saved view and sort settings if they've changed
+foreach (["round_view", "pool_view", "round_sort", "pool_sort"] as $setting) {
+    if ($$setting != $userSettings->get_value("my_projects:$setting")) {
+        $userSettings->set_value("my_projects:$setting", $$setting);
+    }
+}
 
 $page_header = [
     "text_self" => _("My Projects"),
--- a/tools/proofers/my_projects.php
+++ b/tools/proofers/my_projects.php
@@ -3,7 +3,6 @@ $relPath = '../../pinc/';
 include_once($relPath.'base.inc');
 include_once($relPath.'misc.inc'); // array_get(), surround_and_join(), html_safe()
 include_once($relPath.'theme.inc');
-include_once($relPath.'dpsql.inc');
 include_once($relPath.'stages.inc');
 include_once($relPath.'Project.inc');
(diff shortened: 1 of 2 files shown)


tools/proofers/my_projects.php
```diff
@@ -52,12 +51,12 @@
...
 
-// Update saved view and sort settings
-$userSettings->set_value("my_projects:round_view", $round_view);
-$userSettings->set_value("my_projects:pool_view", $pool_view);
-$userSettings->set_value("my_projects:round_sort", $round_sort);
-$userSettings->set_value("my_projects:pool_sort", $pool_sort);
-
+// Update saved view and sort settings if they've changed
+foreach (["round_view", "pool_view", "round_sort", "pool_sort"] as $setting) {
+    if ($$setting != $userSettings->get_value("my_projects:$setting")) {
```
chrismiceli, line 56: is getting the current value expensive or does not writing make up for this cost?

chrismiceli (approved):

tools/proofers/my_projects.php
cpeel: Getting the current value is an in-memory operation from within the `$userSettings` object so it's very cheap.

srjfoo:
> Testable in the [myproject-search-optimizations](https://www.pgdp.org/~cpeel/c.branch/myproject-search-optimizations/) sandbox. The areas to watch out for are making sure the My Projects page correctly "remembers" the table sort settings when changed ...

I've seen two different sorts of persistence (for both the sandbox and the main TEST site:
1. The sort column set in one tab will persist if that column exists in the next tab chosen. In particular, the PP tabs don't all have a uniform set of columns. If the next tab chosen does not have that column, the sort defaults to a title sort for display, but does not reset the sort column, because if no sort column change is made, the last chosen sort column is still used if the user changes to a tab that has that sort column. Sort columns within a tab group apply only to that tab group.
2. Switching between the sandbox and the main TEST site, if the sort is changed in one, and a new tab that has that sort column is chosen on the other, that persists. However, if it is changed on one, but the tab is only refreshed on the other (not actually changed), the tab retains its original setting.

Both make sense to me, but I wanted to confirm that that's expected behavior.

> and that the states available in the drop-downs on the PM and Search pages for PMs & SAs are the same as they were prior.

I did not do exhaustive testing (I can do more if you wish), but as far as I can tell, this is the case. PM only can see only the state changes for their own projects; PFs can see only the state changes they are permitted to make, but for all projects; squirrels can see everything in the sandbox that they can on TEST.

srjfoo (approved):

PR #731, opened 2022-01-01
Faster wordchecks

cpeel:
Shockingly, half the time spent running WordCheck is building the list of available additional dictionaries for the drop-down. This is expensive because `langname_for_langcode3()` is called within several other language functions that get used as part of generating the list, resulting in 69k calls to `substr()`. We can reduce this by only checking the 23 possibly-valid dictionary files that have a 2- or 3-letter code rather than all 84 of them. Creating a reverse lookup cache would be even faster, but I'm not certain the memory overhead is worth the value.

Also correct the logic error where we use `substr()` to split apart combined language codes. Pulling out the first three letters isn't correct and will lead to `qaa` being a valid code when it is actually part of the `qaa-qtz` reserved range.

Testable in the [faster-wordchecks](https://www.pgdp.org/~cpeel/c.branch/faster-wordchecks/) sandbox. Success is that the same list of dictionaries are listed in the dropdown on the WordCheck page.

### DIFF
--- a/pinc/iso_lang_list.inc
+++ b/pinc/iso_lang_list.inc
@@ -623,8 +623,7 @@ function langname_for_langcode3($langcode3)
 // or null if not found.
 {
     foreach (get_iso_language_list() as $entry) {
-        $entry_code = $entry['lang_code'];
-        if (substr($entry_code, 0, 3) == $langcode3 || substr($entry_code, 4, 3) == $langcode3) {
+        if (in_array($langcode3, explode("/", $entry['lang_code']))) {
             return $entry['lang_name'];
         }
     }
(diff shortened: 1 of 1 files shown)


chrismiceli (approved):

70ray (commented):
Would it save time for the function `langname_for_langcode($langcode)` to test whether the code is 2 or 3 letters and then call one of `langname_for_langcode2` or `langname_for_langcode3`?

cpeel:
> Would it save time for the function `langname_for_langcode($langcode)` to test whether the code is 2 or 3 letters and then call one of `langname_for_langcode2` or `langname_for_langcode3`?

It would but only for 1 entry of the 24 so the improvement would be miniscule. The only 3-letter langcode we have that is used in this case (ie that we have a dictionary for) is `grc`.

70ray (approved):

70ray:
If the list rarely changes one could even save it in a file, and only make the file if it doesn't exist. It would only need to be updated when adding new dictionaries.

cpeel:
> If the list rarely changes one could even save it in a file, and only make the file if it doesn't exist. It would only need to be updated when adding new dictionaries.

Yes. We could also pre-create the reverse-lookup index in the code too which would make it O(1) for each dictionary instead of O(n). I'm not certain any of these is worth the update maintenance at this time.

srjfoo (approved):

Issue #732, opened 2022-01-03
Incorrect validation of email domain name during user registration


wf49670:
pinc/email_address.inc includes this line which only allows email domain names that are between 2 and 4 characters in length:
`$domain_re = "($label_re\.)+[A-Za-z]{2,4}";`

A recently registered user pointed out that there are now domain names, like the one she works for, which are longer. Example: samvid.ventures, so we should adjust that validation step.

cpeel:
According [to this](https://developer.mozilla.org/en-US/docs/Learn/Common_questions/What_is_a_domain_name) the maximum length of a TLD is 63 characters.

FYI: [here's the list of all current TLDs](https://data.iana.org/TLD/tlds-alpha-by-domain.txt), not that I think we should try to validate against it.

PR #733, opened 2022-01-04
Accept longer TLDs in email addresses

cpeel:
Increase the length of a valid TLD to 63 characters. While I was here I cleaned up a small localization error and removed an erroneous include.

Fixes https://github.com/DistributedProofreaders/dproofreaders/issues/732.

This isn't easily testable on TEST because we skip email validation when `$testing = true`, but here's validation via script:
```php
<?php
$relPath = "./";
include_once($relPath."email_address.inc");

$addresses = [
    "testing@valid.com",
    "person@something.averylongtld",
    "localhost",
    "averyeryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryverylong@valid.com",
];

foreach ($addresses as $address) {
    echo $address . "\n";
    echo "    " . check_email_address($address) . "\n";
}
```
Output:
```
$ php test.php
testing@valid.com

person@something.averylongtld

localhost
    Your E-mail Address does not contain an '@' sign.
averyeryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryveryverylong@valid.com
    Your E-mail Address is too long. The maximum is 100 characters.
```

### DIFF
--- a/accounts/addproofer.php
+++ b/accounts/addproofer.php
@@ -251,7 +251,7 @@ function _validate_fields($real_name, $username, $userpass, $userpass2, $email, 
 
     // In testing mode, a fake email address is constructed using
     // 'localhost' as the domain. check_email_address() incorrectly
-    // thinks the domain should end in a 2-4 character top level
+    // thinks the domain should end in a 2-63 character top level
     // domain, so disable the address check for testing.
     if (!$testing) {
         $err = check_email_address($email);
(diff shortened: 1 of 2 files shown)


chrismiceli (approved):

70ray (approved):

srjfoo (approved):

wf49670 (commented):
Looks good. And thanks for including the test results and methodology; that will be helpful elsewhere in my work.

PR #734, opened 2022-01-08
Add api proofing functions and test interface

70ray:
This adds some api functions for proofreading and a test interface. It does not do any character validation or wordcheck or format preview. Most of the functions haven't yet been added to the yaml file. Not intended for merging but for preliminary evaluation.
From the project page it moves to a basic proofreading page.
Sandbox at: https://www.pgdp.org/~rp31/c.branch/api_proof_22_01

### DIFF
--- a/api/v1.inc
+++ b/api/v1.inc
@@ -4,6 +4,7 @@ include_once("ApiRouter.inc");
 // DP API v1
 include_once("v1_validators.inc");
 include_once("v1_projects.inc");
+include_once("v1_proof.inc");
 include_once("v1_stats.inc");
 
 $router = ApiRouter::get_router();
@@ -14,6 +15,9 @@ $router->add_validator(":projectid", "validate_project");
 $router->add_validator(":roundid", "validate_round");
 $router->add_validator(":pagename", "validate_page_name");
 $router->add_validator(":pageroundid", "validate_page_round");
+$router->add_validator(":projectstateid", "validate_project_state");
+$router->add_validator(":pagestateid", "validate_page_state");
+$router->add_validator(":versionname", "validate_page_version");
 
 // Add routes
 $router->add_route("GET", "v1/projects", "api_v1_projects");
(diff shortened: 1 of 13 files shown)


api/v1.inc
```diff
@@ -31,6 +35,12 @@ $router->add_route("GET", "v1/projects/imagesources", "api_v1_projects_imagesour
 $router->add_route("GET", "v1/projects/:projectid/pages", "api_v1_project_pages");
 $router->add_route("GET", "v1/projects/:projectid/pagedetails", "api_v1_project_pagedetails");
 $router->add_route("GET", "v1/projects/:projectid/pages/:pagename/pagerounds/:pageroundid", "api_v1_project_page_round");
+$router->add_route("POST", "v1/projects/:projectid/projectstates/:projectstateid/checkout", "api_v1_project_state_checkout");
```
cpeel, line 38: I don't think we want to couple any of the update APIs to the page state. The state is an intrinsic part of the page and not a subselector. The current interface passes the state around to ensure that someone doesn't click on a link to update a page that has changed state from what was expected, but we shouldn't try to codify that within the API definitions.

I'm envisioning that the APIs that we will need are:
```
# save page (mirrors the GET)
POST v1/projects/:projectid/pages/:pagename/pagerounds/:pageroundid

# get the next page to proof; marks a page for the user
# and returns the info needed to load the page (page name, round)
POST v1/projects/:projectid/pages/proofread (or /checkout or /next or something similar)
```

To save a page or return a page, we use the same "save page" function and pass in the state (ie save, done, return).

api/v1_proof.inc
```diff
@@ -0,0 +1,388 @@
...
+// checkout a page to a user
+// update the project table in one operation, then select the page we have
+// checked out. (This could be modified to enable checking out several pages)
+// Copying the text from previous round is different to previous action.
+// POST because not idempotent
+// POST v1/projects/:projectid/projectstates/:projectstateid/checkout
+// (similar to LPage get_available_page())
+function api_v1_project_state_checkout($method, $data, $query_params)
+{
+    global $pguser, $preceding_proofer_restriction;
```
cpeel, line 17: We need to figure out how to reuse existing LPage functionality or do some refactoring rather than duplicating logic. This is a bit of a problem already in the API code (most of it my fault) and I don't want to do that for very critical code that manipulates data.

If we were going to do a wholesale cutover and remove the old way and only use the new way then I wouldn't worry about it, but I think doing a successful UI redesign is going to be a staged process that involves rolling out the new API alongside the current UI, then developing the new UI to use the API, then removing the old UI.

Part of the heavy lift on Task 2011 is that I'm pulling out the project validation logic into a single place for re-use in the UI and the API.

cpeel (commented):
Thanks for working on this Ray! I'm excited to see progress on this important effort!

api/v1_proof.inc
70ray: I'll have a look into LPage.

api/v1.inc
70ray: I see what you mean about the page state.
The checkout you envisage is very close to what I've done.

cpeel: > The checkout you envisage is very close to what I've done.

Yes, exactly.

PR #737, opened 2022-01-09
Streamline log_project_event

cpeel:
Pull `log_project_event()` into the `Project` class and rely on DPDatabase to raise an exception if the INSERT fails. Similarly rely on DPDatabase to raise an exception if the project CREATE TABLE fails. This is a stepping stone for error handling for Task 2011.

Testable in the [streamline_log_project_event](https://www.pgdp.org/~cpeel/c.branch/streamline_log_project_event/) sandbox. I've confirmed that project edit events are logged as are project holds. Project page table creation works as well.


pinc/ProjectTransition.inc
```diff
@@ -294,21 +293,16 @@ class ProjectTransition
         if (!is_null($pool) && $this->to_state == $pool->project_checkedout_state) {
             // we want the project event to record who it's been
             // checked out to
-            $project = new Project($projectid);
```
chrismiceli, line 297: there wasn't any good reason a new project object was created here?

chrismiceli (approved):

pinc/ProjectTransition.inc
cpeel: This isn't necessary because the function is passed a `$project` object. I suspect (but haven't dug into the code) that this was here first and then the function was refactored to take a Project object rather than a `$projectid` and this was just left around.

chrismiceli: That was my idea as well. Just confirming.

pinc/ProjectTransition.inc
```diff
@@ -294,21 +293,16 @@ class ProjectTransition
...
         }
         // log the project event
-        $err = log_project_event(
-            $projectid,
+        $project->log_project_event(
             $who,
             'transition',
             $this->from_state,
             $this->to_state,
             ($checkedoutby != '' ? "out_to: $checkedoutby" : @$extras['details'])
```
70ray, line 304: Should we do something better instead of `@extras`? extras is alawys an empty array so far as I can tell so prehaps remove it.

70ray (approved):

pinc/ProjectTransition.inc
cpeel: Very often, but not quite always. `$extras` is used within `tools/project_manager/autorelease.inc` to include the queue name:
```
    $error_msg = project_transition($projectid, $new_state, PT_AUTO, ['details' => "via_q: $queue_name"]);
calls:
    $error = $transition->do_state_change($project, $who, $extras);
calls:
    $project->log_project_event(...)
```
I'm not 100% certain that it's actually used in this path, but it could be.

chrismiceli (approved):

tools/project_manager/editproject.php
```diff
@@ -630,26 +630,16 @@ public function save_to_db()
...
-                                        PROJ_POST_FIRST_CHECKED_OUT,
-                                        $this->checkedoutby
-                                        );
-                if (!empty($e)) {
-                    die($e);
-                }
+                $project->log_project_event($pguser, 'transition',
+                                            PROJ_POST_FIRST_CHECKED_OUT,
+                                            PROJ_POST_FIRST_CHECKED_OUT,
+                                            $this->checkedoutby);
```
srjfoo, line 639: I don't know where this is happening, but what I'm seeing is that details3 is not getting recorded in your sandbox, but it is in the main TEST sandbox:
image

The top checkout in the screenshot is from your sandbox, the bottom, a couple of minutes later, from TEST. (I checked the `project_events` table, and it's not getting recorded at all.) I double-checked using another project, and got the same results.

I've looked at ` pinc/ProjectTransition.inc`, line 299+ and at `pinc/Project.inc`, line 799+, but I'm just not seeing the problem.

Everything else I was able to test looks good!

cpeel: Can you describe your testing of this path? Was the project in "available for PP" and you edited the project and filled in the PPer?

srjfoo: No, I hadn't tested that -- thanks for the idea. I just checked the projects out in the usual way: look at the list of available projects; pick one; check it out from the project page. I was doing side-by-side comparisons with checking it out via the main TEST sandbox, just to make sure I knew how it should look.

You can see the differences in the last set of events in the project history for [this](https://www.pgdp.org/c/project.php?id=projectID4aeffe8625b41) project. The checkout at `2022-01-16 04:16:08` was done from TEST, in parallel with my first checkout from your sandbox. The checkout at `2022-01-16 04:51:08` was my second checkout from your sandbox, to verify that I hadn't messed up somewhere.

srjfoo: > Was the project in "available for PP" and you edited the project and filled in the PPer?

I've tested this just now to confirm what I thought was the case. Just editing the PP/PPV field (i.e. checkedoutby) in a project page will not automatically check a project out to a PPer if it is in the pool. Squirrels can then use `project_jump.php` to move it into PP checked out to that individual, but PMs and PFs cannot.

There has been some discussion of the issue in the past: editing the field in the project page is an edit operation, but moving it from available for PP to in PP checked out to someone is a project transition. Should a project transition be triggered by editing the project page?

cpeel: > Should a project transition be triggered by editing the project page?

According to line [594](https://github.com/DistributedProofreaders/dproofreaders/blob/master/tools/project_manager/editproject.php#L594) if a project was already in the PP checked out state and the PPer is changed it will check out the project to the new PPer. I just wanted to make sure I knew how you were testing it to see where I broke it. I can see the related code but I haven't figured out why my change broke it. More for me to investigate.

pinc/ProjectTransition.inc
cpeel: @srjfoo, this is the source of the bug you found. So despite the project object existing, reloading it from the DB is required to load the right `checkedoutby` person. I'm digging into *why* the project object doesn't have the right values here.

cpeel: Ha! Found, fixed, and commented why this is needed.

srjfoo: Excellent!

srjfoo (approved):
Unless you want me to go through and re-test, looks good to me, now.

cpeel:
> Unless you want me to go through and re-test, looks good to me, now. (edit: I did re-test the checkout)

The only changes were in that one place, so I'm good as long as that one looks good.

pinc/ProjectTransition.inc
cpeel: One last force commit to remove code cleanup in this file that was not destined for this PR.

srjfoo:
> > Unless you want me to go through and re-test, looks good to me, now. (edit: I did re-test the checkout)
> 
> The only changes were in that one place, so I'm good as long as that one looks good.

Yup, that change looks good.

PR #739, opened 2022-01-12
Add ability to graph subset of pages per month

chrismiceli:
https://www.pgdp.org/~chrismiceli/c.branch/misc_stats1/stats/misc_stats1.php?start=2010-01&end=2015-01&tally_name=P1

@lhamilton1, let me know if this is what you wanted. You'll have to scroll down to the pages per month table to see the form and graph
image

### DIFF
--- a/stats/misc_stats1.php
+++ b/stats/misc_stats1.php
@@ -5,11 +5,14 @@ include_once($relPath.'dpsql.inc');
 include_once($relPath.'theme.inc');
 include_once($relPath.'page_tally.inc'); // get_page_tally_names()
 include_once($relPath.'misc.inc'); // get_enumerated_param()
+include_once($relPath.'graph_data.inc');
 
 require_login();
 
 $valid_tally_names = array_keys(get_page_tally_names());
 $tally_name = get_enumerated_param($_GET, 'tally_name', null, $valid_tally_names);
+$start = get_param_matching_regex($_GET, 'start', null, '/$(\d{4}-\d{1,2})?$/', true);
+$end = get_param_matching_regex($_GET, 'end', null, '/$(\d{4}-\d{1,2})?$/', true);
 
 // -----------------------------------
 
@@ -20,7 +23,38 @@ $curr_year = strftime('%Y', $now_timestamp);
 // -----------------------------------
 
 $title = sprintf(_("Miscellaneous Statistics for Round %s"), $tally_name);
-output_header($title);
+
+$js_data = "";
+if (isset($start) && isset($end)) {
+    $sql = select_from_site_past_tallies_and_goals(
+      $tally_name,
+      "SELECT 
+        {year_month} as 'month',
+        CAST(SUM(tally_delta) AS SIGNED),
+        past_tallies.timestamp - $SECONDS_TO_YESTERDAY as 'timestamp'",
+      "AND timestamp >= UNIX_TIMESTAMP(STR_TO_DATE('$start-01', '%Y-%m-%d')) AND
+      timestamp <= UNIX_TIMESTAMP(LAST_DAY(STR_TO_DATE('$end-01', '%Y-%m-%d')))",
+      "GROUP BY 1",
+      "ORDER BY 1",
+      "");
+    $result = DPDatabase::query($sql);
+    [$datax, $datay] = dpsql_fetch_columns($result);
+    $graphConfig = [
+        "title" => "$tally_name Pages Completed by Month",
+        "data" => [
+            "" => [
+                "x" => $datax,
+                "y" => $datay,
+                "type" => "line",
+            ],
+        ],
+    ];
+    $js_data .= build_svg_graph_inits([["barLineGraph", "graph", $graphConfig]]);
+}
+output_header($title, SHOW_STATSBAR, [
+    "js_files" => get_graph_js_files(),
+    "js_data" => $js_data,
+]);
 
 echo "<h1>$title</h1>\n";
 
@@ -99,7 +133,7 @@ function show_top_days($n, $when)
 
 function show_month_sums($which)
 {
-    global $tally_name, $curr_year_month;
+    global $tally_name, $curr_year_month, $start, $end;
 
     switch ($which) {
         case 'top_ten':
@@ -112,6 +146,15 @@ function show_month_sums($which)
             $sub_title = _("Historical Log of Total Pages Proofread Per Month");
             $order = '1'; // chronological
             $limit = '';
+            $form = "<form>
+            <label>Start <input value='$start' placeholder='YYYY-MM' name='start'></label>
+            <label>End <input value='$end' placeholder='YYYY-MM' name='end'></label>
+            <input type='hidden' name='tally_name' value='$tally_name'>
+            <input type='submit' value='Graph'>
+            </form>
+            <br>
+            <div id='graph' style='max-width: 640px;max-height: 400px'></div>
+            ";
             break;
 
         case 'all_by_pages':
@@ -143,6 +186,10 @@ function show_month_sums($which)
             $limit
         ), 1, DPSQL_SHOW_RANK
     );
+
+    if (isset($form)) {
+        echo $form;
+    }
 
     echo "<br>\n";
 }


lhamilton1:
It looks pretty good. I notice though that if I put a start date of 2015-01 and an end date of 2022-01, the chart includes the data we have in January 2022 -- but the starting month of the data gives us all of December 2014 instead of January 2015

lhamilton1:
I notice too that if I pick 2010-01 to 2015-01 the graph starts at 2010-07

stats/misc_stats1.php
```diff
@@ -5,11 +5,14 @@
 include_once($relPath.'theme.inc');
 include_once($relPath.'page_tally.inc'); // get_page_tally_names()
 include_once($relPath.'misc.inc'); // get_enumerated_param()
+include_once($relPath.'graph_data.inc');
 
 require_login();
 
 $valid_tally_names = array_keys(get_page_tally_names());
 $tally_name = get_enumerated_param($_GET, 'tally_name', null, $valid_tally_names);
+$start = get_param_matching_regex($_GET, 'start', null, '/$(\d{4}-\d{1,2})?$/', true);
```
cpeel, line 14: I think this regex should start with `^` to anchor it at the beginning.

stats/misc_stats1.php
```diff
@@ -20,7 +23,38 @@
 // -----------------------------------
 
 $title = sprintf(_("Miscellaneous Statistics for Round %s"), $tally_name);
-output_header($title);
+
+$js_data = "";
+if (isset($start) && isset($end)) {
+    $sql = select_from_site_past_tallies_and_goals(
+      $tally_name,
```
cpeel, line 30: 4-space indents, please.

stats/misc_stats1.php
cpeel: Since we're passing these in as-is to the SQL we want to be sure and test to get these right.

stats/misc_stats1.php
```diff
@@ -112,6 +146,15 @@ function show_month_sums($which)
             $sub_title = _("Historical Log of Total Pages Proofread Per Month");
             $order = '1'; // chronological
             $limit = '';
+            $form = "<form>
+            <label>Start <input value='$start' placeholder='YYYY-MM' name='start'></label>
```
cpeel, line 150: We probably want to translate `Start` and `End`.

srjfoo:
> I notice too that if I pick 2010-01 to 2015-01 the graph starts at 2010-07

That's because 2010-07 is the first month in 2010 that has any data, I think.

srjfoo:
@chrismiceli -- is the graph associated with the table above it, or below it? I guess it doesn't really matter, as it's the same data, just sorted differently. Minor display issue: maybe a little bit more separation between the bottom of the table above it and the top of the form?

lhamilton1:
> 
Of course  -- it's because the data for that year doesn't start til July. Thanks

srjfoo:
> @chrismiceli -- is the graph associated with the table above it, or below it? I guess it doesn't really matter, as it's the same data, just sorted differently. Minor display issue: maybe a little bit more separation between the bottom of the table above it and the top of the form?

> 

Also, I'd recommend adding an `id` and linking directly to the graph if a graph is generated.

Another question -- there appears to be no way to clear the graph from the page once generated, other than editing the URL. Should there be, or is that something that doesn't really matter?

chrismiceli:
> @chrismiceli -- is the graph associated with the table above it, or below it? I guess it doesn't really matter, as it's the same data, just sorted differently. Minor display issue: maybe a little bit more separation between the bottom of the table above it and the top of the form?

The table above. I put a little space because it was close.

chrismiceli:
> > @chrismiceli -- is the graph associated with the table above it, or below it? I guess it doesn't really matter, as it's the same data, just sorted differently. Minor display issue: maybe a little bit more separation between the bottom of the table above it and the top of the form?
> 
> > 
> 
> Also, I'd recommend adding an `id` and linking directly to the graph if a graph is generated.
> 
> Another question -- there appears to be no way to clear the graph from the page once generated, other than editing the URL. Should there be, or is that something that doesn't really matter?

I added an id. Great idea! As for clearing, I'm not super sure it is important, but I'll defer to Linda if she has an opinion.

chrismiceli:
> It looks pretty good. I notice though that if I put a start date of 2015-01 and an end date of 2022-01, the chart includes the data we have in January 2022 -- but the starting month of the data gives us all of December 2014 instead of January 2015

I'm not sure on this to be honest. I'm not a sql expert, but the query should restrict rows to unix timestamps greater than the start of that first day of the month, and then convert that to a YYYY-MM string. Is the sandbox database not UTC or something strange @cpeel? I even captured the results and converted them myself to YYYY-MM and they were accurate, but the formatted string from mysql was the day before.

srjfoo:
> > @chrismiceli -- is the graph associated with the table above it, or below it? I guess it doesn't really matter, as it's the same data, just sorted differently. Minor display issue: maybe a little bit more separation between the bottom of the table above it and the top of the form?
> 
> The table above. I put a little space because it was close.

The space between the form and the graph looks great -- I was actually thinking that it needed a little more space between the bottom of the table and the form (sorry I wasn't clearer).

lhamilton1:
I think below the table rather than above makes sense. Clearing on page reload makes sense too.

srjfoo:
> > Also, I'd recommend adding an `id` and linking directly to the graph if a graph is generated.   <snip>
> 
> I added an id. Great idea!   <snip>

When I submit the form, the page does not scroll directly to the graph (I hadn't noticed before that you already had an id for the graph form. That would have been fine, too. What I think should be happening is that if someone does ask for data to be graphed, they shouldn't still have to scroll down the page to see it.

lhamilton1:
Thinking about it further -- since the data for that page is all essentially different ways of presenting the same information, having the graph at top or bottom would make sense. Top might be best.

cpeel (approved):

lhamilton1:
By top I mean above the first data table with all the data

chrismiceli:
> > > Also, I'd recommend adding an `id` and linking directly to the graph if a graph is generated.
> > 
> > 
> > I added an id. Great idea!
> 
> When I submit the form, the page does not scroll directly to the graph (I hadn't noticed before that you already had an id for the graph form. That would have been fine, too. What I think should be happening is that if someone does ask for data to be graphed, they shouldn't still have to scroll down the page to see it.

I forgot to update the sandbox apparently...now I think I can remove the id since we'll be moving it to the top :D

chrismiceli:
> > > @chrismiceli -- is the graph associated with the table above it, or below it? I guess it doesn't really matter, as it's the same data, just sorted differently. Minor display issue: maybe a little bit more separation between the bottom of the table above it and the top of the form?
> > 
> > 
> > The table above. I put a little space because it was close.
> 
> The space between the form and the graph looks great -- I was actually thinking that it needed a little more space between the bottom of the table and the form (sorry I wasn't clearer).

Also fixed but didn't update sandbox...silly me.

chrismiceli:
> By top I mean above the first data table with all the data

I agree. Moved, let me know what you think.

srjfoo:
> > By top I mean above the first data table with all the data
> 
> I agree. Moved, let me know what you think.

The first table with all the data is actually **`Historical Log of Total Pages Proofread Per Month`**, but I'm not sure it matters. Part of me likes the graph at the top of the page, proper; part of me thinks it belongs with the table that has the most complete data, and I don't think I have a strong preference.

I did notice one thing, though. I was curious about why the y-axis has a negative range, when there aren't any negative values graphed. I made the assumption until I looked at the actual data that there were no negative numbers. Turns out that there are several, but only one of them is probably large enough to actually show. But at -388 pages for that month, it should show, shouldn't it?

lhamilton1:
I'm happy with it at page top. It still feels odd though to ask it to show me data from April 2012 to July 2021 and have it, altho it shows all of July 2021 as it should, graph from March 2012 rather than from the start of April 2012.

70ray (commented):
If I put in start date 10-01 I get an error message "Fatal error: Uncaught InvalidArgumentException: Parameter 'start' ('10-01') does not match the regex /^\d{4}-\d{1,2}$/ in /home/chrismiceli/public_html/c.branch/misc_stats1/pinc/misc.inc:282 Stack trace: #0 /home/chrismiceli/public_html/c.branch/misc_stats1/stats/misc_stats1.php(14): get_param_matching_regex() #1 {main} thrown in /home/chrismiceli/public_html/c.branch/misc_stats1/pinc/misc.inc on line 282"

srjfoo:
> I'm happy with it at page top. It still feels odd though to ask it to show me data from April 2012 to July 2021 and have it, altho it shows all of July 2021 as it should, graph from March 2012 rather than from the start of April 2012.

I'm going to take a wild guess and say that, with experimentation, it looks like the graphs are somehow constrained to start with an odd-numbered month. I used several even-numbered start months and for all of them, the graph starts with the immediately preceding odd-numbered month.

srjfoo:
> If I put in start date 10-01 I get an error message "Fatal error: Uncaught InvalidArgumentException: Parameter 'start' ('10-01') does not match the regex /^\d{4}-\d{1,2}$/ in /home/chrismiceli/public_html/c.branch/misc_stats1/pinc/misc.inc:282 Stack trace: #0 /home/chrismiceli/public_html/c.branch/misc_stats1/stats/misc_stats1.php(14): get_param_matching_regex() #1 {main} thrown in /home/chrismiceli/public_html/c.branch/misc_stats1/pinc/misc.inc on line 282"

Likewise for choosing an invalid end. And yeah, I totally forgot to test with invalid data. It's very nice to have the hint in the input boxes, but that doesn't validate the input -- just assumes that people are smart enough to take the hint. :grin:

chrismiceli:
> > If I put in start date 10-01 I get an error message "Fatal error: Uncaught InvalidArgumentException: Parameter 'start' ('10-01') does not match the regex /^\d{4}-\d{1,2}$/ in /home/chrismiceli/public_html/c.branch/misc_stats1/pinc/misc.inc:282 Stack trace: #0 /home/chrismiceli/public_html/c.branch/misc_stats1/stats/misc_stats1.php(14): get_param_matching_regex() #1 {main} thrown in /home/chrismiceli/public_html/c.branch/misc_stats1/pinc/misc.inc on line 282"
> 
> Likewise for choosing an invalid end. And yeah, I totally forgot to test with invalid data. It's very nice to have the hint in the input boxes, but that doesn't validate the input -- just assumes that people are smart enough to take the hint. 😁

The error is the validation. What should I do with invalid inoit besides an error? The user can always manually type the query Params. I was avoiding client side validation but maybe could.put the HTML5 RegEx pattern on the input field but still will throw on server if invalid. Is that ok?

cpeel:
> I was avoiding client side validation but maybe could.put the HTML5 RegEx pattern on the input field but still will throw on server if invalid. Is that ok?

I think that is a great compromise.

We don't do a fantastic job of client-side validation site-wide so there's not a great pattern for us to fit into. I do not want to solve the problem in this PR, but I'd welcome some ideas on how to better handle client-side form validation for a better UX.

stats/misc_stats1.php
```diff
@@ -99,7 +146,7 @@ function show_top_days($n, $when)
 
 function show_month_sums($which)
 {
-    global $tally_name, $curr_year_month;
+    global $code_url, $tally_name, $curr_year_month, $start, $end;
```
cpeel, line 149: Looks like these changes are no longer needed in this function?

chrismiceli:
> I'm happy with it at page top. It still feels odd though to ask it to show me data from April 2012 to July 2021 and have it, altho it shows all of July 2021 as it should, graph from March 2012 rather than from the start of April 2012.

I figured this out. It is because I have to convert a string to unix time to compare to page tallies data. The conversion mysql provides is via STR_TO_DATE then convert that DATE object to unix time with UNIX_TIMESTAMP. This conversion converts the date to UTC from local before getting the timestamp. See
https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_unix-timestamp
I need a way to get STR_TO_DATE to return a timestamp, which can have a zone, which I could set to UTC. This would remove the conversion in UNIX_TIMESTAMP. It appears I can set the zone per session too which I could set the zone of the date to UTC, but this is beyond my mysql skills. Perhaps @cpeel could enlighten me.

cpeel:
> > I'm happy with it at page top. It still feels odd though to ask it to show me data from April 2012 to July 2021 and have it, altho it shows all of July 2021 as it should, graph from March 2012 rather than from the start of April 2012.
> 
> I need a way to get STR_TO_DATE to return a timestamp, which can have a zone, which I could set to UTC.

You'll probably have better luck if you calculate the timestamps in PHP and pass that directly into the query. `mktime()` is the function I would recommend. `mktime()` uses the system's current timezone so for TEST that is UTC and for PROD that will be EST.

Something like (note that I removed the unset `$SECONDS_TO_YESTERDAY` variable):
```php
    $start_timestamp = mktime(0, 0, 0, explode("-", $start)[1], 1, explode("-", $start)[0]);
    $end_timestamp = mktime(0, 0, 0, explode("-", $end)[1], 1, explode("-", $end)[0]);
    $sql = select_from_site_past_tallies_and_goals(
        $tally_name,
        "SELECT {year_month} as 'month',
            CAST(SUM(tally_delta) AS SIGNED),
            past_tallies.timestamp as 'timestamp'",
        "AND timestamp >= $start_timestamp AND timestamp < $end_timestamp",
        "GROUP BY 1",
        "ORDER BY 1",
        "");
```

chrismiceli:
> > > I'm happy with it at page top. It still feels odd though to ask it to show me data from April 2012 to July 2021 and have it, altho it shows all of July 2021 as it should, graph from March 2012 rather than from the start of April 2012.
> > 
> > 
> > I need a way to get STR_TO_DATE to return a timestamp, which can have a zone, which I could set to UTC.
> 
> You'll probably have better luck if you calculate the timestamps in PHP and pass that directly into the query. `mktime()` is the function I would recommend. `mktime()` uses the system's current timezone so for TEST that is UTC and for PROD that will be EST.
> 
> Something like (note that I removed the unset `$SECONDS_TO_YESTERDAY` variable):
> 
> ```
>     $start_timestamp = mktime(0, 0, 0, explode("-", $start)[1], 1, explode("-", $start)[0]);
>     $end_timestamp = mktime(0, 0, 0, explode("-", $end)[1], 1, explode("-", $end)[0]);
>     $sql = select_from_site_past_tallies_and_goals(
>         $tally_name,
>         "SELECT {year_month} as 'month',
>             CAST(SUM(tally_delta) AS SIGNED),
>             past_tallies.timestamp as 'timestamp'",
>         "AND timestamp >= $start_timestamp AND timestamp < $end_timestamp",
>         "GROUP BY 1",
>         "ORDER BY 1",
>         "");
> ```

Fixed the query now where the results should match the input. I changed what you had a little @cpeel  to include all days in ending month by using date format 't'. Also, the SECONDS_TO_YESTERDAY is defined in page_tally.inc and used so the where filter matches the data from the select.

srjfoo (approved):
Looks good to me.

And one last 😈 question. It finally occurred to me to test what happens if the beginning date is greater than the ending date. If it is, there's just no graph. I think that's good enough, but wanted to mention it just in case.

cpeel:
> Fixed the query now where the results should match the input. I changed what you had a little @cpeel to include all days in ending month by using date format 't'.

OH! I had a typo in mine for the end date. If you set the day to `0`,  `mktime()` [uses the last day of the month prior](https://www.php.net/manual/en/function.mktime.php) (mktime is weird). I'm not sure either is very clear so whatever you want to use is fine with me.

```php
 $end_timestamp = mktime(0, 0, 0, explode("-", $end)[1], 0, explode("-", $end)[0]);
```

cpeel (approved):

lhamilton1:
Looks good to me -- thanks, Chris!

PR #741, opened 2022-01-16
throw exceptions in LPage and DPage

70ray:
This changes 'abort' and 'die' calls to throw exceptions and catches them elsewhere. Also catches some other cases of uncaught exceptions. This will enable LPage to be used inside an api function while remaining compatible with existing code. Attempting to proofread with wrong project state or page state or unqualified user etc should give same result as before.

Sandbox at https://www.pgdp.org/~rp31/c.branch/proof_exception

### DIFF
--- a/pinc/DPage.inc
+++ b/pinc/DPage.inc
@@ -500,7 +500,7 @@ function _Page_require(
             . "    "
             . sprintf(_("But it is in %s"), $curr_page_state)
             ;
-        abort($err);
+        throw new Exception($err);
     }
 
     if (is_null($allowed_user_colname)) {
@@ -517,7 +517,7 @@ function _Page_require(
                     _("This operation (%s) can only be done by the user who has the page checked out, which you are not."),
                     $op_name)
                 ;
-            abort($err);
+            throw new Exception($err);
         }
     }
 }
--- a/pinc/DPage.inc
+++ b/pinc/DPage.inc
@@ -2,7 +2,6 @@
 include_once($relPath.'misc.inc');
 include_once($relPath.'site_vars.php');
 include_once($relPath.'stages.inc');
-include_once($relPath.'abort.inc');
 include_once($relPath.'user_project_info.inc');
 include_once($relPath.'Project.inc');
 include_once($relPath.'unicode.inc');
@@ -378,14 +377,10 @@ function Page_returnToRound($projectid, $image, $round, $user)
 
 function Page_reclaim($projectid, $image, $round, $user)
 {
-    try {
-        _Page_require($projectid, $image,
-            [$round->page_out_state, $round->page_temp_state],
-            null, null,
-            'reclaim');
-    } catch (ProjectPageException $exception) {
-        abort($exception->getMessage());
-    }
+    _Page_require($projectid, $image,
+        [$round->page_out_state, $round->page_temp_state],
+        null, null,
+        'reclaim');
 
     $setters = join(", ", [
         set_col_str("state", $round->page_avail_state),
@@ -402,14 +397,10 @@ function Page_reclaim($projectid, $image, $round, $user)
 
 function Page_clearRound($projectid, $image, $round, $user)
 {
-    try {
-        _Page_require($projectid, $image,
-            [$round->page_save_state],
-            null, null,
-            'clearRound');
-    } catch (ProjectPageException $exception) {
-        abort($exception->getMessage());
-    }
+    _Page_require($projectid, $image,
+        [$round->page_save_state],
+        null, null,
+        'clearRound');
 
     $setters = join(", ", [
         set_col_str("state", $round->page_avail_state),
@@ -458,14 +449,10 @@ function Page_markAsBad($projectid, $image, $round, $user, $reason)
 
 function Page_eraseBadMark($projectid, $image, $round, $user)
 {
-    try {
-        _Page_require($projectid, $image,
-            [$round->page_bad_state],
-            null, null,
-            'eraseBadMark');
-    } catch (ProjectPageException $exception) {
-        abort($exception->getMessage());
-    }
+    _Page_require($projectid, $image,
+        [$round->page_bad_state],
+        null, null,
+        'eraseBadMark');
 
     $setters = join(", ", [
         set_col_str("state", $round->page_avail_state),
--- a/pinc/DPage.inc
+++ b/pinc/DPage.inc
@@ -2,7 +2,6 @@
 include_once($relPath.'misc.inc');
 include_once($relPath.'site_vars.php');
 include_once($relPath.'stages.inc');
-include_once($relPath.'abort.inc');
 include_once($relPath.'user_project_info.inc');
 include_once($relPath.'Project.inc');
 include_once($relPath.'unicode.inc');
@@ -500,7 +499,7 @@ function _Page_require(
             . "    "
             . sprintf(_("But it is in %s"), $curr_page_state)
             ;
-        throw new Exception($err);
+        throw new ProjectPageException($err);
     }
 
     if (is_null($allowed_user_colname)) {
@@ -517,7 +516,7 @@ function _Page_require(
                     _("This operation (%s) can only be done by the user who has the page checked out, which you are not."),
                     $op_name)
                 ;
-            throw new Exception($err);
+            throw new ProjectPageException($err);
         }
     }
 }
--- a/pinc/DPage.inc
+++ b/pinc/DPage.inc
@@ -2,6 +2,7 @@
 include_once($relPath.'misc.inc');
 include_once($relPath.'site_vars.php');
 include_once($relPath.'stages.inc');
+include_once($relPath.'abort.inc');
 include_once($relPath.'user_project_info.inc');
 include_once($relPath.'Project.inc');
 include_once($relPath.'unicode.inc');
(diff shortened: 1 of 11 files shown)


pinc/DPage.inc
```diff
@@ -501,7 +501,7 @@ function _Page_require(
             . "    "
             . sprintf(_("But it is in %s"), $curr_page_state)
             ;
-        abort($err);
+        throw new Exception($err);
```
cpeel, line 504: We should define and throw/catch a more specific exception for project page errors in all these places. Perhaps a `ProjectPageException`? For now I would probably put this in `Project.inc`.

pinc/LPage.inc
```diff
@@ -117,7 +117,7 @@ function get_available_page_array(
         }
     } else {
         // The empty/null (i.e., no restriction) case has already been dealt with.
-        die("Bad value for 'preceding_proofer_restriction': '$preceding_proofer_restriction'.");
+        throw new Exception("Bad value for 'preceding_proofer_restriction': '$preceding_proofer_restriction'.");
```
cpeel, line 120: Your changes should remove all uses of `abort()` so we can remove the include() as well.

pinc/DPage.inc
cpeel: Also this removes all use of `abort()` in the file so we can remove the include().

cpeel (commented):
Thanks for doing this!

We should probably remove lines 69-71 in `PPage.inc` since `get_indicated_LPage()` cannot return NULL and we're explicitly throwing an exception now. That would remove `abort()` from that file and we can remove the include too.

70ray:
> Thanks for doing this!
> 
> We should probably remove lines 69-71 in `PPage.inc` since `get_indicated_LPage()` cannot return NULL and we're explicitly throwing an exception now. That would remove `abort()` from that file and we can remove the include too.

Yes, done this now.

project.php
```diff
@@ -45,7 +46,11 @@
 $DEFAULT_DETAIL_LEVEL = 3;
 
 // Validate all the input
-$projectid = get_projectID_param($_GET, 'id');
+try {
+    $projectid = get_projectID_param($_GET, 'id');
```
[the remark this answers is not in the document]
70ray: When I was testing this by manipulating the url on a proofing page in page_temp state, if you put in an invalid projectid, then if the projectid was able to get as far as get_indicated_LPage an exception would be thrown and caught. But the invalid projectid is caught before that so it seemed to make sense to display the same abort page in that case.

pinc/DPage.inc
70ray: Done these.

pinc/LPage.inc
70ray: Done this and used more specific exceptions.

project.php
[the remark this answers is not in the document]
cpeel: I appreciate trying to improve the user's error experience but I don't think we want these changes.

The only way an invalid (or not yet created) projectid can be passed in is if the user is manipulating the URL, in which case they are SOL and I would rather barf loudly with an error than try to do so gracefully. Using `abort()` here is perhaps worse since it will show links to a project page that will also be invalid (since the projectid is invalid) and the projectid output in the page isn't escaped since it is presumed to be invalid.

And a reminder that on PROD we never show the exception stack-trace, just the thrown message. eg: https://www.pgdp.net/c/tools/proofers/proof.php?projectid=invalid&proj_state=P1.proj_avail

pinc/LPage.inc
cpeel: `ValueError` would probably be a better exception to throw here but it isn't available until PHP 8. I've opened a PR to backport it into our code (I'm using it in my project validation work): https://github.com/DistributedProofreaders/dproofreaders/pull/742

project.php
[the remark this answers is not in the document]
70ray: > I appreciate trying to improve the user's error experience but I don't think we want these changes.
> 
> The only way an invalid (or not yet created) projectid can be passed in is if the user is manipulating the URL, in which case they are SOL and I would rather barf loudly with an error than try to do so gracefully. Using `abort()` here is perhaps worse since it will show links to a project page that will also be invalid (since the projectid is invalid) and the projectid output in the page isn't escaped since it is presumed to be invalid.
> 
> And a reminder that on PROD we never show the exception stack-trace, just the thrown message. eg: https://www.pgdp.net/c/tools/proofers/proof.php?projectid=invalid&proj_state=P1.proj_avail

Good. undone changes to proof.php.

pinc/LPage.inc
70ray: > `ValueError` would probably be a better exception to throw here but it isn't available until PHP 8. I've opened a PR to backport it into our code (I'm using it in my project validation work): #742

I guess I should wait for this then rebase it.

cpeel (approved):
Code-wise this looks great. Will appreciate some additional eyes on testing the changes from @srjfoo if she has bandwidth.

cpeel:
I confirmed that the throw/catch is working as-designed. Steps that I took:
1. Checked out a page for a project
2. Saved that page as in-progress
3. Opened the page back up
4. Copied the URL
5. Saved the page as done
6. Went to the copied URL
7. Confirmed I received the "Page 016.png has changed state from F1.page_out to F1.page_saved, so your request is invalid." error message on the `abort()`-provided page.

srjfoo:
> Code-wise this looks great. Will appreciate some additional eyes on testing the changes from @srjfoo if she has bandwidth.

Yes, I should -- and my apologies for not getting to it yet.

srjfoo:
I followed @cpeel's lead on the testing, and got the same results. Then I tried a couple of variations. (In both cases, testing in parallel) 

**First variation:**
1. Checked out a page for a project
2. Saved that page as in-progress
3. Opened the page back up
4. Copied the URL
5. Saved the page as done
6. Pasted the copied URL into a different browser with a different logged-in user.
7. Confirmed I received the following for both the sandbox and TEST: "Page 027.png has changed state from P3.page_temp to P3.page_saved, so your request is invalid." (variation in png number because I used different projects between the sandbox and TEST)

I got this in both cases; the error message had nothing to do with me not being the correct user (should it have?).

**Second variation:**

Pretty much the same as the first variation, EXCEPT that I did a reopen from save as done instead of reopening a page saved in progress. Results in screenshots: @70ray's sandbox:
image

TEST:
image

The screenshots are from my second attempt at this particular version, and the second time I tried to be very careful to do exactly the same thing in both sandboxes, so I'm pretty sure that the results are legitimate. It wouldn't hurt for someone to double-check me on that, though. If the difference in behavior is expected, that's fine; I just want to make sure I'm seeing what I'm supposed to be seeing.

70ray:
> I followed @cpeel's lead on the testing, and got the same results. Then I tried a couple of variations. (In both cases, testing in parallel)
> 
> **First variation:**
> 
>     1. Checked out a page for a project
> 
>     2. Saved that page as in-progress
> 
>     3. Opened the page back up
> 
>     4. Copied the URL
> 
>     5. Saved the page as done
> 
>     6. Pasted the copied URL into a different browser with a different logged-in user.
> 
>     7. Confirmed I received the following for both the sandbox and TEST: "Page 027.png has changed state from P3.page_temp to P3.page_saved, so your request is invalid." (variation in png number because I used different projects between the sandbox and TEST)
> 
> 
> I got this in both cases; the error message had nothing to do with me not being the correct user (should it have?).
> 
> **Second variation:**
> 
> Pretty much the same as the first variation, EXCEPT that I did a reopen from save as done instead of reopening a page saved in progress. Results in screenshots: @70ray's sandbox: image
> 
> TEST: image
> 
> The screenshots are from my second attempt at this particular version, and the second time I tried to be very careful to do exactly the same thing in both sandboxes, so I'm pretty sure that the results are legitimate. It wouldn't hurt for someone to double-check me on that, though. If the difference in behavior is expected, that's fine; I just want to make sure I'm seeing what I'm supposed to be seeing.

For the first variation, it would seem better to check that the user owns the page before checking the page state, but this is how it works now.

For the second variation, I haven't managed to get different errors in my branch and on test but I get uncaught errors. 
You've found a case I hadn't considered! I've put in a fix for this and updated the sandbox.

cpeel:
`get_requested_PPage()` is also called in these files which probably also need a try/catch and `abort()` to mirror the prior behavior and any other weird edge cases:
```
tools/proofers/processtext.php
tools/proofers/report_bad_page.php
tools/proofers/image_frame_std.php
tools/proofers/text_frame_std.php
```

70ray:
> `get_requested_PPage()` is also called in these files which probably also need a try/catch and `abort()` to mirror the prior behavior and any other weird edge cases:
> 
> ```
> tools/proofers/processtext.php
> tools/proofers/report_bad_page.php
> tools/proofers/image_frame_std.php
> tools/proofers/text_frame_std.php
> ```

Yes, indeed.

70ray:
file:///media/ray/disk/dp/proof_exception/tree.txt

cpeel:
I appreciate your in-depth review of all the possible code paths and this does highlight all of the paths through `processtext.php` we weren't covering before -- which is *great* -- but I think your last commit is the wrong approach. I do not think DPage or LPage should do more than raise exceptions.

`abort()` is a presentation-level detail. Rather than doing the abort at the lowest level, we should do it at the highest level where we can guarantee the caller is being served an HTML page. Indeed, `automodify.php` is primarily called from a non-HTML context and should just get the exception text.

I would encourage you to revert your prior commit, wrap the `get_requested_PPage()` call in the primary proofer pages,
```
tools/proofers/processtext.php
tools/proofers/report_bad_page.php
tools/proofers/image_frame_std.php
tools/proofers/text_frame_std.php
```
*and* wrap the entire switch block in `processtext.php` which makes all the calls to `$ppage` in a try/catch.

That won't cover some of the other non-`get_requested_PPage()` calls you identified but I think that's OK. The user will still get the correct message from the exception in these non-proofreading edgecases, they just won't be offered the same pretty view back to the project page. If we decide that we want the more graceful approach in those other places we can add them to the correct `*.php` pages.

70ray:
That seems reasonable. I'll do that.

70ray:
Revised and sandbox updated.

cpeel:
This is perfect. I tested various combinations of things that would cause the exception via the proofreading interface and I got the expected `abort()` message every time. 👍

PR #744, opened 2022-01-17
Add project validation logic to the Project class

cpeel:
This copies (but does not move) the project validation logic from `tools/project_manager/editproject.php` into the Project class. This will allow it to be used both for the UI and the API.

The validation logic in `editproject.php` is intertwined with the ProjectHoldInfo class in that file and I decided it was clearer if I did the move in two steps (this PR to create it, and a separate one to use it). I have a second branch prepared that uses this validation and is ready for review once this one goes in.

For validation I added several new unit tests. I also ran the validation code against all of the existing projects on TEST with this code:
```php
<?php
$relPath = "./";
include_once($relPath."Project.inc");

$sql = "
    SELECT projectid
    FROM projects
    ORDER BY projectid
";

$valid_project_count = 0;
$invalid_project_count = 0;
while ($row = mysqli_fetch_assoc($res)) {
    $projectid = $row["projectid"];
    $project = new Project($projectid);
    try {
        $project->validate(true);
        $valid_project_count += 1;
    } catch (Error $exception) {
        echo "$projectid\n";
        echo "    " . $exception->getMessage() . "\n";
        $invalid_project_count += 1;
    }
}
echo "\n";
echo "Valid projects: $valid_project_count\n";
echo "Invalid projects: $invalid_project_count\n";
```

Which produced this output:
```
$ php test_project_validate.php
projectID455833134db51
    Project Manager is required.
projectID45c2c47770793
    birdbird is not a PM.
projectID5d6ab3d7076b1
    r2d2 is not a valid language.
projectID61db3e758a256
    Difficulty is required.

Valid projects: 421
Invalid projects: 4
```

### DIFF
--- a/SETUP/tests/UserTest.php
+++ b/SETUP/tests/UserTest.php
@@ -7,46 +7,14 @@ class UserTest extends PHPUnit\Framework\TestCase
 
     protected function setUp(): void
     {
-        // Attempt to load our test user, if it exists don't create it
-        $sql = "SELECT username FROM users WHERE username = '$this->TEST_USERNAME'";
-        $result = mysqli_query(DPDatabase::get_connection(), $sql);
-        $row = mysqli_fetch_assoc($result);
-        if (!$row) {
-            $sql = "
-                INSERT INTO users
-                SET id = '$this->TEST_USERNAME',
-                    real_name = '$this->TEST_USERNAME',
-                    username = '$this->TEST_USERNAME',
-                    email = '$this->TEST_USERNAME@localhost'
-            ";
-            $result = mysqli_query(DPDatabase::get_connection(), $sql);
-            if (!$result) {
-                throw new Exception("Unable to create test user 1");
-            }
-
-            $sql = "
-                INSERT INTO users
-                SET id = '$this->TEST_USERNAME-2',
-                    real_name = '$this->TEST_USERNAME-2',
-                    username = '$this->TEST_USERNAME-2',
-                    email = '$this->TEST_USERNAME@localhost'
-            ";
-            $result = mysqli_query(DPDatabase::get_connection(), $sql);
-            if (!$result) {
-                throw new Exception("Unable to create test user 2");
-            }
-        } else {
-            mysqli_free_result($result);
-        }
+        create_test_user($this->TEST_USERNAME);
+        create_test_user("$this->TEST_USERNAME-2");
     }
 
     protected function tearDown(): void
     {
-        $sql = "
-            DELETE FROM users
-            WHERE id = '$this->TEST_USERNAME' or id = '$this->TEST_USERNAME-2'
-        ";
-        $result = mysqli_query(DPDatabase::get_connection(), $sql);
+        delete_test_user($this->TEST_USERNAME);
+        delete_test_user("$this->TEST_USERNAME-2");
     }
 
     public function testEmptyConstructor()
--- a/SETUP/tests/ProjectTest.php
+++ b/SETUP/tests/ProjectTest.php
@@ -2,8 +2,215 @@
 
 class ProjectTest extends PHPUnit\Framework\TestCase
 {
+    private $TEST_USERNAME_PM = "ProjectTestPM_php";
+    private $TEST_USERNAME = "ProjectTest_php";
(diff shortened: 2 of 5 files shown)


cpeel:
There is no easy way to test this new code since the validation isn't plugged in anywhere. For above-and-beyond testing, making changes directly in the DB that would make a project invalid and then doing a validation on it would also work. I've created a helper script that can be used on the TEST server against any project as:
```bash
$ ~cpeel/test_project_validate.php projectID4c49f39aa829c
array(0) {
}
```

If no errors are detected it will print an empty array. If errors are detected it will list all errors. For instance:
```bash
$ ~cpeel/test_project_validate.php projectID455833134db51
array(1) {
  [0]=>
  string(28) "Project Manager is required."
}
```

Script contents:
```php
#!/usr/bin/env php
<?php
$relPath = "/home/cpeel/public_html/c.branch/api-project-validation/pinc/";
include_once($relPath."Project.inc");

$projectid = $argv[1];
$project = new Project($projectid);
$errors = $project->validate();
var_dump($errors);
```

chrismiceli (approved):

SETUP/tests/UserTest.php
```diff
@@ -7,46 +7,14 @@ class UserTest extends PHPUnit\Framework\TestCase
...
-            ";
-            $result = mysqli_query(DPDatabase::get_connection(), $sql);
-            if (!$result) {
-                throw new Exception("Unable to create test user 2");
-            }
-        } else {
-            mysqli_free_result($result);
-        }
+        create_test_user($this->TEST_USERNAME);
+        create_test_user("$this->TEST_USERNAME-2");
```
srjfoo, line 11: Silly question: Why is one quoted and the other not? (Since it's that way both here and in the teardown, I assume it's deliberate.)

cpeel: Not a silly question! The "-2" is a string we are appending to the variable so it's in quotes. If it's clearer I can change it to:
```php
create_test_user($this->TEST_USERNAME . "-2");
```

srjfoo: I was staring at it after I sent the question, and wondered if that was what it was. I'm fine with it the way that it is -- it just threw me for a bit (remember, I'm still a beginner 😁 ). Thanks!

srjfoo (approved):
I have not tested every possible combination, but have tested a fairly large sample (thanks for the script!).

70ray (approved):

srjfoo:
I wound up doing a bit of extra testing, and noticed that there's no verification that a `postednum` is actually present if a project has been marked posted. Was that intentional? (Other things that I leave out when editing a project page that are rejected if wrong or missing on save are also rejected using the testing script.)

cpeel:
> I wound up doing a bit of extra testing, and noticed that there's no verification that a `postednum` is actually present if a project has been marked posted. Was that intentional? (Other things that I leave out when editing a project page that are rejected if wrong or missing on save are also rejected using the testing script.)

I had to go back and look why I did it this way. `editproject.php` only does the check when `posted=1` is passed in via the ProjectTransition detour. Indeed, at this point the project isn't even in a posted state -- `editproject.php` does that upon saving the detour edit -- so there was no "easy" logic to lift for the validate function. I'm not currently sure how to implement that check but I agree it needs to be in place before we enable the API. I'd like to punt that down the road a bit for now.

srjfoo:
> > I wound up doing a bit of extra testing, and noticed that there's no verification that a `postednum` is actually present if a project has been marked posted. Was that intentional? (Other things that I leave out when editing a project page that are rejected if wrong or missing on save are also rejected using the testing script.)
> 
> I had to go back and look why I did it this way. `editproject.php` only does the check when `posted=1` is passed in via the ProjectTransition detour. Indeed, at this point the project isn't even in a posted state -- `editproject.php` does that upon saving the detour edit -- so there was no "easy" logic to lift for the validate function. I'm not currently sure how to implement that check but I agree it needs to be in place before we enable the API. I'd like to punt that down the road a bit for now.

That's fine with me. And your explanation makes sense -- looking at the history of a project, the edit is shown before the transition to posted.

Thanks for the explanation!

srjfoo (approved):

PR #746, opened 2022-01-18
Update PHP-CS-Fixer and dev dependencies

cpeel:
Update the PHP-CS-Fixer version to the latest 2.x and all of the dev package dependencies to their latest version (2 commits).

Note that PHP-CS-Fixer 2.x is EOLed. The 3.x version ends up with some additional, reasonable formatting changes that I don't want to inflict upon us quite yet given how much stuff is in-flight. We can worry about that later.

Nothing to test as these are all dev-only changes but validation that it doesn't change any of the formatting:
```
cpeel@ip-172-31-12-95:~/u-cpeel-dp (update-phpcsfixer)$ PHP_CS_FIXER_FUTURE_MODE=1 ./vendor/bin/php-cs-fixer fix
You are running PHP CS Fixer v2, which is not maintained anymore. Please update to v3.
You may find an UPGRADE guide at https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.3.0/UPGRADE-v3.md .
If you need help while solving warnings, ask at https://gitter.im/PHP-CS-Fixer, we will help you!

Loaded config default from "/home/cpeel/u-cpeel-dp/.php-cs-fixer.dist.php".

Fixed all files in 115.324 seconds, 20.000 MB memory used
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ node_modules
 vendor
 .phpunit.result.cache
 .php_cs.cache
+.php-cs-fixer.cache
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -10,8 +10,8 @@ $finder = PhpCsFixer\Finder::create()
     ->name('*.php.example')
 ;
 
-return PhpCsFixer\Config::create()
-    ->setRules([
+$config = new PhpCsFixer\Config();
+return $config->setRules([
         '@PSR12' => true,
         '@PHP74Migration' => true,
         // PHP tags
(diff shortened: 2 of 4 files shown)


srjfoo:
> Nothing to test as these are all dev-only changes but validation that it doesn't change any of the formatting:

Not completely sure what you're recommending here -- I'll cheerfully do some additional validation, but I guess I need more explicit instructions.

cpeel:
> > Nothing to test as these are all dev-only changes but validation that it doesn't change any of the formatting:
> 
> Not completely sure what you're recommending here -- I'll cheerfully do some additional validation, but I guess I need more explicit instructions.

The command run was to illustrate that updating these dev dependencies doesn't change how our formatter works (no files were changed in the run). No additional testing is necessary.

chrismiceli (approved):

70ray (approved):

srjfoo (approved):

PR #749, opened 2022-01-22
Move project deletion into Project class

cpeel:
In prep for some unit tests, move the logic of deleting a project to the Project class. No functional changes.

### DIFF
--- a/pinc/Project.inc
+++ b/pinc/Project.inc
@@ -9,6 +9,7 @@ include_once($relPath.'MARCRecord.inc');
 include_once($relPath.'stages.inc'); // is_formatting_round()
 include_once($relPath.'CharSuites.inc');
 include_once($relPath.'special_colors.inc');
+include_once($relPath.'wordcheck_engine.inc'); // delete_project_wordcheck_events()
 
 class UTF8ConversionException extends Exception
 {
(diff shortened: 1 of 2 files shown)


srjfoo (approved):

pinc/Project.inc
```diff
@@ -258,6 +259,25 @@ class Project
         return $errors;
     }
 
+    public function delete()
+    {
+        exec("rm -rf $this->dir");
```
chrismiceli, line 264: do we need to worry about shell escape here? I guess using php to delete the directory is probably slow?
https://stackoverflow.com/questions/3349753/delete-directory-with-files-in-it

chrismiceli (approved):

pinc/Project.inc
cpeel: UGH. Given how powerful this command is, yes, yes we should. Thanks for pointing it out -- I'll add some better checks and escaping here.

cpeel: Added several more checks:
1. Validate `$this->projectID` to ensure that we aren't pointing to the top-level project directory (which would result in deleting *ALL* projects 😮).
2. Check the directory exists before trying to delete it.
3. Properly escape the directory argument as one final layer of protection (good idea Chris!).

srjfoo (approved):

70ray (approved):

PR #750, opened 2022-01-22
Some small edit project clean-up

cpeel:
In prep for a PR with some big changes to `editproject.php` to use the new `Project->validate()` function I wanted to get these two small commits in to make reviewing that one easier.

This:
* Removes the `get_default_character_suites()` function now that they are defined in `site_vars.php`
* Moves the generic `get_changed_fields_for_objects()` function into `misc.inc`.

Testing that recording of the fields changed for a project still works can be done in the [pre-project-save](https://www.pgdp.org/~cpeel/c.branch/pre-project-save/) sandbox.

### DIFF
--- a/pinc/misc.inc
+++ b/pinc/misc.inc
@@ -13,6 +13,25 @@ function array_get($arr, $key, $default)
     }
 }
 
+function get_changed_fields_for_objects($new, $old)
+// Return an array whose values are the names of the properties
+// whose values differ between the two objects $new and $old.
+{
+    $old_as_array = (array)$old;
+    $new_as_array = (array)$new;
+    // They should have the same set of keys, but just in case,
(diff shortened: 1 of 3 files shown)


pinc/misc.inc
```diff
@@ -13,6 +13,25 @@ function array_get($arr, $key, $default)
     }
 }
 
+function get_changed_fields_for_objects($new, $old)
```
chrismiceli, line 16: this feels ideal for a unit test if you are bored :D

chrismiceli (approved):

srjfoo (approved):

pinc/misc.inc
cpeel: Added some unit tests for this and `array_get()` while I was here!

PR #753, opened 2022-01-23
Set of small API updates

cpeel:
These are a set of small API updates, mostly fixes to docs and edge-cases, that I discovered working on the Project update API.

Notably:
* Return an error if the JSON passed into a request is invalid rather than continuing as if the request input was empty.
* Always return the right HTTP response code in the edgecase where code outputs text before we send the HTTP headers.
* Increase the size of the `api_key` field slightly so that it can contain a full version 4 UUID.
* Update the `curl` docs to not include the unnecessary `-g` param.
* On wordlist PUTs, return the newly-saved set of words rather than an empty array.

Testing:
```bash
# PUTing a new word list with invalid JSON
curl -i -X PUT "https://www.pgdp.org/~cpeel/c.branch/api-updates/api/index.php?url=v1/projects/projectID44de3936807f1/wordlists/good" \
    -H "Content-Type: application/json" \
    -H "X-API-KEY: $API_KEY" \
    -d '[ blearg ]'
HTTP/1.1 400 Bad Request
Date: Sun, 23 Jan 2022 21:15:31 GMT
Server: Apache
Upgrade: h2
Connection: Upgrade, close
Vary: Accept-Language,User-Agent
Access-Control-Allow-Headers: X-Api-Key, Content-Type
Access-Control-Allow-Methods: OPTIONS, GET, POST, PUT, PATCH
X-Rate-Limit-Limit: 3600
X-Rate-Limit-Remaining: 3599
X-Rate-Limit-Reset: 3600
Content-Length: 45
Content-Type: application/json

{
    "error": "Content was not valid JSON"
}


# PUTing a new word list with valid JSON
curl -i -X PUT "https://www.pgdp.org/~cpeel/c.branch/api-updates/api/index.php?url=v1/projects/projectID44de3936807f1/wordlists/good" \
    -H "Content-Type: application/json" \
    -H "X-API-KEY: $API_KEY" \
    -d '[ "word1", "word2" ]'

HTTP/1.1 200 OK
Date: Sun, 23 Jan 2022 21:16:10 GMT
Server: Apache
Upgrade: h2
Connection: Upgrade
Vary: Accept-Language,User-Agent
Access-Control-Allow-Headers: X-Api-Key, Content-Type
Access-Control-Allow-Methods: OPTIONS, GET, POST, PUT, PATCH
X-Rate-Limit-Limit: 3600
X-Rate-Limit-Remaining: 3597
X-Rate-Limit-Reset: 3561
Content-Length: 28
Content-Type: application/json

[
    "word1",
    "word2"
]
```

This is available in the [api-updates](https://www.pgdp.org/~cpeel/c.branch/api-updates/) sandbox but I think a code review with the above is sufficient.

### DIFF
--- a/SETUP/db_schema.sql
+++ b/SETUP/db_schema.sql
@@ -804,7 +804,7 @@ CREATE TABLE `users` (
   `u_profile` int(10) unsigned NOT NULL default '0',
   `u_intlang` varchar(25) default '',
   `u_privacy` tinyint(1) default '0',
-  `api_key` varchar(32) DEFAULT NULL,
+  `api_key` varchar(38) DEFAULT NULL,
   PRIMARY KEY  (`username`),
   UNIQUE KEY `api_key` (`api_key`),
   KEY `u_id` (`u_id`),
--- a/SETUP/upgrade/17/20220123_alter_users_table.php
+++ b/SETUP/upgrade/17/20220123_alter_users_table.php
@@ -1,0 +1,21 @@
+<?php
+$relPath = '../../../pinc/';
+include_once($relPath.'base.inc');
+
+header('Content-type: text/plain');
+
+// ------------------------------------------------------------
+
+echo "Increasing size of users.api_key column..\n";
+$sql = "
+    ALTER TABLE users
+        CHANGE COLUMN api_key api_key varchar(38) default NULL
+";
+
+echo "$sql\n";
+
+mysqli_query(DPDatabase::get_connection(), $sql) or die(mysqli_error(DPDatabase::get_connection()));
+
+// ------------------------------------------------------------
+
+echo "\nDone!\n";
--- a/api/USERS_GUIDE.md
+++ b/api/USERS_GUIDE.md
@@ -106,16 +106,16 @@ available endpoints and their parameters and return values.
 Search for all projects with genre `Other`:
 
 ```bash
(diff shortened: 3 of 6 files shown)


chrismiceli (approved):

srjfoo:
Should the words be being appended? That is, is it reasonable to assume this will only be used to **_create_** the good and bad word lists, and not to **_add to_** them? My initial testing was using your samples on existing good and bad word lists.

cpeel:
> Should the words be being appended? That is, is it reasonable to assume this will only be used to **_create_** the good and bad word lists, and not to **_add to_** them? My initial testing was using your samples on existing good and bad word lists.

The `wordlists` API is all-or-none and replaces the project's word lists with what is passed in. This allows PMs to both add and delete words from the word lists.

srjfoo:
> > Should the words be being appended? That is, is it reasonable to assume this will only be used to **_create_** the good and bad word lists, and not to **_add to_** them? My initial testing was using your samples on existing good and bad word lists.
> 
> The `wordlists` API is all-or-none and replaces the project's word lists with what is passed in. This allows PMs to both add and delete words from the word lists.

That makes sense -- so it's the PM's responsibility, if using this function, to first GET the old word list and include it (plus or minus added or deleted words) before updating.

Thanks!

srjfoo (approved):

70ray (approved):

PR #754, opened 2022-01-23
Move project create and update logic to Project class

cpeel:
This is the big PR that all of the prior ones have been leading up to, and despite those smaller ones prior this is still a big one.

The overarching goal is to move the code that creates and updates project info from the `editproject.php` page to the `Project` class so it can be used both by the UI and the API. Indeed, I have a branch that builds on top of this one that adds project create and update code to the API.

I've added unit tests which cover basic project create, update, and delete functionality but the original Edit Project page is rife with interesting corner cases based on who is doing the editing (PM or SA) and testing this is going to be challenging. I've done a fair amount of testing already but I'd like a lot of eyes on this. If I've done my job right, this should behave virtually identical to what's currently live on TEST and any discrepancy is notable. (I say virtually because my branch keeps the project's Dublin Core file up-to-date on every project edit which was not happening before.)

Interesting use cases:
* As a PM:
  * Create a project from LoC search results
  * Create a project from scratch
  * Clone a project owned by the PM
  * Clone a project owned by another PM
  * Edit a project
* As an SA:
  * Create a project from scratch
  * Clone a project
  * Edit a project

This is available in the [project-save](https://www.pgdp.org/~cpeel/c.branch/project-save/) sandbox.

## A note on `ProjectInfoHolder`
In the original design, `ProjectInfoHolder` holds all project info as the project is edited. In the new model, `ProjectInfoHolder` object has a proper `Project` object (`$this->project`) that holds all of the first-level project information and is used to validate and save the data. `ProjectInfoHolder` still manages meta-level information (is the project posted, original cloned project id) and second-level project information (character suites, etc). This is why many of the changes to that file are changing `$this->blah` to `$this->project->blah`.

One of the things this changes is the names of the changed fields stored in the `project_events` table on edit. The `ProjectInfoHolder` class was used prior to calculate the field names and those fields don't all exactly match up to the ones in the `Project` class. 

Of note:
```txt
# ProjectInfoHolder name  => Project name
difficulty_level          => difficulty
projectmanager            => username
```

This means that going forward the Project field names will be used. The history on the Project page will show the proper name for both old and new fields but the database will use different values. This also means that only changes to first-level project fields will be recorded. Most notably this does not include a project's character suites since they are second-level fields.

### DIFF
--- a/SETUP/tests/ProjectTest.php
+++ b/SETUP/tests/ProjectTest.php
@@ -15,6 +15,7 @@ class ProjectTest extends PHPUnit\Framework\TestCase
         "difficulty" => "average",
         // username and image_source are set in setUp() after creation
     ];
+    private $created_projectids = [];
 
     protected function setUp(): void
     {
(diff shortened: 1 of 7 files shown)


srjfoo:
> Interesting use cases:
> 
>     * As a PM:
>       
>       * Create a project from LoC search results
>       * Create a project from scratch
>       * Clone a project owned by the PM
>       * Clone a project owned by another PM
>       * Edit a project
> 
>     * As an SA:
>       
>       * Create a project from scratch
>       * Clone a project
>       * Edit a project

PMs cannot clone a project owned by another PM. PMs also cannot change the PPer directly if the project is **_in_** PP unless the PM is the PP. (The back door is that they can make it unavailable for PP, whereupon they can make it available for another PP.)

To make it even more complicated than your outline, the in between outline that I don't see mentioned: 

* As a PF:
  * Create a project from LoC search results or from scratch
  * Clone a project owned by them as a PM
  * Clone a project owned by another PM
  * Edit a project owned by them as PM
  * Edit other PMs' projects with the exceptions of
    * Changing the difficulty level to beginner
    * Changing PM assignments

SAs can do all of the above, plus changing PM assignments and changing the difficulty level to beginner. I believe that the BEGIN PMs are the only others who can set the difficulty level to beginner. So yes, a lot of special casing.

cpeel:
While there's still no rush, it would be great to have some eyes on this when folks have time.

srjfoo:
> While there's still no rush, it would be great to have some eyes on this when folks have time.

I've done some poking, and then got distracted -- sorry. It's definitely on my agenda to get back to!

70ray (approved):
I've tried cloning a project. Can't see any problems but as @cpeel says, it is big

srjfoo:
I'm not sure I've tested everything that needs to be tested, but I've tested a range of things ...

As a PM and PF:
* Create a project from LoC search results
* Create a project from scratch
* Clone a project owned by the PM
* Confirm that a PM can't clone a project assigned to a different PM, but a PF can.
* Edit a project (PM can edit own, PF can edit all)
* Confirmed that neither PMs nor PFs can change PM or set difficulty level to beginner (other than the BEGIN PM)

Squirrels can do all of the above, I did not explicitly duplicate the above as a squirrel, but just confirmed that a squirrel can change PMs and set difficulty level to beginner.

Let me know if there's something else!

srjfoo (approved):

PR #755, opened 2022-01-25
Restrict access to the shared postednums script

srjfoo:
The script is somewhat resource intensive, and only site admins are able to deal with any new issues as they occur (and probably should have been restricted to SAs all along).

Testable: [restrict-dup-posteds](https://www.pgdp.org/~srjfoo/c.branch/restrict-dup-posteds).


cpeel (approved):

lhamilton1:
I'm fine with this change

PR #761, opened 2022-01-29
Create project higher

70ray:
This creates the project in the get_available_page() function and passes it to the functions it calls. Also some other changes to parameters, call can_user_get_pages_in_project() in get_available_page(), and move the project continuity test into LPage.php.
This tidies up the logic.
There should be no functional differences. The only difference is that if project creation in get_available_page() throws a NonexistentProjectException it will not get caught (which is good). This is impossible in the normal way because it would get thrown earlier in proof.php. But if you go straight into proof_frame.php it would formerly get as far as the redirection page in proof_frame.php where the error would be thrown again while attempting to create the project again.
(This is in preparation for revising the checkout procedure but seemed like a good thing to do anyway).

Sandbox at: https://www.pgdp.org/~rp31/c.branch/create_project_higher


pinc/LPage.inc
```diff
@@ -8,37 +8,26 @@ include_once($relPath.'user_is.inc');
 include_once($relPath.'DPage.inc');
 include_once($relPath.'Project.inc');
 include_once($relPath.'page_tally.inc');
-include_once($relPath.'project_continuity.inc');
 include_once($relPath.'maybe_mail.inc');
 include_once($relPath.'daily_page_limit.inc'); // get_dpl_count_for_user_in_round
 
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 
 function get_available_page_array(
```
cpeel, line 16: Let's put all of the function args on the same line now that they'll all fit.

cpeel (approved):
Great chanage!

pinc/LPage.inc
70ray: Done this.

70ray:
Sandbox updated.

srjfoo (approved):

PR #764, opened 2022-01-31
change to using fetch for ajax instead of jquery function

70ray:
javascript native fetch uses the promise mechanism. This should make the code less incomprehensible than using two different methods for asynchronous operations.
There should not be any difference in function.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/api_fetch

### DIFF
--- a/SETUP/tests/jsTests/ajaxTests.js
+++ b/SETUP/tests/jsTests/ajaxTests.js
@@ -1,0 +1,60 @@
+/* global QUnit ajax */
+/* exported codeUrl */
+
+let codeUrl = "https://www.dummy.org";
+
+QUnit.module("Ajax test", function() {
+
+    QUnit.test("Return correct data", function (assert) {
+        function fetchPromise() {
+            const blob = new Blob([JSON.stringify("mydata")], {type : 'application/json'});
+            const init = {status: 200, headers: {'content-type': 'application/json'}};
+            return Promise.resolve(new Response(blob, init));
+        }
+
+        return ajax("GET", "myUrl", {}, {}, fetchPromise)
+            .then(function(data) {
+                assert.strictEqual(data, "mydata");
+            });
+    });
+
+    QUnit.test("Wrong type of data", function (assert) {
+        function fetchPromise() {
+            const blob = new Blob([JSON.stringify("mydata")], {type : 'application/json'});
+            const init = {status: 200, headers: {'content-type': 'text/html'}};
+            return Promise.resolve(new Response(blob, init));
+        }
+
+        return ajax("GET", "myUrl", {}, {}, fetchPromise)
+            .then(function() {}, function(data) {
+                assert.strictEqual(data, "Incorrect response type");
+            });
+    });
+
+    QUnit.test("Status 404", function (assert) {
+        function fetchPromise() {
+            const blob = new Blob([JSON.stringify({error: "not found"})], {type : 'application/json'});
+            const init = {status: 404, headers: {'content-type': 'application/json'}};
+            return Promise.resolve(new Response(blob, init));
+        }
+
+        return ajax("GET", "myUrl", {}, {}, fetchPromise)
+            .then(function() {}, function(data) {
+                assert.strictEqual(data, "not found");
+            });
+    });
+
+    QUnit.test("Unknown error", function (assert) {
+        function fetchPromise() {
+            const blob = new Blob([JSON.stringify("")], {type : 'application/json'});
+            const init = {status: 404, headers: {'content-type': 'application/json'}};
+            return Promise.resolve(new Response(blob, init));
+        }
+
+        return ajax("GET", "myUrl", {}, {}, fetchPromise)
+            .then(function() {}, function(data) {
+                assert.strictEqual(data, "Unknown error");
+            });
+    });
+
+});
--- a/SETUP/tests/jsTests/ajaxTests.js
+++ b/SETUP/tests/jsTests/ajaxTests.js
@@ -7,14 +7,14 @@ QUnit.module("Ajax test", function() {
 
     QUnit.test("Return correct data", function (assert) {
         function fetchPromise() {
-            const blob = new Blob([JSON.stringify("mydata")], {type : 'application/json'});
+            const blob = new Blob([JSON.stringify({key: "value"})], {type : 'application/json'});
             const init = {status: 200, headers: {'content-type': 'application/json'}};
             return Promise.resolve(new Response(blob, init));
         }
 
         return ajax("GET", "myUrl", {}, {}, fetchPromise)
             .then(function(data) {
-                assert.strictEqual(data, "mydata");
+                assert.deepEqual(data, {key: "value"});
             });
     });
 
@@ -57,4 +57,20 @@ QUnit.module("Ajax test", function() {
             });
     });
 
+    QUnit.test("Check sent data", function (assert) {
+        function fetchPromise(url, options) {
+            assert.strictEqual(url.href, "https://www.dummy.org/api/index.php?url=myUrl&querykey=queryvalue");
+            assert.strictEqual(options.method, "POST");
+            assert.strictEqual(options.headers['Content-Type'], 'application/json');
+            assert.strictEqual(options.headers["X-API-KEY"], "SESSION");
+            assert.strictEqual(options.headers.Accept, 'application/json');
+            assert.strictEqual(options.body, "{\"datakey\":\"datavalue\"}");
+            const blob = new Blob([JSON.stringify("mydata")], {type : 'application/json'});
+            const init = {status: 200, headers: {'content-type': 'application/json'}};
+            return Promise.resolve(new Response(blob, init));
+        }
+
+        return ajax("POST", "myUrl", {querykey: "queryvalue"}, {datakey: "datavalue"}, fetchPromise)
+            .then();
+    });
 });
--- a/scripts/api.js
+++ b/scripts/api.js
@@ -22,7 +22,11 @@ function ajax(method, apiUrl, queryParams = {}, data = {}) {
                 } else {
                     response.json()
                         .then(function(data) {
-                            reject(data.error);
+                            let message = data.error;
+                            if(!message) {
+                                message = "Unknown error";
+                            }
+                            reject(message);
                         });
                 }
             });
--- a/scripts/api.js
+++ b/scripts/api.js
@@ -24,15 +24,15 @@ function ajax(method, apiUrl, queryParams = {}, data = {}) {
                 } else if(response.ok) {
                     resolve(response.json());
                 } else {
-                    return response.json();
+                    response.json()
+                        .then(function(data) {
+                            let message = data.error;
+                            if(!message) {
+                                message = "Unknown error";
+                            }
+                            reject(message);
+                        });
                 }
-            })
-            .then(function(data) {
-                let message = data.error;
-                if(!message) {
-                    message = "Unknown error";
-                }
-                reject(message);
             })
             .catch(reject);
     });
--- a/scripts/api.js
+++ b/scripts/api.js
@@ -8,6 +8,7 @@ function ajax(method, apiUrl, queryParams = {}, data = {}) {
     method = method.toUpperCase();
     let options = {
         headers: {"X-API-KEY": "SESSION", 'Accept': 'application/json'},
+        credentials: "same-origin",
         method: method,
     };
     if(method !== "GET") {
--- a/scripts/api.js
+++ b/scripts/api.js
@@ -3,8 +3,7 @@
 
 function ajax(method, apiUrl, queryParams = {}, data = {}) {
     let url = new URL(`${codeUrl}/api/index.php`);
-    queryParams.url = apiUrl;
-    url.search = new URLSearchParams(queryParams);
+    url.search = new URLSearchParams(Object.assign({url: apiUrl}, queryParams));
     let upperCaseMethod = method.toUpperCase();
     let options = {
         headers: {"X-API-KEY": "SESSION", 'Accept': 'application/json'},
--- a/scripts/api.js
+++ b/scripts/api.js
@@ -2,8 +2,13 @@
 /* exported ajax */
 
 function ajax(method, apiUrl, queryParams = {}, data = {}, fetchPromise = fetch) {
-    let url = new URL(`${codeUrl}/api/index.php`);
-    url.search = new URLSearchParams(Object.assign({url: apiUrl}, queryParams));
+    let url = new URL("/api/index.php", codeUrl);
+    let searchParams = new URLSearchParams();
+    searchParams.append("url", apiUrl);
+    for (const key in queryParams) {
+        searchParams.append(key, queryParams[key]);
+    }
+    url.search = searchParams;
     let upperCaseMethod = method.toUpperCase();
     let options = {
         headers: {"X-API-KEY": "SESSION", 'Accept': 'application/json'},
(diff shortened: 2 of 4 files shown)


cpeel:
While they probably do, we should double-check that all of our supported browsers support these JS features: https://caniuse.com/?search=promise

scripts/api.js
```diff
@@ -1,17 +1,30 @@
...
+        options.body = JSON.stringify(data);
+    }
+    return new Promise(function(resolve, reject) {
+        fetch(url, options)
+            .then(function(response) {
+                if(response.ok) {
+                    resolve(response.json());
+                } else {
+                    response.json()
+                        .then(function(data) {
```
chrismiceli, line 24: do we want to catch in case response wasn't json?

scripts/api.js
```diff
@@ -1,17 +1,30 @@
...
     };
-    return settings;
+    if(method !== "GET") {
+        // POST or PUT
+        options.headers['Content-Type'] = 'application/json';
+        options.body = JSON.stringify(data);
+    }
+    return new Promise(function(resolve, reject) {
+        fetch(url, options)
+            .then(function(response) {
```
chrismiceli, line 19: not that it can happen often, but the fetch promise could need a catch if network interrupted or something so may want to catch the fetch promise too:
https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch

scripts/page_browse.js
```diff
@@ -336,16 +331,15 @@ function pageBrowse(params, storageKey, replaceUrl, mentorMode = false) {
...
+                    .then(function(rounds) {
                         rounds.forEach(function(round) {
                             let selected = (currentRound === round);
                             roundSelector.add(new Option(round, round, selected, selected));
                         });
                         resolve();
-                    })
-                    .fail(function(jqxhr) {
-                        showError(jqxhr);
+                    }, function(error) {
```
chrismiceli, line 341: would it be cleaner to have a .catch instead of a 2nd param to then? I feel it is more common this way.

scripts/page_browse.js
```diff
@@ -384,11 +378,10 @@ function pageBrowse(params, storageKey, replaceUrl, mentorMode = false) {
                     // by the shown (first) option
                     const round = roundSelector.value;
                     params.set("round_id", round);
-                    $.ajax(makeApiAjaxSettings("v1/projects/" + projectId + "/pages/" + imageFileName + "/pagerounds/" + round))
-                        .done(function(data) {
+                    ajax("GET", `v1/projects/${projectId}/pages/${imageFileName}/pagerounds/${round}`)
```
chrismiceli, line 381: does the image file name ever need encoding to handle characters not valid in url path params like '/' or something that may need encoding?

scripts/page_browse.js
```diff
@@ -336,16 +331,15 @@ function pageBrowse(params, storageKey, replaceUrl, mentorMode = false) {
                 resolve();
             } else {
                 roundSelector = document.createElement("select");
-                $.ajax(makeApiAjaxSettings("v1/projects/pagerounds"))
-                    .done(function(rounds) {
+                ajax("GET", "v1/projects/pagerounds")
+                    .then(function(rounds) {
```
chrismiceli, line 335: if you want, we could use async await, but probably drops support a bit more.

scripts/page_browse.js
cpeel: Currently our image names only contain numbers/letters + period + ext. I'm comfortable with using it as-is in the URL.

scripts/api.js
cpeel: We probably also need to handle the case where `error` isn't a key in the JSON too. All valid API requests should return one on a non-200 response code but it would be bad to fail on a failure.

scripts/api.js
```diff
@@ -1,17 +1,30 @@
...
+function ajax(method, apiUrl, queryParams = {}, data = {}) {
+    let url = new URL(`${codeUrl}/api/index.php`);
+    queryParams.url = apiUrl;
+    url.search = new URLSearchParams(queryParams);
+    let options = {
+        headers: {"X-API-KEY": "SESSION"},
+        method: method,
     };
-    return settings;
+    if(method !== "GET") {
```
chrismiceli, line 12: Might want to consider "get" lowercase too but not critical

chrismiceli:
I love the switch to fetch!

70ray:
> While they probably do, we should double-check that all of our supported browsers support these JS features: https://caniuse.com/?search=promise

Promise and fetch are ok for Chrome 54 Firefox 52 Microsoft Edge  Opera 41  Safari 11

scripts/page_browse.js
70ray: Yes async and await could simplify the ajax function as well. They are supported in firefox 52 and safari 11, not chrome 54 or opera 41.

scripts/api.js
70ray: Rather than use the method as a parameter, is there a case for making separate functions for get, put and post? Then any mistakes would be caught by linting and wouldn't be left until run-time. It would also be better in that "get" can't have a body (or data parameter).

scripts/api.js
70ray: Should we also look at the response status code? Since we are returning several different codes are there circumstances where we want to look at them and take different actions? WE could have a lower level function that returns the response and wrap it in a higher level function.

scripts/api.js
70ray: I put in a catch and tried issuing a request with the network disconnected but nothing ever happens. We could perhaps put in a timer to detect network disconnection.
The article suggests that a network error could result from misconfigured server. Can we simulate this to test it?

scripts/api.js
cpeel: Since this function is just for DP code -- and not for external systems using our code -- I don't think we need quite that level of abstraction. I do like doing a case-insensitive comparison for `get`/`GET`.

We should probably also be setting `options.headers['Accept'] = 'application/json';` for all request types as well.

scripts/page_browse.js
70ray: That seems reasonable. Done this.

scripts/api.js
chrismiceli: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch
Since 2018 fetch changed the default to include same-origin:
https://github.com/whatwg/fetch/pull/585

70ray: Wouldn't hurt to put it in anyway.

scripts/api.js
```diff
@@ -1,17 +1,40 @@
...
-        "url": codeUrl + "/api/index.php?url=" + apiUrl + queryParamString,
-        "dataType": "json",
-        "headers": {
-            "X-API-KEY": "SESSION"
-        }
+function ajax(method, apiUrl, queryParams = {}, data = {}) {
+    let url = new URL(`${codeUrl}/api/index.php`);
+    queryParams.url = apiUrl;
+    url.search = new URLSearchParams(queryParams);
+    method = method.toUpperCase();
```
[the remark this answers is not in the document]
70ray: It seems that the method should be uppercase but lowercase usually works. e.g. https://stackoverflow.com/questions/27423615/http-method-names-upper-or-lower-case
We could make a new variable to overcome your objection.

scripts/api.js
```diff
@@ -1,17 +1,40 @@
...
-    }
-    let settings = {
-        "url": codeUrl + "/api/index.php?url=" + apiUrl + queryParamString,
-        "dataType": "json",
-        "headers": {
-            "X-API-KEY": "SESSION"
-        }
+function ajax(method, apiUrl, queryParams = {}, data = {}) {
+    let url = new URL(`${codeUrl}/api/index.php`);
+    queryParams.url = apiUrl;
```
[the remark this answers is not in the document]
70ray: Another problem with my way is that the url parameters get in the "wrong" order: the query parameters are before the "resource locator". This doesn't matter but if we make a "real" restful url we would have change this anyway. I'll make it so that the "url" part is before the actual query parameters. (This is not noticeable in the page browser because there are never any query parameters)

scripts/api.js
```diff
@@ -1,17 +1,40 @@
...
+                if (!contentType || !contentType.includes('application/json')) {
+                    reject("Incorrect response type");
+                } else if(response.ok) {
+                    resolve(response.json());
+                } else {
+                    return response.json();
+                }
+            })
+            .then(function(data) {
+                let message = data.error;
```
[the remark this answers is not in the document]
70ray: Yes, that would be better. Thanks.

scripts/api.js
```diff
@@ -1,17 +1,40 @@
...
-    if(queryParams) {
-        queryParamString = "&" + $.param(queryParams);
-    }
-    let settings = {
-        "url": codeUrl + "/api/index.php?url=" + apiUrl + queryParamString,
-        "dataType": "json",
-        "headers": {
-            "X-API-KEY": "SESSION"
-        }
+function ajax(method, apiUrl, queryParams = {}, data = {}) {
```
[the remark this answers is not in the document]
70ray: Yes, I'll look at doing this.

chrismiceli:
Looks good to me now. I look forward to the unit tests though if you intend to do them.

70ray:
> Looks good to me now. I look forward to the unit tests though if you intend to do them.

I started working on this but ran into a problem with an interaction with another test, getting the message "validCharacterPattern is not defined". The ajax test works if the characterValidtion test and the file it uses are removed from the unit tests. I'll look into fixing this in another branch since it is not directly related to this MR.

70ray:
> Looks good to me now. I look forward to the unit tests though if you intend to do them.

I started working on this but ran into a problem with an interaction with another test, getting the message "validCharacterPattern is not defined". The ajax test works if the characterValidtion test and the file it uses are removed from the unit tests. I'll look into fixing this in another branch since it is not directly related to this MR.

70ray:
Added some tests and updated the sandbox.

srjfoo:
Browserstack testing:

High Sierra:
* Safari 11 looks okay
* Firefox 52-53 throw an "Invalid API path" exception. FF 54 works okay.
* Chrome 54,-60 throw an "Invalid API path" exception. Chrome 61, 62 okay
* Opera 41-47 throw the exception, Opera 48 seems to be okay.

Windows 7
* FF -- verified, same results as macOS High Sierra

Windows 10
*FF -- verified, same results as macOS High Sierra

For the initial testing, I verified scrolling through pages and resizing a bit, but even so, didn't do exhaustive functional testing; mostly verified that it looked reasonable (i.e. toolbars in place, could scroll pages, could resize window without losing vertical toolbar options.

After the more extensive testing on High Sierra, I mostly verified that the max versions that threw an exception on macOS also did so on Windows, and that the min version that didn't also matched. I also verified that the oldest Edge worked okay, and totally ignored IE 😁 .

I went looking for version information; I didn't look at opera, but found this for Chrome and FF:

Chrome 54: 2016-10-12
Chrome 56: 2017-01-25
Chrome 60: 2017-07-25
Chrome 61: 2017-09-05

Firefox 52: 2017-03-07
Firefox 54: 2017-06-13
Firefox 56: 2017-09-28

Opera 41: 2016-10-25
Opera 43: 2017-02-07
Opera 48: 2017-09-27

So, essentially, functionality for this was added in September of 2017 for all three major cross-platform browsers (Safari may be a major browser for Mac, iPhone and iPad users, but is not cross-platform).

cpeel:
Thanks Sharon!

To reframe what Sharon discovered, our [currently supported browsers](https://www.pgdp.net/wiki/DP_Official_Documentation:General/Supported_Browsers) are:
* Chrome 54
* Firefox 52
* Microsoft Edge
* Opera 41
* Safari 11

And this would change it to:
* Chrome 61 <-- changed
* Firefox 54 <-- changed
* Microsoft Edge
* Opera 48 <-- changed
* Safari 11

We should see what JS feature we are using that doesn't work. The "Invalid API path" error sounds like the URL isn't being constructed correctly. According to caniuse.org [template strings](https://caniuse.com/?search=template%20literals) are supported in all of those versions though which was my first guess.

srjfoo:
> Thanks Sharon!
> 
> To reframe what Sharon discovered, our [currently supported browsers](https://www.pgdp.net/wiki/DP_Official_Documentation:General/Supported_Browsers) are:
> 
>     * Chrome 54
> 
>     * Firefox 52
> 
>     * Microsoft Edge
> 
>     * Opera 41
> 
>     * Safari 11
> 
> 
> And this would change it to:
> 
>     * Chrome 61 <-- changed
> 
>     * Firefox 54 <-- changed
> 
>     * Microsoft Edge
> 
>     * Opera 48 <-- changed
> 
>     * Safari 11
> 
> 
> We should see what JS feature we are using that doesn't work. The "Invalid API path" error sounds like the URL isn't being constructed correctly. According to caniuse.org [template strings](https://caniuse.com/?search=template%20literals) are supported in all of those versions though which was my first guess.

Interesting, though, that browsers that were about 6 months newer than our currently supported set do work with the code as-is?

cpeel:
> Interesting, though, that browsers that were about 6 months newer than our currently supported set do work with the code as-is?

That doesn't surprise me. This code is using some JS features that we haven't used before. I can't find anything obvious at caniuse.com that jumps out as the problem though.

70ray:
> * Invalid API path

Looks like it is the URLSearchParams constructor.

70ray:
The URLSearchParams constructor could not accept a "record". Revised code and updated the sandbox.

srjfoo:
> The URLSearchParams constructor could not accept a "record" but it's possible to append key,value pairs one at a time since FF29, chrome49, opera36. safari10.1, Edge17. Revised code and updated the sandbox.

The previously erroring browser versions work with the new code. I didn't do as much testing, just mostly confirmed that the affected high and low end of the previously not working versions work now on one OS or another, which is hopefully good enough.

srjfoo (approved):

cpeel (approved):
Thanks Ray!

Given the prod deployment tomorrow morning and the code release aimed for the end of the week, I'm going to hold off merging this in until after we cut the release. I'd prefer this change to bake on test before rolling it out and a bit of bake time on prod before including it in a release.

SETUP/tests/jsTests/ajaxTests.js
```diff
@@ -0,0 +1,76 @@
...
+    QUnit.test("Return correct data", function (assert) {
+        function fetchPromise() {
+            const blob = new Blob([JSON.stringify({key: "value"})], {type : 'application/json'});
+            const init = {status: 200, headers: {'content-type': 'application/json'}};
+            return Promise.resolve(new Response(blob, init));
+        }
+
+        return ajax("GET", "myUrl", {}, {}, fetchPromise)
+            .then(function(data) {
+                assert.deepEqual(data, {key: "value"});
```
chrismiceli, line 17: It probably works, but probably a good idea to use the qunit async so the test runner waits for all promises to resolve before marking the test passed or not: https://api.qunitjs.com/assert/async/

70ray: I don't claim to have much understanding of Qunit. I was following the example at the bottom of this page: https://api.qunitjs.com/QUnit/test/#example-test-with-promise

70ray:
> > The URLSearchParams constructor could not accept a "record" but it's possible to append key,value pairs one at a time since FF29, chrome49, opera36. safari10.1, Edge17. Revised code and updated the sandbox.
> 
> The previously erroring browser versions work with the new code. I didn't do as much testing, just mostly confirmed that the affected high and low end of the previously not working versions work now on one OS or another, which is hopefully good enough.

Thanks Sharon, I should have read the doc. https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams more carefully instead of just looking at the top line of Browser compatibility.

PR #765, opened 2022-02-03
Local regex

70ray:
make validCharRegex when needed and use native javascript to make validCharRegex since unicode regex now in supported browsers. So don't need $(document).ready() in character_test.js which causes a problem in unit tests if another test uses promises.

### DIFF
--- a/scripts/character_test.js
+++ b/scripts/character_test.js
@@ -1,37 +1,22 @@
-/*global $ validCharacterPattern XRegExp */
+/*global validCharacterPattern XRegExp */
 /* exported testText */
 
-// regex unicode property escape is supported in Chrome and Safari (Feb 2020)
-// but not in Firefox or Edge. Use 3rd party http://xregexp.com/ instead
+// regex unicode property escape is supported in Chrome 64, Safari 11.1
+// Firefox 78, Edge 79, Opera 51. Use 3rd party http://xregexp.com/ instead
 
 // this matches any character: non-mark codepoint followed by 0 or more marks
 const charMatch = XRegExp("\\PM\\pM*", "Ag");
 
-// this regular expression (constructed below) matches individual good characters
-var validCharRegex;
-
-function makeValidCharRegex(characterPattern) {
-    // need the u flag for Astral plane characters.
-    return XRegExp(characterPattern, "Au");
-}
-
-$(function () {
-    // need to define this after the page has loaded so validCharacterPattern
-    // is available
-    validCharRegex = makeValidCharRegex(validCharacterPattern);
-});
-
-function testChar(character) {
-    return validCharRegex.test(character);
-}
-
 // return false if text contains any bad characters
 function testText(text) {
+    // this regular expression matches individual good characters
+    // unicode support in Chrome 50, Edge 12, Firefox 46, Safari 10, Opera 37
+    let validCharRegex = new RegExp(validCharacterPattern, "u");
     text = text.normalize("NFC");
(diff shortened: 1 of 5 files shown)


cpeel:
Please throw up a sandbox we can sanity check this in.

Thanks for the additional comments about what JS functions work in what browsers!

styles/themes/charcoal.css
```diff
@@ -146,7 +146,7 @@
   font-style: normal;
   font-weight: 700;
   font-display: swap;
-  src: url(https://fonts.gstatic.com/s/ebgaramond/v19/~ChoKC0VCIEdhcmFtb25kOgsI9NCduwcVAAAvRCAA.ttf) format('truetype');
+  src: url(https://fonts.gstatic.com/s/ebgaramond/v24/SlGDmQSNjdsmc35JDF1K5E55YMjF_7DPuGi-DPNUAw.ttf) format('truetype');
```
chrismiceli, line 149: Off topic of the PR, not sure we're GDPR compliant, but there is that recent news about leaking ip addresses to google violating data policies
https://termageddon.com/google-fonts-violates-gdpr/

scripts/character_test.js
```diff
@@ -1,37 +1,22 @@
...
+/*global validCharacterPattern XRegExp */
 /* exported testText */
 
-// regex unicode property escape is supported in Chrome and Safari (Feb 2020)
-// but not in Firefox or Edge. Use 3rd party http://xregexp.com/ instead
+// regex unicode property escape is supported in Chrome 64, Safari 11.1
+// Firefox 78, Edge 79, Opera 51. Use 3rd party http://xregexp.com/ instead
 
 // this matches any character: non-mark codepoint followed by 0 or more marks
 const charMatch = XRegExp("\\PM\\pM*", "Ag");
```
chrismiceli, line 8: are we aiming to get rid of all instances of XRegExp eventually, right?

chrismiceli (approved):

styles/themes/charcoal.css
cpeel: Yeah, saw that. pgdp.net is not GDPR compliant with cookies either.

srjfoo: I remember this issue coming up once before, but I don't remember where or when. I thought that the answer was that for \<reasons\>, it didn't apply to us, but I can't quote those reasons, and don't know whether the assessment might change over time.

srjfoo:
Regarding `tools/proofers/process_diacritcal_markup.js`, I know it's been like that since the file was created, but would it be a good idea to fix the misspelling? I know the code doesn't care, as long as the references in the code match the filename being referenced, but I went looking everywhere for references to `process_diacritical_markup.js` and of course couldn't find it, and it took a while to realize where I was going wrong.

scripts/character_test.js
70ray: Yes.

70ray:
> Regarding `tools/proofers/process_diacritcal_markup.js`, I know it's been like that since the file was created, but would it be a good idea to fix the misspelling?

Yes, we should do that.

70ray:
> Please throw up a sandbox we can sanity check this in.

Sorry, done this now:
Sandbox at: https://www.pgdp.org/~rp31/c.branch/local_regex

70ray:
Updated sandbox after last commit. This is is for quiz.

cpeel (approved):
This LGTM.

PR #767, opened 2022-02-05
make DPDatabase::affected_rows() function

70ray:
Make DPDatabase::affected_rows() function and use it in existing code before new use.
The changed files have other database functions that could be updated. Should I do this in this branch to avoid them being in a partially updated state?

Sandbox at https://www.pgdp.org/~rp31/c.branch/affected_rows_function

I don't know why making less is needed to avoid erroring.


cpeel:
> I don't know why making less is needed to avoid erroring.

The .less URLs point to a generic version that when compiled gets clarified to the actual version. It will be much better if we simply update this in a separate MR and be done with it. One moment.

cpeel:
Fixed and this PR rebased.

> The changed files have other database functions that could be updated. Should I do this in this branch to avoid them being in a partially updated state?

No, keeping this focused and what you have is good.

PR #770, opened 2022-02-07
Fix autorelease logic

cpeel:
There was a bug in `attempt_to_release()` when called from `automodify.php` where it was passed a `$project` array but was expecting a `$project` *object*. This caused `project_pre_release_check()` to return an error (because it, too, was expecting an object). This resulted in every evaluated project being flagged as bad, and an email sent to the PM.

This creates the necessary project object for `project_pre_release_check()` and use in the error case for the emails as well.

This is the code code is hotfixed on PROD.

### DIFF
--- a/tools/project_manager/autorelease.inc
+++ b/tools/project_manager/autorelease.inc
@@ -24,16 +24,16 @@ function attempt_to_release($round, $project, $queue_name)
 {
     $projectid = $project['projectid'];
 
-    $errors = project_pre_release_check($project, $round);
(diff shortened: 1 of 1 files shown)


srjfoo (approved):

70ray (approved):
Perhaps strongly-typed languages were not so bad :-)

chrismiceli (approved):

cpeel:
> Perhaps strongly-typed languages were not so bad :-)

It certainly would have helped in this case!!

PR #772, opened 2022-02-09
Fix date format for safari

chrismiceli:
sandbox: https://www.pgdp.org/~chrismiceli/c.branch/user_logon_graph_safari/stats/user_logon_graphs.php
image

### DIFF
--- a/pinc/graph_data.inc
+++ b/pinc/graph_data.inc
@@ -68,7 +68,7 @@ function user_logging_on($past, $preceding)
     switch ($past) {
       case 'year':
           $min_timestamp = time() - 366 * $seconds_per_day;
-          $date_format = '%Y-%c-%d';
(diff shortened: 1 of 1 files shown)


pinc/graph_data.inc
```diff
@@ -68,7 +68,7 @@ function user_logging_on($past, $preceding)
     switch ($past) {
       case 'year':
           $min_timestamp = time() - 366 * $seconds_per_day;
-          $date_format = '%Y-%c-%d';
```
chrismiceli, line 71: This ensures the month is 2 digits (02 instead of 2), which apparently Safari does not like when doing a new Date('2020-2-08') for example.

srjfoo (approved):
Looks good to me.

lhamilton1:
I approve

PR #776, opened 2022-02-12
Remove the incomplete Metadata code

cpeel:
The Metadata code was added in 2004 but has never been used at pgdp.net. Remove this unused and perpetually untested code.

Available in the [remove-metadata-support](https://www.pgdp.org/~cpeel/c.branch/remove-metadata-support) sandbox.


chrismiceli (approved):

pinc/DPage.inc
```diff
@@ -49,13 +49,7 @@ function project_allow_pages($projectid)
             state         VARCHAR(50) NOT NULL default '', INDEX(state),
             b_user        VARCHAR(25) NOT NULL default '',
             b_code        INT(1)      NOT NULL default '0',
-            orig_page_num VARCHAR(6)  NOT NULL default '',
-            metadata      SET(
```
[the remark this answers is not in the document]
cpeel: Per-project tables are only created via this code, they do not exist in the schema as the default schema does not include any projects.

70ray (approved):

pinc/project_edit.inc
```diff
@@ -6,26 +6,17 @@ include_once($relPath.'project_states.inc');
 
 function user_can_add_project_pages($projectid, $page_type = "normal")
 {
-    global $site_supports_metadata;
-
     $project = new Project($projectid);
     $state = $project->state;
 
     if ($page_type == "tp&v") { // TP&V = Title Page & Verso
```
srjfoo, line 12: As far as I can tell (and I admit that I may be wrong about this) `tp&v` and `tpnv` are both specific to the metadata code. There is a whole section of [`do_early_uploads`](https://github.com/DistributedProofreaders/dproofreaders/blob/443817fcb4c03afaf5aea273f412325588f4d8ef/project.php#L877) in `project.php` that deals only with uploading title page and verso that, as far as I can recall, I've never seen displayed in the project load form when loading files into a project (but I'm not smart enough to figure out why it's not displaying).

There's also a bunch of code in `add_files.php` that deals with loading TP&V. As part of that section of code, the project state is updated to `project_new_waiting_app`. (The bits here, and in `add_files.php` and `project.php` are the three places I've found that refer to loading TP&V; both `tpnv` and `tp&v` are used.)

cpeel: Yup, indeed all the of `tp&v` code needs to come out too, you are correct. Thanks for catching that.

srjfoo: I finally remembered to create a test project and everything went as expected.

srjfoo (approved):

PR #777, opened 2022-02-14
revise checkout procedure

70ray:
This changes any available page to "checked out to the user" in one operation so there can't be any database conflicts. If there are no more pages affected_rows will be 0. Then finds it as the newest page checked out

Sandbox at: https://www.pgdp.org/~rp31/c.branch/robust_checkout

### DIFF
--- a/pinc/DPage.inc
+++ b/pinc/DPage.inc
@@ -253,24 +253,8 @@ function Page_modifyText($projectid, $image, $page_text, $text_column, $user)
 
 function Page_checkout($projectid, $image, $round, $user)
 {
-    _Page_require($projectid, $image,
-        [$round->page_avail_state],
-        null, null,
-        'checkout');
-
-    $timestamp = time();
-    $setters = join(", ", [
-        set_col_str("state", $round->page_out_state),
-        set_col_num($round->time_column_name, $timestamp),
-        set_col_str($round->user_column_name, $user),
-    ]);
-    _Page_UPDATE($projectid, $image, $setters);
-
     _log_page_event($projectid, $image, 'checkout', $user, $round);
-
     _project_adjust_n_available_pages($projectid, -1);
-
-    return $round->page_out_state;
 }
 
 // -----------------------------------------------------------------------------
--- a/pinc/LPage.inc
+++ b/pinc/LPage.inc
@@ -103,22 +103,27 @@ function get_available_page_array($project, $round, $pguser, $preceding_proofer_
         throw new ValueError("Bad value for 'preceding_proofer_restriction': '$preceding_proofer_restriction'.");
     }
 
-    // Find page to be proofread.
-    // (It appears that a simultaneous query from someone else would return the same page.)
-    $dbQuery = "
-        SELECT image, state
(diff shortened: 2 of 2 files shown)


70ray:
> I don't think this is the right solution to this problem. This moves what is currently page data-level logic into the page logic-level logic. `Page_checkout()` no longer actually checks out a page.
> 
> The "find me a page" and "checkout the page to me" happen in a single function call in a single session in `get_available_page()` so the odds of someone getting "in the middle" of those two non-atomic activities is small. But because it's a small window we can guarantee those two things are atomic if we take a table lock.
> 
> Consider creating two new functions in the `Project` class:
> 
> ```
> lock_table(); // will take a write-level table lock
> unlock_table();
> ```
> 
> Then in `LPage.inc`:
> 
> ```diff
> diff --git a/pinc/LPage.inc b/pinc/LPage.inc
> index 090f1092b..14460e8ee 100644
> --- a/pinc/LPage.inc
> +++ b/pinc/LPage.inc
> @@ -142,6 +142,9 @@ function get_available_page($projectid, $proj_state, $pguser, &$err)
>          return null;
>      }
> 
> +    // Lock project table to make checkouts atomic
> +    $project->lock_table();
> +
>      // If the global setting is to not serve the immediately preceding page,
>      // first see if we can serve them a page they haven't at all seen before,
>      // then fall back to not the immediately preceding page.
> @@ -156,6 +159,7 @@ function get_available_page($projectid, $proj_state, $pguser, &$err)
>      }
>      if (!$page_array) {
>          $err = _("No more files available for proofreading for this round of the project.");
> +        $project->unlock_table();
>          return null;
>      }
> 
> @@ -164,6 +168,7 @@ function get_available_page($projectid, $proj_state, $pguser, &$err)
>      $lpage = new LPage($projectid, $imagefile, $state, 0);
> 
>      $lpage->checkout($pguser);
> +    $project->unlock_table();
> 
>      $err = null;
>      return $lpage;
> ```

That seems a better way to do it, but wouldn't you need a read lock rather than a write lock to stop another user from finding the same page before  the first user had checked it out?

70ray:
I haven't been able to get a locking version to work. The lock seems to work but the sql query fails with syntax error.
The mysql manual says "Normally, you do not need to lock tables, because all single UPDATE statements are atomic;". Also it says that "You cannot refer to a locked table multiple times in a single query using the same name. Use aliases instead, and obtain a separate lock for the table and each alias:" There could  be 3? such references in this query if the restriction is complicated so this would add to the complexity if we could get it to work.
Perhaps we could overcome the objection to the 'update' method by removing the LPage and DPage checkout functions altogether and including the _log_page_event() and _project_adjust_n_available_pages() calls in get_available_page().
Alternatively not pursue this since as @cpeel says it is unlkely to cause any problems.

srjfoo (commented):
> Alternatively not pursue this since as @cpeel says it is unlkely to cause any problems.

This seems like it might be the right approach for now, since it's not causing significant problems.

cpeel:
I don't think this is a problem we need to solve right now.

PR #778, opened 2022-02-14
deframeify standard proof frame

70ray:
This uses a splitter for the standard proof frame. There should be no difference otherwise. H/V split and split ratio are controlled by settings as before. Some simplifications have been possible in wordcheck interface because the whole frame is now loaded instead of just the text frame.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/standard_split_3


70ray:
Forgot about the quizzes, but they seem to be working ok. I'll look at the code.

tools/proofers/proof_frame_std.inc
```diff
@@ -1,37 +1,102 @@
...
+
+    echo_std_view();
+    echo_text_area($ppage);
+    echo "</div>\n";
+    echo "</div>\n";
+}
+
+function echo_text_area($ppage)
+{
+    ?>
```
cpeel, line 52: Let's make this all PHP code and not switch back and forth.

cpeel (commented):
Removing frames is excellent -- would be lovely to not have the PIs using them.

Is this a stepping stone to further merging the two interfaces together code-wise? I would be more interested in the change if this was leading to removing the control frame and reducing the code complexity (fewer top-level files outputting the HTML -- really there should be just `proof.php` relying on some `*.inc` files). This might be best going into a separate short-lived branch in that case.

I also question if this helps us to our new proofreading interface goals.

70ray:
> Removing frames is excellent -- would be lovely to not have the PIs using them.
> 
> Is this a stepping stone to further merging the two interfaces together code-wise? I would be more interested in the change if this was leading to removing the control frame and reducing the code complexity (fewer top-level files outputting the HTML -- really there should be just `proof.php` relying on some `*.inc` files). This might be best going into a separate short-lived branch in that case.
> 
> I also question if this helps us to our new proofreading interface goals.

Yes, this is the first step towards merging the two interfaces, so it could go into another branch.

This version retains the functionality in both standard and enhanced interfaces so it could be used for real if we wanted to.
If we develop in a separate branch then we could not bother with the enhanced inerface which would perhaps make life easier.

The next step I would propose would be to make this new standard interface look more like the page browser, with a text widget with text controls and the other buttons in a full width bar. The missing buttons (revert to original and revert to last save could also be added at this stage). The horizontal/vertical switch could be made on the client side as in page browser. Wordcheck and preview could also be coerced into this form with their control buttons in the same bar. It would be possible to have an icon to switch the bar position top or bottom.

It would then be possible to have either buttons or icons in the bar. With icons this would then be sufficiently similar to the enhanced interface that we could call it a new enhanced interface and abandon the old one.

I would suggest making the control frame into a flex box rather than a splitter so it would automatically size itself to fit the contents. I suppose some people would say that they liked to have the resizeable frame so they could maximise the image and text area when they weren't using the control frame. In that case we could make it possible to hide the control frame in a similar way that we did with the character pickers.
The down-side of deframing the control frame is that it would get reloaded whenever the image and text were reloaded but this would not be a big deal with broadband.
At some stage it would also be possible to change to using api functions which would mean only the text or image need get reloaded and not the control frame.

srjfoo:
My preference would be for a separate working branch. I don't think we should be considering incremental updates to either interface on PROD at this point.

> This uses a splitter for the standard proof frame. There should be no difference otherwise.

I did a small amount of testing last night, but not exhaustive. My main comment at this point is that it is noticeably different if you're using the Charcoal theme:
image


> I would suggest making the control frame into a flex box rather than a splitter so it would automatically size itself to fit the contents. I suppose some people would say that they liked to have the resizeable frame so they could maximise the image and text area when they weren't using the control frame. In that case we could make it possible to hide the control frame in a similar way that we did with the character pickers.

I really don't like this idea. I'm one of those who prefers to be able to manually adjust the size of image and text parts of the proofreading interface. Particularly in horizontal (top-bottom) orientation. Consider that when a proofer wishes to limit the number of lines visible in the textarea, they need direct control over how tall the textarea is. There truly are times when it's not appropriate to try to force the container to be resized to fit the contents.

70ray:
> > I would suggest making the control frame into a flex box rather than a splitter so it would automatically size itself to fit the contents. I suppose some people would say that they liked to have the resizeable frame so they could maximise the image and text area when they weren't using the control frame. In that case we could make it possible to hide the control frame in a similar way that we did with the character pickers.
> 
> I really don't like this idea. I'm one of those who prefers to be able to manually adjust the size of image and text parts of the proofreading interface. Particularly in horizontal (top-bottom) orientation. Consider that when a proofer wishes to limit the number of lines visible in the textarea, they need direct control over how tall the textarea is. There truly are times when it's not appropriate to try to force the container to be resized to fit the contents.

Oops, we are at completely cross-purposes, my fault. By control frame I meant the toolbox at the bottom.

srjfoo:
> Oops, we are at completely cross-purposes, my fault. By control frame I meant the toolbox at the bottom.

Ah, okay. And I totally misunderstood, because I was thinking of the textarea. I think the toolbox needs a complete redesign, anyway. I don't think we want to be locked into it containing what it does now.

tools/proofers/proof_frame_std.inc
70ray: Good idea. I'll do that.

70ray:
I've made the text area style stay white and updated the sandbox.

srjfoo:
> I've made the text area style stay white and updated the sandbox.

image

WordCheck interface. The one on the left is your sandbox.

srjfoo:
image

Format Preview. The tool area is not fixed to the bottom of the text pane so that it stays in place when the text scrolls in your sandbox (left) as it is on the main TEST server (right).

tools/proofers/proof_frame_std.inc
```diff
@@ -1,37 +1,95 @@
...
+
+    $project = new Project($ppage->projectid());
+    $valid_character_pattern = build_character_regex_filter($project->get_valid_codepoints(), "js");
+    $switch_confirm = json_encode(_('Are you sure you want to save the current page and change layout?'));
+    $revert_confirm = json_encode(_("Are you sure you want to save the current page and revert to the original text for this round?"));
+
+    $header_args = [
+        "js_files" => [
+            "$code_url/scripts/splitControl.js",
+            "proof_std.js",
```
chrismiceli, line 19: no prefix like the others?

tools/proofers/proof_frame_std.inc
```diff
@@ -1,37 +1,95 @@
...
+            "$code_url/tools/proofers/proof_image.js",
+        ],
+        "css_files" => ["$code_url/styles/preview.css"],
+        "js_data" => get_preview_js_vars() . get_preview_font_data_js()
+        . get_control_bar_texts() . get_image_data($ppage) . get_std_view_data()
+        . "function ldAll() {
+            top.initializeStuff(0);
+        }
+        var validCharacterPattern = '$valid_character_pattern';
+        var switchConfirm = '$switch_confirm';
```
chrismiceli, line 36: I know it is old code, but it'd be nice if we could avoid globals and pass it to proof_validate.js somehow.

tools/proofers/proof_frame_std.inc
```diff
@@ -1,37 +1,95 @@
...
+
+    $ppage->echo_info();
+
+    echo "&nbsp;";
+
+    $comments_url = $ppage->url_for_project_comments(true);
+    echo _("View").": <a href=\"$comments_url\" title=\"". attr_safe(_("View Project Comments in New Window"))."\" target=\"viewcomments\">"._("Project Comments")."</a> ";
+
+    $image_url = $ppage->url_for_display_image(true);
+    echo "| <a href=\"$image_url\" title=\"". attr_safe(_("View Image in New Window"))."\" target=\"lg_image\">"._("Image")."</a> ";
```
chrismiceli, line 90: we should probably html_safe the _("Image") part too in the same spirit as attr_safe, though probably not ever an issue. Same for project comments on 87.

tools/proofers/spellcheck.inc
```diff
@@ -218,10 +220,12 @@ if ($user->profile->i_type == 1) {
 </form>
 </div>
 <?php
-
+if ($user->profile->i_type != 1) {
```
chrismiceli, line 223: we don't have constants for the i_type values we could use? A genuine question, but it'd be nice if we could have some define we reference.

tools/proofers/spellcheck.inc
```diff
@@ -230,7 +234,7 @@ function get_page_js($interface_type_init_value)
 
     return <<<PAGE_JS
                 function ldAll() {
-                    top.initializeStuff($interface_type_init_value);
+                    top.initializeStuff(2);
```
chrismiceli, line 237: Same here, is 2 an i_type? A define would be nice.

70ray:
I've fixed the wordcheck style. The format preview is fixable but it might be easiest to ignore this and go to the next stage.

tools/proofers/proof_frame_std.inc
70ray: This is inconsistent. In some cases there is prefix even though it is in the same directory. Probably best to remove them in that case.

cpeel:
I've created a `pi-redesign` branch and retargeted this PR into it.

tools/proofers/spellcheck.inc
70ray: If we are not planning to deploy this but only use it as a step along the way to a new PI this stuff will all go away.

tools/proofers/proof_frame_std.inc
70ray: It will get done via the API eventually.

tools/proofers/proof_frame_std.inc
70ray: The ppage function puts escaped ampersands in the url. I don't know if anything els eis needed.

srjfoo:
> I've fixed the wordcheck style. The format preview is now fixed but not in a good way. This will get changed in the next stage anyway.

Did you update the sandbox? I'm still seeing the white background for WC and the problem in Preview.

Honestly, though, I'm not sure that it matters at this point, because I do not envision the redesigned PI looking exactly like either of the current PIs. If, for instance, we're going to provide proofers with the option of customizing their proofreading interface's color scheme, shouldn't that be built in from the beginning?

If we're not, should we at least give a choice between light and dark themes, but, again, shouldn't that decision be made at the beginning, and be built in from the beginning?

70ray:
> > I've fixed the wordcheck style. The format preview is now fixed but not in a good way. This will get changed in the next stage anyway.
> 
> Did you update the sandbox? I'm still seeing the white background for WC and the problem in Preview.
> 

maybe not

> Honestly, though, I'm not sure that it matters at this point, because I do not envision the redesigned PI looking exactly like either of the current PIs. If, for instance, we're going to provide proofers with the option of customizing their proofreading interface's color scheme, shouldn't that be built in from the beginning?
> 
> If we're not, should we at least give a choice between light and dark themes, but, again, shouldn't that decision be made at the beginning, and be built in from the beginning?

People work in different ways. Some people plan things out beforehand but I find if I try to do that I always have to change the plans when I see what the real world issues are so I tend to start in the middle and work outwards.

70ray:
> > > If we're not, should we at least give a choice between light and dark themes, but, again, shouldn't that decision be made at the beginning, and be built in from the beginning?
> > 
> > 
> > People work in different ways. Some people plan things out beforehand but I find if I try to do that I always have to change the plans when I see what the real world issues are so I tend to start in the middle and work outwards.
> 
> I can understand that -- though I still think it's a good idea to figure out the non-negotiables ahead of time.

Yes, I suppose we should decide whether to make the colour scheme fully configurable. (It would be possible to use the same sort of interface we are using to configure the format preview). Or whether to have light and dark themes.

Having done some more work on this, I think it could be better to abandon this approach and go straight into an API version.

cpeel:
Please come create and get agreement on a design for a new proofreading interface before we start making changes.

PR #779, opened 2022-02-15
if testing, show database error to user

70ray:


cpeel:
We should not do this, even on the testing server. We put this in place because there are ways for hackers to cause MySQL to leak database information in error messages. I don't want to make this possible even on the test server. You're welcome to tail the `php_errors` log on the test server to get this information.

PR #781, opened 2022-02-19
Api proof round

70ray:
This has a few parts of a proofreading interface using API. No wordcheck, preview or toolbox. Some missing links. Several files are not connected. Most strings not translated.
The PI url uses 'round' instead of project_state.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/api_proof_round


cpeel:
Please break this into 2 separate PRs: one with the API changes we can test in isolation then a second that builds on top of that. The API changes can get into `master` since they are stand-alone building blocks.

cpeel:
Closing this until we have a finalized API. We can't build an interface without one.

PR #782, opened 2022-02-20
add proofing api functions

70ray:
Not sure how well it fits into the restful mould.

I haven't been able to get the Swagger editor to work into my sandbox. But tested substantially the same thing in the api_proof_round branch.

### DIFF
--- a/api/dp-openapi.yaml
+++ b/api/dp-openapi.yaml
@@ -502,6 +502,127 @@ paths:
         default:
           $ref: '#/components/responses/UnexpectedError'
 
+  /projects/{projectid}/codepoints:
+    get:
+      tags:
+      - project
+      description: Returns a string representing valid codepoints for the project
+      parameters:
+      - name: projectid
+        in: path
+        description: ID of project
+        required: true
+        schema:
+          type: string
+      responses:
+        200:
+          description: Codepoints
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/codepoints'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '429':
+          $ref: '#/components/responses/RateLimitExceeded'
+        default:
+          $ref: '#/components/responses/UnexpectedError'
+
+  /projects/{projectid}/rounds/{roundid}:
+    post:
+      tags:
+      - project
+      description: Checks out a page to the user
+      parameters:
+      - name: projectid
+        in: path
+        description: ID of project
+        required: true
+        schema:
+          type: string
+      - name: roundid
+        in: path
+        description: round ID
+        required: true
+        schema:
+          type: string
+      responses:
+        200:
+          description: pagename and pagestate
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/checked_page'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '429':
+          $ref: '#/components/responses/RateLimitExceeded'
+        default:
+          $ref: '#/components/responses/UnexpectedError'
+
+  /projects/{projectid}/pages/{pagename}/rounds/{roundid}:
+    put:
+      tags:
+      - project
+      description: Operates on a page checked out to a user
+      parameters:
+      - name: projectid
+        in: path
+        description: ID of project
+        required: true
+        schema:
+          type: string
+      - name: pagename
+        in: path
+        description: Name of page
+        required: true
+        schema:
+          type: string
+      - name: roundid
+        in: path
+        description: round ID
+        required: true
+        schema:
+          type: string
+      - name: pagestate
+        in: query
+        description: expected page state
+        required: true
+        schema:
+          type: string
+      - name: action
+        in: query
+        description: action to be performed
+        required: true
+        schema:
+          type: string
+          enum: [return, save_temp, save_done, open, revert]
+      requestBody:
+        description: json encoded string to store as page text
+        content:
+          application/json:
+            schema:
+              type: string
+      responses:
+        200:
+          description: pagetext, image_url, pagestate, 'can restore to last save'
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/proof_page_data'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '429':
+          $ref: '#/components/responses/RateLimitExceeded'
+        default:
+          $ref: '#/components/responses/UnexpectedError'
+
   /stats/site:
     get:
       tags:
@@ -657,6 +778,35 @@ components:
           items:
             $ref: '#/components/schemas/project_page_round_detail'
 
+    proof_page_data:
+      required:
+      - text
+      - image_url
+      - pagestate
+      - canRestore
+      type: object
+      properties:
+        text:
+          type: string
+        image_url:
+          type: string
+          format: url
+        pagestate:
+          type: string
+        canRestore:
+          type: boolean
+
+    checked_page:
+      required:
+      - pagename
+      - pagestate
+      type: object
+      properties:
+        pagename:
+          type: string
+        pagestate:
+          type: string
+
     project_page_round_detail:
       type: object
       required:
@@ -710,6 +860,14 @@ components:
             type: string
         enabled:
           type: boolean
+
+    codepoints:
+      required:
+      - codepoints
+      type: object
+      properties:
+        codepoints:
+          type: string
 
     image_source:
       required:
--- a/api/v1.inc
+++ b/api/v1.inc
@@ -30,7 +30,11 @@ $router->add_route("GET", "v1/projects/specialdays", "api_v1_projects_specialday
 $router->add_route("GET", "v1/projects/imagesources", "api_v1_projects_imagesources");
 $router->add_route("GET", "v1/projects/:projectid/pages", "api_v1_project_pages");
 $router->add_route("GET", "v1/projects/:projectid/pagedetails", "api_v1_project_pagedetails");
-$router->add_route("GET", "v1/projects/:projectid/pages/:pagename/pagerounds/:pageroundid", "api_v1_project_page_round");
+$router->add_route("GET", "v1/projects/:projectid/pages/:pagename/pagerounds/:pageroundid", "api_v1_project_page_pageround");
+$router->add_route("GET", "v1/projects/:projectid/codepoints", "api_v1_project_codepoints");
+
+$router->add_route("POST", "v1/projects/:projectid/rounds/:roundid", "api_v1_project_round");
+$router->add_route("PUT", "v1/projects/:projectid/pages/:pagename/rounds/:roundid", "api_v1_project_page_round");
 
 $router->add_route("GET", "v1/stats/site", "api_v1_stats_site");
 $router->add_route("GET", "v1/stats/site/rounds", "api_v1_stats_site_rounds");
--- a/api/v1_projects.inc
+++ b/api/v1_projects.inc
@@ -4,6 +4,8 @@ include_once($relPath.'wordcheck_engine.inc');
 include_once($relPath.'ProjectSearchForm.inc');
 include_once($relPath.'page_table.inc');
 include_once($relPath.'special_colors.inc');
+include_once($relPath.'CharSuites.inc'); // get_valid_codepoints()
+include_once($relPath.'LPage.inc'); // get_available_page()
 include_once("exceptions.inc");
 
 // DP API v1 -- Projects
@@ -225,9 +227,8 @@ function api_v1_project_pagedetails($method, $data, $query_params)
 }
 
 //---------------------------------------------------------------------------
-// projects/:projectid/pages/:pagename/pagerounds/:pageroundid
-
-function api_v1_project_page_round($method, $data, $query_params)
+// GET projects/:projectid/pages/:pagename/pagerounds/:pageroundid
+function api_v1_project_page_pageround($method, $data, $query_params)
 {
     if ($data[":pageroundid"] == "OCR") {
         $text_column = "master_text";
@@ -269,6 +270,94 @@ function render_project_page_json($row)
         "text" => $row["text"],
         "state" => $row["state"],
     ];
+}
+
+// POST v1/projects/:projectid/rounds/:roundid
+// checkout a new page to proofread
+function api_v1_project_round($method, $data, $query_params)
+{
+    global $pguser;
+
+    $project = $data[":projectid"];
+    $round = $data[":roundid"];
+    check_project_user($project, $round->id);
+
+    try {
+        $lpage = get_available_page($project->projectid, $project->state, $pguser, $err);
+    } catch (Exception $exception) {
+        throw new BadRequest($exception->getMessage());
+    }
+
+    if (is_null($lpage)) {
+        throw new BadRequest($err);
+    }
+    return ["pagename" => $lpage->imagefile, "pagestate" => $lpage->page_state];
+}
+
+// PUT projects/:projectid/pages/:pagename/rounds/:roundid
+// the PUTs all can change the page text or page state.
+// they are idempotent in the sense that if the identical request is issued
+// again, either the the state and text will not be changed or the request will
+// be rejected because the page state has changed so there will be no change
+// to the server. Logging side-effects excluded.
+function api_v1_project_page_round($method, $data, $query_params)
+{
+    global $pguser;
+
+    $project = $data[":projectid"];
+    $round = $data[":roundid"];
+    check_project_user($project, $round->id);
+
+    try {
+        $lpage = get_indicated_LPage($project->projectid, $project->state, $data[":pagename"], $query_params["pagestate"], 0);
+
+        // can throw InvalidArgumentException
+        $action = get_enumerated_param($query_params, "action", null, ["return", "save_temp", "save_done", "open", "revert"]);
+
+        switch ($action) {
+        case "return":
+            $lpage->returnToRound($pguser);
+            return [];
+            break;
+        case "save_temp":
+            $lpage->saveAsInProgress(api_get_request_body(), $pguser);
+            return render_page_data($lpage);
+            break;
+        case "save_done":
+            [$page_saved, $limit_reached] = $lpage->attemptSaveAsDone(api_get_request_body(), $pguser);
+            return ["pageSaved" => $page_saved, "limitReached" => $limit_reached];
+        case "open":
+            $lpage->resume_saved_page($pguser);
+            return render_page_data($lpage);
+            break;
+        case "revert":
+            $lpage->saveAsInProgress(api_get_request_body(), $pguser);
+            $lpage->revertToOriginal();
+            return render_page_data($lpage);
+            break;
+        }
+    } catch (Exception $exception) {
+        throw new BadRequest($exception->getMessage());
+    }
+}
+
+function render_page_data($lpage)
+{
+    global $projects_url;
+    return [
+        "text" => $lpage->get_text(),
+        "imageUrl" => "$projects_url/{$lpage->projectid}/{$lpage->imagefile}",
+        "pagestate" => $lpage->page_state,
+        "canRestore" => $lpage->can_be_reverted_to_last_save(),
+    ];
+}
+
+function check_project_user($project, $round_id)
+{
+    [$code, $msg] = $project->can_be_proofed_by_current_user($round_id);
+    if ($code != $project->CBP_OKAY) {
+        throw new BadRequest($msg);
+    }
 }
 
 //---------------------------------------------------------------------------
@@ -403,4 +492,11 @@ function api_v1_projects_imagesources($method, $data, $query_params)
     }
 
     return $return_data;
+}
+
+// GET, v1/projects/:projectid/codepoints
+function api_v1_project_codepoints($method, $data, $query_params)
+{
+    $project = $data[":projectid"];
+    return ["codepoints" => $project->get_valid_codepoints()];
 }
--- a/pinc/Project.inc
+++ b/pinc/Project.inc
@@ -722,8 +722,9 @@ class Project
     public $CBP_REQUESTOR_NOT_LOGGED_IN = 3;
     public $CBP_USER_NOT_QUALIFIED_FOR_ROUND = 4;
     public $CBP_PROJECT_REQUIRES_MAINTENANCE = 5;
+    public $CBP_PROJECT_NOT_IN_REQUIRED_ROUND = 6;
 
-    public function can_be_proofed_by_current_user()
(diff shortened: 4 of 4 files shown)


api/v1.inc
```diff
@@ -30,7 +30,11 @@ $router->add_route("GET", "v1/projects/specialdays", "api_v1_projects_specialday
 $router->add_route("GET", "v1/projects/imagesources", "api_v1_projects_imagesources");
 $router->add_route("GET", "v1/projects/:projectid/pages", "api_v1_project_pages");
 $router->add_route("GET", "v1/projects/:projectid/pagedetails", "api_v1_project_pagedetails");
-$router->add_route("GET", "v1/projects/:projectid/pages/:pagename/pagerounds/:pageroundid", "api_v1_project_page_round");
+$router->add_route("GET", "v1/projects/:projectid/pages/:pagename/pagerounds/:pageroundid", "api_v1_project_page_pageround");
+$router->add_route("GET", "v1/projects/:projectid/codepoints", "api_v1_project_codepoints");
+
+$router->add_route("POST", "v1/projects/:projectid/rounds/:roundid", "api_v1_project_round");
+$router->add_route("PUT", "v1/projects/:projectid/pages/:pagename/rounds/:roundid", "api_v1_project_page_round");
```
cpeel, line 37: The whole pageround/round I created is a little weird (thanks OCR "round") but I think we should make this a PUT against the same `v1/projects/:projectid/pages/:pagename/pagerounds/:pageroundid` URL. You can have it as a separate function still and use the `:roundid` to do the round validation for the PUT separately than the GET though.

api/v1.inc
```diff
@@ -30,7 +30,11 @@ $router->add_route("GET", "v1/projects/specialdays", "api_v1_projects_specialday
 $router->add_route("GET", "v1/projects/imagesources", "api_v1_projects_imagesources");
 $router->add_route("GET", "v1/projects/:projectid/pages", "api_v1_project_pages");
 $router->add_route("GET", "v1/projects/:projectid/pagedetails", "api_v1_project_pagedetails");
-$router->add_route("GET", "v1/projects/:projectid/pages/:pagename/pagerounds/:pageroundid", "api_v1_project_page_round");
+$router->add_route("GET", "v1/projects/:projectid/pages/:pagename/pagerounds/:pageroundid", "api_v1_project_page_pageround");
+$router->add_route("GET", "v1/projects/:projectid/codepoints", "api_v1_project_codepoints");
+
+$router->add_route("POST", "v1/projects/:projectid/rounds/:roundid", "api_v1_project_round");
```
cpeel, line 36: I don't think this URL should include the round. There is no instance where a user can checkout a page for a project in a round different than what the project is currently in. Therefore the round should be inferred from the project's current round, not specified by the user.

I think a better URL for this is `v1/projects/:projectid/pages/checkout`.

api/v1_projects.inc
```diff
@@ -271,6 +272,94 @@ function render_project_page_json($row)
...
+    global $projects_url;
+    return [
+        "text" => $lpage->get_text(),
+        "imageUrl" => "$projects_url/{$lpage->projectid}/{$lpage->imagefile}",
+        "pagestate" => $lpage->page_state,
+        "canRestore" => $lpage->can_be_reverted_to_last_save(),
+    ];
+}
+
+function check_project_user($project, $round_id)
```
cpeel, line 355: `validate_user_can_work_in_project()` or something similar would be a much clearer name. I also suggest that this take a `Round` object instead of a `$round_id` since it takes a `Project` object already.

api/v1_projects.inc
```diff
@@ -271,6 +272,94 @@ function render_project_page_json($row)
...
+        throw new BadRequest($exception->getMessage());
+    }
+}
+
+function render_page_data($lpage)
+{
+    global $projects_url;
+    return [
+        "text" => $lpage->get_text(),
+        "imageUrl" => "$projects_url/{$lpage->projectid}/{$lpage->imagefile}",
```
cpeel, line 349: The API returns JSON objects using `this_format`, we should be consistent and use `image_url`, `page_state`, etc.

api/v1_projects.inc
```diff
@@ -271,6 +272,94 @@ function render_project_page_json($row)
...
+
+    $project = $data[":projectid"];
+    $round = $data[":roundid"];
+    check_project_user($project, $round->id);
+
+    try {
+        $lpage = get_indicated_LPage($project->projectid, $project->state, $data[":pagename"], $query_params["pagestate"], 0);
+
+        // can throw InvalidArgumentException
+        $action = get_enumerated_param($query_params, "action", null, ["return", "save_temp", "save_done", "open", "revert"]);
```
cpeel, line 315: In general, `PUT`s should take their inputs from their content not the query params. The input for this should probably look like:
```json
{
  "action": "return",
  "page_text": "this is page text"
}
```

70ray:
Doesn't fit the swagger description exactly.

api/v1_projects.inc
```diff
@@ -271,6 +272,94 @@ function render_project_page_json($row)
...
+            return render_page_data($lpage);
+            break;
+        case "save_done":
+            [$page_saved, $limit_reached] = $lpage->attemptSaveAsDone(api_get_request_body(), $pguser);
+            return ["pageSaved" => $page_saved, "limitReached" => $limit_reached];
+        case "open":
+            $lpage->resume_saved_page($pguser);
+            return render_page_data($lpage);
+            break;
+        case "revert":
```
cpeel, line 333: Reverting is an interface-level artifact that shouldn't be in LPage to begin with (LPage even says as much on line 240). If the user wants to revert, the client should save the page as in-progress, GET the page text from the prior round, and present that to the user.

cpeel (commented):
I know you just closed this so I'll stop adding to the review I had in-progress. Here's a start though.

70ray:
Thanks @cpeel. I found it very hard to make the functions we need fit into the restful framework. I'll look at your comments and make something new.

api/v1_projects.inc
```diff
@@ -4,6 +4,8 @@ include_once($relPath.'wordcheck_engine.inc');
 include_once($relPath.'ProjectSearchForm.inc');
 include_once($relPath.'page_table.inc');
 include_once($relPath.'special_colors.inc');
+include_once($relPath.'CharSuites.inc'); // get_valid_codepoints()
```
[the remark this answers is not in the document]
70ray: Yes.

70ray:
Not really re-opening this yet, just to enable comments.

api/v1_projects.inc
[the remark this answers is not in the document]
70ray: dine this.

api/v1_projects.inc
70ray: changed name but now we don't always have a round object ready made leaving it as roundid

api/v1.inc
70ray: I'm not sure about this. These are rounds that the user can work in. But OCR is not really the same sort of thing.

api/v1.inc
70ray: Made it a parameter.

api/v1_projects.inc
70ray: The way it is now fits quite well onto how LPage works. Should we change the way LPage works first if we want to do this?

api/v1_projects.inc
70ray: done this

api/v1.inc
cpeel: I agree it's odd, but it fits better with the RESTful notion that a URL uniquely defines a resource to work with. Having one URL to get the page and another to save the page is weirder than having a round that you can only get and not set.

api/v1_projects.inc
cpeel: Which is to say leave it as-is but don't use it in the API. We need to design and use the API in ways that make sense now while enabling the functionality we need in a modern interface.

api/v1_projects.inc
cpeel: I realized why "return" and "open" feel weird to me here: they aren't describing an attribute of a page, they're an action. We already know the action which is "persist new information about the page", we just need to know what new information we need to persist.

We should consider inputs for this function like:
```json
{
  "state": "new state",
  "text": "this is page text"
}
```

This function should take and persist the state and text for the page. Based on the current state of the page and the desired state, it might be returning it (and ignoring the page text), saving it (temp or done), or opening it (current state is done and the request is to save it as temp).

api/v1.inc
70ray: I don't think we can use the same GET function that we use for page browse because we want to check that the user can get the text for proofreading. We can use Lpage->resume saved page first, which just does some checks if the page is 'out' or 'temp' and also changes its state to 'temp' if it was 'saved'.

70ray: How what I've made now uses the 'open' action to get the text.

api/v1_projects.inc
```diff
@@ -271,6 +271,93 @@ function render_project_page_json($row)
...
+        $lpage = get_indicated_LPage($project->projectid, $project->state, $data[":pagename"], $page_state, 0);
+        $action = $request_body->action;
+
+        switch ($action) {
+        case "return":
+            $lpage->returnToRound($pguser);
+            return [];
+            break;
+        case "save_temp":
+            $lpage->saveAsInProgress($request_body->text, $pguser);
```
70ray, line 321: We don't need to return the page text in this case unless we allow that it could be different because the js validator is not identical to the php one. But we not concerned about this in the 'save as done' case.

api/v1_projects.inc
70ray: Yes, that seems like a good way to do it.

api/v1_projects.inc
cpeel: You're making the incorrect assumption that the only user of this API could be the UI you are building that uses the JS validator. Anyone can use this API. We may (or may not) want to return the page text in this call, but the schema should be the same for every call.

70ray: Good point.

api/dp-openapi.yaml
```diff
@@ -502,6 +502,113 @@ paths:
...
+      - project
+      description: Checks out a page to the user
+      parameters:
+      - name: projectid
+        in: path
+        description: ID of project
+        required: true
+        schema:
+          type: string
+      - name: roundid
```
chrismiceli, line 545: Can a project be in more than 1 round at a time?

api/dp-openapi.yaml
```diff
@@ -502,6 +502,113 @@ paths:
...
+        in: query
```
chrismiceli, line 546: The roundid here is a query, but in the path below for `/projects/{projectid}/pages/{pagename}/rounds/{roundid}`, do we want them consistent?

api/dp-openapi.yaml
```diff
@@ -502,6 +502,113 @@ paths:
...
+        in: path
+        description: round ID
+        required: true
+        schema:
+          type: string
+      requestBody:
+        content:
+          application/json:
+            schema:
+              type: object
```
chrismiceli, line 595: we can't be more specific here?

api/dp-openapi.yaml
cpeel: See some of the other discussions on the PR. The gist is that we don't want them consistent because while a project can't be in more than one round we want the ability to say "check out a page and the project should be in round X" and if the project isn't in round X then return an error.

api/dp-openapi.yaml
cpeel: We probably want to hold off on reviewing the swagger spec until we've worked out the final details on the interface, then we can make sure the swagger spec is up-to-date. But yes, we need to be clear on the schema that is being returned from this call.

api/v1_projects.inc
70ray: There is a difficulty with this: if you're opening the page it can go into state 'temp' or stay in 'out' so difficult to use the new state to specify what we want to do. There is very little value in distinguishing 'out' and 'temp' so we could abandon it. (The only effect from the users point of view is that if it is 'out' the 'revert to last save' button is disabled).

70ray: To abandon 'out' we could copy the text from previous round when checking out the page. Then, before the user saves the page as temp,  revert to last save would do the same as revert to original.

srjfoo: I thought that `out` was the page state where nothing had been saved. I.e. if a user checks out the page and then returns it to the round before any interim saves have been done. `temp` is after a save in progress has been done (which can be done manually, or be triggered by WC or changing PI orientation).

The distinction is that currently, if a proofer abandons a page before doing any kind of save, and it's the only page they do in a project, that project does not appear in their "My Projects" list. If an interim save is done, it does. If we abandon `out`, there will be no way to make that distinction. I suppose an argument could be made that, regardless of whether or not a proofer had done an interim save, if they return the page to the round, it shouldn't be counted in their My Projects list. I don't have a strong opinion one way or the other on that.

70ray: Thanks, there is more significance than I was aware of. I think the difficulty I was envisaging could easily be overcome by just specify a "out_or_temp" state we want to go into. This text does not represent the actual page_state text but is just a convenience for fitting us into the restful jacket.

70ray:
Changed from round to pageround using the same url as for 'get' but with 'put'. Using old and new pagestates instead of 'action'.

I haven't changed the revert much. It doesn't seem like we could put 'previous' in the url for round because we'd then lose the ability to check if the user can work in the round. We could put 'previous' in the request data, which would not be much different to what we are doing now. Since the LPage code for reverting exists we might as well use it or we'd be duplicating it.

I haven't changed the swagger spec although it should be changed.

api/v1_projects.inc
jmdyck: > ... currently, if a proofer abandons a page before doing any kind of save, and it's the only page they do in a project, that project does not appear in their "My Projects" list. If an interim save is done, it does.

That's correct.

> If we abandon `out`, there will be no way to make that distinction.

No, "My Projects" makes that distinction without looking at page states. (If it had to look at page states, it would probably be fairly slow.)

When you interact with a project in certain ways, the code creates or updates an entry in the `user_project_info table`. In particular, if you save (in-progress or as-done) a page in that project, the `t_latest_page_event` is updated. If you visit a project but never save a page, `t_latest_page_event` will be the default 0. And generally, `my_projects.php` only considers projects for which `t_latest_page_event > 0`, so it ignores projects where you haven't done a save.

cpeel (commented):
We're getting closer!

api/v1_projects.inc
```diff
@@ -225,40 +226,81 @@ function api_v1_project_pagedetails($method, $data, $query_params)
...
+                [$page_saved, $limit_reached] = $lpage->attemptSaveAsDone($request_body->text, $pguser);
+                return ["pageSaved" => $page_saved, "limitReached" => $limit_reached];
+                break;
+            case "resume":
+                $lpage->resume_saved_page($pguser);
+                return render_page_data($lpage);
+                break;
+            case "temp":
+                $lpage->saveAsInProgress($request_body->text, $pguser);
+                if ($request_body->revert) {
```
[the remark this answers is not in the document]
70ray: Are you saying that there shouldn't be a way to get page text from the previous round in the API to do a "restore to original" or that we should not use LPage for this purpose?
The function used allows for previous skipped rounds.

api/v1_projects.inc
```diff
@@ -271,6 +313,47 @@ function render_project_page_json($row)
...
 }
 
+function render_page_data($lpage)
+{
+    global $projects_url;
+    return [
+        "text" => $lpage->get_text(),
+        "image_url" => "$projects_url/{$lpage->projectid}/{$lpage->imagefile}",
+        "page_state" => $lpage->page_state,
+        "canRestore" => $lpage->can_be_reverted_to_last_save(),
```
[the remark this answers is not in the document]
70ray: I agree it would be better for the caller to handle it. It's only purpose is to disable the "restore to last save" button if the text hasn't yet been saved in the round; i.e. if the page is in page_out state. Is the caller allowed to interpret the meaning of the page state?
Alternatively we could not bother to disable the button. But instead copy the text from the "previous round" into the current round. Then if the user pressed the button before he had saved the text he would just get the original text rather than nothing.

70ray:
I've made a new version which works differently, but I'll await a response to my questions about "revert to original" and the "restorable" information before pushing it.

The new version uses a GET to get the page text for proofreading. This is fine as long as the page is in out or temp which it should always be in a prrofreading situation but I put a Page_require in the api to check. This should really be in the LPage or DPage get_text function but that is also used elsewhere when the page is usually in 'saved'. We could add a new function: get_text_for_proofreading to LPage/DPage which would perhaps be better.

It could be convenient to make an additional function in LPage to change page state from 'temp' to 'saved'. Then to 'save as done' you would send the data to save as temp and then change state to done. This would simplify the situation when the user has exceeded the limit because the page would already be saved as temp, so we could just alert the user to the situation without having to then do a 'save as temp' as happens automatically now but would be more difficult if we raise an exception rather than allowing for this case as a normal occurrence.

api/v1_projects.inc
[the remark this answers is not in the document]
cpeel: I was thinking about not using the LPage for it. However, I had totally forgotten about skipped rounds (because UGH). In that case I think we need to keep the revert capability via the API. I don't want to force the client to encode that logic.

cpeel:
> The new version uses a GET to get the page text for proofreading.

`GET`s shouldn't modify the database in any way so I'm curious what your new approach looks like.

api/v1_projects.inc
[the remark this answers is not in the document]
70ray: The problem is that if you reopen a page you worked on earlier it can be in temp state, meaning you saved it earlier or it could still be in out state if you never saved it but just stopped proofreading.

api/v1_projects.inc
```diff
@@ -271,6 +313,47 @@ function render_project_page_json($row)
...
+    try {
+        $lpage = get_available_page($project->projectid, $project->state, $pguser, $err);
+    } catch (ProjectPageException $exception) {
+        throw new BadRequest($exception->getMessage());
+    }
+
+    if (is_null($lpage)) {
+        throw new BadRequest($err);
+    }
+    return ["pagename" => $lpage->imagefile, "page_state" => $lpage->page_state];
```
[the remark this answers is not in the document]
70ray: done this.

api/v1.inc
```diff
@@ -31,10 +31,13 @@ $router->add_route("GET", "v1/projects/imagesources", "api_v1_projects_imagesour
...
 $router->add_route("GET", "v1/projects/:projectid/pagedetails", "api_v1_project_pagedetails");
 $router->add_route("GET", "v1/projects/:projectid/pages/:pagename/pagerounds/:pageroundid", "api_v1_project_page_pageround");
-$router->add_route("PUT", "v1/projects/:projectid/pages/:pagename/pagerounds/:pageroundid", "api_v1_project_page_pageround");
 $router->add_route("GET", "v1/projects/:projectid/codepoints", "api_v1_project_codepoints");
 
 $router->add_route("POST", "v1/projects/:projectid/pages/checkout", "api_v1_project_checkout");
+$router->add_route("GET", "v1/projects/:projectid/pages/:pagename/pageoptions/:pageoption", "api_v1_project_page_option");
+$router->add_route("PATCH", "v1/projects/:projectid/pages/:pagename", "api_v1_project_page");
+$router->add_route("PUT", "v1/projects/:projectid/pages/:pagename", "api_v1_project_page");
+$router->add_route("DELETE", "v1/projects/:projectid/pages/:pagename", "api_v1_project_page");
```
cpeel, line 40: As a user, I would expect this to delete a page but it does't, it only change the state of the page. We should not use the DELETE verb for this.

api/v1_projects.inc
```diff
@@ -271,6 +271,123 @@ function render_project_page_json($row)
...
+
+    $project = $data[":projectid"];
+    $required_roundid = $query_params["roundid"];
+    validate_user_can_work_in_project($project, $required_roundid);
+    $page_state = $query_params["page_state"];
+    try {
+        $projectid = $project->projectid;
+        $lpage = get_indicated_LPage($projectid, $project->state, $data[":pagename"], $page_state, 0);
+        if ($data[":pageoption"] == "original") {
+            $lpage->revertToOriginal();
```
cpeel, line 309: GETs should not modify anything.

cpeel (changes requested):

70ray:
This change leaves out round from the url since it is implied by what round the project is in.
We are creating a new resource with the POST: a page checked out to a user. If it is returned to the round the resource is deleted so we can use DELETE for this.
It would perhaps not be unreasonable to specify the resource in the url as the page state but I haven't done this yet.

api/v1_projects.inc
70ray: It doesn't change anything on the server. LPage is just a temporary construction.

api/v1.inc
70ray: My comment below was later than yours. But if we specify the resource in the url as  the page state then it will be clearer what we are deleting.

cpeel:
Instead of code that keeps changing, can you please present a design we can review and comment on and come to a consensus before throwing code at it?

api/v1.inc
srjfoo: I'm also uncomfortable using the DELETE verb. Isn't there something that describes it better that can be used?

api/v1_projects.inc
70ray: > Instead of code that keeps changing, can you please present a design we can review and comment on and come to a consensus before throwing code at it?

That would be an excellent plan. How can we do this? I'm not sure if the swagger editor is a good way.
I think the profreading functions are difficult to fit into the noun-based restful model because they have previously been verb-based.
But treating a page-checked-out-to-a-user as a resource which can be created, put into different states and deleted would be a way to go.
The addressing mechanism of the url would identify the resource if it exists.
If we used a url like project/page/round/page_state then the validators would determine if the project exists, if the page exists in the project, if the project was in the specified round and if the page is in the specified state. This would duplicate some of the functions of LPage but that doesn't matter much. (Checks on the user also needed).
We want to keep using LPage because at least one of the things it does (getting an available page) is so complicated that it would be risky to do otherwise.
There are several other things to comment on but we first need to identify the best forum.

api/v1.inc
70ray: There are only a small number of verbs available. I think if we use a url perhaps including something which shows clearly this is a page checked out to a user it would be in the spirit of restful to delete it. Or we can move this function into the PUT case, i.e. put the state to available. We would be returning a page state which would be unused but that is not a problem. I think it is not required to have the same input to all cases of the same PUT url since these can be optional. (The other cases of PUT are saves with page text input).

cpeel:
I've created a Google document for us to collaborate on the API design easier than doing it via a PR. Access available via Slack.

70ray:
I've made an edited version. I don't mean to override what you've written
but putting in suggested edits or comments seemed to get very messy. I
think your original version is still available so my version could be
treated as an alternative suggestion.
Ray

cpeel:
Closing this PR until we have an agreed-upon design to implement.

PR #784, opened 2022-02-22
Copy project holds when cloning

cpeel:
When cloning a project, clone the project holds as well. This is [Task 1993](https://www.pgdp.net/c/tasks.php?task_id=1993). This was super easy now that we hold the project data in a Project object so metadata like this doesn't get in the way.

Testable in the [clone-project-holds](https://www.pgdp.org/~cpeel/c.branch/clone-project-holds/) sandbox.

### DIFF
--- a/tools/project_manager/editproject.php
+++ b/tools/project_manager/editproject.php
@@ -167,6 +167,9 @@ class ProjectInfoHolder
             array_push($this->charsuites, $project_charsuite->name);
         }
 
+        // and the project's holds
+        $this->hold_states = $this->project->get_hold_states();
+
         // reset project values that should not be cloned
         $this->project->projectid = null;
         $this->project->postednum = '';
@@ -304,6 +307,10 @@ class ProjectInfoHolder
         }
 
         $this->original_marc_array_encd = @$_POST['rec'];
+        $this->hold_states = array_intersect(
+            $this->unserialize(@$_POST['hold_states']) ?? [],
+            Project::get_holdable_states(),
+        );
 
         return $errors;
     }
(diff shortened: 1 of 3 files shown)


tools/project_manager/editproject.php
```diff
@@ -304,6 +307,7 @@ public function set_from_post()
         }
 
         $this->original_marc_array_encd = @$_POST['rec'];
+        $this->hold_states = $this->unserialize(@$_POST['hold_states']) ?? [];
```
chrismiceli, line 310: any reason we shouldn't trust the client with whatever hold states it sends here? I can't think of one to be honest.

cpeel: Well, we should never trust user input 😁 

We correctly escape values before putting them into the database, so there is low risk here, but validation is always better and I need to do this when I add this to the API too. Thanks for calling it out.

srjfoo (approved):

tools/project_manager/editproject.php
```diff
@@ -560,4 +573,18 @@ public function normalize_spaces()
         $this->project->clearance = preg_replace('/\s+/', ' ', trim($this->project->clearance));
         $this->project->extra_credits = preg_replace('/\s+/', ' ', trim($this->project->extra_credits));
     }
+
+    private function serialize($value)
+    {
```
70ray, line 578: Could not redefining a function with the same name lead to confusion for future generations?

70ray (approved):

tools/project_manager/editproject.php
cpeel: It's a class function so it isn't overriding the built-in (not that PHP supports that like python does). The only way to call this function is to use the object.

PR #787, opened 2022-02-27
Remove smooth dir on archive

cpeel:
When a project is archived, remove the `smooth/` directory. This directory contains redundant copies of existing files as a convenience to smooth readers and does not need to be retained. [Task 2017](https://www.pgdp.net/c/tasks.php?task_id=2017)

I also updated the archiving files to use the DPDatabase functions which is frankly the bulk of the PR.

I will provide testing instructions directly to @srjfoo.

### DIFF
--- a/pinc/archiving.inc
+++ b/pinc/archiving.inc
@@ -31,10 +31,11 @@ function archive_project($project, $dry_run)
     } elseif ($dry_run) {
         echo "    Move table $projectid to $archive_db_name.\n";
     } else {
-        mysqli_query(DPDatabase::get_connection(), "
+        $sql = "
             ALTER TABLE $projectid
             RENAME AS $archive_db_name.$projectid
-        ") or die(DPDatabase::log_error());
+        ";
+        DPDatabase::query($sql);
     }
 
     $project_dir = $project->dir;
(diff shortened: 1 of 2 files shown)


pinc/archiving.inc
```diff
@@ -26,15 +26,25 @@ function archive_project($project, $dry_run)
...
 
-    if (!does_project_page_table_exist($projectid)) {
+    if (!$project->pages_table_exists) {
         echo "    Table $projectid does not exist.\n";
     } elseif ($dry_run) {
         echo "    Move table $projectid to $archive_db_name.\n";
     } else {
-        mysqli_query(DPDatabase::get_connection(), "
+        $sql = "
             ALTER TABLE $projectid
```
chrismiceli, line 35: you escape projectid below, but not here. Is it intentional?

chrismiceli (approved):

pinc/archiving.inc
cpeel: Because this is a table name, not a string. See [Validating project tables](https://www.pgdp.net/wiki/DP_Code_Best_Practices#Validating_project_tables) in our best practices. In this case the only caller is pulling the fields from the DB so we know they are valid projectids and thus valid tables names.

srjfoo (approved):

70ray (approved):

PR #790, opened 2022-03-05
Better project validation

cpeel:
Sharon's testing in #783 showed that we are missing some validation corner cases. I thought it would be best if I pulled those out to address separately.

Specifically this:
* Validates `comment_format` value (originally in 783 but is better here)
* Correctly validates `image_source` -- the prior code was simply wrong
* Validates Project field types during `validate()`. This is most notable through the API where we can assign an array to a string field which can confuse some other parts of the `validate()` checks and causes warnings as PHP casts the array into a string during `save()`.
* Updates the default value for `postednum` to be `null`. This matches what is actually in the database (NULL) but requires some massaging in the Edit Project page. This is, coincidentally, the only non-string value we allow users to update.

Testable in the [better-project-validation](https://www.pgdp.org/~cpeel/c.branch/better-project-validation/) sandbox.

Suggestions for testing is to ensure setting (and unsetting) `postednum` via the Edit Project page works as expected.

### DIFF
--- a/SETUP/tests/ProjectTest.php
+++ b/SETUP/tests/ProjectTest.php
@@ -197,6 +197,14 @@ class ProjectTest extends PHPUnit\Framework\TestCase
         $this->assertStringContainsString("required", $errors[0]);
     }
 
+    public function test_validate_nameofwork_wrong_type()
+    {
+        $project = new Project($this->valid_project_data);
+        $project->nameofwork = [];
+        $errors = $project->validate();
+        $this->assertStringContainsString("should be of type", $errors[0]);
+    }
+
     public function test_validate_authorsname_missing()
     {
         $project = new Project($this->valid_project_data);
--- a/pinc/Project.inc
+++ b/pinc/Project.inc
@@ -162,6 +162,7 @@ class Project
             "genre",
             "image_source",
             "difficulty",
+            "comment_format",
         ];
 
         foreach ($required_fields as $field) {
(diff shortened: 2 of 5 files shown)


chrismiceli (approved):

srjfoo:
First test, I went into a "posted" project, and removed the posted number, but never got far enough to see whether I got a complaint about that or not -- it got hung up on `Field ppverifier should be of type string`. I get the error if I try to save the project page at all, whether I've made changes or not.

Which ... I think `ppverifier` is something that's set in the code, isn't it? The only thing the PM can change in the project page is `checkedoutby`.

If a potential PPer wishes to reserve a project in the rounds, or checks one out from the pool. the value that goes into that slot for the project page comes from the `checkedoutby` field.

If the PPer still needs PPV, once they upload for PPV, the username in `checkedoutby` is copied into the `postproofer` field, but `checkedoutby` stays the same until the PPVer checks it out, at which point `checkedoutby` contains the PPVerifier's username, and `postproofer` contains the original PPer's username.

`ppverifier` is only set when a project is marked posted, and the PPer has to put their projects through PPV. If the PPer has DU, the `ppverifier` field remains `NULL`. (Yes, I know, it's convoluted -- I'm just trying to explain how it actually works, not how it necessarily _**should**_ work.)

cpeel:
Good catch, Sharon! `ppverifier` is one of the few nullable string fields and is null by default. The `Project->save()` code wasn't correctly handling that case.

70ray (approved):

srjfoo (approved):
Second round of testing using the `postednum` suggestion, all looks good. Let me know if there's anything else you can think of (I poked at setting/unsetting a few of the other fields, too).

PR #794, opened 2022-03-19
change structure of pickerset

70ray:
so subsets are in a numerically keyed array
rather than keyed by name.
This will ensure order is preserved if json encoded

If we use the pickerset in an api we would also have to incorporate the utf8_character_names into it but this is not done in this request.

### DIFF
--- a/SETUP/lint_charsuites.php
+++ b/SETUP/lint_charsuites.php
@@ -57,7 +57,7 @@ foreach (CharSuites::get_all() as $charsuite) {
     validate_hex_codes_lowercase($charsuite->codepoints, "uppercase hex values found in charsuite codepoints");
 
     // Validate pickersets only contain characters within the suite
-    foreach ($charsuite->pickerset->get_subsets() as $title => $picker_subset) {
+    foreach ($charsuite->pickerset->get_subsets() as  [$title, $long_title, $picker_subset]) {
         foreach ($picker_subset as $codepoints) {
             $string = implode("", convert_codepoint_ranges_to_characters($codepoints));
             $invalid_chars = get_invalid_characters($string, $charsuite->codepoints);
--- a/pinc/CharSuites.inc
+++ b/pinc/CharSuites.inc
@@ -11,23 +11,15 @@ class PickerSet
 
     public function add_subset($name, $codepoints, $title = null)
     {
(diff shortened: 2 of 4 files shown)


cpeel:
Can you elaborate on why we're changing this here instead of in the hydration to the API?

This enforces the order by building a numeric-based array when that is needed for returning via the API for instance:
```php
<?php
$relPath="./";

$charsuite = CharSuites::get("basic-latin");
$picker_set = $charsuite->pickerset;
$subset = [];
foreach ($picker_set->get_subsets() as $code => $picker) {
    $subset[] = ["name" => $code, "values" => $picker];
}
echo json_encode($subset, JSON_PRETTY_PRINT | JSON_UNESCAPED_UNICODE |
    JSON_UNESCAPED_SLASHES | JSON_NUMERIC_CHECK);
echo "\n";
```

70ray:
> Can you elaborate on why we're changing this here instead of in the hydration to the API?
> 
> This enforces the order by building a numeric-based array when that is needed for returning via the API for instance:
> 
> ```
> <?php
> $relPath="./";
> include_once($relPath."base.inc");
> include_once($relPath."CharSuites.inc");
> 
> $charsuite = CharSuites::get("basic-latin");
> $picker_set = $charsuite->pickerset;
> $subset = [];
> foreach ($picker_set->get_subsets() as $code => $picker) {
>     $subset[] = ["name" => $code, "values" => $picker];
> }
> echo json_encode($subset, JSON_PRETTY_PRINT | JSON_UNESCAPED_UNICODE |
>     JSON_UNESCAPED_SLASHES | JSON_NUMERIC_CHECK);
> echo "\n";
> ```

You could do like that but the other way seemed simpler, it actually reduces the code by 9 lines.

cpeel:
It may seem simpler, but it's less clear. You've changed an explicit object format into an implicit format (`[$name, $title, $codepoints]`).

70ray:
> It may seem simpler, but it's less clear. You've changed an explicit object format into an implicit format (`[$name, $title, $codepoints]`).

I don't agree. Letting each subset object have a name, title and a set of codepoints seems an obvious way to structure it. If we wanted to get the title or codepoints from a given name then there would be more value in using a name key but we only ever enumerate the whole set of subsets.

PR #795, opened 2022-03-21
Fix two project edit bugs

cpeel:
This fixes two bugs that have been found on the edit project page that came in with a71d4e30a:
* Correctly output a multi-level variable reference within a string. If we don't wrap it in `{}` only the part is used which is an object which raises an error and breaks the page.
* Remove PP-centric fields from cloned projects. These are being incorrectly cloned.

Testable in the [edit-project-updates](https://www.pgdp.org/~cpeel/c.branch/edit-project-updates/) sandbox.

### DIFF
--- a/tools/project_manager/editproject.php
+++ b/tools/project_manager/editproject.php
@@ -490,7 +490,7 @@ class ProjectInfoHolder
         if ($this->project->difficulty == "beginner" && !$can_set_difficulty_tofrom_beginner) {
             // allow PF to edit a BEGIN project, but without altering the difficulty
             $this->row(_("Difficulty"), 'just_echo', _("Beginner"));
-            echo "<input type='hidden' name='difficulty' value='$this->project->difficulty'>";
+            echo "<input type='hidden' name='difficulty' value='{$this->project->difficulty}'>";
         } else {
             $this->row(_("Difficulty"), 'difficulty_list', $this->project->difficulty);
         }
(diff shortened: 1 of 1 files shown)


srjfoo:
I've done some quick testing, and the specific issues that were uncovered have been fixed, but I'd like to do some more extensive testing.

One additional field that I thought of that needs to be cleared/set to default when a project is cloned is the `archived` field, since a project that has been posted to PG and archived can also be cloned.

chrismiceli (approved):

cpeel:
> One additional field that I thought of that needs to be cleared/set to default when a project is cloned is the `archived` field, since a project that has been posted to PG and archived can also be cloned.

`archived` isn't a field that `Project->save()` manages so it can't be edited and won't be included in the clone. However, it would be better to include it in `editproject.php` should that change in the Project class. Update forthcoming.

srjfoo:
> > One additional field that I thought of that needs to be cleared/set to default when a project is cloned is the `archived` field, since a project that has been posted to PG and archived can also be cloned.
> 
> `archived` isn't a field that `Project->save()` manages so it can't be edited and won't be included in the clone. However, it would be better to include it in `editproject.php` should that change in the Project class. Update forthcoming.

I was just about to comment that I've been doing some research, and that it looked like there wasn't a problem with it. I cloned a couple of projects on TEST, and compared the fields in the `projects` table records for before and after for both source projects and their clones. In both cases, the `archived` bit was not set in the clone. Is it likely to cause problems in the future, or is this a case where we know absolutely that the field should not be set on a cloned project, so it makes no difference?

cpeel:
> Is it likely to cause problems in the future, or is this a case where we know absolutely that the field should not be set on a cloned project, so it makes no difference?

It's not likely to cause problems in the future, but its possible that someone updates `Project->save()` to manage `archived` and clone would then clone it if we didn't have the fix in place (which I had pushed up but just realized I hadn't published to my sandbox, one small re-test of cloning an archived project would be appreciated!).

srjfoo (approved):
Everything looks good after the second round of testing!

PR #800, opened 2022-04-01
expand ranges and add titles in pickerset

70ray:
This incorporates the character names (titles) in the picker set so they can be used via an API. PHP has a function for getting the names from the character codes but other languages such as javascript do not. It would be possible to incorporate them in the api function by breaking down the pickerset and building it up again, but putting the names in when it is first built seems a better option and only needs some small changes to existing code.


70ray:
needs more work.

PR #804, opened 2022-04-16
swagger file for pi api

70ray:
This is the swagger definition of a proposed profreading interface. It does not include any code.
The character validation is done only in the server when a page is saved or wordchecked. It returns code 422 with data indicating the bad characters if any are found.
Wordcheck returns data indicating bad words. The client can mark punctuation using a set of punctuation characters it gets from the server.

### DIFF
--- a/SETUP/tests/apiTest.php
+++ b/SETUP/tests/apiTest.php
@@ -1,0 +1,575 @@
+<?php
+
+// api functions in api/v1_projects.inc
+
+class ApiTest extends PHPUnit\Framework\TestCase
+{
+    private $TEST_USERNAME_PM = "ProjectTestPM_php";
+    private $TEST_USERNAME = "ProjectTest_php";
+    private $TEST_USERNAME_2 = "ProjectTest_2_php";
+    private $TEST_OLDUSERNAME = "ProjectTest_old_php";
+    private $TEST_IMAGESOURCE = "PrjTest";
+    private $valid_projectID = "projectID45c225f598e32";
+    private $valid_page_image = "001.png";
+    private $valid_project_data = [
+        "nameofwork" => "War and Peace",
+        "authorsname" => "Bob Smith",
+        "language" => "English",
+        "genre" => "Other",
+        "difficulty" => "average",
+        "clearance" => "123abc",
+        // username and image_source are set in setUp() after creation
+    ];
+    private $created_projectids = [];
+
+    private $TEST_IMAGE_FILE_NAME = "001.png";
+    private $TEST_IMAGE_FILE_NAME_2 = "002.png";
+    private $TEST_TEXT = "This is a test file";
+    private $TEST_MODIFIED_TEXT = "This is a modified test file";
+    private $TEST_BAD_TEXT = "This is a bĀd test file";
+
+    protected function setUp(): void
+    {
+        create_test_user($this->TEST_USERNAME_PM);
+
+        // make the user a PM
+        $settings = new Settings($this->TEST_USERNAME_PM);
+        $settings->set_boolean("manager", true);
+
+        create_test_user($this->TEST_USERNAME);
+
+        // create an old user (new users $days_on_site_threshold = 21)
+        create_test_user($this->TEST_OLDUSERNAME, 22);
+
+        create_test_image_source($this->TEST_IMAGESOURCE);
+
+        $this->valid_project_data["username"] = $this->TEST_USERNAME_PM;
+        $this->valid_project_data["image_source"] = $this->TEST_IMAGESOURCE;
+    }
+
+    protected function tearDown(): void
+    {
+        // clean up the PM value
+        $settings = new Settings($this->TEST_USERNAME_PM);
+        $settings->set_value("manager", null);
+
+        delete_test_user($this->TEST_USERNAME_PM);
+        delete_test_user($this->TEST_USERNAME);
+        delete_test_user($this->TEST_OLDUSERNAME);
+        delete_test_image_source($this->TEST_IMAGESOURCE);
+
+        foreach ($this->created_projectids as $projectid) {
+            $project = new Project($projectid);
+            delete_test_project_remains($project);
+        }
+    }
+
+    // -----------------------------------------------------------
+    // helper functions
+
+    protected function _create_project()
+    {
+        $project = new Project();
+        foreach ($this->valid_project_data as $key => $value) {
+            $project->$key = $value;
+        }
+        $project->save();
+        $this->created_projectids[] = $project->projectid;
+
+        return $project;
+    }
+
+    protected function _create_available_project_with_page($npage = 1)
+    {
+        $project = $this->_create_project();
+        $project->add_charsuite("basic-latin");
+
+        $this->add_page($project, "001");
+        if ($npage > 1) {
+            $this->add_page($project, "002");
+        }
+
+        project_transition($project->projectid, PROJ_P1_UNAVAILABLE, $this->TEST_USERNAME_PM);
+        project_transition($project->projectid, PROJ_P1_WAITING_FOR_RELEASE, $this->TEST_USERNAME_PM);
+        project_transition($project->projectid, PROJ_P1_AVAILABLE, PT_AUTO);
+
+        return $project;
+    }
+
+    protected function _create_available_project_round2()
+    {
+        $project = $this->_create_available_project_with_page();
+        project_transition($project->projectid, PROJ_P1_COMPLETE, PT_AUTO);
+        project_transition($project->projectid, PROJ_P1_COMPLETE, PT_AUTO);
+        project_transition($project->projectid, PROJ_P2_WAITING_FOR_RELEASE, PT_AUTO);
+        project_transition($project->projectid, PROJ_P2_AVAILABLE, PT_AUTO);
+
+        return $project;
+    }
+
+    protected function add_page($project, $base)
+    {
+        $source_project_dir = sys_get_temp_dir();
+        $txt_file_name = "$base.txt";
+        $txt_file_path = "$source_project_dir/$txt_file_name";
+        $fp = fopen($txt_file_path, 'w');
+        fwrite($fp, $this->TEST_TEXT);
+        fclose($fp);
+
+        $image_file_name = "$base.png";
+        $img_file_path = "{$project->dir}/$image_file_name";
+        $fp = fopen($img_file_path, 'w');
+        fwrite($fp, str_repeat("This is a test image file", 10));
+        fclose($fp);
+
+        project_add_page($project->projectid, $base, $image_file_name, $txt_file_path, $this->TEST_USERNAME_PM, time());
+    }
+
+    protected function checkout($roundid, $projectid)
+    {
+        // check out a page
+        $path = "v1/rounds/$roundid/projects/$projectid/pages";
+        $query_params = "";
+        $router = ApiRouter::get_router();
+        $_SERVER["REQUEST_METHOD"] = "POST";
+        return $router->route($path, $query_params);
+    }
+
+    protected function return_to_round($projectid, $page)
+    {
+        global $request_body;
+
+        $path = "v1/rounds/P1/projects/$projectid/pages/$page";
+        $query_params = "";
+        $router = ApiRouter::get_router();
+        $request_body = ["new_page_state" => "avail"];
+        $_SERVER["REQUEST_METHOD"] = "PUT";
+        return $router->route($path, $query_params);
+    }
+
+    protected function save_as_done($projectid, $page, $text)
+    {
+        global $request_body;
+
+        $path = "v1/rounds/P1/projects/{$projectid}/pages/$page";
+        $query_params = "";
+        $router = ApiRouter::get_router();
+        $request_body = ["new_page_state" => "saved", "text" => $text];
+        $_SERVER["REQUEST_METHOD"] = "PUT";
+        return $router->route($path, $query_params);
+    }
+
+    protected function get_text($projectid, $page, $option)
+    {
+        $path = "v1/rounds/P1/projects/{$projectid}/pages/$page";
+        $query_params = ["pageoption" => $option];
+        $router = ApiRouter::get_router();
+        $_SERVER["REQUEST_METHOD"] = "GET";
+        return $router->route($path, $query_params);
+    }
+
+    // ----------------------------------------------------
+    // tests
+
+    public function test_checkout_invalid_round()
+    {
+        global $pguser;
+
+        $this->expectExceptionCode(DP::INVALID_ROUND);
+
+        $this->checkout("P0", "projectId");
+    }
+
+    public function test_checkout_bad_projectid()
+    {
+        global $pguser;
+
+        $this->expectExceptionCode(DP::NO_SUCH_PROJECT_ID);
+
+        $this->checkout("P1", "projectId");
+    }
+
+    public function test_user_access_denied()
+    {
+        global $pguser;
+
+        $this->expectExceptionCode(DP::USER_NOT_QUALIFIED_FOR_ROUND);
+
+        $project = $this->_create_project();
+        $pguser = $this->TEST_USERNAME;
+
+        $this->checkout("P3", $project->projectid);
+    }
+
+    public function test_project_not_in_round()
+    {
+        global $pguser;
+
+        $this->expectExceptionCode(DP::PROJECT_NOT_IN_ROUND);
+
+        $project = $this->_create_project();
+        $pguser = $this->TEST_USERNAME;
+
+        $this->checkout("P1", $project->projectid);
+    }
+
+    public function test_project_unavailable()
+    {
+        global $pguser;
+
+        $this->expectExceptionCode(DP::PROJECT_NOT_AVAILABLE);
+
+        $project = $this->_create_project();
+        $pguser = $this->TEST_USERNAME;
+        project_transition($project->projectid, PROJ_P1_UNAVAILABLE, $this->TEST_USERNAME_PM);
+
+        $this->checkout("P1", $project->projectid);
+    }
+
+    public function test_project_checkout_available()
+    {
+        global $pguser, $ELR_round;
+
+        // new project, new user
+        $project = $this->_create_available_project_with_page();
+        $pguser = $this->TEST_USERNAME;
+
+        $response = $this->checkout("P1", $project->projectid);
+        $expected = [
+            'pagename' => $this->TEST_IMAGE_FILE_NAME,
+            'image_url' => "{$project->url}/{$this->TEST_IMAGE_FILE_NAME}",
+            'text' => $this->TEST_TEXT,
+        ];
+        $this->assertEquals($response, $expected);
+
+        $page = $this->TEST_IMAGE_FILE_NAME;
+        $this->return_to_round($project->projectid, $page);
+
+        // user done many pages but few days on site
+        // $page_tally_threshold 500 for new projects in reserve time
+        page_tallies_add($ELR_round->id, $pguser, 501);
+        $response = $this->checkout("P1", $project->projectid);
+        $this->assertEquals($response, $expected);
+        $this->return_to_round($project->projectid, $page);
+
+        // few pages, many days on site
+        $pguser = $this->TEST_OLDUSERNAME;
+        $response = $this->checkout("P1", $project->projectid);
+        $this->assertEquals($response, $expected);
+        $this->return_to_round($project->projectid, $page);
+
+        // user done many pages, many days on site
+        $this->expectExceptionCode(DP::RESERVED_FOR_NEW_PROOFREADERS);
+        page_tallies_add($ELR_round->id, $pguser, 501);
+        $this->checkout("P1", $project->projectid);
+    }
+
+    public function test_beginner_project_checkout()
+    {
+        global $pguser, $ELR_round;
+
+        $this->expectExceptionCode(DP::BEGINNERS_QUOTA_REACHED);
+
+        $this->valid_project_data["difficulty"] = "beginner";
+        $project = $this->_create_available_project_with_page();
+        $pguser = $this->TEST_USERNAME;
+        // beginners limit is 40 in user_is.inc
+        page_tallies_add($ELR_round->id, $pguser, 50);
+
+        $this->checkout("P1", $project->projectid);
+    }
+
+    public function test_beginner_mentor_project_checkout()
+    {
+        global $pguser, $ELR_round;
+
+        $this->expectExceptionCode(DP::NO_ACCESS_TO_MENTORS_ONLY);
+
+        $this->valid_project_data["difficulty"] = "beginner";
+        $project = $this->_create_available_project_with_page();
+        $project = $this->_create_available_project_round2();
+        $pguser = $this->TEST_USERNAME;
+        $user = new User($pguser);
+        $user->grant_access("P2", $pguser);
+
+        $this->checkout("P2", $project->projectid);
+    }
+
+    public function test_project_checkout_wrong_round()
+    {
+        global $pguser;
+
+        $this->expectExceptionCode(DP::PROJECT_NOT_IN_REQUIRED_ROUND);
+
+        $project = $this->_create_available_project_with_page();
+        $pguser = $this->TEST_USERNAME;
+        $user = new User($pguser);
+        $user->grant_access("P2", $pguser);
+
+        $this->checkout("P2", $project->projectid);
+    }
+
+    public function test_project_checkout_no_more_pages()
+    {
+        global $pguser;
+
+        $this->expectExceptionCode(DP::NO_MORE_PAGES);
+
+        $project = $this->_create_available_project_with_page();
+        $pguser = $this->TEST_USERNAME;
+
+        $this->checkout("P1", $project->projectid);
+        // try to get another page
+        $this->checkout("P1", $project->projectid);
+    }
+
+    public function test_return_to_round()
+    {
+        global $pguser;
+
+        $pguser = $this->TEST_USERNAME;
+        $project = $this->_create_available_project_with_page();
+
+        // check out a page
+        $response = $this->checkout("P1", $project->projectid);
+
+        // return it to round
+        $page = $response['pagename'];
+        $response = $this->return_to_round($project->projectid, $page);
+        $this->assertEquals($response, null);
+    }
+
+    public function test_return_non_existent_page()
+    {
+        global $pguser;
+
+        $this->expectExceptionCode(DP::NO_SUCH_PAGE);
+
+        $pguser = $this->TEST_USERNAME;
+        $project = $this->_create_available_project_with_page();
+
+        // check out a page
+        $this->checkout("P1", $project->projectid);
+
+        // return another page to round
+        $this->return_to_round($project->projectid, $this->TEST_IMAGE_FILE_NAME_2);
+    }
+
+    public function test_return_page_wrong_state()
+    {
+        global $pguser;
+
+        $this->expectExceptionCode(DP::WRONG_PAGE_STATE);
+
+        $pguser = $this->TEST_USERNAME;
+        $project = $this->_create_available_project_with_page();
+
+        // return a page to round without checkout first
+        $this->return_to_round($project->projectid, $this->TEST_IMAGE_FILE_NAME);
+    }
+
+    public function test_return_page_not_owned()
+    {
+        global $pguser;
+
+        $this->expectExceptionCode(DP::WRONG_PAGE_OWNER);
+
+        $pguser = $this->TEST_USERNAME;
+        $project = $this->_create_available_project_with_page();
+
+        // check out a page
+        $this->checkout("P1", $project->projectid);
+
+        // let another user return it to round
+        $pguser = $this->TEST_USERNAME_2;
+        $this->return_to_round($project->projectid, $this->TEST_IMAGE_FILE_NAME);
+    }
+
+    public function test_save_as_in_progress()
+    {
+        global $pguser, $request_body;
+
+        $pguser = $this->TEST_USERNAME;
+        $project = $this->_create_available_project_with_page();
+
+        // check out a page
+        $response = $this->checkout("P1", $project->projectid);
+
+        // save as in progress
+        $page = $response['pagename'];
+        $path = "v1/rounds/P1/projects/{$project->projectid}/pages/$page";
+        $query_params = "";
+        $router = ApiRouter::get_router();
+        $request_body = ["new_page_state" => "temp", "text" => $this->TEST_MODIFIED_TEXT];
+        $_SERVER["REQUEST_METHOD"] = "PUT";
+        $response = $router->route($path, $query_params);
+        $this->assertEquals($response, null);
+
+        // get original text back
+        $query_params = ["pageoption" => "original"];
+        $_SERVER["REQUEST_METHOD"] = "GET";
+        $response = $router->route($path, $query_params);
+        $this->assertEquals($response, ["text" => $this->TEST_TEXT]);
+
+        // get last saved text
+        $query_params = ["pageoption" => "current"];
+        $_SERVER["REQUEST_METHOD"] = "GET";
+        $response = $router->route($path, $query_params);
+        $this->assertEquals($response, ["text" => $this->TEST_MODIFIED_TEXT]);
+
+        // save bad text
+        $query_params = "";
+        $request_body = ["new_page_state" => "temp", "text" => $this->TEST_BAD_TEXT];
+        $_SERVER["REQUEST_METHOD"] = "PUT";
+        $response = $router->route($path, $query_params);
+        $this->assertEquals($response, null);
+
+        // get saved text back - bad char removed
+        $query_params = ["pageoption" => "current"];
+        $_SERVER["REQUEST_METHOD"] = "GET";
+        $response = $router->route($path, $query_params);
+        $this->assertEquals($response, ["text" => "This is a bd test file"]);
+    }
+
+    public function test_invalid_text()
+    {
+        global $request_body;
+
+        $project = $this->_create_available_project_with_page();
+        $query_params = "";
+        $router = ApiRouter::get_router();
+        $request_body = ["text" => $this->TEST_BAD_TEXT];
+        $path = "v1/projects/{$project->projectid}/validatetext";
+        $_SERVER["REQUEST_METHOD"] = "POST";
+        $response = $router->route($path, $query_params);
+        $expected_response = [
+            "mark_array" => [["This is a b", 0], ["Ā", 1], ["d test file", 0]],
+            "good" => false,
+        ];
+        $this->assertEquals($response, $expected_response);
+
+        $request_body = ["text" => $this->TEST_TEXT];
+        $response = $router->route($path, $query_params);
+        $expected_response = [
+            "mark_array" => [[$this->TEST_TEXT, 0]],
+            "good" => true,
+        ];
+        $this->assertEquals($response, $expected_response);
+    }
+
+    public function test_save_as_done()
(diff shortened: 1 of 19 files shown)


api/dp-openapi.yaml
```diff
@@ -784,6 +1132,163 @@ components:
...
+      type: object
+      properties:
+        advice:
+          type: string
+          description: html encoded text
+          example: Common Fixes ...
+        reasons:
+          type: array
+          description: reasons to mark page bad
+          items:
```
chrismiceli, line 1286: we want to include the other reasons too, right?

api/dp-openapi.yaml
```diff
@@ -784,6 +1132,163 @@ components:
...
+          type: array
+          items:
+            type: string
+
+    bad_page_data:
+      type: object
+      properties:
+        advice:
+          type: string
+          description: html encoded text
```
chrismiceli, line 1281: would an array of strings of the common fixes? This feels weird in a REST endpoint to be honest in my opinion.

api/dp-openapi.yaml
```diff
@@ -591,6 +591,354 @@ paths:
...
+      responses:
+        200:
+          description: Data
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/bad_page_data'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '404':
```
chrismiceli, line 608: is this possible?

api/dp-openapi.yaml
```diff
@@ -784,6 +1132,163 @@ components:
...
+            items:
+              type: object
+              properties:
+                name:
+                  type: string
+                  example: ˆ
+                title:
+                  type: string
+                  example: Letters with acute, grave, and circumflex
+                rows:
```
chrismiceli, line 1242: Rows feels like a UI concept that doesn't belong in the REST api.

api/dp-openapi.yaml
```diff
@@ -591,6 +591,354 @@ paths:
...
+
+  /rounds/{roundid}:
+    get:
+      tags:
+      - project
+      description: Gets round information
+      parameters:
+      - name: roundid
+        in: path
+        description: The round identifier (as defined by array_keys($Round_for_round_id_)
```
chrismiceli, line 651: missing close paren?

api/dp-openapi.yaml
```diff
@@ -591,6 +591,354 @@ paths:
...
+
+  /proofrounds/{proofroundid}/proofprojects/{proofprojectid}/proofpages:
+    post:
+      tags:
+      - project
+      description: Checks out a page for proofreading
+      parameters:
+      - name: proofroundid
+        in: path
+        description: The round identifier (as defined by array_keys($Round_for_round_id_)
```
chrismiceli, line 677: same paren

api/dp-openapi.yaml
```diff
@@ -784,6 +1132,163 @@ components:
...
+        text:
+          type: string
+        auxlanguage:
+          type: string
+        adhocgoodwords:
+          type: array
+          items:
+            type: string
+
+    page_save:
```
chrismiceli, line 1195: not sure how swagger would represent this, but should we have 1 type with optional entries for text,  and reason instead of 3 and the oneOf? Not sure if it is even better or not.

api/dp-openapi.yaml
```diff
@@ -591,6 +591,354 @@ paths:
...
+        429:
+          $ref: '#/components/responses/RateLimitExceeded'
+        default:
+          $ref: '#/components/responses/UnexpectedError'
+
+  /proofrounds/{proofroundid}/proofprojects/{proofprojectid}/ownedpages/{ownedpage}:
+    put:
+      tags:
+      - project
+      description: make page available for proofreading, change state to temp if necessary
```
chrismiceli, line 756: is there a way to capture what roles are required to perform certain operations?

api/dp-openapi.yaml
```diff
@@ -591,6 +591,354 @@ paths:
...
+          type: string
+      - name: proofpage
+        in: path
+        description: the name of the page, must be in temp state and owned by user
+        required: true
+        schema:
+          type: string
+      - name: pageoption
+        in: path
+        description: current or original
```
chrismiceli, line 818: should we use enum for the scheama? https://swagger.io/docs/specification/data-models/enums/

chrismiceli, line 818: does original only get previous round? Should it be called previous?

api/dp-openapi.yaml
```diff
@@ -784,6 +1132,163 @@ components:
...
+    wordcheck_data:
+      type: object
+      properties:
+        langsWithDict:
+          type: array
+          items:
+            type: string
+        messages:
+          type: object
+        puncChars:
```
chrismiceli, line 1144: what does this mean?

api/dp-openapi.yaml
```diff
@@ -784,6 +1132,163 @@ components:
...
+          type: string
+        adhocgoodwords:
+          type: array
+          items:
+            type: string
+
+    page_save:
+      type: object
+      properties:
+        new_page_state:
```
chrismiceli, line 1198: we aren't consistent in camel case vs snake case in the properties on these objects

chrismiceli (commented):
My comments are relatively generic and not related to the design, so don't put too much effort into rectifying them until the design gets properly reviewed.

api/dp-openapi.yaml
70ray: Probably not. I just copied all the error responses at the end of each  endpoint.

api/dp-openapi.yaml
70ray: Yes, I suppose it is only important for upper and lower case which we should describe more explicitly.

api/dp-openapi.yaml
70ray: Maybe so.

api/dp-openapi.yaml
70ray: Yes, this is just an example.

api/dp-openapi.yaml
cpeel: We should use snake case everywhere in here to be consistent with the existing API.

api/dp-openapi.yaml
```diff
@@ -591,6 +591,354 @@ paths:
...
+              schema:
+                $ref: '#/components/schemas/round_data'
+        404:
+          $ref: '#/components/responses/NotFound'
+        429:
+          $ref: '#/components/responses/RateLimitExceeded'
+        default:
+          $ref: '#/components/responses/UnexpectedError'
+
+  /proofrounds/{proofroundid}/proofprojects/{proofprojectid}/proofpages:
```
cpeel, line 669: Can you talk about why you went for a `round -> project -> page` hierarchy? It feels a bit awkward since projects shift between rounds but rounds don't shift between projects, so the most stable part of the hierarchy is the project, so: `project -> round -> page`. It could conceivably be `/projects/{projectid}/proofrounds/{proofroundid}/...`

api/dp-openapi.yaml
```diff
@@ -591,6 +591,354 @@ paths:
...
+        404:
+          $ref: '#/components/responses/NotFound'
+        422:
+          $ref: '#/components/responses/InvalidCharacters'
+        429:
+          $ref: '#/components/responses/RateLimitExceeded'
+        default:
+          $ref: '#/components/responses/UnexpectedError'
+
+  /proofrounds/{proofroundid}/proofprojects/{proofprojectid}/ownedpages/{ownedpage}:
```
cpeel, line 752: I don't understand what `ownedpages` is vs `proofpages` - can you clarify how these differ and where this one is used?

api/dp-openapi.yaml
```diff
@@ -591,6 +591,354 @@ paths:
...
+        401:
+          $ref: '#/components/responses/Unauthorized'
+        404:
+          $ref: '#/components/responses/NotFound'
+        429:
+          $ref: '#/components/responses/RateLimitExceeded'
+        default:
+          $ref: '#/components/responses/UnexpectedError'
+
+  /proofrounds/{proofroundid}/proofprojects/{proofprojectid}/proofpages/{proofpage}/pageoptions/{pageoption}:
```
cpeel, line 792: `pageoption` should be an optional URL parameter that gets passed in (defaults to `current` for instance), not part of the URL. That lets the GET and PUT be the same URL which is more RESTful.

api/dp-openapi.yaml
70ray: > I would expect this to be coded within the client and not passed from the server.

That seems reasonable.

api/dp-openapi.yaml
70ray: > We should use snake case everywhere in here to be consistent with the existing API.

I'll do that

api/dp-openapi.yaml
70ray: It seemed more like a hierarchical address if round is put first. In the case of proofreading if you choose a round you can then choose a project which is available in that round. But doing it the other way round, if you choose a project you cannot then choose which round you want to work in because the round has to be the round the project is in.
It makes the validation work in a logical sequence: first check that the user can work in the given round and then check that the project is in the round and available.
The user is really the first component in the hierarchy although not stated explicitly so if the user is not logged in it will fall at the first fence.
Although this is not an argument either way this parallels the way the site works now in that you choose a round and are presented with a list of projects. We could make another endpoint GET `/proofrounds/{proofroundid}/proofprojects` which would get a list of projects.

api/dp-openapi.yaml
70ray: The difference is in the page state. (Assuming we are no longer using the "out" state.) A proofpage just means that it is in the "temp" state and owned by the user.
A proofpage can then be saved as temp (no state chage) or "done" (changes to a "savedpage"), returned to the round (change to "avail", no longer a "proofpage"), or marked as bad (no longer a "proofpage"). None of these require any data to be returned to the client unless something is wrong (such as bad characters or limit exceeded for saving.
I think it would really be logical to define a "savedpage" which is in "saved" state and owned by the user. The only thing that the user can do to a page in this state is to re-open it, that is convert it to a "proofpage" which changes its state to "temp" and make some other database changes. This operation returns page data to the client.
The exising code makes links to pages which can be either saved or temp so in order to avoid adding more complexity to the interface we can define an "ownedpage" which can be in either state. If we "reopen" it, the state may or may not change. (Similar to what LPage:resume_saved_page() does).

api/dp-openapi.yaml
70ray: Good, I'll do that.

api/dp-openapi.yaml
cpeel: If `ownedpages` and `proofpages` are based on states of a page then they aren't separate resources -- the resources is the page -- and should be selectors to an API rather than separate endpoints.

api/dp-openapi.yaml
```diff
@@ -591,6 +591,354 @@ paths:
         default:
           $ref: '#/components/responses/UnexpectedError'
 
+  /projects/badpagedata:
```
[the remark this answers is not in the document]
cpeel: We should think of a more generic way to handle string translation than usecase-specific URLs. Others have solved this problem so lets do some research on how they've done it rather than reinventing our own.

api/dp-openapi.yaml
70ray: We can define a resource in a more abstract way:  a "proofpage" is defined as a page checked out to a user and in "temp" state. We can create this resource by the POST checkout operation. If we return it to the round the resource ceases to exist. If we save it as "done" the "proofpage" also ceases to exist but is converted to an "ownedpage". It is easy to make validators to check whether these resources exist.

api/dp-openapi.yaml
70ray: There is some redundancy in how the system works now. The project state (which is used in the PI now) defines what round, if any, the project is in. The page state also contains round information which is redundant since it must agree with the round defined by the project state. We could use project state in this API instead of round which is how the present system works; it does not matter much.
I do not think it matters that this is different to the  API `v1/projects/:projectid/pages/:pagename/pagerounds/:pageroundid` because they are doing different things. That endpoint is for looking at data and is not concerned with the user's permissions or the project state or page state. For proofreading we have to check all of these which can be done conveniently by the validators. These validators do more checks than the validators for the above endpoint.
Doing the validation in this way means that the _Page_require() functions in DPage are not needed for proofreading operations because we have already checked that everything is as it has to be for the operation we are doing.

api/dp-openapi.yaml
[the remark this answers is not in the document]
70ray: Yes, the way we are handling translated strings in javascript now is somewhat inconvenient.

api/dp-openapi.yaml
cpeel: But nothing is stopping you from doing that enhanced validation on non-GET requests using the same URLs -- you don't need separate URLs for that.

As a user I'm looking at the new API endpoints and it seems like you're adding a lot of complexity for no discernible benefit.

70ray: Yes, that should be possible, I'll try it out. I think it would be desirable to do the enhanced validation also on GET requests; there might be a way.

api/dp-openapi.yaml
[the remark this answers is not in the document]
70ray: Something like https://stackoverflow.com/questions/431672/how-to-translate-strings-in-js might be possible. I'll try something out.

api/dp-openapi.yaml
70ray: What validations are done is determined by the names of the parts of the URL. This seems to me to be a simple way of arranging appropriate validation. I do not see how it adds any complexity.

cpeel: You're adding an entirely new set of endpoints which require new code, new documentation in the Swagger, and new sets of validators for resources that already have existing endpoints -- that's complexity.

70ray: Being able to validate based on the name of each part of the url is one of the good parts of this system so I would be reluctant to reject it. Since we have to validate anyway this seems like the simplest and best way to do it.

cpeel: But you can do that already...
```diff
diff --git a/api/v1_projects.inc b/api/v1_projects.inc
index c2dda153d..b77b74448 100644
--- a/api/v1_projects.inc
+++ b/api/v1_projects.inc
@@ -393,6 +393,26 @@ function api_v1_project_pagedetails($method, $data, $query_params)

 function api_v1_project_page_round($method, $data, $query_params)
 {
+    // handle GET and POST requests
+    if ($method == "GET") {
+        // everyone can get a page
+        return _api_v1_project_page_round_GET($data);
+    } elseif ($method == "POST") {
+        $project = $data[":projectid"];
+        $round = $data[":roundid"];
+        $page_name = $data[":pagename"];
+
+        // pages can only be modified if the project is in the correct
+        // round/state and the user has the page checked out, etc
+        if (! $project->state ... && ...) {
+            throw new UnauthorizedError("You are not authorized to change the page at this time.");
+        }
+
+        // update page
+    }
+}
+
+function _api_v1_project_page_round_GET($data) {
     if ($data[":pageroundid"] == "OCR") {
         $text_column = "master_text";
         $user_column = "NULL";
```

70ray: Although this is not the main issue I'll mention it now in case I forget: there is a case where not doing any validation before getting a page could be a problem: when the user reverts to the last save. If the project had moved to a different state he would not become aware of it.

70ray: If we used round rather than pageround for proofreading purposes then the GET case could validate also.

cpeel: > when the user reverts to the last save. If the project had moved to a different state he would not become aware of it.

This should be handled by the client. The data returned from the GET has the data needed for the client to validate this -- this isn't the responsibility of the endpoint, because it isn't an authorization issue.

70ray: You could look at it like that and I'm quite happy to go along with it but there would have to be two requests instead of one. Can we not make the api requests mean whatever we want them to? Like "Get this page for me to proofread" and the server could reply "No, this page is no longer available to you for proofreading" (though you can look at it).

cpeel: Proofreading is just one of many use-cases for the API. Right now any user can view any page text for any page in any project in any round the project has been in so far regardless of the page's save status (not checked out, checked out, returned, etc) -- that is: they are authorized to do so. Why would we change that just to satisfy one specific use-case?

We can make the API requests mean whatever we want to, but not at the expensive of reducing the functionality of the API just because it makes it easier for one specific use case.

70ray: Good, I think we can proceed on this basis. I'll revise the doc.

70ray:
This moves the validation into the project functions instead of context-free validators.
It uses round urls rather than pageround. It seems not unreasonable to use different url for proofreading page actions and other actions: 1. OCR cannot be a proofreading round. 2. The validation for non-proofreading loading page text will be entirely different to the proofreading case. I can look into this to see what would actually be required to see if it would be possible to combine them. Proofreading access will likely be used hundreds of times more than other accesses.
The order of the url components is unimportant and could be changed if preferred.
Given that the url above is used for put and post it seems reasonable to use the same for get which also makes it possible to validate the users permission for this purpose.

api/dp-openapi.yaml
```diff
@@ -591,6 +591,350 @@ paths:
...
+
+  /rounds/{roundid}:
+    get:
+      tags:
+      - project
+      description: Gets round information
+      parameters:
+      - name: roundid
+        in: path
+        description: The round identifier (as defined by array_keys($Round_for_round_id_))
```
cpeel, line 651: These descriptions should not reference PHP code. They should be useful to users without seeing the code.

api/dp-openapi.yaml
```diff
@@ -591,6 +591,350 @@ paths:
...
+        404:
+          $ref: '#/components/responses/NotFound'
+        422:
+          $ref: '#/components/responses/InvalidCharacters'
+        429:
+          $ref: '#/components/responses/RateLimitExceeded'
+        default:
+          $ref: '#/components/responses/UnexpectedError'
+
+  /rounds/{roundid}/projects/{projectid}/pages/{page}/suggestions:
```
cpeel, line 894: This should be after the one for `/rounds/{roundid}/projects/{projectid}/pages/{page}` in the file.

cpeel (commented):
Please rebase this off of `pi-redesign` and we'll retarget the PR to point to that one instead.

Overall I think this is the right direction. I think we have some fundamental disagreements on what makes an elegant API but this is your rodeo, not mine.

I suggest that you pare down the initial API here to just the endpoints that manipulate a page -- that is: check one out, make a change, return one, etc. Get that in and implement the endpoints in the code. Then build out the additional ones that a UI client would need (wordcheck, suggestions, etc). Start with small building blocks, build them, test them, and then iterate.

70ray:
> Please rebase this off of `pi-redesign` and we'll retarget the PR to point to that one instead.
> 
> Overall I think this is the right direction. I think we have some fundamental disagreements on what makes an elegant API but this is your rodeo, not mine.
> 
> I suggest that you pare down the initial API here to just the endpoints that manipulate a page -- that is: check one out, make a change, return one, etc. Get that in and implement the endpoints in the code. Then build out the additional ones that a UI client would need (wordcheck, suggestions, etc). Start with small building blocks, build them, test them, and then iterate.

Thanks Casey. I'll do what you suggest.

api/dp-openapi.yaml
70ray: If this is linked from existing code these rounds are provided by the links but in general I supose we should add an endpoint to get a list of valid rounds for proofreading in the same way as projects/pagerounds but noit including OCR.

70ray:
Rebased without any changes.

cpeel:
Thanks for the rebase and PR retarget.

70ray:
ProofPage.inc has functions similar to LPage and DPage. The differences are that checkout copies the text from previous round and sets state to temp instead of page_out which is now not used at all, and the checking is done by the ProofPage.validate function instead of Page_require().

70ray:
I've tested it with curl. It would be nice to make some automatic tests but I don't know if it is possible.

70ray:
This is strange: the last two commits checked out in my sandbox.

api/index.php
```diff
@@ -40,7 +40,10 @@ function api()
 
     $router = ApiRouter::get_router();
 
-    api_output_response($router->route($path, $query_params));
```
cpeel, line 43: Can you elaborate on this change and the related changes to all the functions in `v1_projects.inc` to take a 3rd parameter?

70ray: Away from home at the momennt but it was to enable testing. There could be a better way to do it with stub or something like that in phpunit.

70ray: I've had a look at this:

Setting a value for method: this comes from $_SERVER["REQUEST_METHOD"] so we could just set this global variable. This works but is not a good thing to do. If I understand correctly, namespaces will not help because super globals are visible in all contexts, so I think adding a parameter to route() could be the best option for this since it is a very simple change. The code already passes the value for $query_params through route().

Setting a value for api_get_request_body(). This function in index.php is not visible in the tests so we can just make a new global function for it and manipulate the value it returns by a global variable.

api/dp-openapi.yaml
```diff
@@ -893,6 +1057,40 @@ components:
...
+    proof_page:
+      type: object
+      properties:
+        pagename:
+          type: string
+        image_url:
+          type: string
+        text:
+          type: string
+
```
70ray, line 1069: It would perhaps be useful to send also the number of pages the user can save before reaching their daily page limit.

SETUP/tests/apiTest.php
```diff
@@ -0,0 +1,467 @@
...
+
+    // ----------------------------------------------------
+    // tests
+
+    public function test_checkout_invalid_round()
+    {
+        global $pguser;
+
+        $this->expectException(NotFoundError::class);
+        $this->expectExceptionMessage('Invalid round');
```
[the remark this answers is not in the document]
70ray: The code could be text or a number.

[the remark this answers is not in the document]
70ray: There are some cases where we would have to add exception codes to existing code to test in this way for every possibility. e.g. there are several different reasons why a user cannot check out a page in addition to there being no more pages available (in can_user_get_pages_in_project() in stages .inc).

SETUP/tests/phpunit_bootstrap.php
```diff
@@ -1,8 +1,16 @@
 <?php
 global $relPath;
+global $ACCESS_CRITERIA; // HACK see https://github.com/sebastianbergmann/phpunit/issues/325
```
cpeel, line 3: The better fix is to explicitly make these global in the files they are defined rather than implicitly here. This is a general problem in our code where we assume that if we define variables outside of a function they are global -- but they aren't. We've had to make these types of fixes before.

api/v1_projects.inc
```diff
@@ -410,12 +415,15 @@ function api_v1_project_page_round($method, $data, $query_params)
...
         WHERE image = '%s'
-    ", $text_column, $user_column, $data[":projectid"]->projectid,
+    ", $text_column, $user_column, $project->projectid,
         DPDatabase::escape($data[":pagename"])
     );
     $result = DPDatabase::query($sql);
     $row = mysqli_fetch_assoc($result);
-    $row["image_url"] = $data[":projectid"]->url . "/" . $row["image"];
+    if (!$row) {
+        throw new NotFoundError("No such page in project", 1006);
```
cpeel, line 424: We should either treat these strings as canonical, or the numbers as canonical, but not both. That is, if we want to use unique error codes like this we need to define them somewhere with the strings they belong to and treat them as constants in the code. We then need an API endpoint to return them to the API developer so they don't need to slog through the code to see what each one is.

api/v1_projects.inc
```diff
@@ -576,3 +584,114 @@ function api_v1_projects_holdstates($method, $data, $query_params)
...
+    switch ($new_page_state) {
+    case "saved":
+        $text = $request_data["text"] ?? "";
+        $text_checker = new TextChecker($text, $project);
+        if (!$text_checker->all_good) {
+            return $text_checker->response();
+        }
+        $page_saved = $proof_page->attempt_save_as_done($text);
+        if (!$page_saved) {
+            throw new BadRequest(_("You have exceeded your daily page limit, page saved as in progress"), 1009);
```
cpeel, line 648: By design, the API currently only returns non-translated strings and we should continue to do so. Clients are responsible for managing the UI presented to the user -- including translated strings. The reason is that the API may be used by an intermediary that is not the user with the API key. For example, PG is working on a tool that will query the API using an automated account as a proxy to the user. The user's desired language is therefore not the one associated with the API key.

api/v1_projects.inc
```diff
@@ -576,3 +584,114 @@ function api_v1_projects_holdstates($method, $data, $query_params)
...
+    }
+}
+
+function validate_for_proofing($round, $project)
+{
+    global $pguser;
+
+    if (!$round->user_access($pguser)->can_access) {
+        // The current user isn't allowed to work in the round
+        $err = sprintf(
```
cpeel, line 675: See about that strings returned by the API should not be translated / localized.

api/v1_validators.inc
```diff
@@ -35,10 +35,6 @@ function validate_wordlist($wordlist, $data)
 
 function validate_page_name($pagename, $data)
 {
-    $pages = $data[":projectid"]->get_page_names_from_db();
```
cpeel, line 38: We can't neuter this validator -- this needs to still validate page names.

pinc/Project.inc
```diff
@@ -1030,14 +1030,11 @@ class Project
...
     {
-        if (!$this->pages_table_exists) {
-            throw new NoProjectPageTable(_("Project page table does not exist."));
-        }
+        $this->validate_project_page_table();
 
         // if state is null, get all pages
         $where = $state ? sprintf("state = '%s'", DPDatabase::escape($state)) : "1";
 
-        validate_projectID($this->projectid);
```
cpeel, line 1040: We should continue to do this before we use it in the SQL rather than moving it in a new function (we can also have it in the new function if you want). Having it before the SQL where we use is a reinforcement that we are explicitly validating variables in SQL before using them. We've had unvalidated `$projectid` be used in a security incident before so this is something I'm very aware of.

SETUP/tests/phpunit_bootstrap.php
70ray: Thanks, I'll do that.

api/v1_projects.inc
70ray: Yes, I was thinking the same.

api/v1_projects.inc
70ray: ok.

api/v1_validators.inc
70ray: Would it be better to validate it by querying the database directly rather than getting all the page names and checking if it is among them? I guess it woulfd be more efficient but we already have the function to get all the page  names. (Remebering what computers were like 50 years ago I tend to be unduly concerned with doing things most efficiently).

pinc/Project.inc
70ray: ok

api/v1_validators.inc
cpeel: In this case they're probably about the same cost. The image is a unique key in project table so it'll be an index-only read and that index is probably in memory already. It would be more efficient to query to see if a specific page exists if you wanted to add such a function:
```
mysql> explain select image from projectID5e26579e08db8;
+----+-------------+------------------------+------------+-------+---------------+-------+---------+------+------+----------+-------------+
| id | select_type | table                  | partitions | type  | possible_keys | key   | key_len | ref  | rows | filtered | Extra       |
+----+-------------+------------------------+------------+-------+---------------+-------+---------+------+------+----------+-------------+
|  1 | SIMPLE      | projectID5e26579e08db8 | NULL       | index | NULL          | image | 50      | NULL |  420 |   100.00 | Using index |
+----+-------------+------------------------+------------+-------+---------------+-------+---------+------+------+----------+-------------+
1 row in set, 1 warning (0.00 sec)

mysql> explain select image from projectID5e26579e08db8 where image = '420.png';
+----+-------------+------------------------+------------+-------+---------------+-------+---------+-------+------+----------+-------------+
| id | select_type | table                  | partitions | type  | possible_keys | key   | key_len | ref   | rows | filtered | Extra       |
+----+-------------+------------------------+------------+-------+---------------+-------+---------+-------+------+----------+-------------+
|  1 | SIMPLE      | projectID5e26579e08db8 | NULL       | const | image         | image | 50      | const |    1 |   100.00 | Using index |
+----+-------------+------------------------+------------+-------+---------------+-------+---------+-------+------+----------+-------------+
1 row in set, 1 warning (0.00 sec)

```

api/v1_projects.inc
70ray: If we follow the principle that the API should return only data then in the example case of trying to checkout a page there are several possible outcomes:

- success 200 response
-  predictable failure cases where we could return a code and some data, perhaps with the 422 response
-  abnormal errors with other 4xx responses.

The client would have to be able to make a human-readable message from the code and data. This could be done either by pre-loading it with all possible error messages or by an API request to provide a message for for data the client provides, perhaps by another api channel to a possibly different host.

70ray: There has been some discussion of this topic such as https://groups.google.com/g/api-craft/c/U6JMr26Ew6k and http://soabits.blogspot.com/2013/05/error-handling-considerations-and-best.html.
If we are to provide a user experience at least as good as previously then we will have to provide enough data with the error such as:
- a simple default message to use if all else fails.
- a code for the client to use to construct a more verbose message
- any dynamic data: an array of the parameters for the sprintf used to construct the message in the present system.

cpeel: Thanks for the research! I see the problem with being able to provide localized strings with parameter replacements. If we didn't have to worry about parameter replacements this would be much easier.

Hmm...

What about if failures returned both a non-localized string (the current `error` field) and a localized string based on the user's API key? (`error_localized`?). This would present a consistent client experience for non-web-users and the ability for clients who want a localized string to get one. Thoughts?

70ray: We could make an extended exception that contained a code and an array of parameters; the existing code could interpret this so we could implement it without breaking the existing function and the api would return this data. The client would have the information to construct messages in the appropriate language

70ray: That is true, it would be rather complicated. Your proposal would better.

srjfoo: I'm sure I don't understand all the ramifications of what's being discussed, but I've been thinking about this and thought I'd ask what type of scenario we're talking about. The current three main users of the API that I can think of at this point are:
1. DP
2. bfoley's scripts that provide some statistics that DP doesn't
3. the script's rfrank is preparing for PG (won't actually be using the API until next PROD deploy, but he's been building them using TEST).

For 1: Will localization be an issue for us when the new PI is in place. I.e., will the API need to be able to localize strings for the PI, or will that be done elsewhere in the code?

For 2: As far as I know, Brian doesn't do any localization.

For 3: Again, I don't think there's any localization going on, and if there is, it would be at their end.

Consider, also, that (1) at this point, only the French translation is enabled, (2) it is intentionally incomplete in the sense that we made a conscious decision to ignore certain classes of messages (for instance, those only seen by site admins or project facilitators are not localized) and (3) maintenance of the translations is only periodic.

So, what is the problem that we're trying to solve?

cpeel: These are the strings and error messages that will be reported to the user in the new PI -- which include strings in the existing PI that we've already translated to French. So without some sort of localization for these strings we will be regressing existing functionality.

Ray's on the right track here because we need to ensure we have a way for the new PI to have translated strings before we move over to it.

srjfoo: > Ray's on the right track here because we need to ensure we have a way for the new PI to have translated strings before we move over to it.

Thanks -- that's what I was trying to understand, I think.

pinc/MessageCodes.inc
```diff
@@ -0,0 +1,43 @@
...
+    public const NO_SUCH_PAGE = 10;
+    public const WRONG_PAGE_STATE = 11;
+    public const WRONG_PAGE_OWNER = 12;
+    public const DAILY_LIMIT_EXCEEDED = 13;
+    public const REACHED_BEGINNERS_QUOTA = 14;
+    public const REACHED_PROJECT_BEGINNERS_QUOTA = 15;
+    public const NO_ACCESS_TO_BEGINERS_ONLY = 16;
+    public const RESERVED_FOR_NEW_PROOFERS = 17;
+}
+
```
70ray, line 23: Why not define all the non-translated messages in here?

pinc/MessageCodes.inc
```diff
@@ -0,0 +1,44 @@
...
+    public const REACHED_BEGINNERS_QUOTA = "Reached 'Beginners Only' quota.";
+    public const REACHED_PROJECT_BEGINNERS_QUOTA = "Reached project 'Beginners Only' quota.";
+    public const NO_ACCESS_TO_BEGINERS_ONLY = "No access to 'beginner' projects.";
+    public const RESERVED_FOR_NEW_PROOFERS = "Reserved for new proofreaders.";
+    public const NO_ACCESS_TO_MENTORS_ONLY = "No access 'mentors only' projects.";
+}
+
+// localized messages only needed for cases which can occur in normal operation
+// other cases can only happen if the client app. is faulty
+class LocalizedException extends Exception
```
cpeel, line 27: The challenge with throwing a generalized Exception (which is effectively what LocalizedException is) is that a caller can't catch a specific exception and do something with it without catching all of the generalized exceptions and then checking the message string.

While it might seem verbose to code, making each message its own exception type extending a base exception and including its custom string would be better programatically.

These seem like they logically fall into exception with either:
* project
* page
* user access

The first ones could be exceptions that extend the existing `ProjectException` base exception (indeed we already have `InvalidProjectIDException` for an invalid project ID). The second could extend from a new `ProjectPageException` base exception.  And the third extend from a new `UserAccessException` base exception (probably have these defined in `User.inc`).

70ray: I'll try that.

70ray:
I haven't forgotten about this but have been busy with other stuff. I hope to get back to it.

70ray:
New Sandbox at: https://www.pgdp.org/~rp31/c.branch/pi_swag

It is possible to test using curl:

 curl -i -X POST "https://www.pgdp.org/~rp31/c.branch/pi_swag/api/index.php?url=v1/rounds/P2/projects/$projectid" -H "Accept: application/json" -H "X-API-KEY: $key"

where $projectid is the project id and $key is your api key.

cpeel:
This is coming together great, Ray. Please break out this code into separate PRs so we can review and test them individually. It feels like we have these logical groupings:
* API test-related code (including updated cleanup code, declaring globals in `Activity.inc` & `ProjectTransition.inc`, etc) that don't relate to new API endpoints
* API changes to return a code
* Converting error-detection code to throwing Exceptions
* New API endpoints and supporting code

I think we should open these against `master` as they all seem like great functionality we want in the base. I know that some of these build on other so we might need to do them in series not parallel.

cpeel:
> The codes are now defined in project.inc but this may not be the best place. Perhaps a new file?

I think we should tie them directly to the exception and just extend our exceptions to new ones that we can assign unique numbers to.

For instance, project exceptions:
```php
class ProjectException extends Exception
{
    protected int $code = 100; // all project errors extend from here and in range 100-199
}
class NonexistentProjectException extends ProjectException
{
    protected int $code = 101;
}
class InvalidProjectIDException extends ProjectException
{
    protected int $code = 102;
}
// ...
```

Then we can create a new `PageException` with the 200-299 range, etc

PR #806, opened 2022-04-22
Enforce correct project create/update perms in API

cpeel:
@srjfoo did some very thorough testing of the new project create/update API (thanks Sharon!) and found we were not doing the right permissions enforcement. This fixes those problems.

Testable in the [update-project-create-api-perms](https://www.pgdp.org/~cpeel/c.branch/update-project-create-api-perms/) sandbox.

### DIFF
--- a/api/v1_projects.inc
+++ b/api/v1_projects.inc
@@ -142,6 +142,11 @@ function get_updatable_project_fields()
 
 function create_or_update_project($project)
 {
+    // can the user use this update API at all?
+    if (!user_is_PM()) {
+        throw new UnauthorizedError();
+    }
+
     // can the user manage the existing project?
     if ($project->projectid && !$project->can_be_managed_by_current_user) {
         throw new UnauthorizedError();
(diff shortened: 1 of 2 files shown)


api/v1_projects.inc
```diff
@@ -184,6 +192,28 @@ function create_or_update_project($project)
...
     }
 
+    // enforce specific field authorization for non-SAs
+    if (!user_is_a_sitemanager()) {
+        // project creation
+        if (!$orig_project->projectid) {
+            if (User::current_username() != $project->username) {
+                throw new UnauthorizedError("You do not have permission to create projects for another PM");
+            }
+            if (User::current_username() != 'BEGIN' && $project->difficulty == "beginner") {
```
chrismiceli, line 202: There is a user with the name "BEGIN" that is the only person who can set difficulty to beginner? I never knew this.

chrismiceli (approved):

api/v1_projects.inc
srjfoo: Actually, siteadmins should also be able to set the difficulty to beginner.

cpeel: They can. These inner checks only apply to non-SAs.

cpeel: > There is a user with the name "BEGIN" that is the only person who can set difficulty to beginner?

Yeah, this is one of several pgdp.net-specific things that is kind baked into our code and hard to extract.

PR #810, opened 2022-05-03
Allow selecting non-enabled data from endpoints

cpeel:
The API was written for the "enabled" filter to be used as a boolean flag to either return all items or just enabled items. The Swagger UI expects it to accept both a "true" and "false" value as a boolean which is confusing for users who expect "false" to return all disabled items.

Update the 3 endpoints that accept an "enabled" flag to work both ways. This enables them to work with the Swagger UI as well as the original design.

Testable in the [fix-api-enable-disable](https://www.pgdp.org/~cpeel/c.branch/fix-api-enable-disable/) sandbox as well as the "Development" endpoint in the Swagger UI.

### DIFF
--- a/api/v1_projects.inc
+++ b/api/v1_projects.inc
@@ -489,10 +489,15 @@ function api_v1_projects_pagerounds($method, $data, $query_params)
 
 function api_v1_projects_charsuites($method, $data, $query_params)
 {
-    if (isset($query_params["enabled"])) {
+    $enabled_filter = _get_enabled_filter($query_params);
+    if ($enabled_filter === null) {
+        $charsuites = CharSuites::get_all();
+    } elseif ($enabled_filter) {
         $charsuites = CharSuites::get_enabled();
     } else {
-        $charsuites = CharSuites::get_all();
+        $charsuites = array_udiff(CharSuites::get_all(), CharSuites::get_enabled(), function ($a, $b) {
+            return $a->name <=> $b->name;
+        });
     }
 
     $return_data = [];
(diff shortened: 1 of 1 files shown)


api/v1_projects.inc
```diff
@@ -576,3 +591,22 @@ function api_v1_projects_holdstates($method, $data, $query_params)
...
+// Check to see if the user requested an enabled filter. This returns one of
+// three possible states:
+//   null - no enabled flag was set
+//   true - enabled flag was set to a non-"false" value
+//   false - enabled flag was set to "false"
+function _get_enabled_filter($query_params)
+{
+    if (!isset($query_params["enabled"])) {
+        return null;
+    } elseif ($query_params["enabled"] == "false") {
```
chrismiceli, line 607: should we make it case insensitive? Stack overflow also mentions filter_var https://stackoverflow.com/questions/4775294/parsing-a-string-into-a-boolean-value-in-php

chrismiceli (approved):

api/v1_projects.inc
cpeel: Yes, making it case-insensitive makes sense here.

Oh nice, I didn't know about `filer_var()`. Let me change things just a smidge.

chrismiceli (approved):

srjfoo:
> This enables them to work with the Swagger UI as well as the original design.

Just verifying that this means that using `curl`, the only options are "all" (i.e. no parameter) and "enabled", given the syntax of the `curl` command.

srjfoo (approved):

cpeel:
> Just verifying that this means that using `curl`, the only options are "all" (i.e. no parameter) and "enabled", given the syntax of the `curl` command.

Not at all. Anything you can do in Swagger you can do with `curl`. The sentence just means these both work:
```
https://www.pgdp.org/api-dev/v1/projects/imagesources?enabled       (original design)
https://www.pgdp.org/api-dev/v1/projects/imagesources?enabled=true
```

70ray (approved):

PR #815, opened 2022-05-18
Updated API user's guide for ORed searches

cpeel:
The API supports combining project search params as ORs. This is only documented in the [Swagger docs](https://github.com/DistributedProofreaders/dproofreaders/blob/master/api/dp-openapi.yaml#L28) right now, but makes more sense in the User's Guide.

### DIFF
--- a/api/USERS_GUIDE.md
+++ b/api/USERS_GUIDE.md
@@ -109,6 +109,10 @@ available endpoints and their parameters and return values.
 
 ### Project searches
 
+Use URL parameters to search for projects against the `v1/projects` endpoint.
+To send multiple values of the same parameter to OR together, append the name
+with `[]` (e.g. `projectid[]=`). Unique parameters are ANDed together.
+
 Search for all projects with genre `Other`:
 
 ```bash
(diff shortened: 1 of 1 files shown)


srjfoo (approved):

api/USERS_GUIDE.md
```diff
@@ -109,6 +109,11 @@ available endpoints and their parameters and return values.
 
 ### Project searches
 
+Use URL parameters to search for projects against the `v1/projects` endpoint.
+If different parameters are included they are ANDed together. To send multiple
+values of the same parameter to OR together, append the name with `[]`
+(e.g. `projectid[]=`).
```
chrismiceli, line 115: we may want to clarify that the unique parameters are still ANDed, just OR among parameters of the same type.

chrismiceli (approved):

api/USERS_GUIDE.md
cpeel: I've rephrased it -- is that clearer?

chrismiceli (approved):

api/USERS_GUIDE.md
chrismiceli: Much better in my opinion. Thank you

srjfoo (approved):

api/USERS_GUIDE.md
```diff
@@ -117,6 +121,14 @@ curl -i -X GET "https://www.pgdp.org/api/v1/projects?genre=Other" \
...
 ```
 
+Search for all projects with genre `Science` or `Sports`:
+
+```bash
+curl -i -X GET "https://www.pgdp.org/api/v1/projects?genre[]=Science&genre[]=Sports" \
+    -H "Accept: application/json" \
+    -H "X-API-KEY: $API_KEY"
+```
+
```
70ray, line 131: Should we clarify it even more by giving an example of OR and AND in the same query? e.g. monster in title and genre either Science or sports.

cpeel: Sure, that gives us a chance to give a pseudocode example of the AND/ORing too. Just added that.

srjfoo (approved):

70ray (approved):

chrismiceli (approved):

PR #816, opened 2022-05-18
Install phpunit with composer

cpeel:
The Github Action `shivammathur/setup-php@v2` is timing out including `phpunit`. Move `phpunit` to a composer package instead. This provides a more consistent development & test experience and doesn't require a separate phpunit install by the user.


chrismiceli (approved):

composer.json
```diff
@@ -11,7 +11,8 @@
         "voku/portable-utf8": "^5.4"
     },
     "require-dev": {
-        "friendsofphp/php-cs-fixer": "^2.18"
+        "friendsofphp/php-cs-fixer": "^2.18",
+        "phpunit/phpunit": "9.5.0"
```
chrismiceli, line 15: Should we float the version like we do for php-cs-fixer?

srjfoo (approved):

composer.json
cpeel: Yes, good call. Done.

chrismiceli (approved):

PR #817, opened 2022-05-21
Allow client to specify fields for v1/projects results

cpeel:
Allow an API client to limit which fields are returned for project searches using the `v1/projects` endpoint. This is in preparation to support a PG use-case.

Testable with curl:
```bash
    "https://www.pgdp.org/~cpeel/c.branch/limit-fields/api/?url=v1/projects&fields=projectid,title,author"
```

### DIFF
--- a/api/USERS_GUIDE.md
+++ b/api/USERS_GUIDE.md
@@ -146,6 +146,17 @@ curl -i -X GET "https://www.pgdp.org/api/v1/projects?genre[]=Science&genre[]=Spo
     -H "X-API-KEY: $API_KEY"
 ```
 
+To limit which project fields are included, use the `field` parameter. Request
+multiple fields by appending `[]` to the parameter name and passing multiple
+of them.
+
+To return only the `projectid` and `title` fields:
+```bash
+curl -i -X GET "https://www.pgdp.org/api/v1/projects?field[]=projectid&field[]=title" \
+    -H "Accept: application/json" \
+    -H "X-API-KEY: $API_KEY"
+```
+
 ### Project details
 
 Get details about a specific project:
--- a/api/v1_projects.inc
+++ b/api/v1_projects.inc
@@ -111,10 +111,16 @@ function api_v1_projects($method, $data, $query_params)
 
     api_send_pagination_header($query_params, $total_rows, $per_page, $page);
 
+    // restrict to list of desired fields, if set
+    $return_fields = array_get($query_params, "field", null);
+    if ($return_fields && !is_array($return_fields)) {
+        $return_fields = [$return_fields];
+    }
+
     $output = [];
     while ($row = mysqli_fetch_assoc($result)) {
         $project = new Project($row);
-        $output[] = render_project_json($project);
+        $output[] = render_project_json($project, $return_fields);
     }
 
     return $output;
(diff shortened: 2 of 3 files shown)


api/USERS_GUIDE.md
```diff
@@ -125,6 +125,16 @@ curl -i -X GET "https://www.pgdp.org/api/v1/projects?title=Monster" \
     -H "X-API-KEY: $API_KEY"
 ```
 
+To limit which project fields are included, use the `fields` parameter to pass
+in a comma-separated list of fields to return.
+
+To return only the `projectid`, `title`, and `author` fields:
+```bash
+curl -i -X GET "https://www.pgdp.org/api/v1/projects?fields=projectid,title,author" \
```
chrismiceli, line 133: Based on the last MR about multi fields, should we instead do field[]=title&field[]=author etc.?

chrismiceli (approved):

srjfoo (approved):

api/v1_projects.inc
```diff
@@ -104,10 +104,16 @@ function api_v1_projects($method, $data, $query_params)
 
     api_send_pagination_header($query_params, $total_rows, $per_page, $page);
 
+    // restrict to list of desired fields, if set
+    $return_fields = array_get($query_params, "fields", null);
```
70ray, line 108: Not really relevant to this MR but I came across the null coalescing operator recently which does exactly the same as array_get so we could write `$return_fields = $query_params["fields"] ?? null;` in native php.

70ray (approved):

api/USERS_GUIDE.md
cpeel: I get where you're going with that, but I think that will be confusing with this endpoint in particular because we aren't `OR`ing the fields together in any way and they aren't part of the search criteria, for instance:
```
genre[]=Science&genre[]=Sports&field[]=title&field[]=author
```
Thoughts?

chrismiceli: I see where you are coming from. I think it would make sense to a user of the API if we just explained that all fields that accept multiple values are sent with field[] syntax. Then we need to clarify that multivalued search fields are ORed while distinct values that are relevant to search are ANDed. I think that is less confusing then figuring out how to send multi values depending on how they are used by the API.

In reality though, I've seen it done both ways and am open to either.

cpeel: That makes a lot of sense. The comma approach is nice but we can only use it for strings that might not contain commas (like a field name). I agree it makes more sense to be consistent within the API and use `[]` for multiple values. Let me get the branch changed to use that instead. Thanks for bringing it up and talking through it.

cpeel:
Updated syntax for testing:
```bash
    "https://www.pgdp.org/~cpeel/c.branch/limit-fields/api/?url=v1/projects&field[]=projectid&field[]=title&field[]=author"
```

chrismiceli:
I can't find the approve button after your last commit but consider this an approval.

srjfoo (approved):

api/USERS_GUIDE.md
```diff
@@ -146,6 +146,17 @@ curl -i -X GET "https://www.pgdp.org/api/v1/projects?genre[]=Science&genre[]=Spo
     -H "X-API-KEY: $API_KEY"
 ```
 
+To limit which project fields are included, use the `field` parameter. Request
+multiple fields by appending `[]` to the parameter name and passing multiple
+of them.
```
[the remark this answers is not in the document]
cpeel: I don't think that's necessary as the prior examples demonstrate asking for the full project. I included it in the YAML since there aren't examples there to demonstrate with.

PR #818, opened 2022-05-21
Allow SAs and PFs to search on clearance

cpeel:
Allow SAs and PFs to search on clearance. We can't allow PMs to do so without opening up the ability for value-fishing as PMs can only see clearances for their own projects. This is also in prep for an expected PG use-case.

Testable with curl and an `API_KEY` for an SA/PF user:
```bash
  "https://www.pgdp.org/~cpeel/c.branch/api-project-search-clearance/api/?url=v1/projects&clearance=yes"
```

If the `API_KEY` does not belong to an SA/PF the parameter is ignored just as other unrecognized parameters are.

### DIFF
--- a/api/v1_projects.inc
+++ b/api/v1_projects.inc
@@ -29,6 +29,13 @@ function api_v1_projects($method, $data, $query_params)
         "pages_total" => "n_pages",
     ];
 
+    // Allow SAs and PFs to search on clearance. We can't allow PMs to do so
+    // without opening up the ability for value-fishing as PMs can only see
+    // clearances for their own projects.
+    if (user_is_a_sitemanager() || user_is_proj_facilitator()) {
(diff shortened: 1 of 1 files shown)


srjfoo:
Some `clearance` fields have more than just the basic clearance OK key in them. Is there a reason this has to be a strict match? This is primarily true of the remaining gbn clearances, but there are a small handful of others that have more than one clearance associated with the project.

That's really my only question -- looks good in testing.

chrismiceli (approved):

srjfoo (approved):

70ray (approved):
Code looks ok to me.

cpeel:
> Is there a reason this has to be a strict match?

Nope, no reason at all. I've made it a substring match now.

PR #820, opened 2022-05-22
Allow pg-automation user to view clearances

cpeel:
Allow Project Gutenberg's `pg-automation` account to view clearances. Note that this is going into the `pgdp-production` branch.

Testable with:
```bash
   "https://www.pgdp.org/~cpeel/c.branch/allow-pg-automation-clearance/api/?url=v1/projects&clearance=do-not-upload-to-PG"
```

If you use a regular user's `API_KEY` this will ignore the `clearance` field and not show the field in the output. If you use the `pg-automation` user's `API_KEY` this should return projects with the desired clearance *and* show them in the output.

### DIFF
--- a/pinc/Project.inc
+++ b/pinc/Project.inc
@@ -901,6 +901,7 @@ class Project
             $this->can_be_managed_by_current_user
             || $this->PPVer_is_current_user
             || ($this->PPer_is_current_user && user_has_DU_access())
+            || User::current_username() == 'pg-automation'
         );
     }
 
--- a/api/v1_projects.inc
+++ b/api/v1_projects.inc
@@ -32,7 +32,8 @@ function api_v1_projects($method, $data, $query_params)
     // Allow SAs and PFs to search on clearance. We can't allow PMs to do so
     // without opening up the ability for value-fishing as PMs can only see
     // clearances for their own projects.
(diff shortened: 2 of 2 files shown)


pinc/Project.inc
```diff
@@ -901,6 +901,7 @@ class Project
             $this->can_be_managed_by_current_user
             || $this->PPVer_is_current_user
             || ($this->PPer_is_current_user && user_has_DU_access())
+            || User::current_username() == 'pg-automation'
```
srjfoo, line 904: Is this necessary? The plan is for this username to be used only to give PG scripts the ability to see the clearance when accessed via API. This user should not be permitted to actually manage a project.

srjfoo (approved):
API correctly provides clearance info when using the `pg-automation` API key, and ignores the clearance field altogether when the API key is for a user who doesn't have permission to search on the field (that is, in absence of any other restriction, the first 20 records of the whole table are returned in the response (which is what I would expect to happen).

pinc/Project.inc
cpeel: Yes, this controls if the current user can see the clearance line and used by the API, and the project page, to restrict the clearance from the output if the user can't. This isn't used to control managing a project (that's `Project-> can_be_managed_by_user()` further up).

PR #821, opened 2022-05-24
Correctly detect expired quiz attempts

cpeel:
When evaluating if a user has passed a quiz, correctly reject quiz passes that are too old. This bug was introduced in 8cf7ce568. Big thanks to @srjfoo for tracking down the problem and providing a super-focused repro.

Testable in the [fix-quiz-pass-detection](https://www.pgdp.org/~cpeel/c.branch/fix-quiz-pass-detection/) sandbox.

### DIFF
--- a/pinc/Quiz.inc
+++ b/pinc/Quiz.inc
@@ -83,7 +83,7 @@ class Quiz
 
             if ($use_pass_requirements && isset($this->pass_requirements['maximum_age'])) {
                 if ((time() - $attempt->date) > $this->pass_requirements['maximum_age']) {
-                    $pages_required_results[$quiz_page_id] = null;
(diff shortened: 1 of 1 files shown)


srjfoo (approved):

srjfoo:
> When evaluating if a user has passed a quiz, correctly reject quiz passes that are too old. This bug was introduced in [8cf7ce5](https://github.com/DistributedProofreaders/dproofreaders/commit/8cf7ce568a6e68e2ede44824cd5dc6f4961889c5). Big thanks to @srjfoo for tracking down the problem and providing a super-focused repro.

And big thanks to @cpeel for figuring out how to fix it!

chrismiceli (approved):

PR #824, opened 2022-06-09
Explicitly set API value types

cpeel:
Implicitly having `json_encode()` convert numbers-as-strings into numbers means titles like [1914](https://www.pgdp.net/c/project.php?id=projectID40fb078069c51) get rendered as numbers and not strings. @asylumcs discovered this when using the API. Explicitly set the value types instead.

I've used our Swagger schema -- which specifies field types -- as the authoritative source of what fields should be cast to `int`s rather than rendered as strings, which seems to be the default.

Testable with:
```bash
# project with real pg_ebook_number (should return number)
    "https://www.pgdp.org/~cpeel/c.branch/stringify-api-strings/api/?url=v1/projects/projectID4b8bc04f5dc7c"

# project with null pg_ebook_number (should return null, not 0)
    "https://www.pgdp.org/~cpeel/c.branch/stringify-api-strings/api/?url=v1/projects/projectID45c225f598e32"

    "https://www.pgdp.org/~cpeel/c.branch/stringify-api-strings/api/?url=v1/projects/projectID45c225f598e32/pages"

    "https://www.pgdp.org/~cpeel/c.branch/stringify-api-strings/api/?url=v1/projects/projectID45c225f598e32/pagedetails"

    "https://www.pgdp.org/~cpeel/c.branch/stringify-api-strings/api/?url=v1/stats/site"

    "https://www.pgdp.org/~cpeel/c.branch/stringify-api-strings/api/?url=v1/stats/site/rounds"
```

While here I correctly raised an exception if someone requests page details for a project that doesn't have a project table:
```bash
# deleted project projectID61db6d49c1394 should return the right error
    "https://www.pgdp.org/~cpeel/c.branch/stringify-api-strings/api/?url=v1/projects/projectID61db6d49c1394/pagedetails"
```

### DIFF
--- a/api/v1_projects.inc
+++ b/api/v1_projects.inc
@@ -261,6 +261,11 @@ function render_project_json($project, $return_fields = null)
 
     foreach (get_updatable_project_fields() as $api_key => $project_key) {
         $return_array[$api_key] = $project->$project_key;
+
+        // force pg_ebook_number to an int
+        if ($api_key == "pg_ebook_number" && $return_array[$api_key] != null) {
+            $return_array[$api_key] = (int)$return_array[$api_key];
+        }
     }
 
     // prevent clearance from being included for unauthorized users
@@ -280,8 +285,8 @@ function render_project_json($project, $return_fields = null)
             "post_processor" => $project->PPer,
             "post_process_verifier" => $project->PPVer,
             "state" => $project->state,
-            "pages_available" => $project->n_available_pages,
-            "pages_total" => $project->n_pages,
+            "pages_available" => (int)$project->n_available_pages,
+            "pages_total" => (int)$project->n_pages,
             "last_state_change_time" => date(DATE_ATOM, $project->modifieddate),
             "last_page_done_time" => date(DATE_ATOM, $project->t_last_page_done),
             "last_edit_time" => date(DATE_ATOM, $project->t_last_edit),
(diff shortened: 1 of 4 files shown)


srjfoo (approved):

chrismiceli (approved):

api/v1_projects.inc
```diff
@@ -280,8 +285,8 @@ function render_project_json($project, $return_fields = null)
             "post_processor" => $project->PPer,
             "post_process_verifier" => $project->PPVer,
             "state" => $project->state,
-            "pages_available" => $project->n_available_pages,
-            "pages_total" => $project->n_pages,
+            "pages_available" => (int)$project->n_available_pages,
+            "pages_total" => (int)$project->n_pages,
```
chrismiceli, line 289: So these properties at runtime are string or does the encoder coerce everything to string if possible?

cpeel: It *appears* as though the mysqli driver returns all values to PHP as strings, which in most cases is fine since PHP "does the right thing" with numeric strings most of the time except that the `json_encode()` also treats them as strings unless we cast them otherwise:
```php
<?php
$relPath = "./";
include_once($relPath."Project.inc");

$projectid = 'projectID45c225f598e32';
$sql = sprintf("
    SELECT projectid, n_pages, postednum
    FROM projects
    WHERE projectid = '%s'",
    DPDatabase::escape($projectid)
);
$row = mysqli_fetch_assoc($res);
echo json_encode($row) . "\n";
```
returns
```json
{"projectid":"projectID45c225f598e32","n_pages":"418","postednum":null}
```
despite the `n_pages` column being defined as an int:
```
| n_pages                | smallint unsigned  | NO   |     | 0        |       |
```

chrismiceli: You're right! Check the recommendation for example 3 here: https://www.php.net/manual/en/mysqli.quickstart.statements.php
That is very surprising but I guess PHP typing is relatively newer. Glad you confirmed though.

PR #826, opened 2022-07-03
Use escapeshellarg() to escape arguments

cpeel:
The most secure way to escape shell arguments is `escapeshellarg()`, not `escapeshellcmd()`.

Testable in [dont-use-escapeshellcmd](https://www.pgdp.org/~cpeel/c.branch/dont-use-escapeshellcmd/)

### DIFF
--- a/pinc/upload_file.inc
+++ b/pinc/upload_file.inc
@@ -241,7 +241,7 @@ function virus_check($file_path, $verbose)
     $av_test_result = [];
     $av_retval = 0;
 
-    $cmd = "$antivirus_executable -- " . escapeshellcmd($file_path);
+    $cmd = "$antivirus_executable -- " . escapeshellarg($file_path);
     exec($cmd, $av_test_result, $av_retval);
     // $av_retval == 0 is ok
     if ($av_retval == 1) {


chrismiceli (approved):

chrismiceli:
Do we need to revisit the other exec calls? https://github.com/DistributedProofreaders/dproofreaders/search?q=exec

cpeel:
> Do we need to revisit the other exec calls? https://github.com/DistributedProofreaders/dproofreaders/search?q=exec

It wouldn't hurt to revisit all of the `exec()` and `system()` calls and ensure we are escaping all arguments, but a quick review says that all of the args we're currently using in those are code-defined and not based on user input.

srjfoo (approved):

chrismiceli:
> > Do we need to revisit the other exec calls? https://github.com/DistributedProofreaders/dproofreaders/search?q=exec
> 
> It wouldn't hurt to revisit all of the `exec()` and `system()` calls and ensure we are escaping all arguments, but a quick review says that all of the args we're currently using in those are code-defined and not based on user input.

That was my impression as well.

PR #828, opened 2022-07-04
Updates for PHP 8

cpeel:
These are some updates that, after limited testing, appear to make the code work cleanly with PHP 8.0 and 8.1. Thankfully our code is modern enough that we don't need any major changes.

This code should work on PHP 7.4, 8.0, and 8.1. Accordingly there are two sandboxes available:
* The [updates-for-php8](https://www.pgdp.org/~cpeel/c.branch/updates-for-php8/) on TEST which is running PHP 7.4
* ~The [updates-for-php8](https://100.22.14.219/~cpeel/c.branch/updates-for-php8/) on a separate EC2 instance running PHP 8.1 (the cert does not match so you will get an SSL error in your browser)~

### DIFF
--- a/pinc/POFile.inc
+++ b/pinc/POFile.inc
@@ -176,7 +176,7 @@ class POFileIterator implements Iterator
         $this->po_filename = $po_filename;
     }
 
-    public function rewind()
+    public function rewind(): void
     {
         if ($this->file_handle) {
             fseek($this->file_handle, 0);
@@ -187,17 +187,17 @@ class POFileIterator implements Iterator
         $this->next();
     }
 
-    public function current()
+    public function current(): mixed
     {
         return $this->current_block;
     }
 
-    public function key()
+    public function key(): mixed
     {
         return $this->block_count;
     }
 
-    public function next()
+    public function next(): void
     {
         if (!$this->file_handle) {
             return;
@@ -224,7 +224,7 @@ class POFileIterator implements Iterator
         $this->file_handle = null;
     }
 
-    public function valid()
+    public function valid(): bool
     {
         return ($this->current_block != null);
     }
--- a/stats/includes/member.inc
+++ b/stats/includes/member.inc
@@ -310,12 +310,6 @@ function showForumProfile($username)
             "_new",
         ],
         [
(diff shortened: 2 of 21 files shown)


chrismiceli (approved):

chrismiceli (approved):

pinc/POFile.inc
```diff
@@ -176,7 +176,7 @@ class POFileIterator implements Iterator
         $this->po_filename = $po_filename;
     }
 
-    public function rewind()
+    public function rewind(): void
```
chrismiceli, line 179: should we start annotating all our functions?

cpeel: Perhaps as we start going through them. Because `POFileIterator` inherits from `Iterator` (which has return annotations) PHP 8.1 issues notices that the two signatures do not match which is why this is getting updated.

cpeel:
I think this is ready for review.

chrismiceli:
Thanks for setting up the EC2 instance too! Should really help with testing.

stats/includes/member.inc
```diff
@@ -309,12 +309,6 @@ function showForumProfile($username)
             "https://www.twitter.com/" . @$bb_user["twitter"],
             "_new",
         ],
-        [
-            "Google+",
-            @$bb_user["googleplus"],
```
chrismiceli, line 314: is it worth removing it from `$interested_columns` in forum_interface.inc and the bb table?

chrismiceli (approved):

stats/includes/member.inc
cpeel: My inclination is to leave it in both places until phpBB stops supporting it in 4.x. I removed it primarily because it was easier than fixing the "don't do string stuff to a Null value" for a no-longer-around social media site.

chrismiceli (approved):

cpeel:
We've deleted the php8 EC2 instance as we feel good that the changes in this PR are the bulk of what needs to change. The branch is still available for testing on TEST though.

srjfoo (approved):

PR #832, opened 2022-07-08
Fix minor PHP messages from Detectify run

cpeel:
The monthly Detectify security run against TEST did not find any new security errors (🎉) but there were many warnings in the `php_errors` log for  things that we could have caught and handled slightly better. This PR pulls all of those in together. 

In short:
* For project searches, validate that none of the search terms are arrays. This is the "feature" of PHP where any URL params *and values* that end in `[]` are converted to an array in `$_GET`. So `?language[]=English[]` comes into PHP as `language = [English[]]` rather than `language=["English"]`.
* For project language searches (which use regex) use a better regex escaping function to catch all characters.
* Fully escape special day codes in the nightly queue cronjob.
* When inserting / updating values in the `sessions` table, do so atomically. This avoids a very very rare edgecase where two HTTP requests with identical `sid`s come in. Also ensure that the `sid` value is not longer than the database column. This is largely an artifact of the Detectify run but it's an easy fix to make.

There should be zero functional changes, just backend changes.

Testable in [fix-minor-php-error-messages](https://www.pgdp.org/~cpeel/c.branch/fix-minor-php-error-messages/) sandbox.

### DIFF
--- a/crontab/onoff_special_event_queues.php
+++ b/crontab/onoff_special_event_queues.php
@@ -84,7 +84,7 @@ foreach (['open', 'close'] as $which) {
     ";
     echo $specials_query, "\n";
 
-    $res = mysqli_query(DPDatabase::get_connection(), $specials_query) or die(DPDatabase::log_error());
+    $res = DPDatabase::query($specials_query);
     $n = mysqli_num_rows($res);
     echo "
         Found $n special events which '$which' now.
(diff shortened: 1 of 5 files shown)


chrismiceli (approved):

chrismiceli:
Awesome improvements.

srjfoo (approved):

PR #834, opened 2022-07-10
Upgrade portable-utf8 to 6.x

cpeel:
Upgrade portable-utf8 to 6.x to keep us up-to-date. The only major change with 6.0.0 is that it removed deprecated aliases for many functions. The only one we use is `split()` -> `str_split()`. See its full [changelog](https://github.com/voku/portable-utf8/blob/master/CHANGELOG.md) for details.

I put this up as a separate PR since it is a major component of our Unicode functionality and seemed to warrant the option of a more reasonable test pass.

Testable in the [upgrade-portable-utf8](https://www.pgdp.org/~cpeel/c.branch/upgrade-portable-utf8/) sandbox.

Note that the `charsuites.php` page throws an error for a missing include. This is currently a problem on `master` but not `pgdp-production` (and thus PROD and TEST) so the `project.inc` file is getting pulled in from somewhere else. This is fixed on the PHP8 branch so it would probably best if we prioritized https://github.com/DistributedProofreaders/dproofreaders/pull/828 and I can then rebase this branch and sandbox.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
         "ezyang/htmlpurifier" : "v4.*",
         "erusev/parsedown": "1.8.0-beta-7",
         "erusev/parsedown-extra": "0.8.0",
-        "voku/portable-utf8": "^5.4"
+        "voku/portable-utf8": "^6.0"
     },
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^2.18",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "37c5becbb8ed2d5c292bfa0ca2b082b6",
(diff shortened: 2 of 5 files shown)


chrismiceli (approved):
I checked the source for the older version to confirm, they are the same:
https://github.com/voku/portable-utf8/blob/5.4.51/src/voku/helper/UTF8.php#L5847

cpeel:
FWIW, portable-utf8 appears to be ready for PHP 8.2 as well after https://github.com/voku/portable-utf8/pull/181 merges in, so getting us up to 6.x will make it seamless to pick up that change down the line.

cpeel:
> Note that the `charsuites.php page` throws an error for a missing include.

This has been fixed with https://github.com/DistributedProofreaders/dproofreaders/pull/837 so this is ready for testing.

https://github.com/voku/portable-utf8/pull/181 has also been merged in and 6.0.5 released so I've updated this accordingly. This means that after the PHP8 PR merges in we should be good to go for PHP 8.2.

srjfoo (approved):

PR #835, opened 2022-07-10
mysqli improvements

cpeel:
Finish the long slog of moving `mysqli_query()` calls to`DPDatabase::query()`. This allows us to manage query failures in
a centralized place. Also change `dpsql_query()` to `DPDatabase::query()` where reasonable.

The above allows us to to use `mysqli_report()` to raise an exception on DB error rather than silently failing. Since we use `DPDatabase::query()` everywhere we can catch the exceptions from queries and handle them the same way we did before but we get better error detection with other `mysqli_*` functions. The new error level match the default in PHP 8.1.

These updates are largely straightforward but they touch a lot of code, unfortunately. None of these updates are urgent.

Testable in the [mysqli_improvements](https://www.pgdp.org/~cpeel/c.branch/mysqli_improvements/) sandbox.

### DIFF
--- a/pinc/DPDatabase.inc
+++ b/pinc/DPDatabase.inc
@@ -21,15 +21,18 @@ final class DPDatabase
         include('udb_user.php');
         self::$_db_name = $db_name;
 
-        // Ignore warnings from mysqli_connect() so we can handle them
-        // gracefully elsewhere.
-        self::$_connection = @mysqli_connect($db_server, $db_user, $db_password);
+        // Throw exceptions for DB call failures (PHP 8.1 default values)
+        mysqli_report(MYSQLI_REPORT_ERROR | MYSQLI_REPORT_STRICT);
 
-        if (!self::$_connection) {
+        try {
+            self::$_connection = mysqli_connect($db_server, $db_user, $db_password);
+        } catch (mysqli_sql_exception $e) {
             throw new DBConnectionError("Unable to connect to database");
         }
 
-        if (!mysqli_select_db(self::$_connection, $db_name)) {
+        try {
+            mysqli_select_db(self::$_connection, $db_name);
+        } catch (mysqli_sql_exception $e) {
             throw new DBConnectionError("Unable to locate database.");
         }
 
@@ -39,7 +42,9 @@ final class DPDatabase
         // We disable it here at the session level rather than in my.cnf
         // to not impact other database users (forum, wiki, etc).
         $sql = "SET SESSION sql_mode=(SELECT REPLACE(@@sql_mode,'ONLY_FULL_GROUP_BY',''));";
-        if (!mysqli_query(self::$_connection, $sql)) {
+        try {
+            mysqli_query(self::$_connection, $sql);
+        } catch (mysqli_sql_exception $e) {
             throw new DBConnectionError("Unable to set sql_mode");
         }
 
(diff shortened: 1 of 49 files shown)


pinc/DPDatabase.inc
```diff
@@ -90,7 +96,7 @@ final class DPDatabase
 
     public static function affected_rows()
     {
-        return  mysqli_affected_rows(self::$_connection);
+        return mysqli_affected_rows(self::$_connection);
```
chrismiceli, line 99: the formatter didn't catch this before?

stats/includes/team.inc
```diff
@@ -305,7 +302,7 @@ function showTeamMbrs($curTeam, $tally_name)
     $req_team_id = get_integer_param($_GET, 'tid', null, 0, null);
 
     $subtitle = null;
-    $latestMbr = mysqli_query(DPDatabase::get_connection(), "SELECT users.username AS Username, users.u_privacy AS Privacy FROM user_teams INNER JOIN users ON user_teams.latestUser = users.u_id WHERE user_teams.id = ".$curTeam['id']." AND user_teams.latestUser <> 0");
+    $latestMbr = DPDatabase::query("SELECT users.username AS Username, users.u_privacy AS Privacy FROM user_teams INNER JOIN users ON user_teams.latestUser = users.u_id WHERE user_teams.id = ".$curTeam['id']." AND user_teams.latestUser <> 0");
```
chrismiceli, line 305: curTeam['id'] is an integer and doesn't need encoding, right? Should we use sprintf %d here to be safer?

stats/includes/team.inc
```diff
@@ -573,7 +570,7 @@ function uploadImages($preview, $tid, $type)
             $upload_avatar_dir = "$team_avatars_dir/$avatarID";
             move_uploaded_file($_FILES['teamavatar']['tmp_name'], $upload_avatar_dir);
             if ($preview != 1) {
-                mysqli_query(DPDatabase::get_connection(), "UPDATE user_teams SET avatar='$avatarID' WHERE id = $tid") or die(DPDatabase::log_error());
+                DPDatabase::query("UPDATE user_teams SET avatar='$avatarID' WHERE id = $tid");
```
chrismiceli, line 573: same here for avatarID and iconID and tid on 592.

chrismiceli (approved):

70ray (approved):

pinc/DPDatabase.inc
cpeel: Shocking, no.

stats/includes/team.inc
cpeel: It is. It should have `sprintf()` done on it but I opted to not do that lift on the team or member code which is abysmal. 😦

stats/includes/team.inc
cpeel: Same - these are all integers and enforced upstream and should be using `sprint()` but I'm not up for the lift to modernize this code.

chrismiceli: I completely understand.

cpeel:
Rebased to pick up the lessc changes to pass static checks.

chrismiceli (approved):

pinc/dpsql.inc
```diff
@@ -24,11 +24,7 @@ function dpsql_dump_themed_query($query, $start_at = 0, $show_rank = DPSQL_NO_RA
 
 function dpsql_query($query)
```
70ray, line 25: I was wondering why you did it like this but I see mysqli_query is used in noncvs.

70ray (approved):

srjfoo (approved):

srjfoo (approved):

PR #836, opened 2022-07-13
Node dependency updates

chrismiceli:
Just updating versions. Sandbox for font change
https://www.pgdp.org/~chrismiceli/c.branch/node-dependency-updates/

### DIFF
--- a/SETUP/generate_css_from_less.sh
+++ b/SETUP/generate_css_from_less.sh
@@ -39,11 +39,11 @@ function run_less {
     css_file="${less_file%.*}.css"
     if [ $CHECK -eq 0 ]; then
         echo "Generating $css_file"
-        eval "$LESSC_PATH/lessc $less_file $css_file"
+        $LESSC_PATH/lessc $less_file $css_file
     else
         echo "Validating $css_file"
         new_css_file="${css_file}.new"
-        eval "$LESSC_PATH/lessc $less_file $new_css_file"
+        $LESSC_PATH/lessc $less_file $new_css_file
         diff -q $css_file $new_css_file > /dev/null 2>&1
         if [ $? -ne 0 ]; then
             echo "ERROR: $css_file doesn't match generated less file"
--- a/SETUP/generate_css_from_less.sh
+++ b/SETUP/generate_css_from_less.sh
@@ -21,9 +21,6 @@ else
     exit 1
 fi
 
-# Ensure npm-installed lessc is on the path
-PATH=$PATH:$STYLE_DIR/../node_modules/.bin
-
 # Array of .less files that should not turn into a .css file
 NON_CSS_LESS_FILES=("layout.less" "global.less" "page_interfaces.less")
 
@@ -40,11 +37,11 @@ function run_less {
     css_file="${less_file%.*}.css"
     if [ $CHECK -eq 0 ]; then
         echo "Generating $css_file"
-        lessc $less_file $css_file
+        ../node_modules/.bin/lessc $less_file $css_file
     else
         echo "Validating $css_file"
         new_css_file="${css_file}.new"
-        lessc $less_file $new_css_file
+        ../node_modules/.bin/lessc $less_file $new_css_file
         diff -q $css_file $new_css_file > /dev/null 2>&1
         if [ $? -ne 0 ]; then
             echo "ERROR: $css_file doesn't match generated less file"
--- a/SETUP/generate_css_from_less.sh
+++ b/SETUP/generate_css_from_less.sh
@@ -21,6 +21,8 @@ else
     exit 1
 fi
 
+LESSC_PATH=$STYLE_DIR/../node_modules/.bin
+
 # Array of .less files that should not turn into a .css file
 NON_CSS_LESS_FILES=("layout.less" "global.less" "page_interfaces.less")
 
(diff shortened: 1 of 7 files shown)


SETUP/generate_css_from_less.sh
```diff
@@ -40,11 +37,11 @@ function run_less {
     css_file="${less_file%.*}.css"
     if [ $CHECK -eq 0 ]; then
         echo "Generating $css_file"
-        lessc $less_file $css_file
+        ../node_modules/.bin/lessc $less_file $css_file
```
chrismiceli, line 40: despite the path, my system was still picking the system on instead of the node one.

styles/themes/charcoal.css
```diff
@@ -146,7 +146,7 @@
   font-style: normal;
   font-weight: 700;
   font-display: swap;
-  src: url(https://fonts.gstatic.com/s/ebgaramond/v25/SlGDmQSNjdsmc35JDF1K5E55YMjF_7DPuGi-DPNUAw.ttf) format('truetype');
+  src: url(https://fonts.gstatic.com/s/ebgaramond/v26/SlGDmQSNjdsmc35JDF1K5E55YMjF_7DPuGi-DPNUAw.ttf) format('truetype');
```
chrismiceli, line 149: this was failing without even updating less. Seems the url in our less file resolves to a new version now.

chrismiceli: I guess this could impact production, let me sandbox this.

cpeel: This periodically happens and it's a bit frustrating. Appreciate you fixing it.

SETUP/generate_css_from_less.sh
cpeel, line 40: This is going to prevent running the script from an arbitrary directory. Consider using `$STYLE_DIR/../node_modules/.bin/lessc`.

cpeel (commented):
Thanks for rev'ing these!

package.json
```diff
@@ -23,9 +23,9 @@
   },
   "homepage": "https://github.com/DistributedProofreaders/dproofreaders",
   "devDependencies": {
-    "eslint": "^7.32.0",
-    "geckodriver": "^2.0.4",
-    "lessc": "^1.0.2",
-    "selenium-webdriver": "^4.0.0-rc-1"
+    "eslint": "^8.19.0",
+    "geckodriver": "^3.0.2",
+    "less": "^4.1.3",
```
srjfoo, line 28: Has the script been renamed to `less` instead of `lessc`?

chrismiceli: No, the script is still called lessc. It seems the lessc package in npmjs is really old, unmaintained, with low download activity:
https://www.npmjs.com/package/lessc
Where if you read the docs on less's website:
https://lesscss.org/usage/
it says to use the less package which is maintained and has tons of download activity:
https://www.npmjs.com/package/less

I was curious so I looked at the package contents of the lessc artifact (since there is no linked source in npmjs) and it is just a warpper around an older version of less:
```
{
  "name": "lessc",
  "version": "1.0.2",
  "description": "Run lessc via npx",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "techknowlogick",
  "license": "ISC",
  "dependencies": {
    "less": "^3.9.0"
  }
}
```
So really my change just removes an unused middleman and updates the version to the latest. I am curious why we ever went with lessc at all. Perhaps it was recommended at some point? @cpeel seems to have implemented it with https://github.com/DistributedProofreaders/dproofreaders/commit/a8f179bd8d23954fd27f8b994cc5512ee6c02444.

SETUP/generate_css_from_less.sh
chrismiceli: let me take a look at what was going on with my path then.

cpeel: I think your overall "lets use the one in `node_modules` is the right one. The PATH update on line 25 prioritized the system version which is not ideal. I just think we need to use a more flexible way to specify the `node_modules` lessc.

package.json
chrismiceli: :D, I believe you nodejs knowledge is probably 10x my PHP knowledge, but we all try and that's what matters.

SETUP/generate_css_from_less.sh
```diff
@@ -40,11 +39,11 @@ function run_less {
     css_file="${less_file%.*}.css"
     if [ $CHECK -eq 0 ]; then
         echo "Generating $css_file"
-        lessc $less_file $css_file
+        eval "$LESSC_PATH/lessc $less_file $css_file"
```
chrismiceli, line 42: using eval per https://www.shellcheck.net/wiki/SC2091. I am not great at bash.

chrismiceli: Got it thanks! One day I'll convert it all to powershell ;) /s.

cpeel:
Sorry for the delay, I somehow missed the last commit addressing my last bash suggestion. Thanks for this update!

chrismiceli:
> Sorry for the delay, I somehow missed the last commit addressing my last bash suggestion. Thanks for this update!

No worries. It wasn't critical. Let me know if there is something else you want me to look at.

PR #839, opened 2022-09-06
Prepare for the R202209 release

cpeel:
Prepare for our usual September release. This won't be merged until just before we're ready to cut the release.

The full list of commits can be seen with:
```bash
git log R202202..master
```

PR #840, opened 2022-09-09
Create new Math symbols character suite

srjfoo:
Create new character suite containing commonly used mathematical symbols that are not in Basic Latin.

The Math symbols character suite can be viewed here: [math-symbols](https://www.pgdp.org/~srjfoo/c/tools/charsuites.php?charsuite=math-symbols) and tested in this project: [Foundations of & wireless](https://www.pgdp.org/~srjfoo/c/project.php?id=projectID5c13a64918b30)

### DIFF
--- a/pinc/charsuite-math-symbols.inc
+++ b/pinc/charsuite-math-symbols.inc
@@ -1,0 +1,46 @@
+<?php
+include_once($relPath."CharSuites.inc");
+
+use voku\helper\UTF8;
+
+$charsuite = new CharSuite("math-symbols", _("Math symbols"));
+
+$charsuite->description = _("Commonly used mathematical symbols not in Basic Latin");
+
+$charsuite->codepoints = [
+    'U+2202-U+2203',
+    'U+2207-U+2209',
+    'U+221a',
+    'U+221e',
+    'U+2220',
+    'U+2229-U+222b',
+    'U+2234-U+2235',
+    'U+2237',
+    'U+2248',
+    'U+2260-U+2261',
+    'U+2264-U+2265',
+    'U+2282-U+2283',
+    'U+2295',
+    'U+2297',
+];
+
+$charsuite->reference_urls = [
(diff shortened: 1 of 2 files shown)


srjfoo:
@cpeel -- I realized after I pushed the branch up that I still have the comment in place for the reference URL -- I never did get the question answered whether the wiki page [Essential Maths for Proofers](https://www.pgdp.net/wiki/DP_Official_Documentation:Proofreading/Essential_Maths_for_Proofers) should be used. I can go ahead and put it in, or just remove that comment altogether, but wanted to go ahead and get this up for review so that we can maybe get it in for Tuesday's deployment.

pinc/charsuite-math-symbols.inc
```diff
@@ -0,0 +1,48 @@
...
+    'U+2248',
+    'U+2260-U+2261',
+    'U+2264-U+2265',
+    'U+2282-U+2283',
+    'U+2295',
+    'U+2297',
+];
+
+/*
+$charsuite->reference_urls = [
```
cpeel, line 28: I'm not sure https://www.pgdp.net/wiki/DP_Official_Documentation:Proofreading/Essential_Maths_for_Proofers is the best reference. I'd almost rather we create a similar page to https://www.pgdp.net/wiki/Symbols_collection (which could link to the Essential Maths for Proofers page) that talks about the character suite a bit more specifically.

cpeel (commented):
Note to self: update https://github.com/DistributedProofreaders/dproofreaders/pull/839 after this merges in.

pinc/charsuite-math-symbols.inc
srjfoo: I can do that for this PR and just put an "Under construction" message in it for a couple of days, if that works?

cpeel: That works perfectly 👍

cpeel (approved):

chrismiceli (approved):

PR #841, opened 2022-09-24
Streamline project listing filter interface

cpeel:
While working on some unrelated potential updates to the round pages I saw a chance to improve the way we present project filter information. This update moves the current filter into the filter table along side the button to reset it. If there is no filter the filter isn't shown.

Compare:
* [original P1 page](https://www.pgdp.org/c/tools/proofers/round.php?round_id=P1#filter_form)
* [update P1 page](https://www.pgdp.org/~cpeel/c.branch/update-filter-box/tools/proofers/round.php?round_id=P1#filter_form)

Note that the filter form does not show up unless you've [proofread 20 pages](https://github.com/DistributedProofreaders/dproofreaders/blob/master/tools/proofers/round.php#L121), so if you don't see a filter selection box, go proofread 20 pages.

### DIFF
--- a/pinc/filter_project_list.inc
+++ b/pinc/filter_project_list.inc
@@ -151,23 +151,27 @@ class DifficultyElement extends ProjectFilterElement
 {
     public function echo_active_html_control()
     {
-        echo "<p><b>$this->label:</b> ";
-        echo "<label for='diff-all'>" . pgettext("all difficulties", "All") . "</label><input type='checkbox' name='difficulty[]' id='diff-all' value=''";
+        echo "<div class='flex_row'>";
+        echo "<div class='entry'><b>$this->label:</b></div>";
+        echo "<div class='entry'><input type='checkbox' name='difficulty[]' id='diff-all' value=''";
         $difficulty = array_get($this->data, "difficulty", []);
         if (!count($difficulty)) {
             echo " checked";
         }
-        echo ">\n";
+        echo "> <label for='diff-all'>" . pgettext("all difficulties", "All") . "</label>\n";
+        echo "</div>\n";
         foreach (get_project_difficulties() as $value => $option) {
             $escaped_option = html_safe($option);
-            echo "&nbsp;<label for='$value'>$escaped_option</label><input type='checkbox' name='difficulty[]' class='diff-opt' id='$value' value='$value'";
+            echo "<div class='entry'>";
+            echo "<input type='checkbox' name='difficulty[]' class='diff-opt' id='$value' value='$value'";
             if (is_array($difficulty) && in_array($value, $difficulty)) {
                 echo " checked";
                 $this->selected_options[] = $option;
             }
-            echo ">\n";
+            echo "> <label for='$value'>$escaped_option</label>\n";
+            echo "</div>";
         }
-        echo "</p>\n";
+        echo "</div>\n";
     }
 }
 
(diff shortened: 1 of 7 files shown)


srjfoo (approved):
I definitely like the updated look, and it makes a lot more sense to have the list of filters inside the filter box instead of below it and seemingly unconnected to it.

pinc/filter_project_list.inc
```diff
@@ -151,23 +151,27 @@ class DifficultyElement extends ProjectFilterElement
 {
     public function echo_active_html_control()
     {
-        echo "<p><b>$this->label:</b> ";
-        echo "<label for='diff-all'>" . pgettext("all difficulties", "All") . "</label><input type='checkbox' name='difficulty[]' id='diff-all' value=''";
+        echo "<div class='flex_row'>";
+        echo "<div class='entry'><b>$this->label:</b></div>";
```
chrismiceli, line 155: We may want to html encode this to cover any cases where it needs encoding now or in the future or in a different language

pinc/filter_project_list.inc
```diff
@@ -241,10 +245,7 @@ function process_and_display_project_filter_form($username, $filter_type, $filte
 
     $difficulty_element->echo_html_control();
 
-    echo "<input type='submit' name='reset' value='", attr_safe(_("Remove Filter")), "'>";
-    echo "&nbsp;<input type='submit' name='apply' value='", attr_safe(_("Apply Selections")), "'>";
-    echo "</fieldset>";
-    echo "</form>\n";
+    echo "<input type='submit' name='apply' value='", attr_safe(_("Apply Selections")), "'>";
```
chrismiceli, line 248: Not important, but I learned today that you can use `,` to do multiple echos as mentioned here https://stackoverflow.com/questions/1466408/difference-between-period-and-comma-when-concatenating-with-echo-versus-return

pinc/filter_project_list.inc
```diff
@@ -151,23 +151,27 @@ class DifficultyElement extends ProjectFilterElement
...
-        echo "<label for='diff-all'>" . pgettext("all difficulties", "All") . "</label><input type='checkbox' name='difficulty[]' id='diff-all' value=''";
+        echo "<div class='flex_row'>";
+        echo "<div class='entry'><b>$this->label:</b></div>";
+        echo "<div class='entry'><input type='checkbox' name='difficulty[]' id='diff-all' value=''";
         $difficulty = array_get($this->data, "difficulty", []);
         if (!count($difficulty)) {
             echo " checked";
         }
-        echo ">\n";
+        echo "> <label for='diff-all'>" . pgettext("all difficulties", "All") . "</label>\n";
```
chrismiceli, line 161: less likely, but may want to html encode here too.

chrismiceli (approved):

pinc/filter_project_list.inc
cpeel: Good eye - I fixed this, the one below, and a few other existing instances in the file.

pinc/filter_project_list.inc
cpeel: I only recently discovered that myself after @70ray started using them. Most of the code does concatenation and we've got a bit of a mixed bag going on.

chrismiceli (approved):

srjfoo (approved):

PR #845, opened 2022-10-11
Enable bypassing error log on DB query failure

cpeel:
There are rare times when we know a query might fail and we don't want to log an error to the error log. Create a temporary table is one of them because we can't otherwise determine if the temporary table exists as it isn't populated in the schema and we don't want to give elevated privs to run sys.table_exists().

This is testable in the [fix-temp-table-logging-message](https://www.pgdp.org/~cpeel/c.branch/fix-temp-table-logging-message/) sandbox. Success is loading the round pages without generating errors in `php_errors`.

### DIFF
--- a/pinc/DPDatabase.inc
+++ b/pinc/DPDatabase.inc
@@ -82,13 +82,17 @@ final class DPDatabase
         return mysqli_real_escape_string(self::$_connection, $value);
     }
 
-    public static function query($sql, $throw_on_failure = true)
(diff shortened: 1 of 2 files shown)


srjfoo (approved):

cpeel:
I want to do a minor update to make this slightly cleaner, standby.

cpeel:
Code and sandbox updated. @srjfoo can you please do a quick re-validation?

srjfoo (approved):

PR #846, opened 2022-10-15
Fix Author page & XML rendering

cpeel:
If there is no author bio, there is no `$row["bio_id"];` to access and we get a warning of accessing an array offset on a null field, as seen in `php_errors` on PROD. Fix this by simplifying how we output bios on the author page. I did a bit more minor cleanup in `authors.inc` while I was here, none of it functional.

The other functional fix in this PR is making the `authorsxml.php` page render valid XML. Previously we were not closing the `<date>` tag making the document invalid.

Testable in the [fix-author-rendering](https://www.pgdp.org/~cpeel/c.branch/fix-author-rendering/) sandbox.

### DIFF
--- a/tools/authors/authors.inc
+++ b/tools/authors/authors.inc
@@ -12,11 +12,13 @@ $months = [_('Unknown'),
 // and a link to the xml data will be provided.
 function echo_author($last_name, $other_names, $born, $dead, $author_id)
 {
+    global $code_url;
+
     $name = $last_name . ($other_names != '' ? ", $other_names" : '');
 
     echo "<h2>" . html_safe($name);
     if ($author_id) {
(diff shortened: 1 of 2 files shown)


srjfoo (approved):
In testing, I noticed that the pophelp really should be updated -- I'm guessing that should be done in a separate PR, as it's unrelated to this fix.

chrismiceli (approved):
Do we still actively use authors etc.?

srjfoo:
> Do we still actively use authors etc.?

Every once in a while, a PM wanders by and does. That's what triggered the PHP error, was a PM adding a new author and bio. (For the first time since Markdown was enabled by PR #530.)

chrismiceli:
> > Do we still actively use authors etc.?
> 
> Every once in a while, a PM wanders by and does. That's what triggered the PHP error, was a PM adding a new author and bio. (For the first time since Markdown was enabled by PR #530.)

Hey that code looks familiar :)

PR #848, opened 2022-10-31
encode ampersand in two stages

70ray:
to avoid problem with & in upper case small cap markup, add comments and tests.
This addresses issue https://www.pgdp.net/c/tasks.php?action=show&task_id=2040

### DIFF
--- a/tools/proofers/preview.js
+++ b/tools/proofers/preview.js
@@ -856,7 +856,8 @@ $(function () {
             let smallCapRegex = new RegExp(sc1 + "([^]+?)" + sc2, 'g');
 
             function boxHtml(txt) {
-                return txt.replace(/┌/g, "&lt;").replace(/┐/g, "&gt;");
+                return txt.replace(/┌/g, "&lt;").replace(/┐/g, "&gt;")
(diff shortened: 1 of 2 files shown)


tools/proofers/preview.js
```diff
@@ -856,7 +856,8 @@ $(function () {
             let smallCapRegex = new RegExp(sc1 + "([^]+?)" + sc2, 'g');
 
             function boxHtml(txt) {
-                return txt.replace(/┌/g, "&lt;").replace(/┐/g, "&gt;");
+                return txt.replace(/┌/g, "&lt;").replace(/┐/g, "&gt;")
```
chrismiceli, line 859: Any other characters we need to worry about? Just making sure our solution can scale.

70ray:
Sandbox at https://www.pgdp.org/~rp31/c.branch/amp_fix

srjfoo (approved):

tools/proofers/preview.js
70ray: I don't think there are.

chrismiceli (approved):

PR #849, opened 2022-11-01
Add small image check to project load

chrismiceli:
Per request, adding the logic to project load too similar to large image checks.

https://www.pgdp.org/~chrismiceli/c.branch/small-images/

### DIFF
--- a/pinc/image_check.inc
+++ b/pinc/image_check.inc
@@ -1,6 +1,7 @@
 <?php
 
 $page_image_size_limit = 100; // kb
+$page_image_minimum_size = 1000; // pixels
 
 function get_image_size_error($image_size)
 {
@@ -9,6 +10,17 @@ function get_image_size_error($image_size)
 
     if ($image_size > ($page_image_size_limit * 1024)) {
         $result = sprintf(_("Image > %dkb: %dkb"), $page_image_size_limit, $image_size / 1024);
+    }
+    return $result;
+}
+
+function get_image_size_small_error($image_dimensions)
+{
+    global $page_image_minimum_size;
+    $result = null;
+
+    if ($image_dimensions[0] < $page_image_minimum_size || $image_dimensions[1] < $page_image_minimum_size) {
+        $result = _("The image may be too small: the shortest dimension of the images is less than $page_image_minimum_size pixels.");
     }
     return $result;
 }
--- a/pinc/image_check.inc
+++ b/pinc/image_check.inc
@@ -1,7 +1,7 @@
 <?php
 
 $page_image_size_limit = 100; // kb
-$page_image_minimum_dimension = 5000; // pixels
+$page_image_minimum_dimension = 1000; // pixels
 
 function get_image_size_error($image_size)
 {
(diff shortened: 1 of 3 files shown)


pinc/image_check.inc
```diff
@@ -12,3 +13,14 @@ function get_image_size_error($image_size)
...
     return $result;
 }
+
+function get_image_size_small_error($image_dimensions)
+{
+    global $page_image_minimum_size;
+    $result = null;
+
+    if ($image_dimensions[0] < $page_image_minimum_size || $image_dimensions[1] < $page_image_minimum_size) {
+        $result = _("The image may be too small: the shortest dimension of the images is less than $page_image_minimum_size pixels.");
```
cpeel, line 23: To properly translate this string, it needs to use `sprintf()` for the variable.

Ditto the other instance in `project_quick_check.inc`.

pinc/image_check.inc
```diff
@@ -12,3 +13,14 @@ function get_image_size_error($image_size)
     }
     return $result;
 }
+
+function get_image_size_small_error($image_dimensions)
```
cpeel, line 17: I'd recommend this function take `$width` and `$height`, that way it isn't implicitly relying on the output of `getimagesize()`. This way we keep the opaqueness of the data structure immediately adjacent to where it comes from.

Also consider renaming this `get_image_small_dimension_error()` since "size" is used as file size elsewhere in this function. Ditto `$page_image_minimum_size` as `$page_image_minimum_dimension`.

cpeel (changes requested):

pinc/image_check.inc
chrismiceli: Great suggestions.

pinc/project_quick_check.inc
```diff
@@ -191,6 +191,8 @@ function _test_project_for_large_page_images($projectid)
 
 function _test_project_for_corrupt_pngs($projectid)
 {
+    global $page_image_size_limit;
```
cpeel, line 194: Doesn't look like this is used in this function?

cpeel (commented):
Sans extra global the code LGTM but I haven't tested it myself.

pinc/image_check.inc
```diff
@@ -1,6 +1,7 @@
 <?php
 
 $page_image_size_limit = 100; // kb
+$page_image_minimum_dimension = 5000; // pixels
```
srjfoo, line 4: \<cough> ... 1000

pinc/image_check.inc
```diff
@@ -12,3 +13,14 @@ function get_image_size_error($image_size)
...
     return $result;
 }
+
+function get_image_small_dimension_error($width, $height)
+{
+    global $page_image_minimum_dimension;
+    $result = null;
+
+    if ($width < $page_image_minimum_dimension || $height < $page_image_minimum_dimension) {
+        $result = sprintf(_("The image may be too small: the shortest dimension of the images is less than %d pixels."), $page_image_minimum_dimension);
```
srjfoo, line 23: I'd recommend compressing a bit to make the message slightly shorter; maybe: `"Image may be too small: shortest dimension is less than %d pixels."`

srjfoo (changes requested):

pinc/image_check.inc
chrismiceli: haha, I was testing the error messages and the project I had didn't have any small images so I had to bump the number :).

srjfoo (approved):

70ray (approved):

pinc/project_quick_check.inc
cpeel: @chrismiceli - this change looks unnecessary.

chrismiceli: Sorry I forgot to address this and somehow missed it when reviewing. Taken care of now.

PR #852, opened 2023-01-16
Adjust PPV rating algorithm

windymilla:
Instead of Some/Significant use Basic/Average/Complex Treat recording of illos the same as other features Require 3, not 4, complex features for project to be "Difficult" Round permitted number of errors to nearest integer, not downwards Reduce size-per-error values for Very Good rating by 20%

Also minor form changes:
Remove side bar
Remove spaces from br examples
Update the expected title format

### DIFF
--- a/tools/post_proofers/ppv_report.php
+++ b/tools/post_proofers/ppv_report.php
@@ -20,7 +20,7 @@ require_login();
 
 $page_title = _('Post-Processing Verification Reporting');
 
-output_header($page_title);
+output_header($page_title, NO_STATSBAR);
 
 echo "\n<h1>$page_title</h1>\n";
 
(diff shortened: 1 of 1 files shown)


cpeel:
Thanks @windymilla! The dproofreaders codebases uses a prescribed coding style that, helpfully, can be applied automatically. See https://github.com/DistributedProofreaders/dproofreaders/blob/master/SETUP/CODE_STYLE.md for more details.

tools/post_proofers/ppv_report.php
```diff
@@ -186,13 +186,30 @@ function maybe_report_form_problem($message)
...
+
         $problem = "";
+        $basic_id = "bsc_" . $feature_type;
+        $average_id = "avg_" . $feature_type;
+        $complex_id = "cpx_" . $feature_type;
+        $final_label = $feature_label_map[$feature_type];
         if ($action == HANDLE_ENTRY_FORM_SUBMISSION) {
-            if (isset($_POST[$some_id]) && isset($_POST[$sig_id])) {
-                $problem = _("You selected both \"Some\" and \"Significant Amount\".");
+            $numsel = 0;
```
cpeel, line 201: What you have here is clear and I'm not suggesting that you change it, but the optimizer in me kept trying to figure out if there was a simpler way to do this and I think the answer is "no, but here's a more (untested) PHP way":
```php
if (count(array_intersect([$basic_id, $average_id, $complex_id], array_keys($_POST))) > 1)
```

cpeel (approved):
These updates look great to me - thank you for making them!

srjfoo (approved):

tools/post_proofers/ppv_report.php
windymilla: Yes, I can see how that would do the same - array_intersect would return all the keys from `$_POST` that match the basic, average or complex ids for this feature type. If there's more than one match, something's gone wrong?

That code looked less unwieldy when the `if`s were single line commands, but I fell foul of php-cs-fixer :)

cpeel: I think my one-liner is probably harder to understand overall and think we should keep yours. I'll merge this in tomorrow morning my time unless you object between now and then 😁

chrismiceli:
I wonder if we could enhance the existing unit tests to cover this?

tools/post_proofers/ppv_report.php
windymilla: I'm fine with that, but thanks anyway for the teaching point

windymilla:
> I wonder if we could enhance the existing unit tests to cover this?

I've no idea what that involves. I would say this script is a relatively peripheral aspect of the site and doesn't lead to anything that is stored onsite or that can't be adjusted manually. So, it's probably not worth it, if it's a big effort.

PR #854, opened 2023-01-21
Change some project transitions

srjfoo:
The first commit removes the button that squirrels could use to move a project back into the previous round (the effect of which was to change the project state to unavailable for the previous round and the page states to the previous round's `page_avail` state. This was easy to confuse with doing a round repeat (which is done via separate script). The original reason for the button appears to have been to make it easier, after the move from 2 to 3 proofreading rounds, to shove a P3 project that had no P2 back into an empty P2 round.

The second commit removes the ability for disabled project managers to move a project out of the new project state. From the new project state, it is easier for them to fix problems with a project themselves instead of having to request a squirrel to do so.

Testable in [https://www.pgdp.org/~srjfoo/c.branch/change-pm-transitions/](https://www.pgdp.org/~srjfoo/c.branch/change-pm-transitions/).

### DIFF
--- a/pinc/ProjectTransition.inc
+++ b/pinc/ProjectTransition.inc
@@ -434,7 +434,7 @@ function get_transition($from_state, $to_state)
 new ProjectTransition(
     PROJ_NEW,
     PROJ_P1_UNAVAILABLE,
-    'manager',
+    'unrestricted_manager',
     [
         'project_restriction' => null,
         'action_name' => '[default]',
--- a/pinc/ProjectTransition.inc
+++ b/pinc/ProjectTransition.inc
@@ -681,26 +681,6 @@ for ($rn = 1; $rn <= MAX_NUM_PAGE_EDITING_ROUNDS; $rn++) {
         ];
 
         foreach ($from_states as $from_state) {
-            if ($rn > 1) {
-                // Transition to go back 1 round.
-
-                $prev_round = get_Round_for_round_number($rn - 1);
-
-                new ProjectTransition(
-                    $from_state,
-                    $prev_round->project_waiting_state,
-                    'site_manager',
-                    [
-                        'project_restriction' => null,
-                        'action_name' => sprintf(_('Send back to %s'), $prev_round->id),
-                        'confirmation_question' => '[default]',
-                        'detour' => null,
-                        'settings_template' => "", // don't set modifieddate
-                        'collateral_actions' => null,
-                    ]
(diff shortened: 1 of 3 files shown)


pinc/ProjectTransition.inc
```diff
@@ -681,26 +681,6 @@ for ($rn = 1; $rn <= MAX_NUM_PAGE_EDITING_ROUNDS; $rn++) {
         ];
 
         foreach ($from_states as $from_state) {
-            if ($rn > 1) {
-                // Transition to go back 1 round.
-
-                $prev_round = get_Round_for_round_number($rn - 1);
-
-                new ProjectTransition(
```
chrismiceli, line 689: removing this transition doesn't impact the script you mentioned in the description, right?

chrismiceli (approved):

pinc/project_edit.inc
```diff
@@ -58,8 +58,8 @@ function check_user_can_load_projects($exit_if_not)
     global $site_manager_email_addr;
     if (user_has_project_loads_disabled()) {
         echo "
-              <div class='display-flex'><div class='callout'>"
-            ._("You are not currently permitted to create new projects, or move projects out of the unavailable state.")
+              <div class='display-flex' style='margin-bottom:0.5em'><div class='callout'>"
```
srjfoo, line 61: @cpeel -- I'm not sure this is the correct way to do this, but it was the easiest. Currently, the bottom border of the call-out box has no margin between it and the border of the button below it (at least not in FF):
image

No extra margin was needed on top, and I didn't want to mess with the styling on the callout box. I'd be happy to change this if you can point me in the right direction.

pinc/ProjectTransition.inc
srjfoo: Good question. I don't think it should, but it's a good idea for me to verify that, thanks. I'll investigate that.

srjfoo: Double-checked by actually trying it out. No, it doesn't impact the script. The change doesn't actually remove the plumbing that would allow the transition to happen, it just removes the button that would cause the transition to be performed. The `noncvs` script that actually does the round repeat does not need the button to be defined.

pinc/project_edit.inc
cpeel: I think this is a great fix to that problem 👍

cpeel (approved):

PR #856, opened 2023-02-04
All transitive

chrismiceli:
In response to the dependabot PR, this updates all the transitive deps.


cpeel (approved):
Thanks @chrismiceli!

PR #857, opened 2023-02-06
allow inline math inside display math

70ray:
MathJax can compile inline math inside text inside display math but the format preview previously marked it as an error.
Also add a test for it not to cause an error.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/nest_inline


chrismiceli (approved):

70ray:
Previously something like
```\[e=mc^2 \text{ but \(x=\y)}\] would cause an error because it didn't allow for inline markup \( ... \) inside dieplay markup \[ ... \]```

srjfoo (approved):

PR #859, opened 2023-02-23
Set focus on proofing interface S/R Search field

windymilla:
When the Search/Replace tool is popped in the proofing interface, ensure the keyboard focus is in its Search field. This enables the proofer to begin typing the Search term immediately without an additional click.


windymilla:
Available for testing [here](https://github.com/windymilla/dproofreaders/pull/new/sr-focus)

srjfoo (approved):

cpeel (approved):
Thanks for this UX fix!

windymilla:
Thanks - I will benefit from it :)

PR #861, opened 2023-03-03
Make a function to split text into blocks

70ray:
This fixes  task 2038 https://www.pgdp.net/c/tasks.php?action=show&task_id=2038 and simplifies and improves re-wrap.

Classify blocks based on number of preceding blank lines. Use the function for checking all-bold headings and for re-wrap. Allow Headings and sub-headings to have more than one line. 2 blank lines introduce a section. If the block following the 2 blank lines consists of only one line treat it as a section heading (not all bold); else treat like a paragraph. Add different error codes for not all bold in heading, sub-heading or section heading, add tests for them. Revise re-wrap code and remove unused preview styles.

### DIFF
--- a/SETUP/tests/jsTests/formatPreviewTests.js
+++ b/SETUP/tests/jsTests/formatPreviewTests.js
@@ -257,10 +257,28 @@ QUnit.module("Format preview test", function() {
         issueTest(assert, 0, 8, 1, "spaceBeforeEnd", 1);
     });
 
-    QUnit.test("entirely bold heading", function (assert) {
+    QUnit.test("entirely bold heading with two lines", function (assert) {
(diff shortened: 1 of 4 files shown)


charliehoward4dp:
Headings frequently take multiple lines, and they should be treated as headings , not as paragraphs. What complicates the situation is that many things may be preceded by two blank lines, including Section headings, paragraphs, lists, and tables. Post-processors deal with this routinely.

70ray:
> Headings frequently take multiple lines, and they should be treated as headings , not as paragraphs. What complicates the situation is that either may follow two blank lines. Post-processors deal with this routinely.

Yes, whatever the preview re-wrap does in this case will often be wrong.

70ray:
It could be best to remove the re-wrap function altogether. Or change the way it deals with section headings.

PR #862, opened 2023-03-03
Avoid error if proofer dbl-clicks some buttons

windymilla:
Buttons in the proofing interface that change the page's status can cause an error if double-clicked. Since first click changes the status, second click gives an error. Error only appears to occur on some systems.

Solution is to stop the second click from attempting to re-submit by returning false from onClick handler. On first click, set a user-defined data attribute to true, so second click knows it's the second click.

Affects WordCheck, Save as Done and Proofread Next Page (from proofing and WordCheck interfaces), Save as In Progress and Save as Done.

Testable at https://www.pgdp.org/~windymilla/c/

### DIFF
--- a/SETUP/tests/jsTests/dp_proofTests.js
+++ b/SETUP/tests/jsTests/dp_proofTests.js
@@ -35,7 +35,7 @@ QUnit.module("dp_proof tests", function(hooks) {
         });
         form.appendChild(submitButton);
         document.body.appendChild(form);
-        assert.false(submitButton.disabled, "button should be disabled");
+        assert.false(submitButton.disabled, "button should be enabled");
         submitButton.click();
         setTimeout(() => {
             // simulate event loop for clicks
--- a/pinc/button_defs.inc
+++ b/pinc/button_defs.inc
@@ -1,6 +1,7 @@
 <?php
 include_once($relPath.'site_vars.php');  // $charset for javascript_safe()
 include_once($relPath.'misc.inc');  // attr_safe(), javascript_safe()
+include_once($relPath.'misc.inc');  // attr_safe(), javascript_safe()
 
 define('CHANGE_LAYOUT', 'A');
 define('QUIT', 'B');
@@ -113,6 +114,7 @@ function echo_button($button_id, $which_interface)
                 'name' => "button2",
                 $label => attr_safe(_("Save as 'Done' & Proofread Next Page")),
                 'title' => attr_safe(_("Save as 'Done' & Proofread Next Page")),
+                'onClick' => "return submitJustOnce(this);",
                 'src' => "gfx/bt2.png",
                 'class' => "check_button",
             ];
@@ -124,6 +126,7 @@ function echo_button($button_id, $which_interface)
                 'name' => "button5",
                 $label => attr_safe(_("Save as 'Done'")),
                 'title' => attr_safe(_("Save as 'Done'")),
+                'onClick' => "return submitJustOnce(this);",
                 'src' => "gfx/bt13.png",
                 'class' => "check_button",
             ];
@@ -135,6 +138,7 @@ function echo_button($button_id, $which_interface)
                 'name' => "button1",
                 $label => attr_safe(_("Save as 'In Progress'")),
                 'title' => attr_safe(_("Save as 'In Progress'")),
+                'onClick' => "return submitJustOnce(this);",
                 'src' => "gfx/bt3.png",
                 'class' => "check_button",
             ];
@@ -145,6 +149,7 @@ function echo_button($button_id, $which_interface)
                 'name' => "button10",
                 $label => attr_safe(_("WordCheck")),
                 'title' => attr_safe(_("Run WordCheck")),
+                'onClick' => "return submitJustOnce(this);",
                 'src' => "gfx/bt16.png",
             ];
             break;
--- a/pinc/button_defs.inc
+++ b/pinc/button_defs.inc
@@ -148,7 +148,7 @@ function echo_button($button_id, $which_interface)
                 'name' => "button10",
                 $label => attr_safe(_("WordCheck")),
                 'title' => attr_safe(_("Run WordCheck")),
-                'onClick' => "return submitJustOnce(this);",
+                //                'onClick' => "return submitJustOnce(this);",
                 'src' => "gfx/bt16.png",
             ];
             break;
--- a/pinc/button_defs.inc
+++ b/pinc/button_defs.inc
@@ -1,7 +1,6 @@
 <?php
 include_once($relPath.'site_vars.php');  // $charset for javascript_safe()
 include_once($relPath.'misc.inc');  // attr_safe(), javascript_safe()
-include_once($relPath.'misc.inc');  // attr_safe(), javascript_safe()
 
 define('CHANGE_LAYOUT', 'A');
 define('QUIT', 'B');
--- a/tools/proofers/dp_proof.js
+++ b/tools/proofers/dp_proof.js
@@ -1,5 +1,5 @@
 /* eslint-disable no-use-before-define, no-undef, camelcase */
-/* exported insertCharacter, surroundSelection, changeFontFamily, changeFontSize, showNW, replaceAllText, transformText */
+/* exported insertCharacter, surroundSelection, changeFontFamily, changeFontSize, showNW, replaceAllText, transformText, submitJustOnce */
 // This variable is set by initializeStuff() in dp_scroll.js
 var docRef = null;
 
@@ -230,3 +230,17 @@ function transformText(transformType) {
     txtArea.focus();
 }
 
+// Avoid submitting twice if user double-clicks button
+// Usage: onClick="return submitJustOnce(this);"
+//
+// Uses a data attribute to store when button has been clicked for the first time
+// Returns true on first click to continue submission
+// If attribute is already set (second click), return false to avoid re-submitting
+function submitJustOnce(btn) {
+    if (btn.dataset.alreadysubmitted) {
+        return false;
+    } else {
+        btn.dataset.alreadysubmitted = true;
+        return true;
+    }
+}
--- a/tools/proofers/dp_proof.js
+++ b/tools/proofers/dp_proof.js
@@ -237,10 +237,10 @@ function transformText(transformType) {
 // Returns true on first click to continue submission
 // If attribute is already set (second click), return false to avoid re-submitting
 function submitJustOnce(btn) {
-    if (btn.dataset.alreadysubmitted) {
+    if (btn.dataset.alreadysubmitted === 'true') {
         return false;
     } else {
-        btn.dataset.alreadysubmitted = true;
+        btn.dataset.alreadysubmitted = 'true';
         return true;
     }
 }
--- a/tools/proofers/dp_proof.js
+++ b/tools/proofers/dp_proof.js
@@ -235,7 +235,12 @@ function submitForm(form) {
     setTimeout(() => {
         // disable after form submits (setTimeout) since disabled form values aren't sent and we
         // need the submit button to submit to determine which submit button was clicked
+        // standard interface buttons have type 'submit'
         form.querySelectorAll('input[type=submit]').forEach(inputSubmit => {
+            inputSubmit.disabled = true;
+        });
+        // enhanced interface buttons have type 'image'
+        form.querySelectorAll('input[type=image]').forEach(inputSubmit => {
             inputSubmit.disabled = true;
         });
     }, 0);
--- a/tools/proofers/dp_proof.js
+++ b/tools/proofers/dp_proof.js
@@ -1,5 +1,5 @@
 /* eslint-disable no-use-before-define, no-undef, camelcase */
-/* exported insertCharacter, surroundSelection, changeFontFamily, changeFontSize, showNW, replaceAllText, transformText, submitJustOnce */
+/* exported insertCharacter, surroundSelection, changeFontFamily, changeFontSize, showNW, replaceAllText, transformText, submitJustOnce, submitForm */
 // This variable is set by initializeStuff() in dp_scroll.js
 var docRef = null;
 
@@ -243,4 +243,9 @@ function submitJustOnce(btn) {
         btn.dataset.alreadysubmitted = 'true';
         return true;
     }
+}
+
+function submitForm() {
+    document.querySelector('input').disabled = true;
+    return true;
 }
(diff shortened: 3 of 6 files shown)


pinc/button_defs.inc
```diff
@@ -1,6 +1,7 @@
 <?php
 include_once($relPath.'site_vars.php');  // $charset for javascript_safe()
 include_once($relPath.'misc.inc');  // attr_safe(), javascript_safe()
+include_once($relPath.'misc.inc');  // attr_safe(), javascript_safe()
```
srjfoo, line 4: Duplicate include?

tools/proofers/dp_proof.js
```diff
@@ -230,3 +230,17 @@ function transformText(transformType) {
...
+//
+// Uses a data attribute to store when button has been clicked for the first time
+// Returns true on first click to continue submission
+// If attribute is already set (second click), return false to avoid re-submitting
+function submitJustOnce(btn) {
+    if (btn.dataset.alreadysubmitted) {
+        return false;
+    } else {
+        btn.dataset.alreadysubmitted = true;
+        return true;
```
srjfoo, line 244: Is `btn.dataset.alreadysubmitted = true;` necessary, or could this part of the conditional just be `return true;`?

cpeel: That's the flag we need to track if the button has been clicked already.

tools/proofers/spellcheck.inc
```diff
@@ -206,6 +207,7 @@ if ($user->profile->i_type == 1) {
             id="spsaveandnext"
             value="<?php echo attr_safe(_("Save as 'Done' & Proofread Next Page")); ?>"
             title="<?php echo attr_safe(_("Save page as done and proofread the next available page")); ?>"
+            onClick="return submitJustOnce(this);",
```
cpeel, line 210: Errant trailing comma.

cpeel (commented):
Love this clever solution!!

tools/proofers/dp_proof.js
```diff
@@ -230,3 +230,17 @@ function transformText(transformType) {
     txtArea.focus();
 }
 
+// Avoid submitting twice if user double-clicks button
+// Usage: onClick="return submitJustOnce(this);"
+//
+// Uses a data attribute to store when button has been clicked for the first time
+// Returns true on first click to continue submission
+// If attribute is already set (second click), return false to avoid re-submitting
+function submitJustOnce(btn) {
```
chrismiceli, line 239: I would prefer we instead just mark the button disabled. That way for slow connections, there is an indication that something happened. Then we don't need to track anything since the 2nd click won't fire any event.

tools/proofers/dp_proof.js
```diff
@@ -230,3 +230,17 @@ function transformText(transformType) {
...
+    if (btn.dataset.alreadysubmitted) {
+        return false;
+    } else {
+        btn.dataset.alreadysubmitted = true;
```
chrismiceli, line 243: per https://developer.mozilla.org/en-US/docs/web/api/htmlelement/dataset, all dataset properties are strings, to avoid any implicit conversions, I would set `btn.dataset.alreadysubmitted = 'true';`. Then we could do a strict equality above if we continued with this approach
`if (btn.dataset.alreadysubmitted === 'true') {`

tools/proofers/dp_proof.js
chrismiceli, line 239: Can we add a unit test for this method easily?

pinc/button_defs.inc
windymilla: Thanks - will fix

tools/proofers/spellcheck.inc
windymilla: Thanks - copied from different context in `button_defs.inc` and not adjusted properly

tools/proofers/dp_proof.js
windymilla: That's what I tried first :)
However, once you disable the button, the submission doesn't happen, even when you return true.

I had an intermediate version that changed the button label, and used that as the flag detected by the second click. 
I decided the hidden data attribute was better because
1. It doesn't need you to choose suitable text to change the label to (though "Submitting" would probably be suitable for most.
2. The new label text doesn't need translating
3. It looks a bit odd: even when you're on a fast connection, you see the button label text change, which is unusual behavior.
4. This code seems fairly safe for anything that might not support the `dataset`. The double click bug would still be able to happen of course, but I don't think it should break normal single clicking.

tools/proofers/dp_proof.js
windymilla: Good idea, thanks - I'll do that

tools/proofers/dp_proof.js
windymilla: I'm sorry, but I know nothing about that. Happy to do something if someone is able to explain what's needed

70ray: I agree.

windymilla:
I think I've addressed everything except the unit test suggestion - is such a test possible this method?

srjfoo (approved):
I was able to get really old versions of FF and not-so-old versions of Chrome (at least through 92) to fail on the main test server using browserstack. I have not tested all browsers on all OSs, just enough to verify that I could predictably get the error, and that this code fixes the problem.

tools/proofers/dp_proof.js
chrismiceli: Essentially you need to add a file here:
https://github.com/DistributedProofreaders/dproofreaders/tree/master/SETUP/tests/jsTests
Then setup a file as described by
https://qunitjs.com/
You can open https://github.com/DistributedProofreaders/dproofreaders/blob/master/SETUP/tests/qunit.html to run the test. I imagine you'll need to create an <input element programmatically for the test.

chrismiceli: I just think that if clicking the button does nothing, it should be disabled. I tried this and it seems to achieve what you want. We don't need to change the label in my opinion.
https://jsfiddle.net/yo42jueb/
I'm not completely opposed to what we have here, just think we can use browser features (disabled button) instead of implementing something ourselves.

windymilla: Thanks @chrismiceli  - It may well be my naivety regarding the interaction between PHP, HTML and JS, and hence where to put everything, so I may have made the wrong edit, but my attempt to do what you suggested does not fix the bug, i.e. I still get the error if I double click. If I have done it correctly, I think the reason could be that by the time it's submitted, it's too late - the second click has already been made, which is why I had to use `onclick`, rather than `onsubmit`.  Maybe you can spot if I did something wrong in trying to copy the code from your jsfiddle:
I only tried it out on the WordCheck button, by commenting out the `onclick` I had previously added to `button_defs.inc`, adding your `submitForm` function to `dp_proof.js`, and adding `'onsubmit'="return submitForm()"` to the `<form...>` in `text_frame_std.inc`.
I'll try pushing a commit with those changes so you can see them in context and hopefully spot whether it's my error, or just that it doesn't trap the double click early enough.

windymilla:
Note that as it currently stands, this code does not fix the bug - see earlier comments

tools/proofers/dp_proof.js
windymilla: I'm afraid I have no idea at all about any of that - if you had time, would you be willing to guide me through the process at some point, maybe under a different PR @chrismiceli ? Thanks.

srjfoo: > Thanks @chrismiceli - It may well be my naivety regarding the interaction between PHP, HTML and JS, and hence where to put everything, so I may have made the wrong edit, but my attempt to do what you suggested does not fix the bug, i.e. I still get the error if I double click. If I have done it correctly, I think the reason could be that by the time it's submitted, it's too late - the second click has already been made, which is why I had to use `onclick`, rather than `onsubmit`. Maybe you can spot if I did something wrong in trying to copy the code from your jsfiddle: I only tried it out on the WordCheck button, by commenting out the `onclick` I had previously added to `button_defs.inc`, adding your `submitForm` function to `dp_proof.js`, and adding `'onsubmit'="return submitForm()"` to the `<form...>` in `text_frame_std.inc`. I'll try pushing a commit with those changes so you can see them in context and hopefully spot whether it's my error, or just that it doesn't trap the double click early enough.

@chrismiceli -- I can confirm that the code currently in @windymilla's sandbox can still be made to raise the error associated by double-clicking on the WordCheck button.

chrismiceli: Let me find time tomorrow to see what is going on with the disable option and maybe I can send a commit with the unit test change too if that is ok?

windymilla: Thanks - that sounds like a good plan

srjfoo: > I've sent a PR to your branch @windymilla here: [windymilla#1](https://github.com/windymilla/dproofreaders/pull/1) If you are ok with that, you can merge it and it should show up here. Sandbox here: https://www.pgdp.org/~chrismiceli/c.branch/form/

I've tested in @chrismiceli's sandbox, and can confirm that it (as far as I can tell) fixes the issue.

Something oddball that I experienced -- the project that I chose for my testing had French for the language. Now, on TEST, this doesn't mean a whole lot, and as the pages were actually in English, I eventually got tired of adding English as a dictionary to WC for the times it came up successfully. So I changed it to English. Whereupon I couldn't get the double-click to error, even in FF 52. (I tested in FF 52, and Chrome 70 and 90.)

It's actually pretty irrelevant, because regardless, it's a good idea to fix the issue. I mention it only because it's a good example of where speed (i.e. loading dictionaries) does seem to make a difference.

SETUP/tests/jsTests/dp_proofTests.js
```diff
@@ -21,6 +21,33 @@ QUnit.module("dp_proof tests", function(hooks) {
...
+        submitButton.type = 'submit';
+        form.addEventListener('submit', (event) => {
+            event.preventDefault();
+            submitForm(event.target);
+            submitCount++;
+            return false;
+        });
+        form.appendChild(submitButton);
+        document.body.appendChild(form);
+        assert.false(submitButton.disabled, "button should be disabled");
```
windymilla, line 38: Should this message say "button should be enabled"?

tools/proofers/text_frame_std.inc
```diff
@@ -37,7 +37,7 @@ function echo_text_frame_std($ppage)
         "body_attributes" => "id='standard_interface_text' onload='ldAll()'",
     ];
     slim_header(_("Text Frame"), $header_args); ?>
-    <form name="editform" id="editform" method="post" action="processtext.php" target="proofframe" 'onsubmit'="return submitForm()">
+    <form name="editform" id="editform" method="post" action="processtext.php" target="proofframe" onsubmit="return submitForm(this)">
```
windymilla, line 40: Does `proof_frame_enh.inc` need this same edit?

windymilla (commented):
Thanks @chrismiceli
That fix also seems to stop the error from happening. I agree it makes more sense to disable buttons than to ignore clicks programmatically. Also, if I understand correctly, in addition to double-clicking, your solution traps a person clicking on one button, then clicking on another button before the page refreshes.

I've added a couple of questions, but they may reflect more on my inexperience than any changes needed in the code.

SETUP/tests/jsTests/dp_proofTests.js
chrismiceli: yup, that is a copy paste error

tools/proofers/text_frame_std.inc
chrismiceli: yes I would imagine the change is very similar.

windymilla: Thanks - I'll do those and push another commit

windymilla:
Thanks @chrismiceli  - could you check if the edit I made to `submitForm` looks OK. It seems to fix double-clicking in the enhanced interface

tools/proofers/dp_proof.js
```diff
@@ -235,9 +235,14 @@ function submitForm(form) {
     setTimeout(() => {
         // disable after form submits (setTimeout) since disabled form values aren't sent and we
         // need the submit button to submit to determine which submit button was clicked
+        // standard interface buttons have type 'submit'
         form.querySelectorAll('input[type=submit]').forEach(inputSubmit => {
             inputSubmit.disabled = true;
         });
+        // enhanced interface buttons have type 'image'
+        form.querySelectorAll('input[type=image]').forEach(inputSubmit => {
```
chrismiceli, line 243: why do we need to disable images? do they have click handlers on them that submit the form?

windymilla: That was my understanding from looking at the enhanced interface core. Without that, double clicking the wordcheck image button  causes the error we're trying to avoid.

srjfoo: > That was my understanding from looking at the enhanced interface code. Without that, double clicking the wordcheck image button causes the error we're trying to avoid.

I'm glad you were able to verify this -- I couldn't get the error in the enhanced interface at all on browserstack, Windows or maOS, even using the oldest supported browsers. (And good that you remembered -- I was just testing in the standard interface.

chrismiceli: Ah, checked it out, this is correct. input type=image behave like submit buttons per https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/image

chrismiceli (approved):

srjfoo (approved):

tools/proofers/text_frame_std.inc
```diff
@@ -23,6 +23,7 @@ function echo_text_frame_std($ppage)
             "$code_url/scripts/text_validator.js",
             "$code_url/tools/proofers/proof_validate.js",
             "$code_url/tools/proofers/process_diacritcal_markup.js",
+            "$code_url/tools/proofers/dp_proof.js",
```
70ray, line 26: Rather than including dp_proof.js here would it be better to access the one already loaded in tools/proofers/proof.php by prepending "top."? as is done for initializeStuff() referring to dp_scroll. Similarly elsewhere.

70ray (changes requested):

tools/proofers/text_frame_std.inc
windymilla: Thanks @70ray - another new thing I've learned :)

windymilla:
There was another include in `spellcheck.inc` - fixed now. Thanks @srjfoo

cpeel (approved):
Thanks for the teamwork on this solution!

I am going to wait until after I cut the R202303 release (probably next week) before getting this merged in.

PR #864, opened 2023-03-09
Make check for all bold heading consistent with re-wrap

70ray:
Both now allow multi-line headings and sub-headings (previously a multi-line all-bold heading was not marked as an error but the re-wrap still interpreted it as a heading.

This fixes task 2038 https://www.pgdp.net/c/tasks.php?action=show&task_id=2038

The test for an all-bold heading error revised to have a two-line heading.

unused styles removed.

Known bugs (mishandling <tb> in re-wrap) and dubious features (guessing what is a section heading) are still present.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/bold_heading1

### DIFF
--- a/tools/proofers/preview.js
+++ b/tools/proofers/preview.js
@@ -64,6 +64,79 @@ function removeTrail(txt) {
     return txt;
 }
 
+const blockType = {
+    PARA: 0,
+    SECHEAD: 1,
+    SUBHEAD: 2,
+    HEAD: 3,
+    CONTINUATION: 4,
+};
+
+// split bText into blocks surrounded by blank lines
+// and process them by function procBlock
+function blockSplit(bText, procBlock) {
+    // find the number of blank lines preceding the block
+    // at beginning of page assume just past a single \n and thus in a paragraph
+    let prevEnd = -1;
+    // start is first non \n character
+    let beginRegex = /./g;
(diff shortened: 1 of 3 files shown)


tools/proofers/preview.js
```diff
@@ -64,6 +64,79 @@ function removeTrail(txt) {
...
+};
+
+// split bText into blocks surrounded by blank lines
+// and process them by function procBlock
+function blockSplit(bText, procBlock) {
+    // find the number of blank lines preceding the block
+    // at beginning of page assume just past a single \n and thus in a paragraph
+    let prevEnd = -1;
+    // start is first non \n character
+    let beginRegex = /./g;
```
chrismiceli, line 82: in general, I put regular expressions outside functions to avoid them getting parsed and compiled every function invocation since that can be expensive. Obviously this doesn't apply if you are using some of the instance fields of a regular expression like lastIndex.

chrismiceli (approved):

tools/proofers/preview.js
70ray: Good plan. blockSplit only gets called once or twice per invocation of preview so it won't make much difference in this case.

srjfoo (approved):

PR #866, opened 2023-03-20
fix demo splitter jump

70ray:
run code after page competely loaded rather than when DOM loaded which jquey $() does.

The split demonstations [https://www.pgdp.org/~rp31/c/SETUP/tests/manual_web/split_test/vertical_horizontal_split_flex.php](url) and [https://www.pgdp.org/~rp31/c/SETUP/tests/manual_web/split_test/switchable_split.php](url) showed a jump on the horizontal scroll bar when first loaded. Apparently because the splitter code measured the height of the pane before the flex layout had completed.

### DIFF
--- a/SETUP/tests/manual_web/split_test/switchable_split.js
+++ b/SETUP/tests/manual_web/split_test/switchable_split.js
@@ -1,5 +1,5 @@
 /* global $ splitControl */
-$(function () {
+window.addEventListener("load", () => {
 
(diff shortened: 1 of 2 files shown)


chrismiceli (approved):
This isn't something we'd have to worry about in any usages in pgdp right?

I wonder if we could get off jquery in most instances one day.

70ray:
Something is wrong.

PR #868, opened 2023-03-25
Check Special day ID

70ray:
for allowed characters or starts with Birthday or Otherday.

Addresses task https://www.pgdp.net/c/tasks.php?task_id=2050

Sandbox at: https://www.pgdp.org/~rp31/c.branch/birth_other_day

### DIFF
--- a/pinc/misc.inc
+++ b/pinc/misc.inc
@@ -541,7 +541,7 @@ function startswith($subject, $prefix)
 }
 
 function startswithnocase($subject, $prefix)
-// Return TRUE iff $subject starts with $prefix.
+// Return TRUE iff $subject starts with $prefix ignoring case.
 {
     return (strncasecmp($subject, $prefix, strlen($prefix)) == 0);
 }
--- a/pinc/misc.inc
+++ b/pinc/misc.inc
@@ -540,6 +540,12 @@ function startswith($subject, $prefix)
     return (strncmp($subject, $prefix, strlen($prefix)) == 0);
 }
 
+function startswithnocase($subject, $prefix)
+// Return TRUE iff $subject starts with $prefix.
+{
+    return (strncasecmp($subject, $prefix, strlen($prefix)) == 0);
+}
+
 function endswith($subject, $suffix)
 // Return TRUE iff $subject ends with $suffix.
 {
--- a/tools/site_admin/manage_special_days.php
+++ b/tools/site_admin/manage_special_days.php
@@ -33,15 +33,19 @@ if ($action == 'update_oneshot') {
 
         $spec_code = trim($_POST['spec_code']);
 
-        if ((strlen($spec_code) < 1) || (preg_match("/[^A-Za-z0-9. -]/", $spec_code) === 1)) {
-            $errmsgs .= _("Please enter a Special Day Code. Allowable characters are alphanumeric characters and the period, hyphen or space.") . "<br>";
-        } elseif ((startswithnocase($spec_code, "Otherday")) || (startswithnocase($spec_code, "Birthday"))) {
-            $errmsgs .= _('The ID cannot start with "Otherday" or "Birthday".') . "<br>";
-        }
-
         $source = new SpecialDay($spec_code);
-        if (!isset($_POST['editing']) && !$source->new_source) {
-            $errmsgs .= _('A Special Day with this ID already exists. Please choose a different ID for this event.') . "<br>";
+
+        if (!isset($_POST['editing'])) {
+            // we are creating a new special day
+            if (!$source->new_source) {
+                $errmsgs .= _('A Special Day with this ID already exists. Please choose a different ID for this event.') . "<br>";
+            } else {
+                if ((strlen($spec_code) < 1) || (preg_match("/[^A-Za-z0-9. -]/", $spec_code) === 1)) {
+                    $errmsgs .= _("Please enter a Special Day Code. Allowable characters are alphanumeric characters and the period, hyphen or space.") . "<br>";
+                } elseif ((startswithnocase($spec_code, "Otherday")) || (startswithnocase($spec_code, "Birthday"))) {
+                    $errmsgs .= _('The ID cannot start with "Otherday" or "Birthday".') . "<br>";
+                }
+            }
         }
 
         // validate the numeric fields
--- a/tools/site_admin/manage_special_days.php
+++ b/tools/site_admin/manage_special_days.php
@@ -33,14 +33,19 @@ if ($action == 'update_oneshot') {
 
         $spec_code = trim($_POST['spec_code']);
 
-        $source = new SpecialDay($spec_code);
-
-        if (!isset($_POST['editing']) && !$source->new_source) {
-            $errmsgs .= _('A Special Day with this ID already exists. Please choose a different ID for this event.') . "<br>";
-        }
-
         if (strlen($spec_code) < 1) {
             $errmsgs .= _("A value for Special Day ID is required. Please enter one.") . "<br>";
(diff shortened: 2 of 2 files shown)


tools/site_admin/manage_special_days.php
```diff
@@ -33,16 +33,21 @@
...
 
-        $source = new SpecialDay($spec_code);
+        if (strlen($spec_code) < 1) {
+            $errmsgs .= _("A value for Special Day ID is required. Please enter one.") . "<br>";
+        } elseif (preg_match("/[^A-Za-z0-9. -]/", $spec_code) === 1) {
+            $errmsgs .= _('The code must contain only alphanumeric characters . - or space.') . "<br>";
+        } elseif (startswith($spec_code, "Otherday")) {
+            $errmsgs .= _('The ID cannot start with "Otherday".') . "<br>";
+        } elseif (startswith($spec_code, "Birthday")) {
+            $errmsgs .= _('The ID cannot start with "Birthday".') . "<br>";
```
srjfoo, line 43: This check should probably be case insensitive.

tools/site_admin/manage_special_days.php
```diff
@@ -33,16 +33,21 @@
 
         $spec_code = trim($_POST['spec_code']);
 
-        $source = new SpecialDay($spec_code);
+        if (strlen($spec_code) < 1) {
+            $errmsgs .= _("A value for Special Day ID is required. Please enter one.") . "<br>";
+        } elseif (preg_match("/[^A-Za-z0-9. -]/", $spec_code) === 1) {
+            $errmsgs .= _('The code must contain only alphanumeric characters . - or space.') . "<br>";
```
srjfoo, line 39: Should the form field be a required field, with "Please enter a Special Day Code. Allowable characters are alphanumeric characters and the period, hyphen or space." as the error message if an attempt to create an empty special code is made, or one with bad characters? (I realize you just moved the existing check -- but this wouldn't be the first time we've updated a form to require fields.)

srjfoo (changes requested):

tools/site_admin/manage_special_days.php
```diff
@@ -33,16 +33,21 @@
 
         $spec_code = trim($_POST['spec_code']);
 
-        $source = new SpecialDay($spec_code);
+        if (strlen($spec_code) < 1) {
+            $errmsgs .= _("A value for Special Day ID is required. Please enter one.") . "<br>";
+        } elseif (preg_match("/[^A-Za-z0-9. -]/", $spec_code) === 1) {
```
cpeel, line 38: We only want to do this validation for new entries, not for existing entries. This is most important for other sites which might have already created some special days with these characters in them.

cpeel (commented):
This is not from your change, but it's incredibly confusing that the validation is happening outside the class and we "pass in" `$errmsgs` to the class as a global and if it is set to something we output and die. BUT that's something we should fix another day/time/PR.

70ray:
Why don't we automatically generate a unique ID for new special days like we do for new projects? This would avoid the user having to devise it and could avoid "Birthday" etc.

cpeel:
> Why don't we automatically generate a unique ID for new special days like we do for new projects?

I don't know why the "choose your own primary key" was chosen as an implementation. It should really be an auto-incrementing integer but that's going to be a huge change and I don't think the churn is worth the effort.

srjfoo:
> > Why don't we automatically generate a unique ID for new special days like we do for new projects?
> 
> I don't know why the "choose your own primary key" was chosen as an implementation. It should really be an auto-incrementing integer but that's going to be a huge change and I don't think the churn is worth the effort.

As Casey said. We've definitely bemoaned the user-defined unique ID, but changing it at this point would involve fixing at a minimum the `projects` table entries for all projects with special days, and the `queue_defns` table. Realistically, we don't create that many new special days, except in testing, but it's reasonable to prevent those sorts of things from happening in the future. It would also affect anyone who is using our code (though I wouldn't use that as an excuse not to change it if there were really good reasons to go ahead with it.

srjfoo:
> > > > Why don't we automatically generate a unique ID for new special days like we do for new projects?
> > > 
> > > 
> > > I don't know why the "choose your own primary key" was chosen as an implementation. It should really be an auto-incrementing integer but that's going to be a huge change and I don't think the churn is worth the effort.
> > 
> > 
> > As Casey said. We've definitely bemoaned the user-defined unique ID, but changing it at this point would involve fixing at a minimum the `projects` table entries for all projects with special days, and the `queue_defns` table. Realistically, we don't create that many new special days, except in testing, but it's reasonable to prevent those sorts of things from happening in the future. It would also affect anyone who is using our code (though I wouldn't use that as an excuse not to change it if there were really good reasons to go ahead with it.
> 
> I would not have thought this would be necessary since if the Id is just a string of characters it would not affect how any exising ids are used. But it could be more trouble than it's worth to change to automatically generated ids.

I was assuming that converting to autogenerated IDs would mean that all existing IDs would need to be converted and the field type changed. If that were the case, the IDs used in the `projects` and `queue_defns` tables would have to be updated to use the new IDs.

70ray:
Looking at what would be involved in auto-generating it.

70ray:
> > Looking at what would be involved in auto-generating it.
> 
> It just occurred to me that whatever treatment special days gets should also be applied to image sources (though probably in a different PR).
> 
> I still have my doubts about converting to auto-generating the unique IDs because of the potential effect it would have on any other person/group/site using the code.

I looked at re-writing it as Casey mentioned above, which would be quite a large change. There could be a case that using an API to edit/add a new special day would simplify it because after submitting faulty data you would not have to re-draw the form with the faulty data since it would persist in the client, but just show the error messages.

The code for image sources is very similar to special days.

Perhaps best just to make the change we first thought of for now.

tools/site_admin/manage_special_days.php
70ray: Or perhaps leave out this check.

cpeel:
I agree that the approach in this PR is the easiest lift right now. We still need the code to do this additional validation only on adding a new special day, not editing an existing one. Right now, your code does not allow a user to edit the existing `$Day` special day on TEST, for instance.

70ray:
> I agree that the approach in this PR is the easiest lift right now. We still need the code to do this additional validation only on adding a new special day, not editing an existing one. Right now, your code does not allow a user to edit the existing `$Day` special day on TEST, for instance.

Yes, I noticed that after reading your earlier comment. The logic of how this works seemed very confusing, so I did not make any progress but I'll have another look at it.

70ray:
It should now be possible to edit exising special days with bad characters or Birthday.

There is a peculiarity which was there before: if you try to create a new special day but enter an already existing id then you are advised to choose a different one but the field is not then editable and it behaves as if you are editing the pre-existing special day. So you have to close the page and start again.

pinc/misc.inc
```diff
@@ -540,6 +540,12 @@ function startswith($subject, $prefix)
     return (strncmp($subject, $prefix, strlen($prefix)) == 0);
 }
 
+function startswithnocase($subject, $prefix)
+// Return TRUE iff $subject starts with $prefix.
```
cpeel, line 544: `// Return TRUE iff $subject starts with $prefix ignoring case`

cpeel (approved):
Minor update to comment but LGTM.

> There is a peculiarity which was there before

Yeah, not too worried about that since this is existing behavior.

pinc/misc.inc
70ray: done this.

srjfoo (approved):
The sandbox you linked to in the PR is broken for me, but I was able to create one from this branch and confirm that everything works as it should. Not sure what happened to your sandbox, since you did mention later that you'd updated it.

PR #869, opened 2023-03-26
Finish abstracting dpsession API

cpeel:
Finish the dpsession API abstraction so `dpsession.inc` contains the full set of functions that need to be implemented. This has no functional changes but it's a pattern I want to use elsewhere and decided it would be best if this "reference implementation" was complete.

Testable in the [finish-session-abstraction](https://www.pgdp.org/~cpeel/c.branch/finish-session-abstraction/) sandbox. 

Testing suggestion:
* Logging in and logging out works as expected
* Getting a page for proofreading (starting to proofread a page) works -- this calls the `dpsession_page_set()` code.

---

Aside: Unfortunately PHP doesn't support a way to "override" or replace a function once it has been defined (like Python and other languages do) which is why this was implemented like this a long time ago. I've not been able to find a better PHP way to do this kind of abstraction.

I had hoped [namespaces](https://www.php.net/manual/en/language.namespaces.php) might work -- define each implementation in its own namespace and then just set which namespace you want the functions called from -- but I can't figure out how to make that work either because the `use` syntax [cannot be inside a block definition](https://www.php.net/manual/en/language.namespaces.importing.php#language.namespaces.importing.scope), eg:
```php
// namespaces implement the full API so no need for wrappers
if ($use_php_sessions) {
    use DProofreaders\Sessions\PHPSession;
} else {
    use DProofreaders\Sessions\Cookies;
}
```

Instead of using the an implementation pattern of "functions must be named with a trailing `_`" we could use namespaces still eg:
```php
namespace DProofreaders\Sessions\PHPSession;
function dpsession_begin() {
}
```
then
```php
include_once($relPath.'dpsession_via_php_sessions.inc');
include_once($relPath.'dpsession_via_cookies.inc');

if ($use_php_sessions) {
    $session_namespace = "DProofreaders\\Sessions\\PHPSession";
} else {
    $session_namespace = "DProofreaders\\Sessions\\Cookies";
}

function dpsession_begin($userID) {
    $func = "$session_namespace\\dpsession_begin";
    $func($userID);
}
```
But that doesn't seem like it's particularly better.

This isn't particularly important for this sessions code but as I said, I am wanting to use this pattern elsewhere.

Open to ideas on ways to make this "pluggable" in a better way.

### DIFF
--- a/pinc/dpsession.inc
+++ b/pinc/dpsession.inc
@@ -18,11 +18,20 @@
 //     End a session. (The user has logged out.)
 //
 // Page-info...
-// dpsession_page_set($info)
-// dpsession_page_end()
-// dpsession_page_is_set()
-// dpsession_page_get()
+//     dpsession_page_set($info)
+//     dpsession_page_end()
+//     dpsession_page_is_set()
+//     dpsession_page_get()
+//
+// This API is an abstraction around two implementations:
+// * Sessions
+// * Cookies
+//
+// The implementations are controlled by conditionally including one of two
+// PHP files that implement the desired API functions with names that end
+// in an underscore, eg: dpsession_begin_().
 
+global $pguser;
 $pguser = null;
 
 include_once($relPath.'site_vars.php');
@@ -70,6 +79,39 @@ function dpsession_resume()
 
     return $user_is_logged_in;
 }
+
+function dpsession_end()
+{
+    dpsession_end_();
+}
+
+// -----------------------------------------------------------------------------
+
+// The 'debouncer' cookie...
+
+function dpsession_page_set($info)
+{
+    dpsession_page_set_($info);
+}
+
+function dpsession_page_end()
+{
+    dpsession_page_end_();
+}
+
+function dpsession_page_is_set()
+{
+    return dpsession_page_is_set_();
+}
+
+function dpsession_page_get()
+{
+    return dpsession_page_get_();
+}
+
+// XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
+
+// Helper functions that aren't part of the dpsession API.
 
 function _update_user_activity_time($update_login_time_too)
 {
--- a/pinc/dpsession_via_cookies.inc
+++ b/pinc/dpsession_via_cookies.inc
@@ -37,7 +37,7 @@ function dpsession_resume_()
     return true;
 }
 
-function dpsession_end()
+function dpsession_end_()
 {
     user_cookie_disable();
 }
@@ -46,22 +46,22 @@ function dpsession_end()
 
 // The 'debouncer' cookie...
 
-function dpsession_page_set($info)
+function dpsession_page_set_($info)
 {
     local_send_cookie("debouncer", $info);
 }
 
-function dpsession_page_end()
+function dpsession_page_end_()
 {
     local_send_cookie("debouncer", "");
 }
 
-function dpsession_page_is_set()
+function dpsession_page_is_set_()
 {
     return isset($_COOKIE['debouncer']);
 }
 
-function dpsession_page_get()
+function dpsession_page_get_()
 {
     return $_COOKIE['debouncer'];
 }
--- a/pinc/dpsession_via_php_sessions.inc
+++ b/pinc/dpsession_via_php_sessions.inc
@@ -48,7 +48,7 @@ function dpsession_resume_()
     }
 }
 
-function dpsession_end()
+function dpsession_end_()
 {
     session_unset();
     session_destroy();
@@ -58,22 +58,22 @@ function dpsession_end()
 
 // The 'debouncer' variable
 
-function dpsession_page_set($info)
+function dpsession_page_set_($info)
 {
     $_SESSION['debouncer'] = $info;
 }
 
-function dpsession_page_end()
+function dpsession_page_end_()
 {
     $_SESSION['debouncer'] = null;
 }
 
-function dpsession_page_is_set()
+function dpsession_page_is_set_()
 {
     return isset($_SESSION['debouncer']);
 }
 
-function dpsession_page_get()
+function dpsession_page_get_()
 {
     return $_SESSION['debouncer'];
 }


chrismiceli:
This is over my php head so I will defer unfortunately. Does it make sense to allow cookies for the API? then you could just do a check for the cookie, if not present, fallback to api tokens. Kind of wonky though, but would allow 1 auth strategy to work for API and the site.

70ray:
I also do not understand this high-level stuff (i.e. a long way from assembly language). Could you point us to some documentation about what is going on with sessions so we can better understand it? I saw the code in proof_frame.php which uses these functions but, since the comments mention Firefox 1.0.4, had assumed it was intended to fix a problem in the distant past which we do not need to worry about any more. But this seems not to be the case.

cpeel:
There is no good documentation of this stuff. Basically we have 2 ways for the code to manage sessions that the site admin can set when configuring the code:
* PHP Sessions
* Cookies

Since at least 2007 we've been using PHP sessions and I'm not sure if the site has ever used the cookie method. What should probably happen is that we remove sessions-via-cookie support from the code, but before we do that I would want to do a DP release with a warning in it and then remove it.

Only tangentially related this code also includes functions to support the "debouncer" which is the bit in proof_frame that references Firefox 1.x. As the huge code block there indicates, it's more than a little unclear on what edgecase this code use to support and if it's still needed. However, the "debouncer" functions to set/store the page info are very lightweight.

srjfoo:
> Since at least 2007 we've been using PHP sessions and I'm not sure if the site has ever used the cookie method. What should probably happen is that we remove sessions-via-cookie support from the code, but before we do that I would want to do a DP release with a warning in it and then remove it.

Like the others, I don't know enough to have an opinion on the design. The above quote jumped out at me, though. If you did not have to worry about the sessions-via-cookie part of the code, how significantly would it simplify the code? If you proceed with support for both at this point, are you going to then want to rip it out at a later point? Are there good reasons to continue to support sessions-via-cookies?

srjfoo:
Logging out and in went okay; checking out a page went okay, but running WC raised this notice:
> **Notice**: tempnam(): file created in the system's temporary directory in **/home/cpeel/public_html/c.branch/finish-session-abstraction/pinc/wordcheck_engine.inc** on line **243**

windymilla:
@cpeel - just checking I understand...

Previously, wherever you wanted to use a function like `dpsession_end` you would include `dpsession.inc` and (due to the conditional include in that file) you would either get the version of `dpsession_end` that works with php sessions or the one that works with cookies, but both of those were called `dpsession_end`.
What you have done is make it so that there is just one version of `dpsession_end`. It calls `dpsession_end_` (with the trailing underscore) and it is this latter routine that exists in two forms.
So, the convention that you are making is that if you have a function that similarly needs to exist in two forms with the same name, then that name should end with a trailing underscore?

You wrote
> Aside: Unfortunately PHP doesn't support a way to "override" or replace a function once it has been defined
Presumably (almost complete ignorance on my part) using a derived class and overriding methods is not possible/appropriate in this case?

cpeel:
> If you did not have to worry about the sessions-via-cookie part of the code, how significantly would it simplify the code?

A bit, but not a lot. Basically we would replace `dpsession.inc` with `dpsession_via_php_sessions.inc` and delete `dpsession_via_cookies.inc`. The changes I'm suggesting wouldn't be hard to toss out later.

> Are there good reasons to continue to support sessions-via-cookies?

No, and I suspect that the code wouldn't work without PHP sessions to begin with as there are _many_ places we use `$_SESSION`. I'm totally on-board with removing the cookies-based session support, but I want to do one last release with it in and have a changelog message that it will be removed.

> Logging out and in went okay; checking out a page went okay, but running WC raised this notice:

Hmmmm. I can't see how that is related to my code changes -- does it happen with the main TEST site too?

> So, the convention that you are making is that if you have a function that similarly needs to exist in two forms with the same name, then that name should end with a trailing underscore?

That is correct, but note that it is not a new convention as the code does this already for `dpsession_begin()` and `dpsession_resume()`, I'm just extending it to the other functions. (see https://github.com/DistributedProofreaders/dproofreaders/blob/master/pinc/dpsession.inc#L38)

> Presumably (almost complete ignorance on my part) using a derived class and overriding methods is not possible/appropriate in this case?

PHP *does* let you override class methods in an [inherited class](https://www.php.net/manual/en/language.oop5.inheritance.php) so we could convert this into an OO-based design using a singleton or static pattern, something like:
```php
class DPSessionBase
{
    public static function begin($info) {
    }

    public static function resume() {
    }

    public static function end() {
    }
}

class PHPSessions extends DPSessionBase
{
    public static function begin($info) {
    }
}

class CookieSessions extends DPSessionBase
{
    public static function begin($info) {
    }
}
```

Then conditionally instantiate an object:
```php
if ($use_php_sessions) {
    $sessionManager = new PHPSessions()
} else {
    $sessionManager = new CookieSessions()
}

$sessionManager::begin($info);
```

That does feel like a much cleaner design. I'm not sure moving the session code to this makes sense if we're going to delete the cookie-based method later (to Sharon's point above about effort vs rip-out).

If the "use an OO inheritance pattern" feels like the better approach I am happy to close this PR as WONTDO since we'll be deleting the cookie part in ~6 months. But this does give me the pattern that I want to use in another spot.

Relatedly, I created https://github.com/DistributedProofreaders/dproofreaders/pull/871 to log when we hit the debouncer with the thought that it probably isn't used and we can remove it. This will let us confirm that.

windymilla (approved):

srjfoo:
> > Logging out and in went okay; checking out a page went okay, but running WC raised this notice:
> 
> Hmmmm. I can't see how that is related to my code changes -- does it happen with the main TEST site too?

As a matter of fact, yes, it does. But I don't remember seeing it in earlier testing. So, when and why did it start?

srjfoo (approved):

70ray:
> Logging out and in went okay; checking out a page went okay, but running WC raised this notice:
> 
> > **Notice**: tempnam(): file created in the system's temporary directory in **/home/cpeel/public_html/c.branch/finish-session-abstraction/pinc/wordcheck_engine.inc** on line **243**

I seem to remember coming across this message when I was making a wordcheck api. The directory $aspell_temp_dir should be created in processtext.php so perhaps it is not getting created for some reason.

70ray:
> Logging out and in went okay; checking out a page went okay, but running WC raised this notice:
> 
> > **Notice**: tempnam(): file created in the system's temporary directory in **/home/cpeel/public_html/c.branch/finish-session-abstraction/pinc/wordcheck_engine.inc** on line **243**

I deleted the directory /tmp/sp_check. Then the notice goes away.

cpeel:
I've opened https://github.com/DistributedProofreaders/dproofreaders/pull/875 to update the changelog so we announce sessions-via-cookies is deprecated. After we cut the next release in ~6 months we can delete the related code and this abstraction implementation.

PR #873, opened 2023-03-27
Send event notifications in receivers language

### LINKED ISSUE #582: Notifications not sent in receiver's language
`notify_project_event_subscribers()` send emails without configuring gettext first for the recipient's desired language. Usually this function gets called indirectly by `automodify.php` and thus sends emails in English regardless of the user's language which is not ideal, but it can get called indirectly from a user action with that user's language which might be French which is decidedly less ideal for most of our community.

The fix for this is two-fold:
1. In `notify_project_event_subscribers()`, move anything that might generate a translated string into the per-recipient while loop (this includes the call to `get_label_for_event()` and the switch statement) and then call `configure_gettext_for_user($username)` at the beginning of each loop iteration. Remember to call `configure_gettext_for_user()` after the while loop to reset to the current user's language.
2. For the strings generated by `get_label_for_event()` to be translated, the `$subscribable_project_events` global needs to be moved into a function (eg `get_subscribable_project_events()` that returns `$subscribable_project_events`). This ensures that the strings are translated into the correct language at call time rather than at array initialization in the global context (which will always be in English for pgdp.net).
   * Be sure to update the following files that use the global to now call the function instead:
     * `pinc/user_project_info.inc`
     * `project.php`
     * `tools/site_admin/copy_pages.php`
     * `tools/set_project_event_subs.php`

windymilla:
Put string generating code inside recipient loop.
Use function instead of global for event labels so they are translated at call time.

Fixes #582 

[Sandbox](https://www.pgdp.org/~windymilla/c.branch/event-lang)

Note that the popup of email contents on TEST is not seen for event notifications (possibly due to the delayed redirection) so the sandbox has a small additional hack to write the contents of the email to a file in /tmp. Name of file contains email address and timestamp to aid testing.

### DIFF
--- a/pinc/user_project_info.inc
+++ b/pinc/user_project_info.inc
@@ -54,20 +54,25 @@ function upi_set_t_latest_page_event($username, $projectid, $timestamp)
 
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 
-$subscribable_project_events = [
-    'round_available' => _("Project becomes available in a round"),
-    'round_complete' => _("Project finishes a round"),
-    'pp_enter' => _("Project enters Post-Processing stage"),
-    'sr_available' => _("Project becomes available for Smooth Reading"),
-    'sr_reported' => _("User has uploaded a Smooth Reading report"),
-    'sr_complete' => _("Project finishes Smooth Reading"),
-    'ppv_enter' => _("Project enters Post-Processing Verification stage"),
-    'posted' => _("Project posted to Project Gutenberg"),
-];
+
+function get_subscribable_project_events()
+{
+    $subscribable_project_events = [
+        'round_available' => _("Project becomes available in a round"),
+        'round_complete' => _("Project finishes a round"),
+        'pp_enter' => _("Project enters Post-Processing stage"),
+        'sr_available' => _("Project becomes available for Smooth Reading"),
+        'sr_reported' => _("User has uploaded a Smooth Reading report"),
+        'sr_complete' => _("Project finishes Smooth Reading"),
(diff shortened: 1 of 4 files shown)


chrismiceli (approved):

pinc/user_project_info.inc
```diff
@@ -247,10 +253,10 @@ function notify_project_event_subscribers($project, $event, $extras = [])
...
 
 function get_n_users_subscribed_to_events_for_project($projectid)
-// Return an array: keys are the same as $subscribable_project_events,
+// Return an array: keys are the same as array returned by get_subscribable_project_events(),
 // each value is the number of subscriptions to that event for the given project.
 {
-    global $subscribable_project_events;
+    $subscribable_project_events = get_subscribable_project_events();
     $items = [];
     foreach ($subscribable_project_events as $event => $label) {
```
cpeel, line 261: What you have works great. Note that you could have also removed the variable entirely with:
```php
foreach (get_subscribable_project_events() as $event => $label) {
```

cpeel (approved):

srjfoo (approved):

srjfoo:
@cpeel -- there are a handful of messages that aren't localized yet that probably should be. Do you feel like those need to be tested before this code is merged?

70ray (approved):

srjfoo:
> > @cpeel -- there are a handful of messages that aren't localized yet that probably should be. Do you feel like those need to be tested before this code is merged?
> 
> Not sure I understand "tested" in this context. I don't think we need to test this change with updated localized strings before this goes in. I could see us wanting to hold off on this being released into production until the strings in this file are localized (since they are so user visible) but I see that as a function of knowing how much lead time we need between merge and the next rollout to PROD. Does that answer your question?

Yes. And I think we have plenty of lead time. If I don't manage to get the relevant strings updated in time, the announcement can just say that string translation updates are in progress.

PR #874, opened 2023-03-28
Reduce jquery usage

chrismiceli:
@70ray and I were considering the possibility of removing the jQuery dependency. I started the work and it doesn't appear that hard, but wanted to get a review of the approach and ensure the time wouldn't be wasted. This PR probably accounts for about 50% of the work if I'm not bad at estimating (which I am).

You can test these changes in the sandbox:
https://www.pgdp.org/~chrismiceli/c.branch/jquery/

The experiences that would need to be tested are
1. page browser: https://www.pgdp.org/~chrismiceli/c.branch/jquery/tools/page_browser.php?project=projectID4c185eb94cc34&imagefile=046.png&mode=image
2. signup "how did you hear about us": https://www.pgdp.org/~chrismiceli/c.branch/jquery/accounts/addproofer.php
3. Handle bad page from image list if you are PM. The submit button for editing text for a given round to be specific.
4. The server time on the right for the PM page: https://www.pgdp.org/~chrismiceli/c.branch/jquery/tools/project_manager/projectmgr.php
5. font size and font for each proofreading layout (vertical and horiztonal): https://www.pgdp.org/~chrismiceli/c.branch/jquery/userprefs.php?tab=1&origin=https%3A%2F%2Fwww.pgdp.org%2F%7Echrismiceli%2Fc.branch%2Fjquery%2Ftools%2Fproject_manager%2Fprojectmgr.php
6. some character picker in proofing interface

### DIFF
--- a/scripts/file_resume.js
+++ b/scripts/file_resume.js
@@ -1,5 +1,5 @@
-/*global $ Resumable uploadTarget uploadMessages maxResumeSize maxNormSize */
-$(function() {
+/*global Resumable uploadTarget uploadMessages maxResumeSize maxNormSize */
+window.addEventListener("DOMContentLoaded", function() {
 
     // This function has a server-side pair in pinc/upload_file.inc:is_valid_filename()
     // which should be updated if the below logic changes.
@@ -33,22 +33,22 @@ $(function() {
     });
 
     function showProgress(text) {
-        $("#upload_progress").html(text);
+        document.getElementById("upload_progress").innerHTML = text;
     }
 
     if(resumable.support) {
-        resumable.assignBrowse($("#resumable_browse"));
-        resumable.assignDrop($("#upload_form"));
+        resumable.assignBrowse(document.getElementById("resumable_browse"));
+        resumable.assignDrop(document.getElementById("upload_form"));
 
         // Show the resumable div
-        $("#resumable_uploader").show();
+        this.document.getElementById("resumable_uploader").style.display = "";
         // Hide the old uploader
-        $("#old_uploader").hide();
+        document.getElementById("old_uploader").style.display = "none";
     }
 
     // Before we start the upload, prevent the user from hitting upload again.
-    $("#old_submit").click(function(ev) {
-        let file = $("#old_browse")[0].files[0];
+    document.getElementById("old_submit").addEventListener("click", function(ev) {
+        let file = document.getElementById("old_browse").files[0];
         if(file.size >= maxNormSize) {
             alert(uploadMessages.fileTooBig);
             ev.preventDefault();
@@ -56,18 +56,18 @@ $(function() {
         }
         if(validate(file.name)) {
             // won't work if we disable browse button so hide it
-            $("#old_browse").hide();
-            $("#old_submit").prop( "disabled", true);
+            document.getElementById("old_browse").style.display = "none";
+            document.getElementById("old_browse").disabled = true;
             showProgress(uploadMessages.working);
         } else {
             ev.preventDefault();
         }
     });
 
-    $("#resumable_submit").click(function() {
+    document.getElementById("resumable_submit").addEventListener("click", function() {
         // in this case we can prevent the user from selecting another file
-        $("#resumable_browse").prop( "disabled", true);
-        $("#resumable_submit").prop( "disabled", true);
+        document.getElementById("resumable_browse").disabled = true;
+        document.getElementById("resumable_submit").disabled = true;
         if(resumable.files.length) {
             // a file has been selected
             resumable.upload();
@@ -80,7 +80,7 @@ $(function() {
     resumable.on('fileAdded', function(file) {
         let filename = file.fileName;
         if(validate(filename)) {
-            $("#resumable_selected_file").text(filename);
+            document.getElementById("resumable_selected_file").textContent = filename;
         } else {
             // if the validation failed, remove it from the list
             resumable.removeFile(file);
@@ -90,17 +90,17 @@ $(function() {
     // After a file has been successfully uploaded, we update the form
     // and submit it for final validation (AV scan, etc).
     resumable.on('fileSuccess', function(file) {
-        $('input[name="resumable_filename"]').val(file.fileName);
-        $('input[name="resumable_identifier"]').val(file.uniqueIdentifier);
-        $('input[name="mode"]').val("resumable");
+        document.querySelector('input[name="resumable_filename"]').value = file.fileName;
+        document.querySelector('input[name="resumable_identifier"]').value = file.uniqueIdentifier;
+        document.querySelector('input[name="mode"]').value = "resumable";
         showProgress(uploadMessages.finalizingUpload);
-        $("#upload_form").submit();
+        document.getElementById("upload_form").submit();
     });
 
     // If an error occurred, re-enable the upload form and show a message
     resumable.on('fileError', function(file, message) {
-        $("#resumable_browse").show();
-        $("#resumable_submit").show();
+        document.getElementById("resumable_browse").style.display = '';
+        document.getElementById("resumable_submit").style.display = '';
         showProgress(uploadMessages.uploadFailed + "<br>" + message);
     });
 
--- a/scripts/filter_project.js
+++ b/scripts/filter_project.js
@@ -1,25 +1,29 @@
-/*global $*/
-$(function () {
+/* global */
+window.addEventListener('DOMContentLoaded', function() {
     "use strict";
-    var langSelector = $("#language");
-    var langMatch = $("#lang-match");
-    var diffAll = $("#diff-all");
-    var diffOpt = $(".diff-opt");
+    var langSelector = document.getElementById("language");
+    var langMatch = document.getElementById("lang-match");
+    var diffAll = document.getElementById("diff-all");
+    var diffOpt = document.querySelectorAll(".diff-opt");
 
     function showMatcher() {
-        langMatch.css("display", (0 === langSelector.prop("selectedIndex") ? "none" : "block"));
+        langMatch.style.display = 0 === langSelector.selectedIndex ? "none" : "block";
     }
 
-    langSelector.change(function () {
+    langSelector.addEventListener("change", function () {
         showMatcher();
     });
 
-    diffAll.change(function () {
-        diffOpt.prop("checked", false);
+    diffAll.addEventListener("change", function () {
+        diffOpt.forEach(diffOption => {
+            diffOption.checked = false;
+        });
     });
 
-    diffOpt.change(function () {
-        diffAll.prop("checked", false);
+    diffOpt.forEach(diffOption => {
+        diffOption.addEventListener("change", function() {
+            diffAll.checked = false;
+        });
     });
 
     showMatcher();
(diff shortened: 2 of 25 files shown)


70ray (approved):
Iirc isPlainObject was used because earlier implementations may have used a different sort of thing with the same key. I don't remeber if this happened just during development or not but it's probably not necessary to check now.

cpeel:
I love the idea of removing no-longer-useful 3rd party dependencies. We (well, I) added jQuery to help us write more modern JS in 2018 at a time when we supported very old browsers and its abstraction was a real boon. Are our required browser versions modern enough now that we no longer need jQuery as a wrapper abstraction?

I'm worried that removing jQuery is making our JS code less readable and much more verbose. Consider this change:
```diff
-        const textButton = $("<input>", {type: 'button', class: 'control', value: proofIntData.strings.showText});
-        const imageButton = $("<input>", {type: 'button', class: 'control', value: proofIntData.strings.showImage});
-        const imageTextButton = $("<input>", {type: 'button', class: 'control', value: proofIntData.strings.showImageText});
+        const textButton = document.createElement("input");
+        textButton.type = "button";
+        textButton.classList.add('control');
+        textButton.value = proofIntData.strings.showText;
+        const imageButton = document.createElement("input");
+        imageButton.type = "button";
+        imageButton.classList.add('control');
+        imageButton.value = proofIntData.strings.showImage;
+        const imageTextButton = document.createElement("input");
+        imageTextButton.type = "button";
+        imageTextButton.classList.add('control');
+        imageTextButton.value = proofIntData.strings.showImageText;
```

Do we think the value in removing jQuery is worth the increased verbosity?

windymilla:
Although most of the changes in this PR do appear to make the code more verbose, they don't seem to be as extreme as the example you mentioned @cpeel, i.e. 4 less-readable lines for each previous line. Is that just because those types of uses of jQuery aren't addressed in this PR, but will need to be addressed later?

@chrismiceli - is the reason for wanting to remove the jQuery dependency because there are some problems with it, or a concern about its future, or just a general desire to remove dependencies? 

If it was decided to continue with the removal of jQuery, then presumably if there are some uses of jQuery that are very common, some convenience functions could be created, e.g. replacing  
```
const textButton = $("<input>", {type: 'button', class: 'control', value: proofIntData.strings.showText});
```
with
```
const textButton = make_a_text_button();
```
or
```
const textButton = make_a_button(proofIntData.strings.showText);
```
Or would that defeat the whole/main purpose of removing jQuery? Or is there too much variety for that to be a useful/practical strategy?
Please excuse my ignorance in any of the above questions.

70ray:
Like Nigel, I think it could be useful to make helper functions for some common cases. Although this could be regarded as re-inventing jquery it would be entirely under our control.

cpeel:
If we have to make helper functions to replicate jQuery I'd like a strong argument why we shouldn't just keep jQuery.

srjfoo:
I think @cpeel and @windymilla have covered the questions I might have had.

chrismiceli:
> If we have to make helper functions to replicate jQuery I'd like a strong argument why we shouldn't just keep jQuery.

I probably should've been more clear in the description. The main goal was to reduce/remove dependency on jQuery. The reasons being (in no particular order):
1. Remove burden of keeping jQuery up to date as well as testing upgrades
2. Reduce size to client
3. We don't need it for the complexity of the site. Mostly used for selectors outside a few experiences such as page browser, and the querySelector has been supported by browsers since ~2009.
4. Make code more approachable for newer devs who may not be experienced in jQuery since jQuery has fallen out of favor. (I have been having to constantly check the docs to confirm expectations etc.).

These are all minor wins and may not be worth the risk of change/reduction in legibility/ease of creating experiences. That's why I put this review out to demonstrate what the change could look like, and we can weight the pros/cons. We could even just reduce our jQuery usage and get wins 3 & 4 while maintaining the benefit of jQuery for more complex experiences where it may make more sense.

Most of the required changes are very small, but as we have pointed out, the DOM creation stuff can get a bit more verbose. That's why I included page_browser.js here in this review. An alternative mechanism to do this would be use strings and innerHTML, which would closely match what we have now but without jQuery. I was wary of this though because we need to be more careful about encoding in some cases.

I appreciate the feedback and look forward to the discussion. I have no concerns abandoning this PR, but if we think the wins are worth it, I enjoy doing the work.

cpeel:
I think reducing our reliance on jQuery makes sense and I'm game for proceeding in areas that are pretty straightforward to do so. Your changes in `SETUP/tests/jsTests/addprooferTests.js`, `accounts/addproofer.js`, `tools/project_manager/handle_bad_page.js`, `tools/project_manager/projectmgr.js`, all look good so more of that is great.

For now, I'd keep the jQuery for files like `tools/proofers/character_selector.js`, `scripts/control_bar.js`, `scripts/page_browse.js` that result in a lot of changes that, at least to my non-JS eye, seem to make it harder to read. Does that make sense?

chrismiceli:
> 

That makes perfect sense to me. For the more complex files, I may reduce the jQuery dependency where possible, but keep it where it isn't super necessary. Let me work towards that goal.

cpeel:
This is looking great to me, Chris. Let us know when you think this PR is done and I'll do a closer code review and some testing.

chrismiceli:
> This is looking great to me, Chris. Let us know when you think this PR is done and I'll do a closer code review and some testing.

I think this is now in a good spot to review/test. The remaining usages are almost exclusively in the proofing interface/page browser.

I also made a small change with the XRegExp library. It used to load on every page. I have made it load only when needed. This should make most of the site faster to load since less JS to process though and download (~300Kb). It will make the first time it loads when you actually need it possibly slightly longer though as it possibly hasn't been cached yet. Let me know if this is undesired.

scripts/file_resume.js
```diff
@@ -33,41 +33,41 @@ $(function() {
...
         if(file.size >= maxNormSize) {
             alert(uploadMessages.fileTooBig);
             ev.preventDefault();
             return false;
         }
         if(validate(file.name)) {
             // won't work if we disable browse button so hide it
-            $("#old_browse").hide();
-            $("#old_submit").prop( "disabled", true);
+            document.getElementById("old_browse").style.display = "none";
```
windymilla, line 59: No need to explain the reason (I'm a js/jQuery noob) but sometimes it seems like `.hide()/show()` is replaced with `.style.display = "none"`/`style.display = ''`, sometimes with `style.display = ''`, sometimes adding/removing "nodisp", and sometimes a bit of one method and a bit of the other. If that's all OK, I'm happy.

scripts/filter_project.js
```diff
@@ -1,25 +1,29 @@
...
         showMatcher();
     });
 
-    diffAll.change(function () {
-        diffOpt.prop("checked", false);
+    diffAll.addEventListener("change", function () {
+        diffOpt.forEach(diffOption => {
+            diffOption.checked = false;
+        });
     });
```
windymilla, line 22: The blocks below and above were ones where I (as a js/jQuery noob) couldn't see the reason for the differences. One has foreach inside the function, and one outside, and the original didn't have foreach at all. No need for an explanation - just checking it is what you want.

scripts/file_resume.js
chrismiceli: good question. I used the approach that matched what the php output to hide the elements. Some php pages output elements with the "nodisp" class, others set in the stylesheet `display: none`. I just kept the js consistent with the existing page.

windymilla (approved):

scripts/filter_project.js
srjfoo: I'd like someone to check me on this. I don't know whether it's in `tools/project_manager/show_word_context.js` or `tools/project_manager/show_word_context.php`, but I'm seeing a problem with the image display. If I click on the first image listed in the context list, it displays fine. But all subsequent image links do not work. The test project I'm using is `projectID45e0c521f24cd`.

srjfoo: > I'd like someone to check me on this. ...

@chrismiceli and I looked at this via Slack -- he could not reproduce the problem, and upon re-testing, neither could I. I've no idea what was going on unless it was that I was also testing on TEST.

srjfoo:
Apologies for replying to an unrelated comment above when I should have started a new comment above.

> > I'd like someone to check me on this. ...
> 
> @chrismiceli and I looked at this via Slack -- he could not reproduce the problem, and upon re-testing, neither could I. I've no idea what was going on unless it was that I was also testing on TEST.

I went back and tested again, and realized that when Chris and I were not seeing the problem, that was on TEST, not in the branch sandbox. How I got there:
* Edit project word lists ->
* Suggestions from proofreaders: Display ->
* Context 

I've duplicated the same result as before on two projects. That is, the first context link works. Any link after that doesn't.

scripts/file_resume.js
70ray: This seems slightly risky. Jquery hide() doc. says "the value of the display property is saved in jQuery's data cache so that display can later be restored to its initial value". If the php were ever changed to have a different display value then the js function would no longer work correctly if the display value is built into it.

chrismiceli: Yeah, the jquery hide/show has to be generic. This code only has to handle the specific elements it is working with. I confirmed the mechanism they were using to hide and implemented it here. The initial value is '' for display so we can use that to show, and 'none' to hide. Someone can correct me if I'm wrong based on testing.

scripts/filter_project.js
chrismiceli: > Apologies for replying to an unrelated comment above when I should have started a new comment above.
> 
> > > I'd like someone to check me on this. ...
> > 
> > 
> > @chrismiceli and I looked at this via Slack -- he could not reproduce the problem, and upon re-testing, neither could I. I've no idea what was going on unless it was that I was also testing on TEST.
> 
> I went back and tested again, and realized that when Chris and I were not seeing the problem, that was on TEST, not in the branch sandbox. How I got there:
> 
>     * Edit project word lists ->
> 
>     * Suggestions from proofreaders: Display ->
> 
>     * Context
> 
> 
> I've duplicated the same result as before on two projects. That is, the first context link works. Any link after that doesn't.



> Apologies for replying to an unrelated comment above when I should have started a new comment above.
> 
> > > I'd like someone to check me on this. ...
> > 
> > 
> > @chrismiceli and I looked at this via Slack -- he could not reproduce the problem, and upon re-testing, neither could I. I've no idea what was going on unless it was that I was also testing on TEST.
> 
> I went back and tested again, and realized that when Chris and I were not seeing the problem, that was on TEST, not in the branch sandbox. How I got there:
> 
>     * Edit project word lists ->
> 
>     * Suggestions from proofreaders: Display ->
> 
>     * Context
> 
> 
> I've duplicated the same result as before on two projects. That is, the first context link works. Any link after that doesn't.

I can reproduce it and know what is happening. I'll send a commit tonight to resolve this. Essentially I attached a click handler to the first image, but not the others. This was because the selector selected multiple items and I acted like it only selected one. I'll switch to a querySelectorAll and attach an event listener to each image link.

srjfoo (approved):

cpeel (approved):

cpeel:
This LGTM! I'd like @70ray's final thoughts before I merge this as he's a key JS developer for the site.

70ray (approved):
Looks good to me.

PR #875, opened 2023-03-28
Include deprecation notice for sessions-via-cookies

cpeel:
Deprecate the sessions-via-cookies support in the code (`_USE_PHP_SESSIONS=FALSE`) in the change log. After the next release in ~6 months we can remove the code. Doing it this way gives other sites who might still be using this method notice before it happens.

### DIFF
--- a/SETUP/CHANGELOG.md
+++ b/SETUP/CHANGELOG.md
@@ -7,6 +7,12 @@ see the git history.
 [R202009](https://github.com/DistributedProofreaders/dproofreaders/releases/tag/R202009)
 first before upgrading to R202102 or later releases.**
 
+## R......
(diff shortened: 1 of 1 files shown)


srjfoo (approved):

windymilla (approved):
Is it necessary to describe the alternative at all, like "Future releases will only support PHP sessions", or not really?

cpeel:
> Is it necessary to describe the alternative at all, like "Future releases will only support PHP sessions", or not really?

That's a good idea. I've updated the message to include that and the earliest date I could find in the code when the PHP sessions were the default.

70ray (approved):

PR #878, opened 2023-03-31
Changes to facilitate testing

70ray:
Add global declarations for php unit
more cleanup functions
Set date_created for test user so we can do tests for projects only available for new users.

### DIFF
--- a/SETUP/tests/phpunit_test_helpers.inc
+++ b/SETUP/tests/phpunit_test_helpers.inc
@@ -38,7 +38,7 @@ function delete_test_user($username)
     $user = new User($username);
     $sql = sprintf("
         DELETE FROM current_tallies
-        WHERE holder_id  = '%d'
+        WHERE holder_id = %d
     ", $user->u_id);
     DPDatabase::query($sql);
 
--- a/SETUP/tests/phpunit_test_helpers.inc
+++ b/SETUP/tests/phpunit_test_helpers.inc
@@ -1,6 +1,6 @@
 <?php
 
-function create_test_user($username)
+function create_test_user($username, $age = 0)
 {
     // Attempt to load our test user, if it exists don't create it
     $sql = sprintf("
@@ -16,8 +16,9 @@ function create_test_user($username)
             SET id = '%1\$s',
                 real_name = '%1\$s',
                 username = '%1\$s',
-                email = '%1\$s@localhost'
-        ", DPDatabase::escape($username));
+                email = '%1\$s@localhost',
+                date_created = %2\$d
+        ", DPDatabase::escape($username), time() - ($age * 86400));
         $result = DPDatabase::query($sql);
         if (!$result) {
             throw new Exception(sprintf("Unable to create test user %s", $username));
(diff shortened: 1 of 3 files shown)


pinc/Activity.inc
```diff
@@ -6,6 +6,8 @@ include_once($relPath.'quizzes.inc'); // get_Quiz_with_id
 include_once($relPath.'misc.inc'); // startswith
 include_once($relPath.'User.inc');
 
+// this declaration is needed to enable phpunit to work correctly
```
cpeel, line 9: I'd remove these comments (here and in `ProjectTransition.inc`). Because while they are accurate in this case, we should be more generally declaring global variables as global because otherwise they are just globals by accident if the file is included in the global scope.

I'd also move this right above where `$ACCESS_CRITERIA` is defined a few lines below.

SETUP/tests/phpunit_test_helpers.inc
```diff
@@ -16,8 +16,9 @@ function create_test_user($username)
             SET id = '%1\$s',
                 real_name = '%1\$s',
                 username = '%1\$s',
-                email = '%1\$s@localhost'
-        ", DPDatabase::escape($username));
+                email = '%1\$s@localhost',
+                date_created = %2\$d
+        ", DPDatabase::escape($username), time() - ($age * 86400));
```
srjfoo, line 21: Just trying to understand the logic here -- for purposes of the default, `date_created` will always be the current timestamp? How is `$age` determined in a way such that it would be greater than zero?

srjfoo (commented):
Is there anything here that needs testing that I can do?

70ray:
> Is there anything here that needs testing that I can do?

These changes should not affect any of the normal functions so I don't think there is much to test.

SETUP/tests/phpunit_test_helpers.inc
70ray: Good point. I made this several months ago, my comment above was the wrong way round in that it is to test that an old user who has been on the site more than 21 days cannot access projects reserved for new users. An old user is made by giving the age parameter a value of 22.

srjfoo: Will it be used in that way? If so, then is "create" a good description of the function? Also, if so, maybe add comments to clarify?

70ray: I'll do that.

pinc/Activity.inc
70ray: Done this.

SETUP/tests/phpunit_test_helpers.inc
```diff
@@ -1,34 +1,47 @@
...
+        DELETE FROM user_project_info
+        WHERE username  = '%s'
+    ", DPDatabase::escape($username));
+    DPDatabase::query($sql);
+
+    // Delete Tally
+    $user = new User($username);
+    $sql = sprintf("
+        DELETE FROM current_tallies
+        WHERE holder_id  = '%d'
```
cpeel, line 41: You don't need the quotes, integers don't need to be wrapped as strings.
```sql
WHERE holder_id  = %d
```

70ray: done this

cpeel (approved):
> Is there anything here that needs testing that I can do?

I don't think so as `create_test_user()` is already used as part of the automated tests and the others are sensible building blocks.

PR #879, opened 2023-04-01
Api codes

70ray:
Let api return a message and code.

And one example and a test to show pattern.

~Perhaps the message for NonexistentProjectException should not be translated since it would only occur if there is something wrong with the client application or the end user is manipulating the url.~ The user can enter a nonexistent projectid.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/api_codes


pinc/Project.inc
```diff
@@ -18,9 +18,18 @@ class UTF8ConversionException extends Exception
...
 {
+    // all project errors extend from here and in range 100-199
+    public function __construct($message, $code = 100)
+    {
+        parent::__construct($message, $code);
+    }
 }
 class NonexistentProjectException extends ProjectException
 {
+    public function __construct($message)
```
cpeel, line 29: I like the pattern you established of setting the code in the constructor definition and passing `$code` into the parent constructor. Let's continue that here and for the rest of them too.

api/exceptions.inc
```diff
@@ -3,6 +3,11 @@
 // Base exception that all API exceptions should derive from
 class ApiException extends Exception
 {
+    public function __construct($message = "API exception", $code = 0)
```
cpeel, line 6: Let's add a comment that these API error codes should be in range 0-99. Although we should perhaps use 1-99 to avoid some possible confusion since a return code of 0 on unix means success.

SETUP/tests/phpunit_bootstrap.php
```diff
@@ -5,4 +5,8 @@
 include_once($relPath.'misc.inc');
 include_once($relPath.'wordcheck_engine.inc');
 include_once($relPath.'TableDocumentation.inc');
+
+$apiPath = '../../api/';
+include_once($apiPath.'v1.inc');
```
cpeel, line 10:
```php
include_once($relPath.'../api/v1.inc');
```
would be better so we keep everything relative to `$relPath`.

SETUP/tests/apiTest.php
```diff
@@ -0,0 +1,28 @@
+<?php
```
cpeel, line 1: Let's rename this file `ApiTest` so the case matches the class name.

cpeel (commented):
I think this is a great pattern!

api/exceptions.inc
70ray: done this

70ray:
@cpeel We could add more exception classes for the different cases which occur in v1_projects.inc but I don't know if it would be useful.
None of the changes so far should have any impact on non-api functions, but the next stage of Converting error-detection code to throwing Exceptions could have so it could be best not to start on that until after merging this.

cpeel:
> None of the changes so far should have any impact on non-api functions, but the next stage of Converting error-detection code to throwing Exceptions could have so it could be best not to start on that until after merging this.

Agreed. Let me know when you feel this is ready for review and I'll give it a final look.

pinc/Project.inc
cpeel: Addressing this comment is the only remaining thing for me -- after we resolve this I'm happy to get it in.

cpeel (commented):
Loving the API tests 👍

pinc/Project.inc
70ray: Done this now, thanks.

cpeel (approved):

windymilla (approved):
Some of this is beyond me, but what I follow looks good to me.

api/v1_validators.inc
```diff
@@ -35,21 +39,29 @@ function validate_wordlist($wordlist, $data)
...
 {
     global $Round_for_round_id_;
 
-    $pagerounds = array_merge(["OCR"], array_keys($Round_for_round_id_));
+    try {
+        $pagerounds = array_merge(["OCR"], array_keys($Round_for_round_id_));
 
-    if (!in_array($pageround, $pagerounds)) {
-        throw new NotFoundError("Invalid page round");
+        if (!in_array($pageround, $pagerounds)) {
```
srjfoo, line 60: Picky, but there's a blank line at line 59, whereas `function validate_page_name` follows the same pattern but doesn't have a blank line preceding the `if`.

api/exceptions.inc
```diff
@@ -1,8 +1,14 @@
 <?php
 
 // Base exception that all API exceptions should derive from
+// These exception codes range from 1 to 99
 class ApiException extends Exception
 {
+    public function __construct($message = "API exception", $code = 1)
```
srjfoo, line 7: For this and all similar instances -- are these messages supposed to be translatable?

pinc/Project.inc
```diff
@@ -18,24 +18,74 @@ class UTF8ConversionException extends Exception
 
 class ProjectException extends Exception
 {
+    // all project errors extend from here and in range 100-199
+    public function __construct($message, $code = 100)
```
srjfoo, line 22: For this and similar, messages to be added later?

srjfoo (approved):
Most of this was over my head, so what I was mostly looking at was patterns. None of my questions is critical to the PR, but mostly for my own information.

api/v1_validators.inc
70ray: There isn't really any reason to have a blank line here.

api/exceptions.inc
70ray: As I understand it, we should translate messages which are shown to the end user when something 'expected' happens, like when there are no more pages left to proofread. Some messages would only apear if something unexpected went wrong in the server or the client application. Then the message would be more use to someone looking for the bug who might be more likely to understand it if left in English.
There is one case which is affected by this PR: if you select a different project in the page browser and put in an invalid or nonexistent code then formerly you would get `No such project ID` not translatable, but now get `There is no project with projectid 'asdf'` which is translatable.
Composing this reply I tested whether I can trigger another error by manipulating the url to get a nonexistent page (and press return) when the error shown will be `There is no page 00.png in this project` translated. This message comes from the client rather than the server. (I'm not sure that the client should really work like this but this is a different issue.)

I'll make a sandbox should you wish to try it out.

srjfoo: I would have thought that `No such projectID` would be translatable? Or do you mean that it's just in a place where it can't be translated.

Thanks for the sandbox! I tried both your tests, and they look good.

I know you've mentioned translation before -- how will that work with the messages coming from the client instead of the server. And of course ... the answer is not critical at this point, so I'm happy for you to say "we'll get to that later!". 😁

pinc/Project.inc
70ray: The message is generally put in when the exception is thrown because it might include data from the context but it wouldn't hurt to put in default messages for all cases.

srjfoo: Thanks for the explanations.

api/exceptions.inc
70ray: `No such projectID` could have been translated but wasn't.

At present the client messages are downloaded from the server beforehand, having been translated. So far we have just implemented a simple substitution of one %s in some cases like the one above.

70ray:
@cpeel Do you think we should merge this into master, or continue with the next stage, (Converting error-detection code to throwing Exceptions) on top of this branch?

cpeel (approved):

cpeel:
Sorry Ray, I lost track of the question/responses that came up after I last reviewed this.

PR #880, opened 2023-04-01
Notify event subscribers of PP/PPV transitions

windymilla:
Transitions that this commit affects are
PP Unavailable -> PP Available
PP Checked out -> PP Available
PPV Checked out -> PPV Available

Question: In an ideal world, the additional warning "This will clear the reserved PPer if there is one." would only be output if `checkedoutby` was not empty. I couldn't see how to check that at the point where the confirmation question is stored in the `ProjectTransition`. If there's not a simple way, it's not that important, since this transition is relatively rare, and only made by people who should know what's going on regarding any reserved PPer and the project being put into Available rather than Checked Out.

Addresses [Task 1881](https://www.pgdp.net/c/tasks.php?action=show&task_id=1881)

### DIFF
--- a/pinc/ProjectTransition.inc
+++ b/pinc/ProjectTransition.inc
@@ -857,6 +857,24 @@ function round_to_PP_avail_collateral($old_project, $project, $transition, $who)
     notify_project_event_subscribers($project, 'pp_enter');
 }
 
+// From either PP Checked out or PP Unavailable to PP Available
+function to_PP_available_collateral($old_project, $project, $transition, $who)
+{
+    // Always notify the PM
+    configure_gettext_for_user($project->username);
+    if ($transition->from_state == PROJ_POST_FIRST_CHECKED_OUT) {
+        $body_blurb_messages[] = _("This project has been returned to available for someone else to do the post-processing.");
+    } else {
+        $body_blurb_messages[] = _("This project has been made available, from unavailable, for someone else to do the post-processing.");
+    }
+    $body_blurb_messages[] = _("You will be notified once it has completed post-processing.");
+    $body_blurb = implode("\n\n", $body_blurb_messages);
+    maybe_mail_project_manager($project, $body_blurb, _("Post-Processing Started"));
+    configure_gettext_for_user();
+
+    notify_project_event_subscribers($project, 'pp_enter');
+}
+
 function round_to_PP_out_collateral($old_project, $project, $transition, $who)
 {
     global $site_abbreviation;
@@ -944,7 +962,7 @@ new ProjectTransition(
     ]
 );
 
-// PPer can grab a project back from the PP pool
+// PPer can grab a project back from the PPV pool
 new ProjectTransition(
     PROJ_POST_SECOND_AVAILABLE,
     PROJ_POST_FIRST_CHECKED_OUT,
@@ -975,7 +993,7 @@ new ProjectTransition(
         'confirmation_question' => null,
         'detour' => "$code_url/tools/upload_text.php?project=<PROJECTID>&stage=return_1",
         'settings_template' => "modifieddate='<TIMESTAMP>', checkedoutby='', ppverifier=NULL, smoothread_deadline='0', postcomments=CONCAT(postcomments,'<G:postcomments>')",
-        'collateral_actions' => null,
+        'collateral_actions' => 'to_PP_available_collateral',
         'disabled_during_SR' => true,
     ]
 );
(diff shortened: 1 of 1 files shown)


pinc/ProjectTransition.inc
```diff
@@ -1159,10 +1182,10 @@ new ProjectTransition(
     [
         'project_restriction' => null,
         'action_name' => _('Make Project Available for Post-Processing'),
-        'confirmation_question' => '[default]',
+        'confirmation_question' => $pp_unavail_to_avail_confirm_question,
         'detour' => null,
-        'settings_template' => '',
-        'collateral_actions' => null,
+        'settings_template' => "checkedoutby=''",   // Clear reserved PPer
```
cpeel, line 1187: This is the only part of the PR that I'm not 100% certain about, mostly because I don't fully grok all of the times that this transition is used. For instance, does this prevent transitioning a project that has a reserved PPer from being made available to that PPer?

srjfoo: When a project is moved to Unavailable for PP in the first place, any assigned PPer is removed, so the only way there would be an assigned PPer is if the field was edited. At present, we don't even define a transition for Unavailable directly into PP checked out.

The db-req squirrels can fake this (for instance, a hospital project that the originally assigned PPer would like to have back) by using Project Jump to force the transition into PP checked out.

If a reserved PPer is in the PP slot when a project is Unavailable for PP, and you want a formal way to assign it to them, then we probably need a new transition from PP Unavailable to PP checked out, but if the transition is from Unavailable to Available, there should not be a PP assigned.

Tangent (and apologies if I should be bringing this up elsewhere): There's another problematical situation where the PM edits the PPer field and expects the book to be checked out to the person they assign it to. When they do this, it remains available in the pool unless someone notifies db-req about it. What I'd like to see there is blocking everyone but squirrels from editing the PP field if a project is in the PP pool, but that might need some discussion. I think that there was some discussion about an edit of a field in the project page not being something that should cause a project transition.

windymilla: @srjfoo probably knows more about that - I've never transitioned a project from Unavailable to either Available or Checked Out (except testing this). Directly below the snippet you've quoted is a transition from Unavailable to Checked Out, which says `checkedoutby='<WHO>'`. From my test, that appears to mean it will be checked out to the person making the change (assuming they accept the "It's unclear whether this transition should even be possible. Are you sure?" question). 

So, I think your query is that you might want to move an unavailable project with a reserved PPer into Available, so it is available for them to check out. With or without this commit, the PPer would have to then check it out themselves (or perhaps a squirrel could force that - not sure if a PM could force the move from Available to Checked Out). 

At present, the first step in that leaves the project in a weird state, where it is in the PP pool, but has a PPer. With this commit it will have no PPer. I don't think that affects whether the PPer would be able to check it out themselves. So, I think the two differences this commit makes are that it won't leave a project in Available but with a PPer assigned (probably desirable?) and that if a squirrel is jumping a project to get it to be checked out by the reserve PPer, and they go via Available, rather than direct from Unavailable to Checked Out, they would have to re-set the PPer's name.

cpeel: I think I got it, thanks. Both of you know more about this than I do so I'm happy to approve it. All of the other changes in this look great to me.

cpeel (approved):

pinc/ProjectTransition.inc
srjfoo: > @srjfoo probably knows more about that - I've never transitioned a project from Unavailable to either Available or Checked Out (except testing this). Directly below the snippet you've quoted is a transition from Unavailable to Checked Out, which says `checkedoutby='<WHO>'`. From my test, that appears to mean it will be checked out to the person making the change (assuming they accept the "It's unclear whether this transition should even be possible. Are you sure?" question).

Thanks, Nigel, I was forgetting about that little transition. But yes, it doesn't check the project out to whoever (if anyone) is in the PP field, but to the person pushing the button.

> So, I think your query is that you might want to move an unavailable project with a reserved PPer into Available, so it is available for them to check out. With or without this commit, the PPer would have to then check it out themselves (or perhaps a squirrel could force that - not sure if a PM could force the move from Available to Checked Out).
>
> At present, the first step in that leaves the project in a weird state, where it is in the PP pool, but has a PPer. With this commit it will have no PPer. I don't think that affects whether the PPer would be able to check it out themselves. So, I think the two differences this commit makes are that it won't leave a project in Available but with a PPer assigned (probably desirable?) and that if a squirrel is jumping a project to get it to be checked out by the reserve PPer, and they go via Available, rather than direct from Unavailable to Checked Out, they would have to re-set the PPer's name.

A squirrel can force it, yes, with `project_jump`. And an unrestricted project manager (or PF or squirrel) can check it out themselves and then edit the PP field. A project in the PP pool with a username in the PP slot is not reserved for that PPer -- someone else can check it out, so it's better not to even allow there to be a username in the PP field if it's available for PP.

70ray (approved):

windymilla:
Sorry, forgot to say [transitions branch](https://www.pgdp.org/~windymilla/c.branch/transitions/) is available for testing

srjfoo (approved):

PR #882, opened 2023-04-03
Clear ppverifier field if they didn't upload to PG

windymilla:
If PPV returns project to PP, who then uploads, the PPVer's stats differ from the "obvious" search. 

For example, it's the difference between
`select count(*) from projects where ppverifier = 'windymilla' and state like '%posted%' and postproofer <> 'windymilla' and checkedoutby <> '';`
and
`select count(*) from projects where ppverifier = 'windymilla' and state like '%posted%' and checkedoutby <> 'windymilla' and checkedoutby <> '';`
The latter is used by the PPV stats page and omits those projects that windymilla PPVed, but returned to the PPer for uploading.

This PR stops that happening to projects in future, i.e. if windymilla PPVed a project, but returned it to the PPer, who had it checkedout when it was marked as Posted, windymilla will be removed as PPVer. This seems correct, since the PPVing job has not been completed for some reason, most likely because the PPer was awarded DU while the PPVer had the project checked out.

Sandbox at: https://www.pgdp.org/~windymilla/c.branch/ppv-chk

### DIFF
--- a/pinc/ProjectTransition.inc
+++ b/pinc/ProjectTransition.inc
@@ -1261,14 +1261,18 @@ new ProjectTransition(
 
 /*
     The following table describes the required values for checkedoutby,
-    postproofer  and ppverifier after the transition to the named state.
-    The states are listed in their normal order, though projects may
+    postproofer and ppverifier after the transition to the named state.
+    The states are listed in their normal order, though projects that have
+    not yet reached "posted" may step back to the previous row,
+    e.g. 2-->1 or 5-->4. Note that the step from 4 to 5 is equivalent to
+    jumping from 4 to 2 but preserving the PPVer's name so that the
+    PPer can later move the project from 5 back to 4 for another PPV check.
 
     Columns 2, 3 and 4 correspond to values set in fields in the
     "projects" database table:
-    cob - checkedoutby
-    pp  - postproofer
-    ppv - ppverifier
+        cob - checkedoutby
+        pp  - postproofer
+        ppv - ppverifier
 
     a is the PPer
     b is the PPVer, not the same user as the PPer
--- a/pinc/ProjectTransition.inc
+++ b/pinc/ProjectTransition.inc
@@ -1260,31 +1260,53 @@ new ProjectTransition(
 // Transitions to PROJ_SUBMIT_PG_POSTED:
 
 /*
-    pseudo-code:
+    The following table describes the required values for checkedoutby,
+    postproofer  and ppverifier after the transition to the named state.
+    The states are listed in their normal order, though projects may
+
+    Columns 2, 3 and 4 correspond to values set in fields in the
+    "projects" database table:
+    cob - checkedoutby
+    pp  - postproofer
+    ppv - ppverifier
+
+    a is the PPer
+    b is the PPVer, not the same user as the PPer
+
+        cob pp  ppv    state                         description
+
+    1.  -   -   NULL   proj_post_first_available,    available for PP
+    2.  a   -   NULL   proj_post_first_checked_out,  checked out for PP
+    3.  a   a   NULL   proj_post_second_available,   available for PPV
+    4.  b   a   NULL   proj_post_second_checked_out, checked out for PPV
+    5.  a   -   b      proj_post_first_checked_out,  returned to PP by PPV
+    6.  a   a   NULL   proj_submit_pgposted,         posted to PG, PPer has DU
+    7.  b   a   b      proj_submit_pgposted,         posted to PG, project was PPV'd
+
+
+    pseudo-code for transitions to PROJ_SUBMIT_PG_POSTED:
     IF postproofer == '' THEN
-        # No postproofer has been recorded, so the project never reached PPV.
-        # So whoever has it checked out must be cleared to post directly,
-        # and must have been the PPer.
+        # No postproofer has been recorded, so the project has been DUed
+        # by the PPer who has it checked out (not the PPVer).
         postproofer=checkedoutby
+    ENDIF
+    IF postproofer != checkedoutby THEN
+        # checkedoutby must be the PPVer...
+        # (This assumes a project is only posted once!
+        # otherwise, maybe also test that ppverifier is null?)
+        ppverifier=checkedoutby
     ELSE
-        # A postproofer has been recorded, so the project did reach PPV.
-        IF postproofer != checkedoutby THEN
-            # checkedoutby must be PPVer...
-            # (This assumes a project is only posted once!
-            # otherwise, maybe also test that ppverifier is null?)
-            ppverifier=checkedoutby
-        ENDIF
+        # PPer has it checked out, and has therefore DUed the project
+        # so the PPVer did not complete PPV - remove their name
+        ppverifier=NULL
     ENDIF
 
-    i.e.:
-        postproofer = IF( postproofer = '', checkedoutby, postproofer )
-        ppverifier = IF( postproofer != '' AND postproofer != checkedoutby, checkedoutby, ppverifier )
 */
 
 $pg_posted_settings_template =
     "modifieddate='<TIMESTAMP>',
     postproofer = IF( postproofer = '', checkedoutby, postproofer ),
-    ppverifier = IF( postproofer != '' AND postproofer != checkedoutby, checkedoutby, ppverifier )";
+    ppverifier = IF( postproofer != checkedoutby, checkedoutby, NULL )";
 
 
 new ProjectTransition(


windymilla:
Questions/advice please:
There appeared to be a redundant test [here](https://github.com/DistributedProofreaders/dproofreaders/pull/882/files#diff-5a54f5b7d3192c7fdddf3c84daadbbd1bdc277d21880978dba5ac077b21589b9L1287) for `postproofer = ''` which it can't be due to the previous line. I've retained that check, but wondered whether it should be removed or kept for safety.

Also, related to that, I think the pseudo-code didn't/doesn't reflect the code because it's says
```
    IF postproofer == '' THEN
        postproofer=checkedoutby
    ELSE
        do ppverifier checks
    ENDIF
```
implying that the ppverifier checks aren't done if the postproofer was empty at the start. I think that the code sets the postproofer if it was empty and then goes on to do the ppverifier checks. In fact, I think I'm relying on that for this fix to work.

srjfoo:
Another note. @windymilla noticed the discrepancy when comparing the number of PPV projects returned via project search with the number reported on the PPV stats page. Eventually the discrepancy was determined to be a result of the change implemented by PR #410. Prior to that time, if a PPVer returned a project to the PPer, the `ppverifier` field was returned to `NULL`.

70ray:
This is complicated, I'll have to look again later.

srjfoo:
> This is complicated, I'll have to look again later.

In case it helps, here's a description of the various combinations.

```
1 - checkedoutby
2 - postproofer
3 - ppverifier

a, b -- different users

    1   2   3      state, description
       
1.  -   -   NULL   proj_post_first_available,
                   available for PP
2.  a   -   NULL   proj_post_first_checked_out,
                   checked out for PP
3.  a   a   NULL   proj_post_second_available,
                   available for PPV
4.  b   a   NULL   proj_post_second_checked_out,
                   checked out for PPV
5.  a   -   b      proj_post_first_checked_out,
                   returned to PP by PPV
6.  a   a   NULL   proj_submit_pgposted,
                   posted to PG, PPer has DU
7.  b   a   b      proj_submit_pgposted,
                   posted to PG, project was PPV'd
```

The above are all valid. The current PR is designed to deal with another one that can happen with the current code, which is actually invalid, but can occur in the code currently deployed on PROD:
```
8.  a   a   b      proj_submit_pgposted, *
```
\* The project was initially checked out for PPV, and returned to the PPer for more work. Meanwhile, before the PPer could return it to the PPVer, the PPer earned DU, and subsequently uploaded it directly without returning it to the PPVer.

This PR is designed to leave the final condition of a project uploaded to PG by the new DU, as though it had never been in PPV (i.e. item 6 in the top part of the table), since the PPVer didn't upload. That is, the only time the `ppverifier` field should be non-NULL is when the PPVer actually uploaded the project to PG.

cpeel:
> > This is complicated, I'll have to look again later.
> 
> In case it helps, here's a description of the various combinations.

_**Thank you**_. Can you please put up a PR and put that somewhere in the code in a comment or a developer-centric `README.md` or something? I don't know where that should go but holy cow those three fields and their various states have confused the !@#$ out of me time and time again.

srjfoo:
> > > This is complicated, I'll have to look again later.
> > 
> > 
> > In case it helps, here's a description of the various combinations.
> 
> _**Thank you**_. Can you please put up a PR and put that somewhere in the code in a comment or a developer-centric `README.md` or something? I don't know where that should go but holy cow those three fields and their various states have confused the !@#$ out of me time and time again.

Certainly, but do you want it as a separate `README`, or would it work as well to expand the comment at the top of the `Transitions to PROJ_SUBMIT_PG_POSTED` section of the code that starts at line 1260? I can work with @windymilla on the exact wording: since this code is supposed to remove option 8, we probably ought to leave it out of the comment in the code. If you want a separate file, I'll look around, though.

70ray:
If we are considering the transition to posted then from rows 6, 7 and 8 of @srjfoo's table the logic could be simpler:

postproofer = postproofer.
ppverifier = IF( checkedoutby != postproofer, checkedoutby, NULL)

windymilla:
> If we are considering the transition to posted then from rows 6, 7 and 8 of @srjfoo's table the logic could be simpler:
>
> postproofer = postproofer.
> ppverifier = IF( checkedoutby != postproofer, checkedoutby, NULL)

Assuming we're talking about the same thing, I don't think you can simplify both of them as much as that:
Firstly, the postproofer...
If the PPer DUs the project, and it's never been to PPV, then before the transition to Posted occurs, postproofer is empty, and the PPer's name is in checkedoutby, hence the need for the IF in
`postproofer = IF( postproofer = '', checkedoutby, postproofer ),`

Now the ppverifier...
In my first comment, I queried whether it would be OK to lose the first IF in the ppverifier logic, i.e. change it from
```
    ppverifier = IF( postproofer = '',
                        ppverifier, 
                        IF( postproofer != checkedoutby,
                            checkedoutby,
                            IF( postproofer != ppverifier, NULL, ppverifier)))";
```
to
```
    ppverifier = IF( postproofer != checkedoutby,
                            checkedoutby,
                            IF( postproofer != ppverifier, NULL, ppverifier))";
```
I think it would be OK to remove that first IF, because the postproofer field will definitely be non-empty due to the previous line of code that sets the postproofer field. 
If that first IF is removed, then the question is whether we can simplify that further to 
`ppverifier = IF( checkedoutby != postproofer, checkedoutby, NULL)`
as you suggest, Ray.
The difference is that in your case, you always set ppverifier to NULL if it's not the PPer who has it checked out, but my code only NULLs the ppverifier field if that's true *and* the PPer and PPVer are not the same person. So, the difference between the two bits of logic occurs when the PPer and the PPVer are the same person. But we've already said the PPer doesn't have it checked out, and since the PPVer is the same person, the PPVer doesn't have it checked out either. Can that situation arise @srjfoo (i.e. checkedoutby is empty, but postproofer and ppverifier are not empty and are not the same)? If not, then I think Ray's simplified ppverifier code will suffice, but not the simplified postproofer code.

70ray:
Yes, I was still confused. We are really considering the transitions from rows before 6 to posted. That is 2,3,4,5. Not possible to transition to posted from 1.

srjfoo:
> Yes, I was still confused. We are really considering the transitions from rows before 6 to posted. That is 2,3,4,5. Not possible to transition to posted from 1. Your logic seems correct.
> 
> Interestingly ppv_stats.php doesn't use ppverifier at all!

I just took a look at the database, and there are exactly two posted projects where `ppverifier` is not null and `postproofer` == `ppverifier`. I looked at both of them, and I think that the `ppverifier` field can (and should be) set to `NULL` for both of them. However, the `ppverifier` field does have a use, as it is used to store the `ppverifier`'s username if a project is returned to the PPVer for more work, so that it can be returned directly to the PPVer instead of being put back into PPV available for the PPVer to check out again.

windymilla:
I'll put up another commit with a simplified test after checking that it still seems to do what we want.

windymilla:
Thanks everyone for the advice/comments. I've pushed another commit which now has the simplified tests, removing a redundant one, and just setting the ppverifier field to null if the PPer (not the PPVer) had it checked out at the point it transitioned to "Posted". I've also included Sharon's helpful table in the comments, and made the pseudocode reflect the code.

srjfoo (approved):

pinc/ProjectTransition.inc
```diff
@@ -1260,31 +1260,53 @@ new ProjectTransition(
...
 
 /*
-    pseudo-code:
+    The following table describes the required values for checkedoutby,
+    postproofer  and ppverifier after the transition to the named state.
+    The states are listed in their normal order, though projects may
+
+    Columns 2, 3 and 4 correspond to values set in fields in the
+    "projects" database table:
+    cob - checkedoutby
```
cpeel, line 1269: Can we indent these three lines a bit please?

pinc/ProjectTransition.inc
```diff
@@ -1260,31 +1260,53 @@ new ProjectTransition(
 // Transitions to PROJ_SUBMIT_PG_POSTED:
 
 /*
-    pseudo-code:
+    The following table describes the required values for checkedoutby,
+    postproofer  and ppverifier after the transition to the named state.
+    The states are listed in their normal order, though projects may
```
cpeel, line 1265: Incomplete sentence -- don't leave me hanging! 😁 

(also, double space mid-sentence in the line above)

cpeel (commented):
Thank you for incorporating Sharon's (excellent) table! A few small things in the comments.

pinc/ProjectTransition.inc
windymilla: Done! I did complete that sentence once, so either changed my mind and decided to reword it, then forgot, or I messed up syncing the latest version of the file from my local system. Thanks for spotting it anyway.

cpeel (approved):

PR #884, opened 2023-04-05
Factor out $forums_url

cpeel:
Factor out `$forums_url` so it is only used from `forum_interface.inc`. This further encapsulates knowledge about the forum (`$forums_url` and `$forums_dir`) to the forum interface code. This is a stepping stone to supporting other forums in the future (such as the "always just one year away" future phpBB 4 release).

But most interesting to me is that this lays the foundation for having a test-only "forumless" version we could develop that allows much easier local development because it doesn't require installing and setting up phpBB to get going.

Testable in the [abstract-out-forum-url](https://www.pgdp.org/~cpeel/c.branch/abstract-out-forum-url/) sandbox.


chrismiceli (approved):

srjfoo (approved):
I didn't dig into the new user email but tested pretty much everything else. Looks good.

70ray (approved):

PR #885, opened 2023-04-07
convert range parameters to int instead of char

70ray:
This enables digits to appear in the numerical character picker.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/fix_integers

### DIFF
--- a/pinc/unicode.inc
+++ b/pinc/unicode.inc
@@ -220,7 +220,7 @@ function convert_codepoint_ranges_to_characters($codepoints)
             // we have a character range
             [$start, $end] = explode('-', $codepoint);
             $return_array = array_merge($return_array,
-                UTF8::range(UTF8::hex_to_chr($start), UTF8::hex_to_chr($end)));
+                UTF8::range(UTF8::hex_to_int($start), UTF8::hex_to_int($end)));
         } else {
             // a single character or a combined character
             $return_array[] = utf8_combined_chr($codepoint);


chrismiceli (approved):

cpeel (commented):
The `UTF8::range()` function will [also take an int](https://github.com/voku/portable-utf8#rangeintstring-var1-intstring-var2-bool-use_ctype-string-encoding-floatint-step-list) so this seems fine to me. Can you explain more about the problem you are solving so I better understand what to look for in the future?

70ray:
> The `UTF8::range()` function will [also take an int](https://github.com/voku/portable-utf8#rangeintstring-var1-intstring-var2-bool-use_ctype-string-encoding-floatint-step-list) so this seems fine to me. Can you explain more about the problem you are solving so I better understand what to look for in the future?

The digits 0-9 are missing in the character picker. What appears to be happening is that `U+0030` (etc.) is being converted by `hex_to_chr` to a string `"0"` but PHP  and range are interpreting this as a number `0` rather than a character `"0"`  so giving the range of control characters from 0 to 9. After some experimentaation I found chr_to_int() works. It converts `U+0030` to the number 48 which works correctly.

cpeel:
Ah, got it! Thanks for figuring this out!

srjfoo (commented):
I agree -- definitely good catch. (Sorry I wasn't around to do a review earlier.) And I feel embarrassed that I've just been looking at the Basic Latin character picker and making the assumption that we intentionally left the numbers out of that picker.

PR #888, opened 2023-04-09
Allow mentors, etc to view diffs in BB format

windymilla:
In order to aid feedback to mentees and similar situations provide an option on the diffs page to view the diffs as BB-code formatted text instead of the usual table view

Sandbox at: https://www.pgdp.org/~windymilla/c.branch/bbdiff

This addresses [Task #1875](https://www.pgdp.net/c/tasks.php?action=show&task_id=1875)

### DIFF
--- a/pinc/BBDiffFormatter.inc
+++ b/pinc/BBDiffFormatter.inc
@@ -1,0 +1,151 @@
+<?php
+// This file allows formatting of diffs using BB code, inherited from
+// MediaWiki difference formatter base class
+//
+// See 3rdparty/mediawiki/README.md for which MediaWiki version this maps to.
+include_once($relPath."misc.inc"); // html_safe()
+include_once($relPath."3rdparty/mediawiki/DiffFormatter.php");
+
+class BBDiffFormatter extends DiffFormatter
+{
+    protected $diffurl;
+    protected $leadingContextLines = 1;   // Code assumes 1 leading context line - displayed for add/delete only
+    protected $trailingContextLines = 0;  // Code assumes 0 trailing context lines
+
+    public function __construct($diffurl)
+    {
+        $this->diffurl = $diffurl;
+    }
+
+    // Outputs an edit block
+    // Leading context line is output before add/delete changes
+    // Assumes zero trailing context lines
+    protected function block($xbeg, $xlen, $ybeg, $ylen, &$edits)
+    {
+        // We don't display leading context line for change edits, so need to
+        // increment the starting line number for change edits with context
+        $has_context = false;
+        $has_change = false;
+        foreach ($edits as $edit) {
+            if ($edit->type == 'copy') {
+                $has_context = true;
+            } elseif ($edit->type == 'change') {
+                $has_change = true;
+            }
+        }
+        if ($has_context and $has_change) {
+            $xbeg++;
+        }
+
+        $this->startBlock($this->blockHeader($xbeg, $xlen, $ybeg, $ylen));
+
+        $leading_context = null;
+        foreach ($edits as $edit) {
+            if ($edit->type == 'copy') {
+                $leading_context = $edit->orig;
+            } elseif ($edit->type == 'add') {
+                $this->added($edit->closing, $leading_context);
+            } elseif ($edit->type == 'delete') {
+                $this->deleted($edit->orig, $leading_context);
+            } elseif ($edit->type == 'change') {
+                $this->changed($edit->orig, $edit->closing);
+            } else {
+                throw new MWException("Unknown edit type: {$edit->type}");
+            }
+        }
+        $this->endBlock();
+    }
+
+    // Called at the start of a block of connected edits.
+    protected function startBlock($header)
+    {
+        $linenum = preg_replace('/^(\d+).*/', 'Line $1', $header);
+        $difflabel = $this->make_safe(preg_replace('/.*image=(.+\.png).*/', '$1', $this->diffurl));
+        $safeurl = $this->make_safe($this->diffurl);
+        $this->writeOutput("\n----- [b]$linenum, page [url=$safeurl]${difflabel}[/url][/b] -----\n");
+    }
+
+    // Writes all (optionally prefixed) lines to the output buffer, separated by newlines
+    // and enclosed in BBCode [code] markup
+    protected function lines($lines, $prefix = '')
+    {
+        $this->writeOutput("[code]\n");
+        foreach ($lines as $line) {
+            $this->writeOutput($this->make_safe("$prefix$line\n"));
+        }
+        $this->writeOutput("[/code]\n");
+    }
+
+    // Output lines that were added by the diff
+    protected function added($lines, $leading_context = null)
+    {
+        $this->added_deleted('added', $lines, $leading_context);
+    }
+
+    // Output lines that were deleted by the diff
+    protected function deleted($lines, $leading_context = null)
+    {
+        $this->added_deleted('deleted', $lines, $leading_context);
+    }
+
+    // Output lines that were added/deleted by the diff
+    private function added_deleted($adddel_string, $lines, $leading_context)
+    {
+        if (empty($leading_context)) {  // Must be first line of the page if no context
+            $this->writeOutput("At the top of the page, ");
+        } elseif ($this->count_all_blank_lines($leading_context) > 0) {
+            $this->writeOutput("After the blank line, ");   // Context line is a blank line
+        } else {
+            $this->writeOutput("After this line:\n");
+            $this->lines($leading_context);
+        }
+
+        // If all added/deleted lines are blank, tell user about them, rather than literal output
+        $blank_lines = $this->count_all_blank_lines($lines);
+        if ($blank_lines > 1) {
+            $this->writeOutput("$blank_lines blank lines need to be $adddel_string.\n");
+        } elseif ($blank_lines === 1) {
+            $this->writeOutput("$blank_lines blank line needs to be $adddel_string.\n");
+        } else {
+            $this->writeOutput("the following needs to be $adddel_string:\n");
+            $this->lines($lines);
+        }
+    }
+
+    // Count blank lines in given array if they are all blank
+    //
+    private function count_all_blank_lines($lines)
+    {
+        foreach ($lines as &$line) {
+            if (!empty($line)) {
+                return 0;
+            }
+        }
+        return count($lines);
+    }
+
+    // Output lines that were changed by the diff
+    protected function changed($orig, $closing)
+    {
+        $this->lines($orig);
+        $this->writeOutput("is proofread as\n");
+        $this->lines($closing);
+    }
+
+    // Make string safe for this type of output
+    // No changes needed for BBCode output
+    protected function make_safe($string)
+    {
+        return $string;
+    }
+}
+
+// Makes BBCode diffs suitable for displaying within HTML
+class BBDiffFormatterHTML extends BBDiffFormatter
+{
+    // Make string safe for HTML output
+    protected function make_safe($string)
+    {
+        return html_safe($string);
+    }
+}
--- a/pinc/DifferenceEngineWrapper.inc
+++ b/pinc/DifferenceEngineWrapper.inc
@@ -25,12 +25,20 @@ include_once($relPath."3rdparty/mediawiki/DairikiDiff.php");
 include_once($relPath."3rdparty/mediawiki/WordLevelDiff.php");
 include_once($relPath."3rdparty/mediawiki/DiffEngine.php");
 include_once($relPath."3rdparty/mediawiki/DiffFormatter.php");
-include_once($relPath."3rdparty/mediawiki/TableDiffFormatter.php");
 
-class DifferenceEngineWrapper
+abstract class DifferenceEngineWrapper
 {
     private $oldText;
     private $newText;
+    private $formatter;
+
+    protected function __construct(DiffFormatter $formatter = null)
+    {
+        if ($formatter === null) {
+            $formatter = new DiffFormatter();
+        }
+        $this->formatter = $formatter;
+    }
 
     public function setText($oldText, $newText)
     {
@@ -38,7 +46,7 @@ class DifferenceEngineWrapper
         $this->newText = $newText;
     }
 
-    public function showDiff($oldTitle, $newTitle)
+    public function getDiff($oldTitle, $newTitle)
     {
         $otext = $this->oldText;
         $ntext = $this->newText;
@@ -48,7 +56,7 @@ class DifferenceEngineWrapper
         $difftext = $this->localiseLineNumbers($difftext);
 
         // from DifferenceEngine::getDiff()
-        echo $this->addHeader($difftext, $oldTitle, $newTitle);
+        return $this->addHeader($difftext, $oldTitle, $newTitle);
     }
 
     private function generateTextDiffBody($otext, $ntext)
@@ -65,54 +73,15 @@ class DifferenceEngineWrapper
         $ota = explode("\n", $wgContLang->segmentForDiff($otext));
         $nta = explode("\n", $wgContLang->segmentForDiff($ntext));
         $diffs = new Diff($ota, $nta);
-        $formatter = new TableDiffFormatter();
-        $difftext = $wgContLang->unsegmentForDiff($formatter->format($diffs));
+        $difftext = $wgContLang->unsegmentForDiff($this->formatter->format($diffs));
 
         return $difftext;
     }
 
-    private function addHeader($diff, $otitle, $ntitle, $multi = '', $notice = '')
+    // Default implementation adds nothing
+    protected function addHeader($diff, $otitle, $ntitle, $multi = '', $notice = '')
     {
-        // lifted from DifferenceEngine::addHeader()
-        // but without any of the $userLang bits
-
-        $tableClass = 'diff';
-        $header = "<table class='$tableClass'>";
-
-        if (!$diff && !$otitle) {
-            $header .= "
-            <tr class='diff-title'>
-            <td class='diff-ntitle'>{$ntitle}</td>
-            </tr>";
-            $multiColspan = 1;
-        } else {
-            if ($diff) { // Safari/Chrome show broken output if cols not used
-                $header .= "
-                <col class='diff-marker' />
-                <col class='diff-content' />
-                <col class='diff-marker' />
-                <col class='diff-content' />";
-                $colspan = 2;
-                $multiColspan = 4;
-            } else {
-                $colspan = 1;
-                $multiColspan = 2;
-            }
-            $header .= "
-            <tr class='diff-title'>
-            <td colspan='$colspan' class='diff-otitle'>{$otitle}</td>
-            <td colspan='$colspan' class='diff-ntitle'>{$ntitle}</td>
-            </tr>";
-        }
-
-        if ($multi != '') {
-            $header .= "<tr><td colspan='{$multiColspan}' class='diff-multi'>{$multi}</td></tr>";
-        }
-        if ($notice != '') {
-            $header .= "<tr><td colspan='{$multiColspan}' class='diff-notice'>{$notice}</td></tr>";
-        }
-
-        return $header . $diff . "</table>";
+        return $diff;
     }
 
     private function localiseLineNumbers($text)
@@ -167,32 +136,4 @@ if (!function_exists("wfDebug")) {
     {
         // don't log any debug messages
     }
-}
-
-// Override default css for DP customizations
-function get_DifferenceEngine_css_data()
-{
-    [, $font_size, $font_family] = get_user_proofreading_font();
-    if ($font_size != '') {
-        $font_size = "font-size: $font_size;";
-    }
-
-    return "
-.diff-otitle,
-.diff-ntitle {
-    font-weight: bold;
-}
-.diff-marker,
-.diff-addedline,
-.diff-deletedline,
-.diff-context {
-    font-family: $font_family;
-    $font_size
-}
-/* Adjust padding to prevent descenders from being chopped off. Task 1936 */
-.diff-deletedline .diffchange,
-.diff-addedline .diffchange {
-    padding: 0.1em 0;
-}
-";
 }
--- a/pinc/DifferenceEngineWrapper.inc
+++ b/pinc/DifferenceEngineWrapper.inc
@@ -18,7 +18,6 @@
 //
 // See 3rdparty/mediawiki/README.md for which MediaWiki version this maps to.
 include_once($relPath."misc.inc"); // html_safe() in following includes
-include_once($relPath."prefs_options.inc"); // get_user_proofreading_font()
 include_once($relPath."3rdparty/mediawiki/ComplexityException.php");
 include_once($relPath."3rdparty/mediawiki/WordAccumulator.php");
 include_once($relPath."3rdparty/mediawiki/DairikiDiff.php");
--- a/pinc/DifferenceEngineWrapperBB.inc
+++ b/pinc/DifferenceEngineWrapperBB.inc
@@ -1,0 +1,34 @@
+<?php
+// This file inherits from DifferenceEngineWrapper to facilitate
+// formatting diffs using BB code.
+include_once($relPath."DifferenceEngineWrapper.inc");
+include_once($relPath."BBDiffFormatter.inc");
+
+// Format diffs using BB code, without HTML markup
+class DifferenceEngineWrapperBB extends DifferenceEngineWrapper
+{
+    public function __construct($diffurl)
+    {
+        parent::__construct(new BBDiffFormatter($diffurl));
+    }
+}
+
+// Format diffs using BB code, displayed within HTML markup
+class DifferenceEngineWrapperBBHTML extends DifferenceEngineWrapper
+{
+    public function __construct($diffurl)
+    {
+        parent::__construct(new BBDiffFormatterHTML($diffurl));
+    }
+
+    protected function addHeader($diff, $otitle, $ntitle, $multi = '', $notice = '')
+    {
+        $header = "";
+        if (!$diff && !$otitle) {
+            $header .= $ntitle;
+        } else {
+            $header .= "\n$otitle<br>\n$ntitle";
+        }
+        return "$header\n<pre>$diff</pre>\n";
+    }
+}
--- a/tools/project_manager/diff.php
+++ b/tools/project_manager/diff.php
@@ -175,11 +175,8 @@ echo "</p>\n";
 $diffurl = "$code_url/tools/project_manager/diff.php?project=$projectid&image=$image&L_round_num=$L_round_num&R_round_num=$R_round_num";
 
 $diffEngine = $bb_diffs ? new DifferenceEngineWrapperBBHTML($diffurl) : new DifferenceEngineWrapperTable();
-
 $diffEngine->setText($L_text, $R_text);
-$difftext = $diffEngine->getDiff($L_label, $R_label);
-
-echo $difftext;
+echo $diffEngine->getDiff($L_label, $R_label);
 
 // don't print out the navigation bit again if there is no difference
 // at the top of the page it's buttons, then project page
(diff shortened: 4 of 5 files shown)


tools/project_manager/diff.php
```diff
@@ -138,11 +141,26 @@
...
+    $bb_diffs_url = "?" . attr_safe(http_build_query($_GET));
+    echo "| <a href='$bb_diffs_url'>$bb_diffs_label</a>\n";
+}
+
 echo "</p>\n";
 
 // ---------------------------------------------------------
 
-$diffEngine = new DifferenceEngineWrapper();
+$diffurl = $_SERVER["REQUEST_SCHEME"]."//".$_SERVER['HTTP_HOST'].$_SERVER["PHP_SELF"];
```
cpeel, line 161: Since the HTML will be rendered on this page, we should be able to use the relative URL (just the line below) and not need to reconstruct the full URL. Using relative URLs when we can is preferred. When we can't we use `$code_url` global instead (eg: `$code_url/tools/project_manager/diff.php`).

tools/project_manager/diff.php
```diff
@@ -138,11 +141,26 @@
     echo "| <a href='$format_url'>$format_label</a>\n";
 }
 
+if (user_can_mentor_in_any_round()) {
+    // show option to change diff style
+    if ($bb_diffs) {
+        $bb_diffs_label = _("Table diff style");
```
cpeel, line 147: Should we make this text more action-centric since they're links?
`Show table-style diff` and `Show BBCode-style diff`

tools/project_manager/diff.php
```diff
@@ -7,6 +7,7 @@
 include_once($relPath.'links.inc');
 include_once($relPath.'misc.inc'); // get_integer_param(), attr_safe()
 include_once($relPath."DifferenceEngineWrapper.inc");
+include_once($relPath."DifferenceEngineWrapperBB.inc");
```
cpeel, line 10: Your PR is missing at least the `DifferenceEngineWrapperBB.inc` file 😁

tools/project_manager/diff.php
windymilla: It does need the full URL in this case, since it's not rendering on this page. It's passing the diffurl down into the DiffEngineWrapper and thence to the DiffFormatter. From there, it will be output into the BB-style diffs as `[url=$diffurl]001.png[/url]`, i.e. the URL is included in the PM to the mentee, so the full URL is needed. I've changed it back to use `$code_url` (as I had in my noncvs experiment). It's good to know about the relative URL thing though, thanks.

tools/project_manager/diff.php
windymilla: Oops - thanks.

70ray:
Since I have never encountered BB-code before, could you say how it would be useful or what difficulties it overcomes?

windymilla:
> Since I have never encountered BB-code before, could you say how it would be useful or what difficulties it overcomes?

BBcode is the markup used in DP forum messages. When a mentor (for example) wants to give feedback to a mentee via a PM, they frequently want to include the diffs, e.g. the mentee failed to correct a scanno, and the proofer in the next round has corrected it. The mentor therefore wants to send a message something like

> ----- Line 23, 001.png
> `books arid magazines`
> ...is proofread as...
> `books and magazines`

Copying and pasting the diffs from the standard table of diffs is not easy, and prone to error, and even when the text has been extracted, the mentor still wants to mark up the diffs in mono-spaced font and add a link to the regular diffs page that the mentee can then visit to see it in more detail/context.

So this PR gives an alternative style for the diffs, with the BBCode markup `[pre]` around the text and `[url]` around the link. The link to swap between diff styles will only be visible for users who are allowed to mentor (plus SA/PFs)

BTW, I realised hadn't put a link to [Task #1875](https://www.pgdp.net/c/tasks.php?action=show&task_id=1875) in the PR description at the top, so I'll add that too.

tools/project_manager/diff.php
windymilla: By the way, the `DiffFormatter` include file in `pinc/3rdparty/mediawiki` has a `.php` extension, so originally, I gave my `BBDiffFormatter` which extends the base class a `.php` extension too. One of the security checks then failed, and I guessed it was because `.php` files are checked differently to `.inc`. I therefore renamed my include file to `BBDiffFormatter.inc`. Hope that's OK - if not, what do I need to do to make it pass the security test if it's an include file but with a `.php` extension.

windymilla:
I think I've addressed the points raised so far, and it is OK for testing again now. Sandbox https://www.pgdp.org/~windymilla/c.branch/bbdiff has been updated

tools/project_manager/diff.php
cpeel: Ah, but the *entry* page is `diff.php` and that's what the link will be submitting to -- the web browser doesn't know about anything else being included. I really think you can get away with a relative URL.

pinc/DifferenceEngineWrapperBB.inc
```diff
@@ -0,0 +1,29 @@
...
+    protected function addHeader($diff, $otitle, $ntitle, $multi = '', $notice = '')
+    {
+        $header = "";
+
+        if (!$diff && !$otitle) {
+            $header .= "{$ntitle}";
+        } else {
+            $header .= "
+            {$otitle}<br>
+            {$ntitle}";
```
cpeel, line 25: Can we make this and line 27 consistent? Either building the string all in one go and having implicit newlines or doing it piecemeal and being explicit about newlines -- either is fine but the mix is confusing. In general the code is explicit about newlines but we have a mix of building strings piecemeal or doing it all in one go, so whatever you want to do there is fine.

eg:
```php
$header .= "\n$otitle<br>\n$ntitle";
return "$header\n<pre>$diff</pre>\n";
// or
$header .= "\n{$otitle}<br>\n{$ntitle}";
return "{$header}\n<pre>{$diff}</pre>\n";
// or
$header .= "\n" . $otitle . "<br>\n" . $ntitle";
return $header . "\n<pre>" . $diff . "</pre>\n";
```

tools/project_manager/diff.php
cpeel: Sorry, I didn't read your message clearly. Yup, I see now that we need to include the absolute URL and what you have here is good!

pinc/BBDiffFormatter.inc
```diff
@@ -0,0 +1,74 @@
...
+    {
+        $pre_done = false;
+        foreach ($lines as $line) {
+            if ($pre_done) {
+                $this->writeOutput("\n");
+            } else {
+                $this->writeOutput("[pre]");
+                $pre_done = true;
+            }
+            $this->writeOutput(html_safe("$prefix${line}"));
```
cpeel, line 38: Either `$prefix$line` or `{$prefix}{$line}` please.

tools/project_manager/diff.php
```diff
@@ -138,11 +141,25 @@
     echo "| <a href='$format_url'>$format_label</a>\n";
 }
 
+if (user_can_mentor_in_any_round()) {
+    // show option to change diff style
+    if ($bb_diffs) {
+        $bb_diffs_label = _("Show table-style diff");
+        $_GET["bb_diffs"] = "off";
```
chrismiceli, line 148: is writing to `$_GET` good practice? I assumed it was read only until now.

tools/project_manager/diff.php
```diff
@@ -138,11 +141,25 @@
...
+    } else {
+        $bb_diffs_label = _("Show BBCode-style diff");
+        $_GET["bb_diffs"] = "on";
+    }
+    $bb_diffs_url = "?" . attr_safe(http_build_query($_GET));
+    echo "| <a href='$bb_diffs_url'>$bb_diffs_label</a>\n";
```
chrismiceli, line 154: personally I get super confused when the text I see if for something I'm not currently seeing. For example, when viewing "table-style diff", the button says BB style and vice-versa. Perhaps we could use a dropdown, that shows your current mode so it is clear what you are looking at? Or at least indicate the format in a heading somewhere?

chrismiceli (commented):
Since this code probably is meant to be read, maybe we add a copy to clipboard button to facilitate copying this to be pasted in a forum?

tools/project_manager/diff.php
windymilla: I copied it from the code directly above that sets the "With/Without Formatting" flag. I don't know if it's good practice.

tools/project_manager/diff.php
```diff
@@ -138,11 +141,25 @@
...
+}
+
 echo "</p>\n";
 
 // ---------------------------------------------------------
 
-$diffEngine = new DifferenceEngineWrapper();
+$diffurl = "$code_url/tools/project_manager/diff.php?project=$projectid&amp;image=$image&amp;L_round_num=$L_round_num&amp;R_round_num=$R_round_num";
```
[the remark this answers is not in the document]
windymilla: `$image_url` and `$url` in this file use `&amp;` and I followed blindly. If you think the context argument means that this url should be treated differently, I can change it. I'm assuming you mean that I would pass the unencoded version into the constructor, then encode it later when I'm writing it to the webpage?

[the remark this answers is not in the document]
chrismiceli: My concern is this code doesn't know in what context this string will be used. If it is being output as HTML somewhere, that is where the encoding should be. For all we know at this point, this string may be shoved in a SQL statement, in which case its encoding would need to be different. So either we encode when using it or name it so the encoding is clear to consumers. I think $image_url and $url encode premeptively above because they are used at least within the same file, but even that I would think we should do differently to be honest.

tools/project_manager/diff.php
```diff
@@ -169,6 +186,9 @@ function do_navigation($projectid, $image, $L_round_num, $R_round_num, $L_user_c
     $navigation_text .= "\n<input type='hidden' name='L_round_num' value='$L_round_num'>";
     $navigation_text .= "\n<input type='hidden' name='R_round_num' value='$R_round_num'>";
     $navigation_text .= "\n<input type='hidden' name='format' value='$format'>";
+    if ($bb_diffs) {
```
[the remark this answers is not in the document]
windymilla: I may be wrong, but does that mean the URLs linked to in the navigation buttons will then always have the `bb_diffs` parameter with its value either on or off? A reason for not doing that is that general visibility of this feature is restricted to mentor-types. I think the way I currently have it means a non-mentor won't see that parameter in the URL when they use a button like `Proofreader Next`. Sorry if I've misunderstood what you are suggesting.

tools/project_manager/diff.php
windymilla: The button actually says "Show BBCode-style diff" to make it clear that clicking the link will show BBCode, rather than being a description of what is currently shown. This is the same as the adjacent link, which says "Compare without formatting" when the formatting is being shown. Given that there is currently a checkbox for "Skip empty diffs", and a link for "Compare without formatting", using a dropdown for the BBCode style would be a third method of essentially displaying a toggle switch, all of which are in the same area of the page. 
One thing about the BBCode switch is that it's instantly obvious which style of diffs you are looking at, since they are visually so different. That means it's different to the "without formatting" or the "skip empty diffs" one, where you might not notice which way the flag is set. 
I'm happy to go with the majority though. It might also be worth bearing in mind that the target audience hasn't seen it in the current form yet, so may have some usability suggestions.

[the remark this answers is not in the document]
windymilla: Thanks

tools/project_manager/diff.php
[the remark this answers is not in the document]
windymilla: Thanks - I hadn't considered that before you raised it

tools/project_manager/diff.php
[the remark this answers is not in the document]
chrismiceli: I wouldn't worry about it being present for non-mentors since it won't have any impact and it will simplify this form code some. Just my opinion, may want to see what others say.

chrismiceli:
Is it worth it to add php unit tests in any capacity?

windymilla:
> Since this code probably is meant to be read, maybe we add a copy to clipboard button to facilitate copying this to be pasted in a forum?

@chrismiceli - did you mean that the text should remain as it is (i.e. with BBCode markup) and a button copy that to the clipboard, or did you mean that the text should be displayed without markup and the copy to clipboard button would add the markup as it copies the text to the clipboard?

That also raises another possibility...scrap pretty much everything I've done, and just add a "Copy diffs to clipboard with BBCode" to the normal diffs page. So, diffs would continue to be shown as they currently are, but when you click the copy to clipboard button it would convert them to the BBCode style as it copies to the clipboard.

windymilla:
> Is it worth it to add php unit tests in any capacity?

Possibly - that's still an area I know nothing about

70ray (commented):
If the lines are very long then a scroll bar appears at the bottom which scrolls the whole window including the header: file:///home/ray/Pictures/Screenshot_20230410_204942.png
Perhaps there should be an enclsing div so only the inside part scrolls.

chrismiceli:
> > Since this code probably is meant to be read, maybe we add a copy to clipboard button to facilitate copying this to be pasted in a forum?
> 
> @chrismiceli - did you mean that the text should remain as it is (i.e. with BBCode markup) and a button copy that to the clipboard, or did you mean that the text should be displayed without markup and the copy to clipboard button would add the markup as it copies the text to the clipboard?
> 
> That also raises another possibility...scrap pretty much everything I've done, and just add a "Copy diffs to clipboard with BBCode" to the normal diffs page. So, diffs would continue to be shown as they currently are, but when you click the copy to clipboard button it would convert them to the BBCode style as it copies to the clipboard.

I just meant a button in addition to what you have here. But really we should probably get feedback on intended usage patterns.

pinc/DifferenceEngineWrapperBB.inc
```diff
@@ -0,0 +1,24 @@
+<?php
+// This file inherits from DifferenceEngineWrapper to facilitate
+// formatting diffs using BB code.
+include_once($relPath."DifferenceEngineWrapper.inc");
+include_once($relPath."BBDiffFormatter.inc");
+
+class DifferenceEngineWrapperBB extends DifferenceEngineWrapper
```
70ray, line 7: This seems unsymmetrical somehow. Alternatively you could make DifferenceEngineWrapper into an abstract class and extend DifferenceEngineWrapperBB and DifferenceEngineWrapperTable from it. They would have their own constructors and definitions of addHeader() and everything else derived from the parent.

tools/project_manager/diff.php
cpeel: It's not read only and it's not a great practice but it's better to be consistent with the other instance in this file (like you have it).

tools/project_manager/diff.php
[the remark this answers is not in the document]
cpeel: I thought about this too, but I think Nigel's approach of not having the param for users who can't see it is the better approach IMHO. Because one of our eagle-eyed users is going to see this param and then try changing it and ask about it, and while it doesn't reveal anything they *can't* see, it's just more hassle than hiding it to begin with.

70ray:
> > Since this code probably is meant to be read, maybe we add a copy to clipboard button to facilitate copying this to be pasted in a forum?
> 
> @chrismiceli - did you mean that the text should remain as it is (i.e. with BBCode markup) and a button copy that to the clipboard, or did you mean that the text should be displayed without markup and the copy to clipboard button would add the markup as it copies the text to the clipboard?
> 
> That also raises another possibility...scrap pretty much everything I've done, and just add a "Copy diffs to clipboard with BBCode" to the normal diffs page. So, diffs would continue to be shown as they currently are, but when you click the copy to clipboard button it would convert them to the BBCode style as it copies to the clipboard.

Another possibility would be to make the BBCodestyle look like the processed BBCode in the PM preview and add a copy button as Chris suggested which would copy the source BBCode. I don't know how difficult this would be, perhaps just converting [pre] to <pre> etc. would do it.

70ray:
> If the lines are very long then a scroll bar appears at the bottom which scrolls the whole window including the header ...  Perhaps there should be an enclosing div so only the inside part scrolls.

This same effect occurs in other places on the site so probably no need to worry about it.

pinc/DifferenceEngineWrapperBB.inc
windymilla: Done - it also led me to a way of easily providing the suggested Copy to Clipboard button. Thanks

windymilla:
> Another possibility would be to make the BBCodestyle look like the processed BBCode in the PM preview and add a copy button as Chris suggested which would copy the source BBCode. I don't know how difficult this would be, perhaps just converting [pre] to` \<pre\>` etc. would do it.

I haven' t done this for now, though the new class structure easily allows for it. The intended users may have feedback on the visual appearance

tools/project_manager/diff.php
```diff
@@ -135,17 +138,44 @@
...
+}
+
+$diffurl = "$code_url/tools/project_manager/diff.php?project=$projectid&image=$image&L_round_num=$L_round_num&R_round_num=$R_round_num";
+
+// Provide button to copy BBCode diffs not wrapped in HTML
+if ($bb_diffs) {
+    $diffEngineCopy = new DifferenceEngineWrapperBB($diffurl);
+    $diffEngineCopy->setText($L_text, $R_text);
+    $cliptext = javascript_safe($diffEngineCopy->getDiff($L_label, $R_label));
+    echo "| \n<input type='button' value='" . attr_safe(_("Copy BBCode to clipboard")) . "' onClick='navigator.clipboard.writeText(\"$cliptext\");'>";
```
chrismiceli, line 166: `writeText` returns a promise while the user interacts with the permission dialog. If they reject the request or have turned off clipboard access altogether, that promise is rejected. We may want to in that case alert the user the copy failed.
```js
navigator.clipboard.writeText(\"$cliptext\").catch(function () {alert("Unable to copy!");});'
```
but that is getting a bit much. Maybe at this point we inline some js at the top with a click handler and just reference it here in the attribute.

tools/project_manager/diff.php
```diff
@@ -135,17 +138,44 @@
...
+    echo "| <a href='$bb_diffs_url'>$bb_diffs_label</a>\n";
+}
+
+$diffurl = "$code_url/tools/project_manager/diff.php?project=$projectid&image=$image&L_round_num=$L_round_num&R_round_num=$R_round_num";
+
+// Provide button to copy BBCode diffs not wrapped in HTML
+if ($bb_diffs) {
+    $diffEngineCopy = new DifferenceEngineWrapperBB($diffurl);
+    $diffEngineCopy->setText($L_text, $R_text);
+    $cliptext = javascript_safe($diffEngineCopy->getDiff($L_label, $R_label));
```
chrismiceli, line 165: Instead of passing the text directly, maybe just have a function that reads the diff from the DOM so you don't need to encode at all.

tools/project_manager/diff.php
chrismiceli: > @chrismiceli - I added
> 
> ```
>     "js_data" => "
>         function copyToClip(textstring) {
>             navigator.clipboard.writeText(textstring).catch(function () {alert('Unable to copy!');});
>         }
>     ",
> ```
> 
> to `$extra_args` and changed the onClick to
> 
> ```
> onClick='copyToClip(\"$cliptext\");'
> ```
> 
> Is that the kind of thing you had in mind? I am unable to reliably get my browser to reject use of clipboard, but have seen the "Unable to copy" alert. If you are able to control clipboard use easily on your system, this edit is in my [main sandbox](https://www.pgdp.org/~windymilla/c/) but not committed to the bbdiff branch yet if you're able to try it out, or suggest something different. Thanks

That's better. We'd need to encode the $cliptext though for the html attribute context to confirm no characters will escape out of the attribute like ".

tools/project_manager/diff.php
chrismiceli: If we're showing the bbcode, I meant grab it from the < pre > tag in the DOM maybe in the function like
```js
navigator.clipboard.writeText(document.getElementById('#preTag').textcontent).catch(function () {
  alert("Unable to copy!");
});
```

tools/project_manager/diff.php
windymilla: Doesn't `javascript_safe` already do that to the cliptext before it gets passed in as an argument to `copyToClip`? Or did you mean something else?

tools/project_manager/diff.php
windymilla: Thanks for the code snippet - I agree that if we're showing the unmodified bbcode and also not putting the button on the table-style diffs page, that is much neater. From initial user feedback, it may be that the Copy everything to Clipboard button is not really useful for what they do, which is to copy diffs individually into a message template. I'll solicit more feedback after changing/fixing a couple of things, to help make that decision.

tools/project_manager/diff.php
chrismiceli: javascript_safe is for encoding things to be used in a javascript context. Html attribute safe if for things going in html attributes. You are doing both here. This is why I was suggesting we pull the diff from the DOM to avoid all encoding issues.

windymilla: Ah yes. Well if the user feedback means we end up being able to use the DOM idea, that will clearly be better. It still might happen that we don't have the button at all. Depending on what others think, the first feedback I got was that they wouldn't want all the diffs in one block in the clipboard, just manually selecting one at a time. We'll see - thanks for continuing to educate me :)

tools/project_manager/diff.php
```diff
@@ -127,6 +136,7 @@
 
 if ($L_format || $R_format) {
     // show option to change compare method
+    $format_save = array_key_exists("format", $_GET) ? $_GET["format"] : $format;
```
cpeel, line 139: `array_get()` in `misc.inc` is a bit clearer for this:
```php
$format_save = array_get($_GET, "format", $format);
```

tools/project_manager/diff.php
```diff
@@ -135,17 +145,49 @@
...
     $format_url = "?" . attr_safe(http_build_query($_GET));
+    $format_label = html_safe($format_label);
     echo "| <a href='$format_url'>$format_label</a>\n";
+    // Restore $_GET in case needed below
+    $_GET["format"] = $format_save;
+}
+
+if (user_can_mentor_in_any_round()) {
+    // show option to change diff style
+    $bb_diffs_save = array_key_exists("bb_diffs", $_GET) ? $_GET["bb_diffs"] : "off";
```
cpeel, line 156: Ditto `array_get()` here.

tools/project_manager/diff.php
```diff
@@ -135,17 +145,49 @@
...
+    $bb_diffs_label = html_safe($bb_diffs_label);
+    echo "| <a href='$bb_diffs_url'>$bb_diffs_label</a>\n";
+    // Restore $_GET in case needed below
+    $_GET["bb_diffs"] = $bb_diffs_save;
+}
+
+$diffurl = "$code_url/tools/project_manager/diff.php?project=$projectid&image=$image&L_round_num=$L_round_num&R_round_num=$R_round_num";
+
+// Provide button to copy BBCode diffs not wrapped in HTML
+if ($bb_diffs or true) { // ALWAYS SHOW BUTTON
```
cpeel, line 174: Leftover for testing?

cpeel (commented):
I'm not 100% sure where this is in the user-testing phase. I'm happy with whatever y'all decide is best from the UX side of things, so let me know when this is ready for a final code review and to go in.

windymilla:
> I'm not 100% sure where this is in the user-testing phase. I'm happy with whatever y'all decide is best from the UX side of things, so let me know when this is ready for a final code review and to go in.

Still in the user-testing phase at the moment, thanks - as you saw, I left a couple of things untidy as they may be removed completely or need tidying depending on user feedback. I'll fix the `array_get()` suggestions of course when I do the other tidying.

I'll be sure to let people know when it's ready for final review - thanks

windymilla:
For user testing - no need for code review ATM

windymilla:
@cpeel & others - thanks for feedback and suggestions. The behavior of the code now appears to match what the users would like, and has had some testing from @srjfoo too. So, it's ready for a code review from whoever is willing - thanks!

srjfoo (approved):
I'm happy to do more testing as needed. I don't understand this part of the code well enough to be able to give a good code review on it, but as far as it working as expected, it does as far as I can tell.

pinc/DifferenceEngineWrapper.inc
```diff
@@ -25,20 +25,28 @@ include_once($relPath."3rdparty/mediawiki/DairikiDiff.php");
 include_once($relPath."3rdparty/mediawiki/WordLevelDiff.php");
 include_once($relPath."3rdparty/mediawiki/DiffEngine.php");
 include_once($relPath."3rdparty/mediawiki/DiffFormatter.php");
-include_once($relPath."3rdparty/mediawiki/TableDiffFormatter.php");
```
cpeel, line 28: Line 21 (pref_options.inc include) needs to be moved into `pinc/DifferenceEngineWrapperTable.inc` as that's where this function is used.

tools/project_manager/diff.php
```diff
@@ -135,17 +145,41 @@
...
 // ---------------------------------------------------------
 
-$diffEngine = new DifferenceEngineWrapper();
+$diffurl = "$code_url/tools/project_manager/diff.php?project=$projectid&image=$image&L_round_num=$L_round_num&R_round_num=$R_round_num";
+
+$diffEngine = $bb_diffs ? new DifferenceEngineWrapperBBHTML($diffurl) : new DifferenceEngineWrapperTable();
 
 $diffEngine->setText($L_text, $R_text);
-$diffEngine->showDiff($L_label, $R_label);
+$difftext = $diffEngine->getDiff($L_label, $R_label);
```
cpeel, line 180: You might consider just echoing here instead of storing it in a variable and then writing it out.

cpeel (approved):
Two minor comments and this LGTM!

tools/project_manager/diff.php
windymilla: Thanks - should be OK now

windymilla:
> > I'm happy to do more testing as needed. I don't understand this part of the code well enough to be able to give a good code review on it, but as far as it working as expected, it does as far as I can tell.
> 
> One concern I would have is if the diff had some special characters (html, javascript, &, ", etc.) that caused issues when outputting the HTML or javascript parts of the UI.

Thanks, @chrismiceli. 

To check the characters were being escaped properly, I added the following strings to a page: `&amp;`, `<i>`, `</pre>`. They were reported correctly (as were double quotes) and weren't interpreted as HTML entities/commands. 

See the first diff on [this bage](https://www.pgdp.org/~windymilla/c.branch/bbdiff/tools/project_manager/diff.php?project=projectID567f1ea6a1ba6&image=003.png&L_round_num=4&R_round_num=5&format=keep&jumpto=001.png&only_nonempty_diffs=on&bb_diffs=on). Is that OK?

srjfoo:
> > > I'm happy to do more testing as needed. I don't understand this part of the code well enough to be able to give a good code review on it, but as far as it working as expected, it does as far as I can tell.
> > 
> > 
> > One concern I would have is if the diff had some special characters (html, javascript, &, ", etc.) that caused issues when outputting the HTML or javascript parts of the UI.
> 
> Thanks, @chrismiceli.
> 
> To check the characters were being escaped properly, I added the following strings to a page: `&amp;`, `<i>`, `\</pre\>`. They were reported correctly (as were double quotes) and weren't interpreted as HTML entities/commands.
> 
> See the first diff on [this page](https://www.pgdp.org/~windymilla/c.branch/bbdiff/tools/project_manager/diff.php?project=projectID567f1ea6a1ba6&image=003.png&L_round_num=4&R_round_num=5&format=keep&jumpto=001.png&only_nonempty_diffs=on&bb_diffs=on). Is that OK?

And following up on @windymilla's test: see [this page](https://www.pgdp.org/~windymilla/c.branch/bbdiff/tools/project_manager/diff.php?project=projectID4c4a429bed548&image=038.png&L_round_num=4&R_round_num=5&format=keep&jumpto=039.png&bb_diffs=on).

chrismiceli (approved):

Issue #561, opened 2021-01-16
Investigate HTML emails


cpeel:
We should investigate having the site send out HTML emails to look more professional. This should be much easier after we start using composer (https://github.com/DistributedProofreaders/dproofreaders/issues/557).

One possible solution would be to support sending markdown-formatted emails. The rendered version could be used for the HTML version and the text version for the non-HTML version. That would dramatically simplify creating and sending HTML-formatted emails. It should also allow us to gradually start rolling out HTML emails rather than needing to update all the emails in the code at once.

This is probably also the time to use a robust mailing package like [PHPMailer](https://github.com/PHPMailer/PHPMailer).

cpeel:
Fixed in #890

PR #890, opened 2023-04-15
Begin Markdown/HTML/PHPMailer emails

windymilla:
Email body text will use Markdown and email will be sent as HTML along with text version if user's email doesn't display HTML.

PHPMailer is used to send the email instead of PHP's `mail`.

~~Existing function `maybe_mail` remains unchanged, and new function `maybe_email` implements Markdown/PHPMailer changes. This allows for piecemeal conversion of emails to Markdown/HTML format. New function is similar but takes separate From/Reply-to arguments instead of general additional header lines.~~

~~This commit converts the emails used to notify event subscribers and displays the round/phase in italics. Future work to convert other emails needs each email to be checked, especially for paragaraph breaks or line breaks (by adding double space at end of line - see edit to `project.inc`.~~

~~PHPMailer is only implemented using `localhost` on port 2025, and will need the production settings in a later commit.~~


windymilla:
Testing of email can be done with https://github.com/tomwardill/FakeEmail
`pip install fakeemail` installs FakeEmail
`fakeemail 2025 8080 0.0.0.0` runs it to listen on port 2025 and provide web interface on port 8080
Open `http://pgdp.org:8080/` in a browser to see emails.

windymilla:
Sandbox: https://www.pgdp.org/~windymilla/c.branch/email

pinc/maybe_mail.inc
```diff
@@ -4,6 +4,135 @@ include_once($relPath.'misc.inc'); // html_safe()
...
+{
+    $string = rtrim($string, '>');
+    $results = explode(' <', $string);
+    if (sizeof($results) === 1) {
+        array_unshift($results, null);
+    }
+    return $results;
+}
+
 function maybe_mail($to, $subject, $message, $additional_headers = null)
```
cpeel, line 136: I'd love for us to think if there's a way we can make `maybe_mail()` end up being a wrapper around the new function so we only have one function that actually sends email.

It looks like there are only two places where we use `$additional_headers` for something besides `$from` and that's now supported by your new function call: `notify_old_pp.php` and `access_change_emails.inc`:
```php
    $headers = ["Reply-To: $db_requests_email_addr"];
    $mail_accepted = maybe_mail($email, $subject, $message_string, $headers);
```

It would mean that the new function would need to know the format of `$message` or we have a different way of enabling that via a couple of different functions.

pinc/maybe_mail.inc
```diff
@@ -4,6 +4,135 @@ include_once($relPath.'misc.inc'); // html_safe()
...
+    }
+
+    if ($mail_mode === 0) {
+        return success;    // Nothing left to do!
+    }
+
+    $mail = new PHPMailer(true);
+    try {
+        // Server settings
+        $mail->isSMTP();
```
cpeel, line 82: Is it possible to have PHPMailer use the PHP `mail()` function to send email? If it is, that might be a good first version as we know that already works at pgdp.net and we don't have to wire through additional email settings.

cpeel (commented):
Definitely going in the right direction! Some overall thoughts but not a low-level review.

pinc/maybe_mail.inc
```diff
@@ -4,6 +4,135 @@ include_once($relPath.'misc.inc'); // html_safe()
 include_once($relPath.'Project.inc');
 include_once($relPath.'User.inc');
 
+use PHPMailer\PHPMailer\PHPMailer;
+use PHPMailer\PHPMailer\SMTP;
+use PHPMailer\PHPMailer\Exception;
+
+function maybe_email($to, $subject, $message, $from = null, $reply_to = null)
```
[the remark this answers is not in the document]
windymilla: Good idea!

pinc/maybe_mail.inc
```diff
@@ -4,6 +4,135 @@ include_once($relPath.'misc.inc'); // html_safe()
...
+        echo "$html_message\n"   ;
+        echo "<hr>\n";
+        // Uncomment below to test emails using localhost:2025
+        $mail_mode = -1;
+    } else {
+        $mail_mode = 1;
+    }
+
+    if ($mail_mode === 0) {
+        return success;    // Nothing left to do!
```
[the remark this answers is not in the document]
windymilla: Oops - thought I fixed that. Maybe I got caught (again) by the lack of interaction between NP++ remote editing and running php-cs-fixer. Either, after I run the fixer, I force-load the file back into NP++, and lose an edit I made there, or I end up saving a non-fixered version from NP++ back onto the server and get lint errors when I commit. I just need to remember to follow procedure carefully.

pinc/maybe_mail.inc
windymilla: The new function supports `$from` and `$reply-to` - is there anything else that's needed?
Instead of making maybe_mail a wrapper, might it be easier to just fix the text of all the emails so they can all go through the new function - it didn't seem like there are too many to deal with. Then maybe_mail would be deleted, i.e. instead of having an intermediate version where both functions exist, just bring forward the final aim of a single function by fixing all the calls?

pinc/maybe_mail.inc
windymilla: I don't know - I didn't see one straight away, as it occurred to me when I was struggling to get it working in the first place (because I had security enabled which FakeEmail didn't want). I'll have another look.

srjfoo:
Change to testing instructions: @cpeel has set things up so that we can just open [http://pgdp.org:8080/](http://pgdp.org:8080/) to view an email that would have been sent to us. (Which would actually block us being able to run `fakeemail` as described above.)

srjfoo:
I've just done a small amount of testing, but one drawback so far is that it doesn't show the emails sent to other users that I can see. (I set one of my other users to have event subscriptions for the same project as my main one, but in French, but only got the one directed to me.)

windymilla:
> I've just done a small amount of testing, but one drawback so far is that it doesn't show the emails sent to other users that I can see. (I set one of my other users to have event subscriptions for the same project as my main one, but in French, but only got the one directed to me.)

Are you sure you have email addresses set up for your other users? I don't think the FakeEmail knows which is actually you

srjfoo:
> > I've just done a small amount of testing, but one drawback so far is that it doesn't show the emails sent to other users that I can see. (I set one of my other users to have event subscriptions for the same project as my main one, but in French, but only got the one directed to me.)
> 
> Are you sure you have email addresses set up for your other users? I don't think the FakeEmail knows which is actually you
> 
> They can be fake addresses of course

That did it, thanks ... of course I found that I really _**do**_ need to be testing from your sandbox for it to work. 😁

pinc/maybe_mail.inc
cpeel: I'm on-board with doing a wholesale update. I see one use of the old function in noncvs scripts, but that's an easy replacement.

pinc/maybe_mail.inc
```diff
@@ -4,6 +4,135 @@ include_once($relPath.'misc.inc'); // html_safe()
...
+    $success = true;
+
+    // Append a standard footer to all emails sent out by the system.
+    if (!endswith($message, "\n")) {
+        $message .= "\n";
+    }
+
+    $message .= implode("\n", [
+        "",
+        "--  ",
```
cpeel, line 31: Looks like if we make this `---` we'll get a horizontal bar, which might be nice.

pinc/maybe_mail.inc
```diff
@@ -4,6 +4,135 @@ include_once($relPath.'misc.inc'); // html_safe()
...
+        "$site_name  ",
```
cpeel, line 32: We should add a comment here that the trailing spaces is to force the lines into multiple paragraphs, took me a bit to remember why you did it like this.

pinc/maybe_mail.inc
cpeel: Doesn't look like it. Let me give some thought on how we can pass in config info because that's going to be a bit of a challenge I think.

cpeel: I have a commit (pushed up to my repo): https://github.com/cpeel/dproofreaders/commit/224eb4cccd900009a6098296c3b417e0130d5b76 that adds dynamic configuration. I have updated `~cpeel/bin/TEST_setup.sh` and `~cpeel/bin/BRANCH_setup.sh` to use the FakeEmail port, so you should be able to update your `site_vars.php` to use these new settings:
```bash
git checkout email
./SETUP/configure ~cpeel/bin/TEST_setup.sh .
```

You're welcome to take or drop whatever parts of this you want.

windymilla:
@cpeel - I don't know if the failed check is because of something I did or didn't do, relating to the SMTP configuration stuff

Apart from that, I think this is ready for review/testing (noting that it still needs a final commit to stop it sending emails on TEST once testing is complete)

pinc/send_mail.inc
```diff
@@ -0,0 +1,153 @@
...
+
+        // Content
+        $mail->isHTML(true);
+        $mail->Subject = $subject;
+        $mail->Body = $html_message;
+        $mail->AltBody = $message;
+
+        $mail->send();
+    } catch (Exception $e) {
+        echo "<p>Message could not be sent. Mailer Error: {$mail->ErrorInfo}</p>";
```
cpeel, line 85: I suggest we change this to a `error_log()` instead of showing an error to the user for a couple of reasons.
* The user might not even realize that what they did results in emails being sent, so this would be confusing.
* The error might contain system-level info that we don't want to show to the user -- up to and including other people's email addresses.

Something like:
```php
error_log("send_email.inc - Message could not be sent. PHPMailer Error: {$mail->ErrorInfo}");
```

cpeel (commented):
I added a small commit that will fix the tests -- the CI code has a separate configuration file.

pinc/send_mail.inc
windymilla: Oh yes, two good reasons!

pinc/Project.inc
```diff
@@ -1266,8 +1267,8 @@ class Project
             // Should really be localized according to the preferences of the *recipient*.
             _("This message is regarding the following project:"),
             "",
-            "\"$title\"",
-            "    $code_url/project.php?id=$projectid",
+            mdmail_append_linebreak("\"$title\""),
```
cpeel, line 1270: Your changes to this function are what I'd like to see for this PR which makes primarily subtle changes to the email contents.

I'm hoping future iterations, after perhaps some user input, can refine this and other email contents to make them more modern. For instance, I'd love for this to simply be:
```md
[$title]($code_url/project.php?id=$projectid)
```

pinc/send_mail.inc
```diff
@@ -0,0 +1,153 @@
...
+
+    $from = $from ?? $auto_email_addr;
+    $reply_to = $reply_to ?? $auto_email_addr;
+
+    // Append a standard footer to all emails sent out by the system.
+    if (!endswith($message, "\n")) {
+        $message .= "\n";
+    }
+
+    $message .= implode("\n", [
```
cpeel, line 27: Perhaps not for this PR, but I'd love for us to consider are more refined footer for HTML emails since we have complete control over that here.

For instance, perhaps:
```html
<hr>
<a href="$site_url">$site_name</a><br>
<small>This is an automated message.</small>
```

pinc/send_mail.inc
```diff
@@ -0,0 +1,153 @@
...
+        echo html_safe("Reply-To: $reply_to\n");
+        echo "</pre>";
+        echo "<pre style='white-space: pre-wrap'>";
+        echo html_safe("$message\n");
+        echo "</pre>\n";
+        echo "<hr style='border-style:dashed;'>\n";
+        echo "$html_message\n"   ;
+        echo "<hr style='border-style:double;'>\n";
+        // Comment out line below to allow sending of emails when testing
+        // return true;
```
cpeel, line 58: Note to myself: confirm this is uncommented before merging.

cpeel (approved):
This is great. It immediately improves the look of our current emails and lays the foundation for future email text refinements to take advantage of the new formatting options.

cpeel:
Fixes https://github.com/DistributedProofreaders/dproofreaders/issues/561

srjfoo (approved):
Nigel did the necessary hacking of his sandbox to test `notify_old_pp`, which is the only thing I haven't done any testing on. I think everything's good.

70ray (approved):
Tested the new user email. Looks good. (Punctuation is outside the scope of this PR.)

pinc/send_mail.inc
windymilla: Did a small amount of prettifying in latest commit

windymilla:
Sandbox https://www.pgdp.org/~windymilla/c.branch/email updated

chrismiceli (approved):

cpeel:
Footer updates look great, thanks @windymilla! I'm happy for this to be merged in after we remove the comment for the return.

windymilla:
@cpeel - I removed the comment and verified that when I sent an email, it still appeared on the webpage, but did not get sent to FakeEmail server.

PR #892, opened 2023-04-17
Use Page Browser in place of downloadproofed

cpeel:
Replace the use of `downloadproofed.php` with the Page Browser. This enables users to get a lot of additional context about the text very easily. This implements [Task 1259](https://www.pgdp.net/c/tasks.php?action=show&task_id=1259).

This changes links on the following pages:
* diff page text links
* project quick check
* handle bad page (accessible from the "Fix" link on the diff page)

Testable in the [replace-downloadproofed-with-page-browser](https://www.pgdp.org/~cpeel/c.branch/replace-downloadproofed-with-page-browser/) sandbox.


srjfoo (approved):
Thank you!

70ray (approved):

windymilla (approved):
Looks good to me - I've wanted this quite recently when looking at text from the Page Detail page during proofing. I'd got so used to the page browser view, I felt very disappointed to get the old downloadproofed view.

Question (almost certainly outside the scope of this PR) - is there any scope/merit in making the setting of Text/Image/Image&Text on Page Browser be more persistent under some circumstances? I click on the Page Detail page to open a view of the text and turn on Image & Text to see the image too. Then I go back to Page Detail and see another page I want to look at, and it comes up Text only (or Image only if I click the Image link). If the code could read my mind, it would know I definitely want text if I click the Text link, and image if I click the Image link, but also if my last view of the Page Browser showed both, then I probably want both for this click too.

pinc/project_quick_check.inc
```diff
@@ -450,6 +449,12 @@ function _test_project_for_bad_bytes($projectid)
 
     while ([$text, $imagename, $proofers] = page_info_fetch($pages_res)) {
         $round_num = get_round_num_from_proofers($proofers);
+        if ($round_num == 0) {
```
chrismiceli, line 452: I guess it doesn't make sense to add a proper Round for OCR here https://github.com/DistributedProofreaders/dproofreaders/blob/d9e338bacfa4ed8a6547bbafd54dc4b88742cfac/pinc/stages.inc so we could use get_Round_for_round_number like our other rounds for various reasons, but just saying it in case it has just been missed.

pinc/project_quick_check.inc
```diff
@@ -466,17 +471,9 @@ function _test_project_for_bad_bytes($projectid)
 
         // write image and round columns
         $rowspan = $number_bad_chars;
-        $img_url = "$code_url/tools/page_browser.php?project=$projectid&imagefile=$imagename";
-        $text_url = "$code_url/tools/project_manager/downloadproofed.php?project=$projectid&amp;image=$imagename&amp;round_num=$round_num";
+        $page_browser_url = "$code_url/tools/page_browser.php?project=$projectid&imagefile=$imagename&amp;round_id=$round_name&mode=imageText";
         $details_for_this_page = "<tr>\n";
-        $details_for_this_page .= "<td class='top-align' rowspan='$rowspan'><a href='$img_url'>$imagename</a></td>\n";
```
chrismiceli, line 472: should we leave the img link with the `mode=image` to not break anyone's habits?

chrismiceli (approved):

pinc/project_quick_check.inc
cpeel: Good question. Unfortunately the code treats Stages as "things that can be done" and a lot of our code does various things to iterate over them, like the Activity Hub and (IIRC) the navigation header. I think we'd need to do a whole lot of other special casing to make OCR a stage of some sort. Something worth thinking about as we do this "OCR is a sort of stage thing" allllllll over the place.

pinc/project_quick_check.inc
cpeel: I defer to @srjfoo on what she thinks is best. My thought was that having a single view of them together is more useful and the reason they were separate was because we _couldn't_ put them together.

srjfoo: I saw that and squinted at it a little bit before I understood, and then thought "what a good idea, would that work in the page details?" -- but upon investigation, decided that the page details table is probably better off with two separate columns.

I'm good with combining them into one column in Project Quick Check.

cpeel:
> is there any scope/merit in making the setting of Text/Image/Image&Text on Page Browser be more persistent under some circumstances?

I'm happy to consider this if we can describe a way that is intuitive for users. Implementation wise I could see the PB storing some transient local state for "the last someone looked at a page in this project they used the imageText view" so within some time window if they ask for a page in that project again they continue to get the imageText view even if they clicked on just the image or just the text. Consider opening a Task or an Issue with this request.

pinc/project_quick_check.inc
windymilla: Would it be to weird to show Image only if clicking on the Image link (you wouldn't know what text to show anyway - you'd have to show OCR, but maybe not what the user wants), but Text & Image if clicking on the Text link? It feels like showing the image for free when they want to see the text might be helpful and at worst would be irrelevant. Very happy with it

windymilla:
> > is there any scope/merit in making the setting of Text/Image/Image&Text on Page Browser be more persistent under some circumstances?
> 
> I'm happy to consider this if we can describe a way that is intuitive for users. Implementation wise I could see the PB storing some transient local state for "the last time user X looked at a page in this project they used the imageText view" so within some time window if user X asks for a page in that project again they continue to get the imageText view even if they clicked on just the image or just the text. Consider opening a Task or an Issue with this request.

Yes - I think you've hit on the key point. Could be hard to make it intuitive for users - when systems try to outthink you, they can often be more annoying than dumb systems because you come to expect a certain behavior, then slip over the edge to a different behaviour (e.g. out of time window, or on a different project, or some such criterion). Definitely not one to rush into :)

pinc/project_quick_check.inc
windymilla: Or maybe it's not "for free" - on my wide screen with vertical layout, it is for free, but on a narrower screen or with horizontal layout, the image takes up screen space which could be annoying if you didn't want it.

srjfoo: > Would it be to weird to show Image only if clicking on the Image link ...

My comment above applies only to the Project Quick Check change where @cpeel combined the png number and the round into the same column, and the link brings up both image and text for that round.

windymilla: Sorry, I didn't make it clear I was commenting on your "what a good idea, would that work in the page details?"

srjfoo: > Sorry, I didn't make it clear I was commenting on your "what a good idea, would that work in the page details?"

Ah, okay. And apparently I didn't make it clear enough that I decided that it wasn't such a good idea after all 😁

windymilla: You were clear - it just made me think 😁 - definitely leave it as it is.

PR #893, opened 2023-04-18
Raise exceptions

70ray:
This is the first stage of converting the functions we wish to use in a proofreading API to raise exceptions instead of returning messages.
Converts the function project->can_be_proofed_by_current_user();

### DIFF
--- a/pinc/LPage.inc
+++ b/pinc/LPage.inc
@@ -137,7 +137,7 @@ function get_available_page($projectid, $proj_state, $pguser, &$err)
     $round = get_Round_for_project_state($proj_state);
 
     try {
-        $page_array = get_available_proof_page($project, $round, $pguser);
+        $page_array = get_available_proof_page_array($project, $round, $pguser);
     } catch (ProjectException | UserAccessException $exception) {
         $err = $exception->getMessage();
         return null;
--- a/pinc/LPage.inc
+++ b/pinc/LPage.inc
@@ -137,12 +137,11 @@ function get_available_page($projectid, $proj_state, $pguser, &$err)
     $round = get_Round_for_project_state($proj_state);
 
     try {
-        $page_array = get_available_proof_page_array($project, $round, $pguser);
+        [$imagefile, $state] = get_available_proof_page_array($project, $round, $pguser);
     } catch (ProjectException | UserAccessException $exception) {
         $err = $exception->getMessage();
         return null;
     }
-    [$imagefile, $state] = $page_array;
     $lpage = new LPage($project->projectid, $imagefile, $state, 0);
 
     $lpage->checkout($pguser);
--- a/pinc/LPage.inc
+++ b/pinc/LPage.inc
@@ -151,7 +151,7 @@ function get_available_page($projectid, $proj_state, $pguser, &$err)
     return $lpage;
 }
 
-function get_available_proof_page($project, $round, $pguser)
+function get_available_proof_page_array($project, $round, $pguser)
 // Returns an array [image, state] unless no page is available,
 // in which case throw an exception.
 {
--- a/pinc/LPage.inc
+++ b/pinc/LPage.inc
@@ -137,8 +137,10 @@ function get_available_page($projectid, $proj_state, $pguser, &$err)
     }
 
     $round = get_Round_for_project_state($proj_state);
-    $err = can_user_get_pages_in_project($pguser, $project, $round);
-    if ($err) {
+    try {
+        can_user_get_pages_in_project($pguser, $project, $round);
+    } catch (UserAccessException $exception) {
+        $err = $exception->getMessage();
         return null;
     }
 
--- a/pinc/LPage.inc
+++ b/pinc/LPage.inc
@@ -157,7 +157,7 @@ function get_available_proof_page_array($project, $round, $pguser)
 {
     global $preceding_proofer_restriction;
 
-    can_user_get_pages_in_project($pguser, $project, $round);
+    validate_user_can_get_pages_in_project($pguser, $project, $round);
 
     // If the global setting is to not serve the immediately preceding page,
     // first see if we can serve them a page they haven't at all seen before,
--- a/pinc/LPage.inc
+++ b/pinc/LPage.inc
@@ -125,8 +125,6 @@ function get_available_page($projectid, $proj_state, $pguser, &$err)
 // Returns an LPage, unless no page is available,
 // in which case returns NULL and sets $err.
 {
-    global $preceding_proofer_restriction;
-
     // can throw NonexistentProjectException
     $project = new Project($projectid);
 
@@ -137,12 +135,29 @@ function get_available_page($projectid, $proj_state, $pguser, &$err)
     }
 
     $round = get_Round_for_project_state($proj_state);
+
     try {
-        can_user_get_pages_in_project($pguser, $project, $round);
-    } catch (UserAccessException $exception) {
+        $page_array = get_available_proof_page($project, $round, $pguser);
+    } catch (ProjectException | UserAccessException $exception) {
         $err = $exception->getMessage();
         return null;
     }
+    [$imagefile, $state] = $page_array;
+    $lpage = new LPage($project->projectid, $imagefile, $state, 0);
+
+    $lpage->checkout($pguser);
+
+    $err = null;
+    return $lpage;
+}
+
+function get_available_proof_page($project, $round, $pguser)
+// Returns an array [image, state] unless no page is available,
+// in which case throw an exception.
+{
+    global $preceding_proofer_restriction;
+
+    can_user_get_pages_in_project($pguser, $project, $round);
 
     // If the global setting is to not serve the immediately preceding page,
     // first see if we can serve them a page they haven't at all seen before,
@@ -157,18 +172,10 @@ function get_available_page($projectid, $proj_state, $pguser, &$err)
         $page_array = get_available_page_array($project, $round, $pguser, $preceding_proofer_restriction);
     }
     if (!$page_array) {
-        $err = _("No more files available for proofreading for this round of the project.");
-        return null;
+        throw new ProjectNoMorePagesException(_("The project has no more pages available for proofreading in this round."));
     }
 
-    [$imagefile, $state] = $page_array;
-
-    $lpage = new LPage($projectid, $imagefile, $state, 0);
-
-    $lpage->checkout($pguser);
-
-    $err = null;
-    return $lpage;
+    return $page_array;
 }
 
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
--- a/pinc/Project.inc
+++ b/pinc/Project.inc
@@ -1066,9 +1066,7 @@ class Project
 
     public function get_num_pages_in_state($state = null, $counter_sql = "*")
     {
-        if (!$this->pages_table_exists) {
-            throw new NoProjectPageTable(_("Project page table does not exist."));
-        }
+        $this->validate_project_page_table_exists();
 
         // if state is null, get all pages
         $where = $state ? sprintf("state = '%s'", DPDatabase::escape($state)) : "1";
@@ -1092,10 +1090,7 @@ class Project
 
     public function get_page_names_from_db()
     {
-        if (!$this->pages_table_exists) {
-            throw new NoProjectPageTable(_("Project page table does not exist."));
-        }
-
+        $this->validate_project_page_table_exists();
         validate_projectID($this->projectid);
         $page_image_names = [];
         $sql = "
@@ -1109,6 +1104,13 @@ class Project
         }
 
         return $page_image_names;
+    }
+
+    public function validate_project_page_table_exists()
+    {
+        if (!$this->pages_table_exists) {
+            throw new NoProjectPageTable(_("Project page table does not exist."));
+        }
     }
 
     public function get_page_names_from_dir()
--- a/pinc/Project.inc
+++ b/pinc/Project.inc
@@ -88,6 +88,30 @@ class ProjectPageException extends ProjectException
     }
 }
 
+class ProjectNotInRoundException extends ProjectException
+{
+    public function __construct($message, $code = 110)
+    {
+        parent::__construct($message, $code);
+    }
+}
+
+class ProjectRequiresMaintenanceException extends ProjectException
+{
+    public function __construct($message, $code = 111)
+    {
+        parent::__construct($message, $code);
+    }
+}
+
+class ProjectNotAvailableException extends ProjectException
+{
+    public function __construct($message, $code = 112)
+    {
+        parent::__construct($message, $code);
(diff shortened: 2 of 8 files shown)


pinc/Project.inc
```diff
@@ -955,83 +979,38 @@ class Project
...
-    // These should be treated as constants.
-    // (PHP doesn't appear to allow constants as class members.)
-    public $CBP_OKAY = 0;
-    public $CBP_PROJECT_NOT_IN_ROUND = 1;
-    public $CBP_PROJECT_NOT_AVAILABLE = 2;
-    public $CBP_REQUESTOR_NOT_LOGGED_IN = 3;
-    public $CBP_USER_NOT_QUALIFIED_FOR_ROUND = 4;
-    public $CBP_PROJECT_REQUIRES_MAINTENANCE = 5;
-
     public function can_be_proofed_by_current_user()
```
cpeel, line 982: A `can_*` or `is_*` function should return an answer to the question it's asked, not an exception as an answer. We can rename this function to `validate_user_can_get_pages_in_project()` which should either pass or return an exception though.

Or the `validate_()` function can call this one and raise the appropriate exception.

pinc/RoundDescriptor.inc
```diff
@@ -199,6 +199,25 @@ class Round extends Stage
         // Give them the lowest honorific.
         return $honorific;
     }
+
+    // -----------
+
+    public function user_can_access()
```
cpeel, line 205: Similar: `user_can()` should return an answer. `validate_user_can()` can return an exception as the answer.

70ray:
`get_available_proof_page_array()` is like `get_available_page` but omits the project continuity test. The API will specify the round rather than the project state so the check will have to be done differently.

70ray:
It seems to me that this is as far as we can usefully go in converting existing functions to throw exceptions so they can be used in the proofreading API. For the actual functions like checkout, save and return it would be easier to make a new ProofPage class rather than adapting the existing LPage and DPage.
In this situation it will be desirable to manually test the functions that these changes impact before merging them.

`can_be_proofed_by_current_user()` is used in two places:
in `project.php` it gives the text in the grey boxes in the project page if the user cannot work on the project as follows:

-  The project is not in any round -- no text in the box.
- Project has not been converted to UTF-8 yet.
- The project is in a round, but isn't available.
- User is not logged in.
- The user is not qualified for the round.
In `proof.php` if the user cannot work on the page for these reasons it goes to the abort page. In this case it gives a message if the project is not in a round.

70ray:
It seems to me that this is as far as we can usefully go in converting existing functions to throw exceptions so they can be used in the proofreading API. For the actual functions like checkout, save and return it would be easier to make a new ProofPage class rather than adapting the existing LPage and DPage.
In this situation it will be desirable to manually test the functions that these changes impact before merging them.

`validate_can_be_proofed_by_current_user()` is used in two places:
in `project.php` it gives the text in the grey boxes in the project page if the user cannot work on the project for the following reasons:

-  The project is not in any round -- no text in the box.
- Project has not been converted to UTF-8 yet.
- The project is in a round, but isn't available.
- User is not logged in.
- The user is not qualified for the round.

If the user can get to `proof.php` then if the user cannot work on the page for these reasons it goes to the abort page. In this case it gives a message if the project is not in a round.

`validate_user_can_get_pages_in_project` is also used in two places:
It also gives text in the grey boxes in the project page if the user cannot work on the project as follows:

- You have reached your quota of pages from 'Beginners Only' projects in this round.
- You have reached your quota of pages from this 'Beginners Only' project.
- You do not have access to difficulty='beginner' (Mentors Only) projects in this round.
- This project is currently reserved for new proofreaders.

If you get to proof_frame.php via the `Save as done and proofread next page` route, these do not cause an abort but it should go to a new page offering options according to the circumstances of the user.

Similarly for the case that there are no more pages available for proofreading.

If it is possible to get to proof_frame.php with an invalid page state or a user who does not own the page (which would throw a ProjectPageException in the _Page_require() function in $lpage->checkout) then it should go to the abort page as previously.

pinc/LPage.inc
```diff
@@ -137,10 +135,29 @@ function get_available_page($projectid, $proj_state, $pguser, &$err)
     }
 
     $round = get_Round_for_project_state($proj_state);
-    $err = can_user_get_pages_in_project($pguser, $project, $round);
-    if ($err) {
+
+    try {
+        $page_array = get_available_proof_page_array($project, $round, $pguser);
```
cpeel, line 140: Just use `[$imagefile, $state] = ` here instead of using a temp var as that would be clearer.

pinc/RoundDescriptor.inc
```diff
@@ -199,6 +199,25 @@ class Round extends Stage
         // Give them the lowest honorific.
         return $honorific;
     }
+
+    // -----------
+
+    public function validate_user_can_access()
```
cpeel, line 205: We aren't very consistent in the codebase with this, but I would prefer that this either take the user we are validating, or we use `validate_current_user_can_access()` to be clear that it's `$pguser` being used. Ideally the former as that's more extensible.

pinc/stages.inc
```diff
@@ -308,15 +304,13 @@ function can_user_get_pages_in_project($username, $project, $round)
...
-                        _("You have reached your quota of pages from this 'Beginners Only' project, but you're welcome to work on a different one.")
-                        . " "
-                        . _("Or you could try working on an EASY project.");
+                    throw new BeginnersProjectQuotaReachedException(_("You have reached your quota of pages from this 'Beginners Only' project, but you're welcome to work on a different one. Or you could try working on an EASY project."));
                 }
             }
         } elseif ($round->is_a_mentor_round()) {
             if (!user_can_work_on_beginner_pages_in_round($round)) {
-                return _("You do not have access to difficulty='beginner' (Mentors Only) projects in this round.");
+                $message = _("You do not have access to difficulty='beginner' (Mentors Only) projects in this round.");
```
cpeel, line 312: Either assign `$message` and use it, or just put it in the exception like you do above. Probably the latter.

project.php
```diff
@@ -1,7 +1,7 @@
 <?php
 $relPath = './pinc/';
 include_once($relPath.'base.inc');
-include_once($relPath.'stages.inc');
+include_once($relPath.'stages.inc'); // validate_user_can_get_pages_in_project()
```
cpeel, line 4: We use a lot of other stuff in `stages.inc` in this file, I would leave off the comment about this one function lest others think this is the only thing we use in that file. Yeah, we're wildly inconsistent about how we do this 😕

cpeel (commented):
Overall this is great and I like how we're moving into using Exceptions more for error conditions instead of special values. I made a few comments about specific code pieces, but the overall structure of these changes look sound to me.

pinc/stages.inc
```diff
@@ -284,10 +283,7 @@ function can_user_get_pages_in_project($username, $project, $round)
     if ($project->difficulty == 'beginner') {
         if ($round->is_a_mentee_round()) {
             if (!user_can_work_on_beginner_pages_in_round($round)) {
-                return
-                    _("You have reached your quota of pages from 'Beginners Only' projects in this round.")
-                    . " "
-                    . _("Perhaps you could try working on an EASY project.");
+                throw new BeginnersQuotaReachedException(_("You have reached your quota of pages from 'Beginners Only' projects in this round. Perhaps you could try working on an EASY project."));
```
cpeel, line 286: I don't have strong feelings about combining these sentences together vs keeping them apart, however this will impact the site translations for no discernible benefit. I defer to @srjfoo on how she would like to approach these changes.

srjfoo: Oof, that's a good point.

I think I'm going to go for streamlining. I'm not sure why those strings were separated to begin with. Neither piece is used separately anywhere else. Can the code have line breaks to make for easier reading of the code?

pinc/Project.inc
```diff
@@ -88,6 +88,38 @@ class ProjectPageException extends ProjectException
     }
 }
 
+class ProjectNotInRoundException extends ProjectException
+{
+    public function __construct($message, $code = 110)
```
windymilla, line 93: Is there a reason the specific Exceptions like `ProjectNotInRoundException` require a message argument to their constructor, rather than the constructor containing the message ("The project is not in a round") and passing it to the parent constructor. Alternatively, if there is a need for different messages for certain circumstances, could the message argument have "The project is not in a round" as a default? If this type of exception is used in multiple places, it seems as though the default message could end up appearing in the code several times.

pinc/stages.inc
70ray: oops.

pinc/stages.inc
70ray: It seems you cannot escape newlines in strings in PHP like you can in C. But the DP_Code_Best_Practices wiki says you can break lines and the translators will ignore the extra spaces so I have done this.

pinc/Project.inc
70ray: Interesting point; should we provide messages where it seems there could be no dependence on the context? or keep a consistent pattern and always provide default messages or not? I defer to @cpeel 's opinion.

70ray:
Updated sandbox at https://www.pgdp.org/~rp31/c.branch/raise_exceptions

pinc/RoundDescriptor.inc
```diff
@@ -199,6 +199,25 @@ class Round extends Stage
...
+
+    public function validate_user_can_access($user)
+    {
+        global $code_url;
+
+        $uao = $this->user_access($user);
+        if (!$uao->can_access) {
+            $message = sprintf(
+                // TRANSLATORS: %1$s is the round name, %2$s is the round ID (eg: P1)
+                _('You have not yet been cleared to work on projects in %1$s (%2$s). Please visit <a href="%3$s">the %2$s home</a> to find out what happens in this round and how you can qualify to work in it.'),
```
70ray, line 213: This message contains html so would not be ideal for showing in an alert box. It would presumably work in a <dialog> element but we cannot use this yet.
In the pesent code, if it is shown in the abort page it is wrapped in html_safe() which is also not ideal. (It cannot happen normally, only by e.g. doing a page in F2, then revoking your access to F2 and then following the link to the page you did in the project page.)
Since it is quite acceptable to look at the project page of a project in a round you are not qualified to work in, perhaps we should consider making a message without html and catching this exception in the project page and adding the second sentence there.

70ray: Done this.

70ray:
Hang on! I've found something else to do.

srjfoo:
> or a user who does not own the page (which would throw a ProjectPageException in the _Page_require() function in $lpage->checkout) then it should go to the abort page as previously.

Still in the middle of testing, but wanted to note this: I logged in as one of my unprivileged users, proofed pages, kept the project page open and logged out from another browser window and logged into another of my test accounts. I clicked on the link to edit a page and got the following message:
>This operation (reopen) can only be done by the user who has the page checked out, which you are not.

However, if it's possible to open a page that's in an invalid page state, I'm not crafty enough to figure out how to do it 😁 

The message that I got appears to be in `DPage.inc`.

70ray:
> > or a user who does not own the page (which would throw a ProjectPageException in the _Page_require() function in $lpage->checkout) then it should go to the abort page as previously.
> 
> Still in the middle of testing, but wanted to note this: I logged in as one of my unprivileged users, proofed pages, kept the project page open and logged out from another browser window and logged into another of my test accounts. I clicked on the link to edit a page and got the following message:
> 
> > This operation (reopen) can only be done by the user who has the page checked out, which you are not.
> 
> However, if it's possible to open a page that's in an invalid page state, I'm not crafty enough to figure out how to do it grin
> 
> The message that I got appears to be in `DPage.inc`.

The message you quote and the function _Page_require() in DPage.inc should not be affected by this branch, although the purpose of this testing is to ensure that nothing has been accedentally broken.
Although the function _Page_require() does things which are similar to other cases that this branch intercepts, I should explain why it does not go there:
We are proposing to make eliminate the page_state --- to be continued, thunderstorm is happenning.

70ray:
To enable testing the value of $page_tally_threshold in the sandbox has been changed from 500 to 20.
This means that the message shown for projects reserved for new proofreaders will be `This project is currently reserved for proofreaders who joined less than 21 days ago or have done less than 20 pages.`

pinc/Project.inc
```diff
@@ -1124,6 +1105,13 @@ class Project
         return $page_image_names;
     }
 
+    public function validate_page_table_exists()
+    {
+        if (!$this->pages_table_exists) {
+            throw new NoProjectPageTable(_("Project page table does not exist."));
```
srjfoo, line 1111: It took me a while to figure out how to trigger this particular message. I mostly got "Page table does not exist". I was finally able to trigger it by going to the images index for an active project and subbing in a deleted projectid in the URL. Are there other situations that _**should**_ trigger the message that maybe aren't?

pinc/Project.inc
```diff
@@ -955,83 +986,38 @@ class Project
...
+            // The project is not in any round.
+            throw new ProjectNotInRoundException(_("The project is not in a round"));
+        }
         if (!$this->is_utf8) {
             // Project has not been converted to UTF-8 yet
-            return [
-                $this->CBP_PROJECT_REQUIRES_MAINTENANCE,
-                _('The project requires maintenance.') . ' ' . _("Project table is not UTF-8."),
-            ];
+            throw new ProjectRequiresMaintenanceException(_('The project requires maintenance. Project table is not UTF-8.'));
```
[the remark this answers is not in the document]
70ray: > These can be fixed after the testing is done, but for now, if anyone else wants to verify, `projectID459698b8b63ac` has been converted to `latin1` (so has `projectID6229978d368ca`, but it's a new project, and the blurb box with the "requires maintenance" message doesn't appear for new projects; just the error message in the project data on the project page).

It has moved into `Proofreading Round 1: Available` now, and the blurb box appears with the message. But the link to start proofreading does not appear. It looks the same on DPT branch so it looks like nothing is broken.

pinc/Project.inc
70ray: It looks like the message "Page table does not exist" is only made in project_quick_check.
The only ways that  there cannot be a page table seem to be if the project is deleted or archived. Or perhaps a system crash during project creation.

pinc/Project.inc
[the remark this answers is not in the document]
srjfoo: Sorry for the confusion. There were two projects that I converted to `latin1`. The first one was already in P1 Available, and showed the blurb box. The second one was/is still a new project, and doesn't show the blurb box because the blurb box isn't shown in that case.

pinc/Project.inc
srjfoo: I may not have been clear enough here, either. What I wanted to know is if there are other places than the image index where the message in the code here ("Project page table does not exist.") should be seen. It's not in Project Quick Check, and I guess I wanted to make sure that it is not supposed to be.

pinc/Project.inc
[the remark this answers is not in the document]
70ray: Probably my confusion.

70ray:
Considering another change.

PR #897, opened 2023-04-25
Add link in post files for music

70ray:
This addresses https://www.pgdp.net/c/tasks.php?task_id=2059

Sandbox at: https://www.pgdp.org/~rp31/c.branch/post_music.

This project has a music zip (added by @srjfoo )
https://www.pgdp.org/~rp31/c.branch/post_music/project.php?id=projectID44dd012611ab0

### DIFF
--- a/project.php
+++ b/project.php
@@ -1401,6 +1401,7 @@ function do_post_files()
         echo "<ul>";
 
         echo_download_zip(_("Download Zipped Images"), 'images');
+        echo_download_zip(_("Download Zipped Music Files"), '_music');
 
         if ($state == PROJ_POST_FIRST_AVAILABLE || $state == PROJ_POST_FIRST_CHECKED_OUT) {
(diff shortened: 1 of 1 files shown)


srjfoo (approved):

project.php
```diff
@@ -1401,6 +1401,7 @@ function do_post_files()
         echo "<ul>";
 
         echo_download_zip(_("Download Zipped Images"), 'images');
+        echo_download_zip(_("Download Zipped Music Files"), '_music');
```
chrismiceli, line 1404: why `_music`? Also, could we make it a constant since it is referenced in more than one place? It'd make updating it later for any reason easier.

chrismiceli (approved):

project.php
70ray: We could do that. I suppose we should really have a constant for `image` also. I'll leave it like it is for now.

PR #900, opened 2023-05-19
Fix PP Limit

srjfoo:
Obviously not something that's used often, or it would have been fixed long ago. `manage_site_access_privileges.php` and `user_is.inc` were not using the same setting name. Sandbox at [fix-pp-limit](https://www.pgdp.org/~srjfoo/c.branch/fix-pp-limit/).

### DIFF
--- a/pinc/user_is.inc
+++ b/pinc/user_is.inc
@@ -126,7 +126,7 @@ function user_has_project_loads_disabled()
 function that_user_is_over_PP_checked_out_limit($username)
 {
     $userSettings = & Settings::get_settings($username);
-    $limit = $userSettings->get_value("PP_checked_out_limit", 'none');
+    $limit = $userSettings->get_value("pp_limit_value", 'none');
     if (! is_numeric($limit)) {
         return false;
     }


srjfoo:
Testing notes: I've tested it and verified that:
1. If a PPer has had a checked-out limit defined, they can't see the button to check out a project on the project page, nor can they check one out from the search results page.
2. If a PPer is over their limit, and a project reserved for them finishes the rounds, it is sent to PP Available, not added to their PP queue.

If you feel a need to test it, add a PP limit using `manage_site_access_privileges.php` and test it out. The limit, if set, applies even to squirrels.

cpeel (approved):
Thanks for finding this and the PR to fix it!

srjfoo:
> Thanks for finding this and the PR to fix it!

And thanks for the help sorting through what was needed!

chrismiceli (approved):
Do we use the old setting anymore anywhere? Is it worth a database migration to remove those values?

srjfoo:
> Do we use the old setting anymore anywhere? Is it worth a database migration to remove those values?

It hasn't been used in my memory, which doesn't mean that it shouldn't be. I wouldn't remove the setting at this point.

cpeel:
> > Do we use the old setting anymore anywhere? Is it worth a database migration to remove those values?
> 
> It hasn't been used in my memory, which doesn't mean that it shouldn't be. I wouldn't remove the setting at this point.

IIRC, Sharon only found one instance of it on PROD and several of the other one. I don't think we need a DB migration for this.

srjfoo:
> > > Do we use the old setting anymore anywhere? Is it worth a database migration to remove those values?
> > 
> > 
> > It hasn't been used in my memory, which doesn't mean that it shouldn't be. I wouldn't remove the setting at this point.
> 
> IIRC, Sharon only found one instance of it on PROD and several of the other one. I don't think we need a DB migration for this.

Ohhh ... I misunderstood what you meant by DB migration. No, definitely not needed. It would take longer to throw the script together than to fix any differences manually.

chrismiceli:
> > > > Do we use the old setting anymore anywhere? Is it worth a database migration to remove those values?
> > > 
> > > 
> > > It hasn't been used in my memory, which doesn't mean that it shouldn't be. I wouldn't remove the setting at this point.
> > 
> > 
> > IIRC, Sharon only found one instance of it on PROD and several of the other one. I don't think we need a DB migration for this.
> 
> Ohhh ... I misunderstood what you meant by DB migration. No, definitely not needed. It would take longer to throw the script together than to fix any differences manually.

Awesome news

windymilla (approved):

PR #901, opened 2023-05-20
More basic usages

chrismiceli:
The main goal here is to remove jQuery for more usages that don't involve creating and modifying a bunch of DOM elements. This mostly consist of
1. Using normal event listeners
2. Removing $.Callbacks in favor of arrays of methods
3. Removing $.isPlainObject

I also took the opportunity to fix a deprecation warning on event.which.

Again, this is low priority with some risks, so no pressure to merge or review.

Branch for testing
https://www.pgdp.org/~chrismiceli/c.branch/jquery/

### DIFF
--- a/scripts/page_browse.js
+++ b/scripts/page_browse.js
@@ -25,7 +25,7 @@ var maketextControl = function(textArea) {
         fontFaceSelector.add(new Option(fontFaces[index], index));
     });
 
-    $(fontFaceSelector).change(function () {
+    fontFaceSelector.addEventListener("change", function () {
         const fontFaceIndex = this.value;
         setFontFace(fontFaceIndex);
         textData.fontFaceIndex = fontFaceIndex;
@@ -42,7 +42,7 @@ var maketextControl = function(textArea) {
         fontSizeSelector.add(new Option(displayFontSize, fontSize));
     });
 
-    $(fontSizeSelector).change(function () {
+    fontSizeSelector.addEventListener("change", function () {
         const fontSize = fontSizeSelector.value;
         setFontSize(fontSize);
         textData.fontSize = fontSize;
@@ -69,7 +69,10 @@ var maketextControl = function(textArea) {
         setup: function(storageKey) {
             textKey = storageKey + "-text";
             textData = JSON.parse(localStorage.getItem(textKey));
-            if(!$.isPlainObject(textData)) {
+            if(!textData ||
+                typeof textData.textWrap !== "string" ||
+                typeof textData.fontFaceIndex !== "number" ||
+                typeof textData.fontSize !== "string") {
                 textData = {
                     textWrap: "N",
                     fontFaceIndex: 0,
@@ -109,7 +112,7 @@ function makeTextWidget(container, splitter = false, reLayout = null) {
         content.append(topTextDiv, bottomTextDiv);
 
         subSplitter = splitControl(content, {splitVertical: false, reDraw: reLayout});
-        subSplitter.dragEnd.add(function (percent) {
+        subSplitter.dragEnd.push(function (percent) {
             textSplitData.splitPercent = percent;
             localStorage.setItem(splitterKey, JSON.stringify(textSplitData));
         });
@@ -122,7 +125,7 @@ function makeTextWidget(container, splitter = false, reLayout = null) {
             if(splitter) {
                 splitterKey = textWidgetKey + "-split";
                 textSplitData = JSON.parse(localStorage.getItem(splitterKey));
-                if(!$.isPlainObject(textSplitData)) {
+                if(!textSplitData || typeof textSplitData.splitPercent !== 'number') {
                     textSplitData = {splitPercent: 100};
                 }
                 subSplitter.setSplitPercent(textSplitData.splitPercent);
@@ -145,7 +148,7 @@ function makeTextWidget(container, splitter = false, reLayout = null) {
 var viewSplitter = function(container, storageKey) {
     const storageKeyLayout = storageKey + "-layout";
     let layout = JSON.parse(localStorage.getItem(storageKeyLayout));
-    if(!$.isPlainObject(layout)) {
+    if(!layout || layout.splitDirection !== "horizontal" || layout.splitDirection !== "vertical") {
         layout = {splitDirection: "horizontal"};
     }
     const splitVertical = (layout.splitDirection === "vertical");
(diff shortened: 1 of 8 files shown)


scripts/splitControl.js
```diff
@@ -58,8 +60,8 @@ var splitControl = function(container, config) {
     let height;
     let width;
 
-    let reSize = $.Callbacks();
-    let dragEnd = $.Callbacks();
+    let reSize = [];
```
70ray, line 63: Alternatively, here and elsewhere, you could replace $.Callbacks with a Set instead of an array. Then the associated add() functions could remain unchanged instead of changing to push(). (It is avaialable within our range of supported browsers.)

70ray: Perhaps also rename reSize() to onResize() (and prepend on- similarly elsewhwere) to make it clearer what this is doing.

tools/proofers/previewControl.js
```diff
@@ -179,16 +179,23 @@ $( function() {
...
-        $("#" + viewMode).prop("checked", true);
+        document.getElementById(viewMode).checked = true;
         // check if MathJax already loaded. Will break if load more than once
         if(previewStyles.allowMathPreview && (typeof(MathJax) === 'undefined')) {
-            // allow caching which getScript() by default does not
-            return $.ajax({
-                dataType: "script",
-                cache: true,
-                url: "https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-svg.js",
+            const maxjaxScriptElement = document.createElement('script');
```
70ray, line 185: perhaps "mathJaxScriptElement" ?

70ray (commented):
By chance, I was also looking at some changes in the splitControl area. The only one which coincides with this is replacing the jquery callbacks which I have commented on.

I have tested tahat the math preview works.

70ray (approved):

cpeel (approved):
I'm a JS n00b and am fine with these changes as long as Ray is and it passes Sharon's testing.

70ray (approved):
Looks good to me.

srjfoo (changes requested):
Not entirely sure where this comment belongs in the code, so I'm not attaching it to any specific piece of code. In testing the splitter (and it's not just the splitter, it's text-only mode, as well), if I set the font size to other than browser default and refresh the page, the new font size persists. However, if I set a different font face for the text and refresh the page, both the font face and the font size are reset to browser default.

In going to test it in vertical mode (side-by-side) as well as horizontal mode (top-and-bottom), I discovered that when I refreshed the page in vertical mode, the display switched back to horizontal mode with font and font size back at browser default (note that when I switched back to vertical mode, the font size had persisted -- **_unless_** I also changed the font to other than browser default.

70ray:
> Not entirely sure where this comment belongs in the code, so I'm not attaching it to any specific piece of code. In testing the splitter (and it's not just the splitter, it's text-only mode, as well), if I set the font size to other than browser default and refresh the page, the new font size persists. However, if I set a different font face for the text and refresh the page, both the font face and the font size are reset to browser default.
> 
> In going to test it in vertical mode (side-by-side) as well as horizontal mode (top-and-bottom), I discovered that when I refreshed the page in vertical mode, the display switched back to horizontal mode with font and font size back at browser default (note that when I switched back to vertical mode, the font size had persisted -- **_unless_** I also changed the font to other than browser default.

It looks like (at least part of) the problem is with numbers being stored as strings. In the case of the split percent this is because the function uses toFixed() which gives a string.
Sorry I didn't notice this before. There must have been something sub-conscious which didn't surface because when I saw this comment I went straight to look at this.

chrismiceli:
> > Not entirely sure where this comment belongs in the code, so I'm not attaching it to any specific piece of code. In testing the splitter (and it's not just the splitter, it's text-only mode, as well), if I set the font size to other than browser default and refresh the page, the new font size persists. However, if I set a different font face for the text and refresh the page, both the font face and the font size are reset to browser default.
> > In going to test it in vertical mode (side-by-side) as well as horizontal mode (top-and-bottom), I discovered that when I refreshed the page in vertical mode, the display switched back to horizontal mode with font and font size back at browser default (note that when I switched back to vertical mode, the font size had persisted -- **_unless_** I also changed the font to other than browser default.
> 
> It looks like (at least part of) the problem is with numbers being stored as strings. In the case of the split percent this is because the function uses toFixed() which gives a string. Sorry I didn't notice this before. There must have been something sub-conscious which didn't surface because when I saw this comment I went straight to look at this.

Yeah, I saw this in the settings for the split percents too. I'll take a look and ensure there aren't any other instances that can be impacted. It is introduced when I enhanced the strength of the test on the values coming from storage. Before we just checked they were objects, I checked the properties were of the right type.

chrismiceli:
> > > Not entirely sure where this comment belongs in the code, so I'm not attaching it to any specific piece of code. In testing the splitter (and it's not just the splitter, it's text-only mode, as well), if I set the font size to other than browser default and refresh the page, the new font size persists. However, if I set a different font face for the text and refresh the page, both the font face and the font size are reset to browser default.
> > > In going to test it in vertical mode (side-by-side) as well as horizontal mode (top-and-bottom), I discovered that when I refreshed the page in vertical mode, the display switched back to horizontal mode with font and font size back at browser default (note that when I switched back to vertical mode, the font size had persisted -- **_unless_** I also changed the font to other than browser default.
> > 
> > 
> > It looks like (at least part of) the problem is with numbers being stored as strings. In the case of the split percent this is because the function uses toFixed() which gives a string. Sorry I didn't notice this before. There must have been something sub-conscious which didn't surface because when I saw this comment I went straight to look at this.
> 
> Yeah, I saw this in the settings for the split percents too. I'll take a look and ensure there aren't any other instances that can be impacted. It is introduced when I enhanced the strength of the test on the values coming from storage. Before we just checked they were objects, I checked the properties were of the right type.

Fixed the one instance where this could happen. The other places look good to me, but let me know if I missed somewhere.

scripts/page_browse.js
```diff
@@ -145,7 +151,7 @@ function makeTextWidget(container, splitter = false, reLayout = null) {
 var viewSplitter = function(container, storageKey) {
     const storageKeyLayout = storageKey + "-layout";
     let layout = JSON.parse(localStorage.getItem(storageKeyLayout));
-    if(!$.isPlainObject(layout)) {
+    if(!layout || layout.splitDirection !== "horizontal" || layout.splitDirection !== "vertical") {
```
70ray, line 154: Should be
`if(!layout || (layout.splitDirection !== "horizontal" && layout.splitDirection !== "vertical")) {`
as you have elsewhere.

70ray (changes requested):

70ray (commented):
What you have is very rigorous but I think it would be sufficient to test the JSON.parse and set defaults like you have in show_word_context.js.

srjfoo:
It looks like the font and font size are persistent now, but if I switch from horizontal mode to vertical mode, and refresh the page, it's still switching back to horizontal mode when displaying image+text. This does not happen on the main TEST site.

70ray:
> It looks like the font and font size are persistent now, but if I switch from horizontal mode to vertical mode, and refresh the page, it's still switching back to horizontal mode when displaying image+text. This does not happen on the main TEST site.

The change I suggested above should have fixed that. Looks like Chris forgot to update the sandbox.

chrismiceli:
> > It looks like the font and font size are persistent now, but if I switch from horizontal mode to vertical mode, and refresh the page, it's still switching back to horizontal mode when displaying image+text. This does not happen on the main TEST site.
> 
> The change I suggested above should have fixed that. Looks like Chris forgot to update the sandbox.

updated now, busy weekend :).

srjfoo (approved):

PR #902, opened 2023-05-20
Escape sql wildcards for Project Search

windymilla:
As per task #2018, underscore was treated as a wildcard character, e.g. in PPer field in Project Search.

Searching for a username on the Stats page already escaped percent and underscore. Escape the sql
wildcard characters when a LIKE comparison is being done in Project Search

### DIFF
--- a/pinc/DPDatabase.inc
+++ b/pinc/DPDatabase.inc
@@ -82,6 +82,11 @@ final class DPDatabase
         return mysqli_real_escape_string(self::$_connection, $value);
     }
 
+    public static function escape_like_wildcards($value)
+    {
+        return addcslashes($value, "%_[]^-");
+    }
+
     public static function query($sql, $throw_on_failure = true, $log_on_failure = true)
     {
         try {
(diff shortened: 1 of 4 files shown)


windymilla:
Sandbox at: https://www.pgdp.org/~windymilla/c.branch/sql-wildcards

pinc/DPDatabase.inc
```diff
@@ -82,6 +82,11 @@ final class DPDatabase
         return mysqli_real_escape_string(self::$_connection, $value);
     }
 
+    public static function escape_like_wildcards($value)
+    {
+        return addcslashes($value, "%_[]^-");
```
cpeel, line 87: According to https://dev.mysql.com/doc/refman/8.0/en/string-comparison-functions.html#operator_like it's only `_` and `%` as `LIKE` wildcards. The other looks like they're for regex-based patterns which I wouldn't include here.

70ray (approved):

pinc/DPDatabase.inc
windymilla: Thanks - removed superfluous wildcards from new `escape_like_wildcards` function

srjfoo (approved):

chrismiceli (approved):

PR #903, opened 2023-05-21
mark an all-bold paragraph as a possible issue.

70ray:
Do not attempt to identify section headings since the markup does not distinguish between a paragraph and a section heading, but mark an all-bold paragraph as a possible issue.

Continue to mark all-bold headings and sub-headings as definite issues.

add test

Sandbox at: https://www.pgdp.org/~rp31/c.branch/bold_para

### DIFF
--- a/tools/proofers/preview.js
+++ b/tools/proofers/preview.js
@@ -540,15 +540,16 @@ window.addEventListener('DOMContentLoaded', function() {
         }
 
         function testBoldBlock(block) {
-            if((txt.indexOf("<b>", block.start) === block.start) && (txt.indexOf("</b>", block.start) === (block.end - 4))) {
+            if((txt.indexOf("<b>", block.start) === block.start) && (txt.indexOf("</b>", block.start) === (block.end - "</b>".length))) {
                 // heading: issue, paragraph: possible issue
                 if((block.type == blockType.HEAD) || (block.type == blockType.SUBHEAD)) {
                     reportIssue(block.start, 3, "noBold");
                 } else {
                     // highlight after tags to avoid interleaved spans with style markup
-                    let markPoint = block.start + 3;
+                    let markPoint = block.start + "<b>".length;
+                    // if there is another start tag here advance past it.
                     while(txt.charAt(markPoint) == '<') {
-                        markPoint = txt.indexOf('>', markPoint + 2) + 1;
+                        markPoint = txt.indexOf('>', markPoint) + 1;
                     }
                     reportIssue(markPoint, 1, "boldPara");
                 }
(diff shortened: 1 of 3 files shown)


srjfoo (approved):

tools/proofers/preview.js
```diff
@@ -66,7 +66,6 @@ function removeTrail(txt) {
 
 const blockType = {
     PARA: 0,
-    SECHEAD: 1,
```
chrismiceli, line 69: since we removed this entry, should we renumber the later ones?

tools/proofers/preview.js
```diff
@@ -544,29 +539,20 @@ window.addEventListener('DOMContentLoaded', function() {
...
-                case blockType.SECHEAD:
-                case blockType.SUBHEAD:
-                    checkHead("noBold", txt, block);
-                    break;
-                default:
-                    break;
+        function testBoldBlock(block) {
+            if((txt.indexOf("<b>", block.start) === block.start) && (txt.indexOf("</b>", block.start) === (block.end - 4))) {
+                // heading: issue, paragraph: possible issue
+                if((block.type == blockType.HEAD) || (block.type == blockType.SUBHEAD)) {
```
chrismiceli, line 545: *minor* no need for parentheses.

tools/proofers/preview.js
```diff
@@ -544,29 +539,20 @@ window.addEventListener('DOMContentLoaded', function() {
...
-                switch(block.type) {
-                case blockType.HEAD:
-                case blockType.SECHEAD:
-                case blockType.SUBHEAD:
-                    checkHead("noBold", txt, block);
-                    break;
-                default:
-                    break;
+        function testBoldBlock(block) {
+            if((txt.indexOf("<b>", block.start) === block.start) && (txt.indexOf("</b>", block.start) === (block.end - 4))) {
```
chrismiceli, line 543: the indices we see (`- 4`, `+ 3`, `+ 2`) could be easier if we explained them some, or computed them from a fixed value? Like:
```
if((txt.indexOf("<b>", block.start) === block.start) &&
   (txt.indexOf("</b>", block.start) === (block.end - "</b>".length))) {
```
obviously subjective, just thinking out loud.

chrismiceli (approved):

tools/proofers/preview.js
70ray: I don't think we need to since they are just arbitrary numbers.

tools/proofers/preview.js
70ray: That seems like a good idea.

tools/proofers/preview.js
chrismiceli: No, definitely not required, so makes sense to leave as it to reduce risk of change.

PR #904, opened 2023-05-21
Append newline to email footers

windymilla:
Previously, no final newline was included in the text emails. When HTML emails were added, no final `<br>` was added. This only appears to cause an issue where the email is send to a mailing list, e.g. PPV mailing list. It appears that a mailing list footer is appended, which probably includes a newline followed by a line of underscores then some info about the mailing list. This worked OK with text messages since the newline put the underscores onto the line below, but with the HTML version, a newline does nothing, and a `<br>` is needed.

In fact, it seems reasonable that all our emails, text and HTML, should have a terminating newline/`<br>`.

### DIFF
--- a/pinc/send_mail.inc
+++ b/pinc/send_mail.inc
@@ -120,11 +120,12 @@ function get_standard_email_footers()
         $site_url,
         "",
         $automated_message,
+        "",
     ]);
     $html_footer = implode("\n", [
         "<hr style='width:12em; margin-left:0; margin-top:2em;'>",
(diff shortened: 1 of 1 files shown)


windymilla:
Sandbox at: https://www.pgdp.org/~windymilla/c.branch/html-newline
You can't actually see the original problem, since that only showed when the email was sent to a mailing list.

srjfoo (approved):

chrismiceli (approved):

PR #906, opened 2023-05-28
Only update user_project_info.t_latest_home_visit for PP/PPVers

cpeel:
The `user_project_info` tracks the last time a user visits a project page, proofs a project page, or subscribes to a project event. The table definition looks like:
```
+----------------------+--------------+------+-----+---------+-------+
| Field                | Type         | Null | Key | Default | Extra |
+----------------------+--------------+------+-----+---------+-------+
| username             | varchar(25)  | NO   | PRI | NULL    |       |
| projectid            | varchar(22)  | NO   | PRI | NULL    |       |
| t_latest_home_visit  | int unsigned | NO   |     | 0       |       |
| t_latest_page_event  | int unsigned | NO   |     | 0       |       |
| iste_round_available | tinyint(1)   | NO   |     | 0       |       |
| iste_round_complete  | tinyint(1)   | NO   |     | 0       |       |
| iste_pp_enter        | tinyint(1)   | NO   |     | 0       |       |
| iste_sr_available    | tinyint(1)   | NO   |     | 0       |       |
| iste_sr_complete     | tinyint(1)   | NO   |     | 0       |       |
| iste_ppv_enter       | tinyint(1)   | NO   |     | 0       |       |
| iste_posted          | tinyint(1)   | NO   |     | 0       |       |
| iste_sr_reported     | tinyint(1)   | NO   |     | 0       |       |
+----------------------+--------------+------+-----+---------+-------+
```
At pgdp.net the `user_project_info` table is the 3rd largest table and has 4 million rows in it. Almost half of the rows are from users who have visited a project page but have never proofed a page or subscribed to an event.

The only place in the code that uses `user_project_info.t_latest_home_visit` is the PP/PPV notification code in `pinc/post_processing.inc`. Change it so that this field is only updated for current PP/PPVers of a project and add an upgrade script that will delete any rows that don't satisfy this new requirement. This will delete about half the rows in that table.

This is a "quick" fix for what should be a larger architectural change of moving the "is the PP/PPV still paying attention to this project" date in the `projects` table rather than the `user_project_info` table, and using a more active mechanism than just visiting the page, but this change will immediately stop this table from growing so quickly and should decrease the periodic lock timeouts we see on it.

The code is live in the [only-update-visit-for-pper](https://www.pgdp.org/~cpeel/c.branch/only-update-visit-for-pper/) sandbox.

I've tested the update script on 3 projects on TEST and it worked as-expected:
```
projectID44dcf8bc3a74f: rows with non-zero t_latest_home_visit to delete = 11; to retain = 0
projectID44dcfc8b0b855: rows with non-zero t_latest_home_visit to delete = 16; to retain = 0
projectID536a29867c0c4: rows with non-zero t_latest_home_visit to delete = 7; to retain = 1
```
For the last one, `libby` is the PPer (and PPVer) of `projectID536a29867c0c4` and their record was retained along with two other rows with people who had proofed a page in the project.

### DIFF
--- a/project.php
+++ b/project.php
@@ -95,10 +95,13 @@ if (!$user_is_logged_in) {
     return;
 }
 
-// We only want to update the latest home visit if the user is a PPer or PPVer
-// for this project.
-if ($user_is_logged_in && ($project->PPer_is_current_user or $project->PPVer_is_current_user)) {
-    upi_set_t_latest_home_visit($pguser, $project->projectid, time());
+if ($user_is_logged_in) {
+    upi_set_t_latest_home_visit(
+        $pguser,
+        $project->projectid,
+        time(),
+        $project->PPer_is_current_user or $project->PPVer_is_current_user
+    );
 }
 
 if ($detail_level == 1) {
--- a/project.php
+++ b/project.php
@@ -95,7 +95,9 @@ if (!$user_is_logged_in) {
     return;
 }
 
-if ($user_is_logged_in) {
+// We only want to update the latest home visit if the user is a PPer or PPVer
+// for this project.
+if ($user_is_logged_in && ($project->PPer_is_current_user or $project->PPVer_is_current_user)) {
     upi_set_t_latest_home_visit($pguser, $project->projectid, time());
 }
 
--- a/SETUP/upgrade/19/20230601_prune_user_project_info.php
+++ b/SETUP/upgrade/19/20230601_prune_user_project_info.php
@@ -1,0 +1,79 @@
+<?php
+$relPath = '../../../pinc/';
+include_once($relPath.'base.inc');
+include_once($relPath.'Project.inc');
+
+header('Content-type: text/plain');
+
+// ------------------------------------------------------------
+
+// We've changed the t_latest_home_visit field to only track PP/PPV visits.
+// Delete rows from the user_project_info table that ONLY have a value for this
+// field and only if the row is for a non-PP/PPVer of the project. This
+// deletes entries that were only added by users who opened the project page
+// but didn't otherwise interact with the project (proofing a page, subscribing
+// to events, etc).
+
+echo "Pruning entries in user_project_info\n";
+$sql = "
+    SELECT projectid
+    FROM projects
+    ORDER BY projectid
+";
+
+$result = mysqli_query(DPDatabase::get_connection(), $sql) or die(mysqli_error(DPDatabase::get_connection()));
+
+while ([$projectid] = mysqli_fetch_row($result)) {
+    $project = new Project($projectid);
+    $sql = sprintf("
+        SELECT username
+        FROM user_project_info
+        WHERE projectid = '%s'
+            AND t_latest_page_event = 0
+            AND iste_round_available = 0
+            AND iste_round_complete = 0
+            AND iste_pp_enter = 0
+            AND iste_sr_available = 0
+            AND iste_sr_complete = 0
+            AND iste_ppv_enter = 0
+            AND iste_posted = 0
+            AND iste_sr_reported = 0
+        ORDER BY username
+    ", DPDatabase::escape($projectid));
+
+    $upi_result = mysqli_query(DPDatabase::get_connection(), $sql) or die(mysqli_error(DPDatabase::get_connection()));
+
+    $upi_users_to_delete = [];
+    $upi_users_to_retain = [];
+    while ([$username] = mysqli_fetch_row($upi_result)) {
+        if ($username != $project->PPer and $username != $project->PPVer) {
+            $upi_users_to_delete[] = $username;
+        } else {
+            $upi_users_to_retain[] = $username;
+        }
+    }
+    $upi_rows_delete = count($upi_users_to_delete);
+    $upi_rows_retain = count($upi_users_to_retain);
+
+    // skip the table if there are no rows to delete
+    if ($upi_rows_delete == 0) {
+        continue;
+    }
+
+    // delete the row -- because ($projectid, $username) is the primary key
+    // we can delete it just based on those two fields alone
+    $sql = sprintf("
+        DELETE FROM user_project_info
+        WHERE projectid = '%s'
+            AND username in (%s)
+    ", DPDatabase::escape($projectid),
+    surround_and_join(array_map("DPDatabase::escape", $upi_users_to_delete), "'", "'", ","));
+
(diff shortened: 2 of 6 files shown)


70ray (approved):

jmdyck:
In `project.php`, we check whether the project comments have changed 'recently' and alter the "Start Proofreading" blurbs accordingly ("Project Comments have changed! Please scroll down ..."). The test for recency is whether the comments have changed since the user's last page-save in the project. I *think* one of the ideas behind `user_project_info.t_latest_home_visit` was to use it instead of time-of-last-save for that test, since the last visit to the home page was the last time that they could have read the project comments.

(@srjfoo, @lhamilton1: does that sound at all familiar? Or worthwhile?)

That change never happened, but it could, and without preventing the proposed pruning of the upi table. That is, we'd keep logging `t_latest_home_visit` for regular proofers (not just PP/PPV), BUT ONLY if they've already got a upi row (after the pruning) for some other reason (i.e., they've saved a page or subscribed to an event). I think the only code change with respect to this PR would be around `upi_set_t_latest_home_visit`: if the upi row already exists, update the `t_latest_home_visit` column; if it doesn't exist, don't create it just for that purpose (except for the PPer or PPVer, which the caller could convey with a new parameter, say `$ensure_row`). I.e., if `$ensure_row` is true, use the existing `INSERT ... ON DUPLICATE KEY UPDATE` command; if it's false, just use an `UPDATE ... WHERE` command.

lhamilton1:
Are there any stats that would be affected or the calculations for the marking of voters for the DPF Board?

srjfoo:
> I _think_ one of the ideas behind `user_project_info.t_latest_home_visit` was to use it instead of time-of-last-save for that test, since the last visit to the home page was the last time that they could have read the project comments.
> 
> (@srjfoo, @lhamilton1: does that sound at all familiar? Or worthwhile?)

@jmdyck, it doesn't sound familiar, but that doesn't mean anything. I could easily have missed that discussion, or forgotten it. However, I'm not sure that `user_project_info.t_latest_home_visit` would do what the current method does. Isn't one of the points to force the proofreader to at least scroll past the PCs and have to use the "Start Proofreading" link at the bottom of the page? I may be totally wrong, but if `user_project_info.t_latest_home_visit` were used, wouldn't the user just have to refresh the page to be able to see the link at the top? Wouldn't `user_project_info.t_latest_page_event` be more appropriate for that?

> Are there any stats that would be affected or the calculations for the marking of voters for the DPF Board?

@lhamilton1, I don't see that field used in any of the election steps, but @jmdyck would know more about that than I.

cpeel:
That's a great suggestion @jmdyck - I've added a new commit to do just that. This at least enables possible future uses of that column without contributing to the ballooning.

windymilla (approved):
Looks good - thanks

srjfoo:
Note that the following may not be relevant for this PR, but be an existing problem, but I wanted to bring it up.

One of the things I picked up on was this comment:
```
    // To handle the (very rare) case where the user has never visited the
    // project page, we need the IFNULL() function to map the visit time to 0.
```
One of my test users (Sara R) was set up at one point as PPer for a project whose project page the user had never visited (unfortunately, I accidentally caused it to be visited just now when composing this).

This screenshot shows "Days since visit" as zero for "Inanity can be contagious" (projectID564a504488cb6).
image

Checking `user_project_info.t_latest_home_visit` for user `Sara R` showed this, which confirms that there wasn't an entry for that project:
image

Thinking that it might be useful to have another example for testing, I assigned another project to Sara R, [Ask a woman who knows {F2->F2}](https://www.pgdp.org/c/project.php?id=projectID45447792b4a5a). In [user_pp for Sara R](https://www.pgdp.org/noncvs/user_pp.php?username=Sara+R) "Days since visit" shows as an empty cell, not zero as it does on the PP page. (both should probably show the same thing)

So, my questions:

* On the PP page (or wherever some indication of last visit is shown), shouldn't it say "never" instead of just reading zero (which implies that the most recent visit was zero days ago, or "today")?
* Does it mess with the PP notification script?

cpeel:
Good catch, Sharon. I think the problem is that we're forcing a null value into a float on output for the PP listing:
```php
$cell = sprintf("%.1f", $book[$col_id]);
```
which is rendering it as `0.0`. I'll update this to special-case it for null values.

Looks like I'll also need to special-case the PP notification email too.

cpeel:
I've added a new commit that will show `Never` on the PP page and in the PP notification email if the project page has never been visited. I've also updated both `user_pp.php` and `user_ppv.php` page on TEST and PROD to show `Never` for this case as well.

srjfoo:
> I've added a new commit that will show `Never` on the PP page and in the PP notification email if the project page has never been visited. I've also updated both `user_pp.php` and `user_ppv.php` page on TEST and PROD to show `Never` for this case as well.

Did the sandbox get updated? It's still showing `0.0` for my test case instead of `Never`. For `user_pp.php`, you caught the case for the reserved books, but not the "Projects currently being PPed" section, which still shows as an empty cell.

project.php
```diff
@@ -96,7 +96,12 @@
 }
 
 if ($user_is_logged_in) {
-    upi_set_t_latest_home_visit($pguser, $project->projectid, time());
+    upi_set_t_latest_home_visit(
+        $pguser,
+        $project->projectid,
+        time(),
+        $project->PPer_is_current_user or $project->PPVer_is_current_user
```
chrismiceli, line 103: *minor* any reason for `or` vs `||`? I have never seen `or` before, but php isn't my expertise.

chrismiceli (approved):

project.php
cpeel: This is probably my python coming out. I think we've used `||` primarily before although it looks like [both work](https://www.php.net/manual/en/language.operators.logical.php). I'll see what we've done in other parts of the code and mirror that.

cpeel:
> Did the sandbox get updated?

It did not but it is now.

project.php
srjfoo: My understanding is that they have different precedence.

srjfoo (approved):

project.php
cpeel: The code seems to use both and I think `or` is clearer so let's keep it like this.

cpeel:
Follow-up from this change on PROD (posted in Slack, but it's useful to have it here):
> [After this change the] `user_project_info` table is now ~30% of the size that it was originally.
> (down from 977mb to 268mb)

PR #908, opened 2023-06-01
Restore previous Reply-To email behavior

windymilla:
Specifically, if a From address is given to override the default, don't specify the Reply-To address.
Previous HTML email work meant the Reply-To address got set to the default address, unless it was
explicitly set.


windymilla:
Sandbox at: https://www.pgdp.org/~windymilla/c.branch/email-reply-to

cpeel (approved):

srjfoo (approved):

PR #913, opened 2023-06-11
No block blank

70ray:
While considering re-making the re-wrap function, it appeared that blank lines after the block-quote start tag and before the block-quote end tag would have some peculiar effects. Rather trying to work around these, it seemed better to ensure that such blank lines are not present.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/no_block_blank


70ray:
There are some counter -examples in the LoFE. Re-consider.

PR #916, opened 2023-06-15
Permit ampersands in team names

windymilla:
Team names and webpage were passed through `stripAllString` which is intended to convert bb markup in team info to HTML markup, e.g. for italic. This conversion is not needed for team name and webpage.

Since `stripAllString` calls `html_safe`, ampersands were getting encoded, then encoded again via `attr_safe` in `ShowEdit`.

Addresses task #2041

Sandbox at: https://www.pgdp.org/~windymilla/c.branch/team-name

### DIFF
--- a/stats/teams/new_team.php
+++ b/stats/teams/new_team.php
@@ -12,9 +12,9 @@ $user = User::load_current();
 
 $theme_extra_args = ["js_data" => get_newHelpWin_javascript("$code_url/pophelp.php?category=teams&name=edit_")];
 
-$teamname = stripAllString(trim(array_get($_POST, "teamname", "")));
+$teamname = trim(array_get($_POST, "teamname", ""));
 $text_data = stripAllString(array_get($_POST, "text_data", ""));
-$teamwebpage = stripAllString(array_get($_POST, "teamwebpage", ""));
+$teamwebpage = array_get($_POST, "teamwebpage", "");
 $tavatar = array_get($_POST, "tavatar", "");
 $ticon = array_get($_POST, "ticon", "");
 


srjfoo (approved):

chrismiceli (approved):

stats/teams/new_team.php
```diff
@@ -12,9 +12,9 @@
 
 $theme_extra_args = ["js_data" => get_newHelpWin_javascript("$code_url/pophelp.php?category=teams&name=edit_")];
 
-$teamname = stripAllString(trim(array_get($_POST, "teamname", "")));
+$teamname = trim(array_get($_POST, "teamname", ""));
```
cpeel, line 15: Does this change alter how team names appear in team forums when they are created? I forget how this crazy team code works but that was the only other phpBB-interfacing piece that came to mind.

windymilla: It seems to display team names OK in the forum too. I created a team `test & test again` on TEST. 
Team details here: https://www.pgdp.org/c/stats/teams/tdetail.php?tid=68
Team forum thread here: https://www.pgdp.org/phpBB3/viewtopic.php?t=752
Searching for `&` on the Team List page finds the new team and displays the name correctly there.

srjfoo: I checked the team I created in this sandbox, also, and the team discussion was created correctly (the team that has the `&amp;` explicitly in the team name is one I created in the main test site as a control, not a failure of @windymilla's code).

cpeel: Awesome, thanks for confirming that!

PR #917, opened 2023-06-21
Handle month/day given without Birthday/Otherday code

windymilla:
Requested by task #836

If month/day is given on project edit page, but special code field is left empty, set special code to `Birthday` and output error message for user to check before saving.

Fix is not as slick as it might be, because month/date are not stored in the Project structure or database except when appended to Birthday/Otherday as a special code. This is trying to cope with the case when month/date are set, but special code was left blank. Previously, project would have been saved with no special day code at all, leaving it vulnerable to releasing before the PM realised the mistake.

Sandbox at: https://www.pgdp.org/~windymilla/c.branch/special-date


cpeel (approved):
This is a very clever solution and far better than silently failing.

srjfoo (approved):

70ray (approved):

chrismiceli (approved):

PR #918, opened 2023-06-21
Check if team_id exists before creating topic

cpeel:
If someone manipulates the team discussion link on the Teams page to use a numeric, but invalid, team ID the script blindly plunges forward. Instead, if we can't find a team ID just error out. This page is only accessible from the teams page.

Available in https://www.pgdp.org/~cpeel/c.branch/better-checks-team-discussion/


srjfoo (approved):

chrismiceli (approved):

cpeel:
Small update to use the same exception type we use in `misc.inc` in the other parameter validation functions.

srjfoo (approved):

PR #919, opened 2023-06-24
Standardize "Return to x" links

windymilla:
As per task #1871, linked phrases such as
"Return to the project page" were inconsistently worded, puncutated and marked up for translation.

Standardize return to project, round and Activity Hub.

Note that further standardization of capitalization of "project page" could be done in future, where it does not appear just as a simple "Return to" link.

Sandbox at: https://www.pgdp.org/~windymilla/c.branch/return-links

### DIFF
--- a/pinc/links.inc
+++ b/pinc/links.inc
@@ -81,17 +81,36 @@ function private_message_link($proofer_username, $target = "_blank")
 
 // -----------------------------------------------------------------------------
 
-function return_to_project_page_link($projectid, $url_suffix = "")
-// Returns a string containing a snippet of HTML to return to a project page.
-// Optional second argument could be page anchor, e.g "#holds"
-// or additional URL parameter, e.g. "&detail_level=4"
+function project_page_link_url($project_id, $query_param_array = null, $anchor_link = "")
+// Returns HTML-safe URL for the given project page.
+// Optional second argument is array of query parameters (without ?/&), e.g. ["detail_level=4"]
+// Optional third argument is page anchor (without #), e.g "holds"
+// NOTE: query_params and anchor_link should be urlencoded if necessary
 {
     global $code_url;
 
-    $url_suffix = html_safe($url_suffix);
+    $url = "$code_url/project.php?id=$project_id";
+    if (!empty($query_param_array)) {
+        $url .= "&" . implode("&", $query_param_array);
+    }
+    if (!empty($anchor_link)) {
+        $url .= "#" . $anchor_link;
+    }
+    return html_safe($url);
+}
+
+// -----------------------------------------------------------------------------
+
+function return_to_project_page_link($project_id, $query_param_array = null, $anchor_link = "")
+// Returns a string containing an HTML link to return to a project page.
+// Optional second argument is array of query parameters (without ?/&), e.g. ["detail_level=4"]
+// Optional third argument is page anchor (without #), e.g "holds"
+// NOTE: query_params and anchor_link should be urlencoded if necessary
+{
+    $url = project_page_link_url($project_id, $query_param_array, $anchor_link);
     return sprintf(
              _("Return to the <a %s>project page</a>"),
-             "href='$code_url/project.php?id=$projectid$url_suffix' target='_top'"
+             "href='$url' target='_top'"
            );
 }
 
--- a/pinc/links.inc
+++ b/pinc/links.inc
@@ -78,3 +78,46 @@ function private_message_link($proofer_username, $target = "_blank")
 
     return $a;
 }
+
+// -----------------------------------------------------------------------------
+
+function return_to_project_page_link($projectid, $url_suffix = "")
+// Returns a string containing a snippet of HTML to return to a project page.
+// Optional second argument could be page anchor, e.g "#holds"
+// or additional URL parameter, e.g. "&detail_level=4"
+{
+    global $code_url;
+
+    $url_suffix = html_safe($url_suffix);
+    return sprintf(
+             _("Return to the <a %s>project page</a>"),
+             "href='$code_url/project.php?id=$projectid$url_suffix' target='_top'"
+           );
+}
+
+// -----------------------------------------------------------------------------
+
+function return_to_round_page_link($round_id)
+// Returns a string containing a snippet of HTML to return to a round page.
+{
+    global $code_url;
+
+    return sprintf(
+             _("Return to <a %s>round %s</a>"),
+             "href='$code_url/tools/proofers/round.php?round_id=$round_id' target='_top'",
+             $round_id
+           );
+}
+
+// -----------------------------------------------------------------------------
+
+function return_to_activity_hub_link()
+// Returns a string containing a snippet of HTML to return to the Activity Hub.
+{
+    global $code_url;
+
+    return sprintf(
+             _("Return to the <a %s>Activity Hub</a>"),
+             "href='$code_url/activity_hub.php' target='_top'"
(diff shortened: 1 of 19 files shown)


pinc/links.inc
```diff
@@ -78,3 +78,46 @@ function private_message_link($proofer_username, $target = "_blank")
 
     return $a;
 }
+
+// -----------------------------------------------------------------------------
+
+function return_to_project_page_link($projectid, $url_suffix = "")
+// Returns a string containing a snippet of HTML to return to a project page.
+// Optional second argument could be page anchor, e.g "#holds"
+// or additional URL parameter, e.g. "&detail_level=4"
```
chrismiceli, line 87: This interface feels error prone since it takes query params or fragments, but also requires the `&`, but you can't send the `?`. Would it make more sense to split them out?
```
function return_to_project_page_link($projectId, $query_param_array, $fragment)
```
Then the caller doesn't need to know about `?`, etc.?

pinc/links.inc
```diff
@@ -78,3 +78,46 @@ function private_message_link($proofer_username, $target = "_blank")
 
     return $a;
 }
+
+// -----------------------------------------------------------------------------
+
+function return_to_project_page_link($projectid, $url_suffix = "")
```
chrismiceli, line 84: we should mention that the url_suffix needs to be url encoded since we don't encode before we craft the HTML. That or we should encode and make sure the callers don't encode.

pinc/links.inc
```diff
@@ -78,3 +78,46 @@ function private_message_link($proofer_username, $target = "_blank")
...
+    $url_suffix = html_safe($url_suffix);
+    return sprintf(
+             _("Return to the <a %s>project page</a>"),
+             "href='$code_url/project.php?id=$projectid$url_suffix' target='_top'"
+           );
+}
+
+// -----------------------------------------------------------------------------
+
+function return_to_round_page_link($round_id)
```
chrismiceli, line 100: probably not just an issue here, but we use `projectid` above, but `round_id` here. Ideally they'd be the same casing strategy I imagine.

tools/extend_sr.php
```diff
@@ -19,7 +20,8 @@ class ActionException extends Exception
 
 $title = _("Extend the Smooth Reading deadline");
 $new_state = PROJ_POST_FIRST_CHECKED_OUT;
-$back_url = "$code_url/project.php?id=$projectid&amp;expected_state=$new_state#smooth_start";
+$suffix = "&amp;expected_state=$new_state#smooth_start";
```
chrismiceli, line 23: if we took the query params, we could avoid this `& amp ;` stuff (spaces for github)

windymilla: Thanks @chrismiceli 
So, if I understand correctly, for this case, I would end up calling it something like this:
```
$query_param_array = array("expected_state=$new_state");
return_to_project_page_link($projectid, $query_param_array, "#smooth_start");
```
Is that what you're suggesting? Presumably also with `$query_param_array` and `$fragment` being optional arguments to `return_to_project_page_link`, since neither of them is needed for many calls?

Final question (for now): Do you also recommend allowing the function to be called with a single string in place of the array argument when only one query param is needed (most cases), i.e. 
```
return_to_project_page_link($projectid, "expected_state=$new_state", "#smooth_start");
```
I believe that's permitted in PHP if I check it inside the function, isn't it? Or do you think that is an unnecessary complication and prone to errors?

cpeel: I think the query-params-as-array approach is a more robust one. Probably better for the second arg to either be an array or Null and just have the caller wrap a singe arg in `[]`s, eg:
```php
return_to_project_page_link($projectid, ["expected_state=$new_state"], "#smooth_start");

```

srjfoo (commented):
I've tested a bunch of the messages, but if there are going to be changes, I should probably wait to finish the testing until that's in place. So far, the only thing other than a couple I couldn't figure out how to trigger the message on (and nudged @windymilla  about) have worked just fine, and I just had the one query about harmonizing another link in `add_files.php`.

tools/extend_sr.php
chrismiceli: Yes, I like what cpeel has. And yes, if we can, support defaulting the 2nd and 3rd parameters since they aren't always needed. We could also avoid the `#` on the fragment argument since it will always be present.

pinc/links.inc
windymilla: Done

pinc/links.inc
windymilla: You're right - there is inconsistency in the existing code. I chose to keep underscores, since otherwise the query parameter would read `?round_id=$roundid`, so I changed the parameter of the project page functions to `project_id`. The variable `$projectid` is used in many other places though, so difficult to find a perfect solution.

tools/project_manager/add_files.php
```diff
@@ -173,7 +174,7 @@
 }
 
 echo "<hr>\n";
-echo "Return to <a href='$code_url/project.php?id=$projectid&detail_level=4'>Project Page</a>.\n";
+echo return_to_project_page_link($projectid, "&detail_level=4") . "\n";
```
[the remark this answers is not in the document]
windymilla: A function returning just the URL works well, without breaking up the sentence as would happen if it returned `<a href...>project page</a>`, leaving "Return to the" to be translated independently, which is not sufficiently defined for some languages.

tools/upload_text.php
```diff
@@ -153,7 +154,7 @@
 $return_anchor = "<a href='$back_url'>$back_blurb</a>";
 // TRANSLATORS: %s is an already-translated page name, eg: Project Page
 $return_message = "<p>". sprintf(_("Return to the %s"), $return_anchor). "</p>";
-$return_to_project_link = "<a href='$code_url/project.php?id=$projectid'>" . _("Return to the Project Page") . "</a>\n";
+$return_to_project_link = return_to_project_page_link($projectid). "\n";
```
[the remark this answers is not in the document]
windymilla: Done

windymilla:
@srjfoo Sandbox updated: https://www.pgdp.org/~windymilla/c.branch/return-links

pinc/links.inc
cpeel: I appreciate that we are inconsistent between `$projectid` and `$round_id`, but we're pretty consistent in those two forms being used throughput the code and suggest we stick with `$projectid` in the functions above.

tools/project_manager/add_files.php
```diff
@@ -632,7 +633,8 @@ public function display()
         echo "<br>\n";
         echo _("Please review the information, and use the link at the bottom of the page to proceed with the load.");
         echo "<br>\n";
-        echo _("If there's a problem, return to the  <a href='$code_url/project.php?id=$projectid&detail_level=4'>Project Page</a> without loading the files.");
+        $project_url = project_page_link_url($projectid, ["detail_level=4"]);
+        echo _("If there's a problem, return to the  <a href='$project_url'>project page</a> without loading the files.");
```
cpeel, line 637: It was broken when you got here, but we should make this more translatale while we're here:
```php
echo sprintf(_("If there's a problem, return to the <a %s>project page</a> without loading the files."), "href='$project_url'");
```

cpeel (commented):
Oh yeah - I like this direction! Two minor requests.

cpeel (approved):

70ray (approved):

srjfoo (approved):

PR #923, opened 2023-07-01
Do not permit Edit Task to close a task

windymilla:
As per task #1828, tasks aren't closed properly
if their status is just set to Closed via the
Edit (or Create) Task form. It needs to be done
via the button on the main task page.

Therefore, omit Closed from the list of statuses on the Edit page.

Sandbox at: https://www.pgdp.org/~windymilla/c.branch/close-tasks

### DIFF
--- a/tasks.php
+++ b/tasks.php
@@ -1911,7 +1911,8 @@ function title_string_for_task($pre_task)
 // Input string must be a valid property value
 function get_property_key($value, $array)
 {
-    $key = array_search($value, $array);
-    assert($key !== false);
+    if (($key = array_search($value, $array)) === false) {
+        throw new ValueError("Bad task property value: $value");
+    }
     return $key;
 }
--- a/tasks.php
+++ b/tasks.php
@@ -152,21 +152,9 @@ asort($categories_array);
 $tasks_status_array = [
     1 => "New",
     2 => "Accepted",
-    3 => "Duplicate",
-    4 => "Fixed",
-    5 => "Invalid",
-    6 => "Later",
-    7 => "None",
-    8 => "Out of Date",
-    9 => "Postponed",
-    10 => "Rejected",
-    11 => "Remind",
-    12 => "Won't Fix",
-    13 => "Works for Me",
     14 => "Closed",
     15 => "Reopened",
     16 => "Researching",
-    17 => "Implemented",
     18 => "In Progress",
 ];
 asort($tasks_status_array);
@@ -1173,8 +1161,9 @@ function TaskForm($task)
         $tasks_status_array = [get_property_key("New", $tasks_status_array) => "New"];
     }
 
-    // Don't want to permit setting status to "Closed" when creating/editing a task
+    // Don't want to permit setting status to "Closed" or "Reopened" when creating/editing a task
     unset($tasks_status_array[get_property_key("Closed", $tasks_status_array)]);
+    unset($tasks_status_array[get_property_key("Reopened", $tasks_status_array)]);
 
     $task_summary_enc = attr_safe($task->task_summary);
     $task_details_enc = html_safe($task->task_details);
--- a/tasks.php
+++ b/tasks.php
@@ -1160,6 +1160,9 @@ function TaskForm($task)
         $tasks_status_array = [1 => "New"];
     }
 
+    // Don't want to permit setting status to 'Closed' when creating/editing a task
+    unset($tasks_status_array[14]);
+
     $task_summary_enc = attr_safe($task->task_summary);
     $task_details_enc = html_safe($task->task_details);
 
--- a/tasks.php
+++ b/tasks.php
@@ -425,18 +425,27 @@ function SearchParams_get_url_query_string()
 
 function make_default_task_object()
 {
+    global $tasks_array;
+    global $categories_array;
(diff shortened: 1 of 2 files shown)


srjfoo (approved):
Looks good to me, but it occurred to me that there are several other statuses that imply closure, and shouldn't be set via the edit interface:
```
    3 => "Duplicate",
    4 => "Fixed",
    5 => "Invalid",
    8 => "Out of Date",
    10 => "Rejected",
    12 => "Won't Fix",
    13 => "Works for Me",
    17 => "Implemented",
```
`Reopened` should not be set from the edit interface, either.

Several statuses, I'm wondering whether we should get rid of altogether: Invalid, Out of Date, Rejected, Remind are not associated with any tasks on PROD. Is that simply because we chose some other reason for closing a task?

tasks.php
```diff
@@ -1160,6 +1160,9 @@ function TaskForm($task)
         $tasks_status_array = [1 => "New"];
     }
 
+    // Don't want to permit setting status to 'Closed' when creating/editing a task
+    unset($tasks_status_array[14]);
```
chrismiceli, line 1164: I realize it is the way it is currently, but I'm not a huge fan of [unnamed numerical constants](https://en.wikipedia.org/wiki/Magic_number_(programming)#Unnamed_numerical_constants)
It makes it hard to know what the code is doing instead of being self documented.

windymilla: I've done some work along those lines @chrismiceli - does that look better?

windymilla:
Updated https://www.pgdp.org/~windymilla/c.branch/close-tasks

tasks.php
chrismiceli: Must better in my opinion, thank you!

windymilla:
> Looks good to me, but it occurred to me that there are several other statuses that imply closure, and shouldn't be set via the edit interface:
> 
> ```
>     3 => "Duplicate",
>     4 => "Fixed",
>     5 => "Invalid",
>     8 => "Out of Date",
>     10 => "Rejected",
>     12 => "Won't Fix",
>     13 => "Works for Me",
>     17 => "Implemented",
> ```

No PROD tasks have any of the above statuses, apart from [one task](https://www.pgdp.net/c/tasks.php?action=show&task_id=576) which is "Implemented" (in a sandbox in 2009, but presumably never merged - should probably be addressed). 
I can remove all of the above statuses completely, if that seems best - there's no point in just hiding them from the Edit interface, since the only possible other use for them is during a Search (which you wouldn't do if there are none). The "Duplicate", "Implemented", etc., that are stored when a task is closed is a separate list of Closure Reasons - a separate list in the code, and a separate column in the database.

> 
> `Reopened` should not be set from the edit interface, either.
> 

Agreed - that is like "Closed" in that it is set automatically, rather than manually.

> Several statuses, I'm wondering whether we should get rid of altogether: Invalid, Out of Date, Rejected, Remind are not associated with any tasks on PROD. Is that simply because we chose some other reason for closing a task?

Invalid, Out of Date and Rejected are dealt with above. I agree that Remind can be also be removed completely.

What about Later and Postponed - do we need both of those?

And does it make sense to have a status of None - what does it mean? Just [one task](https://www.pgdp.net/c/tasks.php?action=show&task_id=669) on PROD has status None (could easily be edited to give it a non-None status). Then None could also be removed from the list of statuses.

cpeel:
I fully-endorse reducing the number of `closed_reason`s available and agree with the list above for removal. I don't think we need Later, Postponed, or None. If we do remove any reasons we should make an upgrade script that will change any of the removed values to some still-current value, both for PROD but also for any other sites using our code.

windymilla:
> I fully-endorse reducing the number of `closed_reason`s available and agree with the list above for removal. I don't think we need Later, Postponed, or None. If we do remove any reasons we should make an upgrade script that will change any of the removed values to some still-current value, both for PROD but also for any other sites using our code.

I think that's done. Please check the upgrade script carefully, since I have no way to test it out.
Updated https://www.pgdp.org/~windymilla/c.branch/close-tasks

tasks.php
```diff
@@ -1901,3 +1906,12 @@ function title_string_for_task($pre_task)
 {
     return sprintf(_("Task #%d: %s"), $pre_task->task_id, $pre_task->task_summary);
 }
+
+// Get key corresponding to a task properties string
+// Input string must be a valid property value
+function get_property_key($value, $array)
```
cpeel, line 1912: It looks like the value this function has over `array_search()` is that it checks to see if the value exists and fails if it doesn't, but it does so with `assert()` which are often [disabled in production](https://www.php.net/manual/en/ini.core.php#ini.zend.assertions) rendering this effectively the same thing. If we really want to fail on an invalid error we should throw a `ValueError` exception instead of using `assert()`.

I was also wondering if this should be in `misc.inc` as `array_get_key()` to mirror `array_get()` but the value in the latter is the ability to include a default which I don't think we really want to do here so I'd opt for keeping it here with the name you have and just throwing on error.

windymilla: If there's a need in the future for a more general `array_get_key()`, it can be renamed and moved to `misc.inc`, I suppose.
I've made the change you suggested.

srjfoo (approved):

PR #924, opened 2023-07-03
Disallow @s in new usernames

cpeel:
By squirrel request: to further discourage people using emails for their usernames, disallow `@`s in new registrations.

Testable in https://www.pgdp.org/~cpeel/c.branch/disallow-ats-in-new-usernames/

### DIFF
--- a/SETUP/tests/UserTest.php
+++ b/SETUP/tests/UserTest.php
@@ -124,4 +124,43 @@ class UserTest extends PHPUnit\Framework\TestCase
         $user = new User();
         $user->save();
     }
+
+
+    // Username validation tests
+
+    public function testValidRegUsername()
+    {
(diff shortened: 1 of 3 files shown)


srjfoo (approved):

lhamilton1:
I approve

windymilla (approved):
Works fine for me. 

I did wonder if it would be good for the error message to also say "Don't use your email address" to aid those who can't see why their username is being rejected. However, I don't think it's worth it - a user needs to be able to understand "Valid characters are..." not having the at-sign. So, not worth the extra code complication.

70ray (approved):

chrismiceli (approved):
Might be worth adding some phpunit tests to this.

cpeel:
Unit tests added!

PR #926, opened 2023-07-14
Rework forum code to be specific to phpBB3

cpeel:
This is a stepping stone to supporting more forum types than just phpBB, specifically to have a "dev only" forum shim to allow local development without installing phpBB. A follow-up PR will rename `pinc/forum_interface.inc` and it is much easier to review the changes made to that file before doing so.

The PR should have no functional changes in it.

This is in https://www.pgdp.org/~cpeel/c.branch/forum-abstraction/

### DIFF
--- a/pinc/forum_interface.inc
+++ b/pinc/forum_interface.inc
@@ -14,6 +14,8 @@ include_once($relPath."misc.inc"); // array_get()
 // support was removed to handle the possible future case where other
 // version conditionals are needed.
 
+define('PHPBB_VERSION', '3');
+
 /*
 Important notes about using phpBB code within DP
 ------------------------------------------------
@@ -44,7 +46,8 @@ assert(PHPBB_VERSION == 3, "Only phpBB version 3 is supported");
 */
 function phpbb_table($table)
 {
-    return PHPBB_TABLE_PREFIX . "_" . $table;
+    global $forums_phpbb_table_prefix;
+    return $forums_phpbb_table_prefix . "_" . $table;
 }
 
 function phpbb_lang($locale = false)
@@ -106,7 +109,6 @@ function create_forum_user($username, $password, $email, $password_is_digested =
 // Note: Use a strict comparison operator (=== or !==) against TRUE
 // to evaluate the pass/fail status of this function.
 {
-    global $forums_dir;
     if (PHPBB_VERSION == 3) {
         // Use the phpBB3 API to create the user, but do so via phpbb3.inc
         // as to not mix the DP and phpBB3 codespace.
@@ -151,9 +153,9 @@ function login_forum_user($username, $password)
 //         'too_many_attempts'
 //         'unknown'
 {
-    global $forums_dir;
+    global $forums_phpbb_dir;
 
-    if (!is_dir($forums_dir)) {
+    if (!is_dir($forums_phpbb_dir)) {
         return false;
     }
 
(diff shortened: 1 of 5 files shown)


srjfoo (approved):

windymilla (commented):
I just have a question relating to `get_url_for_forum()`: its purpose appears to be to avoid including the global `$forums_phpbb_url` everywhere. 
Are the globals `$forums_phpbb_table_prefix` and  `$forums_phpbb_dir` different in nature/usage so don't warrant a similar access function, or is it just that one hasn't been created yet? Or have I misunderstood the overall plan?

pinc/forum_interface.inc
```diff
@@ -44,7 +46,8 @@ assert(PHPBB_VERSION == 3, "Only phpBB version 3 is supported");
 */
 function phpbb_table($table)
```
chrismiceli, line 47: we're waiting to rename these methods right?

pinc/forum_interface.inc
```diff
@@ -168,7 +170,7 @@ function login_forum_user($username, $password)
 
     if (PHPBB_VERSION == 3) {
```
chrismiceli, line 171: Do we want to check forum type too at this point, or is that again later?

pinc/phpbb3.inc
```diff
@@ -193,7 +190,7 @@ function _insert_post(
     }
     $sql = "
         SELECT *
-        FROM {$phpbb_table_prefix}_users
+        FROM {$forums_phpbb_table_prefix}_users
```
chrismiceli, line 193: do we want to avoid sql concat in case `$username` or (much less likely so) the table prefix is malicious?

chrismiceli (approved):

pinc/forum_interface.inc
cpeel: No, as the whole file will be renamed `forum_interface_phpbb3.inc` in a follow-up PR so everything in here will be phpBB3-specific.

pinc/phpbb3.inc
cpeel: The table prefix is set in a configuration file, so if that has malicious values in it the admin as done it to themselves 😁 

But good call on the username. I'll get a commit up for that.

cpeel:
> I just have a question relating to `get_url_for_forum()`: its purpose appears to be to avoid including the global `$forums_phpbb_url` everywhere. Are the globals `$forums_phpbb_table_prefix` and `$forums_phpbb_dir` different in nature/usage so don't warrant a similar access function, or is it just that one hasn't been created yet? Or have I misunderstood the overall plan?

That's a fair question. I was on the fence about if we should use `get_url_for_forum()` or `$forums_phpbb_url` in those functions. I leaned towards the former because that function is part of the forum interface "API" (ie it's used elsewhere in the code) and not something internal to the phpBB-specific forum interface. But the other two variables are phpBB-specific so I left them as-is. If we think the discrepancy is confusing I'm happy to change the uses back to `$forums_phpbb_url`.

windymilla:
> > I just have a question relating to `get_url_for_forum()`: its purpose appears to be to avoid including the global `$forums_phpbb_url` everywhere. Are the globals `$forums_phpbb_table_prefix` and `$forums_phpbb_dir` different in nature/usage so don't warrant a similar access function, or is it just that one hasn't been created yet? Or have I misunderstood the overall plan?
> 
> That's a fair question. I was on the fence about if we should use `get_url_for_forum()` or `$forums_phpbb_url` in those functions. I leaned towards the former because that function is part of the forum interface "API" (ie it's used elsewhere in the code) and not something internal to the phpBB-specific forum interface. But the other two variables are phpBB-specific so I left them as-is. If we think the discrepancy is confusing I'm happy to change the uses back to `$forums_phpbb_url`.

I'm happy with that reasoning - thanks. It doesn't make sense to use `$forums_phpbb_url` directly when there's a function for it that is used elsewhere.

windymilla (approved):

chrismiceli (approved):

srjfoo (approved):

PR #929, opened 2023-07-27
Avoid duplicate project thread creation

windymilla:
Addresses Task #1353
If the "Start a Discussion" link on the project page is double-clicked, two project threads may be created.
Avoid this by replacing the link with plain text when it is clicked.

Sandbox:  https://www.pgdp.org/~windymilla/c.branch/double-forum

### DIFF
--- a/project.php
+++ b/project.php
@@ -548,7 +548,7 @@ function do_project_info_table()
             if ($topic_id == "") {
                 $blurb = html_safe(_("Start a discussion about this project"));
                 $url = "$code_url/tools/proofers/project_topic.php?project=$projectid";
-                echo_row_a(_("Forum"), "<a href='$url'>$blurb</a>");
+                echo_row_a(_("Forum"), "<a href='$url' onclick='this.outerHTML=\"$blurb\";'>$blurb</a>");
             } else {
                 $details = get_topic_details($topic_id);
                 if ($details) {


srjfoo (approved):
On testing, this seemed to work. Given a small amount of testing, without the change, one of two double-clicks created an extra topic. With the change, three project topics created correctly with no duplicate on double-click.

project.php
```diff
@@ -548,7 +548,7 @@ function do_project_info_table()
             if ($topic_id == "") {
                 $blurb = html_safe(_("Start a discussion about this project"));
                 $url = "$code_url/tools/proofers/project_topic.php?project=$projectid";
-                echo_row_a(_("Forum"), "<a href='$url'>$blurb</a>");
+                echo_row_a(_("Forum"), "<a href='$url' onclick='this.outerHTML=\"$blurb\";'>$blurb</a>");
```
chrismiceli, line 551: While this may work, it feels like the wrong place to fix this. Shouldn't we have logic in project_topic that prevents double thread creation? Like, maybe some sort of [double check lock](https://en.wikipedia.org/wiki/Double-checked_locking) during creation?

windymilla: Possibly, though I would not be confident implementing something like that, for fear of unintended consequences. It feels like a big solution to solve a small problem. I don't know if anything like that is currently used in the code, but if something like that is needed, I'm happy for this PR to be closed and for someone else to tackle it at some point.

cpeel: We have logic in [Project->ensure_topic()](https://github.com/DistributedProofreaders/dproofreaders/blob/master/pinc/Project.inc#L1251) that attempts to prevent double creation. The problem is that there's a race condition between the "create the topic" code we call out to phpBB for and the "link the new topic ID to the project" code immediately following. [Note that it's compounded by the topic_id being loaded when the Project object is created in project_topic.] I think it would be hard to implement a lock around this problem outside of the DB because the two page loads operate in different PHP process contexts.

chrismiceli: I'm definitely ignorant on a lot of the php complexities, but some quick googling lead me to https://www.php.net/manual/en/function.flock.php.
Would that be useful at all across the php processes?

cpeel: flock is really designed to be used against a workflow that is file-centric. We'd have to introduce some temporary file to lock against in one process to check on the second one and then delete when it was done on the first one. That doesn't feel like a good approach to this DB-centric workflow.

cpeel (approved):

project.php
windymilla: Thanks both for pondering it

PR #930, opened 2023-07-30
Raise exceptions 1

70ray:
This is the first stage of converting the functions we wish to use in a proofreading API to raise exceptions instead of returning messages.
It is similar to the previous version 'raise_exceptions' except that the validate_page_table_exists() function is omitted because it is not needed for the API.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/raise_exceptions_1

### DIFF
--- a/pinc/LPage.inc
+++ b/pinc/LPage.inc
@@ -137,8 +137,10 @@ function get_available_page($projectid, $proj_state, $pguser, &$err)
     }
 
     $round = get_Round_for_project_state($proj_state);
-    $err = can_user_get_pages_in_project($pguser, $project, $round);
-    if ($err) {
+    try {
+        validate_user_can_get_pages_in_project($pguser, $project, $round);
+    } catch (UserAccessException $exception) {
+        $err = $exception->getMessage();
         return null;
     }
 
--- a/pinc/LPage.inc
+++ b/pinc/LPage.inc
@@ -125,8 +125,6 @@ function get_available_page($projectid, $proj_state, $pguser, &$err)
 // Returns an LPage, unless no page is available,
 // in which case returns NULL and sets $err.
 {
-    global $preceding_proofer_restriction;
-
     // can throw NonexistentProjectException
     $project = new Project($projectid);
 
@@ -137,12 +135,28 @@ function get_available_page($projectid, $proj_state, $pguser, &$err)
     }
 
     $round = get_Round_for_project_state($proj_state);
+
     try {
-        validate_user_can_get_pages_in_project($pguser, $project, $round);
-    } catch (UserAccessException $exception) {
+        [$imagefile, $state] = get_available_proof_page_array($project, $round, $pguser);
+    } catch (ProjectException | UserAccessException $exception) {
         $err = $exception->getMessage();
         return null;
     }
+    $lpage = new LPage($project->projectid, $imagefile, $state, 0);
+
+    $lpage->checkout($pguser);
+
+    $err = null;
+    return $lpage;
+}
+
+function get_available_proof_page_array($project, $round, $pguser)
+// Returns an array [image, state] unless no page is available,
+// in which case throw an exception.
+{
+    global $preceding_proofer_restriction;
+
+    validate_user_can_get_pages_in_project($pguser, $project, $round);
 
     // If the global setting is to not serve the immediately preceding page,
     // first see if we can serve them a page they haven't at all seen before,
@@ -157,18 +171,10 @@ function get_available_page($projectid, $proj_state, $pguser, &$err)
         $page_array = get_available_page_array($project, $round, $pguser, $preceding_proofer_restriction);
     }
     if (!$page_array) {
-        $err = _("No more files available for proofreading for this round of the project.");
-        return null;
+        throw new ProjectNoMorePagesException(_("The project has no more pages available for proofreading in this round."));
     }
 
-    [$imagefile, $state] = $page_array;
-
-    $lpage = new LPage($projectid, $imagefile, $state, 0);
-
-    $lpage->checkout($pguser);
-
-    $err = null;
-    return $lpage;
+    return $page_array;
 }
 
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
--- a/pinc/ProjectTransition.inc
+++ b/pinc/ProjectTransition.inc
@@ -163,7 +163,7 @@ class ProjectTransition
 
     // ---------------------------------------------------------------
 
-    public function do_state_change($project, $who, $extras)
+    public function do_state_change($project, $who, $extras, $quiet = false)
     // Apply the transition to the given project,
     // and perform any attendant processing.
     // If there are any problems, return a string containing an error message.
@@ -176,7 +176,7 @@ class ProjectTransition
     // It should perhaps also be the only place that modifies these columns:
     //     modifieddate checkedoutby postproofer postcomments
     {
-        global $testing;
+        $verbose = $quiet ? false : $GLOBALS["testing"];
 
         // make sure this transition is valid
         assert($this->is_valid_for($project, $who));
@@ -252,7 +252,7 @@ class ProjectTransition
         }
         // finished expanding the template, so we can go ahead and update the table
 
-        if ($testing) {
+        if ($verbose) {
             echo "    settings: $settings<br>\n";
         }
 
@@ -314,7 +314,7 @@ class ProjectTransition
                 get_forum_id_for_project_state($this->to_state));
         }
 
-        if ($this->collateral_actions) {
+        if ($this->collateral_actions && !$quiet) {
             // Because of the "UPDATE projects" command, $project is now out of date.
             // *** but maybe not, see above "$project = new Project($projectid);"
             // So get an up-to-date version of the project info (in
--- a/pinc/ProjectTransition.inc
+++ b/pinc/ProjectTransition.inc
@@ -163,7 +163,7 @@ class ProjectTransition
 
     // ---------------------------------------------------------------
 
-    public function do_state_change($project, $who, $extras, $quiet = false)
+    public function do_state_change($project, $who, $extras)
     // Apply the transition to the given project,
     // and perform any attendant processing.
     // If there are any problems, return a string containing an error message.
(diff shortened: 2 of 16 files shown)


70ray:
(This comment copied from previous version and revised)

These changes should not have any effect on the function of the code but are intended to enable some functions to be used in the proofreading API.
The functions which it impacts are:

validate_can_be_proofed_by_current_user() is used in two places:
in project.php it gives the text in the grey boxes in the project page if the user cannot work on the project for the following reasons:

    The project is not in any round -- no text box shown in this case.
    Project has not been converted to UTF-8 yet.
    The project is in a round, but isn't available.
    User is not logged in.
    The user is not qualified for the round the project is in (possibly not the same as the round the user thinks he is working in?)

If the user can get to proof.php then if the user cannot work on the page for these reasons it goes to the abort page. In this case it gives a message if the project is not in a round.

`validate_user_can_get_pages_in_project` is also used in two places:
It also gives text in the grey boxes in the project page if the user cannot work on the project as follows:

    You have reached your quota of pages from 'Beginners Only' projects in this round.
    You have reached your quota of pages from this 'Beginners Only' project.
    You do not have access to difficulty='beginner' (Mentors Only) projects in this round.
    This project is currently reserved for new proofreaders.

If you get to proof_frame.php via the Save as done and proofread next page route, these do not cause an abort but it should go to a new page offering options according to the circumstances of the user.

Similarly for the case that there are no more pages available for proofreading.

If it is possible to get to proof_frame.php with an invalid page state or a user who does not own the page (which would throw a ProjectPageException in the _Page_require() function in $lpage->checkout) then it should go to the abort page as previously.

70ray:
Rebased onto master in case any conflicts had arisen (none had).

pinc/ProjectTransition.inc
```diff
@@ -314,7 +314,7 @@ class ProjectTransition
                 get_forum_id_for_project_state($this->to_state));
         }
 
-        if ($this->collateral_actions) {
+        if ($this->collateral_actions && !$quiet) {
```
cpeel, line 317: The collateral action might very well be something that is more than just echo'ing something to the screen. If you want to just ensure that nothing from the collateral function is printed to the screen, you could use `ob_start()` / `ob_end_clean()` to wrap the function to discard anything it would otherwise output.

cpeel (commented):
Except for the comment about the collateral and quiet mode, this looks good to me. I'd appreciate another once-over testing from @srjfoo before merging though.

srjfoo:
> * The project is not in any round -- no text box shown in this case. 
> * Project has not been converted to UTF-8 yet.
> * The project is in a round, but isn't available.
> * User is not logged in.
> * The user is not qualified for the round the project is in (possibly not the same as the round the user thinks he is working in?)
> 
> 
> If the user can get to proof.php then if the user cannot work on the page for these reasons it goes to the abort page. In this case it gives a message if the project is not in a round.
> 
> `validate_user_can_get_pages_in_project` is also used in two places: It also gives text in the grey boxes in the project page if the user cannot work on the project as follows:
> * You have reached your quota of pages from 'Beginners Only' projects in this round.
> * You have reached your quota of pages from this 'Beginners Only' project.
> * You do not have access to difficulty='beginner' (Mentors Only) projects in this round.
> * This project is currently reserved for new proofreaders.

Apologies for being so slow to get this reviewed. @70ray, does raising an exception for reaching the daily_page_limit belong in one of these categories? (Since I've been working on that for real, it was one of the first things I thought to test, and since I don't see it on your list, figured I ought to ask. (btw, I set the the daily page limit to 10 pages in P1 and P2 in this sandbox, to make for easier testing.)

70ray:
> ... @70ray, does raising an exception for reaching the daily_page_limit belong in one of these categories? (Since I've been working on that for real, it was one of the first things I thought to test, and since I don't see it on your list, figured I ought to ask. (btw, I set the the daily page limit to 10 pages in P1 and P2 in this sandbox, to make for easier testing.)

I haven't raised an exception for this. This is a bit more complicated than the other cases because the existing code gives a warning when the limit is reached and when it is passed it saves the page as in progress instead and informs the user of the situation. It might be easier to treat these cases as normal outcomes and return data which allows for these situations. 
Of course the same argument could be made for other cases which can occur normally. One could say that an exception doesn't necessarily mean that something bad has happened but is just one way of dealing with a situation.

pinc/LPage.inc
```diff
@@ -155,18 +171,10 @@ function get_available_page($projectid, $proj_state, $pguser, &$err)
         $page_array = get_available_page_array($project, $round, $pguser, $preceding_proofer_restriction);
     }
     if (!$page_array) {
-        $err = _("No more files available for proofreading for this round of the project.");
-        return null;
+        throw new ProjectNoMorePagesException(_("The project has no more pages available for proofreading in this round."));
```
70ray, line 174: Should we re-word this? For the situation where there are some pages available, but not for the particular user because he proofed them in the previous round.

cpeel:
Thanks for fixing the quiet / ob_start stuff, change looks good 👍

cpeel (approved):

pinc/LPage.inc
srjfoo: Is the message specific to that situation? That is, for the situation where there are still pages, but not pages that are available to the current user? If so, then I'd change it to say "... available to you ...". But unless that's the case, I'd probably leave it.

srjfoo (commented):
Note -- I changed `$page_tally_threshold` to 10 pages in the sandbox to make it easier to test that aspect of the changes.

> `validate_can_be_proofed_by_current_user()` is used in two places: in project.php it gives the text in the grey boxes in the project page if the user cannot work on the project for the following reasons:
> 
> * The project is not in any round -- no text box shown in this case.

Just confirming, States such as SR/PP/PPV/New?
> 
> * Project has not been converted to UTF-8 yet.
> 
> * The project is in a round, but isn't available.

I had trouble with this one, and I'm not sure I found the right instance. I was able to trigger
```
Round Complete
There are no pages available for proofreading
```
If that's not what was meant here, let me know what to look at and I'll do further testing
> 
> * User is not logged in.
> 
> * The user is not qualified for the round the project is in (possibly not the same as the round the user thinks he is working in?)

Other than the items I commented on above, everything checked out okay

> `validate_user_can_get_pages_in_project` ...
> * You have reached your quota of pages from 'Beginners Only' projects in this round.
> * You have reached your quota of pages from this 'Beginners Only' project.
> * You do not have access to difficulty='beginner' (Mentors Only) projects in this round.
> * This project is currently reserved for new proofreaders.

All good

> If you get to proof_frame.php via the Save as done and proofread next page route, these do not cause an abort but it should go to a new page offering options according to the circumstances of the user.
> 
> Similarly for the case that there are no more pages available for proofreading.

These seemed to work as expected.

> 
> If it is possible to get to proof_frame.php with an invalid page state or a user who does not own the page (which would throw a ProjectPageException in the _Page_require() function in $lpage->checkout) then it should go to the abort page as previously.

I think I tried to get to invalid pages this way in some previous testing, but never did figure out how to. I'm game to test it some more, though, if anyone else can think of a way! 😁 

Let me know if this covers it all, or if I need to do more testing. 

Regarding `daily_page_limit`, given your comment, I can see it doesn't belong in this PR, but it does need to be taken into consideration.

pinc/LPage.inc
70ray: It is the same message for all cases so I'll leave it.

70ray:
> Note -- I changed `$page_tally_threshold` to 10 pages in the sandbox to make it easier to test that aspect of the changes.
> 
> > `validate_can_be_proofed_by_current_user()` is used in two places: in project.php it gives the text in the grey boxes in the project page if the user cannot work on the project for the following reasons:
> > 
> > * The project is not in any round -- no text box shown in this case.
> 
> Just confirming, States such as SR/PP/PPV/New?
> 
Yes it seems like the states in project search form that do not have a round.
> > * Project has not been converted to UTF-8 yet.
It might be difficult to test this
> > * The project is in a round, but isn't available.
> 
> I had trouble with this one, and I'm not sure I found the right instance. I was able to trigger
> 
> ```
> Round Complete
> There are no pages available for proofreading
> ```
> 
> If that's not what was meant here, let me know what to look at and I'll do further testing
> 
> > * User is not logged in.
> > * The user is not qualified for the round the project is in (possibly not the same as the round the user thinks he is working in?)
> 
> Other than the items I commented on above, everything checked out okay
> 
> > `validate_user_can_get_pages_in_project` ...
> > 
> > * You have reached your quota of pages from 'Beginners Only' projects in this round.
> > * You have reached your quota of pages from this 'Beginners Only' project.
> > * You do not have access to difficulty='beginner' (Mentors Only) projects in this round.
> > * This project is currently reserved for new proofreaders.
> 
> All good
> 
> > If you get to proof_frame.php via the Save as done and proofread next page route, these do not cause an abort but it should go to a new page offering options according to the circumstances of the user.
> > Similarly for the case that there are no more pages available for proofreading.
> 
> These seemed to work as expected.
> 
> > If it is possible to get to proof_frame.php with an invalid page state or a user who does not own the page (which would throw a ProjectPageException in the _Page_require() function in $lpage->checkout) then it should go to the abort page as previously.
> 
> I think I tried to get to invalid pages this way in some previous testing, but never did figure out how to. I'm game to test it some more, though, if anyone else can think of a way! 😁
> 
> Let me know if this covers it all, or if I need to do more testing.
> 
> Regarding `daily_page_limit`, given your comment, I can see it doesn't belong in this PR, but it does need to be taken into consideration.

srjfoo:
> > * Project has not been converted to UTF-8 yet.
> 
> It might be difficult to test this

Actually, not so much. We've had other instances where we needed to test that, and have [Nothing & co {not UTF-8}](https://www.pgdp.org/c/project.php?id=projectID459698b8b63ac) -- whose purpose is to provide us with a non-UTF-8 project when needed.

srjfoo:
> > * Project has not been converted to UTF-8 yet.
> 
> It might be difficult to test this

Actually, not so much. We've had other instances where we needed to test that, and have [Nothing & co {not UTF-8}](https://www.pgdp.org/c/project.php?id=projectID459698b8b63ac) -- whose purpose is to provide us with a non-UTF-8 project when needed.

---

> > > * The project is in a round, but isn't available.
> 
> I think this means any state in a round except available

I tried an assortment of project states: Unavailable, Bad, Waiting, in multiple; the message for all of them is `Users are not allowed to work on the project in its current state.` Just making sure that's what was meant.

70ray:
> > > * Project has not been converted to UTF-8 yet.
> > 
> > 
> > It might be difficult to test this
> 
> Actually, not so much. We've had other instances where we needed to test that, and have [Nothing & co {not UTF-8}](https://www.pgdp.org/c/project.php?id=projectID459698b8b63ac) -- whose purpose is to provide us with a non-UTF-8 project when needed.

Thanks, I tested it also now. It gives the right error messages.

srjfoo (approved):

cpeel:
Thank you for working on this Ray. I appreciate the work you put into it and your patience as we worked to get it into the code.

PR #933, opened 2023-08-14
Widget 1

70ray:
This merges two subroutines into the functions that call them, hopefully simplifying the code.
It should not make any functional changes.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/widget_1

### DIFF
--- a/scripts/control_bar.js
+++ b/scripts/control_bar.js
@@ -261,7 +261,8 @@ function makeImageWidget(container, align = "C") {
         C: "center",
         R: "right"
     };
-    const content = $("<div>").css("text-align", alignment[align]);
+    const content = document.createElement("div");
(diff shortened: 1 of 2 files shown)


chrismiceli (approved):
Code changes look reasonable to me.

srjfoo:
Bearing in mind that JS is pretty opaque to me in parts, I've a couple of questions ...

Is this a case where merging two separate functions (if there's a difference between a subroutine and a function, I think I need it to be explained to me; I thought they were just different names for the same thing) will make things work better? Or easier to read?

You don't mention what to test -- I tested in the page browser, in the WC context area and the image pane in the proofreading interface. Is there anything else I should be looking at?

cpeel (approved):

PR #936, opened 2023-08-22
Basic latin character suite modifications

srjfoo:
The first commit adds `Ẽ`, `ẽ`, `Ĩ`, `ĩ`, `Ũ`, and `ũ` to Basic Latin, and the appropriate picker set, completing the set of vowels with tilde above.

The second commit removes the superscript numbers, fractions, ordinals and standalone diacritical marks from the picker sets. This is a removal from the picker sets, only; the characters will still be valid for Basic Latin.

Compare [the new version in my sandbox](https://www.pgdp.org/~srjfoo/c.branch/basic-latin-mods/tools/charsuites.php?charsuite=basic-latin ) with [the currently deployed version](https://www.pgdp.net/c/tools/charsuites.php?charsuite=basic-latin).

### DIFF
--- a/pinc/charsuite-basic-latin.inc
+++ b/pinc/charsuite-basic-latin.inc
@@ -12,22 +12,22 @@ $charsuite->codepoints = [
     'U+00a1-U+00ac',
     'U+00ae-U+00ff',
     // codepoints that map to Windows-1252 + extras
-    'U+0128', // [~I] (not in Windows-1252)
-    'U+0129', // [~i] (not in Windows-1252)
-    'U+0152', // OE ligature
-    'U+0153', // oe ligature
-    'U+0160', // [vS]
-    'U+0161', // [vs]
-    'U+0168', // [~U] (not in Windows-1252)
-    'U+0169', // [~u] (not in Windows-1252)
-    'U+017d', // [vZ]
-    'U+017e', // [vz]
-    'U+0178', // [:Y]
-    'U+0192', // latin small letter f with hook florin
-    'U+1ebc', // [~E] (not in Windows-1252)
-    'U+1ebd', // [~e] (not in Windows-1252)
-    'U+2039', // open single guillemet
-    'U+203a', // close single guillemet
+    'U+0128', // Ĩ - I with tilde (not in Windows-1252)
+    'U+0129', // ĩ - i with tilde (not in Windows-1252)
+    'U+0152', // Œ - OE ligature
+    'U+0153', // œ - oe ligature
+    'U+0160', // Š - S with caron
+    'U+0161', // š - s with caron
+    'U+0168', // Ũ - U with tilde (not in Windows-1252)
+    'U+0169', // Ũ - u with tilde (not in Windows-1252)
+    'U+0178', // Ÿ - Y with diaeresis
+    'U+017d', // Ž - Z with caron
+    'U+017e', // ž - z with caron
+    'U+0192', // ƒ - latin small letter f with hook florin
+    'U+1ebc', // Ẽ - E with tilde (not in Windows-1252)
+    'U+1ebd', // ẽ - e with tilde (not in Windows-1252)
+    'U+2039', // ‹ - single left-pointing guillemet
+    'U+203a', // › - single right-pointing guillemet
 ];
 $charsuite->reference_urls = [
     "https://en.wikipedia.org/wiki/Basic_Latin_(Unicode_block)",
--- a/pinc/charsuite-basic-latin.inc
+++ b/pinc/charsuite-basic-latin.inc
@@ -11,15 +11,21 @@ $charsuite->codepoints = [
     // without the soft hyphen
     'U+00a1-U+00ac',
     'U+00ae-U+00ff',
-    // codepoints that map to Windows-1252
+    // codepoints that map to Windows-1252 + extras
+    'U+0128', // [~I] (not in Windows-1252)
+    'U+0129', // [~i] (not in Windows-1252)
     'U+0152', // OE ligature
     'U+0153', // oe ligature
     'U+0160', // [vS]
     'U+0161', // [vs]
+    'U+0168', // [~U] (not in Windows-1252)
+    'U+0169', // [~u] (not in Windows-1252)
     'U+017d', // [vZ]
     'U+017e', // [vz]
     'U+0178', // [:Y]
     'U+0192', // latin small letter f with hook florin
+    'U+1ebc', // [~E] (not in Windows-1252)
+    'U+1ebd', // [~e] (not in Windows-1252)
     'U+2039', // open single guillemet
     'U+203a', // close single guillemet
 ];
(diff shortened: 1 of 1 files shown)


pinc/charsuite-basic-latin.inc
```diff
@@ -11,15 +11,21 @@ $charsuite->codepoints = [
     // without the soft hyphen
     'U+00a1-U+00ac',
     'U+00ae-U+00ff',
-    // codepoints that map to Windows-1252
-    'U+0152', // OE ligature
+    // codepoints that map to Windows-1252 + extras
+    'U+0128', // [~I] (not in CP-1252)
```
chrismiceli, line 15: does php support unicode? We could put the actual characters in the comment too if so.

chrismiceli (approved):

pinc/charsuite-basic-latin.inc
srjfoo: I'm not sure if this answers your question or not, but from the PHP 8.2 manual on strings -- 
>A string is a series of characters, where a character is the same as a byte. This means that PHP only supports a 256-character set, and hence does not offer native Unicode support.

chrismiceli: I meant in the .php file, not in a string the php language. I wonder if that is more os/git/php interpreter specific. Anyways, it is a minor idea, not even a comment :).

cpeel: Like all things PHP, this is complicated. By default, PHP source files are assumed to be ISO-8859-1 unless [zend.script-encoding](https://www.php.net/manual/en/ini.core.php#ini.zend.script-encoding) is configured on the server or unless [declare(encoding=)](https://www.php.net/manual/en/control-structures.declare.php#control-structures.declare.encoding) is set in the file. We do neither, so the PHP interpreter assumes that all of our scripts are ISO-8859-1 and we can't use anything outside of that for function names, variables, or other "code" structures.

_However_, as Sharon found, PHP just treats strings as a series of bytes, so we could use a literal UTF-8 string inside some PHP string inside a source file and it would happily just treat it as a set of bytes and leave it up to the code as to what to do with it (output it, process it, etc).

During our conversion to UTF-8 we updated all of the dproofreader source files to be explicitly UTF-8 and even have a check as part of our linting that ensures files stay that way (see https://github.com/DistributedProofreaders/dproofreaders/commit/505f00d3fcae558cea6a7f29265a7eec51937254). So we _can_ use UTF-8 characters in source files for strings and comments safely if we want.

pinc/charsuite-basic-latin.inc
```diff
@@ -11,15 +11,21 @@ $charsuite->codepoints = [
...
+    'U+0129', // [~i] (not in CP-1252)
+    'U+0152', // OE ligature~
```
cpeel, line 17: Errant `~` at the end of the comment?

pinc/charsuite-basic-latin.inc
cpeel, line 15: I'd like for use to use either `CP-1252` or `Windows-1252` in this file's comments, but not both. The former is the more common so my preference would be to use that instead.

```
cpeel@ip-172-31-12-95:~/u-cpeel-dp$ grep -R Windows-1252 * | grep -v vendor | wc -l
7
cpeel@ip-172-31-12-95:~/u-cpeel-dp$ grep -R CP-1252 * | grep -v vendor | wc -l
2
```

srjfoo: > does php support unicode? We could put the actual characters in the comment too if so.

I'm perfectly happy to add another commit to update those characters -- it would likely make it easier to read the comments.

pinc/charsuite-basic-latin.inc
srjfoo: Fixed, thanks -- no clue how that got in there.

pinc/charsuite-basic-latin.inc
srjfoo: Done. I also harmonized the comments in general.

srjfoo: Characters added to comments and comments harmonized.

cpeel:
I think the answer to this question is "no" (or at least "not in this PR and maybe not ever") but I thought it worth asking: Should we include Č č to the character suite and the set Č č Š š Ž ž to a pickerset to round out support for Slovenian?

cpeel (approved):

srjfoo:
> I think the answer to this question is "no" (or at least "not in this PR and maybe not ever") but I thought it worth asking: Should we include Č č to the character suite and the set Č č Š š Ž ž to a pickerset to round out support for Slovenian?

See [Extended European Latin B](https://www.pgdp.net/c/tools/charsuites.php?charsuite=extended-european-latin-b) -- I think we can safely leave them out of the Basic Latin picker sets, and let PMs use EEL-B. I think they were added to Basic Latin from the start, because they're in the collection of extras from Windows-1252, but they were never included in picker sets. (Mind you, except for Greek, I imagine some PMs who work in non-English languages would be happy for everything they need to be included in Basic Latin so they don't have to worry about it. 😁 )

cpeel:
> See [Extended European Latin B](https://www.pgdp.net/c/tools/charsuites.php?charsuite=extended-european-latin-b) 

Ah, yup - totally makes sense. 👍

pinc/charsuite-basic-latin.inc
```diff
@@ -11,17 +11,23 @@ $charsuite->codepoints = [
...
+    'U+0161', // š - s with caron
+    'U+0168', // Ũ - U with tilde (not in Windows-1252)
+    'U+0169', // Ũ - u with tilde (not in Windows-1252)
+    'U+0178', // Ÿ - Y with diaeresis
+    'U+017d', // Ž - Z with caron
+    'U+017e', // ž - z with caron
+    'U+0192', // ƒ - latin small letter f with hook florin
+    'U+1ebc', // Ẽ - E with tilde (not in Windows-1252)
+    'U+1ebd', // ẽ - e with tilde (not in Windows-1252)
+    'U+2039', // ‹ - open single guillemet
```
70ray, line 29: According to wikipedia some languages use these the other way round for opening and closing. Should we name them as "SINGLE LEFT-POINTING ANGLE QUOTATION MARK" etc. instead?

70ray (commented):
Comment on naming guillemets. Perhaps not relevant since it is not something this request changes.

pinc/charsuite-basic-latin.inc
srjfoo: I'm perfectly happy to update those comments -- I kind of abbreviated the others, except for the `ƒ`, with something that was close, but not quite as long as the full Unicode description. Should I go for the full long version, or summarize them as "single left(/right)-pointing guillemet"?

cpeel: I think an updated summary is good enough.

srjfoo: Done.

70ray (approved):

windymilla (approved):

PR #937, opened 2023-08-25
Update DB schema tooling

cpeel:
The `dump_db_schema` script is used in concert with `check_db_schema` during a code release to ensure that the prior release + new upgrade scripts = what's currently in the DB. The order mysqldump outputs tables is inconsistent and there's no guarantee that the tables will be output in an order that allows them to be created due to foreign key constraints.

@chrismiceli addressed this for the `users` table in a54b13b266317a477c0e15f78f4b590719811a30 for initial creation, but we still have that problem during the pre-release check. This PR rolls back Chris' commit and uses the `FOREIGN_KEY_CHECKS` variable to control how the checks are managed. This is another belt-and-suspenders approach where we make the changes in `db_schema.sql` (in case someone wants to just load it directly as it _is_ valid SQL), `install_db.php` which is how users are directed to load the schema, and in `dump_db_schema.template` so that it outputs SQL that can actually be loaded.

This is all non-live code and I have tested it all locally as part of my pre-release work. I'm looking for "does this pass the sniff test" review.

### DIFF
--- a/SETUP/dump_db_schema.template
+++ b/SETUP/dump_db_schema.template
@@ -96,6 +96,12 @@ get_table_names()
 tables=`get_table_names`
 
 if true; then
+    # The order tables are dumped is not consistent and it's possible to have
+    # tables with foreign keys before the foreign key tables are created.
+    # To make sure the dump is loadable, we have to temporarily disable
+    # foreign key checks for this session.
+    echo "SET FOREIGN_KEY_CHECKS=0;"
+
     mysqldump --user=$db_user --password="$db_password" \
         --no-data --quote-names --add-drop-table=FALSE --force \
         $db_name $tables |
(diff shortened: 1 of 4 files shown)


SETUP/dump_db_schema.template
```diff
@@ -96,6 +96,12 @@ get_table_names()
 tables=`get_table_names`
 
 if true; then
+    # The order tables are dumped is not consistent and its possible to have
```
srjfoo, line 99: (picky) -- "`it's` possible"

srjfoo (approved):

chrismiceli (approved):

SETUP/dump_db_schema.template
cpeel: Thanks - fixed!

PR #938, opened 2023-08-25
Update the CHANGELOG for the upcoming R202309 release

cpeel:
Update the `CHANGELOG.md` file for the upcoming R202309 release happening next month. I'll hold off getting this in until just before the release but wanted to get the doc updates out for review.

A reminder that we only call out the "highlights" in the changelog and rely on the git history for the details. That said, I'm open to suggestions on if this hits the right balance or if things should be added or removed!

### DIFF
--- a/SETUP/CHANGELOG.md
+++ b/SETUP/CHANGELOG.md
@@ -7,11 +7,25 @@ see the git history.
 [R202009](https://github.com/DistributedProofreaders/dproofreaders/releases/tag/R202009)
 first before upgrading to R202102 or later releases.**
 
-## R......
+## R202309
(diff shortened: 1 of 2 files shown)


windymilla (approved):

chrismiceli (approved):
I'm not even sure the jQuery line is necessary, but I'm not opposed to it being there.

srjfoo (approved):
Looks good to me. The only question I have is whether the BBcode diffs for mentors is significant enough for a mention.

cpeel:
> The only question I have is whether the BBcode diffs for mentors is significant enough for a mention.

I don't think so, because while it was a really useful change, it is a mentors-only feature which seems narrow enough not to include in the release notes.

PR #940, opened 2023-09-18
Rev Google Font URLs

cpeel:
It's that time for another Google font URL update. Last one was in Apr 2022 which isn't so bad.

Sandbox at https://www.pgdp.org/~cpeel/c.branch/rev-google-font-urls/

### DIFF
--- a/styles/themes/charcoal.css
+++ b/styles/themes/charcoal.css
@@ -146,7 +146,7 @@
   font-style: normal;
   font-weight: 700;
   font-display: swap;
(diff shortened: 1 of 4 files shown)


srjfoo (approved):

chrismiceli (approved):

70ray (commented):
I don't know if this is relevant but when I pushed a recent change to raise_exceptions_1 it failed because of  a problem with these files and I had to rerun "less" to fix it and the numbers changed to v27.

The same thing has happened on #941

cpeel:
> I don't know if this is relevant but when I pushed a recent change to raise_exceptions_1 it failed because of a problem with these files and I had to rerun "less" to fix it and the numbers changed to v27.

It's the same problem. We'll probably have to rebase your branch to drop that commit after this merges in to avoid a merge conflict.

PR #941, opened 2023-09-19
Remove cookie-based DP session management

cpeel:
Remove the cookie-based DP session management code. This PR effectively moves the contents of `pinc/dpsession_via_php_sessions.inc` into `pinc/dpsession.inc` with a small bit of cleanup.

The cookie-based session code has been completely untested for probably 2 decades as pgdp.net has used PHP sessions since before 2004.

Testable in https://www.pgdp.org/~cpeel/c.branch/remove-dp-sessions-via-cookies/

Testing hints: if logging in, staying logged in, and being able to log out and stay logged out all work, this is good to go.


srjfoo (approved):

chrismiceli (approved):
I love it

PR #943, opened 2023-09-20
Upgrade Github action versions

cpeel:
Upgrade the versions of `actions/checkout` and `actions/setup-node`. This addresses a deprecation warning we're seeing in the CI output about node12 (which is used in the old actions) being deprecated. This is only a CI-related change and does not impact the DP code at all. If the tests pass, this should be good to go.


chrismiceli (approved):
we could probably go to node 18 too, the current LTS. Let me see if I can get that review out eventually.

windymilla (approved):

srjfoo (approved):

PR #944, opened 2023-09-21
Move PM view preferences to user_settings table

cpeel:
The `users.i_pmdefault` setting is only applicable to PMs so move the setting to the `user_settings` table rather than in the table for all users. This follows the pattern we user for other PM-specific settings and further simplifies the user table to be applicable to all users.

Also remove the unused and unreferenced `users.i_perfs` column. The only thing I think this is currently tracking is if the user has saved their preferences at any point and that doesn't seem useful.

Testable in the https://www.pgdp.org/~cpeel/c.branch/move-pm-pref-to-usersettings/ sandbox.

_Note that the new setting has been populated on the TEST system but the column has not yet been dropped from the `users` table. After this PR is approved I'll confirm than the update script works in-full rather than just the data population half._

### DIFF
--- a/tools/project_manager/projectmgr.php
+++ b/tools/project_manager/projectmgr.php
@@ -16,16 +16,8 @@ include_once('projectmgr.inc'); // echo_manager_links();
 require_login();
 
 $user = User::load_current();
-switch ($user->i_pmdefault) {
-    case 0:
-        $default_view = "user_all";
-        break;
-    case 1:
-        $default_view = "user_active";
-        break;
-    default:
-        $default_view = "blank";
-}
+$user_settings = Settings::get_Settings($user->username);
+$default_view = $user_settings->get_value("pm_view", "user_all");
 
 try {
     $show_view = get_enumerated_param($_GET, 'show', $default_view,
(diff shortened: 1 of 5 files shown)


srjfoo (approved):

windymilla (approved):

tools/project_manager/projectmgr.php
```diff
@@ -16,16 +16,8 @@
...
-        $default_view = "user_all";
-        break;
-    case 1:
-        $default_view = "user_active";
-        break;
-    default:
-        $default_view = "blank";
-}
+$user_settings = Settings::get_Settings($user->username);
+$default_view = $user_settings->get_value("pm_view", "user_all");
```
chrismiceli, line 20: does this change the default from "blank" to "user_all"? Not that I expect the db value to be unset or outside the range of [0, 1, 2]

chrismiceli (approved):

tools/project_manager/projectmgr.php
cpeel: It might seem like it changes the default, but the fall-through was never actually used because the value is always set by a `DEFAULT` value in the database:
```
`i_pmdefault` smallint(1) NOT NULL default '0',
```

And `0` is `user_all` which is the new default if no value is set (such as a new PM that hasn't specified a value in their preferences).

chrismiceli: Makes perfect sense.

70ray (approved):

cpeel:
Both upgrade scripts have been run on TEST to populate the new settings and drop the two columns.

PR #947, opened 2023-09-23
Update to node 18 and latest dependencies

chrismiceli:
This should only impact tests. I confirmed the less utility we use for the css generation produces the same results as well. No need for a sandbox unless we think it is necessary.


srjfoo (approved):

cpeel (approved):
Thanks for these updates, Chris!

70ray (approved):

PR #949, opened 2023-09-28
Highlight non-Latin unicode scripts in WC

cpeel:
To improve proofreaders ability to differentiate look-alike characters, highlight non-Latin characters in WordCheck like we do for punctuation. This removes the current behavior where words containing multiple Unicode scripts are flagged as bad in lieu of highlighting the characters instead.

Testable in the [highlight-non-latin-scripts](https://www.pgdp.org/~cpeel/c.branch/highlight-non-latin-scripts/) sandbox -- in particular the [UTF-8 Practice: Dogs of all nations](https://www.pgdp.org/~cpeel/c.branch/highlight-non-latin-scripts/project.php?id=projectID5e20bc6bd5786&expected_state=P1.proj_avail) project which has some additional character suites and custom characters.

### DIFF
--- a/tools/proofers/spellcheck_text.inc
+++ b/tools/proofers/spellcheck_text.inc
@@ -51,6 +51,13 @@ function spellcheck_text($orig_text, $projectid, $imagefile, $aux_language, $acc
 
     [$uncommonScriptWords, $uncommonScripts] =
         get_words_with_uncommon_scripts(array_keys(get_distinct_words_in_text($orig_text)));
+
+    // If Greek is present in $uncommonScripts, we always want it to be first
+    // so it always has the same color across pages and projects. We do this
+    // only for Greek as it is the most common non-Latin script at pgdp.net.
+    if (in_array("Greek", $uncommonScripts)) {
+        $uncommonScripts = array_merge(["Greek"], array_diff($uncommonScripts, ["Greek"]));
+    }
     $uncommonScriptMap = array_flip($uncommonScripts);
 
     // Prepare the style block which requires some dynamic content from
--- a/tools/proofers/spellcheck_text.inc
+++ b/tools/proofers/spellcheck_text.inc
@@ -66,9 +66,10 @@ function spellcheck_text($orig_text, $projectid, $imagefile, $aux_language, $acc
                    "  pre { line-height: 2.7; }".
                    "  .proofingfont { font-family: $font_family; $font_size }" .
                    "  .hl-punc { background-color: yellow; color: black; }" .
-                   "  .hl-script-0 { background-color: lightgreen; color: black; }" .
-                   "  .hl-script-1 { background-color: aqua; color: black; }" .
-                   "  .hl-script-2 { background-color: mistyrose; color: black; }" .
+                   "  .hl-script-0 { background-color: CadetBlue; color: black; }" .
+                   "  .hl-script-1 { background-color: lightpink; color: black; }" .
+                   "  .hl-script-2 { background-color: lightgreen; color: black; }" .
+                   "  .hl-script-3 { background-color: mediumpurple; color: black; }" .
                    "  img.aw { border: 0; margin-left: 5px; }" .
                    "  span.aw { background-color: white; color: black; }" .
                    "</style>";
--- a/tools/proofers/spellcheck_text.inc
+++ b/tools/proofers/spellcheck_text.inc
@@ -53,23 +53,27 @@ function spellcheck_text($orig_text, $projectid, $imagefile, $aux_language, $acc
         get_words_with_uncommon_scripts(array_keys(get_distinct_words_in_text($orig_text)));
     $uncommonScriptMap = array_flip($uncommonScripts);
 
-    // ok, at this point we have a finalized list of bad words.
-    // start preparing the page
+    // Prepare the style block which requires some dynamic content from
+    // the user's preferences and the page itself.
     [$font_style, $font_size, $font_family] = get_user_proofreading_font();
     if ($font_size != '') {
         $font_size = "font-size: $font_size";
     } else {
         $font_size = "font-size: unset";
     }
-    // define the styles used for the interface (highlight for the punctuation, AW button, etc)
+
+    $highlight_colors = ["lightpink", "lightgreen", "CadetBlue", "mediumpurple"];
+    $highlight_styles = "  .hl-punc { background-color: yellow; color: black; }\n";
+    for ($index = 0; $index < count($uncommonScripts); $index++) {
+        $color_index = min($index, count($highlight_colors) - 1);
+        $color = $highlight_colors[$color_index];
+        $highlight_styles .= "  .hl-script-{$index} { background-color: $color; color: black; }\n";
+    }
+
     $returnString .= "<style>" .
                    "  pre { line-height: 2.7; }".
                    "  .proofingfont { font-family: $font_family; $font_size }" .
-                   "  .hl-punc { background-color: yellow; color: black; }" .
-                   "  .hl-script-0 { background-color: CadetBlue; color: black; }" .
-                   "  .hl-script-1 { background-color: lightpink; color: black; }" .
-                   "  .hl-script-2 { background-color: lightgreen; color: black; }" .
-                   "  .hl-script-3 { background-color: mediumpurple; color: black; }" .
+                   $highlight_styles .
                    "  img.aw { border: 0; margin-left: 5px; }" .
                    "  span.aw { background-color: white; color: black; }" .
                    "</style>";
--- a/tools/proofers/spellcheck_text.inc
+++ b/tools/proofers/spellcheck_text.inc
@@ -24,7 +24,6 @@ use voku\helper\UTF8;
 function spellcheck_text($orig_text, $projectid, $imagefile, $aux_language, $accepted_words)
 {
     global $puncCharacters;
-    global $code_url;
 
     // variable holding final string
     $returnString = "";
@@ -50,6 +49,10 @@ function spellcheck_text($orig_text, $projectid, $imagefile, $aux_language, $acc
     [$badWordHash, $languages, $messages] =
         get_bad_word_levels_for_project_text($orig_text, $projectid, $aux_language, $accepted_words);
 
+    [$uncommonScriptWords, $uncommonScripts] =
+        get_words_with_uncommon_scripts(array_keys(get_distinct_words_in_text($orig_text)));
+    $uncommonScriptMap = array_flip($uncommonScripts);
+
     // ok, at this point we have a finalized list of bad words.
     // start preparing the page
     [$font_style, $font_size, $font_family] = get_user_proofreading_font();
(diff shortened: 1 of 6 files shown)


tools/proofers/spellcheck_text.inc
```diff
@@ -125,54 +132,26 @@ function spellcheck_text($orig_text, $projectid, $imagefile, $aux_language, $acc
...
             $wordLen = UTF8::strlen($word);
 
-            // see if we are punctuation
-            if (in_array($word, $puncArray)) {
-                $newLine = UTF8::substr_replace($newLine, _wrapPunc($word), $lineIndex, $wordLen);
-            } elseif (in_array($word, $accepted_words)) {
-                // see if we're an AW word
-                $newLine = UTF8::substr_replace($newLine, _wrapAW($word), $lineIndex, $wordLen);
-            } else {
-                // not punctuation, handle word
```
cpeel, line 135: This code was lifted into the `_wrapBadWord()` function below and largely unchanged.

tools/proofers/spellcheck_text.inc
```diff
@@ -62,17 +65,17 @@ function spellcheck_text($orig_text, $projectid, $imagefile, $aux_language, $acc
     $returnString .= "<style>" .
                    "  pre { line-height: 2.7; }".
                    "  .proofingfont { font-family: $font_family; $font_size }" .
-                   "  .hl { background-color: yellow; color: black; }" .
+                   "  .hl-punc { background-color: yellow; color: black; }" .
+                   "  .hl-script-0 { background-color: lightgreen; color: black; }" .
```
cpeel, line 69: These three new highlighting colors are random placeholders and we need to select at least 4, probably 5 or 6, highlighting colors that will work well before this merges in.

70ray: Could you instead use just one colour and add a tooltip to indicate what script the offending character belongs to?

70ray (commented):
For the case where there are many words in another script such as polytonic greek would it be possible to not highlight a word if it is entirely in the same script? Or if most of the characters in a word are in polytonic greek, to highlight the characters which are not? There could be some special cases such as a single capital A.

tools/proofers/spellcheck_text.inc
srjfoo: I don't think that would work visually if a word, for instance, included both Greek and Cyrillic characters.

cpeel:
> For the case where there are many words in another script such as polytonic greek would it be possible to not highlight a word if it is entirely in the same script? Or if most of the characters in a word are in polytonic greek, to highlight the characters which are not? There could be some special cases such as a single capital A.

These are possible but it is likely to be confusing for users because what's highlighted won't be consistent on the page and might change page-to-page depending on what is on each page.

srjfoo:
Bear in mind that this is quick and dirty, but I've done some color testing.

Color version: 
image

Greyscale version: 
image

Legend:
A — CadetBlue #5f9ea0
B — SpringGreen #00ff7f
C — lightgreen #90EE90
D — MistyRose  #FFE4E1
E — LightSkyBlue #87CEFA
F — mediumpurple #9370DB
G — darkgrey #A9A9A9
H — lightpink #FFB6C1
I — peru #CD853F

This is by no means a perfect test -- I really don't know how color blindness works. What I was doing was using a contrast checker on the web, and then seeing what combinations looked when converted to default greyscale. It might be worth saying when we put it up for testing that we would welcome input from anyone who suffers from color blindness or other vision problems affecting color perception to offer input on the colors. (Side note: we already have had one user with macular degeneration (I think) who couldn't see the yellow highlighting. I ran some tests, and adding a smidge of red to the yellow and lightening the WC background did help, but they didn't follow up, and I was afraid to push making that kind of change to WC -- and I got distracted.)

This last screenshot is a kind of mishmash. The legend is the same except that I dropped off MistyRose -- it's too close to the background color. My recommendation would be one of (ABI), H (or something with the same contrast level, if pink is too offensive 😁 ), E or C and F, to give us 4 to start with. Again, while my eyes may not be as good as they used to be, I don't have that kind of eye problems, so I'm not a good judge. If you can see the colors, there's enough difference that they're all relatively easy to distinguish (except, to me, the MistyRose).
image

srjfoo (approved):
[Special Character Test Project](https://www.pgdp.org/~cpeel/c.branch/highlight-non-latin-scripts/project.php?id=projectID5deed201ccd06) has an additional script added to the project custom characters. I added a bit of hiragana, and the Cyrillic A so that I could see how three scripts did with the highlighting. Hiragana doesn't have any lookalike characters, so probably wouldn't be a big deal being recognized regardless, though.

image

tools/proofers/spellcheck_text.inc
70ray: That's true. If a greek word contained one Cyrillic character it wouldn't be obvious anything was wrong.

70ray: Unless you used two colours and an algorithm so that adjacent characters in different scripts were in different colours. But this would mean there was no relation between the colour and the script.

windymilla:
> This last screenshot is a kind of mishmash. The legend is the same except that I dropped off MistyRose -- it's too close to the background color. My recommendation would be one of (ABI), H (or something with the same contrast level, if pink is too offensive 😁 ), E or C and F, to give us 4 to start with. Again, while my eyes may not be as good as they used to be, I don't have that kind of eye problems, so I'm not a good judge. If you can see the colors, there's enough difference that they're all relatively easy to distinguish (except, to me, the MistyRose). image
A or I, H, C, F looks good to me in color or grayscale, but I agree it would be good to get wider opinions before deciding.

cpeel:
Thanks @srjfoo! I've updated the colors to be A, C, H, and then F.

srjfoo:
Just goes to show that it's good to take a look at it at the proper scale.
image
and 
image

At the larger scale, I found the cadet blue and the medium purple quite readable. Not so at my default text size (browser default, which, I think is 16px).

Comments?

(The fourth script I added is katakana, because I wanted to see all four colors. Project where the characters can be tested: [Special Character Test Project](https://www.pgdp.org/~cpeel/c.branch/highlight-non-latin-scripts/project.php?id=projectID5deed201ccd06))

srjfoo:
Ooh -- I just discovered a tool in Photoshop I didn't know about:
View->Proof setup->Color Blindness -- Protanopia-type:
image
and
View->Proof setup->Color Blindness -- Deuteranopia-type:
image

That doesn't cover everything we might run into, but it's a couple more data points.

70ray (approved):

windymilla (approved):

srjfoo:
> I think if there are more scripts than colours, it would currently show unhighlighted 

I stumbled over this comment again just now, and have confirmed that if there's not a highlight color defined, the script isn't highlighted:
image

I guess the question is, how many scripts is any one page likely to have where there are potentially confusing letters used? Of the 5 that I put on that page, only Coptic, Cyrillic and Greek will have any chance of being confusing. Hiragana and Katakana should stand out on their own.

There is also this: if a page is that busy with multiple scripts that have been approved for the project, let's hope that it makes the proofreader very, very careful!

windymilla:
Very happy with that logic Sharon

cpeel:
@windymilla the code does work as you have described (yay for it being somewhat understandable!).

I do like the idea of having a more graceful failure mode for > the number of defined colors (currently 4, not including the one for punctuation). While it is an edgecase it's a pretty glaring one and your solution is an elegant one. Let me ponder how to make that work. It does mean we'll need one more color 😁 

Speaking of colors: are we happy with the 4 presented (and coded)?

srjfoo:
> Speaking of colors: are we happy with the 4 presented (and coded)?

I'm not completely happy, but those who have responded in the forums have not yet registered complaints. It may be that we need to just be prepared to make changes later if we find out otherwise. Most of the users probably don't even know that the test site exists.

cpeel:
I've taken a slightly different approach and just re-used the last highlight color if we have more scripts than we have defined colors. If this approach makes sense then my suggestion is that we don't bother adding an additional color and go with this. We can add & adjust colors after we get feedback from the field.

srjfoo:
> I've taken a slightly different approach and just re-used the last highlight color if we have more scripts than we have defined colors. If this approach makes sense then my suggestion is that we don't bother adding an additional color and go with this. We can add & adjust colors after we get feedback from the field.

For reference:
image

image

[Special Character Test Project](https://www.pgdp.org/~cpeel/c.branch/highlight-non-latin-scripts/project.php?id=projectID5deed201ccd06)

Bear in mind that I'm picking non-Latin scripts at random for custom chars for the above test project. I don't think limiting it to 4 colors is a bad idea, but one thing jumps out at me. Since the scripts are listed and associated with colors in alphabetical order, there's no good way to, say, associate Greek with a particular color (and, as I'm about to get started on pulling Cyrillic together, ditto that) unless those are the only two scripts being used.

Is that likely to be a problem? My feeling is, probably not, and I think we can punt that off to the future if it ever becomes one. The key thing is that we've got highlighting for non-Latin scripts, and that anything over 2 on a given page is likely to be rare (at least I hope so). If a PM starts a project that contains more than two (essentially, more than Greek and Cyrillic and whatever they can stuff into the 32 characters available for custom chars) should be aware of the potential for problems, and warning proofreaders. For a lot of scripts, the differences will be obvious for the overflow, and even if not, they're at least obviously marked as non-Latin.

All this is the long way of saying that I'm good with it as-is, unless you think we need to add another color.

cpeel:
Yes, the downside of it being entirely programmatic is that we can't actually guarantee that Greek will be highlighted in a distinct color without creating a *lot* more colors. I don't think this is likely to be an issue at roll-out as I have a hard time imagining that there will be many (any?) projects that have more than 4 possible scripts. I haven't done an exhaustive search but it looks like projects only have up to one additional script using custom characters on PROD and that coupled with our current character suites says that 4 should be enough initially.

I really suspect that after we roll this out people will Have Opinions about the colors we've chosen and we can use that opportunity to select a wider range of them which will also

windymilla:
> All this is the long way of saying that I'm good with it as-is, unless you think we need to add another color.

I think it's OK as-is too. A page can have 4 different scripts (plus Latin) and each one will have its own color. A page with 5 scripts must be extremely rare. 

> Since the scripts are listed and associated with colors in alphabetical order, there's no good way to, say, associate Greek with a particular color

I do quite like the idea of Greek (if it is on the page) always being in script-0, mostly because it is by far the most common confusable script (and the reason for the user requests). It will be the one that everyone will see (maybe the only one for lots of people) and would be nice to be consistent across pages and across projects. That said, since most pages/projects only have Greek, then Greek would be script-0 for those.

To make Greek always script 0, would it work instead of sorting the scripts alphabetically using `sort`, using `usort`, with the callback function putting "Greek" before any other script names (which would themselves continue to be alphabetically sorted). Given that the list of scripts in the legend is going to be relatively short, people will still be able to find the word "Greek" despite it being out of alphabetical order. If you think otherwise, I'm OK as it is though.

srjfoo:
> I really suspect that after we roll this out people will Have Opinions about the colors we've chosen and we can use that opportunity to select a wider range of them too.

Me too. Because though we do have a few regular testers, most people don't bother, and don't even see the announcements. So I find it difficult to get really excited until others see it.

And thanks for special-casing Greek. That looks good!

srjfoo (approved):

cpeel:
Reordered the colors around slightly so that CadetBlue is 3rd instead of 1st thereby making Greek always lightpink. Going to merge this in and we can futz with colors after we roll this out and get more feedback.

PR #953, opened 2023-10-02
Introduce JSON "forum" backend for development

cpeel:
I've been thinking about ways to improve the developer onboarding experience. I want to make it easier for developers to clone the repo and get the code working on their computers with less work. One of the hurdles right now is needing to install and configure phpBB which for most development purposes we only need for authentication.

This PR introduces a test-only JSON "forum" backend. With this developers can point the configuration to two JSON paths that will store users and posts for development and ignore phpBB altogether.

To implement this, the existing `forum_interface.inc` file was moved to `forum_interface_phpbb3.inc` and the former replaced with a conditional load based on the forum type. Sadly, this makes for a very noisy PR because it sees it as a file addition and a mostly-deleted file. This PR defines all of the forum required functions for the JSON backend, but most of the not-required ones are neutered. That said, account registration and authentication all work as do creating new forum topics.

Testing this on TEST is a little tricky because creating accounts in the JSON backend will create accounts in the main TEST database that do not have versions in phpBB. I will follow-up on Slack with the sandbox and some additional details.

I'm interested in the following feedback on this PR:
1. Does this seems like a worthwhile endeavor?
2. Does the code in `forum_interface_json.inc` looks reasonable?


70ray:
I find it quite convenient to work in a sandbox on the test server. This has some advantages if one wishes to work from different computers or change computers, but there could be some benefit in working on my own computer. I'll have a look at the code.

windymilla:
That's what I thought happened. AFK now, but will try the same again later.

windymilla:
@cpeel - you're right - it's working as you describe - if I click on "Discuss this project", it just refreshes the project page. I obviously fooled myself the previous time I did it. Sorry for the false alarm!

chrismiceli (approved):
I think this will be super useful! I imagine it could also be useful for some unit testing in some capacity.

PR #954, opened 2023-10-04
decode JSON objects as arrays for api_get_request_body()

70ray:


cpeel:
Can we get a sandbox to test this? This function is used in two other places (PUTing word lists and PUTing project holds) and it would be good to validate that those locations still work (although I expect they will).

70ray:
> Can we get a sandbox to test this? This function is used in two other places (PUTing word lists and PUTing project holds) and it would be good to validate that those locations still work (although I expect they will).

Forgot this. I've made it now: Sandbox at: https://www.pgdp.org/~rp31/c.branch/json_array

I tested it with curl in my c sandbox but it said method not allowed in the c.branch sandbox.

cpeel (approved):
Thanks for the sandbox - confirmed both of those paths also work.

PR #955, opened 2023-10-10
simplify the split direction control button

70ray:
Use just one image and change its source rather than showing and hiding alternate jquery object buttons.

Sandbox at https://www.pgdp.org/~rp31/c.branch/sim_split


chrismiceli (approved):
I love moving away from jQuery more :D.

srjfoo (approved):

cpeel (approved):

PR #958, opened 2023-10-13
Introduce code documentation standards and tooling

cpeel:
Introduce DocBlock as the preferred function code comment style. This PR includes an example of what this would look like in `activity_hub.php` and docs on how to generate the code-level docs using phpDocumentor. You can see an example of the rendered docs at https://www.pgdp.org/~cpeel/c/build/docs/files/activity-hub.html

As a bonus, many editors (VSCode, Sublime) can parse DocBlocks and provide both syntax color highlighting and function completion.

# Why

This is another case of [yak-shaving](https://en.wiktionary.org/wiki/yak_shaving#/media/File:Yak_shaving.jpg).

`php-cs-fixer` v3 has Strong Opinions on where function comments should be located in functions with multi-line parameters.

For instance, this:
```php
function example_with_multiline_params(
    $param1,
    $param2
)
// Function comment
{
}
```

Becomes this:
```php
function example_with_multiline_params(
    $param1,
    $param2
) {
    // Function comment
}
```

We do this "function comment after the signature and before the open bracket" a lot and moving the function documentation comment into the block is simply incorrect. Before I run `php-cs-fixer` v3 on all of the codebase I need to update all of the functions that follow this pattern. Since I have to touch pretty much every function anyway this seemed like a good time to settle on some code doc standards.

If this PR is approved and we agree on this format going forward I'll merge this in and follow up with another PR that updates function comments to DocBlock format with the least amount of work possible (not adding `@params` descriptions where we don't do so already, etc). That will unblock our upgrade to `php-cs-fixer` v3.

### DIFF
--- a/activity_hub.php
+++ b/activity_hub.php
@@ -140,17 +140,26 @@ activity_descriptions();
 
 // ----------------------------------
 
+/**
+ * Print out the snapshot table.
+ *
+ * The table contains a row for each stage including the user's access for
+ * that stage as well as project and page metrics.
+ *
+ * @param bool $show_filtered_projects
+ *   If TRUE, the table will show numbers based on the user's project filter
+ *   for that stage.
+ *
+ * @param bool $show_filtering_links
+ *   If TRUE, the table will allow the user to toggle between viewing numbers
+ *   for ALL projects and just those for filters.
+ *
+ * @param bool $show_beginner_help
+ *   If TRUE, the table will be prefaced with an informational paragraph.
+ *
+ * @return void
+ */
 function progress_snapshot_table($show_filtered_projects, $show_filtering_links, $show_beginner_help)
-// Prints out a table containing a row for each stage including the user's
-// access for that stage as well as project and page metrics.
-// Arguments:
-//   $show_filtered_projects - if TRUE, the table will show numbers based on
-//                             the user's project filter for that stage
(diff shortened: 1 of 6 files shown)


windymilla (approved):
LGTM - great idea!
Lucky that yaks are large animals - there's always some part that needs it when you want to do some shaving

srjfoo (approved):

70ray (approved):

activity_hub.php
```diff
@@ -140,17 +140,26 @@
...
+ * @param bool $show_filtering_links
+ *   If TRUE, the table will allow the user to toggle between viewing numbers
+ *   for ALL projects and just those for filters.
+ *
+ * @param bool $show_beginner_help
+ *   If TRUE, the table will be prefaced with an informational paragraph.
+ *
+ * @return void
+ */
 function progress_snapshot_table($show_filtered_projects, $show_filtering_links, $show_beginner_help)
```
chrismiceli, line 162: should we start using php type declarations too since we're specifying types in the phpdocs now? I realize this changes runtime behavior with TypeError, so maybe safer to not. Just an idea.

chrismiceli (approved):

activity_hub.php
cpeel: I think so, yes. We'll have to do it gradually and intentionally but we're at a point where that makes sense with our minimum PHP version, etc.

PR #963, opened 2023-10-17
Update odd formatting

cpeel:
This is another prep PR for the upcoming php-cs-fixer v3 update. The new version Has Opinions about switch statements and a few other things that I thought would be easier to handle in a separate PR.

This has 4 commits which touch different things:
* Standardizes switch statements across the codebase
* Remove errant `;`s
* Massage multiline `global` statements (v3 wants to format these oddly)
* Tweak comments (v3 wants to place these in interesting places)

The PR is best viewed with the "ignore whitespace" option.

I think a code review should be sufficient for this PR, but because it's easy to do, here's a sandbox: [update-odd-formatting](https://www.pgdp.org/~cpeel/c.branch/update-odd-formatting/)

### DIFF
--- a/quiz/generic/quiz_page.inc
+++ b/quiz/generic/quiz_page.inc
@@ -167,8 +167,6 @@ function qp_text_contains_anticipated_error($text, $test)
             }
             break;
 
-        // -----------------------------------------------------------
-
         case "markupmissing":
             // It's an error if the text does not contain both $opentext and $closetext.
             // (A correct text will contain $opentext or $closetext [or both, really].)
@@ -209,8 +207,6 @@ function qp_text_contains_anticipated_error($text, $test)
                 return $test["error"];
             }
             break;
-
-        // -----------------------------------------------------------
 
         case "expectedlinebreaks":
             // It's an error if the portion of the text between $starttext and $stoptext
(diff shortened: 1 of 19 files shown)


chrismiceli (approved):

srjfoo (approved):

70ray (approved):

quiz/generic/quiz_page.inc
```diff
@@ -344,10 +340,9 @@ function qp_echo_error_html($message_id)
     // Challenge the quiz-taker to fix the error.
     echo "<p>";
     if (isset($message["challengetext"])) {
+        // XXX: Currently unused. Might be discontinued.
```
windymilla, line 343: Should the "unused & might be discontinued" refer to the `if` or the `else`? It looks like the `else` from the original positioning of the comment, but the `if` with this edit. I'm sure you know which is correct - I don't, but it looked like a change.

cpeel: Good catch. My read is that it actually does belong with the `if`. `challengetext` is not used anywhere in the quizzes so I decided the comment must have applied to the `if`.

windymilla: Fair enough - thanks

PR #964, opened 2023-10-17
Move bulk of spellcheck.inc into a function

cpeel:
`spellcheck.inc` is one of the few remaining oddities where we `include()` this file in the middle of another file (`processtext.php`) to inject its contents. It's also got an annoying mix of PHP and straight HTML. This is step 1 of 2 to addressing those issues by moving the contents into two pure PHP functions. I intend to lift all the functions in this file into `spellcheck_text.inc` and delete the `spellcheck.inc` entirely but it's going to be far easier to PR the changes if we do this first. The move will happen in a follow-up PR after this goes in.

The impetus for this is that php-cs-fixer wants to left-justify most of this file and I decided actually addressing the problem was a better approach.

Testable in [wordcheck-cleanup-1](https://www.pgdp.org/~cpeel/c.branch/wordcheck-cleanup-1/).

### DIFF
--- a/tools/proofers/spellcheck.inc
+++ b/tools/proofers/spellcheck.inc
@@ -34,7 +34,7 @@ output_wordcheck_interface($user, $ppage, $text_data, $is_changed, $accepted_wor
 
 function output_wordcheck_interface($user, $ppage, $text_data, $is_changed, $accepted_words, $aux_language)
 {
-    global $word_check_messages;
+    global $code_url, $word_check_messages;
 
     $revert_text = $_POST['revert_text'] ?? $text_data;
 
--- a/tools/proofers/spellcheck.inc
+++ b/tools/proofers/spellcheck.inc
@@ -16,69 +16,39 @@ include_once('image_block_enh.inc');
 // * $accepted_words
 // * $aux_language
 
-$revert_text = $_POST['revert_text'] ?? $text_data;
-
-$is_header_visible = array_get($_SESSION, "is_header_visible", 1);
-
-$project = new Project($ppage->projectid());
-$valid_character_pattern = build_character_regex_filter($project->get_valid_codepoints(), "js");
-
-$keep_corrections = _("Keep corrections and return to proofreading this page");
-$rerun = _("Check page against an additional language");
-
-$word_check_messages = json_encode([
+$word_check_messages = [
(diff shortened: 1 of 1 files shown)


tools/proofers/spellcheck.inc
```diff
@@ -16,69 +16,39 @@ include_once('image_block_enh.inc');
...
+$word_check_messages = [
     "disableAwLabel" => _("Word has been edited; unable to Suggest"),
     "enableAwLabel" => _("Unflag All & Suggest Word"),
     "confirmExitString" => _("Changes have been made. OK to quit without saving?"),
     "pageChangedError" => _("Cannot save page from WordCheck. Words have been modified."),
     "badCharsError" => _("There are invalid characters in the page"),
-    "keepCorrectons" => $keep_corrections,
-    "rerun" => $rerun,
-]);
+    "keepCorrectons" => _("Keep corrections and return to proofreading this page"),
```
cpeel, line 25: I kept the original spelling since it is used in a JS file like this.

tools/proofers/spellcheck.inc
```diff
@@ -184,57 +150,42 @@ if ($user->profile->i_type == 1) {
...
-</tr>
-</table>
-</form>
-</div>
-<?php
+    echo '</td>';
+    echo '</tr>';
+
+    echo '<tr>';
+    echo '<td id="tdbottom">';
```
cpeel, line 168: The ID was changed from `tdtop` to `tdbottom` as `tdtop` already exists above. Found this after running the resulting HTML through the W3C validator to confirm I didn't screw up any of the tags.

srjfoo:
Operation looks good, but looks like the button styling is off (good thing I normally use Charcoal. 😁 

Top is in the main TEST sandbox; bottom is the PR sandbox, in Charcoal. looks as expected in the PG theme.
image

Now that I look at it more closely, the buttons look better when I switch to the default theme, but still don't match the current WC colors:
image

cpeel:
> Operation looks good, but looks like the button styling is off (good thing I normally use Charcoal. 😁

Oh, good catch - thank you! Can you try it again in the sandbox? If it works now I will push up that change to undo my `tdtop` -> `tdbottom` as mentioned above.

srjfoo (approved):
LGTM

70ray (approved):

PR #967, opened 2023-10-30
redefine task_status_array after unsetting keys

70ray:
This enables ordinary users to create tasks.

Perhaps not an entirely satisfactory solution to the problem. Probably better to avoid redefining the task_status_array somehow, but seems to fix the problem for now.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/task_fix

### DIFF
--- a/tasks.php
+++ b/tasks.php
@@ -1169,14 +1169,14 @@ function TaskForm($task)
     global $task_assignees_array;
     global $priority_array, $tasks_url;
 
-    // Don't want to permit setting status to "Closed" or "Reopened" when creating/editing a task
-    unset($tasks_status_array[get_property_key("Closed", $tasks_status_array)]);
-    unset($tasks_status_array[get_property_key("Reopened", $tasks_status_array)]);
-
     // Non-managers can only set the task status to New.
     if (!user_is_a_sitemanager() && !user_is_taskcenter_mgr()) {
         $tasks_status_array = [get_property_key("New", $tasks_status_array) => "New"];
     }
+
+    // Don't want to permit setting status to "Closed" or "Reopened" when creating/editing a task
+    unset($tasks_status_array[get_property_key("Closed", $tasks_status_array)]);
+    unset($tasks_status_array[get_property_key("Reopened", $tasks_status_array)]);
 
     $task_summary_enc = attr_safe($task->task_summary);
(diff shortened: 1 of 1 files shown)


chrismiceli (approved):

srjfoo (approved):

tasks.php
```diff
@@ -1169,15 +1169,15 @@ function TaskForm($task)
...
     global $priority_array, $tasks_url;
 
+    // Don't want to permit setting status to "Closed" or "Reopened" when creating/editing a task
+    unset($tasks_status_array[get_property_key("Closed", $tasks_status_array)]);
+    unset($tasks_status_array[get_property_key("Reopened", $tasks_status_array)]);
+
     // Non-managers can only set the task status to New.
     if (!user_is_a_sitemanager() && !user_is_taskcenter_mgr()) {
         $tasks_status_array = [get_property_key("New", $tasks_status_array) => "New"];
     }
```
cpeel, line 1179: Can you elaborate on how this change resolves the problem? This redefines the array for ordinary users so it shouldn't matter if we unset fields above or below the section. I'm very likely missing some nuance which is a code-smell.

70ray: The problem is the get_property_key(). If the array is redifined to be just "New" then the values "Closed" or "Reopened" no longer exist and the function get_property_key() throws an error.
Perhaps a better fix would be to say
```
    // Non-managers can only set the task status to New.
    if (!user_is_a_sitemanager() && !user_is_taskcenter_mgr()) {
        $tasks_status_array = [get_property_key("New", $tasks_status_array) => "New"];
    } else {
      // Don't want to permit setting status to "Closed" or "Reopened" when creating/editing a task
      unset($tasks_status_array[get_property_key("Closed", $tasks_status_array)]);
      unset($tasks_status_array[get_property_key("Reopened", $tasks_status_array)]);
}
```

cpeel: Ah, got it. I do like the if/else block better 👍

70ray: I'll make it so.

70ray:
Sandbox re-made.

chrismiceli (approved):

cpeel (approved):

srjfoo (approved):

PR #968, opened 2023-11-02
Squirrels delete illos from images index

srjfoo:
This extends the ability to delete illustrations from the images index to all rounds/pools for squirrels. All others should still be limited to the new project state or P1 Waiting.

Testable in the [image-index-squirrel-delete-illos/](https://www.pgdp.org/~srjfoo/c.branch/image-index-squirrel-delete-illos/) sandbox.

### DIFF
--- a/pinc/Project.inc
+++ b/pinc/Project.inc
@@ -600,6 +600,8 @@ class Project
                 return $this->image_source_name;
             case "can_be_managed_by_current_user":
                 return $this->can_be_managed_by_user($pguser);
+            case "user_can_delete_nonpage_images":
+                return $this->user_can_delete_nonpage_images($pguser);
             case "names_can_be_seen_by_current_user":
                 return $this->names_can_be_seen_by_user($pguser);
             case "PPer":
@@ -977,6 +979,27 @@ class Project
             ($username == $this->username
             || that_user_is_a_sitemanager($username)
             || that_user_is_proj_facilitator($username));
+    }
+
+    public function user_can_delete_nonpage_images($username)
+    {
+        if (is_null($username)) {
+            return false;
+        }
+        // SAs can always delete nonpage images
+        if (that_user_is_a_sitemanager($username)) {
+            return true;
+        }
+
+        // PMs/PFs can only delete nonpage images in New
+        // or P1 Unavailable stages
+        if ($this->can_be_managed_by_user($username)) {
+            return $this->state == PROJ_NEW || $this->state == PROJ_P1_UNAVAILABLE;
+        } else {
+            return false;
+        }
+        // Nobody else can delete images
+        return false;
     }
 
     // -------------------------------------------------------------------------
--- a/tools/project_manager/update_illos.php
+++ b/tools/project_manager/update_illos.php
@@ -62,7 +62,9 @@ if (!$project->can_be_managed_by_current_user) {
     exit;
 }
 
-if (($is_delete_all_operation || $operation == 'delete') && $project->state != PROJ_NEW && $project->state != PROJ_P1_UNAVAILABLE) {
+if (($is_delete_all_operation || $operation == 'delete') && !$project->user_can_delete_nonpage_images) {
+    // Allow squirrels to delete illustrations in any stage
+    // and PMs to do so in project_new or P1.proj_unavail
     echo "<p>", _('You can only delete illustrations for a project in the new or P1 unavailable states.'), "</p>\n";
     provide_escape_links();
     exit;
--- a/tools/proofers/images_index.php
+++ b/tools/proofers/images_index.php
@@ -110,7 +110,8 @@ function list_images($project, $image_names, $these_are_page_images)
     echo "<h4 class='center-align'>$header</h4>";
 
     $show_replace_links = $project->can_be_managed_by_current_user;
-    $show_delete_links = $project->can_be_managed_by_current_user && !$these_are_page_images && ($project->state == PROJ_NEW || $project->state == PROJ_P1_UNAVAILABLE);
+    // If user is allowed to delete nonpage images, show links
+    $show_delete_links = $project->user_can_delete_nonpage_images && !$these_are_page_images;
 
     echo "<table>\n";
 
@@ -196,7 +197,7 @@ function show_delete_all_link($project, $image_names)
 {
     global $code_url;
 
-    if ($project->can_be_managed_by_current_user && ($project->state == PROJ_NEW || $project->state == PROJ_P1_UNAVAILABLE) && !empty($image_names)) {
+    if ($project->user_can_delete_nonpage_images && !empty($image_names)) {
         $form_target = "$code_url/tools/project_manager/update_illos.php";
         $submit_label = _("Delete Illustrations");
         echo "<form action='$form_target' method='POST' style='display: inline'>\n";


srjfoo:
Testing notes: I've backed up the "illustration" files for all projects in P3, F1 and F2 that have them so that I can restore them later if you'll let me know what you deleted.

The important thing is to do an extra confirmation that non-squirrels don't see the delete button or links except where appropriate (I.e. New and P1 Waiting for project PMs and all PFs, and not at all for others.

tools/proofers/images_index.php
```diff
@@ -110,7 +111,7 @@ function list_images($project, $image_names, $these_are_page_images)
     echo "<h4 class='center-align'>$header</h4>";
 
     $show_replace_links = $project->can_be_managed_by_current_user;
-    $show_delete_links = $project->can_be_managed_by_current_user && !$these_are_page_images && ($project->state == PROJ_NEW || $project->state == PROJ_P1_UNAVAILABLE);
+    $show_delete_links = $project->can_be_managed_by_current_user && !$these_are_page_images && ($project->state == PROJ_NEW || $project->state == PROJ_P1_UNAVAILABLE || user_is_a_sitemanager());
```
chrismiceli, line 114: These conditionals are getting a little complex. I wonder if it makes sense to break them up some? Or perhaps extract them to a function so we can unit tests the truth table?

chrismiceli (approved):

windymilla (approved):
Checked squirrel/non-squirrel users do/do not get the Delete button correctly

tools/proofers/images_index.php
cpeel: This could be a function of the Project class, something like:
```php
public function can_user_delete_images($image_type = null)
{
    // SAs can always delete images, page or illustrations
    if (user_is_a_sitemanager()) {
        return true;
    }

    // all other actions can only be done by users who can manage the project
    if (!$this->can_be_managed_by_current_user) {
        return false;
    }

    // if page isn't an illustration but the project is New that's ok
    if ($page_type == "page" && ..
}
```

That just breaks apart this one-line into a readable function.

srjfoo: This PR is only for illustration images, so, I'm guessing the function name should be `can_user_delete_nonpage_images` or something like that? SAs can already delete proofing images from the page details page at any point. If a page image is deleted, the text should be deleted, too, and I don't think we want to deal with that from the images index, so unless you think this should be a multi-purpose function, I wouldn't mention page images at all.

As I thought back after I posted the above, I realized that I did ponder breaking it up into an `if` block, but ultimately opted to keep the compact form.

I like the idea of a function, but my experience with sticking functions into classes and then invoking them is zero, so, I need help with that, too. (I don't mind -- it's a good learning experience, but the learning curve must be traversed.)

cpeel: > I like the idea of a function, but my experience with sticking functions into classes and then invoking them is zero, so, I need help with that, too. (I don't mind -- it's a good learning experience, but the learning curve must be traversed.)

You can model it off of the [can_be_managed_by_user()](https://github.com/DistributedProofreaders/dproofreaders/blob/master/pinc/Project.inc#L956) function in the Project class. It's just like a regular function and you already have a Project object where you're calling it.

srjfoo: *Finally* updated. The sandbox has been updated. I'm happy to do further changes -- this was new to me, so I may not have gotten everything done that I need to. My testing comes back clean, but good to have other testers.

70ray (approved):

tools/project_manager/update_illos.php
```diff
@@ -62,7 +62,9 @@
     exit;
 }
 
-if (($is_delete_all_operation || $operation == 'delete') && $project->state != PROJ_NEW && $project->state != PROJ_P1_UNAVAILABLE) {
+if (($is_delete_all_operation || $operation == 'delete') && !$project->user_can_delete_nonpage_images) {
```
chrismiceli, line 65: don't you have to invoke the function `!$project->user_can_delete_nonpage_images()`? My php is bad though so I may be wrong here.

pinc/Project.inc
```diff
@@ -979,6 +981,31 @@ class Project
...
+        // or P1 Unavailable stages
+        if ($this->can_be_managed_by_user($username)) {
+            if ($this->state == PROJ_NEW || $this->state == PROJ_P1_UNAVAILABLE) {
+                return true;
+            } else {
+                return false;
+            }
+        }
+        // Anyone else can't delete images
+        else {
```
chrismiceli, line 1004: *minor* You don't need the else before this return since all code paths would've returned before already.

pinc/Project.inc
```diff
@@ -979,6 +981,31 @@ class Project
...
+        // SAs can always delete nonpage images
+        if (that_user_is_a_sitemanager($username)) {
+            return true;
+        }
+
+        // PMs/PFs can only delete nonpage images in New
+        // or P1 Unavailable stages
+        if ($this->can_be_managed_by_user($username)) {
+            if ($this->state == PROJ_NEW || $this->state == PROJ_P1_UNAVAILABLE) {
+                return true;
```
chrismiceli, line 998: Up to you if you think this is more clear than just
```
return $this->state == PROJ_NEW || $this->state == PROJ_P1_UNAVAILABLE;
```

tools/project_manager/update_illos.php
cpeel: This is a PHP feature. The `Project` class uses the `__get()` [magic function](https://www.php.net/manual/en/language.oop5.overloading.php#object.set) that allows this to be accessed like a property rather than a function. Sharon wired this into that magic function at line 603 in `Project.inc`.

cpeel (approved):

tools/project_manager/update_illos.php
srjfoo: Say, rather, that it seemed like I **_had_** to wire it into the magic function. 😁  -- I was following the pattern of what had been done before, and that broke things without fixing that, too. (And then, too, even if there's another way to do things, it's better to follow the current pattern if possible, unless there's a reason to rip it all out and redo it.)

chrismiceli (approved):

pinc/Project.inc
srjfoo: I like that better, thanks. I do plan to do it, but I should wait until tomorrow.

pinc/Project.inc
srjfoo: I waffled over this one a bit, but ultimately came down on the side of making the change. But ... tomorrow.

tools/project_manager/update_illos.php
chrismiceli: Oh I agree, I just wasn't aware of this __get thing. Thanks for explaining with links @cpeel

pinc/Project.inc
srjfoo: Done and done! Sandbox has been updated (after each change, I like to verify that I didn't break anything making the changes).

pinc/Project.inc
```diff
@@ -979,6 +981,27 @@ class Project
...
+
+        // PMs/PFs can only delete nonpage images in New
+        // or P1 Unavailable stages
+        if ($this->can_be_managed_by_user($username)) {
+            return $this->state == PROJ_NEW || $this->state == PROJ_P1_UNAVAILABLE;
+        } else {
+            return false;
+        }
+        // Nobody else can delete images
+        return false;
```
chrismiceli, line 1002: *minor* I'm not sure this line is ever reachable since all code paths should've entered at least one of the above ifs or elses. I'd just remove the else on line 999

chrismiceli (approved):

PR #970, opened 2023-11-10
Proposed Swagger file for proofreading

70ray:
This is only the Swagger file without the php code so it is not for merging but for commenting on. It has functions for checking out a number of pages at once, for ordinary users and for mentors.

There is a Sandbox at: https://www.pgdp.org/~rp31/c.branch/proof_api_8. It is possible to load the file into the [Swagger editor](https://editor.swagger.io/) from there.

### DIFF
--- a/api/dp-openapi.yaml
+++ b/api/dp-openapi.yaml
@@ -700,6 +700,10 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/checkout_status'
+              example:
+                status: 303
+                reason: You have reached your quota of pages from 'Beginners Only' projects in this round.
+                npages: 2
         '400':
           $ref: '#/components/responses/InvalidValue'
         '401':
--- a/api/dp-openapi.yaml
+++ b/api/dp-openapi.yaml
@@ -684,24 +684,15 @@ paths:
       - project
       description: Check out some pages for proofreading
       parameters:
-      - name: projectid
-        in: path
-        description: ID of project
-        required: true
-        schema:
-          type: string
-      - name: projstate
-        in: query
-        description: Project state
-        required: true
-        schema:
-          type: string
+      - $ref: '#/components/parameters/projectid'
+      - $ref: '#/components/parameters/projectstate'
       - name: npage
         in: query
-        description: number of pages to checkout, defaults to one
+        description: number of pages to checkout
         required: false
         schema:
           type: integer
+          default: 1
       responses:
         200:
           description: result status
@@ -726,18 +717,8 @@ paths:
       - project
       description: Finds the 'oldest' mentee and checks out all his/her pages.
       parameters:
-      - name: projectid
-        in: path
-        description: ID of project
-        required: true
-        schema:
-          type: string
-      - name: projstate
-        in: query
-        description: Project state
-        required: true
-        schema:
-          type: string
+      - $ref: '#/components/parameters/projectid'
+      - $ref: '#/components/parameters/projectstate'
       responses:
         200:
           description: no content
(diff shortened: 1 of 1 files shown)


api/dp-openapi.yaml
```diff
@@ -678,6 +678,474 @@ paths:
...
+      tags:
+      - project
+      description: Check out some pages for proofreading
+      parameters:
+      - name: projectid
+        in: path
+        description: ID of project
+        required: true
+        schema:
+          type: string
```
chrismiceli, line 692: should we create a custom project type since it often ends up in query params?

api/dp-openapi.yaml
```diff
@@ -678,6 +678,474 @@ paths:
...
+      - name: projstate
+        in: query
+        description: Project state
+        required: true
+        schema:
+          type: string
```
chrismiceli, line 698: same here, maybe an enum of all the states type?

70ray: We could create a reference also.
Although one can write a Swagger definition first and write the code afterwards, I wrote the code first and it does validate the project state against an array of all possible project states (the global variable $PROJECT_STATES_IN_ORDER). It would be possible to put these into an enum in the Swagger file.
On the other hand, if the Swagger file is used by someone to write an application which uses the API they do not need to know what the states actually are since this value will be received from the application. Its only purpose is to ensure the project is still in the same state it was when they received the value.

windymilla:
This probably isn't the place for a discussion of my Swagger/YAML ignorance, but...
I was able to load the yaml file into the Swagger editor only by importing from the [Github URL](https://raw.githubusercontent.com/DistributedProofreaders/dproofreaders/621c51f34127d9e57f31c27587706ef46a7d55a4/api/dp-openapi.yaml) - not sure if there is an easier/better way.
Once in the editor, I guess I would need to Authorize before I could try out anything - I don't know the API key, so couldn't do that - and you also said there's no php code anyway.
The new yaml code makes sense to my limited level of understanding as being what is required for proofreading.

windymilla (approved):

api/dp-openapi.yaml
chrismiceli: Looks like we could use pattern to futher enhance the strength of the type:
https://swagger.io/docs/specification/data-models/data-types/#pattern
and there are a few other places in the swagger we could use the new ref projectid type if we want.

chrismiceli (approved):

70ray:
> I was able to load the yaml file into the Swagger editor only by importing from the [Github URL](https://raw.githubusercontent.com/DistributedProofreaders/dproofreaders/621c51f34127d9e57f31c27587706ef46a7d55a4/api/dp-openapi.yaml) - not sure if there is an easier/better way.

I'm glad that works. I thought there must be a way to import it from Github but I didn't manage to do it myself.

> Once in the editor, I guess I would need to Authorize before I could try out anything - I don't know the API key, so couldn't do that - and you also said there's no php code anyway.

I will upload some code. I was waiting in case any comments on the yaml file would lead to any code changes. I never managed to get the Swagger editor to work into a sandbox.

api/dp-openapi.yaml
70ray: Swagger is able to generate code although we are not using it for that purpose. Presumably the pattern would get built into the code. There is a function `validate_projectID()` in `pinc/Project.inc` which validates a projectid using a regular expression but the validator for projectid in `api/v1_validators.inc` does not use it but uses the function `_load_from_db()` which either loads the project or throws a nonexistentProject exception. Perhaps it would be useful to use `validate_projectID()` in our API validator. What does @cpeel  think?

70ray:
#973 has the same yaml file as this.

PR #971, opened 2023-11-13
separate the text and splitter widgets from page_browse

70ray:
so they can be used elsewhere


chrismiceli (approved):

scripts/widgets.js
```diff
@@ -0,0 +1,214 @@
+/*global $ proofIntData splitControl makeControlDiv */
```
cpeel, line 1: Can we get a less generic name for this file? @chrismiceli may be able to provide better guidance on the JS-way of doing things but a general catch-all file for all widgets doesn't seem like the best approach.

70ray: Thanks. Perhaps split it up even more so there are files text-view.js and view-splitter.js. It it possible a use could arise for a view splitter without the text widget.

chrismiceli: I like that idea!

chrismiceli (approved):

srjfoo (approved):

cpeel (approved):

PR #973, opened 2023-11-17
add proofing api functions and tests

70ray:
The checkout procedures are intended to be used with a revised meaning for save as done and proofread next -- the next already-checked-out page. (It can be used in a way compatible with the present procedure).

The exceptions are all derived from BadRequest so they already have a response code. The code may not be entirely appropriate but so far as our client ajax function (which uses fetch()) is concerned it does not make any difference: Response.ok is true if the response code is in the range 200-299 and false otherwise. This can easily be changed if appropriate.

This generally follows the pattern that if what you want to do does not succeed it throws an exception and returns a response code 400, otherwise 200. Deviations from this pattern:

Checkout a number of pages: If none are available, returns 400. If the number requested can be checked out, returns 200 with status 0. If some can be checked out but not as many as requested, returns 200 with status 1 and a reason string and the corresponding code and the number actually checked out.

When saving, if the text contains invalid characters, does not save the text but returns 200 with data indicating what the problems are. This could be changed to return a not-ok response code together with changing the client ajax function to be able to handle this. Or alternatively add another endpoint to validate the text which can be invoked if the save fails with bad characters. This would need an extra trip to the server but doesn't matter much since this will probably happen rarely.

If the daily limit for saving is reached the text is saved but returns response 400 with a message. Perhaps it would be better to return 200 with a status code and message since the request has succeeded but the user needs to be warned.

If the daily limit for saving is exceeded the text is saved as in progress but returns response 400 with a message.

### DIFF
--- a/api/dp-openapi.yaml
+++ b/api/dp-openapi.yaml
@@ -678,6 +678,315 @@ paths:
         default:
           $ref: '#/components/responses/UnexpectedError'
 
+  /projects/{projectid}/checkout:
+    put:
+      tags:
+      - project
+      description: Check out some pages for proofreading
+      parameters:
+      - $ref: '#/components/parameters/projectid'
+      - $ref: '#/components/parameters/projectstate'
+      - name: npage
+        in: query
+        description: number of pages to checkout
+        required: false
+        schema:
+          type: integer
+          default: 1
+      responses:
+        200:
+          description: result status
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/checkout_status'
+              example:
+                status: 303
+                reason: You have reached your quota of pages from 'Beginners Only' projects in this round.
+                npages: 2
+        '400':
+          $ref: '#/components/responses/InvalidValue'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '429':
+          $ref: '#/components/responses/RateLimitExceeded'
+        default:
+          $ref: '#/components/responses/UnexpectedError'
+
+  /projects/{projectid}/mentorcheckout:
+    put:
+      tags:
+      - project
+      description: Finds the 'oldest' mentee and checks out all his/her pages.
+      parameters:
+      - $ref: '#/components/parameters/projectid'
+      - $ref: '#/components/parameters/projectstate'
+      responses:
+        200:
+          description: no content
+        '400':
+          $ref: '#/components/responses/InvalidValue'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '429':
+          $ref: '#/components/responses/RateLimitExceeded'
+        default:
+          $ref: '#/components/responses/UnexpectedError'
+
+  /projects/{projectid}/returnpages:
+    put:
+      tags:
+      - project
+      description: Returns all the user's checked out pages to the round.
+      parameters:
+      - $ref: '#/components/parameters/projectid'
+      - $ref: '#/components/parameters/projectstate'
+      responses:
+        200:
+          description: no content
+        '400':
+          $ref: '#/components/responses/InvalidValue'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '429':
+          $ref: '#/components/responses/RateLimitExceeded'
+        default:
+          $ref: '#/components/responses/UnexpectedError'
+
+  /projects/{projectid}/nextpage:
+    get:
+      tags:
+      - project
+      description: Gets data for the next checked out page.
+      parameters:
+      - $ref: '#/components/parameters/projectid'
+      - $ref: '#/components/parameters/projectstate'
+      responses:
+        200:
+          description: page data
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/proof_page_data'
+        '400':
+          $ref: '#/components/responses/InvalidValue'
+        '401':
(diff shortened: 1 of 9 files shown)


api/dp-openapi.yaml
```diff
@@ -678,6 +678,315 @@ paths:
...
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '429':
+          $ref: '#/components/responses/RateLimitExceeded'
+        default:
+          $ref: '#/components/responses/UnexpectedError'
+
+  /projects/{projectid}/returnpages:
```
cpeel, line 740: This feels like an unnecessary abstraction and should be left up to the client to return each page individually.

cpeel (commented):
I think we're trying to cram too much application logic into the API. The API should probably be simpler thereby allowing the client more flexibility.

api/dp-openapi.yaml
srjfoo: > This feels like an unnecessary abstraction and should be left up to the client to return each page individually.

Bear in mind that I know nothing about designing an API, but it makes a lot of sense to me that the API be simple and straightforward, and that the software using the API determine how it's implemented.

api/dp-openapi.yaml
```diff
@@ -678,6 +678,315 @@ paths:
...
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '429':
+          $ref: '#/components/responses/RateLimitExceeded'
+        default:
+          $ref: '#/components/responses/UnexpectedError'
+
+  /projects/{projectid}/pages/{pagename}/saveasdone:
```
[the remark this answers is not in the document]
70ray: Yes, I started off doing it like that. This method seemed simpler because it uses the mechanism already present to select different functions rather than having to add another selection mechanism and validate the options.

api/dp-openapi.yaml
```diff
@@ -678,6 +678,315 @@ paths:
...
+    put:
+      tags:
+      - project
+      description: Check out some pages for proofreading
+      parameters:
+      - $ref: '#/components/parameters/projectid'
+      - $ref: '#/components/parameters/projectstate'
+      - name: npage
+        in: query
+        description: number of pages to checkout
```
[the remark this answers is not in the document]
70ray: There were two motivations for checking out multiple pages at a time.

- Ordinary proofreaders often like to work on consecutive pages and it is possible to checkout several pages quickly in the present system by refreshing the proofreading page. This is probably not something we want to do so to pre-empt complaints about reducing functionality we can have a way to checkout several pages. We could do this by checking out one at a time and let the client application call the function several times. The server would still have to impose a limit on how many can be checked out at once and it would still be best to have different functions for checking out and getting the page data to avoid excessive traffic.
- Mentors need to check out all the pages for one proofreader. Finding the 'oldest proofreader' and checking out a block seemed like a simple way to do it. If you only checkout one page then the 'oldest' proofreader can then become another person so you would have to have a mechanism for selecting a particular proofreader.

> Checking out one page at a time has the benefit that we can mitigate a denial of service (someone checking out all of the pages in a round) because we limit the number of API calls a user can do per unit of time.

This will not be a problem because it is not possible to checkout another block of pages if you already have some checked out.

api/dp-openapi.yaml
70ray: > This feels like an unnecessary abstraction and should be left up to the client to return each page individually.

This was actually requested in an earlier demonstration of a method of implementing a mentor's proofreading interface.

srjfoo: > > This feels like an unnecessary abstraction and should be left up to the client to return each page individually.
> 
> This was actually requested in an earlier demonstration of a method of implementing a mentor's proofreading interface.

Come to think of it, I was probably the one who pointed out that if a proofer has several pages to return, they might not bother. Of course that's true now, also. Does the "return block" have to be specified in the API?

70ray: > Does the "return block" have to be specified in the API?

Not necessarily. But it would make life easier. Otherwise the client software would have to get each checked-out page and return it until there were no more left. So returning 5 pages would need 11 api calls, 5 returning the text payload, and the last reporting there were no more pages checked out.

(In the demo referred to this function is called from the project page rather than the Proofreading interface).

70ray:
I'm inclined to make the text validation a separate endpoint.

api/dp-openapi.yaml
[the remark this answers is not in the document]
70ray: > What's the use-case for checking out multiple pages at a time and why can't we fulfill it via multiple single calls? I see that the implementation has `MAX_REQUESTED_PAGES` to restrict how many pages a single call can check out, but I'm not clear on why we need more than one to begin with.

You are right. I will revise this to make it work with single  page checkout.

The mentors use-case is a bit different but I'll consider it further.

pinc/ProofProject.inc
```diff
@@ -0,0 +1,598 @@
...
+        $reason = "";
+        $status_code = 0;
+        $this->validate_no_outpages();
+        $out_pages = 0;
+        validate_user_can_get_pages_in_project($pguser, $this->project, $this->round);
+        while ($out_pages < $requested_pages) {
+            try {
+                $this->checkout_available_proof_page();
+                $out_pages += 1;
+            } catch (ProjectNoMorePagesException | BeginnersQuotaReachedException | BeginnersProjectQuotaReachedException $exception) {
```
70ray, line 132: This is unsatisfactory because future changes to why no more pages can be checked out could break this.

70ray:
Going to remove the multi-page checkout.

PR #974, opened 2023-11-26
enable logging errors for validating zip and add tests

70ray:
Rather than logging errors from within  'is_valid_zip_file()', the function 'validate_zip_file()' is used to enable testing. The tests are the same as the tests for 'is_valid_zip_file()' in misc_ZipMethodsTest.php plus an extra one to test for a file which is not a zip archive.


pinc/misc.inc
```diff
@@ -882,14 +882,30 @@ function flatten_directory(string $directory_to_flatten)
...
+{
+}
+
 function is_valid_zip_file(string $file_path, bool $ignore_extension_check = false): bool
 {
-    if (!file_exists($file_path)) {
+    try {
+        validate_zip_file($file_path, $ignore_extension_check);
+    } catch (ZipException $exception) {
+        error_log("is_valid_zip_file: " . $exception->getMessage());
```
cpeel, line 895: Our usual `error_log()` format string in the code is `<filename> - $error` but that doesn't make a lot of sense here as `misc.inc` is a common include. Consider instead: `is_valid_zip_file() - $error`.

In general I'm a little dubious of logging an error here in a generic validation function. I suspect as a pattern that will generate a fair amount of noise compared to the value. It might be better to have this function just return true/false as it did before without logging the exception and update `zip_check()` to run the new validation function and capture and log the error there with the filename in the message like we do elsewhere.

SETUP/tests/ZipValidateTest.php
```diff
@@ -0,0 +1,82 @@
+<?php
```
cpeel, line 1: Is there a reason we didn't add these functions to `misc_ZipMethodsTest.php`? If we want them in a different file it should follow the `misc_<class>.php` naming convention.

cpeel (commented):
Great job on the new validation function and the new tests for it!

pinc/misc.inc
```diff
@@ -898,9 +914,21 @@ function is_valid_zip_file(string $file_path, bool $ignore_extension_check = fal
 
     if ($result === true) {
         $zip->close();
+    } else {
+        $zip_results = [
```
chrismiceli, line 918: should we define this as a global so we don't have to reconstruct it every error? I suppose it doesn't matter much.

pinc/misc.inc
```diff
@@ -898,9 +914,21 @@ function is_valid_zip_file(string $file_path, bool $ignore_extension_check = fal
...
+            ZipArchive::ER_INCONS => "Zip archive inconsistent.",
+            ZipArchive::ER_INVAL => "Invalid argument.",
+            ZipArchive::ER_MEMORY => "Malloc failure.",
+            ZipArchive::ER_NOENT => "No such file.",
+            ZipArchive::ER_NOZIP => "Not a zip archive.",
+            ZipArchive::ER_OPEN => "Can't open file.",
+            ZipArchive::ER_READ => "Read error.",
+            ZipArchive::ER_SEEK => "Seek error.",
+        ];
+        $error = $zip_results[$result];
```
chrismiceli, line 929: do we want to handle the case some other number was in `$result`? can we add some default case to the map or should we use a switch case?

SETUP/tests/ZipValidateTest.php
70ray: The reason is that when I tried to run the `misc_ZipMethodsTest.php` alone from the command line phpUnit said 
> Class misc_ZipMethodsTest could not be found in /home/rp31/public_html/u-rp75-dproofreaders/SETUP/tests/misc_ZipMethodsTest.php

The documentation says
> When invoked as shown above, the PHPUnit command-line test runner will look for a ArrayTest.php source code file in the current working directory, load it, and expect to find an ArrayTest test case class. It will then run the tests found in that class.

so I thought the easiest way round the problem was to make a new file. But, as you say, it would be better to put them in the existing file. Is there away to run it from the command line?

pinc/misc.inc
70ray: I don't know what is best.

pinc/misc.inc
70ray: Yes, I suppose we should allow for that case. These were the codes in the [PHP doc](https://www.php.net/manual/en/ziparchive.open.php). One of the user contributions on the page gave a longer list.

pinc/misc.inc
70ray: I'll change it to work like that.

70ray:
I've made a Sandbox at: https://www.pgdp.org/~rp31/c.branch/log_zip

It would be useful to try uploading a bad zip file to see if an appropriate entry appears in the log.

chrismiceli (approved):

SETUP/tests/ZipValidateTest.php
cpeel: Huh. I think the older version of phpunit we started with worked differently (you could just give it the filename and it would just run the tests in that file). You can run specific classes like this:
```
../../vendor/bin/phpunit --filter MiscUtils
```

pinc/misc.inc
cpeel: If you make it a static variable it is only constructed on first use, but I don't think we need that optimization here because the odds that this function will be used many times on a single page and return errors all of those times is low.

windymilla (approved):
Tested:
1. Good zip file - uploaded OK - no entry in log
2. Bad zip file - error displayed on webpage and logged correctly

srjfoo (approved):
I think this is okay as far as it goes (and can't hurt), but I wanted to add some additional comments. I asked Rick a few more questions via Slack, for clarification.
* It doesn't happen all the time
* It happens on both his desktop and laptop, both on his home network and on the out of town network
* The zips he's uploading are apparently valid zips, because the upload succeeds on a subsequent attempt[1]
* It hasn't happened always, but seems to have happened regularly (though not with every project) over about the past year.
* It's only happened for SR uploads. Not uploads to dpscans, and not PP backup files: he did some `first_in_prog` uploads while we were chatting (5 in a row with no failures).

I've asked him when it happens (before or after the virus check), and if he can share a screenshot the next time it happens.

70ray:
I've moved the tests into misc_ZipMethodsTest.php. Perhaps we could delete some of them since they are duplicated at different levels. But they don't take much time.

PR #975, opened 2023-11-30
Proofreading api with single page checkout

70ray:
Sandbox at: https://www.pgdp.org/~rp31/c.branch/proof_api_10

### DIFF
--- a/api/dp-openapi.yaml
+++ b/api/dp-openapi.yaml
@@ -1029,32 +1029,6 @@ paths:
           $ref: '#/components/responses/RateLimitExceeded'
         default:
           $ref: '#/components/responses/UnexpectedError'
-    get:
-      tags:
-      - project
-      description: Gets the current (last saved) page text
-      parameters:
-      - $ref: '#/components/parameters/projectid'
-      - $ref: '#/components/parameters/state_query'
-      - $ref: '#/components/parameters/pagename'
-      - $ref: '#/components/parameters/pagestate_query'
-      responses:
-        200:
-          description: Page text
-          content:
-            application/json:
-              schema:
-                $ref: '#/components/schemas/proof_page_text'
-        '400':
-          $ref: '#/components/responses/InvalidValue'
-        '401':
-          $ref: '#/components/responses/Unauthorized'
-        '404':
-          $ref: '#/components/responses/NotFound'
-        '429':
-          $ref: '#/components/responses/RateLimitExceeded'
-        default:
-          $ref: '#/components/responses/UnexpectedError'
 
   /stats/site:
     get:
@@ -1545,13 +1519,6 @@ components:
       type: array
       items:
         type: string
-
-    proof_page_text:
-      type: object
-      properties:
-        text:
-          description: The page text from current round.
-          type: string
 
     proof_page:
       type: object
--- a/api/dp-openapi.yaml
+++ b/api/dp-openapi.yaml
@@ -1030,6 +1030,33 @@ paths:
         default:
           $ref: '#/components/responses/UnexpectedError'
 
+    get:
+      tags:
+      - project
+      description: Get page text and state
+      parameters:
+      - $ref: '#/components/parameters/projectid'
+      - $ref: '#/components/parameters/pagename'
+      responses:
+        200:
+          description: Page data
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/page_text_state'
+        '400':
+          $ref: '#/components/responses/InvalidValue'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '403':
+          $ref: '#/components/responses/Forbidden'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '429':
+          $ref: '#/components/responses/RateLimitExceeded'
+        default:
+          $ref: '#/components/responses/UnexpectedError'
+
   /stats/site:
     get:
       tags:
@@ -1571,9 +1598,8 @@ components:
           description: The page state, it is now in a 'temp' state since this is returned by save-as-in-progress or save-and-revert-to-original
           type: string
         saved:
-          description: Now true (since page has been saved)
+          description: true if the page has been saved.
           type: boolean
-          enum: [true]
 
     text_analysis:
       type: object
--- a/api/dp-openapi.yaml
+++ b/api/dp-openapi.yaml
@@ -924,6 +924,138 @@ paths:
         default:
           $ref: '#/components/responses/UnexpectedError'
 
+  /projects/{projectid}/checkout:
+    put:
+      tags:
+      - project
+      description: Check out a page for proofreading.
+      parameters:
+      - $ref: '#/components/parameters/projectid'
+      - $ref: '#/components/parameters/state_query'
+      responses:
+        200:
+          description: page data
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/proof_page'
+        '400':
+          $ref: '#/components/responses/InvalidValue'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '403':
+          $ref: '#/components/responses/Forbidden'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '429':
+          $ref: '#/components/responses/RateLimitExceeded'
+        default:
+          $ref: '#/components/responses/UnexpectedError'
+
+  /projects/{projectid}/validatetext:
+    put:
+      tags:
+      - project
+      description: Checks the given text for invalid characters, returns data indicating them
+      parameters:
+      - $ref: '#/components/parameters/projectid'
+      requestBody:
+        $ref: '#/components/requestBodies/page_text'
+      responses:
+        200:
+          description: analysis data
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/text_analysis'
+        '400':
+          $ref: '#/components/responses/InvalidValue'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '429':
+          $ref: '#/components/responses/RateLimitExceeded'
+        default:
+          $ref: '#/components/responses/UnexpectedError'
+
+  /projects/{projectid}/pages/{pagename}:
+    put:
+      tags:
+      - project
+      description: Various operations on the page as determined by the 'pageaction_query' parameter
+      parameters:
+      - $ref: '#/components/parameters/projectid'
+      - $ref: '#/components/parameters/state_query'
+      - $ref: '#/components/parameters/pagename'
+      - $ref: '#/components/parameters/pagestate_query'
+      - name: pageaction_query
+        in: query
+        description: Action to perform on the page.
+            <br><br>
+            resume - If the page is saved-as-done 'unsaves' it. Response:- proof_page object.
+            <br><br>
+            checkin - Checks the page for invalid characters and if valid, saves it as done. Response:- checkin_response object.
+            <br><br>
+            save - Checks the page for invalid characters and if valid, saves it as in progress. Response:- page_text_state object.
+            <br><br>
+            revert - Checks the page for invalid characters and if valid, saves it as in progress. Response:- page_text_state object.
+            <br><br>
+            abandon - Returns the page to the round, response:- none.
+        required: true
+        schema:
+          type: string
+      requestBody:
+        $ref: '#/components/requestBodies/page_text'
+      responses:
+        200:
+          description: Page data
+          content:
+            application/json:
+              schema:
+                oneOf:
+                  - $ref: '#/components/schemas/checkin_response'
+                  - $ref: '#/components/schemas/proof_page'
+                  - $ref: '#/components/schemas/page_text_state'
+        '400':
+          $ref: '#/components/responses/InvalidValue'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '403':
+          $ref: '#/components/responses/Forbidden'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '429':
+          $ref: '#/components/responses/RateLimitExceeded'
+        default:
+          $ref: '#/components/responses/UnexpectedError'
+    get:
+      tags:
+      - project
+      description: Gets the current (last saved) page text
+      parameters:
+      - $ref: '#/components/parameters/projectid'
+      - $ref: '#/components/parameters/state_query'
+      - $ref: '#/components/parameters/pagename'
+      - $ref: '#/components/parameters/pagestate_query'
+      responses:
+        200:
+          description: Page text
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/proof_page_text'
+        '400':
+          $ref: '#/components/responses/InvalidValue'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '429':
+          $ref: '#/components/responses/RateLimitExceeded'
+        default:
+          $ref: '#/components/responses/UnexpectedError'
+
   /stats/site:
     get:
       tags:
@@ -1008,6 +1140,50 @@ components:
       type: apiKey
       in: header
       name: X-API-KEY
+
+  parameters:
+    projectid:
+      name: projectid
+      in: path
+      description: ID of project
+      required: true
+      schema:
+        type: string
+
+    state_query:
+      name: state
+      in: query
+      description: Project state
+      required: true
+      schema:
+        type: string
+
+    pagename:
+      name: pagename
+      in: path
+      description: Name of page
+      required: true
+      schema:
+        type: string
+
+    pagestate_query:
+      name: pagestate
+      in: query
+      description: Page state
+      required: true
+      schema:
+        type: string
+
+  requestBodies:
+    page_text:
+      description: Text to save or analyse
+      content:
+        application/json:
+          schema:
+            type: object
+            properties:
+              text:
+                type: string
 
   schemas:
     project:
@@ -1370,6 +1546,82 @@ components:
       items:
         type: string
 
+    proof_page_text:
+      type: object
+      properties:
+        text:
+          description: The page text from current round.
+          type: string
+
+    proof_page:
+      type: object
+      properties:
+        text:
+          description: The page text. If page state is 'out' from OCR or previous round, else from current round
+          type: string
+        pagestate:
+          description: The page state, for checking continuity. Either out or temp
+          type: string
+        saved:
+          description: True if the page state is 'temp'
+          type: boolean
+        pagename:
+          description: The filename of the page with extension
+          type: string
+        image_url:
+          description: URL of the image of the page
+          type: string
+        language_direction:
+          description: left-to-right or right-to-left
+          type: string
+          enum: [RTL, LTR]
+        pagenum:
+          description: Page "number" without extension
+          type: string
+        round_info_array:
+          description: Array of names of proofreaders in earlier rounds
+          type: array
+          items:
+            type: string
+
+    checkin_response:
+      type: object
+      properties:
+        message:
+          description: message if reached limit, else empty string.
+          type: string
+        status:
+          description: 1 if limit reached, else 0
+          type: integer
+
+    page_text_state:
+      type: object
+      properties:
+        text:
+          description: The page text, current or original.
+          type: string
+        pagestate:
+          description: The page state, it is now in a 'temp' state since this is returned by save-as-in-progress or save-and-revert-to-original
+          type: string
+        saved:
+          description: Now true (since page has been saved)
+          type: boolean
+          enum: [true]
+
+    text_analysis:
+      type: object
+      properties:
+        valid:
+          type: boolean
+        mark_array:
+          type: array
+          items:
+            type: array
+            items:
+              oneOf:
+                - type: string
+                - type: integer
+
     Error:
       required:
       - message
@@ -1390,6 +1642,12 @@ components:
             $ref: '#/components/schemas/Error'
     Unauthorized:
       description: Unauthorized
+      content:
+        application/json:
+          schema:
+            $ref: '#/components/schemas/Error'
+    Forbidden:
+      description: Forbidden
       content:
         application/json:
           schema:
--- a/api/dp-openapi.yaml
+++ b/api/dp-openapi.yaml
@@ -1572,7 +1572,7 @@ components:
         pagenum:
           description: Page "number" without extension
           type: string
-        round_info_array:
+        round_info:
           description: Array of names of proofreaders in earlier rounds
           type: array
           items:
--- a/api/dp-openapi.yaml
+++ b/api/dp-openapi.yaml
@@ -928,7 +928,7 @@ paths:
     put:
       tags:
       - project
-      description: Check out a page for proofreading.
+      description: Check out a page for proofreading and return page-level details (current text, image name, etc).
       parameters:
       - $ref: '#/components/parameters/projectid'
       - $ref: '#/components/parameters/state_query'
@@ -983,17 +983,17 @@ paths:
     put:
       tags:
       - project
-      description: Various operations on the page as determined by the 'pageaction_query' parameter
+      description: Various operations on the page as determined by the 'pageaction' parameter
       parameters:
       - $ref: '#/components/parameters/projectid'
       - $ref: '#/components/parameters/state_query'
       - $ref: '#/components/parameters/pagename'
       - $ref: '#/components/parameters/pagestate_query'
-      - name: pageaction_query
+      - name: pageaction
         in: query
         description: Action to perform on the page.
             <br><br>
-            resume - If the page is saved-as-done 'unsaves' it. Response:- proof_page object.
+            resume - The page state can be 'out', 'temp' or 'saved'. If the state is 'saved' it is changed to 'temp'. Returns page-level details (current text, image name, etc). Response:- proof_page object.
             <br><br>
             checkin - Checks the page for invalid characters and if valid, saves it as done. Response:- checkin_response object.
             <br><br>
--- a/api/exceptions.inc
+++ b/api/exceptions.inc
@@ -81,6 +81,19 @@ class InvalidValue extends ApiException
     }
 }
 
+class ForbiddenError extends ApiException
+{
+    public function __construct($message = "Forbidden", $code = 7)
+    {
+        parent::__construct($message, $code);
+    }
+
+    public function getStatusCode()
+    {
+        return 403;
+    }
+}
+
 //---------------------------------------------------------------------------
 // Exceptions that shouldn't happen unless someone is futzing with something
 // they shouldn't be, or there's an unexpected problem.
--- a/api/v1_projects.inc
+++ b/api/v1_projects.inc
@@ -775,6 +775,14 @@ function _get_enabled_filter($query_params)
 //---------------------------------------------------------------------------
 // Proofreading functions
 
+function api_v1_project_page_data(string $method, array $data, array $query_params)
+{
+    $project_page = $data[":pagename"];
+    $proof_project = new ProofProject($data[":projectid"]);
+    $proof_project_page = new ProofProjectPage($proof_project, $project_page);
+    return $proof_project_page->get_page_text_data();
+}
+
 // checkout a page
 function api_v1_project_checkout($method, array $data, array $query_params)
 {
--- a/api/v1_projects.inc
+++ b/api/v1_projects.inc
@@ -775,19 +775,14 @@ function _get_enabled_filter($query_params)
 //---------------------------------------------------------------------------
 // Proofreading functions
 
-function api_v1_project_page_data(string $method, array $data, array $query_params)
-{
-    $project_page = $data[":pagename"];
-    $proof_project = new ProofProject($data[":projectid"]);
-    $proof_project_page = new ProofProjectPage($proof_project, $project_page);
-    return $proof_project_page->get_page_text_data();
-}
-
 // checkout a page
 function api_v1_project_checkout($method, array $data, array $query_params)
 {
     try {
-        $proof_project = create_proof_project($data, $query_params);
+        $project = $data[":projectid"];
+        $state = $query_params["state"] ?? null;
+        validate_project_state($project, $state);
+        $proof_project = new ProofProject($project);
         return $proof_project->checkout();
     } catch (ProjectException $exception) {
         throw new NotFoundError($exception->getMessage(), $exception->getCode());
@@ -807,10 +802,19 @@ function api_v1_project_page(string $method, array $data, array $query_params)
     global $pguser;
 
     try {
-        $proof_project = create_proof_project($data, $query_params);
+        $project = $data[":projectid"];
         $project_page = $data[":pagename"];
-        $page_state = $query_params['pagestate'] ?? null;
+        if ($method == "GET") {
+            $proof_project = new ProofProject($project);
+            $proof_project_page = new ProofProjectPage($proof_project, $project_page);
+            return $proof_project_page->get_page_text_data();
+        }
+        // $method is "PUT"
+        $state = $query_params["state"] ?? null;
+        validate_project_state($project, $state);
+        $proof_project = new ProofProject($project);
 
+        $page_state = $query_params['pagestate'] ?? null;
         if (null === $page_state) {
             throw new InvalidValue("No page state found in request.");
         }
@@ -853,30 +857,23 @@ function api_v1_project_page(string $method, array $data, array $query_params)
     }
 }
 
-function create_proof_project(array $data, array $query_params)
+function validate_project_state($project, $state)
 {
-    $project = $data[":projectid"];
-    $state = $query_params["state"] ?? null;
-    try {
-        if (null === $state) {
-            throw new InvalidValue("No project state found in request.");
-        }
-        if (!in_array($state, ProjectStates::get_states())) {
-            throw new InvalidValue("Invalid project state");
-        }
-        if ($state != $project->state) {
-            $err = sprintf(
-                _('Project "%1$s" is not in state "%2$s"; it is now in state "%3$s".'),
-                $project->nameofwork,
-                project_states_text($state),
-                project_states_text($project->state)
-            );
-            throw new ProjectNotInRequiredStateException($err);
-        }
-    } catch (ProjectException $exception) {
-        throw new NotFoundError($exception->getMessage(), $exception->getCode());
+    if (null === $state) {
+        throw new InvalidValue("No project state found in request.");
     }
-    return new ProofProject($project);
+    if (!in_array($state, ProjectStates::get_states())) {
+        throw new InvalidValue("Invalid project state");
+    }
+    if ($state != $project->state) {
+        $err = sprintf(
+            _('Project "%1$s" is not in state "%2$s"; it is now in state "%3$s".'),
+            $project->nameofwork,
+            project_states_text($state),
+            project_states_text($project->state)
+        );
+        throw new ProjectNotInRequiredStateException($err);
+    }
 }
 
 function receive_project_text_from_request_body($field): string
--- a/api/v1_projects.inc
+++ b/api/v1_projects.inc
@@ -820,10 +820,6 @@ function api_v1_project_page(string $method, array $data, array $query_params)
         }
 
         $proof_project_page = new ProofProjectPage($proof_project, $project_page);
-        if ($method == "GET") {
-            return $proof_project_page->render_page_text();
-        }
-        // must be PUT
         $page_action = $query_params['pageaction'] ?? null;
         if (null === $page_action) {
             throw new BadRequest("No page action found in request.");
--- a/api/v1_projects.inc
+++ b/api/v1_projects.inc
@@ -4,6 +4,7 @@ include_once($relPath.'wordcheck_engine.inc');
 include_once($relPath.'ProjectSearchForm.inc');
 include_once($relPath.'page_table.inc');
 include_once($relPath.'special_colors.inc');
+include_once($relPath.'ProofProject.inc');
 include_once("exceptions.inc");
 include_once("api_common.inc");
 
@@ -548,7 +549,7 @@ function api_v1_project_page_round($method, $data, $query_params)
         $user_column,
         implode(",", $user_cols),
         $project->projectid,
-        DPDatabase::escape($data[":pagename"])
+        DPDatabase::escape($data[":pagename"]->page_name)
     );
     $result = DPDatabase::query($sql);
     $row = mysqli_fetch_assoc($result);
@@ -769,4 +770,117 @@ function api_v1_projects_holdstates($method, $data, $query_params)
 function _get_enabled_filter($query_params)
 {
     return get_flag_value($query_params, "enabled");
+}
+
+//---------------------------------------------------------------------------
+// Proofreading functions
+
+// checkout a page
+function api_v1_project_checkout($method, array $data, array $query_params)
+{
+    try {
+        $proof_project = create_proof_project($data, $query_params);
+        return $proof_project->checkout();
+    } catch (ProjectException $exception) {
+        throw new NotFoundError($exception->getMessage(), $exception->getCode());
+    } catch (UserAccessException $exception) {
+        throw new ForbiddenError($exception->getMessage(), $exception->getCode());
+    }
+}
+
+function api_v1_project_validatetext($method, $data, array $query_params)
+{
+    $text_checker = new TextChecker();
+    return $text_checker->analyse(receive_project_text_from_request_body('text'), $data[":projectid"]);
+}
+
+function api_v1_project_page(string $method, array $data, array $query_params)
+{
+    global $pguser;
+
+    try {
+        $proof_project = create_proof_project($data, $query_params);
+        $project_page = $data[":pagename"];
+        $page_state = $query_params['pagestate'] ?? null;
+
+        if (null === $page_state) {
+            throw new InvalidValue("No page state found in request.");
+        }
+        if (!in_array($page_state, Rounds::get_page_states())) {
+            throw new InvalidValue(sprintf("%s is not a valid page state", $page_state));
+        }
+        if ($page_state != $project_page->page_state) {
+            $err = sprintf(
+                _('Warning: Page  "%1$s" is no longer in state "%2$s"; it is now in state "%3$s".'),
+                $project_page->page_name,
+                $page_state,
+                $project_page->page_state
+            );
+            throw new ProjectPageInconsistentStateException($err);
+        }
+
+        $proof_project_page = new ProofProjectPage($proof_project, $project_page);
+        if ($method == "GET") {
+            return $proof_project_page->render_page_text();
+        }
+        // must be PUT
+        $page_action = $query_params['pageaction'] ?? null;
+        if (null === $page_action) {
+            throw new BadRequest("No page action found in request.");
+        }
+        switch ($page_action) {
+            case 'resume':
+                return $proof_project_page->pp_resume_page();
+            case 'checkin':
+                return $proof_project_page->attempt_checkin(receive_project_text_from_request_body('text'));
+            case 'save':
+                return $proof_project_page->save(receive_project_text_from_request_body('text'));
+            case 'revert':
+                return $proof_project_page->save_and_revert(receive_project_text_from_request_body('text'));
+            case 'abandon':
+                return $proof_project_page->returnToRound($pguser);
+            default:
+                throw new BadRequest("$page_action is not a valid page action.");
+        }
+    } catch (ProjectException $exception) {
+        throw new NotFoundError($exception->getMessage(), $exception->getCode());
+    } catch (UserAccessException $exception) {
+        throw new ForbiddenError($exception->getMessage(), $exception->getCode());
+    }
+}
+
+function create_proof_project(array $data, array $query_params)
+{
+    $project = $data[":projectid"];
+    $state = $query_params["state"] ?? null;
+    try {
+        if (null === $state) {
+            throw new InvalidValue("No project state found in request.");
+        }
+        if (!in_array($state, ProjectStates::get_states())) {
+            throw new InvalidValue("Invalid project state");
+        }
+        if ($state != $project->state) {
+            $err = sprintf(
+                _('Warning: Project "%1$s" is not in state "%2$s"; it is now in state "%3$s".'),
+                $project->nameofwork,
+                project_states_text($state),
+                project_states_text($project->state)
+            );
+            throw new ProjectNotInRequiredStateException($err);
+        }
+    } catch (ProjectException $exception) {
+        throw new NotFoundError($exception->getMessage(), $exception->getCode());
+    }
+    return new ProofProject($project);
+}
+
+function receive_project_text_from_request_body($field): string
+{
+    $request_data = api_get_request_body();
+    $page_text = $request_data[$field] ?? null;
+    if (null === $page_text) {
+        throw new InvalidValue("There is no text");
+    }
+    return $page_text;
 }
--- a/api/v1_projects.inc
+++ b/api/v1_projects.inc
@@ -823,7 +823,7 @@ function api_v1_project_page(string $method, array $data, array $query_params)
         }
         if ($page_state != $project_page->page_state) {
             $err = sprintf(
-                _('Page  "%1$s" is not in state "%2$s"; it is now in state "%3$s".'),
+                _('Page "%1$s" is not in state "%2$s"; it is now in state "%3$s".'),
                 $project_page->page_name,
                 $page_state,
                 $project_page->page_state
@@ -833,10 +833,9 @@ function api_v1_project_page(string $method, array $data, array $query_params)
 
         $proof_project_page = new ProofProjectPage($proof_project, $project_page);
         $page_action = $query_params['pageaction'] ?? null;
-        if (null === $page_action) {
-            throw new BadRequest("No page action found in request.");
-        }
         switch ($page_action) {
+            case null:
+                throw new BadRequest("No page action found in request.");
             case 'resume':
                 return $proof_project_page->pp_resume_page();
             case 'checkin':
--- a/api/v1_projects.inc
+++ b/api/v1_projects.inc
@@ -811,7 +811,7 @@ function api_v1_project_page(string $method, array $data, array $query_params)
         }
         if ($page_state != $project_page->page_state) {
             $err = sprintf(
-                _('Warning: Page  "%1$s" is no longer in state "%2$s"; it is now in state "%3$s".'),
+                _('Page  "%1$s" is not in state "%2$s"; it is now in state "%3$s".'),
                 $project_page->page_name,
                 $page_state,
                 $project_page->page_state
@@ -862,7 +862,7 @@ function create_proof_project(array $data, array $query_params)
         }
         if ($state != $project->state) {
             $err = sprintf(
-                _('Warning: Project "%1$s" is not in state "%2$s"; it is now in state "%3$s".'),
+                _('Project "%1$s" is not in state "%2$s"; it is now in state "%3$s".'),
                 $project->nameofwork,
                 project_states_text($state),
                 project_states_text($project->state)
--- a/api/v1_validators.inc
+++ b/api/v1_validators.inc
@@ -40,7 +40,7 @@ function validate_page_name($pagename, $data)
 {
     try {
         return $data[":projectid"]->get_project_page($pagename);
-    } catch (NonexistentPageException $exception) {
+    } catch (NonexistentPageException | NoProjectPageTable $exception) {
         throw new NotFoundError($exception->getMessage(), $exception->getCode());
     }
 }
--- a/api/v1_validators.inc
+++ b/api/v1_validators.inc
@@ -39,12 +39,8 @@ function validate_wordlist($wordlist, $data)
 function validate_page_name($pagename, $data)
 {
     try {
-        $pages = $data[":projectid"]->get_page_names_from_db();
-        if (!in_array($pagename, $pages)) {
-            throw new NonexistentPageException("No such page in project");
-        }
-        return $pagename;
-    } catch (NonexistentPageException | NoProjectPageTable $exception) {
+        return $data[":projectid"]->get_project_page($pagename);
+    } catch (NonexistentPageException $exception) {
         throw new NotFoundError($exception->getMessage(), $exception->getCode());
     }
 }
--- a/pinc/Project.inc
+++ b/pinc/Project.inc
@@ -1901,6 +1901,10 @@ class Project
 
     public function get_project_page($pagename)
     {
+        if (!$this->pages_table_exists) {
+            throw new NoProjectPageTable(_("Project page table does not exist."));
+        }
+
         $sql = sprintf(
             "
             SELECT state
--- a/pinc/Project.inc
+++ b/pinc/Project.inc
@@ -155,7 +155,10 @@ class ProjectInvalidTextException extends ProjectException
 
 class ProjectPage
 {
-    public function __construct($pagename, $pagestate)
+    public $page_name;
+    public $page_state;
+
+    public function __construct(string $pagename, string $pagestate)
     {
         $this->page_name = $pagename;
         $this->page_state = $pagestate;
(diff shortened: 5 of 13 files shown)


pinc/ProofProject.inc
```diff
@@ -0,0 +1,362 @@
...
+                    );
+                    throw new DailyLimitReachedException($sentence);
+                }
+            }
+        } else {
+            $this->save_as_done();
+        }
+    }
+
+    private function save_as_done(): void
```
chrismiceli, line 212: is there a reason some private functions start with `_` and others don't?

pinc/ProofProject.inc
```diff
@@ -0,0 +1,362 @@
...
+        if ($page_option == "current") {
+            $desired_column_name = $this->round->text_column_name;
+        } else {
+            $desired_column_name = $this->round->prevtext_column_name;
+        }
+
+        $sql = "
+            SELECT $desired_column_name
+            FROM $this->projectid
+            WHERE image='$this->page_name'
```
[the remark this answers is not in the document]
70ray: Yes, there are other places where image is escaped so I'll do that. The other things are predefined so shouldn't need escaping but it wouldn't hurt.

pinc/ProofProject.inc
```diff
@@ -0,0 +1,362 @@
...
+        $expected_state = $query_params["projstate"] ?? null;
+        if (null === $expected_state) {
+            throw new ProjectNoStateGivenException("No expected state found in request.");
+        }
+        if (!in_array($expected_state, $PROJECT_STATES_IN_ORDER)) {
+            throw new ProjectInvalidStateException(sprintf("%s is not a valid project state", $expected_state));
+        }
+        if ($expected_state != $project->state) {
+            $err = sprintf(
+                _('Warning: Project "%1$s" is not in state "%2$s"; it is now in state "%3$s".'),
```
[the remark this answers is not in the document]
70ray: It may not be consistent but the plan was to localize messages that the user would see in normal use, but not messages that would only appear if there was something wrong with the client application.

pinc/ProofProject.inc
70ray: Not really. I did that with _save_as_in_progress so I could re-use the same function for the normal save_as_in_progress and save_as_in_progress_and_revert_to_original. It would perhaps be better to use a more obviously different name. Or possibly do save_as_in_progress_and_revert_to_original by first doing save_as_in_progress and then getting the original text in another api call.

api/dp-openapi.yaml
```diff
@@ -733,6 +845,63 @@ components:
...
+        saveasdone Checks the page for invalid characters and if good, saves it as done. Else throws an exception.<br><br>
+        savetemp Checks the page for invalid characters and if good, saves it as in progress, returns page text and state. Else throws an exception.<br><br>
+        saverevert Checks the page for invalid characters and if good, saves it as in progress, returns original page text and current state. Else throws an exception.<br><br>
+        return Returns the page to the round.
+      required: true
+      schema:
+        type: string
+
+  requestBodies:
+    TextBody:
```
cpeel, line 895: Consider `page_text` so it mirrors other `page_text_*` parameters.

api/dp-openapi.yaml
```diff
@@ -1020,6 +1189,69 @@ components:
...
+        saved:
+          description: true if the page state is 'temp'
+          type: boolean
+        pagename:
+          description: the filename of the page with extension
+          type: string
+        image_url:
+          description: url of the image of the page
+          type: string
+        pageNum:
```
cpeel, line 1217: I was, unfortunately, inconsistent in how I named properties -- sometime I used an `_` and sometime I didn't. That's on me. But let's not introduce snakeCase as a third. Suggest this be `pagenum` to match `pagename`.

api/dp-openapi.yaml
```diff
@@ -1020,6 +1189,69 @@ components:
...
+          description: page "number" without extension
+          type: string
+        roundInfoArray:
```
cpeel, line 1220: Ditto no snakeCase - consider `round_info_array`.

api/dp-openapi.yaml
```diff
@@ -1020,6 +1189,69 @@ components:
...
+      type: object
+      properties:
+        text:
+          description: the page text. If page state is 'out' from OCR or previous round, else from current round
+          type: string
+        pagestate:
+          description: the page state, for checking continuity. Either out or temp
+          type: string
+        saved:
+          description: true if the page state is 'temp'
```
cpeel, line 1209: Will this not also return `true` if the page was saveAsDone -- that also saves the page.

api/dp-openapi.yaml
cpeel: Why do we want this separate from pagename?

api/dp-openapi.yaml
```diff
@@ -1020,6 +1189,69 @@ components:
...
         type: string
 
+    proof_page_text:
+      type: object
+      properties:
+        text:
+          description: the page text from current round.
+          type: string
+
+    proof_page_data:
```
cpeel, line 1199: Consider calling this `proof_page` to mirror `project_page` -- `data` doesn't really add anything here.

pinc/ProofProject.inc
```diff
@@ -0,0 +1,365 @@
...
+                "pageNum" => pathinfo($this->page_name, PATHINFO_FILENAME),
+                "roundInfoArray" => $round_info_array,
+                "pagename" => $this->page_name,
+                "image_url" => "{$this->project->url}/{$this->page_name}",
+            ];
+    }
+
+    private function get_username_for_round(object $round): string
+    {
+        $res = DPDatabase::query("
```
cpeel, line 28: This needs an `sprintf()` for the `image` parameter to escape it like you do in `get_page_text_array()`.

api/dp-openapi.yaml
```diff
@@ -733,6 +845,63 @@ components:
...
+    projectid:
+      name: projectid
+      in: path
+      description: ID of project
+      required: true
+      schema:
+        type: string
+
+    projectstate:
+      name: projstate
```
cpeel, line 858: In the `project` API this is `state` and I suggest we keep that for consistency.

api/v1_projects.inc
```diff
@@ -656,3 +657,47 @@ function _get_enabled_filter($query_params)
...
+
+function api_v1_project_validatetext($method, $data, array $query_params)
+{
+    $text_checker = new TextChecker($data[":projectid"]);
+    return $text_checker->analyse();
+}
+
+function api_v1_project_page($method, $data, $query_params)
+{
+    $proof_project_page = new ProofProjectPage($data[":projectid"], $data[":pagename"], $query_params);
```
cpeel, line 679: We should not pass opaque arrays down into other functions, particularly class constructors. It's the API's responsibility as the front-line to the user to validate API input are valid values -- they could be the _wrong_ values (it's a valid state but the wrong state) and the underlying code could decide it's wrong and raise something, but the API should sanitize user data.

pinc/ProofProject.inc
```diff
@@ -0,0 +1,365 @@
...
+        $expected_page_state = $query_params['pagestate'] ?? null;
+        if (null === $expected_page_state) {
+            throw new ProjectNoPageStateGivenException("No expected page state found in request.");
+        }
+        if (!in_array($expected_page_state, $PAGE_STATES_IN_ORDER)) {
+            throw new ProjectInvalidPageStateException(sprintf("%s is not a valid page state", $expected_page_state));
+        }
+
+        $user_column = $this->round->user_column_name;
+        $sql = "
```
cpeel, line 142: Needs `sprintf()` escaping.

api/dp-openapi.yaml
```diff
@@ -678,6 +678,118 @@ paths:
...
+        '429':
+          $ref: '#/components/responses/RateLimitExceeded'
+        default:
+          $ref: '#/components/responses/UnexpectedError'
+
+  /projects/{projectid}/validatetext:
+    put:
+      tags:
+      - project
+      description: Checks the given text for bad characters, returns data indicating them
```
cpeel, line 711:
```suggestion
      description: Checks the given text for invalid characters, returns data indicating them
```
I thought this was WordCheck, not the unicode checker at first, `invalid` might be slightly clearer?

pinc/ProofProject.inc
```diff
@@ -0,0 +1,365 @@
...
+            if (1 != preg_match("/$this->pattern_string/u", $grapheme)) {
+                throw new ProjectBadTextException(_("The text contains characters which are not allowed for this project"));
+            }
+        }
+        return _normalize_page_text($this->page_text, $this->project->projectid);
+    }
+
+    private function receive_text(): void
+    {
+        $request_data = api_get_request_body();
```
cpeel, line 359: This needs to be in the api and passed into the class. This is effectively user input which should always be handled by the top-most entity to ensure components, like this class, are reusable.

cpeel (commented):
This is looking great. Lots of feedback but most if it in the weeds stuff.

api/dp-openapi.yaml
```diff
@@ -1020,6 +1189,69 @@ components:
...
+    page_text_state:
+      type: object
+      properties:
+        text:
+          description: the page text, current or original.
+          type: string
+        pagestate:
+          description: the page state, it is now temp
+          type: string
+        saved:
```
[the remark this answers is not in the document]
70ray: Yes, this needs more explanation. Although we were at one time considering removing the page_out_state, it could be useful to keep it in case it could be used in the future in connection with checking out several pages to distinguish the ones that haven't yet been looked at. This version stays close to the original and, when checking out a page, does not change the current page text.
If we have a button for 'Revert to last save' then it is disabled if the page hasn't yet been saved (as it works now). That is the only purpose of this 'saved' value.

api/dp-openapi.yaml
```diff
@@ -1020,6 +1189,69 @@ components:
...
+            type: string
+
+    page_text_state:
+      type: object
+      properties:
+        text:
+          description: the page text, current or original.
+          type: string
+        pagestate:
+          description: the page state, it is now temp
```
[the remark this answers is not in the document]
70ray: The Swagger file doesn't tie up the responses to the different functions. 'page_text_state' is returned by 'save_as_in_progress' or 'save_and_revert_to_original' so the page state is the 'temp' state and 'saved' is true in these cases.

api/dp-openapi.yaml
```diff
@@ -1020,6 +1189,69 @@ components:
...
+          description: now true
+          type: boolean
+          enum: [true]
+
+    text_analysis:
+      type: object
+      properties:
+        good:
+          type: boolean
+          enum: [false]
```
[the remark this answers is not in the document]
70ray: No this doesn't make any sense. I think it was left over from a previous version that returned this from 'save' if there were any invalid characters, but it seems better to check the text as a separate endpoint if 'save' fails.

api/dp-openapi.yaml
```diff
@@ -1020,6 +1189,69 @@ components:
...
+        text:
+          description: the page text, current or original.
+          type: string
+        pagestate:
+          description: the page state, it is now temp
+          type: string
+        saved:
+          description: now true
+          type: boolean
+          enum: [true]
```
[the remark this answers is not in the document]
70ray: We could leave it out. The same parameter is returned in proof_page_data where it could be true or false.

api/dp-openapi.yaml
```diff
@@ -733,6 +845,63 @@ components:
...
+    pagestate:
+      name: pagestate
+      in: query
+      description: page state
+      required: true
+      schema:
+        type: string
+
+    pagefunction:
+      name: pagefunction
```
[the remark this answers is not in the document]
70ray: I'll do that.

70ray:
> This is looking great. Lots of feedback but most if it in the weeds stuff.

Thanks Casey, working on it ...

api/dp-openapi.yaml
70ray: done this

api/dp-openapi.yaml
70ray: changed to pagenum. This is how the existing PI shows the page number (in the text below the buttons in standard interface) Should we keep it like this?

api/dp-openapi.yaml
70ray: done this.

api/dp-openapi.yaml
70ray: It doesn't return anything for saved as done because the PI then goes back to the project page so anything returned would be discarded. But we could return something since the PI doesn't have to work like that.

api/dp-openapi.yaml
70ray: done this

api/dp-openapi.yaml
70ray: done this

pinc/ProofProject.inc
70ray: done this

pinc/ProofProject.inc
70ray: done this

api/v1_projects.inc
70ray: I am inclined to move the project state and page state into the URI. I'll do this in a separate commit.

cpeel: I think having them in the request body is the right place. Having a resource locator that looks like `v1/projects/:projectid/pages/:pagename/state/:state/` where there is only one legal value for state doesn't really make any sense -- ditto project state. We should just pull them out in the api code and not here.

api/dp-openapi.yaml
70ray: made some changes to clarify this

api/v1_projects.inc
70ray: I see what you mean. I am quite happy to make it work however you like.
One could make sense of this by saying that the project can be in different states at different times and making a request for a resource which is a project in a given state will either succeed or fail in the sense that the resource does not exist.

cpeel: I think your URL schema for this is good -- I would just pull all of the ProofProjectPage constructor's parameters out of $query_params here in the API rather than in the class itself.

70ray: I'll do that.

pinc/ProofProject.inc
70ray: done this

api/v1_projects.inc
```diff
@@ -656,3 +657,85 @@ function _get_enabled_filter($query_params)
...
+
+function get_page_state($query_params)
+{
+    global $PAGE_STATES_IN_ORDER;
+
+    $page_state = $query_params['pagestate'] ?? null;
+    if (null === $page_state) {
+        throw new ProjectNoPageStateGivenException("No page state found in request.");
+    }
+    if (!in_array($page_state, $PAGE_STATES_IN_ORDER)) {
```
70ray, line 727: This also seems pointless.

api/dp-openapi.yaml
cpeel: If the page has been saved this should be set to `true` regardless of what the PI currently does or doesn't do. We need to think about a general endpoint user and not just the current PI.

pinc/User.inc
```diff
@@ -2,6 +2,7 @@
 include_once($relPath.'misc.inc');
 include_once($relPath.'access_log.inc');
 include_once($relPath.'SettingsClass.inc');
+include_once($relPath.'../api/exceptions.inc');
```
cpeel, line 5: I'm not loving how the API exceptions are being pulled into `Project.inc` and `User.inc`. The API should use those classes / files not the other way around. I need to ponder this some more because I see _why_ you did it.

pinc/ProofProject.inc
```diff
@@ -0,0 +1,354 @@
...
+        _Page_UPDATE($this->projectid, $this->page_name, $setters);
+
+        _log_page_event($this->projectid, $this->page_name, 'checkout', $pguser, $this->round);
+        _project_adjust_n_available_pages($this->projectid, -1);
+
+        return $this->render_page_data();
+    }
+}
+
+class ProofProjectPage extends ProofProject
```
cpeel, line 120: It looks like this class is duplicating a lot of the logic in DPage. Do you envision removing DPage after we have an entirely API-based UI? What's our plan for supporting two separate sets of code that do the same thing? I thought one of the reasons for 40b07f0801 was to use the DPage functions directly?

api/dp-openapi.yaml
```diff
@@ -733,6 +845,63 @@ components:
...
+      required: true
+      schema:
+        type: string
+
+    pageaction:
+      name: pageaction
+      in: query
+      description: action to perform on the page.<br><br>
+        resume If the page is saved-as-done 'unsaves' it, gets data for the given page.<br><br>
+        saveasdone Checks the page for invalid characters and if valid, saves it as done. Else throws an exception.<br><br>
```
cpeel, line 886: I would try to hide some of our internal workings from the API client. Consider the following verbs instead (same order as here):
```
checkout - Puts the saved page in a checked-out state for changing
checkin - Saves the page as completed
save - Saves the page and keeps it checked out
revert - Reverts to the original page text
abandon - Returns the page to the round
```

These mirror the current actions but hides the confusing complexity. `checkout` (ie `resume`) pairs with `return` and mirrors how the API got the page to begin with. You can `save` something but until you `checkin` or `abandon` it it's still checked out to you. This way we avoid the confusing `temp` descriptions here -- it's just "checked out" to the user.

Also no need to include "Else throws an exception" as that's a given for any non-success case.

cpeel (commented):
Very happy with the OpenAPI schema overall. 🎉

api/v1_projects.inc
```diff
@@ -656,3 +657,85 @@ function _get_enabled_filter($query_params)
...
+
+function get_project_state($query_params)
+{
+    global $PROJECT_STATES_IN_ORDER;
+
+    $state = $query_params["state"] ?? null;
+    if (null === $state) {
+        throw new ProjectNoStateGivenException("No project state found in request.");
+    }
+    if (!in_array($state, $PROJECT_STATES_IN_ORDER)) {
```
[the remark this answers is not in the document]
70ray: That looks like a good way do do it.
I was about to reply to my comment that if the validation against all possible states is removed then the validation against what the project state actually is should be moved up to this level so we don't pass un-validated user input down to the lower function.
An objection to validating against all possible states is that you have to generate the array of all possible states, which we do anyway now, but it seems like a lot of unnecessary work.

api/dp-openapi.yaml
70ray: I'll do that.

pinc/ProofProject.inc
70ray: Earlier versions I made were more different from DPage than this. The only remaining reason for not using the DPage functions is that they use the _Page_require() function which we don't need because we have already validated the data. It could be better to use the DPage functions directly and then remove the _Page_require some time in the future. I'll look at making it so.

api/v1_projects.inc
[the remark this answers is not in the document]
70ray: Done this

api/dp-openapi.yaml
70ray: Are you saying we should return all this data if the page is saved as done?

api/v1_projects.inc
70ray: Done this now.

api/v1_projects.inc
70ray: Done this.

api/v1_projects.inc
```diff
@@ -656,3 +657,85 @@ function _get_enabled_filter($query_params)
...
+    if (null === $page_state) {
+        throw new ProjectNoPageStateGivenException("No page state found in request.");
+    }
+    if (!in_array($page_state, $PAGE_STATES_IN_ORDER)) {
+        throw new ProjectInvalidPageStateException(sprintf("%s is not a valid page state", $page_state));
+    }
+    return $page_state;
+}
+
+function receive_text(): string
```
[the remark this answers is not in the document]
70ray: done these.

api/v1_projects.inc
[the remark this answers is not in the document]
70ray: I'm quite happy with this but still think it is not entirely logical. We alreay know what the project state is at this point so we could validate it here in a validator. In our case there are only 35 different project states and we already have a list of them so it is not a big deal. (Whether we need have created this list is a different question).

api/v1_projects.inc
70ray: Done this. There are, I think, only 25 possible page states.
We could go further and validate against the actual page state. Consider that when we validate the page name we check it against the array of all the pages in the project. But we could instead search for it in the projectid table and create a Page object containing its name and state and user for the round since we now know the round. We could then validate also that the logged-in user owns the page. This parallels what we do for projects: rather than get a list of all projects and check if it is in the list, we search for it by id and create an object containing relevant data.

api/v1_projects.inc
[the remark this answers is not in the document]
70ray: Done this

api/dp-openapi.yaml
```diff
@@ -733,6 +845,63 @@ components:
       in: header
       name: X-API-KEY
 
+  parameters:
+    projectid:
```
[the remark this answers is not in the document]
70ray: done this

api/dp-openapi.yaml
```diff
@@ -733,6 +845,63 @@ components:
...
+
+    pagestate:
+      name: pagestate
+      in: query
+      description: page state
+      required: true
+      schema:
+        type: string
+
+    pageaction:
```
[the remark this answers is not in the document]
70ray: done this

api/dp-openapi.yaml
70ray: done these.

70ray: I've left resume as it is bedause checkout could be confusing since we already have a checkout function to get a new page. Perhaps we could change the word for that function and use checkout and checkin  for a given page.

pinc/ProofProject.inc
70ray: One possible approach would be to split up some of the DPage functions so that we could use the parts we needed in here.

pinc/ProofProject.inc
```diff
@@ -0,0 +1,354 @@
...
+                _('Warning: Project "%1$s" is not in state "%2$s"; it is now in state "%3$s".'),
+                $project->nameofwork,
+                project_states_text($state),
+                project_states_text($project->state)
+            );
+            throw new ProjectNotInRequiredStateException($err);
+        }
+
+        $project_round = $project->get_project_available_round();
+        $project_round->validate_user_can_access($pguser);
```
70ray, line 91: Why not put all this validation in a validator?

pinc/ProofProject.inc
```diff
@@ -0,0 +1,354 @@
...
+                $page_name,
+                $page_state,
+                $current_page_state
+            );
+            throw new ProjectPageInconsistentStateException($err);
+        }
+
+        if ($row[$user_column] != $pguser) {
+            throw new PageNotOwnedException(_("You do not own this page."));
+        }
```
70ray, line 155: The preceding validation could also go in a validator.

pinc/ProofProject.inc
```diff
@@ -0,0 +1,354 @@
...
+
+        $this->page_name = $page_name;
+        $this->page_state = $current_page_state;
+    }
+
+    public function return_to_round(): void
+    {
+        global $pguser;
+
+        $this->validate_page(false, "returnToRound");
```
70ray, line 165: This should not go in a validator because it depends on the page action, but we could split out the relevant piece of the DPage `_Page_require` function to do this.

pinc/ProofProject.inc
```diff
@@ -0,0 +1,354 @@
...
+        $setters = join(", ", [
+            set_col_str("state", $this->round->page_avail_state),
+            set_col_num($this->round->time_column_name, time()),
+        ]);
+        _Page_UPDATE($this->projectid, $this->page_name, $setters);
+        _log_page_event($this->projectid, $this->page_name, 'returnToRound', $pguser, $this->round);
+        _project_adjust_n_available_pages($this->projectid, +1);
```
70ray, line 172: This could be done as a piece split out of DPage. And several other similar places.

pinc/User.inc
70ray: Yes, it would involve lots of throwing and catching.

70ray (commented):
I'll try changing to a different `validate_page_name` function that returns some data that we can use. If we don't like it we can revert it.

pinc/ProofProject.inc
70ray: > Why not put all this validation in a validator?

If you are just suggesting putting this in a separate function somewhere else, what would be the point? If you were to define a new endpoint component that required this validation, such as `proofable_project` then you could create this ProofProject object in a validator. It's another way of doing it that would probably work just as well.

pinc/ProofProject.inc
70ray: > The preceding validation could also go in a validator.

Again, it wouldn't acieve anything.

pinc/ProofProject.inc
70ray: Or maybe use LPage directly. Perhaps change it a bit if necessary.

70ray:
> It is unfortunate that in the Swagger system "there is no way to link specific schemas to certain parameter combinations".

We have to specify exactly what data each action returns. Will a description in the Swagger file suffice?

api/exceptions.inc
```diff
@@ -81,6 +81,19 @@ class InvalidValue extends ApiException
     }
 }
 
+class ForbiddenError extends ApiException
```
70ray, line 84: This should be added to the Swagger file in cases where it can occur.

pinc/Project.inc
```diff
@@ -11,6 +11,7 @@ include_once($relPath.'special_colors.inc');
 include_once($relPath.'wordcheck_engine.inc'); // delete_project_wordcheck_events()
 include_once($relPath.'DPage.inc'); // project_allow_pages()
 include_once($relPath.'send_mail.inc');
+include_once($relPath.'../api/exceptions.inc');
```
70ray, line 14: This line can be removed now.

pinc/User.inc
```diff
@@ -1,6 +1,7 @@
 <?php
 include_once($relPath.'access_log.inc');
 include_once($relPath.'Settings.inc');
+include_once($relPath.'../api/exceptions.inc');
```
70ray, line 4: ditto remove this line.

70ray:
The responses for 'abandon' and 'checkin' could perhaps be the page state rather than nothing. I haven't changed this yet.

cpeel:
@70ray, let me know when the dust settles on this and you're like for me to look at it again!

70ray:
Sandbox updated.

70ray:
Resolved conflicts.
This still returns nothing after successfully saving as done (checkin) or return to round (abandon). I'll add another commit to make it return just the page state.

Sandbox updated.

70ray:
rebased and conflicts resolved.

pinc/ProofProject.inc
```diff
@@ -0,0 +1,217 @@
...
+        return $this->render_page_data();
+    }
+
+    public function render_page_text(): array
+    {
+        // validate user and page? page should be in out or temp state only
+        return $this->get_page_text_array();
+    }
+
+    private function render_page_data(): array
```
70ray, line 109: This duplicates some of the function of PPage.inc. Why not make a function in LPage to get the data which we can also use in the api and use it in PPage to echo the data?

70ray: Or make a separate PR to do that.

pinc/ProofProject.inc
```diff
@@ -0,0 +1,217 @@
...
+                );
+                throw new DailyLimitExceededException($sentence);
+            } else {
+                $this->save_as_done();
+                if ($pre_save_dpl_count + 1 >= $this->round->daily_page_limit) {
+                    $sentence = sprintf(
+                        _("Your page has been saved as 'Done'. However, you have now reached the daily page limit for %s."),
+                        $this->round->id
+                    );
+                    throw new DailyLimitReachedException($sentence);
```
cpeel, line 48: We should probably think of something besides throwing an exception for this, because an exception is going to return a non-200 message to the client which by definition is a failure. Instead of throwing an exception this function (and endpoint) should probably return some structure that the client can parse to figure this out.

pinc/ProofProject.inc
```diff
@@ -0,0 +1,217 @@
...
+
+class TextChecker
+{
+    private const VALID_TEXT = 0;
+    private const INVALID_TEXT = 1;
+
+    private static $text_array;
+    private static $valid_text_block;
+
+    public static function analyse(string $page_text, object $project): array
```
cpeel, line 187: Consider making this a non-static class. It feels odd that this class is collecting & persisting data statically.

api/dp-openapi.yaml
```diff
@@ -917,6 +917,137 @@ paths:
         default:
           $ref: '#/components/responses/UnexpectedError'
 
+  /projects/{projectid}/checkout:
+    put:
+      tags:
+      - project
+      description: Check out a page for proofreading. Response - schema - proof_page.
```
cpeel, line 924: `Response - schema - proof_page.`?

api/dp-openapi.yaml
```diff
@@ -917,6 +917,137 @@ paths:
...
+      parameters:
+      - $ref: '#/components/parameters/projectid'
+      - $ref: '#/components/parameters/state_query'
+      - $ref: '#/components/parameters/pagename'
+      - $ref: '#/components/parameters/pagestate_query'
+      - name: pageaction_query
+        in: query
+        description: Action to perform on the page.
+            <br><br>
+            resume - If the page is saved-as-done 'unsaves' it. Response - schema - proof_page.
```
cpeel, line 989:
```suggestion
            resume - If the page is saved-as-done 'unsaves' it. Response: proof_page object.
```

api/dp-openapi.yaml
```diff
@@ -917,6 +917,137 @@ paths:
...
+            <br><br>
+            checkin - Checks the page for invalid characters and if valid, saves it as done. Response - none.
```
cpeel, line 991:
```suggestion
            checkin - Checks the page for invalid characters and if valid, saves it as done. Response: none.
```

api/dp-openapi.yaml
```diff
@@ -917,6 +917,137 @@ paths:
...
+            <br><br>
+            save - Checks the page for invalid characters and if valid, saves it as in progress. Response - schema - page_text_state.
```
cpeel, line 993:
```suggestion
            save - Checks the page for invalid characters and if valid, saves it as in progress. Response: page_text_state object.
```

api/dp-openapi.yaml
```diff
@@ -917,6 +917,137 @@ paths:
...
+            <br><br>
+            revert - Checks the page for invalid characters and if valid, saves it as in progress. Response - schema - page_text_state.
```
cpeel, line 995:
```suggestion
            revert - Checks the page for invalid characters and if valid, saves it as in progress. Response: page_text_state object.
```

api/dp-openapi.yaml
```diff
@@ -917,6 +917,137 @@ paths:
...
+            <br><br>
+            abandon - Returns the page to the round, response - none.
```
cpeel, line 997:
```suggestion
            abandon - Returns the page to the round. Response: none.
```

api/dp-openapi.yaml
```diff
@@ -1002,6 +1133,50 @@ components:
...
+    projectid:
+      name: projectid
+      in: path
+      description: ID of project
+      required: true
+      schema:
+        type: string
+
+    state_query:
+      name: state_query
```
cpeel, line 1146: This name needs to match what you use in your code which is `state`.

api/dp-openapi.yaml
```diff
@@ -1002,6 +1133,50 @@ components:
...
+    pagename:
+      name: pagename
+      in: path
+      description: Name of page
+      required: true
+      schema:
+        type: string
+
+    pagestate_query:
+      name: pagestate_query
```
cpeel, line 1162: Ditto this needs to match the code, so `pagestate`.

pinc/ProofProject.inc
```diff
@@ -0,0 +1,217 @@
...
+        }
+        return
+            $this->get_page_text_array() +
+            $this->get_page_state_array() +
+            [
+                "pagenum" => pathinfo($this->page_name, PATHINFO_FILENAME),
+                "round_info_array" => $round_info_array,
+                "pagename" => $this->page_name,
+                "image_url" => "{$this->project->url}/{$this->page_name}",
+                "langDir" => lang_direction(langcode2_for_langname($this->project->languages[0])),
```
cpeel, line 125: This isn't included in the OpenAPI spec. If we want to include it we should call it `language_direction`.

cpeel (commented):
Structurally this is looking really good. I like how we're inheriting from LPage and calling DPage functions 👍 

Looks like this will need to be rebased or have master merged into it to replace `$PAGE_STATES_IN_ORDER`.

pinc/Project.inc
```diff
@@ -1801,6 +1882,23 @@ class Project
             $this->delete_file($dir);
         }
     }
+
+    public function get_project_page($pagename)
+    {
+        $sql = "
+            SELECT state
+            FROM $this->projectid
+            WHERE image='$pagename'
```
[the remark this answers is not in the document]
70ray: Done this.

api/v1_validators.inc
```diff
@@ -72,3 +68,54 @@ function validate_release_queue($queueid, $data)
     }
     return $queue_data;
 }
+
+function validate_project_state($state, $project)
```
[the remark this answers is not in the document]
70ray: These validators are used to validate query parameters rather than url components so perhaps they should be moved somewhere else.

api/v1_validators.inc
```diff
@@ -72,3 +68,54 @@ function validate_release_queue($queueid, $data)
...
+                project_states_text($project->state)
+            );
+            throw new ProjectNotInRequiredStateException($err);
+        }
+    } catch (ProjectException $exception) {
+        throw new NotFoundError($exception->getMessage(), $exception->getCode());
+    }
+}
+
+function validate_page_state($page_state, $project_page)
```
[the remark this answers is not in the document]
70ray: moved the validator.

pinc/Project.inc
```diff
@@ -123,6 +123,87 @@ class ProjectNoMorePagesException extends ProjectException
...
+
+class ProjectNotInRequiredStateException extends ProjectException
+{
+    public function __construct($message, $code = 120)
+    {
+        parent::__construct($message, $code);
+    }
+}
+
+class ProjectInvalidStateException extends ProjectException
```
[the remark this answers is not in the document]
70ray: Good point. I'll do that.

pinc/ProofProject.inc
```diff
@@ -0,0 +1,217 @@
+<?php
+declare(strict_types=1);
+
+class ProofProjectPage extends LPage
+{
+    public function __construct(object $proof_project, object $project_page)
+    {
+        global $pguser;
```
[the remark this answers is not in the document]
70ray: Removed this.

pinc/ProofProject.inc
70ray: Yes, I that is probably best.

pinc/ProofProject.inc
```diff
@@ -0,0 +1,217 @@
...
+
+        $this->page_state = $project_page->page_state;
+        $this->page_name = $project_page->page_name;
+        $this->project = $proof_project->project;
+        $this->projectid = $proof_project->project->projectid;
+        $this->round = $proof_project->round;
```
70ray, line 14: There is also a round variable inside Lpage which is derived from the page_state. Although they will be the same this is not good. Perhaps we should consider revising LPage so that the constructor receives a round (which is easily avaiable in the old code derived from project_state)

pinc/ProofProject.inc
70ray: Yes I can do that.

pinc/ProofProject.inc
70ray: We could change the procedure so that the client has to `save` then `checkin`. `checkin` would return a status code and a message or throw an exception if already over the limit.

cpeel: I don't think we need to make it multi-step. We could have the API call catch these two exceptions and return something in the 200 response to tell the client.

70ray: Or throw an exception only in the over-the-limit case. Then the client application could detect this and do a save (as in progress) which would return the normalised text as usual for such saves. The client would be behaving in a similar way to PPage.

api/dp-openapi.yaml
70ray: Removed this since no amiguity in this case.

api/dp-openapi.yaml
70ray: Unfortunately the colon upsets the swagger editor.

api/dp-openapi.yaml
70ray: this has now changed.

api/dp-openapi.yaml
70ray: done.

api/dp-openapi.yaml
70ray: done.

api/dp-openapi.yaml
70ray: done this.

api/dp-openapi.yaml
70ray: also done.

pinc/ProofProject.inc
70ray: Included this. It was in the old code.

70ray:
More changes involving changes to LPage to come later.

70ray:
LPage gets the Round from the page state. The database has some redundancy in that the project state also implies the round. We assume these can never become conficted.

70ray:
Sandbox updated.

70ray:
Sandbox updated.

api/v1_projects.inc
```diff
@@ -752,3 +753,116 @@ function _get_enabled_filter($query_params)
...
+                _('Warning: Page  "%1$s" is no longer in state "%2$s"; it is now in state "%3$s".'),
+                $project_page->page_name,
+                $page_state,
+                $project_page->page_state
+            );
+            throw new ProjectPageInconsistentStateException($err);
+        }
+
+        $proof_project_page = new ProofProjectPage($proof_project, $project_page);
+        if ($method == "GET") {
```
cpeel, line 805: From a usability perspective, the GET shouldn't need to pass any any of the state information to get the page text. And it should probably return the page state in the return -- probably the `page_text_state` object, actually.

api/dp-openapi.yaml
```diff
@@ -924,6 +924,138 @@ paths:
...
+    put:
+      tags:
+      - project
+      description: Various operations on the page as determined by the 'pageaction_query' parameter
+      parameters:
+      - $ref: '#/components/parameters/projectid'
+      - $ref: '#/components/parameters/state_query'
+      - $ref: '#/components/parameters/pagename'
+      - $ref: '#/components/parameters/pagestate_query'
+      - name: pageaction_query
```
cpeel, line 992: This needs to match what's in the code for the argument name.
```suggestion
      - name: pageaction
```

api/dp-openapi.yaml
```diff
@@ -924,6 +924,138 @@ paths:
...
+        '429':
+          $ref: '#/components/responses/RateLimitExceeded'
+        default:
+          $ref: '#/components/responses/UnexpectedError'
+
+  /projects/{projectid}/pages/{pagename}:
+    put:
+      tags:
+      - project
+      description: Various operations on the page as determined by the 'pageaction_query' parameter
```
cpeel, line 986:
```suggestion
      description: Various operations on the page as determined by the 'pageaction' parameter
```

api/dp-openapi.yaml
```diff
@@ -924,6 +924,138 @@ paths:
...
+      parameters:
+      - $ref: '#/components/parameters/projectid'
+      - $ref: '#/components/parameters/state_query'
+      - $ref: '#/components/parameters/pagename'
+      - $ref: '#/components/parameters/pagestate_query'
+      - name: pageaction_query
+        in: query
+        description: Action to perform on the page.
+            <br><br>
+            resume - If the page is saved-as-done 'unsaves' it. Response:- proof_page object.
```
cpeel, line 996: I can't think of a better description, but one of the use-cases for this is for the API user to get the page-level details (current text, image name, etc) for a "temp" page too.

cpeel (commented):
First of all, this is super cool. It's really amazing to be able to manipulate page information via the API.

After using this some, I think we need to relax two of the state checks specifically:
* GETting a page shouldn't require passing in any state information -- project or page (more in this in the code-level review)
* PUTing a page shouldn't require the project state, just the page state.

The above allows an API user to GET the page (which includes the state) and using only that information, PUT the page back with an action.

---

We're getting really close. If we can, let's try to focus on just the issues raised in the PR so we can get this locked in and merged!

api/dp-openapi.yaml
70ray: done this.

api/dp-openapi.yaml
70ray: ditto.

api/dp-openapi.yaml
70ray: Put in a longer description.

api/v1_projects.inc
70ray: This function can be used for the 'revert to last save' in the enhanced PI. If the PM changes the project to unavailable while a user has a page open in his PI then at present if he does a 'revert to last save'  he gets a warning that the project has changed state and is unable to proceed. Do we need a different function to get the text etc. for non-proofreading purposes?

api/v1_projects.inc
```diff
@@ -752,3 +753,116 @@ function _get_enabled_filter($query_params)
...
+    try {
+        if (null === $state) {
+            throw new InvalidValue("No project state found in request.");
+        }
+        if (!in_array($state, ProjectStates::get_states())) {
+            throw new InvalidValue("Invalid project state");
+        }
+        if ($state != $project->state) {
+            $err = sprintf(
+                _('Warning: Project "%1$s" is not in state "%2$s"; it is now in state "%3$s".'),
```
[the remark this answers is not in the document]
70ray: Done this.
There are two similar strings in LPage.inc in get_indicated_LPage() and project_continuity_test(). Should we change them also?

70ray (commented):
I'm not sure about removing the state validations. I've put a comment about the 'GET' action. 'PUT' can have a similar case when the project is changed to unavailable.

api/v1_projects.inc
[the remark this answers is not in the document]
cpeel: I think we should change the one in `project_continuity_test()` as it is project-state related (and let's remove the `\n` prefix before that string too). I'd leave the one in `get_indicated_LPage()` as-is since it's page-state related.

api/v1_projects.inc
70ray: I see what you mean about the 'GET'. Let me think about that.

70ray: In fact we don't need that GET function at all. It was intended to be used for the 'revert to last save' in the enhanced PI, but you can get the same effect in the standard PI by 'stop proofreading' and resuming the page so we can use the resume_page PUT function instead.
That leaves the question of the PUT functions. They can all change the database in some way so I think we need to check that the project is available as you said.

cpeel: No no, we need the GET. Because a GET is the only way we have to get the state of a page that we need to do anything with it. Without the GET we are blind to how to interact with the page.

70ray: It seems like you are envisaging a different use  for the GET function than I had anticipated. I may be going about it the wrong way but the following is the workflow I am considering. (using the new funcion names)

Checkout a page. Returns all page data or an exception if it can't be done.

Save. (formerly as_in_progress) returns the text (could be more normalised than what was submitted) and page state which could have changed. Throws exception if it contains invalid characters.

Revert. Saves submitted text and returns original. (This supports the coresponding function in enhanced PI. The standard PI can't do this)

Checkin. Saves page as done.

Abandon. Return page to the round.

Resume. Returns all page data. If the page was in the saved_as_done state, changes it to 'temp' and makes appropriate changes to the database.

cpeel: That workflow works great if it's done all in one "session". But consider the case where the API client:
1. checks out page (gets 001)
2. saves page 001
3. checks out another page (gets 002)
4. saves page 002
5. checks in page 002

If I'm the client in order to do something with 001 I have to keep track of what the last state of the page is (and the text itself). But if I have a GET function I could just get the pagetext and  state and then save it or check it in or whatever.

Put another way, if we have the GET function for a page, then we're just one more endpoint away (one that returns all of the pages the user has worked on in a project in the current round) from the client not needing to retain any state beyond a page it's currently working with.
1. client GETs list of all pages the user has worked on in the project in the current round (new endpoint, we should not add it to this PR)
2. client can GET a page they want to do something with (checkout, revert, save, whatever)
3. client now has the data it needs to do does something with that page
4. client does some action with that page (checkout, revert, save, etc)
5. lather, rinse, repeat

70ray: I'd only considered it as if interfacing from the existing project page where you either checkout a page or access a page you have worked on before. If you want to go further back then, as you say, more is required. I'll consider this.

70ray: The new endpoint could return a list of the pages the user has worked on together with their states and timestamps. Similar to the code used to make the lists of pages the user has worked on in the project page. The client could make similar lists and the user could select a page and do what he wanted using the existing endpoints.

For this new endpoint the question arises of how the client would understand the page state? Should we also supply a more concise string than the one used in the database, perhaps just "done", "temp" or "out" (which are the only relevant ones here and the endpoint would have explicitly selected)

cpeel: I think I see the disconnect. We aren't designing the API to replicate the UX we already have. We're designing an API that can be used as a data layer primarily, but not exclusively for, a new proofreading interface. The API should "stand on its own" in a sense.

I'm not at all prescribing what we should present to the user, etc -- more that as an API consumer I want to be able to use these new endpoints stand-alone as a data transport. We should 100% keep the needs of a future API as our primary design focus, but we should also make sure the API endpoints are useful outside of that.

For "list all pages that I worked on in this project" endpoint, it should return the same set of data that the client then needs to work with the PUT endpoint -- so the exact same state name.

cpeel: But let's not add that new endpoint to this PR. Let's focus on wrapping up the endpoints you have.

70ray: I'll resolve the conflicts and push it again.

cpeel: Any progress on adding the GET endpoint back? I'd been waiting on that to re-review this.

cpeel: You are correct, it did not do what I wanted but we still need it. It needs to not require passing in any state (project or page) and it should return a `page_text_state` object so that it includes the page text and the current page state. We need a GET to get the current state of a page so I can do something with it. I can't call resume without the page state.

Separately, and not in this PR, we need an endpoint that will return all pages for a specific user that the API could use to present a user with pages they have done in the past and can work on again.

70ray: I can do that.
I've been maintaining a basic PI based on this branch to show how the functions can be used. I've pushed it up now.

cpeel (commented):
Here's happy-path testing demonstrating all of the new APIs (but not each variation):
```bash
# fill in your API key and select a project you want to proofread
export SANDBOX_URL=https://www.pgdp.org/~rp31/c.branch/proof_api_10
export API_KEY=xxx
export PROJECT=projectID....

# get project details -- we need to know its state to check out a page
curl -s -H"Accept: application/json" -H "X-API-KEY: $API_KEY" \
    -X GET "$SANDBOX_URL/api/v1/projects/$PROJECT" \
    > project.json
PROJECT_STATE=$(cat project.json | jq -r '.state')
echo "Project state: $PROJECT_STATE"

# checkout a page
curl -s -H"Accept: application/json" -H "X-API-KEY: $API_KEY" \
    -X PUT "$SANDBOX_URL/api/v1/projects/$PROJECT/checkout?state=$PROJECT_STATE" \
    > proof_page.json
# pull out the page text into one file and save the page name
PAGE_NAME=$(cat proof_page.json | jq -r .pagename)
PAGE_STATE=$(cat proof_page.json | jq -r .pagestate)
cat proof_page.json | jq '{text}' > page_text.json
echo "Page name: $PAGE_NAME"
echo "Page state: $PAGE_STATE"
echo "Page text (in JSON format):"
cat page_text.json

# Manual: edit the page text in page_text.json if you want

# let's validate the page text doesn't have any weird characters
curl -s -H"Accept: application/json" -H "X-API-KEY: $API_KEY" \
    -X PUT "$SANDBOX_URL/api/v1/projects/$PROJECT/validatetext" \
    -d @page_text.json

# save the page as in-progress
curl -s -H"Accept: application/json" -H "X-API-KEY: $API_KEY" \
    -X PUT "$SANDBOX_URL/api/v1/projects/$PROJECT/pages/$PAGE_NAME?state=$PROJECT_STATE&pagestate=$PAGE_STATE&pageaction=save" \
    -d @page_text.json

# actually, you know I'm done with that page entirely, lets check it in
# but I can't remember what state it is in now, so lets get that
curl -s -H"Accept: application/json" -H "X-API-KEY: $API_KEY" \
    -X GET "$SANDBOX_URL/api/v1/projects/$PROJECT/pages/$PAGE_NAME" \
    > page_info.json
PAGE_STATE=$(cat page_info.json | jq -r .pagestate)
echo "Page state: $PAGE_STATE"

# save it as done
curl -s -H"Accept: application/json" -H "X-API-KEY: $API_KEY" \
    -X PUT "$SANDBOX_URL/api/v1/projects/$PROJECT/pages/$PAGE_NAME?state=$PROJECT_STATE&pagestate=$PAGE_STATE&pageaction=checkin" \
    -d @page_text.json

# get the current page state
curl -s -H"Accept: application/json" -H "X-API-KEY: $API_KEY" \
    -X GET "$SANDBOX_URL/api/v1/projects/$PROJECT/pages/$PAGE_NAME" \
    > page_info.json
PAGE_STATE=$(cat page_info.json | jq -r .pagestate)
echo "Page state: $PAGE_STATE"
```

This makes use of `jq` which is installed on TEST already and is available for Homebrew (and likely Windows I guess).

api/v1_projects.inc
```diff
@@ -770,3 +771,117 @@ function _get_enabled_filter($query_params)
...
+        $page_state = $query_params['pagestate'] ?? null;
+        if (null === $page_state) {
+            throw new InvalidValue("No page state found in request.");
+        }
+        if (!in_array($page_state, Rounds::get_page_states())) {
+            throw new InvalidValue(sprintf("%s is not a valid page state", $page_state));
+        }
+        if ($page_state != $project_page->page_state) {
+            $err = sprintf(
+                _('Page  "%1$s" is not in state "%2$s"; it is now in state "%3$s".'),
```
cpeel, line 826:
```suggestion
                _('Page "%1$s" is not in state "%2$s"; it is now in state "%3$s".'),
```

api/v1_projects.inc
```diff
@@ -770,3 +771,117 @@ function _get_enabled_filter($query_params)
...
+            throw new ProjectPageInconsistentStateException($err);
+        }
+
+        $proof_project_page = new ProofProjectPage($proof_project, $project_page);
+        $page_action = $query_params['pageaction'] ?? null;
+        if (null === $page_action) {
+            throw new BadRequest("No page action found in request.");
+        }
+        switch ($page_action) {
+            case 'resume':
```
cpeel, line 840:
```suggestion
        switch ($page_action) {
            case null:
                throw new BadRequest("No page action found in request.");
            case 'resume':
```

api/dp-openapi.yaml
```diff
@@ -1370,6 +1547,74 @@ components:
...
+          description: URL of the image of the page
+          type: string
+        language_direction:
+          description: left-to-right or right-to-left
+          type: string
+          enum: [RTL, LTR]
+        pagenum:
+          description: Page "number" without extension
+          type: string
+        round_info_array:
```
cpeel, line 1575: To match what you're returning:
```suggestion
        round_info:
```

cpeel (commented):
I've checked that the OpenAPI schema matches what the code is accepting and returning except for one small thing.

cpeel:
I've tested the following negative-paths that all worked as expected
* trying to checkout a page for a project in a round the user does not have access to
* trying to resume a page in a project that did not belong to the user
* trying to manipulate a page without giving the proper state (thank goodness for that GET that I can always get the current state!)

This looks good to me! 🎉

pinc/ProofProject.inc
```diff
@@ -0,0 +1,170 @@
...
+declare(strict_types=1);
+
+class ProofProjectPage extends LPage
+{
+    public function __construct(ProofProject $proof_project, ProjectPage $project_page)
+    {
+        parent::__construct($proof_project->project, $project_page->page_name, $project_page->page_state, 0);
+    }
+
+    public function pp_checkout($user): array
```
cpeel, line 11: Any reason this and `pp_resume_page()` have a `pp_` prefix and the other actions don't? the `pp_` seems superfluous.

cpeel (commented):
There are virtually no non-API changes in this PR (great job, Ray) that require testing in the UI with the possible exception of hacking a "resume" proofreading URL to try and resume a page the user does not own.

And with that except for the super minor things I made comments on, this is ready to go after review by others.

70ray:
> Here's happy-path testing demonstrating all of the new APIs (but not each variation):
>  ...

This is much better than my tests because it tests the whole API not just part of it. Is there any way we can use it in an automated test?

api/v1_projects.inc
70ray: done this.

api/v1_projects.inc
70ray: done this.

pinc/ProofProject.inc
70ray: I originally didn't have these prefixes but at one point I must have added type declarations to the corresponding functions in LPage (returning void). Then phpunit complains that the functions must be compatible with the parent functions so I changed the names. The type declarations in LPage go removed again at some point. Perhaps I should put them back in.

cpeel: Ah, I get it. Let's leave it as-is for now then.

cpeel (approved):
> > Here's happy-path testing demonstrating all of the new APIs (but not each variation):
> > ...
> 
> This is much better than my tests because it tests the whole API not just part of it. Is there any way we can use it in an automated test?

You cover all of these already in your unit tests, I think these are most useful for humans to understand what's going on. Some variation of these probably go in the API user's guide -- I will open up a PR to do that after this merges in.

windymilla (approved):
Apart from my mistypes, etc., this worked fine for me.

The only thing that would have helped me in my incompetence was the error message received when I tried to "PUT" a page back, but with a mistyped JSON fileanme as argument to `-d wrong_name.json`. The message was "Content was not valid JSON", rather than "JSON file not found". There may be nothing that needs doing about it, but I thought I'd mention it.

70ray:
> The only thing that would have helped me in my incompetence was the error message received when I tried to "PUT" a page back, but with a mistyped JSON fileanme as argument to `-d wrong_name.json`. The message was "Content was not valid JSON", rather than "JSON file not found". There may be nothing that needs doing about it, but I thought I'd mention it.

I think this is a problem with 'curl' rather than our software. `-d @filenane` means it gets data from a file to send. The documentation I have been able to find doesn't reveal what curl does if you specify a file that does not exist but whatever it sends is then not valid JSON

pinc/ProofProject.inc
srjfoo: Sorry to be so late to this conversation. I don't feel strongly about changing this for this PR, but I'd _**really**_ like to see it changed at some point. Elsewhere in the code, `pp_` indicates post-processing. If a prefix is necessary, could it be changed to something like `pr_`, `pf_` or `ppp_`?

srjfoo (approved):

pinc/ProofProject.inc
70ray: Good point, this pp_was based on the pp in PPage (presentation page) but someting else would be better.

cpeel: The reason I'm not _too_ worried about this at this time is because eventually we'll be removing the non-API uses of the PPage code and we can update the definitions in PPage to match what we need for the API and remove the `pp_` prefix.

cpeel:
Resolved merge conflict, will merge after CI completes.

PR #979, opened 2023-12-06
Handle deleted/merged related topics

cpeel:
If a task is linked to a forum topic and that topic no longer exists (because it was deleted or because it was merged into another one) we should handle this gracefully. This was discovered in the `php_errors` log on PROD and discussed in Slack.

Testable in https://www.pgdp.org/~cpeel/c/tasks.php?task_id=2300 (specifically task 2300).

PR best reviewed with ignoring whitespace.

To put task 2300 back into this state you can use the following SQL:
```sql
update tasks set related_postings = 'YToxOntpOjA7czozOiI3OTUiO30=' where task_id=2300;
```

### DIFF
--- a/tasks.php
+++ b/tasks.php
@@ -1692,22 +1692,31 @@ function RelatedPostings($tid)
     echo "<table class='themed theme_striped'>\n";
     $related_postings = decode_array($related_postings);
     asort($related_postings);
-    foreach ($related_postings as $val) {
-        $row = get_topic_details($val);
-        $forum_url = get_url_to_view_forum($row["forum_id"]);
-        $topic_url = get_url_to_view_topic($row["topic_id"]);
+    foreach ($related_postings as $topic_id) {
 
         echo "<tr><td>";
-        echo "<a href='$forum_url'>" . $row['forum_name'] . "</a>";
-        echo "&nbsp;&raquo;&nbsp;";
-        echo "<a href='$topic_url'>" . $row['title'] . "</a>";
-        echo " (" . sprintf(_('Posted by: %1$s - %2$d replies'), $row['creator_username'], $row['num_replies']) . ")\n";
+
+        if (does_topic_exist($topic_id)) {
+            $row = get_topic_details($topic_id);
+            $forum_url = get_url_to_view_forum($row["forum_id"]);
+            $topic_url = get_url_to_view_topic($row["topic_id"]);
+            echo "<a href='$forum_url'>" . $row['forum_name'] . "</a>";
+            echo "&nbsp;&raquo;&nbsp;";
+            echo "<a href='$topic_url'>" . $row['title'] . "</a>";
+            echo " (" . sprintf(_('Posted by: %1$s - %2$d replies'), $row['creator_username'], $row['num_replies']) . ")\n";
+        } else {
+            // If the topic doesn't exist, most likely because it was merged
+            // into another topic, we should log an error and still give the
+            // user a way to remove it.
+            error_log("tasks.php - task $tid is linked to non-existent topic $topic_id");
+            echo sprintf(_("Topic %d is linked but no longer exists"), $topic_id);
+        }
 
         echo " ";
         echo "<form action='$tasks_url' method='post' style='display: inline;'>";
         echo "<input type='hidden' name='action' value='remove_related_topic'>";
         echo "<input type='hidden' name='task_id' value='$tid'>";
-        echo "<input type='hidden' name='related_posting' value='$val'>";
+        echo "<input type='hidden' name='related_posting' value='$topic_id'>";
         echo "<input type='submit' value='" . attr_safe(_("Remove")) . "'>\n";
         echo "</form>";
         echo "</td></tr>";


cpeel:
Whoops, I fat-fingered something in here and this isn't working like I thought it was (my dev sandbox was set to `json` forum instead of `phpbb3`). Will get the PR fixed here soon, sorry.

cpeel:
Turns out phpBB lets you delete the only post in a topic but that doesn't actually delete the topic -- which is what I did to test this not realizing the difference. This PR addresses both the case where the topic doesn't exist as well as the edgecase where the topic exists with no posts.

pinc/ProjectSearchResults.inc
```diff
@@ -4,7 +4,7 @@ include_once($relPath.'ProjectTransition.inc');
 include_once($relPath.'forum_interface.inc');
 include_once($relPath.'wordcheck_engine.inc');
 include_once($relPath.'ProjectSearchResultsConfig.inc');
-include_once($relPath.'forum_interface.inc'); // get_topic_details() & get_url_to_view_topic()
+include_once($relPath.'forum_interface.inc'); // get_url_to_view_topic()
```
cpeel, line 7: This change is intentional as `get_topic_details()` is not used in this file so the comment is incorrect.

70ray (approved):
Looks good to me.

srjfoo (approved):
@cpeel, looking at the moderator log, you actually soft deleted that topic. But since soft deleting leaves the door open for someone to come along and restore it, it's not really permanently gone. (Edit: And actually, to everyone except forums admins or global moderators, it looks like soft deleted posts are permanently gone.)

I picked on the next task in order (2301) and added a new task, deleted the only post, and it was reported in the MCP as soft-deleting the topic. I went back and permanently deleted it, and it was recorded as "deleted".

Next, I created another topic, added a second post, then deleted the first post first, and then the second. Both were reported as soft deletes of posts -- no soft deleting of the topic this time (I don't know if that's supposed to be a bug or a feature in phpbb). In the case of your code, though, I don't think that it matters. I think you've covered teh bases that need to be covered.

tasks.php
```diff
@@ -1692,22 +1692,31 @@ function RelatedPostings($tid)
...
-        echo "<a href='$topic_url'>" . $row['title'] . "</a>";
-        echo " (" . sprintf(_('Posted by: %1$s - %2$d replies'), $row['creator_username'], $row['num_replies']) . ")\n";
+
+        if (does_topic_exist($topic_id)) {
+            $row = get_topic_details($topic_id);
+            $forum_url = get_url_to_view_forum($row["forum_id"]);
+            $topic_url = get_url_to_view_topic($row["topic_id"]);
+            echo "<a href='$forum_url'>" . $row['forum_name'] . "</a>";
+            echo "&nbsp;&raquo;&nbsp;";
+            echo "<a href='$topic_url'>" . $row['title'] . "</a>";
```
chrismiceli, line 1705: Do we want to html encode the user provided content (post title) to avoid any XSS/rendering issues?

cpeel: Surprisingly no because phpBB escapes HTML entities upon save, so they are escaped in the database:
```
mysql> select topic_title from phpbb_topics where topic_id=53;
+----------------------------------------------------+
| topic_title                                        |
+----------------------------------------------------+
| &lt;i&gt;i&lt;/i&gt; &amp;amp; ampersand &lt;b&gt; |
+----------------------------------------------------+
1 row in set (0.00 sec)
```
for [topic 53](https://www.pgdp.org/phpBB3/viewtopic.php?f=1&t=53&p=113#p113). Good catch though because my first thought was yes we should!

chrismiceli (approved):

tasks.php
```diff
@@ -1692,22 +1692,31 @@ function RelatedPostings($tid)
...
+            echo " (" . sprintf(_('Posted by: %1$s - %2$d replies'), $row['creator_username'], $row['num_replies']) . ")\n";
+        } else {
+            // If the topic doesn't exist, most likely because it was merged
+            // into another topic, we should log an error and still give the
+            // user a way to remove it.
+            error_log("tasks.php - task $tid is linked to non-existent topic $topic_id");
+            echo sprintf(_("Topic %d is linked but no longer exists"), $topic_id);
```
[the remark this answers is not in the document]
cpeel: Maybe you added one of the soft-deleted IDs instead? I'm pretty sure the code will let you do that right now but will show you the "Topic # is linked but no longer exists" message.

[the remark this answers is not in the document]
srjfoo: > Maybe you added one of the soft-deleted IDs instead? I'm pretty sure the code will let you do that right now but will show you the "Topic # is linked but no longer exists" message.

Sorry it took so long to get back to this -- I wanted to check the database. I confirmed that the most recent topics created on TEST are still several away from 999. Current latest topic ID is 798 according to the database, and I'm pretty sure I haven't added and deleted 200 topics in the last couple of days 😁.

I'm out of ideas at this point. If it weren't for that screenshot, I'd think I had imagined the whole thing.

It is, however, consistently working as it's supposed to, so I think this is good to go.

PR #980, opened 2023-12-08
Fix non-resumable upload

70ray:
While looking into uploading files, I found that the non-resumable upload did not work. The first commit enables it to work. The others tidy up other features.

It is probably not possible to test this without hacking the code to make it appear that the resumable upload is not available. I suppose two sandboxes are needed: one to make sure this does not break the resumable upload and another hacked sandbox to test this. I can make these if anyone wishes.

### DIFF
--- a/pinc/upload_file.inc
+++ b/pinc/upload_file.inc
@@ -108,7 +108,7 @@ function show_upload_form($form_content, $submit_label)
     echo "<div id='old_uploader'>
     <p class='warning'>", sprintf(_('Maximum file size is %s.'), humanize_bytes(get_max_upload_size())), "</p>
     <p>", get_big_upload_blurb(), "</p>",
-    _("File to upload"), ": <input type='file' id='old_browse' accept='.zip' name='uploaded_file'>
+    _("File to upload"), ": <span id='file_name'></span><input type='file' id='old_browse' accept='.zip' name='uploaded_file'>
     <input type='submit' id='old_submit' value='", attr_safe($submit_label), "'>",
     "<p>", sprintf(_("After you click the '%s' button, the browser will appear to be slow getting to the next page. This is because it is uploading the file."), $submit_label), "</p>",
     "</div>";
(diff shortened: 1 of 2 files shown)


srjfoo:
> I suppose two sandboxes are needed: one to make sure this does not break the resumable upload and another hacked sandbox to test this. I can make these if anyone wishes.

I'd feel happier if I could test it, so yes, please.

chrismiceli (approved):
We may one day want to remove the old code for when resumable isn't supported. Last time I tried, I don't think it even worked.

70ray:
The progress indicator is moved in the resumable mode but it should show in both modes now.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/nr_upload (this will use resumable)

Another Sandbox at: https://www.pgdp.org/~rp31/c.branch/no_resume (this will not use resumable upload)

chrismiceli (approved):

srjfoo:
> The progress indicator is moved in the resumable mode but it should show in both modes now.

Just to make sure that I am understanding correctly, the non-resumable doesn't actually give an indication of progress, but just displays an "Uploading, please wait ..." message, whereas the resumable actually shows percentage progress. The key being that they both do display a message of some sort while the file is uploading?

70ray:
> > The progress indicator is moved in the resumable mode but it should show in both modes now.
> 
> Just to make sure that I am understanding correctly, the non-resumable doesn't actually give an indication of progress, but just displays an "Uploading, please wait ..." message, whereas the resumable actually shows percentage progress. The key being that they both do display a message of some sort while the file is uploading?

Yes, previously the progress indicator was above the message "Your browser supports uploading large files ..." and inside the resumable-only div so it did not appear at all for non-resumable. I moved it ouside the div. (At an earlier time it must have been different (probably I broke it a few years ago but no-one noticed because resumable was always available)).

70ray:
A comment says "prevent the user from hitting upload again" but it doesn't actually do this. I'll look into it.

chrismiceli (approved):

70ray:
use #997 instead.

PR #985, opened 2023-12-16
Mark no-wrap and block-quote tags in different colours

70ray:
Mark no-wrap and block-quote tags in different colours. This addresses issue [#2063](https://www.pgdp.net/c/tasks.php?action=show&task_id=2063)

Sandbox at: https://www.pgdp.org/~rp31/c.branch/ool_colour


tools/proofers/previewStyles.js
```diff
@@ -0,0 +1,28 @@
+/* exported defaultStyles */
+
+// these are the default values. If the user changes anything the new
+// styles are saved in local storage and reloaded next time.
+// the foreground and background colours for plain text, italic, bold,
+// gesperrt, smallcaps, font change, other tags, highlighting issues
+// and possible issues.
+// An empty color string means use default color
+var defaultStyles = {
```
chrismiceli, line 9: *minor* we could use `const` here.

chrismiceli (approved):

tools/proofers/previewStyles.js
```diff
@@ -0,0 +1,28 @@
+/* exported defaultStyles */
```
cpeel, line 1: I'd prefer this remain in `previewControl.js`. Every separate JS file requires another round-trip for the browser which decreases site performance.

70ray: I tried that at first but another problem occured. I think I can work round it a different way.

70ray:
I removed the unnecessary addEventListener("DOMContentLoaded") from preview.js. This wasn't really necessary for this purpose but I happened to notice it was not needed. The file can be viewed ignoring white space changes because most of it is unindented.

70ray:
Sandbox updated again.

cpeel (approved):

srjfoo (approved):

PR #986, opened 2023-12-17
check if javascript variables not null

70ray:
On the Preferences page, if using tabs other than proofreading, the font preferences do not appear and javascript exceptions were shown in the developer tools console if it is open.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/prefs_exceptions

### DIFF
--- a/userprefs.php
+++ b/userprefs.php
@@ -166,16 +166,13 @@ $theme_extra_args["js_data"] =
     // font_sample when they change.
     window.addEventListener('DOMContentLoaded', () => {
         for (const orientation of ['v', 'h']) {
-            let radios = document.querySelectorAll('input[name=\"' + orientation + '_fntf\"]');
-            if(radios) {
-                radios.forEach(radio => {
-                    radio.addEventListener('change', function () {
-                        do_font_sample_update(this.value, null, orientation);
-                    });
+            document.querySelectorAll('input[name=\"' + orientation + '_fntf\"]').forEach(radio => {
+                radio.addEventListener('change', function () {
+                    do_font_sample_update(this.value, null, orientation);
                 });
-            }
+            });
             let otherInput = document.querySelector('input[name=\"' + orientation + '_fntf_other\"]');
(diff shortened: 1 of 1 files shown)


userprefs.php
```diff
@@ -166,19 +166,28 @@ function do_font_sample_update(font_index, size_index, layout) {
     // font_sample when they change.
     window.addEventListener('DOMContentLoaded', () => {
         for (const orientation of ['v', 'h']) {
-            document.querySelectorAll('input[name=\"' + orientation + '_fntf\"]').forEach(radio => {
-                radio.addEventListener('change', function () {
-                    do_font_sample_update(this.value, null, orientation);
+            let radios = document.querySelectorAll('input[name=\"' + orientation + '_fntf\"]');
```
chrismiceli, line 169: querySelectorAll shouldn't return a falsy object, just an empty list per https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll. So we shouldn't need this check, right?

userprefs.php
```diff
@@ -166,19 +166,28 @@ function do_font_sample_update(font_index, size_index, layout) {
...
+            if(radios) {
```
chrismiceli, line 170: *minor* other places we put a space after `if`, or `for` if we want to be consistent.

chrismiceli (approved):

userprefs.php
70ray: You're right. I'll remove it.

userprefs.php
70ray: I've changed this. It would be nice to make this a javascript file so we could lint it.

chrismiceli: Yeah. I wish lunters worked better on online scripts :/

70ray:
Sandbox updated.

chrismiceli (approved):

cpeel (approved):
Good catch - thanks for the fix!

srjfoo (approved):

PR #987, opened 2023-12-20
round up smooth read deadline to next hour

70ray:
if starting or extending smooth reading.

This is the first stage of a revised method of sending SR end notifications suggested by @cpeel . It is compatible with the old notification sender.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/sr_roundup

### DIFF
--- a/pinc/smoothread.inc
+++ b/pinc/smoothread.inc
@@ -194,14 +194,24 @@ function handle_smooth_reading_change($project, $postcomments, $days, $extend)
     $projectid = $project->projectid;
 
     if ($days) { // will be 0 if parameter not supplied when only replacing text
-        $seconds = $days * 60 * 60 * 24;
         if ($extend) {
             // extend deadline if not yet passed
-            $deadline = $project->smoothread_deadline + $seconds;
+            $start_time = $project->smoothread_deadline;
         } else {
             // if starting sr with deadline=0, or if sr ended
-            $deadline = time() + $seconds;
+            $start_time = time();
         }
+
+        // Extend the deadline to the next whole hour after $days
+        $datetime = new DateTime("@$start_time");
+        // Add our days
+        $datetime->modify("+{$days} day");
+        // round up to the next hour
+        $datetime = $datetime->modify("next hour");
+        // replace the time with just the hour
+        $datetime->modify($datetime->format('H:00:00'));
+        $deadline = $datetime->format('U');
+
         $details1 = $extend ? "deadline extended" : "text available";
         $smoothread_deadline_sql = "smoothread_deadline = $deadline, ";
         $project->log_project_event($pguser, 'smooth-reading', $details1, $deadline);


cpeel (changes requested):
It would be better if we replaced the original hacky date math with real date functions rather than compound it with more bad date math.
```php
        if ($extend) {
            // extend deadline if not yet passed
            $start_time = $project->smoothread_deadline;
        } else {
            // if starting sr with deadline=0, or if sr ended
            $start_time = time();
        }

        // Extend the deadline to the next whole hour after $days 
        $datetime = new DateTime("@$start_time");
        // Add our days
        $datetime->modify("+{$days} day");
        // round up to the next hour
        $datetime = $datetime->modify("next hour");
        // replace the time with just the hour
        $datetime->modify($datetime->format('H:00:00'));
        $deadline = $datetime->format('U');
```

70ray:
Changed to @cpeel 's recommendation.

Sandbox updated

cpeel (approved):
It occurred to me last night that if we added an upgrade script to round up all existing SR deadlines to the next hour we could get https://github.com/DistributedProofreaders/dproofreaders/pull/988 merged in and it all rolled out at the same time.

I'm happy to help with that upgrade script if you'd like.

srjfoo:
> It occurred to me last night that if we added an upgrade script to round up all existing SR deadlines to the next hour we could get #988 merged in and it all rolled out at the same time.
> 
> I'm happy to help with that upgrade script if you'd like.

I agree that having an upgrade script sounds safer than not. I've seen books put in SR for more than 42 days, because the PPer extended the deadline not too long after they put the book into SR.

pinc/smoothread.inc
```diff
@@ -194,14 +194,24 @@ function handle_smooth_reading_change($project, $postcomments, $days, $extend)
...
-            $deadline = time() + $seconds;
+            $start_time = time();
         }
+
+        // Extend the deadline to the next whole hour after $days
+        $datetime = new DateTime("@$start_time");
+        // Add our days
+        $datetime->modify("+{$days} day");
+        // round up to the next hour
+        $datetime = $datetime->modify("next hour");
```
70ray, line 210: Not very important but this doesn't actually do what I'd intended. If it is already a whole hour it increases to the next. To avoid this we could add 
```
$datetime = $datetime->modify("previous second");
```
before this line.

70ray:
> It occurred to me last night that if we added an upgrade script to round up all existing SR deadlines to the next hour we could get #988 merged in and it all rolled out at the same time.
> 
> I'm happy to help with that upgrade script if you'd like.

I'll try to write something.

pinc/smoothread.inc
cpeel: I think it's fine as-is -- I don't see that an extra hour when we're setting things in days is going to be a big deal.

chrismiceli (approved):

pinc/smoothread.inc
srjfoo: I saw that when looking at one of the test projects. I don't see it as a big deal.

windymilla (approved):

srjfoo:
> > It occurred to me last night that if we added an upgrade script to round up all existing SR deadlines to the next hour we could get #988 merged in and it all rolled out at the same time.
> > I'm happy to help with that upgrade script if you'd like.
> 
> I'll try to write something.

I just had a request from the current Smooth Reading coordinator to not do a script to update past projects, so I think we're back to delaying before deploying #988 in until after everything in the SR pool has the rounded timestamp:

> In applying this change, make sure it is only applies to new entries to  the SR Pool, do not change existing pool projects or it will destroy my entire SR worksheet! Thank you.
>
> I pick up the ending date and post it into my spreadsheet as each project comes into SR or is extended. If those dates change, my whole spreadsheet becomes corrupt.

cpeel:
Thanks for checking, Sharon. We'll merge the cron updates _after_ the Feb push to PROD.

I'd love to know if there is tooling we can develop (noncvs or otherwise) that would help the SR coordinators such that they wouldn't need to maintain a separate spreadsheet.

srjfoo:
> Thanks for checking, Sharon. We'll merge the cron updates _after_ the Feb push to PROD.
> 
> I'd love to know if there is tooling we can develop (noncvs or otherwise) that would help the SR coordinators such that they wouldn't need to maintain a separate spreadsheet.

Undoubtedly there is, but I'm guessing that needs to be discussed with the current SR coordinator. He will likely want to continue making the reports -- that's one thing, I think, that keeps the team going. But having easier and quicker access to the data would undoubtedly make it easier.

PR #988, opened 2023-12-20
revise sending SR notifications assuming whole hour deadlines

70ray:
This is the second stage of the revised method of sending SR end notifications. It should not be merged until at least 42 days after #987 or SR periods that have not been rounded up may fail to send notifications.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/improve_sr_notify

### DIFF
--- a/crontab/finish_smoothreading.php
+++ b/crontab/finish_smoothreading.php
@@ -14,55 +14,38 @@ if ($dry_run) {
     echo "This is a dry run.\n";
 }
 
-// Interval relative to current time
-//   Select all projects whose smooth reading deadline is within this interval.
+// This should run at 5 minutes past each hour
+// Assume smooth read deadlines are at exact hours
+// Select all projects whose smooth reading deadline is within the past hour.
 $from = -60 * 60;
-$to = 60 * 60;
-
 $sql = sprintf(
     "
     SELECT *
     FROM projects
     WHERE
         smoothread_deadline >= (UNIX_TIMESTAMP() + %d)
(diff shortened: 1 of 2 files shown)


crontab/finish_smoothreading.php
```diff
@@ -15,55 +15,38 @@
...
         }
     }
-}
-
-$output .= "finish_smoothreading.php executed.\n";
-
-// Don't output anything if no work was done
-if ($any_work_done) {
-    echo $output;
+    echo "finish_smoothreading.php executed.\n";
```
cpeel, line 51: If we're removing the starting "Checking..." line we should remove this one too.

cpeel (approved):

crontab/finish_smoothreading.php
70ray: > If we're removing the starting "Checking..." line we should remove this one too.

I changed the "Checking..." line to "Found..." and moved it into the conditional block. So this line should also be moved. I'll do that.

cpeel: You are correct, I missed that - thanks!

chrismiceli (approved):

srjfoo (approved):

windymilla (approved):

cpeel:
Note: We need to merge this in _after_ the February deployment to PROD.

cpeel:
Note to self: need to update dpadmin's crontab with the time change post-deploy.

PR #989, opened 2023-12-22
Correct cell bg color for striped rows

cpeel:
Fix setting the has-diff cell background color in Review Work for striped tables.

[Task 2067](https://www.pgdp.net/c/tasks.php?task_id=2067)

This is testable in: https://www.pgdp.org/~cpeel/c.branch/fix-review-work-style/

Specifically see srjfoo's diff [in the sandbox](https://www.pgdp.org/~cpeel/c.branch/fix-review-work-style/tools/proofers/review_work.php?username=srjfoo&use_eval_query=1&work_round_id=P1&review_round_id=P2&days=1200&sample_limit=0) vs the [main TEST code](https://www.pgdp.org/c/tools/proofers/review_work.php?username=srjfoo&use_eval_query=1&work_round_id=P1&review_round_id=P2&days=1200&sample_limit=0) -- you need to be a site admin or PF on TEST to see another user's diffs.

### DIFF
--- a/styles/layout.less
+++ b/styles/layout.less
@@ -1540,7 +1540,8 @@ li.spaced {
 /* Review Work */
 
 td.has-diff {
-    background-color: @table-cell-ok-lowc;
+    /* we use !important so this overrides the bg color in striped tables */
+    background-color: @table-cell-ok-lowc !important;
 }
 
 /* ------------------------------------------------------------------------ */
--- a/styles/themes/charcoal.css
+++ b/styles/themes/charcoal.css
@@ -2524,7 +2524,8 @@ li.spaced {
 /* ------------------------------------------------------------------------ */
 /* Review Work */
 td.has-diff {
-  background-color: #009900;
+  /* we use !important so this overrides the bg color in striped tables */
+  background-color: #009900 !important;
 }
 /* ------------------------------------------------------------------------ */
 /* CharSuite Explorer */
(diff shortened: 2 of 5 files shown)


srjfoo (approved):

70ray (approved):

windymilla (commented):
Sorry I didn't get to review this yesterday.

Although this works, I was under the impression that `!important` should be avoided unless there is no alternative. Just looking online now, an example quote is at [W3 schools](https://www.w3schools.com/css/css_important.asp): 

> Tip: It is good to know about the !important rule. You might see it in some CSS source code. However, do not use it unless you absolutely have to.

I think an alternative (which may have been considered) would be to (re-)specify more precisely than the striped-ness, something like (untested):
```
td.has-diff {
    background-color: @table-cell-ok-lowc;
}
...
table.theme_striped {
    tr:nth-child(even) {
        background-color: @page-background;
        td.has-diff {
            background-color: @table-cell-ok-lowc;
        }
    }
    tr:nth-child(odd) {
        background-color: @sidebar-background;
        td.has-diff {
            background-color: @table-cell-ok-lowc;
        }
    }
}

cpeel:
> I think an alternative (which may have been considered) would be to (re-)specify more precisely than the striped-ness, something like (untested):

I did consider that but I didn't like how the table theme CSS then had to "know" about the has-diff -- I wanted to keep them separate. That's why I went with the `!important` to have the has-diff override whatever else comes its way. It could very well be a poor solution we'll need to undo later but I wanted to start there.

windymilla:
OK - I'm happy with that logic

PR #990, opened 2023-12-23
Factor out user project reserve validation

cpeel:
Factor out the logic used to determine if a user is able to work on a project that might be restricted to a round reserve. Require a User object to be passed in so we don't have to recreate it (with the DB hit) on every call -- which with the current interface is just once per page load, but won't be when I re-use this in the My Recommendations page. There is no functional change.

Testable in https://www.pgdp.org/~cpeel/c.branch/abstract-out-reserve-calculations/

Testing this is a bit hairy because you have to have a user with a certain time on site and a certain number of pages done in P1. I managed to do so on TEST but more importantly we have automated tests! Thanks to @70ray for the `validate_user_can_get_pages_in_project()` tests which caught a bug in my initial implementation!

### DIFF
--- a/SETUP/tests/ProjectTest.php
+++ b/SETUP/tests/ProjectTest.php
@@ -444,21 +444,21 @@ class ProjectTest extends ProjectUtils
         $round = get_Round_for_round_id("P1");
 
         // user done no pages and few days on site
-        validate_user_can_get_pages_in_project($pguser, $project, $round);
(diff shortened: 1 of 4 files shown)


cpeel:
Note to self: get `dont-assume-pguser-in-checks` rebased and put up for PR after this one merges in.

srjfoo (approved):

chrismiceli (approved):

PR #994, opened 2023-12-24
Validate requested user's access to project, not pguser

cpeel:
Fix functions that `validate_user_can_get_pages_in_project()` depend on to operate against the specified user and not assume `$pguser`. This makes the function more generally useful and more consistent.

Testable in https://www.pgdp.org/~cpeel/c.branch/dont-assume-pguser-in-checks/

The only place that the function is called right now is in `project.php` and `LPage.inc` during the proofreading process which all takes place only for the logged-in user so testing this just confirming its enforcing the reserve in the branch just like it does in `master`. The new recommendation page will optionally use this as another user.

### DIFF
--- a/SETUP/tests/ProjectTest.php
+++ b/SETUP/tests/ProjectTest.php
@@ -438,54 +438,50 @@ class ProjectTest extends ProjectUtils
 
     public function test_can_user_get_pages_reserved_for_new_proofreaders()
     {
-        global $pguser;
-        $pguser = $this->TEST_USERNAME;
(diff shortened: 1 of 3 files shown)


70ray (approved):

srjfoo (approved):
Actually, it works as it should (under testing circumstances) in the branch sandbox, but the main TEST site doesn't have the threshold doctored the way the sandbox does, so it doesn't actually trigger the message.

windymilla (approved):

PR #996, opened 2023-12-25
Add new My Suggestions page

cpeel:
A new page for users to find new projects to work on. Includes a very (very) rudimentary recommendation algorithm. The page was originally titled "My Recommendations" but "Suggestions" is both shorter and sounds a little less formal.

The algorithm -- and the bulk of the code -- is the same one currently in noncvs on PROD, however it is not the exact same as this one relies on a few recent PRs.

Code in the PR is testable in https://www.pgdp.org/~cpeel/c.branch/my-suggestions/

### DIFF
--- a/tools/proofers/my_suggestions.php
+++ b/tools/proofers/my_suggestions.php
@@ -1,0 +1,760 @@
+<?php
+$relPath = '../../pinc/';
+include_once($relPath.'base.inc');
+include_once($relPath.'misc.inc'); // array_get(), surround_and_join(), html_safe()
+include_once($relPath.'theme.inc');
+include_once($relPath.'stages.inc');
+include_once($relPath.'Project.inc');
+include_once($relPath.'User.inc');
+include_once($relPath.'project_states.inc');
+include_once($relPath.'js_newwin.inc'); // get_js_for_links_to_project_pages(),  get_onclick_attr_for_link_to_project_page()
+include_once($relPath.'gradual.inc'); // get_pages_proofed_maybe_simulated()
+include_once($relPath.'graph_data.inc'); // get_round_backlog_stats()
+
+require_login();
+
+$verbose = get_integer_param($_GET, "verbose", 0, 0, 1);
+$flush_cache = get_integer_param($_GET, "flush_cache", 0, 0, 1);
+
+$username = User::current_username();
+if (user_is_a_sitemanager() || user_is_proj_facilitator()) {
+    $username = array_get($_GET, 'username', $username);
+    if (!User::is_valid_user($username)) {
+        die("Invalid username.");
+    }
+}
+
+$round_view_options = get_view_options($username);
+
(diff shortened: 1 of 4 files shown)


tools/proofers/my_suggestions.php
```diff
@@ -0,0 +1,760 @@
...
+        }
+
+        // Build the project from the row so we don't have to go back to
+        // the database again for every project. This requires all the necessary
+        // fields used here, and in dependent functions, be populated in the row.
+        $project = new Project($project_row);
+
+        try {
+            $round = $Round_for_project_state_[$project->state];
+            validate_user_against_project_reserve($user, $project, $round);
```
70ray, line 537: Couldn't find the definitionof this function.

windymilla (approved):
That looks ready for general release to me. Minor weight/criteria tweaking can always be done later of course if needed.
Great new feature!

tools/proofers/my_suggestions.php
cpeel: It's in [stages.inc on line 333](https://github.com/DistributedProofreaders/dproofreaders/blob/master/pinc/stages.inc#L333). This was pulled out of `validate_user_can_get_pages_in_project()` in a recent PR that landed a couple of days ago.

chrismiceli (approved):

tools/proofers/my_suggestions.php
70ray: right. that's why I couldn't find it.

tools/proofers/my_suggestions.php
```diff
@@ -0,0 +1,760 @@
...
+            $return_projects[] = $project_row;
+        } catch(Exception $exception) {
```
70ray, line 539: I was originally wondering if any other exceptions can occur that we should not ignore?

tools/proofers/my_suggestions.php
```diff
@@ -0,0 +1,760 @@
...
+            SELECT
+                $project_select_fields
+            FROM projects
+            WHERE $avail_state_clause
+                AND projectid NOT IN (
+                    SELECT projectid
+                    FROM user_project_info
+                    WHERE
+                        t_latest_page_event > 0
+                        AND username='$username'
```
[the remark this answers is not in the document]
cpeel: We 100% should. I've added a second commit which does this here and in My Projects (which was where I sourced the initial version of this at the very beginning).

tools/proofers/my_suggestions.php
cpeel: Right now this can only return `ReservedForNewProofreadersException` but I thought catching any exception was a safer approach here as any exception very likely indicates that validation failed for some reason or another.

70ray (approved):

chrismiceli (approved):

srjfoo (approved):

PR #997, opened 2023-12-26
Upload check

70ray:
This makes a checksum for the uploaded file, uploads it and checks it with the file received by the server.
The javascript function is based on the example in [this mozilla document](https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API/Non-cryptographic_uses_of_subtle_crypto) reworked for the browsers we support which cannot use async, await, File.arrayBuffer() or String.padStart().
This checking should not really be necessary because lower levels of the internet protocol should already validate data but two users have encountered errors given by the zip validator so this is intended to discover if the problem is caused by data corruption or something else.
The change to upload_resumable_file.php is recommended in the resumable docs to avoid unnecessary error messages in the console.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/upload_check

### DIFF
--- a/scripts/file_resume.js
+++ b/scripts/file_resume.js
@@ -72,6 +72,11 @@ window.addEventListener("DOMContentLoaded", function() {
         document.getElementById("upload_progress").innerHTML = text;
     }
 
+    function submitWithHash(hashString) {
+        document.getElementById("hash_code").value = hashString;
+        document.getElementById("upload_form").submit();
+    }
+
     if(resumable.support) {
         resumable.assignBrowse([document.getElementById("resumable_browse")]);
         resumable.assignDrop([document.getElementById("upload_form")]);
@@ -91,10 +96,13 @@ window.addEventListener("DOMContentLoaded", function() {
             return false;
         }
         if(validate(file.name)) {
-            // won't work if we disable browse button so hide it
+            // won't work if we disable the buttons so hide them
             document.getElementById("old_browse").style.display = "none";
-            document.getElementById("old_browse").disabled = true;
+            document.getElementById("old_submit").style.display = "none";
             showProgress(uploadMessages.working);
+            ev.preventDefault();
+            fileHash(file)
+                .then(submitWithHash);
         } else {
             ev.preventDefault();
         }
@@ -131,13 +139,7 @@ window.addEventListener("DOMContentLoaded", function() {
         document.querySelector('input[name="mode"]').value = "resumable";
         showProgress(uploadMessages.finalizingUpload);
         fileHash(file.file)
-            .then(function (hashString) {
-                document.getElementById("hash_code").value = hashString;
-                document.getElementById("upload_form").submit();
-            })
-            .catch( function () {
-                alert("Failed to create checksum");
-            });
+            .then(submitWithHash);
     });
 
     // If an error occurred, re-enable the upload form and show a message
--- a/scripts/file_resume.js
+++ b/scripts/file_resume.js
@@ -23,6 +23,42 @@ window.addEventListener("DOMContentLoaded", function() {
         return true;
     }
 
+    // polyfill for file.arrayBuffer()
+    function fileToArrayBuffer(file) {
+        if ('arrayBuffer' in file) {
+            return file.arrayBuffer();
+        }
+        return new Promise (function (resolve, reject) {
+            const reader = new FileReader();
+            reader.addEventListener("loadend", function() {
+                resolve(reader.result);
+            });
+            reader.addEventListener("error", function() {
+                reject();
+            });
+            reader.readAsArrayBuffer(file);
+        });
+    }
+
+    // polyfill for String.padStart(2, "0")
+    function pad2(s) {
+        return (s.length < 2) ? `0${s}` : s;
+    }
+
+    function fileHash(file) {
+        return fileToArrayBuffer(file)
+            .then(function (arrayBuffer) {
+                return crypto.subtle.digest("SHA-256", arrayBuffer);
+            })
+            .then(function(hashAsArrayBuffer) {
+                const uint8ViewOfHash = new Uint8Array(hashAsArrayBuffer);
+                const hashAsString = Array.from(uint8ViewOfHash)
+                    .map((b) => pad2(b.toString(16)))
+                    .join("");
+                return hashAsString;
+            });
+    }
+
     var resumable = new Resumable({
         target: uploadTarget,
         testTarget: uploadTarget,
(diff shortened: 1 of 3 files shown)


cpeel:
> The change to upload_resumable_file.php is recommended in the resumable docs to avoid unnecessary error messages in the console.

Looks like the original website and repo for this went away. Were you able to find another copy somewhere that we can update `pinc/3rdparty/resumablejs/README.md`?

scripts/file_resume.js
```diff
@@ -23,6 +23,42 @@ window.addEventListener("DOMContentLoaded", function() {
...
+                resolve(reader.result);
+            });
+            reader.addEventListener("error", function() {
+                reject();
+            });
+            reader.readAsArrayBuffer(file);
+        });
+    }
+
+    // polyfill for String.padStart(2, "0")
```
chrismiceli, line 43: do we need a polyfil? The support looks greater than our supported browser matrix:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart
https://www.pgdp.net/wiki/DP_Official_Documentation:General/Supported_Browsers

scripts/file_resume.js
```diff
@@ -23,6 +23,42 @@ window.addEventListener("DOMContentLoaded", function() {
         return true;
     }
 
+    // polyfill for file.arrayBuffer()
+    function fileToArrayBuffer(file) {
```
chrismiceli, line 27: same here, do we need a polyfil? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer

scripts/file_resume.js
```diff
@@ -55,10 +96,13 @@ window.addEventListener("DOMContentLoaded", function() {
...
         }
         if(validate(file.name)) {
-            // won't work if we disable browse button so hide it
+            // won't work if we disable the buttons so hide them
             document.getElementById("old_browse").style.display = "none";
-            document.getElementById("old_browse").disabled = true;
+            document.getElementById("old_submit").style.display = "none";
             showProgress(uploadMessages.working);
+            ev.preventDefault();
+            fileHash(file)
```
chrismiceli, line 104: do we want a .catch?

70ray:
> > The change to upload_resumable_file.php is recommended in the resumable docs to avoid unnecessary error messages in the console.
> 
> Looks like the original website and repo for this went away. Were you able to find another copy somewhere that we can update `pinc/3rdparty/resumablejs/README.md`?

The one I looked at has gone away now. [this one](https://www.npmjs.com/package/resumablejs?activeTab=readme) seems to have the same information.

scripts/file_resume.js
70ray: Looks like we would have to support Chrome 57 but we only do 54

chrismiceli: Ah, missed that. I'm fine with keeping this, but do we know why we support chrome 54 specifically? Was it the last versino supported on xp or something like that? It was released in 2016.

scripts/file_resume.js
70ray: It's File:arrayBuffer()  we need which seems to be same as Blob: arrayBuffer() which is not supported by any of the browsers we can use.

scripts/file_resume.js
70ray: Good point. We should. Perhaps just upload an empty string in this case and let the server say there was no checksum provided.

scripts/file_resume.js
70ray: I don't know. According to browserstack xp only supports chrome 49. xp supports firefox 52 which is the reason for that and firefox 52 does PadStart.

cpeel:
> > > The change to upload_resumable_file.php is recommended in the resumable docs to avoid unnecessary error messages in the console.
> > 
> > 
> > Looks like the original website and repo for this went away. Were you able to find another copy somewhere that we can update `pinc/3rdparty/resumablejs/README.md`?
> 
> The one I looked at has gone away now. [this one](https://www.npmjs.com/package/resumablejs?activeTab=readme) seems to have the same information.

As luck would have it, I actually have a checkout of the old repo on TEST. And I found https://github.com/QuanTemplate/resumable.js/commits/master/ which appears to be a newer fork of the original repo (last commit is by original repo owner). I'll get a fork of that made in DistributedProofreaders so we have it.

cpeel:
We now have a fork in https://github.com/DistributedProofreaders/resumable.js with an updated `README.md` about its lineage for future reference.

scripts/file_resume.js
srjfoo: DP's supported browser list was last updated two-ish years ago in October of 2021. See [Supported Browsers](https://www.pgdp.net/wiki/DP_Official_Documentation:General/Supported_Browsers), and note the comment about Safari, Chrome and Opera on macOS: I couldn't find an official end-of-life schedule for macOS, but it looks like a three-year cycle. Big Sur (macOS 11) is no longer supported now.

How rigorous do we want to be in updating our supported browser list?

windymilla (approved):

scripts/file_resume.js
cpeel: To not boil the ocean, for this PR I think we should bump our minimum Chrome version to 57 and drop the polyfill. Based on the [platform support numbers for Chrome](https://en.wikipedia.org/wiki/Google_Chrome#Platforms) 52 -> 57 doesn't cross any threshold and seems very sensible.

cpeel: I'm happy to revisit all of our minimum browser versions if we need to, but it doesn't sound like this warrants that extra effort.

chrismiceli (approved):

cpeel (approved):
Note to whoever merges this in (maybe me): bump our minimum Chrome version to 57 upon merge.

70ray:
Sandbox updated.

70ray:
We should also bump minimum Safari version from 10 to 11 and Opera from 41 to 44.
I just tried the older versions and chrome 54 on browserstack and they upload the file without any problems but the server says "No checksum provided".

srjfoo:
> We should also bump minimum Safari version from 10 to 11 and Opera from 41 to 44. I just tried the older versions and chrome 54 on browserstack and they upload the file without any problems but the server says "No checksum provided".

Minimum version for Safari is already at 11. And confirmed the other two.

I was already checking out browserstack, because I was thinking I still needed to be checking out the broken non-resumable issue (only to discover that I had to go to XP on a browser no longer supported to be able to find a browser version that didn't support resumables (of the browsers that we support). And I didn't get an error message from it.

srjfoo (approved):

cpeel:
https://www.pgdp.net/wiki/DP_Official_Documentation:General/Supported_Browsers has been updated with the new minimum versions.

PR #998, opened 2023-12-27
Generalize function caching code

cpeel:
Introduce a generalized version of `query_graph_cache()` that can be used for any data-collecting function. As for the name, see the [Memoization](https://en.wikipedia.org/wiki/Memoization) wikipedia article if you're not familiar with it. The updated graph pages are best viewed ignoring whitespace as the code was just indented when added to a function.

Testable in https://www.pgdp.org/~cpeel/c.branch/memoize-function/

I've added a new `memcache-stats` feature to the Squirrel Workbench on TEST that shows the memcache statistics. This can be used to confirm the new code is hitting memcache to get the cached data.

### DIFF
--- a/pinc/misc.inc
+++ b/pinc/misc.inc
@@ -1272,6 +1272,38 @@ function factor_strings($strings)
     return [$left_common, $middles, $right_common];
 }
 
+//--------------------------------------------------------------------------
+
+/**
+ * Given a function and arguments that will generate data, attempt
+ * to load the data from memcached first and fall back to the function and
+ * cache the response. If memcache is not running, this will call the requested
+ * function and return its data without generating a notice/warning/error.
+ *
+ * If the requested function returns data that has localized strings, it's
+ * important that $key_salt = get_desired_language();
+ */
+function memoize_function($function, $args = [], $expire_from_now = 3600, $key_salt = "")
+{
+    $memcache = new Memcached();
+    $memcache->addServer('localhost', 11211);
+
+    $key = hash("sha256", $function . serialize($args) . $key_salt);
+
+    // if the key exists, just return the data uncompressed
+    $data = $memcache->get($key);
+    if ($data !== false) {
+        return unserialize(gzuncompress($data));
+    }
+
+    // if not, call the function to return the data
+    $data = call_user_func_array($function, $args);
+
+    $memcache->set($key, gzcompress(serialize($data)), time() + $expire_from_now);
+
+    return $data;
+}
+
 // -----------------------------------------------------------------------------
 
 /**
--- a/pinc/misc.inc
+++ b/pinc/misc.inc
@@ -1285,8 +1285,12 @@ function factor_strings($strings)
  */
 function memoize_function($function, $args = [], $expire_from_now = 3600, $key_salt = "")
 {
-    $memcache = new Memcached();
-    $memcache->addServer('localhost', 11211);
+    // cache this connection for possible re-use during this page load
+    static $memcache = null;
+    if (!$memcache) {
+        $memcache = new Memcached();
+        $memcache->addServer('localhost', 11211);
+    }
 
     $key = hash("sha256", $function . serialize($args) . $key_salt);
 
--- a/SETUP/API.md
+++ b/SETUP/API.md
@@ -54,8 +54,10 @@ ensure that 3rd parties don't obtain API keys from the requests.
 ## Rate Limiting
 
 API requests can be rate limited to help prevent the site from being
-overwhelemed. To use Rate Limiting, you must have the PHP memcached module
+overwhelmed. To use Rate Limiting, you must have the PHP memcached module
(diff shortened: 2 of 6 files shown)


pinc/misc.inc
```diff
@@ -1272,6 +1272,38 @@ function factor_strings($strings)
...
+ * cache the response. If memcache is not running, this will call the requested
+ * function and return its data without generating a notice/warning/error.
+ *
+ * If the requested function returns data that has localized strings, it's
+ * important that $key_salt = get_desired_language();
+ */
+function memoize_function($function, $args = [], $expire_from_now = 3600, $key_salt = "")
+{
+    $memcache = new Memcached();
+    $memcache->addServer('localhost', 11211);
```
chrismiceli, line 1289: should we unconditionally add? Per https://www.php.net/manual/en/memcached.addserver.php it seems not advised to add the same server multiple times to the same server pool.

pinc/misc.inc
```diff
@@ -1272,6 +1272,38 @@ function factor_strings($strings)
...
+
+    $key = hash("sha256", $function . serialize($args) . $key_salt);
```
chrismiceli, line 1291: sha256 is resistant to collision attacks, this is a property we probably don't need for caching. Should we instead use something faster?

pinc/misc.inc
chrismiceli, line 1289: what if there is no memcache server? Is this setup in configuration somewhere?

chrismiceli (approved):

pinc/misc.inc
cpeel: This function creates the memcached object at every call so we need to add it for every call. The object itself doesn't persist.

chrismiceli: oh yes, my bad. I guess the next question is should we not destroy it every call and instead keep the pool of servers alive? Does it have a strong impact to performance?

pinc/misc.inc
cpeel: I changed it from md5 to sha256 to reduce the collision space since this has no control over the function name, args, or salt. I don't know what the odds are of a collision with md5 though, so it might not matter and md5 might be good enough.

pinc/misc.inc
cpeel: If there is no local memcache server running, or it's not connected, all of the memcache calls return `false` and the code falls through to just calling the function, I tried to convey this in the function header with:
```
If memcache is not running, this will call the requested
function and return its data without generating a notice/warning/error.
```

So having memcache is effectively optional and if it isn't there or isn't working there won't be any caching.

The API uses memcache for rate limiting and this is documented in the `API.md`. I should update `INSTALL.md` and mention the use of memcache for other caching too, I'll do that.

cpeel: Connection pooling is difficult with PHP as each page load is a distinct "execution" in most cases (this is not entirely accurate depending on how PHP is wired into the web server (eg Apache) and what multi-processing model is used (ie [MPM](https://httpd.apache.org/docs/2.4/mpm.html))). It's this complexity that has us not pooling or using persistent DB connections either.

The per-instances loopback connections (for both memcache and the database) should be pretty lighweight and at least in this case we're just trading the MySQL connection + query complexity for a memcached connection + constant-time hash lookup. I will see if we can cache the connection within a page load via a static variable though, for the case where we have multiple calls to this function per page load -- let me test that.

cpeel: I've cached the connection in a static function variable so we will at least re-use it in a single page load if there are multiple calls to the function.

cpeel: I've updated both `API.md` and `INSTALL.md` to include much more details about memcached usage.

chrismiceli: Perfect, thank you!

pinc/misc.inc
```diff
@@ -1272,6 +1272,42 @@ function factor_strings($strings)
...
+ * If the requested function returns data that has localized strings, it's
+ * important that $key_salt = get_desired_language();
+ */
+function memoize_function($function, $args = [], $expire_from_now = 3600, $key_salt = "")
+{
+    // cache this connection for possible re-use during this page load
+    static $memcache = null;
+    if (!$memcache) {
+        $memcache = new Memcached();
+        $memcache->addServer('localhost', 11211);
```
chrismiceli, line 1292: minor point, but should we make the port and server configurable similar to other dependencies?

chrismiceli (approved):

pinc/misc.inc
cpeel: We probably should but I'm going to punt on that right now. The install docs talk about this being accessible on localhost so they at least reflect reality.

windymilla (approved):
Tested on TEST, that when `round_backlog_days.php` is refreshed in the sandbox, the cache hit number increases, but the deployed version of that file doesn't affect the cache.
Code looks good too.

srjfoo (approved):

PR #999, opened 2023-12-27
Improve performance for neighbor stats

cpeel:
Currently on PROD, the round pages take ~2.5 seconds each to load. >2 seconds of this time is coming from the logic to calculate a user's neighbors for the stats bar.

Not all users proof in all rounds, so the size of the users table is always going to be bigger than the number of tallies in a round. So instead of generating tallies in a round for all users and trimming out those who aren't present, only get users with tallies and special-case when the current user isn't in the tally round. Performance is now based on how many users have proofread in a round rather than the number of entries in the user table.

I've simplified the function which makes it specific to this one use-case, but it's only used in one place so the extensibility isn't necessary.

Here are the timings of the SQL for the 5 round pages on PROD:
| Round        | Current   | PR        |
|--------------|-----------|-----------|
| P1           | 2.07 sec  | 1.12 sec  |
| P2           | 1.75 sec  | 0.13 sec  |
| P3           | 1.90 sec  | 0.02 sec  |
| F1           | 1.77 sec  | 0.10 sec  |
| F2           | 1.82 sec  | 0.00 sec  |

Testable in https://www.pgdp.org/~cpeel/c.branch/improve-neighbor-performance/

Results should be identical to what is live on TEST now, although the performance difference will be negligible there. Be sure to look at a round page for a user who hasn't proofread anything in that round and confirm user's 0-entry is present and last.

### DIFF
--- a/pinc/TallyBoard.inc
+++ b/pinc/TallyBoard.inc
@@ -240,60 +240,53 @@ class TallyBoard
      * Parameters after `$radius` are an optimization for extracting
      * other info from the database at the same time.
      */
-    public function get_neighborhood(
-        $target_holder_id,
-        $radius,
-        $other_table,
-        $holder_id_expr,
-        $other_select_exprs,
-        $tally_alias,
-        $rank_alias
-    ) {
+    public function get_neighborhood($target_holder_id, $radius)
+    {
         assert($radius >= 0);
 
         $sql = sprintf(
             "
             SELECT
-                $holder_id_expr AS _holder_id,
-                IFNULL(tally_value,0) AS $tally_alias,
-                $other_select_exprs
+                IFNULL(tally_value, 0) AS current_tally,
+                u_id, username, u_privacy, date_created
             FROM
-                $other_table
-                LEFT OUTER JOIN current_tallies
-                ON (
-                    tally_name      = '%s'
-                    AND holder_type = '%s'
-                    AND holder_id   = $holder_id_expr
-                )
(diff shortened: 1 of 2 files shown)


pinc/TallyBoard.inc
```diff
@@ -240,60 +240,53 @@ class TallyBoard
...
+            WHERE
+                tally_name = '%s'
+                AND holder_type = '%s'
+                AND tally_value > 0
+            ORDER BY tally_value DESC, u_id ASC
             ",
             DPDatabase::escape($this->tally_name),
-            DPDatabase::escape($this->holder_type)
+            DPDatabase::escape($this->holder_type),
+            $target_holder_id
```
chrismiceli, line 263: is this used in the query? I only see 2 placeholders in the string, but I could be missing it.

chrismiceli (approved):

chrismiceli:
This performance improvement will be amazing!

pinc/TallyBoard.inc
cpeel: Good eyes! It is not and was left over from a prior iteration. I've pushed up a commit to remove it. 🙏

chrismiceli (approved):

windymilla (approved):
LGTM, including testing with a user who has not proofed in a round.

70ray (approved):
I can't pretend to understand the sql stuff but it works as expected.

srjfoo (approved):

PR #1000, opened 2023-12-27
Don't incur neighborhood overhead if u_neigh=0

cpeel:
If users don't want to see the neighborhood, don't incur the performance penalty of loading it. Also use u_neigh to show users a wider range on their personal stats page.

Testable in https://www.pgdp.org/~cpeel/c.branch/improve-stats-for-no-neighbors/

Should show identical stats to TEST with the exception of Rank which will now show actual rank on TEST and not "Valued Tester".

---

I also strongly considered setting the default value for new accounts so `u_neigh=0`. I still think that's a good idea but we can evaluate that separate from this optimization PR.

### DIFF
--- a/pinc/theme.inc
+++ b/pinc/theme.inc
@@ -622,38 +622,21 @@ function show_tally_specific_stats($tally_name)
         // The requestor is a logged-in user.
         // Show the user's personal statistics
 
-        //Get the personal statistics array
-        $neighbors =
-            user_get_page_tally_neighborhood(
-                $tally_name,
-                $user->username,
-                $user->u_neigh
-            );
-        $usern = $neighbors[0];
+        $current_tally = $tallyboard->get_current_tally($user->u_id);
+        $current_rank = $tallyboard->get_rank($user->u_id);
 
         //get rank
-        global $testing;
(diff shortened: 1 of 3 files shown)


pinc/theme.inc
```diff
@@ -622,38 +622,21 @@ function show_tally_specific_stats($tally_name)
...
-                $user->username,
-                $user->u_neigh
-            );
-        $usern = $neighbors[0];
+        $current_tally = $tallyboard->get_current_tally($user->u_id);
+        $current_rank = $tallyboard->get_rank($user->u_id);
 
         //get rank
-        global $testing;
-        if ($testing) {
```
cpeel, line 636: Removing this was intentional. The source code is publicly accessible -- no one needs to inflate pagecounts on TEST to get a list of them.

chrismiceli (approved):

cpeel:
Per a discussion in #pgdpdev, I've added a second commit that sets the default u_neigh for new users to 0.

windymilla (approved):

70ray (approved):

srjfoo (approved):

PR #1001, opened 2023-12-30
Move misc.inc include to bootstrap

cpeel:
A _vast_ majority of files use functions in `misc.inc`, just include that in `bootstrap.inc` so we can assume functions in there are available everywhere and remove other includes.

This also adds a CI/CD check to make sure `base.inc` (which pulls in `bootstrap.inc`) is correctly being included in all `.php` files per our [best practices](https://www.pgdp.net/wiki/DP_Code_Best_Practices#Every_PHP_file_must_include_base.inc).

I've spun up a sandbox as a sanity check but due to the new CI/CD check we can confirm that all pages are pulling in `base.inc` which pulls in `bootstrap.inc` which pulls in `misc.inc` so there is no need for extensive testing.

https://www.pgdp.org/~cpeel/c.branch/move-misc-to-bootstrap/

### DIFF
--- a/.github/workflows/static-checks.yml
+++ b/.github/workflows/static-checks.yml
@@ -94,3 +94,5 @@ jobs:
         run: cd SETUP && make lint_charsuites && cd ..
       - name: Run less/CSS checks
         run: cd SETUP && make lint_css && cd ..
+      - name: Run best practice checks
+        run: cd SETUP && make best_practice_checks && cd ..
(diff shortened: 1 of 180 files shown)


srjfoo (approved):
I tested a few random things, including logging out and back in again, with no problems.

chrismiceli (approved):

windymilla (approved):

PR #1006, opened 2024-01-02
Move function definitions from stage declarations

cpeel:
Move round-specific functions to `Round.inc` and project page-specific function to `project.php`. This makes `stages.inc` just declare the stages and not house function definitions.

This is a no-op change as `stages.inc` includes `Round.inc` at the top which guarantees that the functions will be defined after `stages.inc` loads, just like they are now. And `project.php` is the only place where  `echo_postcomments_instructions()` is used.

Sanity check sandbox: https://www.pgdp.org/~cpeel/c.branch/separate-defs-from-uses/


windymilla (approved):

chrismiceli (approved):

srjfoo (approved):
I customized `$page_tally_threshold` and `$days_on_site_threshold` in `Round.inc`, and some of the threshold values in `gradual.inc` in the sandbox, and did some spot testing. Looked good!

PR #1007, opened 2024-01-03
Allow generic URL for user's own details

cpeel:
Allow a generic URL for accessing a user's own details. This allows us to include it in forum posts and documentation. I want to include a link to a user's own P1 stats in a forum post and this allows that.

Diff best viewed ignoring whitespace.

Example in sandbox: https://www.pgdp.org/~cpeel/c.branch/allow-generic-mdetail-url/stats/members/mdetail.php?tally_name=P1

### DIFF
--- a/stats/members/mdetail.php
+++ b/stats/members/mdetail.php
@@ -11,23 +11,32 @@ include_once('../includes/member.inc');
 
 require_login();
 
-$id = get_integer_param($_GET, 'id', null, 0, null);
-$user = User::load_from_uid($id);
+$id = get_integer_param($_GET, 'id', null, 0, null, true);
+if ($id === null) {
+    $user = User::load_current();
+} else {
+    $user = User::load_from_uid($id);
+}
 
 $valid_tally_names = array_keys(get_page_tally_names());
 $tally_name = get_enumerated_param($_GET, 'tally_name', null, $valid_tally_names, true);
 
 $can_reveal = can_reveal_details_about($user->username, $user->u_privacy);
-if ($can_reveal) {
-    $user_referent = "'" . $user->username . "'";
+if ($user == User::load_current()) {
+    $desc = _("Your user details");
 } else {
-    $user_referent = "#" . $user->u_id;
-    // Note that this doesn't reveal anything;
-    // the requestor already knows the subject's u_id,
-    // because it was included in the request.
+    if ($can_reveal) {
+        $user_referent = "'" . $user->username . "'";
+    } else {
+        $user_referent = "#" . $user->u_id;
+        // Note that this doesn't reveal anything;
+        // the requestor already knows the subject's u_id,
+        // because it was included in the request.
+    }
+
+    $desc = sprintf(_("Details for user %s"), $user_referent);
 }
 
-$desc = sprintf(_("Details for user %s"), $user_referent);
 output_header($desc, SHOW_STATSBAR, [
     "js_files" => get_graph_js_files(),
 ]);


cpeel:
@chrismiceli - any ideas on the failing jsunit tests? The geckodriver versions look wildly different from each other:
> The geckodriver version (20240103070030.081) detected in PATH at /home/runner/work/dproofreaders/dproofreaders/node_modules/.bin/geckodriver might not be compatible with the detected firefox version (121.0); currently, geckodriver 0.33.0 is recommended for firefox 121.*, so it is advised to delete the driver in PATH and retry

windymilla (approved):

stats/members/mdetail.php
```diff
@@ -11,23 +11,32 @@
...
-    $user_referent = "'" . $user->username . "'";
+if ($user == User::load_current()) {
+    $desc = _("Your user details");
 } else {
-    $user_referent = "#" . $user->u_id;
-    // Note that this doesn't reveal anything;
-    // the requestor already knows the subject's u_id,
-    // because it was included in the request.
+    if ($can_reveal) {
+        $user_referent = "'" . $user->username . "'";
```
chrismiceli, line 29: do we need html encoding? Apologies to keep bringing this up and causing issues with the sql statement for instance.

chrismiceli (approved):

chrismiceli:
> @chrismiceli - any ideas on the failing jsunit tests? The geckodriver versions look wildly different from each other:
> 
> > The geckodriver version (20240103070030.081) detected in PATH at /home/runner/work/dproofreaders/dproofreaders/node_modules/.bin/geckodriver might not be compatible with the detected firefox version (121.0); currently, geckodriver 0.33.0 is recommended for firefox 121.*, so it is advised to delete the driver in PATH and retry

Looks like maybe the compat issues between firefox version and driver version for selenium. It could be transient as the firefox version associated with the geckdriver is updated by the tool. Otherwise we may need to fix some versions to known good values instead of relying on the latest.

If it still occurs tonight, I can investigate further.

chrismiceli:
Do we need to update the privacy documentation in the preferences or wiki any?

stats/members/mdetail.php
cpeel: The problems with the SQL statement were entirely mine because I didn't adequately test it and you were completely right about needing it, so please don't apologize. Given what happened some years ago I'm always interested in possible SQL escaping failures.

We should _probably_ be `html_safe()`ing some of the stuff in this file -- like `$desc` when outputting the heading -- but honestly I just don't have the gumption for changing unrelated code right now. I'm leaving it no worse off than I found it which is my minimum bar.

Related, I think we should revisit our philosophy on needing `html_safe()` around translated strings that don't contain user input. We manage the string translations like we do the code: only a very limited number of people have access to it and we can enforce correctness in them. Also, sometimes our translated strings contain HTML code (like links) that we can't escape which means we're being inconsistent with this right now anyway. We still need `attr_safe()` on all translated strings though as they can and will contain single quotes.

cpeel:
> Do we need to update the privacy documentation in the preferences or wiki any?

Users are always allowed to see their own statistics and this code only gives users more ready access to their own stats so I don't think so.

cpeel:
> > @chrismiceli - any ideas on the failing jsunit tests? The geckodriver versions look wildly different from each other:
> > > The geckodriver version (20240103070030.081) detected in PATH at /home/runner/work/dproofreaders/dproofreaders/node_modules/.bin/geckodriver might not be compatible with the detected firefox version (121.0); currently, geckodriver 0.33.0 is recommended for firefox 121.*, so it is advised to delete the driver in PATH and retry
> 
> Looks like maybe the compat issues between firefox version and driver version for selenium. It could be transient as the firefox version associated with the geckdriver is updated by the tool. Otherwise we may need to fix some versions to known good values instead of relying on the latest.

I re-ran the failed test this morning and it ran successfully!

srjfoo (approved):

stats/members/mdetail.php
chrismiceli: Perfect. Works for me.

PR #1009, opened 2024-01-04
Replace round globals with class statics

cpeel:
_I'm putting this up first for a conceptual "does this seem like a good approach" review rather than an in-depth code review for testing. See `Round.inc` for the important bits._

Global variables are a pain to work with and in general an anti-pattern when they can be avoided. I don't think they could when this code was written, but PHP has come a long way in 20 years. One approach to avoiding globals is to use a class -- either a singleton or static variables/methods within the class.

Rather than using globals to store access to round objects, use static variables & functions within the Round class. This is a more robust approach to what we currently do for the [`Quiz` class](https://github.com/DistributedProofreaders/dproofreaders/blob/master/pinc/Quiz.inc#L9). Note that we do the same global approach for Stages and Pools, and perhaps others, so this should be a pattern we want to reuse elsewhere.

The upsides to using a class are:
1. No need to `global $varname` in a function to access it.
2. The autoloader will pull in the right include file so you don't need to explicitly do so.
3. We hide the implementation details as all users access the data via functions. This lets us change the implementation at-will without breaking consumers.
4. Because class definitions, like function definitions, are global regardless of how they are included, we don't have to worry about an include file getting pulled into a non-global context and making not-explicitly-defined globals non-globals. (We've hit this numerous times with tests in phpunit and why you now see global definitions in places like `Round.inc` where they weren't there before.)

One could argue that the `Round` class shouldn't be the one keeping track of objects that are created from it. It feels a little odd to say `Round::get_rounds()`. We could create a new singleton "holder" class for these created objects (stages, rounds, pools) instead, eg `Structure::get_rounds()`. A downside to that approach is that at least for rounds, we'd end up encoding some of the round-specific logic inside the holder class.

I welcome all feedback but in particular:
* using classes over globals
* using the existing classes (Round) to store the objects rather than a new singleton


windymilla (commented):
Although not a php expert, I think the class rather than global variable approach is definitely a good thing for the reasons laid out in the PR description.

I do think storing the list of created rounds within the `Round` class is unusual in that it surprised me when I first came across it. Regarding the suggestion to have a `Structure` class, would it be possible/better to have a `Rounds` class that holds the list of rounds, and could therefore "know a bit about rounds" where absolutely necessary. Maybe `Rounds` could inherit from a `Structure` base class if there are similar types that need a `MyClass` stored in `MyClasses` approach?

70ray:
The way the rounds and so on worked always seemed very strange to me. I need to try out a few things.

cpeel:
> The way the rounds and so on worked always seemed very strange to me. I need to try out a few things.

It's a bit unique. The class structures are Activity, Stage, Round, and Pool, the inheritance looks something like:
```
Activity
  -> Stage
     -> Round
     -> Pool
```

The creation of objects from these classes are all in `stages.inc` via `new <classname>`. Each class constructor adds itself to some global variable to access the object later (`$Activity_for_id_`, `$Stage_for_id_`, several `$Round_for_*`, `$PAGE_STATES_IN_ORDER`, `$Pool_for_id_`, and `$Pool_for_state_`). Each class file has some function (not part of the class) that can be used to access one of the objects, eg `get_Round_for_round_id()`.

What's really interesting (and frustrating) is that `stages.inc` is critical to creating all of the site structure via the object creation but that isn't consistently included in files. We mostly avoid problems because most pages include `theme.inc` which pulls in `stages.inc`. But for pages that use a `slim_header()` it's super easy to get errors unless you include `stages.inc` or happen to include something that includes it (`Project.inc`).

I'm working at untangling this (and indeed have a branch that attempts to resolve it, but I'm not fully happy with it yet).

70ray:
> > The way the rounds and so on worked always seemed very strange to me. I need to try out a few things.
> 
> It's a bit unique. The class structures are Activity, Stage, Round, and Pool, the inheritance looks something like:
> 
> ```
> Activity
>   -> Stage
>      -> Round
>      -> Pool
> ```
> 
> The creation of objects from these classes are all in `stages.inc` via `new <classname>`. Each class constructor adds itself to some global variable to access the object later (`$Activity_for_id_`, `$Stage_for_id_`, several `$Round_for_*`, `$PAGE_STATES_IN_ORDER`, `$Pool_for_id_`, and `$Pool_for_state_`). Each class file has some function (not part of the class) that can be used to access one of the objects, eg `get_Round_for_round_id()`.
> 
> What's really interesting (and frustrating) is that `stages.inc` is critical to creating all of the site structure via the object creation but that isn't consistently included in files. We mostly avoid problems because most pages include `theme.inc` which pulls in `stages.inc`. But for pages that use a `slim_header()` it's super easy to get errors unless you include `stages.inc` or happen to include something that includes it (`Project.inc`).
> 
> I'm working at untangling this (and indeed have a branch that attempts to resolve it, but I'm not fully happy with it yet).

Yes there has to be a better way.

srjfoo:
Moving away from globals as much as possible sounds good, and reading through the code, it makes sense to me. However I don't think I have the knowledge to really say what the best approach to the problem is. I'm looking forward to testing it, though 😁 .

70ray:
I don't like the way the existing code creates all the rounds and stages (which seem to be anonymous global objects which are acceseed by tables) because most of the data is never used, it would be better to just create the data that you actually need. But changing that would perhaps be a major undertaking too far.
Getting the round you want from the various pieces of information (number, id, project_state etc.) is probably more efficient using tables. Perhaps you could wrap the tables in an accessor object so there would only be one extra global object.
I thought at first it could be possible to extend the generic Round to new classes P1, P2 etc. Then if you wanted to get the round for a project state such as "P1.proj_avail" you could 'parse' the string to get "P1" and make the round with "new P1". But this would be such an entirely different approach that it would be very difficult to do now.

cpeel:
>  Getting the round you want from the various pieces of information (number, id, project_state etc.) is probably more efficient using tables. Perhaps you could wrap the tables in an accessor object so there would only be one extra global object.

Using an associative array rather than the iterator I currently have implemented would be more efficient, but the class abstraction allows us to implement lookup by associative array within the class and retain the abstraction. I don't think the overhead of accessing a static function in a class to access an associative array vs just accessing an associative array is meaningful given the benefits of removing the globals that are often not used at all.

cpeel:
I've added a new commit that creates a new `Rounds` singleton class with similar static functions to the first iteration of this PR. Once again, the `Round.inc` file is the one to pay the most attention to.

If this is tilting in the right direction I'll look and see if this should be a derived class from an Activity "collections" class -- which if feels like it should since we collect Activities and Stages and Pools too.

cpeel:
I had an inspiration earlier: all `Stage`, `Round`, and `Pool` classes drive from `Activity`, so we already have a single collection of all objects instantiated -- we just want easy access methods to look them up without using globals.

I've added a new `Activities` access class that stores all `Activity` and derived objects (just like the `$Activity_for_id_` global does now) and created access classes `Stages`, `Rounds`, and `Pools` that access exactly what they sound like.

The relevant files in this PR are then:
* `Activity.inc`
* `Stage.inc`
* `Round.inc`
* `Pool.inc`

Everything else is just changing uses of the globals to the relevant access classes which is admittedly a lot. Still holding off on putting up a testing sandbox until we think this is a worthwhile direction.

chrismiceli (approved):

70ray (commented):
I quite like this. I've been looking further at an alternative way of defining the rounds etc. and it has converged with this in using Rounds etc. classes although it makes them first rather than afterwards. So it would be similar at higher levels.
The way mine does get_all() uses yield from to delegate to lower levels. I need to do some more work on it to see if it all makes sense.

cpeel:
I've simplified this PR by removing the codebase-wide globals replacement so we can better compare it to @70ray's alternate approach in https://github.com/DistributedProofreaders/dproofreaders/pull/1031.

pinc/Stage.inc
```diff
@@ -10,7 +10,29 @@
 
 include_once($relPath.'Activity.inc');
 
-$Stage_for_id_ = [];
+/*
+ * A singleton container class for accessing Stage objects.
+ * See the Activities class definition for more information.
+ */
+class Stages extends Activities
```
70ray, line 17: It would be simpler not to extend Activities but have Activities, Stages, Pools and Rounds to be independent classes. Then when the different activities are created each one will get the appropriate activities set in it: Stages will get all the Rounds and Pools and Activities will get everything. When you create a Round it creates its parent Stage which will put the Round into its Stages class and so on.

pinc/Round.inc
```diff
@@ -1,20 +1,55 @@
 <?php
+include_once($relPath.'Stage.inc');
 
-// $Id$
+/*
+ * A singleton container class for accessing Round objects.
+ * See the Activities class definition for more information.
+ */
+class Rounds extends Stages
```
70ray, line 8: This will be more complicated than Activity but could use the trait with an override like
```
class Rounds
{
    use StatObj {set as obj_set;}

    private static $round_for_number = [];
    private static $round_for_page_state = [];
    private static $page_states = [];

    public static function set($object)
    {
        self::obj_set($object);
        self::$round_for_number[$object->round_number] = $object;
    }

    public static function set_page_state($page_state, $object)
    {
        self::$page_states[] = $page_state;
        self::$round_for_page_state[$page_state] = $object;
    }

    public static function get_page_states()
    {
        return self::$page_states;
    }

    public static function get_for_page_state($page_state)
    {
        return self::$round_for_page_state[$page_state] ?? null;
    }

    public static function get_for_number($number)
    {
        return self::$round_for_number[$number] ?? null;
    }
}
```

pinc/Round.inc
```diff
@@ -105,12 +138,29 @@ class Round extends Stage
...
 
+        // Populate $this->page_states with all page states that apply to this
+        // round for lookup later. This order is retained for code that needs
+        // to know the order.
+        $this->page_states = [];
+        foreach (["avail", "out", "temp", "save", "bad"] as $state) {
+            $attribute = "page_{$state}_state";
+            $this->page_states[] = $this->$attribute;
+        }
+
```
70ray, line 163: Or you can put them into the Rounds class like:
```php
        Rounds::set($this);
        Rounds::set_page_state($this->page_avail_state, $this);
        Rounds::set_page_state($this->page_out_state, $this);
        Rounds::set_page_state($this->page_temp_state, $this);
        Rounds::set_page_state($this->page_save_state, $this);
        Rounds::set_page_state($this->page_bad_state, $this);
```

pinc/Activity.inc
```diff
@@ -95,8 +179,7 @@ class Activity
                 sprintf(_("'%s' pages completed"), $this->id);
         }
 
-        global $Activity_for_id_;
-        $Activity_for_id_[$id] = & $this;
+        Activities::add($this, $id);
```
70ray, line 182: or have 
```php
        Activities::set($this);
```
since the acivity knows its own id.

70ray:
> I've simplified this PR by removing the codebase-wide globals replacement so we can better compare it to @70ray's alternate approach in #1031.

@70ray  has realised that there is a simpler way than #1031 which is quite close to this so he has closed #1031 and added some comments here.

pinc/Activity.inc
cpeel: Good point. I originally had this at the top of the function which was before the object knew about its ID.

pinc/Activity.inc
```diff
@@ -18,7 +18,91 @@ $ACCESS_CRITERIA = [
...
+ * (Stages, Rounds, Pools).
+ *
+ * The Activity constructor adds the constructed object to this class and
+ * because Stage, Round, and Pool classes inherit from the Activity class,
+ * those objects are accessible from this class too. For code readability,
+ * those classes have accessor classes of their own (Stages, Rounds, Pools)
+ * that derive from Activities and specify which objects to access.
+ */
+class Activities
+{
```
70ray, line 32: (I put this comment in the wrong place earlier)
It would be simpler not to extend Activities but have Activities, Stages, Pools and Rounds to be independent classes. Then when the different activities are created each one will get the appropriate activities set in it: Stages will get all the Rounds and Pools, Activities will get everything. When you create a Round it creates its parent Stage which will put the Round into its Stages class and so on.
The classes would be the same shape so you could define them with a trait like
```php
trait StatObj
{
    private static $objects = [];

    public static function set($object)
    {
        self::$objects[$object->id] = $object;
    }

    public static function get_all()
    {
        return self::$objects;
    }

    public static function get_by_id($id)
    {
        return self::$objects[$id] ?? null;
    }
}

class Activities
{
    use StatObj;
}
```

pinc/Pool.inc
```diff
@@ -1,11 +1,34 @@
 <?php
-
 include_once($relPath.'Stage.inc');
 
-// -----------------------------------------------------------------------------
+/*
+ * A singleton container class for accessing Pool objects.
+ * See the Activities class definition for more information.
+ */
+class Pools extends Stages
```
70ray, line 8: In this case
```php
class Pools
{
    use StatObj;
}
```

pinc/Activity.inc
70ray: It looks like get_ids() is also needed, which could be put in the trait
```php
    public static function get_ids()
    {
        return array_keys(self::$objects);
    }
```

70ray (commented):
What I have suggested doesn't include anything involving project states but we could do this in the same way where it is needed for rounds and pools.
We may need to change it later when reworking project sates to use `rounds->get_all()` etc.

pinc/Activity.inc
cpeel: > It would be simpler not to extend Activities but have Activities, Stages, Pools and Rounds to be independent classes.

Perhaps, but I rather like that as-is the "collection" classes have inheritance like the objects they contain (Activity, Stage, Round, Pool). The current way also is more efficient that we only actually map the object once at the base level -- Activities -- rather than duplicate it everywhere. This should be very minor though as we are storing a reference to the same object, not a new object every time (this was true in the original implementation too).

 I'll have to give it more thought.

pinc/Round.inc
cpeel: I'm uncertain that accesses by number, page_state, etc are oft-used enough that we care about using an assoc array for them. We'll probably spend more time initializing them for every page load than we do using them after they are initialized.

I'll add some logging to some of the current access functions and see how often they are used on average, but my intuition is that they aren't used enough to be worth the trouble of creating an assoc array.

cpeel:
My latest commits simplifies both the main Activities class and the child classes (Stages is 100% derived from Activities now) by making Activities class-type-aware. This replicates the spirit of what @70ray was going for but continues to keep the standard OOP model of subclasses that match with the underlying objects they are managing.

70ray (commented):
I see what you are doing with this.

70ray (commented):
Creating the static classes with functions like `get_by_xxx()` and `get_all_xxx` is a great idea. There are obviously many different ways of imlementing their internals, differing in how much memory they use to initialise and how efficiently the functions execute. So let's reinstate all the commits and remove the globals hacks and see how it runs.

cpeel:
Codebase-wide change of globals to new access methods is back in. This code is live in https://www.pgdp.org/~cpeel/c.branch/remove-round-globals/

Note to self: squash the heck out of this PR if/when it merges in, but not before pulling out the global variable workaround commit for `stages.inc` for a temp commit into `pgdp-production` so we can fix `noncvs` async from when this gets deployed.

70ray (approved):

cpeel:
To make testing this easier, I've:
1. reduced this PR to just the new classes and the globals recreate patch
2. added @70ray's excellent RoundTest unittest from his work in this area

The updated sandbox is at https://www.pgdp.org/~cpeel/c.branch/remove-round-globals/

The `get_Round_*` accessor functions which use the new access functions are used on the project page and on the page details pages so if those look like TEST this PR should be good to go (thanks in large part to Ray's unittest).

After this goes in I'll put up a set of smaller PRs to replace the globals in batches so it's easier to review & test.

srjfoo (approved):

PR #1010, opened 2024-01-06
Allow image to scroll beyond edges

70ray:
This allows the image to scroll to halfway across the image to left and right and 80% of the way upwards.
These amounts could be changed and it could be made to work differently for vertical and horizontal layouts.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/image_overscroll

### DIFF
--- a/scripts/control_bar.js
+++ b/scripts/control_bar.js
@@ -257,7 +257,6 @@ function makeControlDiv(container, content, controls) {
 
 function makeImageWidget(container, align = "C") {
     const content = document.createElement("div");
-    content.style.scrollbarWidth = 'none';
     const grabCursor = "grab";
     // use plain js image so width or style.width is clearly differentiated
     const image = document.createElement("img");
@@ -351,6 +350,11 @@ function makeImageWidget(container, align = "C") {
         yOffset += vertOffset;
         // image rotates about centre. Offset moves it to correct position
         image.style.transform = `matrix(${cosine}, ${-sine}, ${sine}, ${cosine}, ${xOffset}, ${yOffset})`;
+
+        // show scrollbar only if image is bigger than window
+        content.style.overflowX = (imageWidth > contentWidth) ? "auto" : "hidden";
+        // seems to need +1 to hide bar when fitting to height
+        content.style.overflowY = (imageHeight > contentHeight + 1) ? "auto" : "hidden";
     }
 
     function initScroll() {
--- a/scripts/control_bar.js
+++ b/scripts/control_bar.js
@@ -346,8 +346,6 @@ function makeImageWidget(container, align = "C") {
         image.style.transform = `matrix(${cosine}, ${-sine}, ${sine}, ${cosine}, ${xOffset}, ${-yOffset})`;
     }
 
-    image.addEventListener("load", setImageStyle);
-
     function initScroll() {
         // assuming width is 2*contentWidth
         if(align == "C") {
@@ -359,6 +357,13 @@ function makeImageWidget(container, align = "C") {
         }
         content.scrollTop = 0;
     }
+
+    function initAll() {
+        setImageStyle();
+        initScroll();
+    }
+
+    image.addEventListener("load", initAll);
 
     function setZoom() {
         if(percent < minPercent) {
@@ -434,16 +439,14 @@ function makeImageWidget(container, align = "C") {
         .append($("<i>", {class: 'fas fa-redo-alt'}))
         .click( function () {
             [sine, cosine] = [-cosine, sine];
-            setImageStyle();
-            initScroll();
+            initAll();
         });
 
     const counterclockRotateInput = $("<button>", {title: texts.counterclockRotate})
         .append($("<i>", {class: 'fas fa-undo-alt'}))
         .click( function () {
             [sine, cosine] = [cosine, -sine];
-            setImageStyle();
-            initScroll();
+            initAll();
         });
 
     const controls = [
@@ -461,8 +464,7 @@ function makeImageWidget(container, align = "C") {
     const controlDiv = makeControlDiv(container, $(content), controls);
     // content width can change when moving controls
     controlDiv.onChange.add(function () {
-        setImageStyle();
-        initScroll();
+        initAll();
     });
 
     return {
@@ -483,7 +485,6 @@ function makeImageWidget(container, align = "C") {
             sine = 0;
             cosine = 1;
             image.src = src;
-            initScroll();
         },
 
         reScroll: reScroll,
--- a/scripts/control_bar.js
+++ b/scripts/control_bar.js
@@ -358,8 +358,8 @@ function makeImageWidget(container, align = "C") {
         // (2 * windowWidth): half left to half right,
         // (2 * windowWidth + imageWidth): scroll to far edges,
         // similarly for vertical
-        contentWidth = parseFloat(getComputedStyle(content).width);
-        contentHeight = parseFloat(getComputedStyle(content).height);
+        contentWidth = content.clientWidth;
+        contentHeight = content.clientHeight;
         image.style.width = `${10 * percent}px`;
         image.style.height = "auto";
         let imageHeight, xOffset, yOffset, imDivHeight;
--- a/scripts/control_bar.js
+++ b/scripts/control_bar.js
@@ -391,7 +391,7 @@ function makeImageWidget(container, align = "C") {
 
     function reScroll () {
         // keep same scroll proportion when contentWidth changes
-        let oldScroll = content.scrollLeft / contentWidth;
+        let oldScroll = (contentWidth > 20) ? content.scrollLeft / contentWidth : 0.5;
         setImageStyle();
         content.scrollLeft = oldScroll * contentWidth;
     }
--- a/scripts/control_bar.js
+++ b/scripts/control_bar.js
@@ -332,7 +332,7 @@ function makeImageWidget(container, align = "C") {
     let cosine = 1;
 
     const percentInput = $("<input>", {type: 'number', value: percent, title: texts.zoomPercent});
-    let contentWidth, imageWidth, imDivWidth, vertOffset;
+    let contentWidth, contentHeight, imageWidth, imDivWidth, vertOffset;
 
     function setImageStyle() {
         // To allow scrolling when image does not overflow the window
@@ -343,7 +343,7 @@ function makeImageWidget(container, align = "C") {
         // (2 * windowWidth + imageWidth): scroll to far edges,
         // similarly for vertical
         contentWidth = parseFloat(getComputedStyle(content).width);
-        const contentHeight = parseFloat(getComputedStyle(content).height);
+        contentHeight = parseFloat(getComputedStyle(content).height);
         image.style.width = `${10 * percent}px`;
         image.style.height = "auto";
         let imageHeight, xOffset, yOffset, imDivHeight;
@@ -402,19 +402,23 @@ function makeImageWidget(container, align = "C") {
             percent = maxPercent;
         }
         percentInput.val(Math.round(percent));
-        setImageStyle();
     }
 
     function setDrawSave() {
         setZoom();
+        setImageStyle();
         localStorage.setItem(imageKey, JSON.stringify({zoom: percent}));
     }
 
     function reScroll () {
-        // keep same scroll proportion when contentWidth changes
-        let oldScroll = (contentWidth > 20) ? content.scrollLeft / contentWidth : 0.5;
+        // keep image stationary when window changes size
+        // x is distance of left edge of image to left of window etc.
+        let x = contentWidth - content.scrollLeft;
+        let y = contentHeight - content.scrollTop;
         setImageStyle();
-        content.scrollLeft = oldScroll * contentWidth;
+        // contentWidth and height have changed
+        content.scrollLeft = contentWidth - x;
+        content.scrollTop = contentHeight - y;
     }
 
     percentInput.change(function() {
@@ -510,6 +514,8 @@ function makeImageWidget(container, align = "C") {
             setZoom();
             controlDiv.setupControls(imageWidgetKey);
         },
+
+        initAll,
 
         setImage: function (src) {
             sine = 0;
--- a/scripts/control_bar.js
+++ b/scripts/control_bar.js
@@ -316,22 +316,26 @@ function makeImageWidget(container, align = "C") {
     let cosine = 1;
 
     const percentInput = $("<input>", {type: 'number', value: percent, title: texts.zoomPercent});
-    let contentWidth, imageWidth;
+    let contentWidth, imageWidth, imDivWidth, vertOffset;
 
     function setImageStyle() {
-        // To allow scrolling beyond edges make imageDiv bigger.
-        // half left to half right: 2 * contentWidth
-        // to get .8 of way up: contentHeight + .8 * imageHeight
+        // To allow scrolling when image does not overflow the window
+        // make imageDiv bigger than image, different options are:
+        // windowWidth: no extra scrolling,
+        // (2 * windowWidth - imageWidth): scroll to edges,
+        // (2 * windowWidth): half left to half right, (using this)
+        // (2 * windowWidth + imageWidth): scroll to far edges,
+        // similarly for vertical
         contentWidth = parseFloat(getComputedStyle(content).width);
         const contentHeight = parseFloat(getComputedStyle(content).height);
         image.style.width = `${10 * percent}px`;
         image.style.height = "auto";
-        let imageHeight, xOffset, yOffset;
+        let imageHeight, xOffset, yOffset, imDivHeight;
         if(sine != 0) {
             // rotated 90 or 270 degrees
             imageWidth = image.height;
             imageHeight = image.width;
-            yOffset = (image.height - image.width) / 2;
+            yOffset = - (image.height - image.width) / 2;
         } else {
             // rotated 0 or 180 degrees
             imageWidth = image.width;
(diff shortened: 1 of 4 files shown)


chrismiceli (approved):
Initial opinion using the sandbox, my remembered preference for scroll position put the image halfway not visible. Do we need to migrate the saved local storage data?

70ray:
> Initial opinion using the sandbox, my remembered preference for scroll position put the image halfway not visible. Do we need to migrate the saved local storage data?

Thanks for testing. That shouldn't happen. I'll look into it.

70ray:
Sandbox updated.

cpeel:
This is _very_ cool. It functions the way I was hoping with one bug: I can pan the image past the left, top, and right sides of the viewport but not the bottom.

srjfoo:
> This is _very_ cool. It functions the way I was hoping with one bug: I can pan the image past the left, top, and right sides of the viewport but not the bottom.

From reading the code, I assumed that was intentional. I would put it differently -- I can scroll past the bottom of the page, but not above the top of the page.

There's other thing that I'm not quite sure how to describe. The amount of pan room seems to be dependent on the height or width of the image pane, not the image (or maybe it has to do with the relationship between the image and the image pane). For example: if I resize the proofing window to make it shorter, but don't resize the image correspondingly, I can't pan as far in any direction. If I'm in horizontal mode and I make the image the full width of the image pane, I can't scroll past either the top or the bottom.

If I resize the image to fit the width of the image pane (in horizontal mode), I can scroll the correct percentages left and right, but can't scroll past the top or the bottom of the page. Similarly, if I am silly enough to resize the image to fit the height of the pane in horizontal mode, the top and bottom match the coded percentages pretty well, but left and right are pretty far off from the correct percentages (although I think the point is to scroll the image halfway off the screen. It does slightly more than halfway off the screen, but close enough).

Let me know if my descriptions don't make sense, and I can provide screenshots of what I'm seeing.

70ray:
> This is _very_ cool. It functions the way I was hoping with one bug: I can pan the image past the left, top, and right sides of the viewport but not the bottom.

As Sharon said, that was intentional, but it can easily be changed. Perhaps the simplest option would be for the same proportions  horizontal or vertical.

70ray:
> There's other thing that I'm not quite sure how to describe.

While experimenting with it I found that if you zoom image 200 or 300% you can't scroll right to the edges. I'm not sure if this is the same problem you were seeing but I'll try to fix it.

srjfoo:
> > There's other thing that I'm not quite sure how to describe.
> 
> While experimenting with it I found that if you zoom image 200 or 300% you can't scroll right to the edges. I'm not sure if this is the same problem you were seeing but I'll try to fix it.

I didn't try 200-300%, but just fitting the width to the image pane. Under those circumstances, I could scroll right to the edge (top and bottom).

In the clearer light of day, maybe it makes the most sense to just document that the image can be scrolled to a certain percentage of the height if the image is fit to the height of the image pane, and a certain percentage of the width if the image is fit to the width of the image pane, but is proportional under other circumstances.

70ray:
It should work the same way for horizontal and vertical now and always scroll to the edges: Perhaps it could be described in this way: The image can be scrolled in any direction until only half of it is visible, but if the half is still bigger than the window then it can be scrolled further right to the edge.
The "half" is just arbitrarily chosen based on Casey's use case for a text in two columns. It could be changed to some other proportion. The comments in the code suggest how this would work.
Sanbox updated.

cpeel:
This is closer but I think it needs to account for the zoom level. If I'm zoomed in it still limits panning to the top and the bottom of the image. For testing I went to [this project](https://www.pgdp.org/~rp31/c.branch/image_overscroll/project.php?id=projectID4c49f39aa829c&expected_state=F1.proj_avail), proofread a page, hit the `<->` button to zoom in and then try to pan the image up and down.

srjfoo:
> The image can be scrolled in any direction until only half of it is visible, but if the half is still bigger than the window then it can be scrolled further right to the edge.

Horizontal interface:
1. Fit to width: does not scroll past top and bottom edges of image; left and right scroll is fine.
2. Fit to height: all scrolling past the edge works as I would expect. I.e. I can see half the image in height and width if I scroll all the way to the top, bottom or either side.

Vertical interface:
1. Fit to width: works as expected (see point 2 above)
2. Ditto fit to height.

It just seems to be scrolling past the top and bottom in the horzontal mode that causes the problem. And I'm not quite sure how to describe the circumstances. It seems to have something to do with the height of the image pane as compared with the actual image height? I can't tell for sure; I haven't analyzed it well enough.

70ray:
It seems to be working as intended. It can be changed to make it scroll until the image is almost off the screen, but why would anyone want to do this?

srjfoo:
> It seems to be working as intended. It can be changed to make it scroll until the image is almost off the screen, but why would anyone want to do this?

In the Standard interface, proofers can dynamically change the height of the image pane, so that if they're proofreading in the horizontal interface, the image height can be narrowed to the point that they only have two or three lines visible, which can help focus if the text is particularly dense. This is not an option in the Enhanced interface, so I can see a proofer wanting to scroll the image such that the top or bottom of the image is as close to the corresponding lines of text as it can get (this would most likely be the top of the page, probably).

70ray:
> ... so I can see a proofer wanting to scroll the image such that the top or bottom of the image is as close to the corresponding lines of text as it can get (this would most likely be the top of the page, probably).

That seems a good reason. I've changed it so it can scroll to within 60 pixels of the edge. This can be changed easily.

scripts/control_bar.js
```diff
@@ -322,28 +316,64 @@ function makeImageWidget(container, align = "C") {
...
+        // adjust yOffset to be able to scroll downwards
+        vertOffset = Math.max(0.5 * (imDivHeight - imageHeight), 0);
+        yOffset += vertOffset;
         // image rotates about centre. Offset moves it to correct position
-        image.style.transform = `matrix(${cosine}, ${-sine}, ${sine}, ${cosine}, ${offset}, ${-offset})`;
+        image.style.transform = `matrix(${cosine}, ${-sine}, ${sine}, ${cosine}, ${xOffset}, ${yOffset})`;
+    }
+
+    function initScroll() {
+        if(align == "C") {
```
chrismiceli, line 359: since `"C"` is used twice, should we keep a constant for it at least? Even if we never reference any others such as left or right anymore?

scripts/control_bar.js
```diff
@@ -359,6 +389,13 @@ function makeImageWidget(container, align = "C") {
         localStorage.setItem(imageKey, JSON.stringify({zoom: percent}));
     }
 
+    function reScroll () {
+        // keep same scroll proportion when contentWidth changes
+        let oldScroll = content.scrollLeft / contentWidth;
```
chrismiceli, line 394: I know it is an edge case, but do we need to worry about if contentWidth is 0? Say if someone drags too much on accident? I don't want to get in an exception case and the user is in some bad state.

chrismiceli (approved):

scripts/control_bar.js
70ray: I suppose we should at least check that the value provided is in the set we have allowed for.

scripts/control_bar.js
70ray: I'll look into this.

scripts/control_bar.js
70ray: I don't think we need to change this. Nothing bad can happen.

scripts/control_bar.js
70ray: Caught this case.

70ray:
Sandbox updated.

srjfoo (approved):
Looks like I got distracted on this one, too -- my apologies.

srjfoo:
This is going to be a fairly obvious change -- should it be offered for public review before final approval?

cpeel (approved):
This functions beautifully.

It's probably a good idea that we start a thread in Future Features to give folks a heads up about this change.

chrismiceli (approved):

srjfoo:
> Would it be possible to make the background draggable? That way if the image somehow ended up in a corner the drag target isn't 50x50px box? I suppose you can also use the "fit to x" options too to fix this situation.

I just poked at it a bit, and I think it's probably quicker and easier to just push one of the buttons that will recenter it. However, that might be a good question to ask potential users.

70ray:
> Would it be possible to make the background draggable? That way if the image somehow ended up in a corner the drag target isn't 50x50px box? I suppose you can also use the "fit to x" options too to fix this situation.

Done that now.

70ray:
The scroll bars are now turned off and drag outside the image enabled as Chris suggested.

Sandbox updated.

cpeel (approved):
This works brilliantly.

srjfoo (approved):

chrismiceli:
There is a small blue border in firefox when clicked that probably can be removed:
image

I also worry some people may complain there are no scrollbars anymore, so may be worth running by the community.

70ray:
This commit leaves the focus in the text area even when dragging the image. The blue border Chris mentioned indicates that the image has focus so this has now gone.

70ray:
> > This commit leaves the focus in the text area even when dragging the image. The blue border Chris mentioned indicates that the image has focus so this has now gone.
> > Sandbox updated.
> > (The focus() call worked differently in different browsers. This is probably because it is an HTMLelement method and I was calling it incorrectly on a JQuery object)
> 
> Thanks for fixing the blue border.
> 
> As for restoring focus to textarea after dragging, I'm kind of against that from an accessibility point of view. I don't think we should be doing anything fancy with focus as it could confuse a user who would expect the text area to lose focus when clicking something else. We should let the browser handle focus normally I would think. The current prod behavior also doesn't have this, is it necessary for the image overscroll?

All focus manipulations have been removed by the last commit.

chrismiceli:
> > > This commit leaves the focus in the text area even when dragging the image. The blue border Chris mentioned indicates that the image has focus so this has now gone.
> > > Sandbox updated.
> > > (The focus() call worked differently in different browsers. This is probably because it is an HTMLelement method and I was calling it incorrectly on a JQuery object)
> > 
> > 
> > Thanks for fixing the blue border.
> > As for restoring focus to textarea after dragging, I'm kind of against that from an accessibility point of view. I don't think we should be doing anything fancy with focus as it could confuse a user who would expect the text area to lose focus when clicking something else. We should let the browser handle focus normally I would think. The current prod behavior also doesn't have this, is it necessary for the image overscroll?
> 
> All focus manipulations have been removed by the last commit.

Is the sandbox updated? I still see the focus stay in the text area after dragging the image around.

70ray:
> Is the sandbox updated? I still see the focus stay in the text area after dragging the image around.

Yes, this seems to be what the users want. I had originally (before this branch) tried to set the focus to the image when clicking on it so that the arrow keys would pan the image. But it seems I did not do it correctly because it only worked like that in some browsers.
Users could pan the image by using the scroll bars which did not move the focus out of the text, but now the scrollbars have gone they could no longer do this.

chrismiceli:
> > Is the sandbox updated? I still see the focus stay in the text area after dragging the image around.
> 
> Yes, this seems to be what the users want. I had originally (before this branch) tried to set the focus to the image when clicking on it so that the arrow keys would pan the image. But it seems I did not do it correctly because it only worked like that in some browsers. Users could pan the image by using the scroll bars which did not move the focus out of the text, but now the scrollbars have gone they could no longer do this.

That makes sense, especially the part about dragging with the scroll bars without losing focus now being lost. I just always get worried when manipulating natural events like tab and focus.

srjfoo:
I poked at the sandbox last night and noticed that the page did not come up in the proofreading interface with focus already in the text. I had to put it there. Is that expected? (Once it was there, it stayed there.) FF, latest version (122.0.1)

70ray:
> I poked at the sandbox last night and noticed that the page did not come up in the proofreading interface with focus already in the text. I had to put it there. Is that expected? (Once it was there, it stayed there.) FF, latest version (122.0.1)

It doesn't in the master branch either, but it is possible to make it so it does.

70ray:
Considering the request for just the vertical scrollbar to be retained, the setting `style.scrollbarWidth = 'none'` applies to both scroll bars. Instead we could set `overflow-x: hidden; overflow-y: auto;`. This would keep the vertical scroll bar only; the drag would still work and scroll wheel for vertical scroll but shift-scrollwheel would no longer work for horizontal scrolling.

srjfoo:
> Considering the request for just the vertical scrollbar to be retained, the setting `style.scrollbarWidth = 'none'` applies to both scroll bars. Instead we could set `overflow-x: hidden; overflow-y: auto;`. This would keep the vertical scroll bar only; the drag would still work and scroll wheel for vertical scroll but shift-scrollwheel would no longer work for horizontal scrolling.

I'm concerned that having only one set of scrollbars would be confusing. The places I see that are literally in places where you can scroll in one direction, but not the other, so I'm inclined to say "all or nothing". (My personal preference is for no scrollbar, honestly. It looks cleaner.

https://caniuse.com/mdn-css_properties_scrollbar-width shows limited coverage for `scrollbar-width` (which explains why the change didn't work in Opera or Safari), and although it's been in FF since v.64, that's still not as low as our minimum supported browser. And it's only recently supported in Chrome and Edge.

srjfoo:
With his latest commit, @70ray has made a change so that scrollbars show only if the image is larger in either or both dimensions than the image pane that it's in.

I think we're close to being ready to open this up for general evaluation again, but before we do, I'd like to see another round of as broad a range as possible of browser testing on different flavors of browsers being actively used among us as we can.

There's a system-level setting in macOS for how scrollbars are handled. From doing a web search, there should also be one for at least Win 10 and 11. It would be good to have confirmation of this, because that may be important information for other users.

If anyone has access to Win 7, that's also a good place to test: I'm pretty sure we have some people still using Win 7, though I don't know how many.

I tried some testing on browserstack, and whether it's because their OS settings are different (maybe I just didn't know how, but I couldn't seem to check the OS settings), but comparing against my current OS and the browser versions I'm using, I got different results, so I don't completely trust the browserstack results on this.

On my main macOS machine, running Sonoma, and depending on the OS settings for scrollbars, everything works to my satisfaction in FF, Chrome, Opera and Safari. Where there isn't a scrollbar, I can't scroll, but I can click and drag the image in any direction if I need to. With a keyboard and stylus, I'm satisfied that it should be okay on iPads using those tools.

@wfarrell, @lhamilton1, (and indeed, anyone who has a tablet of any kind) can y'all give feedback especially on the iPad. I've talked to another DPer who proofs and formats on the iPad. They use a keyboard, and a stylus in place of mouse. I tried it with a track-pad, and it was pretty miserable, but the stylus seems to work well.

theshriek (approved):

windymilla (approved):
Tested fine on Chrome 124.0.6367.91 (Windows 11).
In Chrome on Android phone, I don't think the image grab & drag works. You can scroll horizontally and vertically by dragging when there are scrollbars, but if the image is small and there are no scrollbars, but part of the image is off-screen, it is not possible to drag the image back on-screen. I don't know if this will turn out to be a general issue for devices that don't use a mouse, as I noticed the code was listening for mouse events. It's not a major issue, because it is easy to zoom in a bit, then drag to where you want it, and realistically, if an image is on a phone but is so small that scrollbars are not needed, it's unlikely that you would do serious proofing.

ellinora (approved):

wf49670 (commented):
I have not yet looked at the code. However, I've played in the sandbox a bit using my iPad. 

Results are not pleasant if in Landscape mode and Enhanced Interface. Seems OK if in Portrait mode, but with only minimal testing so far.

In Landscape mode, it's fine if I use the trackpad to try to move the image, but if I touch the screen with finger or Apple Pencil 2 the image immediately shifts all the way to the right. At that point it can only be recovered by using the <-> button at the top of the window.

wf49670 (changes requested):
I'm not sure how important iPad support is, or how important it is to use without a keyboard (such that using a finger or Apple Pencil is required), but it would probably be better to try to fix this before integrating the code.

70ray:
> I'm not sure how important iPad support is, or how important it is to use without a keyboard (such that using a finger or Apple Pencil is required), but it would probably be better to try to fix this before integrating the code.

It is probably fixable if we add handlers for `touch` events in the same way as is done in splitControl.js. I'll try it.

windymilla:
> > I'm not sure how important iPad support is, or how important it is to use without a keyboard (such that using a finger or Apple Pencil is required), but it would probably be better to try to fix this before integrating the code.
> 
> It is probably fixable if we add handlers for `touch` events in the same way as is done in splitControl.js. I'll try it.

That would maybe fix the Android phone thing too, I guess 🤞

70ray:
> > > I'm not sure how important iPad support is, or how important it is to use without a keyboard (such that using a finger or Apple Pencil is required), but it would probably be better to try to fix this before integrating the code.
> > 
> > 
> > It is probably fixable if we add handlers for `touch` events in the same way as is done in splitControl.js. I'll try it.
> 
> That would maybe fix the Android phone thing too, I guess 🤞

Yes, I'll be able to try that.

70ray (commented):
Just noticed another unfortunate feature: When changing between vertical and horizontal in the page browser, the image goes into the corner.

srjfoo:
The below followup on Walt's testing is pretty long, and provides context, but I wanted to mention in a more visible place that I noticed something that I hope may help figure out the issue. In the two browsers that have problems, the problem occurs only if the width of the image is either smaller than or equal to the width of the image pane. If I increase the width by as little as 1%, I can move the image around by finger, stylus or Apple Pencil.

> Results are not pleasant if in Landscape mode and Enhanced Interface. Seems OK if in Portrait mode, but with only minimal testing so far.
> 
> In Landscape mode, it's fine if I use the trackpad to try to move the image, but if I touch the screen with finger or Apple Pencil 2 the image immediately shifts all the way to the right. At that point it can only be recovered by using the <-> button at the top of the window.

Starting with Walt's notes here, instead of bouncing back and forth between browsers and trying to remember exactly what happened, and where, I finally gave in and set up testing notes. I tested all three of Safari, Chrome and Firefox, with the iPad in both vertical and horizontal orientations.

Testing done on second generation 10.5 inch iPad Pro with Apple Pencil (first gen), non-Apple stylus (that doesn't need to be charged!), or finger. I gave up on my Magic Trackpad; it's either too old, or just doesn't work well with my iPad, and was a disaster when I was testing earlier.

I saw no issues in the Enhanced interface. Either with the iPad vertical or horizontal, and in either vertical or horizontal versions of the proofreading interface, all three browsers would allow me to move the image back and forth, up and down, with finger, stylus or Apple Pencil.

In the Standard interface, Safari seemed to work okay, across the board. In Firefox and Chrome, iPad in vertical orientation, both horizontal and vertical interfaces worked fine.

With the iPad in the horizontal orientation, the vertical interface was fine, but the horizontal interface works as Walt describes: if I try to scroll/move the image with finger, stylus or Apple Pencil, it scoots over to the right-hand edgi. I can move it up and down with finger, stylus or Apple Pencil, but I can't get the left edge of the image away from the right edge of the image pane except by using the buttons to re-center, and even then if I try to move it, the left edge of the image scoots back over to the right edge of the image pane (or as far as it can go, anyway). (Possible work-around noted above.)

70ray:
I'll get back to it towards the end of next week.

70ray:
Touch scroll should now work.
Switching between horizontal and vertical split in the page browser should work the same as if refreshing the page; the image should not almost disappear into a corner.
When changing the window size the image should keep still.

windymilla:
Dragging looks good on my Android phone now. Thanks

70ray:
There could be a way to fix oustanding issues of centering and ipad.

srjfoo:
> There could be a way to fix oustanding issues of centering and ipad.

@70ray, can you describe exactly what the iPad issue is that you're seeing, and how to reproduce it? I watched three different people proofread on three different iPad models, and I've done so on mine, and a couple on browserstack, as well, and the only thing I've seen is the left alignment of the image (and that only on my own iPad and on browserstack, and I can't reproduce anything. (It's hard to tell exactly what version I'm using on browserstack, honestly.)

If the problem is as subtle as this seems to be, we may need more testing than we're likely to get among our normal testers and devs.

70ray:
> > There could be a way to fix oustanding issues of centering and ipad.
> 
> @70ray, can you describe exactly what the iPad issue is that you're seeing, and how to reproduce it? I watched three different people proofread on three different iPad models, and I've done so on mine, and a couple on browserstack, as well, and the only thing I've seen is the left alignment of the image (and that only on my own iPad and on browserstack, and I can't reproduce anything. (It's hard to tell exactly what version I'm using on browserstack, honestly.)
> 
> If the problem is as subtle as this seems to be, we may need more testing than we're likely to get among our normal testers and devs.

Some other changes I made might have fortuitously fixed it but I'll get back to you on this.

70ray:
The image should now remain centred for compatiblity with the way the satndard interface works.
The iPad issue has gone away.
Sandbox updated.

70ray:
Sandbox updated.

cpeel:
Many thanks to @70ray for his work on this and patience as we worked through various browser / device oddities. 🎉

Issue #995, opened 2023-12-25
Revisit the new user experience


cpeel:
In an attempt to help a new user navigate the site, the code shows, or hides, information based on the number of pages the user has proofread in P1 (technically the ELR -- entry level round -- but I'll use P1 in this issue). While the logic to get the number of P1 pages is centralized (`get_pages_proofed_maybe_simulated()`), how this is used as a threshold varies pretty widely.

These are the current thresholds, what happens at them and on what pages:

`>= 10`
* will show random rule -- round page

`>= 15 && < 200`
* will ask proofreaders about their mentor feedback -- activity hub & round page

`> 20`
* will show release queues -- round page
* will show filter block -- round page
* will show filtered projects on progress snapshot -- activity hub
* will show filtering links on progress snapshot -- activity hub

`> 80 && <= 100`
* will show message about simple proofreading rules going away soon -- round page

`< 100`
* will show the string "Activities" in the navbar before the list of activities -- navbar
* will show user block to point them to P1 -- search page
* will show user block telling them to read project comments and hit start proofreading -- project page

`<= 100`
* will show simple proofreading rules -- round page
* will redirect user to ELR or tell them to select a beginner project -- activity hub & round page

`< 300`
* will tell a user about an unread PM -- activity hub & round page

`<= 300`
* will show beginner help in progress snapshot -- activity hub

Note that a proofreader needs to proof 300 pages in P1 before being eligible for P2, that's presumably where the 300 is from, but they are defined in two separate places.

We should -- at a minimum -- standardize on 3 or 4 standard cutoffs, eg: 50, 100, 300.

What would be better is if we conceptualized what it is we're trying to do here and create some centralized code that implements it. For instance, perhaps we break the new-user experience down into 3 levels that they "graduate" from:
1. `guided` steps through the site
2. `simple` versions of certain information or excluding some information
3. `final` last few things before we stop treating them as a beginner

We could then have functions that do gradual reveals map to *one* of those levels and a centralized function that can accept a query with a level and return true/false if the user is in that level. Then we have one place we can define what those levels are.

A more concrete example:
```php
# in gradual.inc
function user_is_tenure_level($level, $username = null)
{
    if ($username === null) {
        $username = User::current_username;
    }

    $LEVEL_BREAKPOINTS = [
        "guided" => 50,
        "simple" => 100,
        "final" => 300,
    ];

    $pages = get_pages_proofed_maybe_simulated($username);
[...]

PR #1011, opened 2024-01-08
Standardize new proofreader page thresholds

cpeel:
This provides some minor standardization for our new proofreader cutoffs so we have fewer of them. This is a minor improvement for issue #995 but is not a fix.

Not putting up a sandbox as this is a pretty simple review via PR.

### DIFF
--- a/activity_hub.php
+++ b/activity_hub.php
@@ -128,7 +128,7 @@ $show_filtering_links = true;
 
 // Proofreaders with fewer than 21 pages can't see the filter box on the Round
 // pages so prevent those users from selecting the filtered option.
-if ($pagesproofed <= 20) {
+if ($pagesproofed < 20) {
     $show_filtered_projects = false;
     $show_filtering_links = false;
 }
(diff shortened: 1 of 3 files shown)


activity_hub.php
```diff
@@ -128,7 +128,7 @@
 
 // Proofreaders with fewer than 21 pages can't see the filter box on the Round
 // pages so prevent those users from selecting the filtered option.
-if ($pagesproofed <= 20) {
+if ($pagesproofed < 20) {
```
chrismiceli, line 131: These constants (20, 100, 300) sprinkled throughout multiple files makes me want to ask if we can put these in a shared location to make updating them in the future easier?

chrismiceli (approved):

activity_hub.php
cpeel: I'm hoping to resolve that more holistically via #995 -- right now there's not a lot of cohesion and I can't come up with constant names that actually describe them as they are all very bespoke. This PR is decidedly only a bandaid.

srjfoo (approved):

windymilla (approved):

PR #1012, opened 2024-01-08
Refer new users to My Suggestions page

cpeel:
This does 2 things:
* Update the new proofreader message (<100 pages proofread) that gets shown on the Search page to point them to the My Suggestions page.
* Update the post-quiz pass logic to automatically grant them access to the round if they are eligible for auto-grant -- this is what already happens if they pass the quiz and then go to the round page. Doing this here allows us to refer them to the My Suggestions page in addition to the round page and ensure they can work in both when they get there.

Sandbox at https://www.pgdp.org/~cpeel/c.branch/finesse-new-proofreader-messages/

Note: The sandbox has the `pgdp-production` commit that requires passed quizzes to work in P1 so that this can be tested easily.

### DIFF
--- a/pinc/Quiz.inc
+++ b/pinc/Quiz.inc
@@ -143,14 +143,25 @@ class Quiz
                 // if the user can't currently access it but they have satisfied
                 // all of the requirements to do so, point them in the right direction
                 if (!$uao->can_access && $uao->all_minima_satisfied) {
+                    // first see if we can auto-grant them access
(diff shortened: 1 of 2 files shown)


chrismiceli (approved):

windymilla (approved):
Behavior as expected - code LGTM

srjfoo (approved):
Whoops -- sorry I lost track!

70ray (approved):

PR #1013, opened 2024-01-09
Escape warnings from WordCheck

cpeel:
Escape warnings from WordCheck before outputting them to the user.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/escape-warnings/


windymilla (approved):
Code looks good (of course). Tested removing all dictionaries and got a correct warning about it in sandbox.

srjfoo (approved):

PR #1016, opened 2024-01-11
Validate chunk size in resumable uploads

cpeel:
This addresses two edgcases in an attempt to fix the upload problems we're seeing:
* Only ever return a single HTTP response code in GET requests
* Validate every chunk upload against its expected size.

To handle the edgecase where a file is fully uploaded but not reassembled, we attempt reassembly both on testChunks (GETs) as well as the uploads (POSTs). The original code responded with a 200 that the chunk exists but the reassembly could have failed and not reported a 5xx code. Now the code sets one and only one return code, or falls through to the script completion which is an implicit 200.

Also validate that every chunk is at least chunkSize to attempt and detect cases where the client didn't successfully uploading an entire chunk.

This PR is best reviewed ignoring whitespace as I moved some code into a function which increased its indent.

Testable in https://www.pgdp.org/~cpeel/c.branch/more-resumable-upload-checks/

I don't have confidence that this addresses the problems that we're seeing, but these were the two corner cases I found from code inspection.

### DIFF
--- a/tools/upload_resumable_file.php
+++ b/tools/upload_resumable_file.php
@@ -58,8 +58,9 @@ elseif ($_SERVER['REQUEST_METHOD'] === 'POST') {
         if (!move_uploaded_file($file['tmp_name'], $chunk_filename)) {
             report_error("Error saving chunk $chunk_filename for $filename");
             exit;
-        } elseif (filesize($chunk_filename) < $chunk_size) {
-            // if the filesizes don't match, unlink the file and report an error
+        } elseif (filesize($chunk_filename) < $chunk_size && $total_size > $chunk_size) {
+            // if the filesize is smaller than expected, unlink the file
+            // and report an error
             report_error("Error saving chunk $chunk_filename for $filename, uploaded size did not match $chunk_size");
             unlink($chunk_filename);
             exit;
--- a/tools/upload_resumable_file.php
+++ b/tools/upload_resumable_file.php
@@ -44,7 +44,7 @@ if ($_SERVER['REQUEST_METHOD'] === 'GET') {
 elseif ($_SERVER['REQUEST_METHOD'] === 'POST') {
     foreach ($_FILES as $file) {
         if ($file["error"] != UPLOAD_ERR_OK) {
-            report_error(get_upload_err_msg($file['error']), false);
+            report_error(get_upload_err_msg($file['error']));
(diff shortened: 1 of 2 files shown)


srjfoo (approved):
About all I knew to do was to upload something to verify that nothing's broken. Rick's problems with his uploads are intermittent, and generally succeed on the second try if they fail the first time.

If you can think of some way to test the edge cases, let me know; I'm happy to do more testing.

70ray (approved):

70ray (commented):
I got an error testing:
Upload progress: Upload failed
Error saving chunk /tmp/resumable_uploads/bdbe374e5efb621b062a0e76448e08f0dir/bdbe374e5efb621b062a0e76448e08f0.part.1 for anessay.zip, uploaded size did not match 1048576.

windymilla (approved):

70ray (changes requested):
Fails in some cases

cpeel:
Thanks for that testing, @70ray. I'm hoping my last commit will address it by telling resumable.js to retry failed chunk uploads instead of bailing.

tools/upload_resumable_file.php
```diff
@@ -23,86 +24,119 @@
...
 
-    if (!is_dir($staging_dir)) {
-        // Suppress warning by mkdir because despite the check above
-        // race conditions from other processes can result in the
-        // directory existing when we try to create it here.
-        @mkdir($staging_dir, 0777, true);
+        if (!move_uploaded_file($file['tmp_name'], $chunk_filename)) {
+            report_error("Error saving chunk $chunk_filename for $filename");
+            exit;
+        } elseif (filesize($chunk_filename) < $chunk_size) {
```
70ray, line 61: I think it is this line that is causing the problem I am seeing. If the whole file size is less than the chunk size it will always fail here.

cpeel: Yup. I was able to reproduce this too. Pushed up a new commit to handle this case.

chrismiceli (approved):

cpeel:
The last commit ignores the size of the last chunk entirely as it can be both smaller and larger than the chunk size in some cases. This is the version that okrick tested and seemed to work for him in the dev sandbox.

srjfoo (approved):

PR #1017, opened 2024-01-12
Enable filtering page details API by round

cpeel:
The [Gap Games](https://www.pgdp.net/wiki/Gap_Game) use the DP API to pull some information from the `v1/projects/:projectid/pagedetails` endpoint (which is the same underlying function that renders the Page Details table for a project). This returns data for every round which
1. is a lot of data if you only need one round of it
2. is a pretty expensive DB query for later-round projects with lots of pages

This PR allows the user to just request the round(s) they care about making the data smaller and the SQL simpler. By default the endpoint still returns all rounds. The new `pageroundid` query param works similarly to the [`field` query param for the `v1/projects` endpoint](https://github.com/DistributedProofreaders/dproofreaders/blob/master/api/dp-openapi.yaml#L79) by allowing multiple to be specified if desired.

This is testable in https://www.pgdp.org/~cpeel/c.branch/api-pagedetail-single-round/ -- confirm that a project's page details hasn't changed -- and via curl, eg:
```bash
API_KEY=xxxxxxx  # your API key
PROJECTID=projectID4c49f39aa829c

# all rounds
curl -i -X GET \
    -H "X-API-KEY: $API_KEY" \
    "https://www.pgdp.org/~cpeel/c.branch/api-pagedetail-single-round/api/index.php?url=v1/projects/$PROJECTID/pagedetails"

# just P3
curl -i -X GET \
    -H "X-API-KEY: $API_KEY" \
    "https://www.pgdp.org/~cpeel/c.branch/api-pagedetail-single-round/api/index.php?url=v1/projects/$PROJECTID/pagedetails&pageroundid=P3"

# just OCR and F2
curl -i -X GET \
    -H "X-API-KEY: $API_KEY" \
    "https://www.pgdp.org/~cpeel/c.branch/api-pagedetail-single-round/api/index.php?url=v1/projects/$PROJECTID/pagedetails&pageroundid[]=OCR&pageroundid[]=F2"

```

### DIFF
--- a/api/dp-openapi.yaml
+++ b/api/dp-openapi.yaml
@@ -620,6 +620,13 @@ paths:
         required: true
         schema:
           type: string
+      - name: pageroundid
+        in: query
+        description: Page round IDs to return, if not set all page rounds are
+          returned. To request multiple page rounds, append with [] and pass
+          one per desired page round (e.g. pageroundid[]=OCR&pageroundid[]=P3).
+        schema:
+          type: string
       responses:
         200:
           description: Page details
--- a/api/v1_projects.inc
+++ b/api/v1_projects.inc
@@ -403,8 +403,26 @@ function api_v1_project_pages($method, $data, $query_params)
 
 function api_v1_project_pagedetails($method, $data, $query_params)
 {
+    // optional page round IDs (one or more) to filter down to
+    $only_rounds = null;
+    $pageroundids = array_get($query_params, "pageroundid", null);
+    if ($pageroundids) {
+        $only_rounds = [];
+        if (!is_array($pageroundids)) {
+            $pageroundids = [$pageroundids];
+        }
+        foreach ($pageroundids as $pageroundid) {
+            validate_page_round($pageroundid, null);
+            if ($pageroundid === "OCR") {
+                $only_rounds[] = "OCR";
+            } else {
+                $only_rounds[] = get_Round_for_round_id($pageroundid);
+            }
+        }
+    }
+
     $return_data = [];
-    foreach (fetch_page_table_data($data[":projectid"]) as $image) {
+    foreach (fetch_page_table_data($data[":projectid"], null, null, $only_rounds) as $image) {
         $page_rounds_data = [];
         // Remove proofer names and adjust timestamp format
         foreach ($image["pagerounds"] as $round_id => $round_data) {
--- a/pinc/Activity.inc
+++ b/pinc/Activity.inc
@@ -99,6 +99,11 @@ class Activity
         $Activity_for_id_[$id] = & $this;
     }
 
+    public function __toString()
+    {
+        return $this->id;
+    }
+
     /**
      * Return a User Access object
      *
--- a/pinc/page_table.inc
+++ b/pinc/page_table.inc
@@ -102,7 +102,7 @@ function fetch_page_table_data($project, $page_selector = null, $work_round = nu
         $rounds_to_display = get_rounds_to_display($project);
         $include_ocr = true;
     } else {
-        $rounds_to_display = array_diff($only_rounds, ["OCR"]);
+        $rounds_to_display = array_unique(array_diff($only_rounds, ["OCR"]));
         $include_ocr = in_array("OCR", $only_rounds);
     }
 
--- a/pinc/page_table.inc
+++ b/pinc/page_table.inc
@@ -106,14 +106,21 @@ function fetch_page_table_data($project, $page_selector = null, $work_round = nu
         $include_ocr = in_array("OCR", $only_rounds);
     }
 
+    $rounds_with_data = get_rounds_with_data($project->projectid);
+
     $fields_to_get = "{$project->projectid}.image AS image, length(master_text), state";
     $tables = $project->projectid;
     foreach ($rounds_to_display as $round) {
         $rn = $round->round_number;
-        if ($rn == 1) {
+
+        // find the previous round with data to use as the diff column
+        $rounds_with_data_index = array_search($round, $rounds_with_data);
+        if ($rounds_with_data_index === 0 || $rn == 1) {
             $prev_text_column_name = 'master_text';
-        } else {
+        } elseif ($rounds_with_data_index === false) {
             $prev_text_column_name = get_Round_for_round_number($rn - 1)->text_column_name;
+        } else {
+            $prev_text_column_name = $rounds_with_data[$rounds_with_data_index - 1]->text_column_name;
         }
 
         $tallyboard = new TallyBoard($round->id, 'U');
--- a/pinc/page_table.inc
+++ b/pinc/page_table.inc
@@ -72,20 +72,49 @@ function get_rounds_to_display($project)
  *     }
  * }
  * ```
+ *
+ * @param string $projectid
+ *   ID of project
+ * @param string|array $page_selector
+ *   Select which pages to return in the query, possible values are:
+ *   * null - all pages are returned
+ *   * string - this user's pages from all rounds are returned
+ *   * array - array of specific image names
+ * @param object $work_round
+ *   Round to limit a user's pages -- must be used with $page_selector set
+ *   to a username.
+ * @param array $only_rounds
+ *   Only return page details for a specific round, possible values are:
+ *   * null - all rounds with data are returned
+ *   * array of round objects and/or "OCR" - only these rounds are returned
+ *
+ * @return array
  */
-function fetch_page_table_data($project, $page_selector = null, $work_round = null)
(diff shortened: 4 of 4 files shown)


windymilla (approved):
LGTM

Testing notes (to check this is what you would expect - none of it was surprising to me):
1. Requesting an empty round name `&pageroundid[]=` gives "Invalid page round", code 105.
2. Requesting the same round twice `&pageroundid[]=F1&pageroundid[]=F1` gives a generic "An error occurred during a database query and has been logged.", code 0. Not sure if this should have been (or needs to be) trapped as a specific error?
3. Not requesting a round gives the same as previously, i.e. all rounds to date.
4. Explicitly requesting the same rounds as were given by 3. above,  gives identical results bar a few timestamps, as expected.
5. Adding a request for rounds that haven't been proofed yet works as expected, i.e. page details are returned but with default/empty/zero data in the fields.

cpeel:
> Requesting the same round twice &pageroundid[]=F1&pageroundid[]=F1 gives a generic "An error occurred during a database query and has been logged.", code 0. Not sure if this should have been (or needs to be) trapped as a specific error?

Good edgecase! I've added something to the function to unique the inputs so there are no duplicates. The resulting SQL does not handle duplicate rounds.

cpeel:
I just pushed up a commit to fix an edgecase where a prior round didn't have data. I found this edgecase testing something in `noncvs` on PROD where a project -- for whatever reason -- didn't have any P1 data which caused incorrect diff detection.

srjfoo:
> I just pushed up a commit to fix an edgecase where a prior round didn't have data. I found this edgecase testing something in `noncvs` on PROD where a project -- for whatever reason -- didn't have any P1 data which caused incorrect diff detection.

It's a fairly common edgecase, Round repeats where the repeat doesn't go all the way back to P1 will always start with the round that the project is sent back to (for example, F2->P2 will put the F2 text in the OCR slot and the first round that will be proofread in will be P2). Skips will cause empty rounds, also (Italian projects commonly skip P3 and F2).

srjfoo (commented):
From `page_table.inc` (lines 51-73, in the comment section), though not part of this PR,
```php
{
    "image": "018.png",
    "image_url": "https://www.pgdp.net/projects/projectID460978cb8e116/018.png",
    "state": "P3.page_saved",
    "image_size": 45541,
    "pagerounds": {
        "OCR": {
            "page_size": "1445"
        },
        "P1": {
            "page_size": "1446",
            "is_diff": "0",
            "username": "username",
            "user_page_tally": "117"
            "wordcheck_ran": "0",
            "modified_timestamp": "1175026200",
        },
        "P2": {
            ...
        },
        ...
    }
}
```

Line 63 implies that the username is supposed to be included in the results. Was this changed later in the code and just not removed from the comment? If this is to provide enough information to be used for generating a page details table, shouldn't it include the username? (Or is it omitted because then there would need to be logic to restrict who can see the username and when?)

I'm guessing that some things changed without the comment being updated, since the above shows `modified_timestamp` with the numeric timestamp, and the curl results show `last_modified` with a human readable timestamp.

cpeel:
> Line 63 implies that the username is supposed to be included in the results. Was this changed later in the code and just not removed from the comment? If this is to provide enough information to be used for generating a page details table, shouldn't it include the username? (Or is it omitted because then there would need to be logic to restrict who can see the username and when?)
> 
> I'm guessing that some things changed without the comment being updated, since the above shows `modified_timestamp` with the numeric timestamp, and the curl results show `last_modified` with a human readable timestamp.

`fetch_page_table_data()` is used both to populate the page table for the UI and for the API, but its job is just to return all of the data. The presentation layers are responsible for how it is presented and what parts of it are allowed to be shown to the user. In the API that's done in [api_v1_project_pagedetails(()](https://github.com/DistributedProofreaders/dproofreaders/blob/master/api/v1_projects.inc#L404) which removes the username if the proofreader is not allowed to see it and changes the timestamp to match our API-standard `DATE_ATOM` format. (The UI uses this function from [echo_page_table()](https://github.com/DistributedProofreaders/dproofreaders/blob/master/pinc/page_table.inc#L202) to show the other half of that.)

I need to add a comment to `fetch_page_table_data()` letting devs know that the format is used largely as-is by the API and changes there could impact the API.

srjfoo:
> > Line 63 implies that the username is supposed to be included in the results. Was this changed later in the code and just not removed from the comment? If this is to provide enough information to be used for generating a page details table, shouldn't it include the username? (Or is it omitted because then there would need to be logic to restrict who can see the username and when?)
> > I'm guessing that some things changed without the comment being updated, since the above shows `modified_timestamp` with the numeric timestamp, and the curl results show `last_modified` with a human readable timestamp.
> 
> `fetch_page_table_data()` is used both to populate the page table for the UI and for the API, but its job is just to return all of the data. The presentation layers are responsible for how it is presented and what parts of it are allowed to be shown to the user. In the API that's done in [api_v1_project_pagedetails(()](https://github.com/DistributedProofreaders/dproofreaders/blob/master/api/v1_projects.inc#L404) which removes the username if the proofreader is not allowed to see it and changes the timestamp to match our API-standard `DATE_ATOM` format. (The UI uses this function from [echo_page_table()](https://github.com/DistributedProofreaders/dproofreaders/blob/master/pinc/page_table.inc#L202) to show the other half of that.)
> 
> I need to add a comment to `fetch_page_table_data()` letting devs know that the format is used largely as-is by the API and changes there could impact the API.

Thanks -- it sounds like it should be working as I expected, then. The thing is, the API key I was using should have allowed me to see usernames. I'll do some more poking; it's probably user error.

srjfoo:
With another round of testing, I got the same results.

Using the API key for my normal squirrel account, I saw no usernames in the results. So, I guess what I'm asking is, is it even retrieved? (For page details to be complete, it should be available to those with the permissions to access the information, shouldn't it?)

Just to make sure it hadn't gotten somehow backwards, I confirmed that I wasn't getting the username information by double-checking with a totally unprivileged account. I compared the API query against the page details display while logged in as that unprivileged user, and noticed something else: the user column isn't even displayed for users who don't have access to username information via page details, so the `user_page_tally` information isn't displayed there, either. However, it was retrieved via API. If it's not displayed in the page details for a project, should it be being retrieved via API for an unprivileged user? (Except for those pages they proof, that is.)

cpeel:
> Using the API key for my normal squirrel account, I saw no usernames in the results.

It helps if I actually _look_ at the code. The API [removes all usernames](https://github.com/DistributedProofreaders/dproofreaders/blob/master/api/v1_projects.inc#L416) before returning results to the user and that matches with the [current API schema](https://github.com/DistributedProofreaders/dproofreaders/blob/master/api/dp-openapi.yaml#L865) we publish. I presume I decided it was easier to just not include it at all in the API vs conditionally including based on the users permissions.

It's included in what `fetch_page_table_data()` returns which allows `echo_page_table()` to [conditionally include](https://github.com/DistributedProofreaders/dproofreaders/blob/master/pinc/page_table.inc#L372) it in the Page Details view.

Sorry for the confusion, I'm doing too much at one time over here.

srjfoo:
> > Using the API key for my normal squirrel account, I saw no usernames in the results.
> 
> It helps if I actually _look_ at the code. The API [removes all usernames](https://github.com/DistributedProofreaders/dproofreaders/blob/master/api/v1_projects.inc#L416) before returning results to the user and that matches with the [current API schema](https://github.com/DistributedProofreaders/dproofreaders/blob/master/api/dp-openapi.yaml#L865) we publish. I presume I decided it was easier to just not include it at all in the API vs conditionally including based on the users permissions.
> 
> It's included in what `fetch_page_table_data()` returns which allows `echo_page_table()` to [conditionally include](https://github.com/DistributedProofreaders/dproofreaders/blob/master/pinc/page_table.inc#L372) it in the Page Details view.
> 
> Sorry for the confusion, I'm doing too much at one time over here.

Hang in there -- I think this takes back seat!

Yes, that helps immensely, thanks! And a totally legitimate reason -- I just wanted to make sure that I wasn't missing something. Is there a reason for including the `user_page_tally` if `username` isn't there, or should that be left out, also?

cpeel:
> Is there a reason for including the `user_page_tally` if `username` isn't there, or should that be left out, also?

I am guessing the reason I included it is that we could -- it doesn't violate any of our privacy policies to include it. Since it's already a part of our established API I'm loath to remove it unless we think it's necessary.

srjfoo:
> > Is there a reason for including the `user_page_tally` if `username` isn't there, or should that be left out, also?
> 
> I am guessing the reason I included it is that we could -- it doesn't violate any of our privacy policies to include it. Since it's already a part of our established API I'm loath to remove it unless we think it's necessary.

🤷‍♀️ -- good enough. It just didn't seem useful without the username.

srjfoo (approved):

PR #1021, opened 2024-01-19
Preload wordcheck_events for project details

cpeel:
To load the "wordcheck ran" boolean for a project's page detail we currently do a subquery against `wordcheck_events` for each round. As the table has grown the query has become less and less performant despite there being an index for `projectid` already. The more pages a project has and the more rounds it has gone through increases the work required by the SQL. For some projects on PROD page details can take over 30 seconds to load (all of which is spent in the mysql server). 

I evaluated a few different ways to solve this, including changing to do a full table join (much worse performance) and adding a compound `(projectid, image)` or `(projectid, username)` index (ruled out as it makes the tablespace even bigger, and it's already at 1GB -- one of our biggest).

I settled on preloading the project's wordcheck events into an associative array and determining the "wordcheck ran" boolean in the code instead of in the SQL. This improves performance on some pages on PROD by 5x. The only downside to this method is that it uses memory for the hash. Despite the worst-case cardinality of the hash being `(min(num_rounds, 3) * num_unique_proofreaders * num_project_pages)` we're only storing an integer and PHP uses a [hashmap under the covers](https://stackoverflow.com/a/25114841) so it's not that bad.

The code can be tested on TEST at https://www.pgdp.org/~cpeel/c.branch/better-pagedetail-perf/ . It is very unlikely to see performance differences on TEST, even with projects in many rounds with many pages, but it should render the same HTML as what's in the main TEST sandbox.

I have a copy of `page_detail.php` on PROD in noncvs that uses the code in this PR and can be used to compare performance against the main PROD site. For instance:
* [noncvs version against projectID5f70409ee5bbb](https://www.pgdp.net/noncvs/cpeel/page_detail.php?project=projectID5f70409ee5bbb) -- <7 seconds
* [PROD version against projectID5f70409ee5bbb](https://www.pgdp.net/c/tools/project_manager/page_detail.php?project=projectID5f70409ee5bbb) -- ~ 37 seconds

### DIFF
--- a/pinc/page_table.inc
+++ b/pinc/page_table.inc
@@ -106,6 +106,35 @@ function fetch_page_table_data($project, $page_selector = null, $work_round = nu
         $include_ocr = in_array("OCR", $only_rounds);
     }
 
+    // load wordcheck_event counts for this table for lookup in building the
+    // page table later, this is faster than doing subqueries in single SQL
+    $round_ids_to_load_wordcheck_events = [];
+    foreach ($rounds_to_display as $round) {
+        if (!is_formatting_round($round)) {
+            $round_ids_to_load_wordcheck_events[] = $round->id;
+        }
+    }
+    if ($round_ids_to_load_wordcheck_events) {
+        $sql = sprintf(
+            "
+            SELECT image, username, round_id, count(*)
+            FROM wordcheck_events
+            WHERE projectid = '%s' and round_id in (%s)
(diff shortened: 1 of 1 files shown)


chrismiceli (approved):

70ray (approved):

srjfoo (approved):

windymilla (approved):
*Much* faster.

PR #1022, opened 2024-01-19
Avoid repeating round definitions when defining project states

70ray:
Define rounds, activities and project states inside functions.  Use the defined rounds when defining project states. Postpone defining 'round project states' until defining project states to avoid duplication. Then define other activities since they use constants defined when defining project states.
This does not make any attempt to reduce the use of global variables but it could be a step on the way to doing so.

### DIFF
--- a/pinc/project_states.inc
+++ b/pinc/project_states.inc
@@ -22,6 +22,8 @@ include_once($relPath.'site_vars.php');
 
   (Actually, the naming convention is in transition.)
 */
+global $PROJECT_STATES_IN_ORDER, $project_state_medium_label_, $project_state_long_label_;
+global $project_state_phase_, $project_states_for_star_metal_;
 
 $PROJECT_STATES_IN_ORDER = [];
 $project_state_medium_label_ = [];
--- a/pinc/project_states.inc
+++ b/pinc/project_states.inc
@@ -58,7 +58,52 @@ function define_project_states($rounds, $pools)
     );
 
     foreach($rounds as $round) {
-        declare_project_states_for_round($round);
+        declare_project_state(
+            "PROJ_{$round->id}_BAD_PROJECT",
+            $round->project_bad_state,
+            "$round->id: " .  _("Bad Project"),
+            "$round->name: " . _("Bad Project"),
+            $projects_forum_idx,
+            'PAGE_EDITING',
+            ''
+        );
+
+        declare_project_state(
+            "PROJ_{$round->id}_UNAVAILABLE",
+            $round->project_unavailable_state,
+            "$round->id: " . _("Unavailable"),
+            "$round->name: " . _("Unavailable"),
+            ($round->id == 'P1' ? $waiting_projects_forum_idx : $projects_forum_idx),
+            'PAGE_EDITING',
+            ''
+        );
+        declare_project_state(
+            "PROJ_{$round->id}_WAITING_FOR_RELEASE",
+            $round->project_waiting_state,
+            "$round->id: " . _("Waiting"),
+            "$round->name: " . _("Waiting for Release"),
+            ($round->id == 'P1' ? $waiting_projects_forum_idx : $projects_forum_idx),
+            'PAGE_EDITING',
+            ''
+        );
+        declare_project_state(
+            "PROJ_{$round->id}_AVAILABLE",
+            $round->project_available_state,
+            "$round->id: " . _("Available"),
+            "$round->name: " . _("Available"),
+            $projects_forum_idx,
+            'PAGE_EDITING',
+            'BRONZE'
+        );
+        declare_project_state(
+            "PROJ_{$round->id}_COMPLETE",
+            $round->project_complete_state,
+            "$round->id: " . _("Completed"),
+            "$round->name: " . _("Completed"),
+            $projects_forum_idx,
+            'PAGE_EDITING',
+            ''
+        );
     }
 
     // POST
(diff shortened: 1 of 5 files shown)


cpeel:
I like moving the object construction inside functions but it feels like we're just moving the circular dependency around rather than solving it.

I'm trying to wrap my head around if using constants is the right approach in the code at all -- it isn't as though using `PROJ_POST_FIRST_CHECKED_OUT` vs `"proj_post_first_checked_out"` is all that useful. Unfortunately the constants aren't consistent (heh) as `PROJ_NEW` is actually `"project_new"`.

pinc/project_states.inc
```diff
@@ -22,17 +22,180 @@ include_once($relPath.'site_vars.php');
...
+    $project_state_forum_ = [];
+    $project_state_phase_ = [];
+    $project_states_for_star_metal_ = [
+        'BRONZE' => [],
+        'SILVER' => [],
+        'GOLD' => [],
+    ];
+
+    $project_round_states = [
+        ["BAD_PROJECT", "bad", "bad", _("Bad Project"), _("Bad Project")],
```
windymilla, line 46: This line made me lol. I imagined our dog was called "Project" and had just eaten my slippers: "Bad Project! Bad, bad, Bad Project!"

pinc/stages.inc
```diff
@@ -261,110 +45,337 @@ $ELR_round = get_Round_for_round_number(1);
...
+        "tools/post_proofers/smooth_reading.php"
+    );
+
+    // -----------------------------------------------------------------------------
+
+    new Pool(
+        'PPV',
+        _('Post-Processing Verification'),
+        [],
+        'NOREQ', // "Peer approval. Also gives F2 access."
```
windymilla, line 331: Is the comment true? PPV gives F2 access? I didn't know that. Or maybe I've misread it.

windymilla (approved):

70ray:
> I like moving the object construction inside functions but it feels like we're just moving the circular dependency around rather than solving it.
> 
> I'm trying to wrap my head around if using constants is the right approach in the code at all -- it isn't as though using `PROJ_POST_FIRST_CHECKED_OUT` vs `"proj_post_first_checked_out"` is all that useful. Unfortunately the constants aren't consistent (heh) as `PROJ_NEW` is actually `"project_new"`.

I think we need to postpone some of the POOL definitions to project_states (as this PR does for rounds). Then it would not seem so convoluted. I'll work on this.

cpeel:
> I think we need to postpone some of the POOL definitions to project_states (as this PR does for rounds). Then it would not seem so convoluted. I'll work on this.

That doesn't help the fact that we're still defining global variables but now inside functions which seems actually weirder and less intuitive.

I don't have the solution but this approach doesn't seem like the right one.

70ray:
> > I think we need to postpone some of the POOL definitions to project_states (as this PR does for rounds). Then it would not seem so convoluted. I'll work on this.
> 
> That doesn't help the fact that we're still defining global variables but now inside functions which seems actually weirder and less intuitive.
> 
> I don't have the solution but this approach doesn't seem like the right one.

The stages would then all be created before the project states.
Wait till you've seen how wonderful it is. It would make it adding new pools just as easy as adding new rounds.

pinc/stages.inc
srjfoo: No, PPV no longer gives F2 access. It used to until somewhere around 2007-2009, but apparently the comment was never revised. By the time I came along in 2009, it was no longer true.

70ray:
I don't think the constants defined in `declare_project_state` add much value but they are used quite extensively. I think the only value in constants rather than strings in general is that they enable mis-spelling mistakes to be caught at "compile time" and in the old days constants could be defined as something small which would save memory and speed things up while still having a descriptive name in the code but none of these seem to apply here.

pinc/project_states.inc
```diff
@@ -22,17 +22,171 @@ include_once($relPath.'site_vars.php');
...
```
70ray, line 46: These lines also make one wonder whether we really need so many different strings. Leaving aside the question of whether we really need the constants at all, do we really need both short and long states and labels? since in most cases they are the same.

70ray, line 46: It would be possible to write the translated strings like `_("%s: Bad Project")` where %s will be replaced by something like "P1" in the short case or _('Proofreading Round 3').

70ray:
> > I think we need to postpone some of the POOL definitions to project_states (as this PR does for rounds). Then it would not seem so convoluted. I'll work on this.
> 
> That doesn't help the fact that we're still defining global variables but now inside functions which seems actually weirder and less intuitive.
> 
> I don't have the solution but this approach doesn't seem like the right one.

You are right. I ended up doing it quite differently.

70ray:
If we were to consider merging this it would be best to write some more comprehensive tests in a new PR and merge it first so that we could be more certain that this PR doesn't break anything.

70ray (commented):
Although this removes the duplicated round definitions it still seems somewhat circular. Is it possible to more clearly separate the definitions of the stages the project moves through and defining the qualifications required to work in each stage? Or if this is not possible then make the definitions more unified rather than having separate definition files (stages.inc and project_states.inc).

70ray:
All references to project states have now been removed from stage files and put in project_states.inc

cpeel:
Love this! Having the project states be derived from the activities makes a lot of sense. I think this combined with https://github.com/DistributedProofreaders/dproofreaders/pull/1009 gets us in a much better & clearer space on how the site structure is defined and accessed.

For looking up rounds for a project state I'm on the fence if this should be "owned" by the round or by project state -- I'll have to noodle that one part over.

I think I'd like to get 1009 in to get rid of the globals and then this one.

70ray:
> I think I'd like to get 1009 in to get rid of the globals and then this one.

I've extended this to get rid of some globals in a different way. I can make a PR for it so anyone can look at it but I should really re-make it because I could have accidentally reverted some things by copying from an out-of-date master file.

70ray:
> I think I'd like to get 1009 in to get rid of the globals and then this one.

It will be easier to re-make this after removing the globals so I'll close this.

PR #1023, opened 2024-01-21
New charsuite cyrillic

srjfoo:
The Basic Cyrillic character suite covers Russian. Discussion in the forums concluded with recommending that PMs who run projects for other Cyrillic-based languages add the few additional letters as project custom characters, and wait to see if a need develops later for another Cyrillic-based character suite.

[This project](https://www.pgdp.org/~srjfoo/c.branch/new-charsuite-cyrillic/project.php?id=projectID65ac79d59af7b) has been set up in the sandbox with pages that contain some Cyrillic upon which to practice (note that no Cyrillic-based dictionaries have yet been installed).

### DIFF
--- a/SETUP/UNICODE.md
+++ b/SETUP/UNICODE.md
@@ -54,6 +54,7 @@ The code currently ships with the following character suites:
 * Semitic and Indic transcriptions - Romanized forms of Arabic, Hebrew, Sanskrit
 * Symbols collection - Astronomical, Music, Zodiac and Apothecary symbols
 * Math symbols - Commonly used mathematical symbols not in Basic Latin
+* Basic Cyrillic
 
 All character suites can be viewed using the [All Character Suites](../tools/charsuites.php)
 page.
--- a/pinc/charsuite-basic-cyrillic.inc
+++ b/pinc/charsuite-basic-cyrillic.inc
@@ -1,0 +1,49 @@
+<?php
+include_once($relPath."CharSuites.inc");
+
+use voku\helper\UTF8;
+
+$charsuite = new CharSuite("basic-cyrillic", _("Basic Cyrillic"));
+
+$charsuite->codepoints = [
+    // https://www.pgdp.net/phpBB3/viewtopic.php?p=1321508#p1321508
+    'U+0401',
+    'U+0406',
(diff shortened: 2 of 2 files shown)


SETUP/UNICODE.md
```diff
@@ -54,6 +54,7 @@ The code currently ships with the following character suites:
 * Semitic and Indic transcriptions - Romanized forms of Arabic, Hebrew, Sanskrit
 * Symbols collection - Astronomical, Music, Zodiac and Apothecary symbols
 * Math symbols - Commonly used mathematical symbols not in Basic Latin
+* Basic Cyrillic
```
windymilla, line 57: Is it worth saying
```
 * Basic Cyrillic - Russian including pre-revolutionary characters
```
to explain which Cyrillic characters are included

70ray (commented):
Would it be worth considering putting all 33 basic letters in one pickerset? Are there many people using narrower screens where this would be a problem?
Or alternatively putting the most common characters in the primary set.

windymilla:
> Would it be worth considering putting all 33 basic letters in one pickerset? Are there many people using narrower screens where this would be a problem? Or alternatively putting the most common characters in the primary set.

I made a similar observation in the forum discussion thread

SETUP/UNICODE.md
srjfoo: > Is it worth saying
> 
> ```
>  * Basic Cyrillic - Russian including pre-revolutionary characters
> ```
> 
> to explain which Cyrillic characters are included

I thought about it, but in the end went with just the name of the character suite, matching the fact that Basic Greek and Polytonic Greek do not have anything there. I certainly have no objections to adding it.

srjfoo: I'll plan to do that, but I'm inclined to wait until the discussion in the forums comes to some conclusion regarding the character suite itself.

cpeel (approved):

70ray (approved):

srjfoo:
After very little discussion following [this post](https://www.pgdp.net/phpBB3/viewtopic.php?p=1331908#p1331908) in the forums, the decision was made to go ahead with what we have.

PR #1026, opened 2024-01-25
Standardize some close-but-not-quite strings

cpeel:
To help reduce the translation burden, standardize a few strings to match others.

Testable in https://www.pgdp.org/~cpeel/c.branch/massage-strings/


srjfoo (changes requested):
Just found the one (`qd_p_basicx_1.inc`) that apparently got overlooked.

chrismiceli (approved):

srjfoo (approved):

PR #1027, opened 2024-01-25
Make all activities then make access tables

70ray:
Add activities to an accessor object then make some more accessors for extended activities. Use rounds and pools to make project states. Add some tests.
This has only changed the core part and not used the new objects in the rest of the code but made aliases for compatibilty which all need to be changed to do it properly.

### DIFF
--- a/SETUP/tests/RoundTest.php
+++ b/SETUP/tests/RoundTest.php
@@ -1,0 +1,52 @@
+<?php
+
+class RoundTest extends ProjectUtils
+{
+    //------------------------------------------------------------------------
+
+    public function test_get_round()
+    {
+        $this->assertEquals("P1", get_Round_for_round_number(1)->id);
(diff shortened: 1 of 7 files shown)


cpeel (changes requested):
I'm not a fan of the accessor implementation because it still requires globals, we've just moved them around and abstract them (or try to) behind functions.

I can see where you're going with an explicit assignment (create the object, pass it into a container) rather than an implicit one (having the object add itself to a container upon creation) but we aren't getting the goal of removing globals.

70ray:
> 
> I can see where you're going with an explicit assignment (create the object, pass it into a container) rather than an implicit one (having the object add itself to a container upon creation) but we aren't getting the goal of removing globals.

Oops, I must have completely missed the point. I'll look at your implementation again.

70ray:
I've made a new branch with static classes.

PR #1029, opened 2024-01-25
api/v1/projects: Add in_smoothreading flag to query

bpfoley:
This requires a new boolean flag type which can be one of:
* absent (implicitly false)
* present but with no value, ie "&flag" (implicitly true)
* explicitly true or false, ie "&flag=false" "&flag=true"

Also, add smoothread_deadline as a valid datetime field in the response.

```bash
API_KEY=xxxx # update
API_ROOT=https://www.pgdp.org/api
FILTER="&in_smoothreading"
FIELDS="&field[]=title&field[]=smoothread_deadline"
curl -i -X GET "$API_ROOT/v1/projects?state=proj_post_first_checked_out${FILTER}${FIELDS}" \
    -H "X-API-KEY: $API_KEY"
```

### DIFF
--- a/api/v1_projects.inc
+++ b/api/v1_projects.inc
@@ -29,6 +29,12 @@ function api_v1_projects($method, $data, $query_params)
         "pages_total" => "n_pages",
     ];
 
+    // parameters in the query where the value can be "false", "true" or absent (==true)
+    // When true, a WHERE clause is added to the SQL query.
+    $valid_flags = [
+        "in_smoothreading" => "smoothread_deadline > UNIX_TIMESTAMP()",
+    ];
+
     // Allow SAs and PFs to search on clearance. We can't allow PMs to do so
     // without opening up the ability for value-fishing as PMs can only see
     // clearances for their own projects.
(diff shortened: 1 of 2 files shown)


api/v1_projects.inc
```diff
@@ -294,6 +306,7 @@ function render_project_json($project, $return_fields = null)
             "last_state_change_time" => date(DATE_ATOM, $project->modifieddate),
             "last_page_done_time" => date(DATE_ATOM, $project->t_last_page_done),
             "last_edit_time" => date(DATE_ATOM, $project->t_last_edit),
+            "smoothread_deadline" => date(DATE_ATOM, $project->smoothread_deadline),
```
cpeel, line 309: We need to add this to the `#/components/schemas/project` schema as well.

cpeel (commented):
Thanks Brian - nice abstraction with the flags.

Can you spin up a sandbox for folks to test this against as well?

bpfoley:
> Thanks Brian - nice abstraction with the flags.
> 
> Can you spin up a sandbox for folks to test this against as well?

ACK. There's a sandbox at https://www.pgdp.org/~bfoley/c.branch/api-projects-sr/

api/v1_projects.inc
bpfoley: TIL. Will do!

70ray (approved):

cpeel (approved):

srjfoo (approved):

PR #1031, opened 2024-01-26
Add activities to a static class.

70ray:
Then make static classes for extended activities. Use rounds and pools to make a static class for project states. Add some tests.

I didn't realise what the significance of a static class was before. This uses  static classes but also has some temporary functions and global variables to enable it to work without actually using them in other files (except one instance (my_projects.php)where a different function could be substituted).

### DIFF
--- a/pinc/Round.inc
+++ b/pinc/Round.inc
@@ -1,20 +1,6 @@
 <?php
 
-// $Id$
-
 include_once($relPath.'Stage.inc');
-
-// -----------------------------------------------------------------------------
-
-global $n_rounds, $Round_for_round_id_, $Round_for_round_number_;
-global $Round_for_project_state_, $Round_for_page_state_, $PAGE_STATES_IN_ORDER;
-
-$n_rounds = 0;
-$Round_for_round_id_ = [];
-$Round_for_round_number_ = [];
-$Round_for_project_state_ = [];
-$Round_for_page_state_ = [];
-$PAGE_STATES_IN_ORDER = [];
 
 // A container for various constants relating to a particular round of proofreading.
 class Round extends Stage
@@ -62,6 +48,7 @@ class Round extends Stage
      *   (Needn't be in a particular order.)
      */
     public function __construct(
+        $round_number,
         $id,
         $name,
         $access_minima,
@@ -89,21 +76,12 @@ class Round extends Stage
             "tools/proofers/round.php?round_id=$round_id"
         );
 
-        global $n_rounds;
-        $n_rounds++;
-        $this->round_number = $n_rounds;
-
+        $this->round_number = $round_number;
         $this->pi_tools = $pi_tools;
         $this->daily_page_limit = $daily_page_limit;
         $this->other_rounds_with_visible_usernames = $other_rounds_with_visible_usernames;
         $this->honorifics = $honorifics;
         krsort($this->honorifics);
-
-        $this->project_unavailable_state = constant("PROJ_{$round_id}_UNAVAILABLE");
-        $this->project_waiting_state = constant("PROJ_{$round_id}_WAITING_FOR_RELEASE");
-        $this->project_bad_state = constant("PROJ_{$round_id}_BAD_PROJECT");
-        $this->project_available_state = constant("PROJ_{$round_id}_AVAILABLE");
-        $this->project_complete_state = constant("PROJ_{$round_id}_COMPLETE");
 
         $this->page_avail_state = "{$round_id}.page_avail";
         $this->page_out_state = "{$round_id}.page_out";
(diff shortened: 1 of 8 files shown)


pinc/stages.inc
```diff
@@ -11,360 +11,537 @@ include_once($relPath.'Pool.inc');
...
+    Activities::set(new Activity(
+        "P2_mentor",
+        sprintf(_("Mentoring in round %s"), 'P2'),
+        [],
+        'NOREQ',
+        '',
+        null // access_change_callback
+    ));
+
+    // now construct various access tables
```
cpeel, line 435: "tables" isn't quite the right word. I believe you're thinking of "lookup tables" but in PHP those are just associative arrays (in python they're called dictionaries). "tables" in our code with the DB backend really implies "database table" and I'd encourage us to avoid using "table" for the array.

pinc/stages.inc
```diff
@@ -11,360 +11,537 @@ include_once($relPath.'Pool.inc');
...
-// This particular assignment assumes that the ELR is the first round
-// defined in this file. This is probably correct, but your site is
-// welcome to use a different round as your ELR.
-$ELR_round = get_Round_for_round_number(1);
-
-// On the other hand, if your site doesn't have a distinguished entry-level
-// round (e.g., everyone can work in multiple rounds from their first day),
-// then you may have to do some hacking (sorry).
-
+trait StatObj
```
cpeel, line 14: These definition and the access functions don't belong in `stages.inc` which should just be the creation of objects. They probably belong in their respective files with the traits in `Activity.inc`.

cpeel (commented):
Yes, this is what I was thinking of on removing the global variables.

But I have to ask: why didn't you just comment on what you didn't like about my PR #1009 which pretty much did a very similar thing on the static classes? I like that we're converging but I have to ask: why?

70ray:
> Yes, this is what I was thinking of on removing the global variables.
> 
> But I have to ask: why didn't you just comment on what you didn't like about my PR #1009 which pretty much did a very similar thing on the static classes? I like that we're converging but I have to ask: why?

Thank Casey. I think there are two reasons why I went ahead with this. I must admit I found your PR difficult to understand and at the time I obviously didn't understand what static classes were for so I didn't make much effort. The other reason is that this evolved in a very convoluted way from removing the duplication of round definitions in project states definition.  A third reason is that I just enjoyed re-inventing the wheel. It may be useful to explore different ways of dong the same thing from different directions.

pinc/Round.inc
```diff
@@ -89,22 +76,13 @@ class Round extends Stage
             "tools/proofers/round.php?round_id=$round_id"
         );
 
-        global $n_rounds;
-        $n_rounds++;
-        $this->round_number = $n_rounds;
-
+        $this->round_number = $round_number;
```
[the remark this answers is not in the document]
70ray: I think passing it in from the function creating the rounds achieves this just as well.

pinc/stages.inc
```diff
@@ -11,360 +11,537 @@ include_once($relPath.'Pool.inc');
...
+        if($activity instanceof Pool) {
+            Pools::set($activity);
+        }
+        if($activity instanceof Stage) {
+            Stages::set($activity);
+        }
+        if($activity instanceof Round) {
+            $round = $activity;
+            Rounds::set($round);
+            Rounds::set_for_number($round);
```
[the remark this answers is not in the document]
70ray: Yes, This came about probably because Rounds::set($round); is in the trait which all the activities use and the others are added on for the special case of the Round.
I have to admit that the way this works duplicates the hierarchical structure of the activtity-stage-round definition to some extent and it would probably be easier to create these 'tables' inside the acivity objects like the present code does. I was reluctant to do this because it somehow didn't sem like the right thing to do. Perhaps I should get over it and try doing like that.

pinc/stages.inc
70ray: Yes, I did occur to me earlier that what this stuff does is like creating a database in memory of all the activity information with Key lookup tables to find the information in various ways. I'll change that comment.

pinc/stages.inc
70ray: I'll try going back to how the old code does it from inside the activity classes to see how it works out.

cpeel:
Makes sense! Let me reduce my proposal to a smaller set (remove the codebase-wide updates to remove the globals) so we can review these two proposals more closely since we'll need to decide on one of them 😁

70ray:
> Makes sense! Let me reduce my proposal to a smaller set (remove the codebase-wide updates to remove the globals) so we can review these two proposals more closely since we'll need to decide on one of them 😁
I started working on adding activities to look-up tables from inside the activity classes. I think this may work out to be much simpler and will probably end up quite close to yours so I'll probably delete all my PRs and comment on yours if my method seems to have any advantages over yours.

70ray:
Adding activities to look-up tables from inside the activity classes is much simpler so I'll close this.

Issue #1025, opened 2024-01-25
The pages proofed graphs take a long time if data is not cached


srjfoo:
In recent experimentation, `stats/pages_proofed_graphs.php` for all rounds on the production site take over 70 seconds if there isn't a recent cache of the data. Is there a way to improve performance (or cause it to have less impact on users viewing the graphs)? When it takes that long for a page to display, it gets worrisome.

I assume that the culprit is the cumulative graph on each page, and this will likely only to continue to get worse.

cpeel:
We should start by caching the graphs until the next server-time midnight, that way we only incur one slow-load a day. I don't think any of those graphs are important enough to retain current-day data.

I'm dubious if we can do anything about the long generation time without changing what's in the charts -- it's just a lot of data.

srjfoo:
> We should start by caching the graphs until the next server-time midnight, that way we only incur one slow-load a day. I don't think any of those graphs are important enough to retain current-day data.

I can't say for sure, because I'm not adept at reading that part of the code, but I don't think current day data is used, anyway. Looking (for example) at P1's "Pages Done Per Day (Jan 2024)" and "Cumulative Pages Done (Jan 2024)", the last bar on each of those graphs is current-day-minus-one. 

> I'm dubious if we can do anything about the long generation time without changing what's in the charts -- it's just a lot of data.

Yeah. The only idea I've had for the last two graphs on the page is to store the data and make those static graphs that can be refreshed once a month, maybe. But I don't have any idea whether that would be workable or not.

cpeel:
#1033 should mitigate this. Let's see how the page perf feels after we get this deployed.

cpeel:
#1033 seems to make the page feel snappy for most use-cases so I'm going to close this issue.

PR #1033, opened 2024-01-27
Cache page stats graphs as long as possible

cpeel:
Generating the page stats is very resource intensive and can take a very long time. Cache the stats for as long as is reasonable which in some cases is up to a month. This mitigates https://github.com/DistributedProofreaders/dproofreaders/issues/1025

Clarify that memcached accepts both a timestamp *and* seconds from now: https://github.com/memcached/memcached/wiki/Programming#expiration

Testable in https://www.pgdp.org/~cpeel/c.branch/cache-graphs-longer/ -- note that the HTML for the page as a comment that shows exactly what the expiration times are evaluated to.

### DIFF
--- a/pinc/graph_data.inc
+++ b/pinc/graph_data.inc
@@ -876,7 +876,7 @@ function get_round_backlog_stats($interested_phases)
  * Note that the graph function might include _() so it's important that we
  * include the user's language in the key.
  */
-function query_graph_cache($function, $args = [], $expire_from_now = 3600)
+function query_graph_cache($function, $args = [], $expiration = 3600)
 {
-    return memoize_function($function, $args, $expire_from_now, get_desired_language());
+    return memoize_function($function, $args, $expiration, get_desired_language());
 }
--- a/pinc/misc.inc
+++ b/pinc/misc.inc
@@ -1290,7 +1290,7 @@ function factor_strings($strings)
  * If the requested function returns data that has localized strings, it's
  * important that $key_salt = get_desired_language();
  */
-function memoize_function($function, $args = [], $expire_from_now = 3600, $key_salt = "")
+function memoize_function($function, $args = [], $expiration = 3600, $key_salt = "")
 {
     // cache this connection for possible re-use during this page load
     static $memcache = null;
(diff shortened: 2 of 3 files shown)


srjfoo:
A couple of questions:

In the comment text, I'm seeing
```html
    <!--          Now: 2024-01-27T02:24:26+00:00
    Tomorrow midnight: 2024-01-28T00:00:00+00:00
     First next month: 2024-02-01T02:24:26+00:00
      First next week: 2024-01-28T00:00:00+00:00
```
"First next month" time of day is not based on midnight, but on "Now". Is that what you wanted?

Second question -- would it be useful to have the expiration time on the graphs themselves, or would that just be confusing to viewers?

chrismiceli (approved):

windymilla:
Just checking - I wasn't quite sure from reading the memcached docs and the code...
Is a number of seconds that is greater than 30 days interpreted as a unix timestamp instead? Where we do "first day of next month" are we using the "number of seconds"? 
If both of those are true, does the following cause an issue?
If it is Jan 1st, then the first day of next month would be Feb 1st, i.e. 31 days later, which is more than 30 days, so a number of seconds from "now" on Jan 1st would get interpreted as a timestamp.

cpeel:
> Where we do "first day of next month" are we using the "number of seconds"?

We are not. We're using some human-readable strings to create a DateTimeImmutable object and then from that outputting a timestamp to pass as the expiration.

Interestingly I was originally calculating the number of seconds before looking at the memcached docs to see how big a number it would accept and learning we could pass in timestamps instead.

windymilla (approved):
Thanks for confirmation

cpeel:
> "First next month" time of day is not based on midnight, but on "Now". Is that what you wanted?

It was not and I've fixed it - thanks for catching that.

> Second question -- would it be useful to have the expiration time on the graphs themselves, or would that just be confusing to viewers?

I think having the expiration time on the graphs is going to be confusing (and make the code more confusing because the graph could would need to know when they expire, which is handled by a layer higher up). We could update the graphs to contain when they were generated and that would probably make more sense to users but I'm not sure if the effort is worth the return?

70ray (approved):

srjfoo:
> > Second question -- would it be useful to have the expiration time on the graphs themselves, or would that just be confusing to viewers?
> 
> I think having the expiration time on the graphs is going to be confusing (and make the code more confusing because the graph could would need to know when they expire, which is handled by a layer higher up). We could update the graphs to contain when they were generated and that would probably make more sense to users but I'm not sure if the effort is worth the return?

No, I don't think it's worth going to a lot of effort. And even that would likely be confusing. Since different graphs have different refresh times, it would probably be much more likely to be confusing than helpful. Thanks!

srjfoo (approved):

PR #1035, opened 2024-01-28
Prevent iOS issues

chrismiceli:
https://www.pgdp.org/~chrismiceli/c.branch/bug_2071/

Note, I left spellcheck off because I think it probably catches a ton of scannos, but I'm not sure if that means iOS will still use smart quotes, etc. Can someone verify for me? I tried in browser stack and I think it helps, but I am not familiar with iOS to be honest.


srjfoo:
I tried testing it:
* Turned smart punctuation on (I usually keep it off -- if I want smart quotes, etc, I'll enter them manually!)
* connected my external bluetooth keyboard
* typed single and double-quotes and double-hyphens.
All are still converted when I have smart punctuation on.

But I'd like independent verification. My iPad is coming up on 7 years old, and the keyboard cover that I used to use broke years ago, and I didn't replace it.

chrismiceli:
> I tried testing it:
> 
>     * Turned smart punctuation on (I usually keep it off -- if I want smart quotes, etc, I'll enter them manually!)
> 
>     * connected my external bluetooth keyboard
> 
>     * typed single and double-quotes and double-hyphens.
>       All are still converted when I have smart punctuation on.
> 
> 
> But I'd like independent verification. My iPad is coming up on 7 years old, and the keyboard cover that I used to use broke years ago, and I didn't replace it.

Digging deeper, it seems it is tied only to the `spellcheck="false"`` attribute
https://github.com/WebKit/webkit/blob/main/Source/WebKit/UIProcess/ios/WKContentViewInteraction.mm#L6772

I think though that this is a useful feature otherwise to catch scannos. That means the next idea would be to catch the smart quote events in JS, but that is a little bit hackier. Are there scenarios where we ever want smart quotes in the text? If not, we could just fix it on save maybe or after ever keypress event? @cpeel

srjfoo:
> Are there scenarios where we ever want smart quotes in the text? If not, we could just fix it on save maybe or after ever keypress event? @cpeel

Single and double curly quotes and em-dashes are among the characters that are rejected on project load, because we decided when prepping for the UTF-8 work that we were going to continue to require straight quotes/apostrophes and double hyphen-minus characters in the proofreading interface. So, no, as things stand, ther will never be a case where we want smart quotes.

I think there was reluctance to fix characters like that on save, because then you're changing what the user input without them having a chance to preview and approve it. That's why we have the built-in check for disallowed characters before saves.

I don't know how advisable it is to check for disallowed characters after each keypress event. Is that really a viable solution?

cpeel:
The idea of changing curly quotes to straight quotes at entrypoint is an interesting one, because that effectively undoes the problem but well before the user tries to save the page -- so we maintain our goal of not changing the page after the user hits save / done.

I've no idea how practical / possible it is but I think that's worth pursuing if you want.

chrismiceli:
> The idea of changing curly quotes to straight quotes at entrypoint is an interesting one, because that effectively undoes the problem but well before the user tries to save the page -- so we maintain our goal of not changing the page after the user hits save / done.
> 
> I've no idea how practical / possible it is but I think that's worth pursuing if you want.

Ok, I think I've implemented this, though hard to test since browserstack doesn't let me configure iOS's smart quote settings. Can we pass this along to someone with a real iPad?

tools/proofers/proof_validate.js
```diff
@@ -44,4 +44,18 @@ window.addEventListener('DOMContentLoaded', function() {
             }
         });
     });
+
+    const conversionMap = {
+        '‘': '\'',
+        '“': '"',
+        '’': '\'',
+        '”': '"',
+        '—': '-'
```
srjfoo, line 53: The em-dash needs to be replaced with a double hyphen-minus (U+002d), not a single one.

srjfoo (changes requested):

tools/proofers/proof_validate.js
srjfoo: I just spotted another potential issue. The cursor is moved to the end of the page after the substitution.

chrismiceli: Thanks, I should've addressed both of these. Let me know how your testing goes.

srjfoo (approved):
LGTM! (both on computer and iPad.

It would probably be a good idea for it to be confirmed on Windows that nothing weird happens?

cpeel (approved):
This is _very_ snazzy!

PR #1036, opened 2024-01-28
Remove moment dependency

chrismiceli:
https://www.pgdp.org/~chrismiceli/c.branch/bug_2072/tools/project_manager/projectmgr.php
Note the format is maintained

### DIFF
--- a/tools/project_manager/projectmgr.js
+++ b/tools/project_manager/projectmgr.js
@@ -6,7 +6,7 @@ window.addEventListener("DOMContentLoaded", () => {
     const timeSpan = document.getElementById("server-time");
     const dateTimeFormat = new Intl.DateTimeFormat(userLanguage, {
         timeZone: serverTimezone,
-        hourCycle: 'h24',
+        hourCycle: 'h23',
         hour: '2-digit',
         minute: '2-digit',
         weekday: 'short'
--- a/tools/project_manager/projectmgr.js
+++ b/tools/project_manager/projectmgr.js
@@ -1,14 +1,15 @@
-/*global serverTimezone*/
+/*global serverTimezone userLanguage*/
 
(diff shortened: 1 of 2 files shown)


tools/project_manager/projectmgr.js
```diff
@@ -1,12 +1,18 @@
-/*global moment serverTimezone*/
+/*global serverTimezone*/
 
 window.addEventListener("DOMContentLoaded", () => {
     "use strict";
 
-    var timeSpan = document.getElementById("server-time");
+    const timeSpan = document.getElementById("server-time");
+    const dateTimeFormat = new Intl.DateTimeFormat('en-US', {
```
cpeel, line 7: This needs to match the user's locale so French users get this in French. You probably want `get_user_language()` but that uses an underscore instead of a dash (eg: `en_US` vs `en-US`).

cpeel (commented):
The new date format is fine with me but I defer to Sharon who has a better idea on what our users would be happy with.

srjfoo (approved):
> The new date format is fine with me but I defer to Sharon who has a better idea on what our users would be happy with.

My main concern was that there might have been those who complained about losing the day of the week. Looks like @chrismiceli managed to find the right magic incantation, and the current version looks great. I've checked it out in both French and German on the test server.

cpeel (approved):

70ray (approved):

chrismiceli:
We'll want to confirm the timezone sent from the server in prod is compatible with the Intl datetimeformatter after deployment. I don't have PM access to check myself @srjfoo, @cpeel

srjfoo:
> We'll want to confirm the timezone sent from the server in prod is compatible with the Intl datetimeformatter after deployment. I don't have PM access to check myself @srjfoo, @cpeel

Definitely will look at that after deployment.

PR #1037, opened 2024-01-28
api1/v1/queues: Add a first cut read-only API for querying release qu…

bpfoley:
…eues

Much of this can probably be implemented better, but this is a first pass for discussion.

Example use:
API_KEY=...
ROOT=https://www.pgdp.org/~bfoley/c.branch/api-queue FILTER="show=enabled&group[]=Uber&group[]=PM&name[]=Eng&name[]=LOTE" FIELDS="field[]=name&field[]=group&field[]=ordering&field[]=release_criterion" curl -X GET "$ROOT/api/v1/queues/P1?$FILTER&$FIELDS" \
    -H "X-API-KEY: $API_KEY"

### DIFF
--- a/api/dp-openapi.yaml
+++ b/api/dp-openapi.yaml
@@ -690,6 +690,202 @@ paths:
         default:
           $ref: '#/components/responses/UnexpectedError'
 
+  /queues:
+    get:
+      tags:
+      - queues
+      description: Gets the release queues for a round (or all rounds)
+      parameters:
+      - name: roundid
+        in: query
+        description: Round ID for queues (all rounds if not provided)
+        schema:
+          type: string
+      - name: field
+        description: Queue field to return, if not set all fields are returned.
+          <br><br>
+          To request multiple fields, append with [] and pass one per desired
+          field (e.g. field[]=projectid&field[]=author).
+        in: query
+        schema:
+          type: string
+      - name: show
+        description: Which queues to show.
+          <dl>
+          <dt>all</dt> <dd>all queues</dd>
+          <dt>enabled</dt> <dd>queues that are enabled</dd>
+          <dt>populated</dt> <dd>queues that are enabled and have projects to release</dd>
+          </dl>
+        in: query
+        schema:
+          type: string
+          enum: [all, enabled, populated]
+      - name: name
+        description: Release queue name substring to match. If not set, all queues are returned.
+          <br><br>
+          To request multiple possible queue names (results ORed together), append with []
+          and pass one per desired name (e.g. name[]=Klingon&name[]=Shakespeare).
+        in: query
+        schema:
+          type: string
+      - name: group
+        description: Release queue group name substring to match. If not set all groups are
+          returned (as well as all matching queues in those groups).
+          <br><br>
+          To match multiple possible group names (results ORed together), append with []
+          and pass one per desired name (e.g. group[]=PM&group[]=Math).
+        in: query
+        schema:
+          type: string
+
+      responses:
+        200:
+          description: Requested release queue list
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/release_queue_detail'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '429':
+          $ref: '#/components/responses/RateLimitExceeded'
+        default:
+          $ref: '#/components/responses/UnexpectedError'
+
+  /queues/{queueid}/stats:
+    get:
+      tags:
+      - queues
+      description: Get release queue stats
+      parameters:
+      - name: queueid
+        in: path
+        description: Queue ID
+        required: true
+        schema:
+          type: integer
+      responses:
+        200:
+          description: Release queue stats
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  type: object
+                  properties:
+                    days_ago:
+                      type: integer
+                    projects_released:
+                      type: integer
+                    pages_released:
+                      type: integer
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '429':
+          $ref: '#/components/responses/RateLimitExceeded'
+        default:
+          $ref: '#/components/responses/UnexpectedError'
+
+  /queues/{queueid}/projects:
+    get:
+      tags:
+      - queues
+      description: Get projects in release queue
+      parameters:
+      - name: queueid
+        in: path
+        description: Queue ID
+        required: true
+        schema:
+          type: integer
+      - name: unheld_only
+        in: query
+        schema:
+          type: boolean
+      - name: field
+        description: Project field to return, if not set all fields are returned.
+          <br><br>
+          To request multiple fields, append with [] and pass one per desired
+          field (e.g. field[]=title&field[]=author).
+        in: query
+        schema:
+          type: string
+
+      responses:
+        200:
+          description: Release queue projects
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  type: object
+                  properties:
+                    projectid:
+                      type: string
+                    title:
+                      type: string
+                    author:
+                      type: string
+                    languages:
+                      type: array
+                      items:
+                        type: string
+                      minItems: 1
+                      maxItems: 2
+                    genre:
+                      type: string
+                    difficulty:
+                      type: string
+                    last_state_change_time:
+                      type: string
+                      format: dateTime
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '429':
+          $ref: '#/components/responses/RateLimitExceeded'
+        default:
+          $ref: '#/components/responses/UnexpectedError'
+
+  /queues/{queueid}:
+    get:
+      tags:
+      - queues
+      description: Get release queue information
+      parameters:
+      - name: queueid
+        in: path
+        description: Queue ID
+        required: true
+        schema:
+          type: integer
+      responses:
+        200:
+          description: Release queue information
+          content:
+            application/json:
+              schema:
+
+                type: object
+                $ref: '#/components/schemas/release_queue_detail'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '429':
+          $ref: '#/components/responses/RateLimitExceeded'
+        default:
+          $ref: '#/components/responses/UnexpectedError'
+
   /stats/site:
     get:
       tags:
@@ -914,6 +1110,31 @@ components:
         text:
           type: string
         state:
+          type: string
+
+    release_queue_detail:
+      properties:
+        id:
+          type: integer
+        round_id:
+          type: string
+        ordering:
+          type: integer
+        name:
+          type: string
+        enabled:
+          type: boolean
+        populated:
+          type: boolean
+        length:
+          type: integer
+        unheld_length:
+          type: integer
+        comment:
+          type: string
+        project_selector:
+          type: string
+        release_criterion:
           type: string
 
     charsuite:
--- a/api/v1.inc
+++ b/api/v1.inc
@@ -4,6 +4,7 @@ include_once("ApiRouter.inc");
 // DP API v1
 include_once("v1_validators.inc");
 include_once("v1_projects.inc");
+include_once("v1_queues.inc");
 include_once("v1_stats.inc");
 
 $router = ApiRouter::get_router();
@@ -14,6 +15,7 @@ $router->add_validator(":projectid", "validate_project");
 $router->add_validator(":roundid", "validate_round");
 $router->add_validator(":pagename", "validate_page_name");
 $router->add_validator(":pageroundid", "validate_page_round");
+$router->add_validator(":queueid", "validate_release_queue");
 
 // Add routes
 $router->add_route("GET", "v1/projects", "api_v1_projects");
@@ -36,6 +38,12 @@ $router->add_route("PUT", "v1/projects/:projectid/wordlists/:wordlist_type", "ap
 $router->add_route("GET", "v1/projects/:projectid/pages", "api_v1_project_pages");
 $router->add_route("GET", "v1/projects/:projectid/pagedetails", "api_v1_project_pagedetails");
 $router->add_route("GET", "v1/projects/:projectid/pages/:pagename/pagerounds/:pageroundid", "api_v1_project_page_round");
+
+$router->add_route("GET", "v1/queues", "api_v1_queues");
+$router->add_route("GET", "v1/queues/:queueid", "api_v1_queue");
+$router->add_route("GET", "v1/queues/:queueid/stats", "api_v1_queue_stats");
+$router->add_route("GET", "v1/queues/:queueid/projects", "api_v1_queue_projects");
+
 
 $router->add_route("GET", "v1/stats/site", "api_v1_stats_site");
 $router->add_route("GET", "v1/stats/site/rounds", "api_v1_stats_site_rounds");
--- a/api/v1_queues.inc
+++ b/api/v1_queues.inc
@@ -1,0 +1,109 @@
+<?php
+include_once($relPath.'release_queue.inc');
+include_once($relPath.'user_is.inc');
+include_once("exceptions.inc");
+include_once("api_common.inc");
+
+// DP API v1 -- Queues
+
+//---------------------------------------------------------------------------
+// queues -- list queues, optionally filtered by round
+function api_v1_queues($method, $data, $query_params)
+{
+    $roundid = array_get($query_params, "roundid", null);
+    $round = !is_null($roundid) ? validate_round($roundid, null) : null;
+
+    $show = get_enumerated_param($query_params, "show", "all", ["enabled", "populated", "all"]);
+
+    // Fields we  can query on
+    $valid_fields = [
+        "name",
+        "group",
+    ];
+
+    $valid_return_fields = [
+        "id",
+        "round_id",
+        "group",
+        "ordering",
+        "name",
+        "enabled",
+        "populated",
+        "length",
+        "unheld_length",
+        "comment",
+    ];
+
+    if (user_can_see_queue_settings()) {
+        array_push($valid_return_fields, "project_selector", "release_criterion");
+    }
+
+    // restrict to list of desired fields, if set
+    $return_fields = array_get_as_array($query_params, "field", $valid_return_fields);
+    $return_fields = array_intersect($return_fields, $valid_return_fields);
+
+    $name_selector = array_get($query_params, "name", null);
+    $group_selector = array_get($query_params, "group", null);
+
+    $output = [];
+    foreach (fetch_queues_data($round, $show, false, $name_selector, $group_selector) as $queue_data) {
+        array_remove_invalid_fields($queue_data, $return_fields);
+        $output[] = $queue_data;
+    }
+    return $output;
+}
+
+//---------------------------------------------------------------------------
+// queues/:queueid -- return release queue info
+function api_v1_queue($method, $data, $query_params)
+{
+    $queue = $data[":queueid"];
+
+    return $queue;
+}
+
+//---------------------------------------------------------------------------
+// queue/:queueid/stats -- return release queue stats
+function api_v1_queue_stats($method, $data, $query_params)
+{
+    $queue = $data[":queueid"];
+
+    $round = validate_round($queue["round_id"], null);
+    $name = $queue["name"];
+
+    return fetch_queue_stats_data($round, $name);
+}
+
+//---------------------------------------------------------------------------
+// queue/:queueid/projects -- list projects in release queue
+function api_v1_queue_projects($method, $data, $query_params)
+{
+    $queue = $data[":queueid"];
+    $unheld_only = (bool) get_flag_value($query_params, "unheld_only");
+
+    $valid_return_fields = [
+        "projectid",
+        "title",
+        "author",
+        "languages",
+        "genre",
+        "difficulty",
+        "last_state_change_time",
+    ];
+
+    // restrict to list of desired fields, if set
+    $return_fields = array_get_as_array($query_params, "field", $valid_return_fields);
+    $return_fields = array_intersect($return_fields, $valid_return_fields);
+
+    $round = validate_round($queue["round_id"], null);
+
+    $output = [];
+    foreach(fetch_queue_projects_data($round, $queue["project_selector"], $unheld_only) as $proj_data) {
+        array_remove_invalid_fields($proj_data, $return_fields);
+        if (isset($proj_data["last_state_change_time"])) {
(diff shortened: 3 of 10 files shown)


api/v1_queues.inc
```diff
@@ -0,0 +1,161 @@
+<?php
+include_once($relPath.'user_is.inc');
+include_once("exceptions.inc");
+
+// TODO(bpfoley): There is a lot of duplicated logic betweeen this file and
```
cpeel, line 5: Yeah, we should probably separate the data-gathering from the presentation and re-use the data bits here and in `stats/relese_queue.php`. I had to do this for the page tables in 81af9d4b5c614983fb1d09cc2e947d07dd136e03 for the API too and we could use a similar pattern here. These new data-specific functions probably belong in `pinc/release_queue.inc`.

api/v1_queues.inc
```diff
@@ -0,0 +1,161 @@
...
+//       stats/release_queue.php's _show_round_queues
+// TODO(bpfoley): Add openAPI schema
+// NB because of how the queue_defns table is read in order, and queues
```
cpeel, line 8: <nod> At this point I wouldn't worry about pagination for the (good) reasons you've highlighted.

api/v1_queues.inc
```diff
@@ -0,0 +1,161 @@
...
+    $where .= " OR (SUBSTR(name, 1, 1) = '*')";
+    $q_sql = "
+        SELECT *
+        FROM queue_defns
+        WHERE $where
+        ORDER BY ordering
+    ";
+    $result = DPDatabase::query($q_sql);
+    if (!$result) {
+        throw new ServerError(DPDatabase::log_error());
```
cpeel, line 61: We intentionally do not display DB messages to the end-user (ask me in Slack if you care to know the details on why). The `DPDatabase::query()` call above will throw an exception on query error (see https://github.com/DistributedProofreaders/dproofreaders/blob/master/pinc/DPDatabase.inc#L24 and https://github.com/DistributedProofreaders/dproofreaders/blob/master/pinc/DPDatabase.inc#L90) unless you tell it otherwise so this will never get used. In general we should rely on `DPDatabase::query()` doing The Right Thing on failure and if it returns then its return value is valid.

api/v1_queues.inc
cpeel: The initial data function version should probably just return all the data to the caller and the API can filter out the parts the user doesn't care about, but a future optimization could be to have the data function not generate the data that isn't used as some (like total number of projects matching the queue definition) are more DB-taxing than others. But that's an optimization step and I wouldn't bother in the MVP.

cpeel (commented):
Looks like it's all in the right direction to me! Ideally we'd have a PR that just does the refactor that pulls out the data-collection from the current UI presentation and then a follow-up with this one that then uses it. I would probably develop them together to make sure the data-collection function returns everything in the way I would want for the API too. Reviewing them seprately, however, would be appreciated.

api/v1_queues.inc
bpfoley: Ah, interesting. I copied the pattern from `v1_projects.inc` without really understanding what it was doing.

cpeel: I am, sadly, not at all surprised that you pulled it from elsewhere in the codebase. And without `git blame` I can tell you that it was my code you copied. 😕

api/misc.inc
```diff
@@ -0,0 +1,24 @@
+<?php
+/**
+ * Check to see if the user requested a flag parameter.
+ *
+ * This returns one of three possible states:
+ * - null - no parameter with this key was set
+ * - true - parameter was set to "true" or an empty string (for ?flagname)
+ * - false - parameter was set to "false"
+ */
+function _get_flag_value($query_params, $flagname)
```
cpeel, line 10: Let's remove the initial `_` as that's suppose to be an indicator that it's only used in the file it was declared in.

api/v1.inc
```diff
@@ -37,6 +39,9 @@ $router->add_route("GET", "v1/projects/:projectid/pages", "api_v1_project_pages"
 $router->add_route("GET", "v1/projects/:projectid/pagedetails", "api_v1_project_pagedetails");
 $router->add_route("GET", "v1/projects/:projectid/pages/:pagename/pagerounds/:pageroundid", "api_v1_project_page_round");
 
+$router->add_route("GET", "v1/queues/:roundid", "api_v1_queues");
+$router->add_route("GET", "v1/queue/:queueid", "api_v1_queue");
```
cpeel, line 43: Our convention, for better or worse, has been to use plural nouns even for pulling a singular object (eg: `v1/projects/:projectid`. For consistency, we might consider using:
```
v1/queues/rounds/:roundid
v1/queues/:queueid
```
That mirrors how we do stats too: `v1/stats/site/rounds/:roundid`.

pinc/release_queue.inc
```diff
@@ -20,3 +25,322 @@ function cook_project_selector($project_selector)
...
+ * {
+ *    "is_group": true,
+ *    "group": "PM Queues (experimental)",
+ * }
+ * ```
+ * For rows that represent queues, the returned format looks like:
+ * ```
+ * {
+ *     "is_group": false,
+ *     "group": "PM Queues (experimental)",
```
cpeel, line 66: Also `id` and `round_id`?

pinc/release_queue.inc
```diff
@@ -20,3 +25,322 @@ function cook_project_selector($project_selector)
...
+ * ```
+ *
+ * @param Round $round
+ *   Return the queues for the specified round.
+ * @param string $show
+ *   Select which queues to return:
+ *   * all - Don't filter out any queues
+ *   * enabled - Filter out queues which aren't enabled
+ *   * populated - Filter out queues which aren't enabled *OR* contain no projects
+ * * @param null|string|array $name_selector
```
cpeel, line 85: s/* * / * /

api/dp-openapi.yaml
```diff
@@ -690,6 +690,160 @@ paths:
...
+          type: string
+      - name: field
+        description: Queue field to return, if not set all fields are returned.
+          To request multiple fields, append with [] and pass one per desired
+          field (e.g. field[]=projectid&field[]=author).
+        in: query
+        schema:
+          type: string
+      - name: show
+        description: Which queues to show.
```
cpeel, line 713: We should confirm how this is rendered as I think it's just going to all run together as-is. IIRC we use HTML in other descriptions to force formatting.

api/dp-openapi.yaml
```diff
@@ -690,6 +690,160 @@ paths:
...
+      parameters:
+      - name: queueid
+        in: path
+        description: Queue ID
+        required: true
+        schema:
+          type: integer
+      - name: unheld_only
+        in: query
+        required: false
```
cpeel, line 769: `required: false` is the default so you don't need this string and we don't explicitly define it in most (but not all?) places in this file when its applicable.

pinc/release_queue.inc
```diff
@@ -1,8 +1,13 @@
 <?php
 // A place for shared functions dealing with release queues.
 
+include_once($relPath.'misc.inc');
+include_once($relPath.'user_is.inc');
+
 $today_MMDD = date('md', strtotime('today'));
 $tomorrow_MMDD = date('md', strtotime('tomorrow'));
+$user_can_see_queue_settings = $ordinary_users_can_see_queue_settings ||
```
cpeel, line 9: If you want this to be really global, we need to include `global $user_can_see_queue_settings;` before this. Otherwise it could be inadvertently not global if this file is included inside another function, for instance.

I don't love that we're shoving another global out there but we're just moving it around from where it was before I suppose. Ideally this would be something like:
```diff
--- a/pinc/User.inc
+++ b/pinc/User.inc
@@ -152,6 +152,8 @@ class User

     public function __get($name)
     {
+        global $ordinary_users_can_see_queue_settings;
+
         switch ($name) {
             case "profile":
                 if ($this->current_profile) {
@@ -191,6 +193,9 @@ class User
                 }

                 return $this->current_profile;
+            case "can_see_queue_settings":
+                return $ordinary_users_can_see_queue_settings ||
+                    user_is_a_sitemanager() || user_is_proj_facilitator();
             default:
                 return $this->table_row[$name];
         }
```
and then here:
```php
User::load_current()->can_see_queue_settings;
```

pinc/release_queue.inc
```diff
@@ -20,3 +25,322 @@ function cook_project_selector($project_selector)
...
+        // Check for project in group.
+        if (!str_contains_any_of($group, $group_selector)) {
+            continue;
+        }
+
+        [$length, $unheld_length] = _get_queue_length($row["project_selector"], $round->project_waiting_state);
+        if ($show == "populated" && $length == 0) {
+            continue;
+        }
+        $data = _queue_data($row);
```
cpeel, line 159: Running `_get_queue_length()` against a whole lot of queues on a site with a whole lot of projects (eg: PROD) is non-trivial, so we should try not to do it twice for a single row -- here on line 155 and again in `_queue_data()`. It's probably worth special casing it, something like:
```php
function _queue_data($row, $length = null, $unheld_length = null)
{
    if ($length === null || $unheld_lenth === null) {
        [$length, $unheld_length] = ...
    }
```
and then here: `_queue_data($row, $length, $unheld_length)`

stats/release_queue.php
```diff
@@ -291,75 +234,35 @@ function _show_queue_details($round, $name, $unheld_only)
...
-        ",
-        DPDatabase::escape($waiting_state),
-        DPDatabase::escape($name),
-        $seconds_ago
-    );
-    $result = DPDatabase::query($sql);
-    $pages_released = mysqli_fetch_row($result)[0] ?? 0;
-
-    return [$projects_released, $pages_released];
+    // TODO(bpfoley) Add a dump_themed_array(), perhaps?
```
cpeel, line 237: Totally grok the desire but this gets back into `dpsql_dump_themed_query_result()`-esque which is not a road I want to go down because turns out we have very few standard tables to output. I have to believe there's a better, more frameworky way to handle this.

cpeel (commented):
Overall this looks _excellent_. I haven't done any testing on it yet but here's a code-level review.

api/dp-openapi.yaml
bpfoley: OK. I'll remove the ones I've introduced for now.

pinc/release_queue.inc
bpfoley: Good idea! Done.

pinc/release_queue.inc
bpfoley: OK. I've moved it to a free function in user_is.inc, since it already relies on functions from that file.

stats/release_queue.php
```diff
@@ -291,75 +231,34 @@ function _show_queue_details($round, $name, $unheld_only)
...
+        _("Date Last Modified"),
+        _("Hold?"),
+    ];
+    echo "<table class='themed theme_striped stats' style='width: 100%; table-layout: fixed'>\n";
+    echo "<tr>\n";
+    echo surround_and_join(array_map("html_safe", $headers), "<th>", "</th>\n", "");
+    echo "</tr>\n";
+    foreach(fetch_queue_projects_data($round, $queue["project_selector"], $unheld_only) as $p) {
+        echo "<tr>\n";
+        echo "<td><a href='$code_url/project.php?id=", html_safe($p["project_id"]), "'>", $p["title"], "</a></td>\n";
```
cpeel, line 250: `$p["title"]` needs `html_safe()` and `html_safe($p["project_id"])` should use `attr_safe()` instead although honestly I'd just output it directly since we control those and know they don't have anything special in them.

stats/release_queue.php
```diff
@@ -291,75 +231,34 @@ function _show_queue_details($round, $name, $unheld_only)
...
+        echo "<td>", html_safe($p["genre"]), "</td>\n";
+        echo "<td>", html_safe($p["difficulty"]), "</td>\n";
+        echo "<td>", html_safe($p["username"]), "</td>\n";
+        echo "<td>", date("Y-m-d H:i:s", $p["modified_date"]), "</td>\n";
+        echo "<td>", is_null($p["holds_state"]) ? "&nbsp;" : "Y", "</td>\n";
+        echo "</tr>";
+    }
+    echo "</table>";
 }
-
```
cpeel, line 362: Missing newline between functions.

cpeel (commented):
Saw a couple of super minor stuff but this LGTM! Still haven't had a chance to test it yet but intend to.

srjfoo:
> Still haven't had a chance to test it yet but intend to.

Ditto.

stats/release_queue.php
bpfoley: <shakes head> It's so easy to miss these! Thanks.

pinc/release_queue.inc
```diff
@@ -20,3 +20,324 @@ function cook_project_selector($project_selector)
...
+            if ($show_groups && str_contains_any_of($group, $group_selector)) {
+                yield [
+                    "is_group" => true,
+                    "group" => $group,
+                ];
+            }
+            continue;
+        }
+
+        if (!is_null($group) && !str_contains_any_of($group, $group_selector)) {
```
cpeel, line 145: I think something in here needs tweaking because right now queues without a group (whose return value is `null`) are always included when requesting members of a specific group, which was unexpected.
```bash
curl -i -H "Accept: application/json" -H "X-API-KEY: $API_KEY"
     -X GET "https://www.pgdp.org/~bfoley/c.branch/api-queue/api/v1/queues/rounds/P1?group=General"
```
for instance, includes:
```json
    {
        "id": 14,
        "round_id": "P1",
        "ordering": 540,
        "name": "George Sand Centennary",
        "enabled": true,
        "populated": false,
        "length": 0,
        "unheld_length": 0,
        "comment": "Special queue for Centennary of George Sand",
        "project_selector": "username = 'PM-Sand'",
        "release_criterion": "projects < 10",
        "group": null
    },
```

api/v1_queues.inc
```diff
@@ -0,0 +1,101 @@
...
+        "modified_date",
+    ];
+
+    // restrict to list of desired fields, if set
+    $return_proj_fields = array_get_as_array($query_params, "proj_field", $valid_proj_return_fields);
+    $return_proj_fields = array_intersect($return_proj_fields, $valid_proj_return_fields);
+
+    $output = [];
+
+    if ($show_queue) {
```
cpeel, line 81: When using it, this was unexpected. Having the endpoint not return any data on a valid request was super confusing. Upon reflection I wonder if either
1. this should be 3 separate RESTful endpoints:
   * `v1/queues/:queueid` -- $show_queue=true
   * `v1/queues/:queueid/stats` -- $show_stats=true
   * `v1/queues/:queueid/projects` -- $show_projects=true
2. the endpoint should always return queue details ($show_queue=true) and keep the separate args for $show_stats and $show_projects
3. the endpoint should return everything by default and you can select which field you want from ["queue", "stats", "projects"]

The first option feels the most RESTful, but it adds 3x the round trips for the same data. What parts of this endpoint are useful to you as an end-user, Brian? Do you want/need all of this and you'd rather do 1 query instead of 3?

api/dp-openapi.yaml
```diff
@@ -690,6 +690,162 @@ paths:
...
+                        projects_released:
+                          type: integer
+                        pages_released:
+                          type: integer
+                  projects:
+                    type: array
+                    items:
+                      type: object
+                      properties:
+                        project_id:
```
cpeel, line 824: This should be `projectid` (here and in the code) to match what the v1/projects endpoint returns.

api/dp-openapi.yaml
```diff
@@ -690,6 +690,162 @@ paths:
...
+                          type: string
+                        title:
+                          type: string
+                        author:
+                          type: string
+                        language:
```
cpeel, line 830: This should probably be `languages` and return an array of the languages like the v1/projects endpoint output.

api/dp-openapi.yaml
```diff
@@ -690,6 +690,162 @@ paths:
...
+                          type: string
+                        genre:
+                          type: string
+                        difficulty:
+                          type: string
+                        modified_date:
```
cpeel, line 836: `last_state_change_time` to match v1/projects endpoint output.

cpeel (commented):
Nothing like using it to _really_ look at it!

pinc/release_queue.inc
srjfoo: I was playing with this last night -- and queues without a group didn't show up at all; should we add a miscellaneous group for those that don't have one? If I'm not mistaken, everything after the beginning of the first group has to be in a group. So, why doesn't that first set of queues have a group?

bpfoley: Yeah, this was the scenario I was trying to handle, but I didn't think through all the corner cases. I think we have 3 cases to worry about (and really cases 2 & 3 are the same).

1. If group is null (ie the query parameter isn't provided), all queues in a group, and all queues with no group should be shown. This is what the `!is_null($group)` workaround was meant to handle, but it's wrong, it should be `!is_null($group_selector)`.
2. If group is set once, then all queues with str_contains($queue_group, $group) should be returned. This will never include queues in no group.
3. If group is set multiple times, then queues with where str_contains($queue_group, $group[i]) for any group will be returned

This does mean that we have no way of explicitly filtering just the 'no group' queues, but, eh... I'm not sure that's a big problem.

api/v1_queues.inc
bpfoley: Really, I just need the projects part, so perhaps that should be a separate endpoint.

There's definitely some attraction to having a `section` parameter so you could say `&section[]=info&section[]=stats&section[]=projects`, but I see that this does start to get murkier if we ever want to add PUT verb to the endpoint. It's meaningless to write to either the projects or stats section.

So yeah, 3 endpoints sounds like it's the right way to go.

api/dp-openapi.yaml
bpfoley: Fixed

api/dp-openapi.yaml
bpfoley: Aah! I didn't realize this was done by Project.inc. I assumed it was in the database. Will do.

api/dp-openapi.yaml
bpfoley: Done.

api/v1_queues.inc
bpfoley: Done.

api/v1.inc
```diff
@@ -37,6 +39,12 @@ $router->add_route("GET", "v1/projects/:projectid/pages", "api_v1_project_pages"
 $router->add_route("GET", "v1/projects/:projectid/pagedetails", "api_v1_project_pagedetails");
 $router->add_route("GET", "v1/projects/:projectid/pages/:pagename/pagerounds/:pageroundid", "api_v1_project_page_round");
 
+$router->add_route("GET", "v1/queues/rounds/:roundid", "api_v1_queues_rounds");
```
cpeel, line 42: Now that we have a unique ID for each queue we can treat them more like objects with round_id and name rather than only unique by (round_id, name) pairs. So what this means is that we can now have
`v1/queues` to return all of the queues (which include id, roundid, and name) and pass in `roundid` as a selection parameter to just get the queues for a specific round (v1/queues?roundid=P1). This then mirrors what we do for projects. This should be pretty easy to support by making `round_id` in the fetch function default to `null` and then just not filtering by roundid when null _but_ in all cases sorting by `round_id, ordering` instead of just `ordering`. It does mean that we need to keep track of when the results switch from one round to another to reset the group to null.

Also to mirror the project endpoints the function names should be massaged to:
```php
$router->add_route("GET", "v1/queues", "api_v1_queues");
$router->add_route("GET", "v1/queues/:queueid", "api_v1_queue");  # singular queue
$router->add_route("GET", "v1/queues/:queueid/stats", "api_v1_queue_stats");  # singular queue
$router->add_route("GET", "v1/queues/:queueid/projects", "api_v1_queue_projects");  # singular queue
```

cpeel (commented):
I realize this is stupidly late in the review cycle to suggest this change but it wasn't until I was playing with all of the queues endpoints that I realized this non-parallel to `v1/projects`.

api/v1.inc
bpfoley: It's good that we've gone through a few iterations on this. We've ended up with a much better API.

Done.

api/v1_queues.inc
```diff
@@ -0,0 +1,125 @@
...
+
+//---------------------------------------------------------------------------
+// queues -- list queues, optionally filtered by round
+function api_v1_queues($method, $data, $query_params)
+{
+    global $Round_for_round_id_;
+
+    $roundid = array_get($query_params, "roundid", null);
+    if (!is_null($roundid)) {
+        try {
```
cpeel, line 18: FWIW, you can use `validate_round()` here:
```php
$round = validate_round($roundid, null);
```

api/v1_queues.inc
```diff
@@ -0,0 +1,125 @@
...
+}
+
+//---------------------------------------------------------------------------
+// queue/:queueid/stats -- return release queue stats
+function api_v1_queue_stats($method, $data, $query_params)
+{
+    global $Round_for_round_id_;
+    $queue = $data[":queueid"];
+
+    $round = $Round_for_round_id_[$queue["round_id"]];
```
cpeel, line 86: You could probably use `validate_round()` here too.

cpeel (approved):
SHIP IT! 🚀

api/v1_queues.inc
bpfoley: Done

api/v1_queues.inc
bpfoley: Done

api/misc.inc
```diff
@@ -0,0 +1,24 @@
+<?php
```
70ray, line 1: Would it be better to give this file a different name since there is already a /pinc/misc.inc. Or even a name reflecting what it does.

cpeel (approved):

api/misc.inc
bpfoley: Hmm. Naming is hard, and all that. The problem is, I just needed somewhere to put a function that's used by two files and I don't know what else might end up in there. `misc.inc` is terribly vague, I'll give you that. How about `common.inc`?

cpeel: Other name brainstorming: `api_misc.inc`? `api_common.inc`? (yes, they are already in `api/` so it's a bit redundant)

srjfoo: Either of these sound reasonable to me. Even if it's redundant, it can be helpful if all you have to go by is the filename, instead of the full path.

pinc/release_queue.inc
```diff
@@ -20,3 +21,324 @@ function cook_project_selector($project_selector)
...
+        return [$row[0], $row[1] ?? 0];
+    } catch (DBQueryError $error) {
+        return [null, null];
+    }
+}
+
+/**
+ * Generator to return queue_defns table data for queues, one queue at a time.
+ *
+ * Data is returned as an array of associative array that abstracts away
```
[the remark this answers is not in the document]
bpfoley: Done.

api/misc.inc
bpfoley: `api_common.inc` it is...

srjfoo (approved):
Verified that if `ordinary_users_can_see_queue_settings` is set to `false`, a non-admin, non-PF user cannot see the `project_selector` or the `release_criterion`.

PR #1040, opened 2024-01-30
lint cleanups to SETUP docs, templates, and scripts

bpfoley:
These make TRUE/FALSE lower case and adjust whitespace so that php-cs-fix doesn't complain about the site config we use in sandboxes.


cpeel (approved):
Thanks for updating these!

srjfoo (approved):

chrismiceli (approved):

PR #1042, opened 2024-02-01
api/v1: Add stats/site/projects endpoint to count projects-in-state.

bpfoley:
Similar to the project counts on activity_hub.php.

Since it's two small DB queries for all the stats, and the output is <700 bytes, don't bother filtering or grouping the stats by round.


api/dp-openapi.yaml
```diff
@@ -1005,6 +1020,13 @@ components:
         enabled:
           type: boolean
 
+    project_stats:
```
srjfoo, line 1023: Should this be `projects_states`? To me, `project`, singular, is about an individual project's stats. (Ditto the descriptions where you're using `project`, singular.

This is just my confusion, as I'm not really 100% clear on when the singular should be used and when the plural.

api/v1_stats.inc
```diff
@@ -33,6 +33,54 @@ function api_v1_stats_site($method, $data, $query_params)
...
+        SELECT state, COUNT(*)
+        FROM projects
+        GROUP BY state
+    ";
+    $result = DPDatabase::query($sql);
+    while([$project_state, $count] = mysqli_fetch_row($result)) {
+        $output[$project_state] = (int)$count;
+    }
+
+    $sql = "
```
cpeel, line 71: Aside: I hate the way SR is hacked into our code.

SR isn't a _state_ of any kind (although it is a `Stage`) and unlike all of the others returned from this endpoint, it isn't a project state and thus isn't included in `v1/projects/states`. And although this endpoint doesn't _say_ "states" anywhere, it feels like we're mushing stuff together here.

Given that, what do you think about changing this endpoint to `v1/stats/projects/states` (which should include return values for all of the possible project states, with 0 for those without any current entries) and creating an additional endpoint `v1/stats/projects/stages` that groups all the projects into the stage (round, pool, or SR) and return the counts that way?

eg (but with actual counts):
```json
{
    "P1": 1,
    "P2": 2,
    "P3": 3,
    "F1": 4,
    "F2": 5,
    "PP": 7,
    "PV": 8,
    "SR": 9
}
```

api/dp-openapi.yaml
bpfoley: Well, I'm just basing it on the `round_stats` schema name directly below.

And... re singular/plural: would you say 'height stats' or 'heights stats' for (eg) short/average/tall people? 'Income stats' or 'incomes stats'? For me, I'd use singular, but I don't know if there's a strict grammatical rule to follow.

FWIW, the schema name is just used as an internal label within the OpenAPI definition (the `$ref: '#/components/schemas/project_stats'` earlier in the file). It doesn't make any difference to the queries generated by tools like Swagger.

api/v1_stats.inc
bpfoley: Thanks. I've failed to make the distinction between Stages and States multiple times before, so the reminder is helpful. I'll do as you suggested.

api/dp-openapi.yaml
srjfoo: Totally fair -- as I said, to a certain extent, it's about my confusion. And the fact that an individual project's stats would look different from stats about a group of projects.

api/v1_stats.inc
srjfoo: @cpeel , are you intentionally lumping Unavailable, Bad, Available and Waiting together for the rounds, and similar for PP, or was that just an example?

cpeel: I am, yes. At this level they're all part of that "stage". If the user really cares about the nuanced view of available/waiting/etc they can process the project states and get whatever combination of data they are looking for.

bpfoley: OK, done, but see caveat in commit message.

api/dp-openapi.yaml
```diff
@@ -703,6 +703,36 @@ paths:
...
+              schema:
+                type: object
+                additionalProperties:
+                  $ref: '#/components/schemas/project_stats'
+
+  /stats/site/projects/states:
+    get:
+      tags:
+      - stats
+      description: Gets site statistics for the number of projects in all stages
```
cpeel, line 725:
```suggestion
      description: Gets site statistics for the number of projects in all states
```

api/dp-openapi.yaml
```diff
@@ -1005,6 +1035,13 @@ components:
         enabled:
           type: boolean
 
+    project_stats:
+      type: object
+      additionalProperties:
+        type: integer
+      description: Project state statistics.
+        Keys are state ID, values are the number of projects in that state
```
cpeel, line 1043: This description is not true for both, we could probably generalize it and make it work though.

cpeel (approved):
ack the commit caveats, I'm fine with that limitation.

srjfoo (approved):

windymilla (approved):

PR #1043, opened 2024-02-02
api: Add v1/projects/:projectid/transitions

bpfoley:
Useful (amongst other things) to see when a project was created and when it left a round.

Try with:

ROOT=https://www.pgdp.org/~bfoley/c.branch/api-transitions curl -X GET "$ROOT/api/v1/projects/projectID5e23a810ef693/transitions" \
    -H "X-API-KEY: $API_KEY"

### DIFF
--- a/api/dp-openapi.yaml
+++ b/api/dp-openapi.yaml
@@ -690,6 +690,35 @@ paths:
         default:
           $ref: '#/components/responses/UnexpectedError'
 
+  /projects/{projectid}/transitions:
+    get:
+      tags:
+      - project
+      description: Gets a list of a project's transitions
+      parameters:
+      - name: projectid
+        in: path
+        description: ID of project
+        required: true
+        schema:
+          type: string
+      responses:
+        200:
+          description: List of a project's transitions, ordered by ascending time
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/project_transition'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '429':
+          $ref: '#/components/responses/RateLimitExceeded'
+        default:
+          $ref: '#/components/responses/UnexpectedError'
   /queues:
     get:
       tags:
@@ -1127,6 +1156,32 @@ components:
         last_modified:
           type: string
           format: dateTime
+
+    project_transition:
+      required:
+      - timestamp
+      - event_type
+      - details
+      type: object
+      properties:
+        timestamp:
+          type: string
+          format: dateTime
+        event_type:
+          type: string
+          enum: [creation, transition, archive]
+          description:
+            <dl>
+            <dt>creation</dt> <dd>Project created</dd>
+            <dt>transition</dt> <dd>Project moves between two stages, eg P3.proj_waiting to P3.proj_avail</dd>
+            <dt>archive</dt> <dd>Project content has been archived and is no longer directly accessible</dd>
+            </dl>
+        details:
+          type: array
+          items:
+            type: string
+            minItems: 3
+            maxItems: 3
 
     project_page:
       required:
--- a/api/v1_projects.inc
+++ b/api/v1_projects.inc
@@ -499,6 +499,33 @@ function render_project_page_json($row)
 }
 
 //---------------------------------------------------------------------------
+// projects/:projectid/transitions
+
+function api_v1_project_transitions($method, $data, $query_params)
+{
+    $sql = sprintf(
+        "
+        SELECT *
+        FROM project_events
+        WHERE projectid = '%s'
+        AND event_type in ('creation', 'transition', 'archive')
+        ORDER BY timestamp
+        ",
+        $data[":projectid"]->projectid
+    );
+    $result = DPDatabase::query($sql);
+    $return_data = [];
+    while($row = mysqli_fetch_assoc($result)) {
+        $return_data[] = [
+            "timestamp" => date(DATE_ATOM, (int)$row["timestamp"]),
+            "event_type" => $row["event_type"],
+            "details" => [$row["details1"], $row["details2"], $row["details3"]],
+        ];
+    }
+    return $return_data;
+}
+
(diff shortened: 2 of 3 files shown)


api/v1_projects.inc
```diff
@@ -498,6 +498,33 @@ function render_project_page_json($row)
...
+// projects/:projectid/transitions
+
+function api_v1_project_transitions($method, $data, $query_params)
+{
+    $sql = sprintf(
+        "
+        SELECT *
+        FROM project_events
+        WHERE projectid = '%s'
+        AND event_type in ('creation', 'transition')
```
cpeel, line 511: After looking at all of the possible `event_type`s* I think we should include `deletion` and `archive` as they seem like interesting "transitions" unlike the others which seem more like just "events".

`*` full set from PROD:
```
mysql> select count(*), event_type from project_events group by event_type;
+----------+----------------+
| count(*) | event_type     |
+----------+----------------+
|    67611 | creation       |
|  1020315 | transition     |
|   218565 | edit           |
|     1068 | deletion       |
|    38120 | archive        |
|    24320 | smooth-reading |
|    13872 | add_holds      |
|    13925 | remove_holds   |
+----------+----------------+
8 rows in set (4.04 sec)
```

api/dp-openapi.yaml
```diff
@@ -1128,6 +1157,26 @@ components:
...
+      type: object
+      properties:
+        timestamp:
+          type: string
+          format: dateTime
+        event_type:
+          type: string
+          enum: [creation, transition]
+        details:
+          type: array
```
cpeel, line 1174: I think these objects take a `description` and it might be worth adding something like:
```
description: Additional details about the transition.
  The schema of what is contained in these strings is undefined
  and the contents may change.
```

api/v1_projects.inc
srjfoo: I agree about adding deletion and archive events. This is nit intended to be used for the project history,?

cpeel: > This is not intended to be used for the project history?

No, not really. By focusing on (mostly) textual-free transitions we avoid the issue where `edit` events are both numerous and include what changed in the details which is further unstructured and involves a lot of [localization gymnastics](https://github.com/DistributedProofreaders/dproofreaders/blob/master/project.php#L1207).

api/dp-openapi.yaml
bpfoley: Done.

api/v1_projects.inc
srjfoo: > After looking at all of the possible `event_type`s* I think we should include `deletion` and `archive` as they seem like interesting "transitions" unlike the others which seem more like just "events".
> 
> `*` full set from PROD:
> 
> ```
> mysql> select count(*), event_type from project_events group by event_type;
> +----------+----------------+
> | count(*) | event_type     |
> +----------+----------------+
> |    67611 | creation       |
> |  1020315 | transition     |
> |   218565 | edit           |
> |     1068 | deletion       |
> |    38120 | archive        |
> |    24320 | smooth-reading |
> |    13872 | add_holds      |
> |    13925 | remove_holds   |
> +----------+----------------+
> 8 rows in set (4.04 sec)
> ```

Footnote: I got curious, because mostly we see a `transition` to `project_delete`. In fact, it looks like `deletion` as an `event_type` was restricted to new projects that had not moved out of the `project_new` state until early 2017. And all deletions (including from `project_new`) since then are logged as transitions to `project_delete`.

srjfoo: In fact, digging further, I pulled a handful of projectids on TEST that had an event_type of `deletion`, and all threw a nonexistent project exception (which would have been expected then because the projectids were not kept), and both via `curl` and Swagger, I get

```
{
    "error": "There is no project with projectid 'projectID44dcf4500f8fe'",
    "code": 101
}
```

Is that the expected response?

cpeel: It is, because that project doesn't exist in the `projects` table:
```
mysql> select * from projects where projectid='projectID44dcf4500f8fe';
Empty set (0.00 sec)
```

bpfoley: OK, but in that case does it mean we'll never be able to see a deleted project_event, because any projects that were deleted no longer have a row in `projects` and thus won't pass the validator?

srjfoo: That's what I would expect. It means that you can't really tell the difference between a real nonexistent projectid (i.e. one that never existed) and one that once existed and was then deleted.

70ray (approved):

windymilla (approved):

api/v1_projects.inc
cpeel: > OK, but in that case does it mean we'll never be able to see a deleted project_event, because any projects that were deleted no longer have a row in `projects` and thus won't pass the validator?

That's correct. Ideally a delete would have _actually_ deleted all the projectid records from all tables but we don't have the foreign keys set up (MyISAM didn't support them at the time they were created) and it couldn't delete the projectid-named table either.

bpfoley: So... if `deletion` can never be seen, I guess I should take it out of the SQL query and the openapi YAML?

cpeel: Yeah, good call.

bpfoley: OK. Done.

srjfoo (approved):

api/dp-openapi.yaml
```diff
@@ -1128,6 +1157,33 @@ components:
...
+          type: string
+          format: dateTime
+        event_type:
+          type: string
+          enum: [creation, transition, archive]
+          description:
+            <dl>
+            <dt>creation</dt> <dd>Project created</dd>
+            <dt>transition</dt> <dd>Project moves between two stages, eg P3.proj_waiting to P3.proj_avail</dd>
+            <dt>deletion</dt> <dd>Project content has been deleted</dd>
```
cpeel, line 1177: I think we want to delete this line?

bpfoley: Argh. Indeed we do. Thanks!

cpeel (approved):

PR #1044, opened 2024-02-04
faq: Remove incorrect argument to output_header for builtin FAQs

bpfoley:
Because this was in the place of the positional show_stats arg it caused the sidebar to appear.

This was missed in the refactor a2d59202ae483 (by me!)

It was never noticed because on PGDP these FAQs redirect to the wiki.


srjfoo (approved):

cpeel (approved):
Ha! I totally assumed this was my fault when you mentioned it on Slack!

PR #1049, opened 2024-02-05
my_suggestions.php: get_desired_language() doesn't take an arg

bpfoley:
Even back in the dawn of time (ie 786723b9c96), it didn't.

Found by PHPStan

### DIFF
--- a/tools/proofers/my_suggestions.php
+++ b/tools/proofers/my_suggestions.php
@@ -602,7 +602,7 @@ function get_user_suggestion_criteria($username, $flush_cache = false)
     // proofread in the language the UI is in and fall back to English
     if (!$user_languages) {
         $user_languages = [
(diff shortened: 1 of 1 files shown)


tools/proofers/my_suggestions.php
```diff
@@ -602,7 +602,7 @@ function get_user_suggestion_criteria($username, $flush_cache = false)
     // proofread in the language the UI is in and fall back to English
     if (!$user_languages) {
         $user_languages = [
-            lang_name(short_lang_code(get_desired_language($username))) => 5,
+            lang_name(short_lang_code(get_desired_language())) => 5,
```
cpeel, line 605: This is my bad -- I used the wrong function.
```suggestion
            lang_name(short_lang_code(get_user_language($username))) => 5,
```

PR #1053, opened 2024-02-05
DPDatabase.inc: Make (de)serialization funcs protected too

bpfoley:
...to remove another way of creating instances of the class

And fix the return type of __sleep (should be a list of strings) to quell PHPStan.


bpfoley:
Not very sure about how this should be done. I know nothing about PHP serialization.

PR #1054, opened 2024-02-05
CharSuites.__set: Has a void return type so shouldn't return null

bpfoley:
TIL that PHP is unlike Python (in this way, as well as so many others :) ). Since 7.1 it has an explicit void return type, which differs from null.

You can either let a function implicitly return, or explicitly use `return;`

https://wiki.php.net/rfc/void_return_type

### DIFF
--- a/pinc/CharSuites.inc
+++ b/pinc/CharSuites.inc
@@ -72,7 +72,7 @@ class CharSuite
             $this->_pickerset = $value;
             $this->_pickerset->name = $this->name;
         } else {
-            // If we don't recognize the property, raise a notice and return null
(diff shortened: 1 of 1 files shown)


pinc/CharSuites.inc
```diff
@@ -80,7 +80,6 @@ class CharSuite
                 ' on line ' . $trace[0]['line'],
                 E_USER_NOTICE
             );
-            return null;
```
cpeel, line 83: We need to edit the comment in line 75, probably just remove `and return null`

bpfoley: Good spot.

PR #1058, opened 2024-02-05
avoid creating project transitions unless needed

70ray:
Whenever `ProjectTranstion.inc` was included it made all the project transition data whether it was needed or not. (`LPage` includes `ProjectTranstion.inc` because it is needed for `markAsBad()` and `v1.inc` includes `LPage.inc` indirectly.)
This removes the global `PROJECT_TRANSITIONS` and makes them once if they are needed.

Best viewed with hide whitespace.
The interleaved functions have been moved to the end.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/avoid_proj_trans

### DIFF
--- a/pinc/ProjectTransition.inc
+++ b/pinc/ProjectTransition.inc
@@ -12,9 +12,6 @@ include_once($relPath.'User.inc');
 
 define('PT_AUTO', '[AUTO]');
 
-global $PROJECT_TRANSITIONS;
-$PROJECT_TRANSITIONS = [];
-
 class ProjectTransition
 {
     /**
@@ -61,9 +58,6 @@ class ProjectTransition
                 project_states_text($this->to_state)
             );
         }
-
-        global $PROJECT_TRANSITIONS;
-        $PROJECT_TRANSITIONS[] = & $this;
     }
 
     public function is_valid_for($project, $who)
@@ -418,8 +412,6 @@ function project_pre_release_check($project, $round)
  */
 function get_valid_transitions($project, $username)
 {
-    global $PROJECT_TRANSITIONS;
-
     // Keep a cache of viable transitions for a given "from" state
     // so we don't have to iterate over every transition for every project/user
     // combo. We must still do the check for the specific project and user, but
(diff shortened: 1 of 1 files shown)


pinc/ProjectTransition.inc
```diff
@@ -462,120 +453,474 @@ function get_transition($from_state, $to_state)
...
-    [
-        'project_restriction' => null,
-        'action_name' => '[default]',
-        'confirmation_question' => null,
-        'detour' => null,
-        'settings_template' => '',
-        'collateral_actions' => null,
-    ]
-);
+function create_all_transitions()
```
cpeel, line 456: I like the idea that we don't create them unless we actually need them. And also that we're storing the set when we create them so we don't create them more than once.

Given the PR with the new Activities, Rounds, etc classes, should we (very lightly) mirror that here? Something like:
```php
class ProjectTransitions
    private static $_project_transitions = [];

    public static get_all()
    {
        if (!self::$_project_transitions) {
            self::$_project_transitions = create_project_transitions();
        }
        return self::$_project_transitions;
    }
}
```
And have this function just return the array of transitions and the use of `$PROJECT_TRANSITIONS` becomes `ProjectTransitions::get_all()`.

70ray: That makes it seem more sensible, rather than having a function called `create` that only creates if they don't already exist. I'll make it so.

pinc/ProjectTransition.inc
```diff
@@ -587,178 +942,257 @@ for ($rn = 1; $rn <= MAX_NUM_PAGE_EDITING_ROUNDS; $rn++) {
...
-                        'detour' => null,
-                        'settings_template' => "modifieddate='<TIMESTAMP>', nameofwork=CONCAT(nameofwork,' {{$round->id} skipped}'), n_available_pages=0, checkedoutby=IF(checkedoutby!='',checkedoutby,username)",
-                        'collateral_actions' => 'round_to_PP_out_collateral',
-                    ]
-                );
-            }
-        }
-    }
+    return $project_transitions;
 }
```
cpeel, line 1197: Can we please add an "`<hr>`" here and a comment that the functions below this are used in the creation of ProjectTransition objects for either `project_restriction` or `collateral_actions` callbacks? eg:
```

// The functions below are used as callbacks for ProjectTransitions, either
// as the project_restriction or the collateral_actions.

```

cpeel (commented):
Please spin up a sandbox for sanity checking. This LGTM code-wise.

cpeel (approved):
Confirmed that `$PROJECT_TRANSITIONS` is not used elsewhere in the codebase or `noncvs`.

cpeel:
I pushed a small commit to fix an incoming php-cs-fixer error that would have come up post-merge due to https://github.com/DistributedProofreaders/dproofreaders/pull/1069. Will merge after successful CI.

PR #1066, opened 2024-02-06
In queue_defns table, replace release_criterion with 2 'targets'

jmdyck:
In practice (at pgdp.net, anyhow), values in the `release_criterion` column generally have one of the forms:
- `projects < M`
- `pages < N`
- `projects < M or pages < N`

So replace `release_criterion` with `projects_target` and `pages_target`, which just store the relevant `M` and `N`.

(For a form like `projects < M` that doesn't mention pages, we can use 0 for `pages_target`, because `projects < M or pages < 0` is equivalent to `projects < M`, because `pages` is never negative, so `pages < 0` is always false.)


bpfoley (approved):

srjfoo (approved):

SETUP/upgrade/20/20240205_alter_queue_defns.php
```diff
@@ -0,0 +1,106 @@
...
+ *   An expression-style release_criterion,
+ *   typically a condition involving 'pages' and/or 'projects'.
+ *   (e.g., "pages < 60 or projects < 3")
+ *
+ * @return array
+ *   An array that gives the targets for
+ *   (possibly a subset of) 'pages' and 'projects'.
+ *   (e.g., ['pages' => '60', 'projects' => '3'])
+ */
+function convert($rc, $queue_ident)
```
cpeel, line 70: Let's `trim()` the `$rc` to sidestep some errant spacing at the beginning and end. That's the cause of the 3 conversion failures on TEST at least:
```
Error: For queue 371 (round_id='P2', name='All'), unable to convert condition 'pages < 31000 '
Error: For queue 510 (round_id='P3', name='English'), unable to convert condition '(pages < 25000) '
Error: For queue 523 (round_id='P3', name='All'), unable to convert condition 'pages < 31000 '
```

pinc/release_queue.inc
```diff
@@ -242,7 +243,8 @@ function _queue_data($row, $length = null, $unheld_length = null)
         "unheld_length" => (int)$unheld_length,
         "comment" => $row["comment"],
         "project_selector" => $row["project_selector"],
-        "release_criterion" => $row["release_criterion"],
+        "projects_target" => $row["projects_target"],
```
cpeel, line 246: We have to cast these as an int here (eg `unheld_length` above), otherwise they will be returned as a string in the JSON.

stats/release_queue.php
```diff
@@ -109,8 +109,9 @@ function _show_round_queues($round, $listing_view_mode)
         echo "<th>", html_safe(_("Current length")), "</th>\n";
         echo "<th>", html_safe(_("Length without holds")), "</th>\n";
         if (user_can_see_queue_settings()) {
-            echo "<th>", html_safe(_("Release criterion")), "</th>\n";
-            $columns += 1;
+            echo "<th>", html_safe(_("Projects target")), "</th>\n";
```
cpeel, line 112: We might consider formatting these as expressions for display (here and on the queue detail page) rather than just displaying the numbers in columns. I find `projects < 30 or pages < 200` more intuitive. The formatting function could do special things for some of our edgecases like projects < $MAGIC_NUMBER.

tools/project_manager/autorelease.inc
```diff
@@ -126,27 +126,26 @@ function autorelease_for_round($round)
...
         // Release available projects of a given type (defined by $qd->project_selector)
-        // as long as the $qd->release_criterion is true for projects of that type.
+        // as long as the condition implied by the targets is true for projects of that type.
 
         echo "\n";
         echo "Considering {$qd->name} projects...\n";
         echo "\n";
         echo "    For projects matching: {$qd->cooked_project_selector}\n";
-        echo "    release if: {$qd->release_criterion}\n";
+        echo "    release if: projects < {$qd->projects_target} or pages < {$qd->pages_target}\n";
```
cpeel, line 145: Could re-use the "convert queue criteria to display" here too.

SETUP/upgrade/20/20240205_alter_queue_defns.php
```diff
@@ -0,0 +1,106 @@
...
+$sql = "
+    SELECT id, round_id, name, release_criterion
+    FROM queue_defns
+    ORDER BY id
+";
+$result = mysqli_query(DPDatabase::get_connection(), $sql) or die(mysqli_error(DPDatabase::get_connection()));
+$n_rows = mysqli_num_rows($result);
+echo "Converting the release_criterion in {$n_rows} rows ...\n";
+
+$n_conversion_errors = 0;
```
cpeel, line 26: I don't want to over-engineer the conversion script, but putting on my "roll this out to PROD" hat it would be nice to know about the errors before the database alterations happened. That way I'm notified about them, can go fix those, and then re-run the script rather than go update the pages/projects column for the failures and then drop the column.

cpeel (commented):
Thanks for the code fix -- glad to get this `eval()` removed.

stats/release_queue.php
```diff
@@ -150,7 +151,8 @@ function _show_round_queues($round, $listing_view_mode)
         echo "<td>$length</td>\n";
         echo "<td>$unheld_length</td>\n";
         if (user_can_see_queue_settings()) {
-            echo "<td>", html_safe($queue_data["release_criterion"]), "</td>\n";
+            echo "<td>", $queue_data["projects_target"], "</td>\n";
+            echo "<td>", $queue_data["pages_target"], "</td>\n";
         }
```
bpfoley, line 156: As an alternative to Casey's "display them as a formatted expression", displaying the thresholds with  ≥ and aligning them, might make them read more clearly to someone who hasn't looked at the docs.
You could do something like
`echo "<td class='right-align'> &ge;", number_format($queue-data[...]), "</td>";`

SETUP/upgrade/20/20240205_alter_queue_defns.php
```diff
@@ -0,0 +1,106 @@
...
+    if (preg_match('/^(1 *= *0|FALSE)$/', $rc)) {
+        // The release_criterion is always False.
+        // That is, the queue never releases.
+        return [];
+
+    } elseif ($rc == "TRUE") {
+        // The release_criterion is always True.
+        // That is, the queue always releases.
+        // Set a ridiculously high limit for projects.
+        return ['projects' => '1000'];
```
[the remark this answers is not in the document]
jmdyck: Done.

SETUP/upgrade/20/20240205_alter_queue_defns.php
jmdyck: Odd, I know I had a `trim` there originally. Anyway, done.

pinc/release_queue.inc
jmdyck: Yup, should have noticed that.

stats/release_queue.php
jmdyck: Done. It occurs to me that we could just add
```
"release_criterion" => format_queue_targets_as_condition(...)
```
in `_queue_data()`, and then `release_queue.php` wouldn't have to change from the status quo at all. It depends whether you want that 'computed field' to appear everywhere that calls `_queue_data`.

tools/project_manager/autorelease.inc
jmdyck: Done.

SETUP/upgrade/20/20240205_alter_queue_defns.php
jmdyck: Done.

cpeel (approved):
Excellent, thanks Michael!

Sandbox updated with these changes to view the UI bits.

srjfoo (approved):

jmdyck:
Deploying this should be basically invisible to users.

E.g., if you compare:
- https://www.pgdp.org/c/stats/release_queue.php?show=all&round_id=P1
- https://www.pgdp.org/~cpeel/c.branch/split_release_criterion/stats/release_queue.php?show=all&round_id=P1

the differences are things like:
- `1=0` vs `FALSE`
- `(pages < 6000)` vs `pages < 6000` (dropping parens)
- `pages < 120 or projects < 4` vs `projects < 4 or pages < 120` (swapping the order)

PR #1071, opened 2024-02-08
Phpstan cleanups

bpfoley:
Opening this as WIP for review, and there's a sandbox at https://www.pgdp.org/~bfoley/c.branch/phpstan-cleanups.

This eliminates a bunch of the warnings at PHPStan level 4, and there's currently 17 more I haven't got around to yet.

There are different reasons for many of the changes and I've tried to group related changes into separate patches in the patchset, so it might be worth reviewing them one at a time. You might disagree with some of my reasons, and we can let the uncontroversial ones through, and nuke/rework the others.

### DIFF
--- a/crontab/import_pg_catalog.php
+++ b/crontab/import_pg_catalog.php
@@ -223,7 +223,7 @@ foreach (scandir($local_catalog_dir) as $filename) {
             }
             $display_format = $sub_type;
         } else {
-            $display_format = @$display_mapping[$mime_type];
+            $display_format = $display_mapping[$mime_type] ?? null;
             if (empty($display_format)) {
                 @$mime_types_not_in_display_mapping[$mime_type] += 1;
                 $display_format = $mime_type;
--- a/pinc/Round.inc
+++ b/pinc/Round.inc
@@ -546,7 +546,7 @@ function round_project_listing_presort($round)
             ) DESC
         ";
     } else {
-        assert(false);
+        throw new UnexpectedValueException("Unexpected round $round->id");
     }
 }
 
--- a/pinc/abort.inc
+++ b/pinc/abort.inc
@@ -113,17 +113,7 @@ function dump_request_info()
     // give them a last chance to save the page text.
 
     foreach (['_GET', '_POST'] as $array_name) {
-        switch ($array_name) {
-            case '_GET':
-                $array = $_GET;
-                break;
-            case '_POST':
-                $array = $_POST;
-                break;
-            default:
-                $array = [];
-                assert(false);
-        }
+        $array = ${$array_name}; // PHP is a dynamic language! :)
         if ($array) {
             echo "$array_name\n";
             echo html_safe(json_encode($array, JSON_PRETTY_PRINT));
--- a/pinc/misc.inc
+++ b/pinc/misc.inc
@@ -1314,8 +1314,6 @@ function factor_strings($strings)
     $middles = [];
 
     foreach ($strings as $string) {
-        assert($left_match_length >= 0);
-        assert($right_match_length >= 0);
         assert($left_match_length + $right_match_length <= strlen($string));
 
         if ($left_match_length == strlen($string)) {
--- a/pinc/showavailablebooks.inc
+++ b/pinc/showavailablebooks.inc
@@ -469,7 +469,7 @@ function show_project_listing(
         [$ext_sort_col, $label] = $col_info;
 
         $style_attribute = "";
-        if (@$column_styles[$col_id]) {
+        if ($column_styles[$col_id] ?? false) {
             $style_attribute = "style='{$column_styles[$col_id]}'";
         }
 
--- a/pinc/showavailablebooks.inc
+++ b/pinc/showavailablebooks.inc
@@ -320,7 +320,7 @@ function show_projects_for_pool($pool, $checkedout_or_available)
 
         $columns["days_avail"] = ['Days', _("Days Available")];
     } else {
-        assert(false);
+        throw new UnexpectedValueException("Unexpected checkedout_or_available $checkedout_or_available");
     }
 
     $filtered_project_selector = "$initial_project_selector $projects_filter";
--- a/pinc/wordcheck_engine.inc
+++ b/pinc/wordcheck_engine.inc
@@ -729,17 +729,8 @@ function load_wordcheck_events($projectid, $start_time = 0)
         $words = $resultSet["suggestions"];
         $corrections = $resultSet["corrections"];
 
-        if ($words != "") {
-            $words = explode("\n", $words);
-        } else {
-            $words = [];
-        }
-
-        if ($corrections != "") {
-            $corrections = explode("\n", $corrections);
-        } else {
-            $corrections = [];
-        }
+        $words = _explode_no_empty("\n", $words);
+        $corrections = _explode_no_empty("\n", $corrections);
 
         $eventArray[] = [$time, $round, $page, $proofer, $words, $corrections];
     }
@@ -924,16 +915,24 @@ function load_word_list($path)
 
     // remove any trailing spaces, although there shouldn't be any for
     // lists saved with save_word_list(), and \r
-    $file_contents = preg_replace("/\s*\r*\n/", "\n", $file_contents);
-    $words = explode("\n", $file_contents);
+    $file_contents = rtrim(preg_replace("/\s*\r*\n/", "\n", $file_contents));
 
-    // remove the last "empty" word caused by the last newline
-    if (count($words) && $words[count($words) - 1] == "") {
-        array_pop($words);
-    }
-
-    return $words;
+    return _explode_no_empty("\n", $file_contents);
 }
+
+/**
+  * @param string $sep
+  *     The separator between substrings
+  * @param string $string
+  *     The input string to split
+  *
+  * @return array<string>
+  */
+function _explode_no_empty($sep, $string)
+{
+    return empty($string) ? [] : explode($sep, $string);
+}
+
 
 /**
  * Save a list of words (one per line) to a file.
--- a/pinc/wordcheck_engine.inc
+++ b/pinc/wordcheck_engine.inc
@@ -168,7 +168,7 @@ function _get_bad_words_for_project_text($which_result_values, $text, $projectid
         // Associate each bad word with the most specific level at which it was deemed bad.
         $bad_words_in_text = array_intersect_key($bad_words, $input_words_w_freq);
     } else {
-        assert(false);
+        throw new UnexpectedValueException("Unexpected which_result_values $which_result_values");
     }
 
     return [$bad_words_in_text, $languages, $messages];
@@ -1055,7 +1055,7 @@ function get_site_possible_bad_word_lists()
  *   If blank all files will be returned.
  * @param bool $urlbase
  *   If TRUE files will be turned as a URL,
- *   I FALSE absolute filesystem paths are returned
+ *   If FALSE absolute filesystem paths are returned
  */
 function get_site_word_files($pattern = "", $urlbase = true)
 {
@@ -1103,6 +1103,12 @@ function get_site_word_files($pattern = "", $urlbase = true)
  * This function is equivalent to
  * `array_count_values(get_all_words_in_text($text))`
  * but (when appropriate) avoids reifying the intermediate array.
+ *
+ * @param string|array<string> $text
+ *    The text or texts to histogram
+ *
+ * @return array<string,int>
+ *    A histogram of word frequencies in $text
  */
 function get_distinct_words_in_text($text)
 {
@@ -1155,7 +1161,7 @@ function get_distinct_words_in_text($text)
             }
         }
     } else {
-        assert(false);
+        throw new UnexpectedValueException("Unexpected text type " . get_type($text));
     }
 
     return $input_words_w_freq;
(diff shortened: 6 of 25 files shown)


pinc/wordcheck_engine.inc
```diff
@@ -707,6 +707,19 @@ function count_wordcheck_suggestion_events($projectid, $start_time = 0)
...
 
+/**
+  * @ param string sep
+  *     The separator between substrings
+  * @ param string $string
+  *     The input string to split
+  *
+  * @ return Array<string>
+  */
+function _explode_no_empty($sep, $string)
```
70ray, line 718: According to the DP Code best practices, the function declaration would be after the use.

pinc/ProjectTransition.inc
```diff
@@ -713,7 +713,7 @@ function create_project_transitions()
             );
         }
 
-        if (1) {
+        if (1) { /** @phpstan-ignore-line */
```
cpeel, line 716: Is PHPStan ok with just `{` here without the if? I don't love that either, but I might like it better for the `if (true)` use-case.

pinc/Round.inc
```diff
@@ -546,7 +546,7 @@ function round_project_listing_presort($round)
             ) DESC
         ";
     } else {
-        assert(false);
+        assert(false); /** @phpstan-ignore-line */
```
cpeel, line 549: I think we should rethink the original approach to using asserts like this that are intended to detect developer errors vs user errors. For cases like this where it's a param passed in with a known set, throwing `UnexpectedValueException` might be a better approach.

pinc/abort.inc
```diff
@@ -122,7 +122,7 @@ function dump_request_info()
                 break;
             default:
```
cpeel, line 123: In cases like this, I think we should remove the assert entirely.

pinc/forum_interface_phpbb3.inc
```diff
@@ -61,7 +61,7 @@ function phpbb_lang($locale = false)
         $locale = get_desired_language();
     }
 
-    if (PHPBB_VERSION == 3) {
+    if (PHPBB_VERSION == 3) { /** @phpstan-ignore-line */
```
cpeel, line 64: For this file we should just rip out this check. Probably in a separate PR as it'll be space-noisy.

pinc/page_tally.inc
```diff
@@ -21,7 +21,7 @@ include_once($relPath.'User.inc');
 function get_page_tally_names()
 {
     $page_tally_names = [];
-    if (true) {
+    if (true) { /** @phpstan-ignore-line */
```
cpeel, line 24: I think this needs to get pulled out into `pgdp-production`, but I need to look at the other `R*` code.

pinc/page_tally.inc
```diff
@@ -380,7 +380,7 @@ function select_from_site_past_tallies_and_goals(
     $s = preg_replace('/{year_month}/', $ym_expr, $s);
     $s = preg_replace('/{year}/', $y_expr, $s);
 
-    if (0) {
+    if (0) { /** @phpstan-ignore-line */
```
cpeel, line 383: We should just remove this block.

pinc/project_trans.inc
```diff
@@ -22,7 +22,7 @@ function project_transition($projectid, $new_state, $who, $extras = [])
     $transition_str = "$projectid: $current_state -> $new_state";
     if ($testing) {
         $indent = '    ';
-        if (false) {
+        if (false) { /** @phpstan-ignore-line */
```
cpeel, line 25: I'd remove this block (and the commented-out echo on line 35).

pinc/showavailablebooks.inc
```diff
@@ -320,7 +320,7 @@ function show_projects_for_pool($pool, $checkedout_or_available)
 
         $columns["days_avail"] = ['Days', _("Days Available")];
     } else {
-        assert(false);
+        assert(false); /** @phpstan-ignore-line */
```
cpeel, line 323: Probably `UnexpectedValueException` here.

pinc/wordcheck_engine.inc
```diff
@@ -707,6 +707,19 @@ function count_wordcheck_suggestion_events($projectid, $start_time = 0)
     return $row["numevents"];
 }
 
+/**
+  * @ param string sep
```
cpeel, line 711: No space between `@` and `param`.

cpeel: Ditto `@ return` below.

phpstan.neon
```diff
@@ -0,0 +1,60 @@
...
+# Easy fixes
+# https://github.com/phpstan/phpstan/milestone/8
+#
+parameters:
+    fileExtensions:
+        - php
+        - inc
+    paths:
+        - .
+    excludePaths:
```
cpeel, line 30: Probably `pinc/3rdparty` too.

pinc/ProjectTransition.inc
bpfoley: Yup. Done.

pinc/wordcheck_engine.inc
bpfoley: Done.

pinc/Round.inc
bpfoley: OK. I've sprinkled a bunch of `UnexpectedValueException`s and `LogicException`s.

pinc/abort.inc
bpfoley: Or simplify it radically. You can do variable lookup by string with `$array = ${$array_name};`

pinc/page_tally.inc
bpfoley: Done.

pinc/project_trans.inc
bpfoley: Done.

pinc/showavailablebooks.inc
bpfoley: Done.

pinc/wordcheck_engine.inc
bpfoley: Done.

phpstan.neon
bpfoley: I'm 50/50 on this one. On the one hand, adding it to the exclude path means we don't have to fix 3rd party code, which is good. On the other hand, it means we don't _see_ it either, and get a handful of errors in our interfacing code like
`Class BBDiffFormatter extends unknown class DiffFormatter`

cpeel:
Lots of good stuff here! If we can, let's pull out these commits into a separate PR as they seem like no-brainers that we want without question (or much thought on my part):
```
Quell PHPStan warnings about implict string -> int conversions. 
add_files.php: Fix non-nullable PHPStan warning 
release_queue.inc: PHPDocs fix 
copy_pages.php: Remove redundant checks to quell PHPStan 
project.php: Replace redundant check with else arm. 
Fix some cases of break isn't required after a return 
manage_image_sources.php: Remove always-true allow_unknown fields. 
ppv_report.php: Remove check not required after PHP 5.1 
page_table.inc: Hardcode previously conditional Users column 
SortUtility.inc: Remove 15 year old(!) bug comment and no-op code. 
project.php: Quell undefined variable warning for check_pages_table_e… 
graph_data.inc: Report correct variable on error 
```
(I know they aren't in order, sorry, juggling GH UI)

I will look at the others in closer detail.

bpfoley:
OK. Done in #1074. I'll remove those patches from this patchset and re-push.

phpstan.neon
```diff
@@ -0,0 +1,60 @@
...
+    bootstrapFiles:
+        - pinc/bootstrap.inc
+
+    # https://phpstan.org/user-guide/rule-levels
+    level: 4
+
+    # Sadly, these don't seem to work at the moment
+    dynamicConstantNames:
+        - DEBUG
+        - PHPBB_VERSION
```
cpeel, line 42: Think we can drop this line now.

cpeel (commented):
I'm requesting one more carve-out, please. Let's split the following PHPStan-centric commits into a separate PR (could be they move out, could be the others move, either way)
```
Add phpstan to lock files
DONOTCOMMIT (yet): Proof of concept phpstan.neon.
Quell PHPStan with annotations until we figure out a good way of asse… 
misc.inc: Remove some type asserts that are ensured by PHPStan and PH…
Add PHPStan annotations for bug not yet fixed in PHPStan 1.10 
Annotate deliberate always true/false checks to quell PHPStan warnings 
```
To the PR containing ^^ commits let's add a CI/CD pipeline that runs and validates PHPStan (or add it to an existing script that runs in CI/CD, I haven't looked recently to see how those are configured).

The PR with changes to the validation functions in`misc.inc` are my biggest question right now and I need to think about those some more.

pinc/archiving.inc
```diff
@@ -229,7 +229,7 @@ function move_project_rows_to_archive_db($projectid, $table_name, $indent, $dry_
     }
 
     $t_elapsed = $watch->read();
-    if (0) {
+    if (0) { /** @phpstan-ignore-line */
```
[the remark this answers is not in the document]
bpfoley: Done.

phpstan.neon
bpfoley: Done.

bpfoley:
> I'm requesting one more carve-out, please. Let's split the following PHPStan-centric commits into a separate PR (could be they move out, could be the others move, either way)
> 
> ```
> Add phpstan to lock files
> DONOTCOMMIT (yet): Proof of concept phpstan.neon.
> Quell PHPStan with annotations until we figure out a good way of asse… 
> misc.inc: Remove some type asserts that are ensured by PHPStan and PH…
> Add PHPStan annotations for bug not yet fixed in PHPStan 1.10 
> Annotate deliberate always true/false checks to quell PHPStan warnings 
> ```
> 
> To the PR containing ^^ commits let's add a CI/CD pipeline that runs and validates PHPStan (or add it to an existing script that runs in CI/CD, I haven't looked recently to see how those are configured).
> 
> The PR with changes to the validation functions in`misc.inc` are my biggest question right now and I need to think about those some more.

That's fair. I'll move the PHPStan addition to a separate request, and figure out the github actions to call it. Apart from getting the remaining patches resolved, there are a few more issues we'll need to work out before we can enable it for real though.

pinc/misc.inc
```diff
@@ -130,7 +130,6 @@ function get_enumerated_param($arr, $key, $default, $choices, $allownull = false
     {
         // sanity checks on the args
         assert(is_array($arr));
-        assert(!is_array($key));
```
cpeel, line 133: I've landed on only being comfortable removing these checks from our param validation functions if we add type declarations to the function calls such that if an invalid type is passed in (almost always developer error) we will get a very loud exception rather than a silent failure. In fact, if we add PHP type declarations to these we can probably remove most of these assert checks.

If we do that, we should do it in a separate PR to all `get_*_param()` functions (including `get_round_param()` in `Round.inc`) and probably want a sandbox to sanity check. We could also add unit tests to these in `SETUP/tests/misc_ParamValidatorTest.php`.

crontab/import_pg_catalog.php
```diff
@@ -67,7 +67,7 @@
 
 $trace = true;
 
-if ($trace) {
+if ($trace) { /** @phpstan-ignore-line */
```
cpeel, line 70: Let's move the commits with the special phpstan comments to the upcoming PR with the phpstan CI/CD integration, I'd rather not sprinkle the codebase with them until we have the checks in place to use them.

pinc/Activity.inc
```diff
@@ -145,7 +145,7 @@ class Activity
         $this->access_change_callback = $access_change_callback;
 
         global $testing;
-        if ($testing && !is_null($this->access_minima)) {
```
cpeel, line 148: Feels like we shouldn't remove this check until we update the Activity constructor with type declarations so that we can guarantee `access_minima` is an array. Because while it happens to be with today's code, it doesn't mean that it will always be and I don't want to rely on phpstan alone.

pinc/misc.inc
bpfoley: Requiring type declarations makes sense. I assume you mean either explicitly typed PHPDocs `@param`s and `@return`s (which PHPStan already uses for type inference), or PHP typehints on method/function signatures (which PHPStan can also read).

This particular function already has `@params` which is how PHPStan was able to tell us some of the asserts were redundant. Of course for ahead of time static type checking to be an effective guard, we need to use it for all the callers, and actually pay attention to the warnings.

crontab/import_pg_catalog.php
bpfoley: OK. Will do.

pinc/misc.inc
cpeel: I mean [PHP type declarations](https://www.php.net/manual/en/language.types.declarations.php) because PHP will raise an exception if what passes in doesn't match the signature.

cpeel: I can work on getting those done if you haven't started down that path already.

bpfoley: OK. I haven't started yet. Whichever of us gets to it first! :)

cpeel: See https://github.com/DistributedProofreaders/dproofreaders/pull/1079

pinc/Activity.inc
bpfoley: OK

cpeel (commented):
Perfect, thank you very much for working through all the iterations of this with me.

PR #1072, opened 2024-02-08
forum_interface_phpbb3.inc: Remove PHPBB_VERSION checks

bpfoley:
We only support version 3, and the always-true triggers PHPStan warnings.

Diff with 'git diff --ignore-all-space' or github's hide whitespace.


cpeel (approved):
LGTM after whitespace cleanup to satisfy php-cs-fixer

PR #1078, opened 2024-02-09
SETUP: Make PHP/JSON linting run in parallel

bpfoley:
Our github action runners have 2 CPUs so this will speed up the linting step a bit.

### DIFF
--- a/SETUP/lint_json_files.sh
+++ b/SETUP/lint_json_files.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
+set -euo pipefail
 
-if [ "_$1" != "_" ]; then
+if [ -n "${1-}" ]; then
     BASE_DIR=$1
 elif [ -d SETUP ]; then
     BASE_DIR=.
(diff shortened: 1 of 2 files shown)


SETUP/lint_php_files.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -euo pipefail
```
cpeel, line 2: This needs to be after line 13 or `./lint_php_files.sh` fails because `$1` isn't set.

SETUP/lint_php_files.sh
```diff
@@ -11,26 +12,32 @@ else
...
+    if ! echo $OUTPUT | grep -Eq 'ASCII|UTF-8'; then
         echo "ERROR: Unexpected encoding in $file"
         echo "$OUTPUT"
         exit 1
     fi
+}
+
+echo "Checking all .php and .inc files under $BASE_DIR for linting errors..."
+N=$(nproc)
+for file in $(find $BASE_DIR -name "*.php" -o -name "*.inc" | grep -v "^$BASE_DIR/vendor"); do
```
cpeel, line 36: `^$BASE_DIR/vendor` doesn't work unless `$BASE_DIR` is passed in fully qualified, which isn't always the case.

SETUP/lint_json_files.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -euo pipefail
```
cpeel, line 2: Needs to be after line 13 or `$1` fails.

cpeel (commented):
These scripts should succeed being run like:
```bash
# from checkout base
./SETUP/lint_php_files.sh
cd SETUP
./lint_php_files.sh
./lint_php_files.sh /path/to/repo. # this is most useful to run against noncvs too
```

SETUP/lint_php_files.sh
bpfoley: Or we can use BASHisms for handling possibly unbound variables. Done!

SETUP/lint_php_files.sh
bpfoley: Ah. OK. I'll undo this because it's an unnecessary (and wrong) tidyup.

bpfoley:
> These scripts should succeed being run like:
> 
> ```shell
> # from checkout base
> ./SETUP/lint_php_files.sh
> cd SETUP
> ./lint_php_files.sh
> ./lint_php_files.sh /path/to/repo. # this is most useful to run against noncvs too
> ```

Thanks! Tested against all these cases.

cpeel (approved):
Thanks for the speedup!

PR #1079, opened 2024-02-11
Add type declarations for get_*_param functions

cpeel:
Improve the robustness of our `get_*_param()` function calls by adding type definitions rather than assert checking where we can. If the caller passes in the wrong type, PHP will raise an exception which most likely means a developer coded something wrong. Change all non-type validation asserts to raise exceptions. This introduces a slew of new unit tests as well as updating some exiting tests to check the exception message to make sure we're testing the right thing.

Because union types and the `mixed` type didn't enter PHP until 8.0 we are limited in a couple of cases and some comments have been added accordingly.

Code inspection and unit tests should be sufficient but for a sanity check, here's a sandbox: https://www.pgdp.org/~cpeel/c.branch/type-decls-in-get-params/

### DIFF
--- a/SETUP/tests/misc_ParamValidatorTest.php
+++ b/SETUP/tests/misc_ParamValidatorTest.php
@@ -7,11 +7,21 @@ class ParamValidatorTest extends PHPUnit\Framework\TestCase
         "i10" => "10",
         "f10" => "10.0",
         "s10" => "ten",
+        "date" => "2024-02-10",
     ];
     private $ENUM_CHOICES = ["a", "b"];
 
     //------------------------------------------------------------------------
     // get_enumerated_param() tests
+
+    public function testEnumInvalidArray()
+    {
+        // sanity check PHP's built-in type checking, we're not going
+        // to do this for every param and every function
+        $this->expectException(TypeError::class);
+        $this->expectExceptionMessage("must be of the type array, null given");
+        get_enumerated_param(null, 'c', null, $this->ENUM_CHOICES);
+    }
 
     public function testEnum()
     {
@@ -30,8 +40,9 @@ class ParamValidatorTest extends PHPUnit\Framework\TestCase
     public function testEnumNoDefault()
     {
         $this->expectException(InvalidArgumentException::class);
+        $this->expectExceptionMessage("is required");
         $default = null;
-        $result = get_enumerated_param($this->GET, 'none', $default, $this->ENUM_CHOICES);
+        get_enumerated_param($this->GET, 'none', $default, $this->ENUM_CHOICES);
     }
 
     public function testEnumNull()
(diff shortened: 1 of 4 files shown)


pinc/misc.inc
```diff
@@ -298,41 +307,56 @@ function get_float_param($arr, $key, $default, $min, $max, $allownull = false)
...
-        assert(!is_array($key));
-
-        if ($type == 'integer') {
-            assert(is_null($default) || is_int($default));
-        } else {
-            assert(is_null($default) || is_float($default));
+    // sanity checks on the args
+    if ($type == 'integer') {
+        if (! (is_null($default) || is_int($default))) {
+            throw new TypeError("\$default must be of the type int or null");
```
srjfoo, line 322: One last `int`.

srjfoo (commented):
Not sure I'm testing properly -- I tried mucking with the round IDs, but did not manage to raise any of the newly added errors. All of them defaulted to the message at line 500 in `Round.inc`. Not sure how to properly test the rest.

pinc/misc.inc
cpeel: This string is meant to replicate the form that PHP generates when it enforces function type definition.

SETUP/tests/misc_ParamValidatorTest.php
```diff
@@ -44,9 +55,71 @@ public function testEnumNull()
...
+    //------------------------------------------------------------------------
+    // get_numeric_param() tests
+
+    // The testInteger* and testFloat* functions test most of this already,
+    // so just handle the interesting edgecases
+
+    public function testNumericType()
+    {
+        $this->expectException(InvalidArgumentException::class);
+        $this->expectExceptionMessage("must be 'integer' or 'float'");
```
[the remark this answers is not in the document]
cpeel: No, because this is a separate parameter that must be one of those two strings.

cpeel (commented):
> Not sure I'm testing properly -- I tried mucking with the round IDs, but did not manage to raise any of the newly added errors.

That's a good sign! All of these errors should only be seen if a developer is calling the function wrong. None of these should be visible to the end-user.

pinc/Round.inc
```diff
@@ -565,19 +565,18 @@ function is_formatting_round($round)
  *
  * @param array $arr
  * @param string $key
- * @param mixed $default
+ * @param Round|null $default
  * @param bool $allownull
  *
+ * @return Round|null
+ *
```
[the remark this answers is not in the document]
bpfoley: To follow up to myself. I see there _is_ in fact one other you might want PHPDocs: when PHPDocs type constraints can express something that type hints can't. The main examples of this are `array<string>`, which can also be written as `string[]`; and `array<k,v>` for associative arrays. None of these are supported by PHP at all yet. Another significant example is union types (eg `string|int`) which were added in PHP 8.0.

pinc/misc.inc
```diff
@@ -123,25 +123,30 @@ function get_changed_fields_for_objects($new, $old, $remove_non_public = true)
  * @param array $choices
  * @param bool $allownull
```
bpfoley, line 124: PHPStan understands generics, so we could say that the return type has to be the same as the choices and the optional default with something like
```
@template T 
@param array $arr   # The array can be (and probably is) heterogeneous. Don't constraint the key/value types.
@param string $key
@param T|null $default
@param array<T> $choices
@param bool $allownull
@return T
```
In practice, I'm sure `T` will be `string` most of the time.

See https://phpstan.org/writing-php-code/phpdoc-types and https://phpstan.org/blog/generics-by-examples

pinc/misc.inc
```diff
@@ -123,25 +123,30 @@ function get_changed_fields_for_objects($new, $old, $remove_non_public = true)
...
-        assert(is_array($arr));
-        assert(!is_array($key));
+    // TODO: On move to PHP 8.0 as a minimum version, add 'mixed' type def to
+    // $default and a return of mixed
 
-        assert(is_null($default) || is_string($default));
-        assert(is_array($choices));
-        assert(count($choices) > 0);
+    // sanity checks on the args
+    if (count($choices) <= 0) {
```
bpfoley, line 136: `if (empty($choices))`?

pinc/misc.inc
```diff
@@ -251,11 +256,13 @@ function get_enumerated_param($arr, $key, $default, $choices, $allownull = false
  * @param int|null $max
  * @param bool $allownull
  *
+ * @return string|null
```
bpfoley, line 259: Should be `int|null` (to match with the `?int` below)

pinc/misc.inc
bpfoley, line 322: If you wanted to make this marginally more efficient, you could pull the is_null($default) to the top level and say...
```
if (!is_null($default)) {
   if ($type == 'integer') {
     if (!is_int($default)) { throw ... }
   } elseif ($type == 'float') {
    if  (!is_float($default)) { throw ... }
   } else {
    throw "invalid type..."
   }
}
```

cpeel: That would only validate `$type` in the case that `$default` is not null though.

pinc/misc.inc
cpeel: Good catch.

pinc/Round.inc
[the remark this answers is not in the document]
cpeel: Probably not? My instinct is to leave them in for now until we're on PHP 8.0 (for the union types and mixed) and we have a bit more critical mass of type defs in the functions.

pinc/misc.inc
bpfoley: Oh. Yeah, true!

pinc/Round.inc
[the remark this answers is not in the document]
bpfoley: Sounds reasonable. PHPStan has warnings at the higher warning levels about un-annotated functions etc. I've disabled those warnings for now, but this will give us a path to gradually add them (on a per file basis) and to not backslide.

pinc/misc.inc
bpfoley: Nits: (I can't figure out how to add suggestions to lines that are outside the diff), but lines 105 & 106 talk about 'string'. This should be 'value'.

pinc/misc.inc
bpfoley, line 1: Line 153: We're requiring the value to be a string, when, it could legitimately be an int. I think this check should become something like
`if (is_string($s) { $s = trim($s); }`
And of course we leave the later conversion of strings etc to int which will still work.

The same observation applies to line 363 in `get_numeric_param`.

bpfoley: Line 478: We're trimming whitespaces from the value retrieved by `get_param_matching_regex`. Do we actually want this, or was it copy and pasted from the numeric functions?

pinc/misc.inc
cpeel: Good catch, fixed in incoming commit.

pinc/misc.inc
cpeel: Fixed. I also added int-specific testcases for `get_enumerated_param()`

bpfoley (approved):

PR #1082, opened 2024-02-12
Allow control bar buttons to change size when font size changes

70ray:
This removes the style settings for these buttons.

### DIFF
--- a/styles/page_interfaces.less
+++ b/styles/page_interfaces.less
@@ -508,9 +508,6 @@
     button, input[type=submit], input[type=button] {
         box-shadow: 1px 1px grey;
         border-radius: 3px;
-        width: 25px;
-        height: 25px;
-        padding: 0;
     }
 
     input[type=checkbox] {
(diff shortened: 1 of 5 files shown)


chrismiceli (approved):

srjfoo (approved):
Looks good in all of horizontal and vertical, standard and advanced.

cpeel (changes requested):
The buttons are no longer square though, which I think is a regression:
Screen Shot 2024-02-12 at 9 16 23 PM
If we specify the dimensions in `em`s I think they will scale with the font size.

70ray:
> The buttons are no longer square though, which I think is a regression: ... If we specify the dimensions in `em`s I think they will scale with the font size.

Brilliant idea,
I'll do that.

70ray:
Without padding:0 the icons are offset in Chrome with normal font size. The checkbox now also resizes. With a large font, the horizontal/vertical split icon is still small. I will try replacing it with a font-awesome button in another PR.

cpeel (approved):
This works perfectly for me!

srjfoo (approved):

PR #1084, opened 2024-02-13
Add phpstan

bpfoley:


pinc/phpstan_exceptions.inc
```diff
@@ -0,0 +1,8 @@
+<?php
```
cpeel, line 1: Would phpstan let us put this in `SETUP/` to put it with the rest of the dev-centric stuff? Not sure if it'll let us since we're excluding that path.

Either way I'd rename it `phpstan_bootstrap.inc`.

Also please put a newline between the function defs.

cpeel (commented):
Minor question about name/location of bootstrap file, otherwise LGTM.

pinc/phpstan_exceptions.inc
bpfoley: Good idea. Done.

cpeel (approved):

PR #1086, opened 2024-02-13
bootstrap.inc: Allow DPDatabase::connect to be skipped

bpfoley:
So we can run PHPStan without needing to set up mysql.

### DIFF
--- a/pinc/bootstrap.inc
+++ b/pinc/bootstrap.inc
@@ -43,13 +43,15 @@ if (!headers_sent()) {
 include_once($relPath.'gettext_setup.inc');
 include_once($relPath.'DPDatabase.inc');
 
-try {
-    DPDatabase::connect();
-} catch (Exception $e) {
-    // If we're in maintenance mode, don't die here - we'll more gracefully
-    // error out later
-    if (!$maintenance) {
-        throw $e;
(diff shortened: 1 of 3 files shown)


pinc/bootstrap.inc
```diff
@@ -43,13 +43,15 @@ if (!headers_sent()) {
...
 include_once($relPath.'DPDatabase.inc');
 
-try {
-    DPDatabase::connect();
-} catch (Exception $e) {
-    // If we're in maintenance mode, don't die here - we'll more gracefully
-    // error out later
-    if (!$maintenance) {
-        throw $e;
+if (!defined('DISABLE_DATABASE_CONNECT_CHECK')) {
```
cpeel, line 46: This isn't actually just a check it's the sole connection to the DB.
```suggestion
if (!defined('SKIP_DB_CONNECT')) {
```

cpeel (commented):
I noodled over various other ways to do this and I can't come up with a better one so lets roll with it.

pinc/bootstrap.inc
bpfoley: Noted!

cpeel (approved):

PR #1089, opened 2024-02-15
api/v1/queues/:id/projects: Add project_manager field to output

bpfoley:
This is already visible on stats/release_queue.php and in api/v1/projects/:pid, and it's useful (read: I have a use for :)), so make it visible in the queue projects API too.

### DIFF
--- a/api/v1_queues.inc
+++ b/api/v1_queues.inc
@@ -85,6 +85,7 @@ function api_v1_queue_projects($method, $data, $query_params)
         "projectid",
         "title",
         "author",
+        "project_manager",
         "languages",
         "genre",
         "difficulty",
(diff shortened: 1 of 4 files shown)


api/v1_queues.inc
```diff
@@ -82,28 +82,34 @@ function api_v1_queue_projects($method, $data, $query_params)
     $unheld_only = (bool) get_flag_value($query_params, "unheld_only");
 
     $valid_return_fields = [
```
cpeel, line 84: We should either have `fetch_queue_projects_data()` return fields that match the database columns (probably) and generalize them here for the API, or return generalized forms used in the API, but not a mix of the two.

bpfoley: OK. Since `fetch_queue_projects_data` is currently renaming some of the not-consistent-with-project-table fields, we may as well do the username -> project_manager rename there too.

cpeel (approved):

PR #1092, opened 2024-02-18
api/v1/project/:id/pagedetails: Show username if permitted

bpfoley:
...using the same permissions checks that page_detail.php uses for each page, which is now factored out into a shared function.

### DIFF
--- a/api/dp-openapi.yaml
+++ b/api/dp-openapi.yaml
@@ -1058,6 +1058,8 @@ components:
           type: string
           enum: [markdown, html]
         clearance:
+          description: This field will not be included if the requestor
+            is not authorized to see it
           type: string
         custom_characters:
           type: string
@@ -1158,6 +1160,10 @@ components:
         last_modified:
           type: string
           format: dateTime
+        username:
+          description: This field will not be included if the requestor
+            is not authorized to see it
+          type: string
 
     project_transition:
       required:
@@ -1197,6 +1203,10 @@ components:
         text:
           type: string
         state:
+          type: string
+        username:
+          description: This field will not be included if the requestor
+            is not authorized to see it
           type: string
 
     release_queue_detail:
--- a/api/v1_projects.inc
+++ b/api/v1_projects.inc
@@ -428,8 +428,14 @@ function api_v1_project_pagedetails($method, $data, $query_params)
         }
     }
 
+    $project = $data[":projectid"];
+    $rounds_to_display = get_rounds_to_display($project);
+    if (!is_null($only_rounds)) {
+        $rounds_to_display = array_intersect($rounds_to_display, $only_rounds);
+    }
+
     $return_data = [];
-    foreach (fetch_page_table_data($data[":projectid"], null, null, $only_rounds) as $image) {
+    foreach (fetch_page_table_data($project, null, null, $only_rounds) as $image) {
         $page_rounds_data = [];
         // Remove proofer names and adjust timestamp format
         foreach ($image["pagerounds"] as $round_id => $round_data) {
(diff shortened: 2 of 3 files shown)


api/dp-openapi.yaml
```diff
@@ -1158,6 +1158,8 @@ components:
         last_modified:
           type: string
           format: dateTime
+        username:
```
cpeel, line 1161: Let's add a description that is something like "this field may not be included if the requestor is not authorized to see it".

Bonus: add a similar description for the `clearance` field in the `project` schema which I think is the only other place we change the schema based on privs.

api/v1_projects.inc
```diff
@@ -438,7 +444,9 @@ function api_v1_project_pagedetails($method, $data, $query_params)
                 $round_data["last_modified"] = date(DATE_ATOM, $round_data["modified_timestamp"]);
                 unset($round_data["modified_timestamp"]);
             }
-            unset($round_data["username"]);
+            if (!can_user_see_usernames_for_page($project, $image, $rounds_to_display)) {
```
cpeel, line 447: If we're solving this here, we should make the solution work for `api_v1_project_page_round()` too, eg:
```php
    // We can't show the username here unless the user has proofed the page
    // in one round or they are an PF/SA. We need to abstract the conditional
    // logic out of tools/project_manager/page_detail.inc
    unset($row["user"]);
```

pinc/page_table.inc
```diff
@@ -548,6 +531,39 @@ function changeSelection(sel) {
     }
 }
 
+/**
+ * Can the currently logged in user see the usernames of users who have
+ * worked on a particular page in a project?
+ *
+ *
+ * @param Project $project
+ * @param array $page_res
```
cpeel, line 540: I don't love that this take the page data schema from `fetch_page_table_data()` as it seems very limited to its use with that function (see also the comment about needing this same auth data for `api_v1_project_page_round()`. Generalizing it so that it works for `api_v1_project_page_round()` too will probably fix this though.

pinc/page_table.inc
```diff
@@ -548,6 +531,39 @@ function changeSelection(sel) {
...
+ *   The page data returned by `fetch_page_table_data`
+ * @param Round[] $rounds_to_display
+ *   Which rounds to check for user work.
+ *
+ * @return bool
+ */
+function can_user_see_usernames_for_page(Project $project, array $page_res, array $rounds_to_display): bool
+{
+    global $pguser;
```
cpeel, line 549: `User::current_username()` is preferable. Death to (most) globals.

api/dp-openapi.yaml
bpfoley: Done.

pinc/page_table.inc
bpfoley: Huzzah! :)

Done.

api/v1_projects.inc
bpfoley: Done.

pinc/page_table.inc
bpfoley: Yeah, I've changed it to work with a project and list of usernames.

cpeel (approved):
Code LGTM but can haz sandbox please?

bpfoley:
> Code LGTM but can haz sandbox please?

https://www.pgdp.org/~bfoley/c.branch/page-deets-users/

cpeel:
Thanks, confirmed both API endpoints only show the user the usernames they are allowed to see.

srjfoo (approved):

PR #1093, opened 2024-02-19
Remove use of MAX_NUM_PAGE_EDITING_ROUNDS

cpeel:
Replace numeric iteration of the rounds by a more natural foreach over the objects. This ultimately removes `MAX_NUM_PAGE_EDITING_ROUNDS`.

Testable in https://www.pgdp.org/~cpeel/c.branch/remove-max-rounds-const/

Aside: This started as work on Task 2000 but it got into a bit of yak shaving.

### DIFF
--- a/tools/project_manager/automodify.php
+++ b/tools/project_manager/automodify.php
@@ -252,11 +252,11 @@ while ([$projectid] = mysqli_fetch_row($allprojects)) {
     if ($state == $round->project_complete_state) {
         // The project is ready to exit this round.
 
-        if ($round->round_number < MAX_NUM_PAGE_EDITING_ROUNDS) {
+        if ($round->round_number < Rounds::get_last()->round_number) {
             // It goes to the next round.
             $next_round = get_Round_for_round_number(1 + $round->round_number);
             $new_state = $next_round->project_waiting_state;
-        } elseif ($round->round_number == MAX_NUM_PAGE_EDITING_ROUNDS) {
+        } elseif ($round->round_number == Rounds::get_last()->round_number) {
             // It goes into post-processing.
             if (is_null(project_get_auto_PPer($projectid))) {
                 $new_state = PROJ_POST_FIRST_AVAILABLE;
@@ -264,7 +264,7 @@ while ([$projectid] = mysqli_fetch_row($allprojects)) {
                 $new_state = PROJ_POST_FIRST_CHECKED_OUT;
             }
         } else {
-            die("round_number is {$round->round_number}???\n");
+            throw new ValueError("round_number is {$round->round_number}???");
         }
 
         if ($verbose) {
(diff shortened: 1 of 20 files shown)


tools/project_manager/automodify.php
```diff
@@ -252,19 +252,19 @@ function ensure_project_blurb($project)
     if ($state == $round->project_complete_state) {
         // The project is ready to exit this round.
 
-        if ($round->round_number < MAX_NUM_PAGE_EDITING_ROUNDS) {
+        if ($round->round_number < Rounds::get_last()->round_number) {
             // It goes to the next round.
             $next_round = get_Round_for_round_number(1 + $round->round_number);
             $new_state = $next_round->project_waiting_state;
-        } elseif ($round->round_number == MAX_NUM_PAGE_EDITING_ROUNDS) {
+        } elseif ($round->round_number == Rounds::get_last()->round_number) {
```
bpfoley, line 259: This could now be shortened to `($round == Rounds::get_last())`

bpfoley (approved):
Definitely nicer to be able to do it this way.

tools/project_manager/automodify.php
cpeel: We could, but I'd like to leave this as-is so it's parallel with the `<` comparison in the first `if()` block. Unfortunately it's not possible to make custom comparison functions in PHP objects like you can do in Python which would let us do `$round1 < $round2`.

bpfoley: A third option, although in this case it's probably more obscure, is to switch on the spaceship operator
```
switch ($round->round_number <=> Rounds::get_last()->round_number) {
 case -1: // increment to next round
 case 0:  // go to postprocessing
 case 1: // should be impossible
}
```

tools/project_manager/show_project_wordcheck_stats.php
```diff
@@ -26,9 +26,7 @@
 // load bad words for the site for this project
 $site_bad_words = load_site_bad_words_given_project($projectid);
 
-// get the latest possible round
-$last_possible_round = get_Round_for_round_number(MAX_NUM_PAGE_EDITING_ROUNDS);
-$pages_res = page_info_query($projectid, $last_possible_round->id, 'LE');
+$pages_res = page_info_query($projectid, Rounds::get_last()->id, 'LE');
```
srjfoo, line 29: I'm getting 

```Notice: Undefined variable: last_possible_round in /home/cpeel/public_html/c.branch/remove-max-rounds-const/tools/project_manager/show_project_wordcheck_stats.php on line 62

Notice: Trying to get property 'id' of non-object in /home/cpeel/public_html/c.branch/remove-max-rounds-const/tools/project_manager/show_project_wordcheck_stats.php on line 62
'' is not a valid round-id 
```
(but don't know how to quote that line specifically -- guess I need some instruction!)

srjfoo (commented):
Everything else looks good.

tools/project_manager/show_project_wordcheck_stats.php
cpeel: Whoops, I missed the second use of that variable on the page. Fixed!

srjfoo (approved):

PR #1094, opened 2024-02-19
use a span to hold controls instead of giving them a class

70ray:
This is "Yak-shaving" getting ready to replace the horizontal/vertical splitter icon  with a font-awesome icon button so it will change size with the font size.
The 'control' class which the icon was given is only used to enable it to be selected and removed later in page_browse.js. This branch makes page_browse.js handle adding and removing controls entirely within itself.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/control_span

### DIFF
--- a/scripts/page_browse.js
+++ b/scripts/page_browse.js
@@ -4,7 +4,7 @@
 function makePageControl(pages, selectedImageFileName, changePage) {
     // changePage is a callback to act when page changes
     const pageSelector = document.createElement("select");
-    const controls = $("<span>", {class: "nowrap control"}).append(proofIntData.strings.page + " ", pageSelector);
+    const controls = $("<span>", {class: "nowrap"}).append(proofIntData.strings.page + " ", pageSelector);
 
     if(!selectedImageFileName) {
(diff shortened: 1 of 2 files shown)


cpeel (approved):

srjfoo (approved):

scripts/page_browse.js
```diff
@@ -4,7 +4,7 @@
 function makePageControl(pages, selectedImageFileName, changePage) {
     // changePage is a callback to act when page changes
     const pageSelector = document.createElement("select");
-    const controls = $("<span>", {class: "nowrap control"}).append(proofIntData.strings.page + " ", pageSelector);
```
chrismiceli, line 7: we don't need to remove or adjust css any, right?

chrismiceli (approved):

scripts/page_browse.js
cpeel: It doesn't looks like we don't style anything with this class, just use it for DOM manipulation, so no CSS changes needed.

PR #1095, opened 2024-02-19
Remove duplicate definition of rounds in project_states.inc

70ray:
This is the first part of the process, using a ProjectState object to hold the $medium_label, $long_label, $forum and $phase and a global $project_states array to associate the state names with the project states.

This has been remade with conficts resolved.

The next part will be to use a static class to remove globals.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/project_states_3

### DIFF
--- a/pinc/Pool.inc
+++ b/pinc/Pool.inc
@@ -40,8 +40,6 @@ class Pool extends Stage
         $access_change_callback,
         $description,
         $document,
-        $project_checkedout_state,
-        $project_available_state,
         $foo_Header,
         $foo_field_name,
         $blather
@@ -58,16 +56,9 @@ class Pool extends Stage
             "tools/pool.php?pool_id=$id"
         );
 
-        $this->project_checkedout_state = $project_checkedout_state;
-        $this->project_available_state = $project_available_state;
         $this->foo_Header = $foo_Header;
         $this->foo_field_name = $foo_field_name;
         $this->blather = $blather;
-
-        $this->states = [
-            $this->project_checkedout_state,
-            $this->project_available_state,
-        ];
     }
 }
 
--- a/pinc/Pool.inc
+++ b/pinc/Pool.inc
@@ -55,8 +55,9 @@ class Pool extends Stage
         );
 
         // The following project-state related attributes are created & set in project_states.inc
-        // $this->project_checkedout_state
-        // $this->project_available_state
+        //     $this->project_checkedout_state
+        //     $this->project_available_state
+        //     $this->states
 
         $this->foo_Header = $foo_Header;
         $this->foo_field_name = $foo_field_name;
--- a/pinc/Pool.inc
+++ b/pinc/Pool.inc
@@ -20,8 +20,6 @@ class Pool extends Stage
      *
      * A container for various constants relating to a particular pool.
      *
-     * @param string $project_checkedout_state
-     * @param string $project_available_state
      * @param $foo_Header
      * @param $foo_field_name
      *   The relevant person to display in pool listings, both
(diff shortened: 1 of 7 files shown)


pinc/Round.inc
```diff
@@ -115,20 +115,6 @@ class Round extends Stage
         $this->honorifics = $honorifics;
         krsort($this->honorifics);
 
-        $this->project_unavailable_state = constant("PROJ_{$round_id}_UNAVAILABLE");
```
cpeel, line 118: Let's add a comment that project state attributes are defined later so that someone who sees an attribute being used on a Round object and looks here has a breadcrumb trail. eg:
```php
// The following project-state related attributes are created & set in project_states.inc
//     $this->project_unavailable_state
//     $this->project_waiting_state
//     $this->project_bad_state
//     $this->project_available_state
//     $this->project_complete_state
//     $this->project_states
```

pinc/Pool.inc
```diff
@@ -58,16 +54,9 @@ class Pool extends Stage
             "tools/pool.php?pool_id=$id"
         );
 
-        $this->project_checkedout_state = $project_checkedout_state;
-        $this->project_available_state = $project_available_state;
         $this->foo_Header = $foo_Header;
         $this->foo_field_name = $foo_field_name;
         $this->blather = $blather;
-
-        $this->states = [
```
cpeel, line 67: Ditto the review made in `Round.inc` about a breadcrumb comment.

pinc/project_states.inc
```diff
@@ -43,247 +49,245 @@ function declare_project_state(
...
     global $project_states_for_star_metal_;
 
     define($constant_name, $constant_value);
-    $PROJECT_STATES_IN_ORDER[] = $constant_value;
-    $project_state_medium_label_[$constant_value] = $medium_label;
-    $project_state_long_label_[$constant_value] = $long_label;
-    $project_state_forum_[$constant_value] = $forum;
-    $project_state_phase_[$constant_value] = $phase;
+
+    $project_states[$constant_value] = new ProjectState(
```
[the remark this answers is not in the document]
70ray: constant_name is the string defining the constant so it resolves to the same thing as constant_value. constant_value is used as the key for the array of objects so there doesn't seem to be a use for the same thing inside the object.

We could put star_metal in the object and postpone constructing the sql condition until we need it. I'll try doing that.

pinc/Round.inc
70ray: Done this.

pinc/Pool.inc
70ray: Done this also.

pinc/project_states.inc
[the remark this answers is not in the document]
70ray: > We could put star_metal in the object and postpone constructing the sql condition until we need it. I'll try doing that.

This would be possible but would involve changing some other files so for now just move the sql constructor into a function that uses the objects.

pinc/Pool.inc
```diff
@@ -58,16 +54,13 @@ class Pool extends Stage
             "tools/pool.php?pool_id=$id"
         );
 
-        $this->project_checkedout_state = $project_checkedout_state;
-        $this->project_available_state = $project_available_state;
+        // The following project-state related attributes are created & set in project_states.inc
+        // $this->project_checkedout_state
+        // $this->project_available_state
```
cpeel, line 59:
```suggestion
        //     $this->project_checkedout_state
        //     $this->project_available_state
        //     $this->states
```

cpeel (approved):
One minor comment and this LGTM.

pinc/Pool.inc
70ray: Done this.

70ray:
Sandbox updated.

srjfoo (approved):

PR #1096, opened 2024-02-19
Add `any` and `all` functions

bpfoley:


SETUP/tests/misc_ParamValidatorTest.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-class ParamValidatorTest extends PHPUnit\Framework\TestCase
+class misc_ParamValidatorTest extends PHPUnit\Framework\TestCase
```
cpeel, line 3: The original intent behind the `misc_` convention was to indicate what files the tested functionality was in. But we aren't really doing that, so I'd much rather just rename the files to not have the `misc_` than to rename the class to include it.

Because I assume your goal is to have the class names match the filename?

bpfoley: That's right. phpunit doesn't seem happy if you don't.

bpfoley: Done.

bpfoley:
> Code looks good but it would be nice to have a sandbox to sanity check the non-test changes.

https://www.pgdp.org/~bfoley/c.branch/any-all/

srjfoo (approved):

PR #1097, opened 2024-02-20
Improve external catalog search

bpfoley:


tools/project_manager/external_catalog_search.php
```diff
@@ -169,11 +172,15 @@ function do_search_and_show_hits()
...
         }
 
-        //if it's not a book don't display it.  we might want to uncomment in the future if there are too many records being returned - if (substr(yaz_record($id, $start, "raw"), 6, 1) != "a") { $start++; continue; }
         /** @phpstan-ignore-next-line */
         $marc_record = new MARCRecord();
         $marc_record->load_yaz_array($rec);
-
+        // If it's not a 'language material' (printed or typed) or
+        // 'manuscript language material' (written), don't display it.
+        // if (!in_array($marc_record->get_type_of_record(), ["Language material", "Manuscript language material"])) {
```
cpeel, line 180: Why the commented code?

tools/project_manager/external_catalog_search.php
```diff
@@ -108,11 +108,14 @@ function do_search_and_show_hits()
     $id = yaz_connect($external_catalog_locator, ["charset" => "UTF-8"]);
     yaz_syntax($id, "usmarc");
     yaz_element($id, "F");
-    yaz_search($id, "rpn", trim($fullquery));
+    yaz_search($id, "rpn", trim(str_replace("\n", " ", $fullquery)));
     $extra_options = ["timeout" => 60];
     yaz_wait($extra_options);
     $errorMsg = yaz_error($id);
 
+    echo "<b>Raw query:</b>";
```
cpeel, line 116: Leftover debugging?

tools/project_manager/external_catalog_search.php
```diff
@@ -200,6 +208,8 @@ function do_search_and_show_hits()
             echo   "<td class='left-align top-align'>$value</td>";
             echo "</tr>\n";
         }
+        // For adding test cases
```
cpeel, line 211: Leftover debugging?

cpeel (commented):
We should get a sandbox spun up for this as well I think.

bpfoley:
> We should get a sandbox spun up for this as well I think.

Yes. It's at https://www.pgdp.org/~bfoley/c.branch/task1900-better-yaz/

tools/project_manager/external_catalog_search.php
bpfoley: Yeah, deliberately, since it's slightly non-obvious. But I can remove it.

tools/project_manager/external_catalog_search.php
bpfoley: Feature request from Sharon in task 1900.

tools/project_manager/external_catalog_search.php
bpfoley: I was initially trying to figure out how to express 'language material only' in the query to the LoC server which would avoid the need to filter the response,  but I was getting weird and inconsistent results.

The original was commented, and I was undecided if I wanted to enable it or not.

Then I thought I might use this logic to decorate each table entry with an HTML class so we can selectively show/hide them in the browser with some javascript.

So... I'm undecided what I want to do with this, basically.

tools/project_manager/external_catalog_search.php
srjfoo: Well, more or less, anyway. 😁 

If a PM creating a project has to wade 10 pages of results to find the record they want, it's a useful reminder. Most PMs are probably not going to know what `@attr 1=1003` means, but it's pretty easy to figure out.

I also really like seeing the media type the record is for. One of my tests, the first three or four hits were all either sound recordings or video of some kind.

cpeel: I defer to Sharon as someone who actually uses this. To me this feels like it will generate more confusion than value for the majority of our PMs.

tools/project_manager/external_catalog_search.php
cpeel: Perhaps instead add a comment in `MARCRecordTest.php` about how to generate the encoded data, eg:
```php
// To generate data for these test functions, use:
// base64_encode(serialize($rec)) where $rec is the output of yaz_record()
```

cpeel (approved):

tools/project_manager/external_catalog_search.php
srjfoo: If you want to hold off merging for a bit, I'll ask Rick what he thinks. If he doesn't like the idea, we can ditch it. I don't want to make things harder for PMs.

bpfoley: No rush. Let's figure out what we're all happy with.

One possible alternative would be to hide the raw query in a `<details>` element that's closed by default.

srjfoo (approved):

tools/project_manager/external_catalog_search.php
```diff
@@ -108,16 +194,18 @@ function do_search_and_show_hits()
     $id = yaz_connect($external_catalog_locator, ["charset" => "UTF-8"]);
     yaz_syntax($id, "usmarc");
     yaz_element($id, "F");
-    yaz_search($id, "rpn", trim($fullquery));
+    yaz_search($id, "rpn", trim(str_replace("\n", " ", $fullquery)));
     $extra_options = ["timeout" => 60];
     yaz_wait($extra_options);
     $errorMsg = yaz_error($id);
 
+    echo "<details style='border: 1px solid #aaa; border-radius: 4px; padding: 0.5em 0.5em 0'>\n";
```
cpeel, line 202: Should we move some of this styling into CSS? I'd rather have labeled-snowflakes there than style dusted around the code.

tools/project_manager/external_catalog_search.php
```diff
@@ -108,16 +194,18 @@ function do_search_and_show_hits()
...
+    echo "<summary style='font-weight: bold; margin: -0.5em -0.5em 0; padding: 0.5em;'>Raw query</summary>\n";
+    echo "<pre style='padding-left: 4em'>$fullquery</pre>\n";
+    echo "</details>\n";
+
     if (!empty($errorMsg)) {
-        echo "<p class='error'>";
-        echo _("The following error has occurred:");
-        echo " $errorMsg";
-        echo "</p>";
+        echo "<p class='error'>", _("The following error has occurred:"), " $errorMsg", "</p>";
```
cpeel, line 208: We should `html_safe($errorMsg)` because while it's probably OK we shouldn't risk it as it comes from an external source. I ack that this was in the original.

tools/project_manager/external_catalog_search.php
```diff
@@ -129,186 +217,227 @@ function do_search_and_show_hits()
...
-            echo "<tr>";
-            echo   "<th class='left-align top-align' style='width: 20%;'>$label:</th>";
-            echo   "<td class='left-align top-align'>$value</td>";
-            echo "</tr>\n";
+    // Display results
+    echo "<p>", sprintf(_("%d results returned."), $total_hits), "</p>\n";
+
+    if ($total_hits > 0) {
+        $encoded_fullquery = base64_encode(serialize($fullquery));
+        $url_base = "external_catalog_search.php?action=do_search_and_show_hits&fq=$encoded_fullquery&" . search_query_params();
```
cpeel, line 255: `$url_base` is output in the HTML directly so we either need to `&` -> `&amp;` here, or in the output function -- the latter is probably preferred. Ditto below where we continue constructing it.

tools/project_manager/external_catalog_search.php
```diff
@@ -129,186 +217,227 @@ function do_search_and_show_hits()
...
+                echo "</td>";
+
+                // Subtable for record
+                echo "<td class='left-align top-align' style='width: 45%;'>";
+
+                // Compresss the record so we don't run over URL length limits
+                $rec = base64url_encode(gzencode(serialize($r), 9));
+                echo "<div style='text-align:right; font-size:small'>\n";
+                echo   "<a style='text-decoration: none' target='_blank' href='marc_inspector.php?rec=$rec'>\n";
+                echo     "inspect MARC <i class='fa fa-external-link-alt'></i>\n";
```
cpeel, line 294: re: FA icon - I like this. We should replace our custom "open in a new window" icon with this site-wide. Not in this PR, obvs. Probably should think about doing this for all or our little inline graphics.

tools/project_manager/external_catalog_search.php
```diff
@@ -129,186 +217,227 @@ function do_search_and_show_hits()
...
-    echo "</td>";
-    echo "</tr>\n";
 
-    // -----------------------------------------------------
-
-    echo "</table><p class='center-align'>";
-    if (yaz_hits($id) != 0) {
-        echo "<input type='submit' value='", attr_safe(_("Create the Project")), "'>&nbsp;";
+    foreach ([
+        [_('Refine Search'), 'external_catalog_search.php?action=show_query_form&' . search_query_params()],
```
cpeel, line 323: Ditto raw `&` in final HTML.

tools/project_manager/external_catalog_search.php
```diff
@@ -129,186 +217,227 @@ function do_search_and_show_hits()
...
-    echo "<input type='button' value='$label' onclick='javascript:location.href=\"$url\";'>";
-    echo "&nbsp;";
-
-    $label = attr_safe(_('No Matches'));
-    $url = "editproject.php?action=createnew";
-    echo "<input type='button' value='$label' onclick='javascript:location.href=\"$url\";'>";
-    echo "&nbsp;";
+function link_or_text(string $url_base, string $label, bool $show_link, int $start): string
+{
+    return $show_link ? "<a href='$url_base&start=$start'>" . html_safe($label) . "</a>" : $label;
```
cpeel, line 337: This too, no?
```suggestion
    return $show_link ? "<a href='$url_base&start=$start'>" . html_safe($label) . "</a>" : html_safe($label);
```

tools/project_manager/external_catalog_search.php
```diff
@@ -129,186 +217,227 @@ function do_search_and_show_hits()
...
+    // https://www.loc.gov/z3950/agency/defns/bib1.html
+    // https://www.loc.gov/z3950/agency/bib1.html
+
+    $fullquery = [];
 
     if ($_REQUEST['title']) {
-        $fullquery = $fullquery.' @attr 1=4 "'.$_REQUEST['title'].'"';
-        $attr_set++;
+        $attr = get_enumerated_param($_REQUEST, 'title_attr', null, array_keys($title_attrs));
+        $fullquery[] = '@attr 1=' . $attr . ' "' . $_REQUEST['title'] . '"';
```
cpeel, line 410: tip: using `sprintf()` would probably make this easier to read.

cpeel: Ditto the others in this function.

tools/project_manager/marc_inspector.php
```diff
@@ -0,0 +1,129 @@
...
+$title = _("MARC record inspector");
+output_header($title, NO_STATSBAR);
+
+echo "<h1>$title</h1>\n";
+echo "<a href='https://www.loc.gov/marc/bibliographic/'>MARC 21 Format for Bibliographic Data</a>\n";
+echo "<p>\n";
+
+$loc_bib = "https://www.loc.gov/marc/bibliographic";
+$field_info = [
+    "leader" => ["tooltip" => "Record header",  "url" => "$loc_bib/bdleader.html"],
```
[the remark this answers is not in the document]
bpfoley: I deliberately didn't translate any of this because, with the exception of the leader (which, TBH, I should probably change back) they're taken verbatim from the titles in the MARC 21 spec.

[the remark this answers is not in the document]
cpeel: Makes sense, let's add a comment to that effect please.

tools/project_manager/marc_inspector.php
```diff
@@ -0,0 +1,129 @@
+<?php
+$relPath = '../../pinc/';
+include_once($relPath.'base.inc');
+include_once($relPath.'theme.inc');
+
+require_login();
+
+$rec = @$_REQUEST['rec'];
```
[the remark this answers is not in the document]
bpfoley: Done. I've no idea why I did that...

tools/project_manager/marc_inspector.php
```diff
@@ -0,0 +1,129 @@
...
+foreach ($record as $r) {
+    if (count($r) == 1) {
+        $r[] = "";
+    }
+    [$label, $value] = $r;
+    if (preg_match('/^(\(3,)([^)]*)(\).*)$/', $label, $match)) {
+        $field = $match[2];
+        if (array_key_exists($field, $field_info)) {
+            $info = $field_info[$field];
+            $attr = " style='font-weight: bold' title='" . attr_safe($info["tooltip"]) . "'";
```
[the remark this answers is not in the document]
bpfoley: Nice! Done.

tools/project_manager/external_catalog_search.php
bpfoley: Done, after a smaller adventure in learning about `lessc`.

tools/project_manager/external_catalog_search.php
bpfoley: ACK. Sprinkled `attr_safe` where `$url_base` is used. Also I'd left an accidental (and now unnecessary) `if ($hide_nontext) { $url_base .= "&hide_nontext=on"; }` in the code.

tools/project_manager/external_catalog_search.php
bpfoley: Thanks. Ray pointed me in the right direction. The minor niggle I have about this is I don't like the size. I couldn't figure out how to make it size proportional to the `font-size` of the `div`.

tools/project_manager/external_catalog_search.php
bpfoley: Handled along with the fix above.

tools/project_manager/external_catalog_search.php
bpfoley: Yup! :)

tools/project_manager/external_catalog_search.php
bpfoley: Now that you say it, it probably would! Done.

tools/project_manager/marc_inspector.php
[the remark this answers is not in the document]
bpfoley: Done.

styles/layout.less
```diff
@@ -352,6 +352,30 @@ footer, #footer {
     }
 }
 
+/* ------------------------------------------------------------------------ */
+/* Disclosure boxes with optional info */
+
+details {
+    border: 1px solid #aaa;
```
cpeel, line 359: We should try not to hard-code color in `layout.less` because it may not work for all themes. See if one of the pre-defined `@border-color-base*` values will work for what you want as the themes override these.

bpfoley: ACK. Yay for CSS variables.

cpeel (approved):
LGTM. @srjfoo - any last review/thoughts re CSS before I merge this?

tools/project_manager/marc_inspector.php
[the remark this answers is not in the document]
srjfoo: Agreed. Not everything needs to be translated,

srjfoo (approved):
Changes look good to me. For those who really like going through everything, they can still get the full results, but this allows a lot more flexibility than we had before.

PR #1098, opened 2024-02-20
Simplify dealing with "OCR" round

cpeel:
Instead of doing gymnastics with round numbers when dealing with the "OCR" round, create a pseudoround object for it with the important details for the diff page (and other uses) and use that instead.

Now we can pass round IDs (and "OCR") into the diff page.

Testable in https://www.pgdp.org/~cpeel/c.branch/diff-by-round-id/

Old `diff.php` URLs should work just fine, but new ones will use the round IDs instead of round numbers.

This was further yak shaving while working towards fixing Task 2000.

### DIFF
--- a/pinc/Round.inc
+++ b/pinc/Round.inc
@@ -572,12 +572,13 @@ function is_formatting_round($round)
  * @param string $key
  * @param Round|null $default
  * @param bool $allownull
+ * @param bool $allowOCR
  *
  * @return Round|null
  *
  * @throws InvalidArgumentException
  */
-function get_round_param(array $arr, string $key, ?Round $default = null, bool $allownull = false): ?Round
+function get_round_param(array $arr, string $key, ?Round $default = null, bool $allownull = false, bool $allowOCR = false)
 {
     // sanity checks on the args
     if (!is_null($default) && $allownull) {
@@ -600,6 +601,8 @@ function get_round_param(array $arr, string $key, ?Round $default = null, bool $
         $round = Rounds::get_by_id($s);
         if ($round) {
             return $round;
+        } elseif ($allowOCR && $s === "OCR") {
+            return get_OCR_pseudoround();
         } else {
             throw new InvalidArgumentException(sprintf(
                 _("Parameter '%1\$s' ('%2\$s') is not a valid round ID"),
@@ -619,4 +622,14 @@ function get_round_param(array $arr, string $key, ?Round $default = null, bool $
             return $default;
         }
     }
+}
+
+function get_OCR_pseudoround()
+{
+    $round = new stdClass();
+    $round->round_number = 0;
+    $round->id = "OCR";
+    $round->text_column_name = 'master_text';
+    $round->user_column_name = "'none'";  // string literal, not column name
+    return $round;
 }
--- a/tools/project_manager/diff.php
+++ b/tools/project_manager/diff.php
@@ -23,7 +23,6 @@ $bb_diffs = (@$_GET['bb_diffs'] === 'on' and user_can_mentor_in_any_round());
 $project = new Project($projectid);
 $state = $project->state;
 $project_title = $project->nameofwork;
-$navigation_text = "";
 
 if (!$project->pages_table_exists) {
     // This shouldn't normally happen --
@@ -127,7 +126,7 @@ output_header("$title: $project_title", NO_STATSBAR, $extra_args);
 echo "<h1>" . html_safe($project_title) . "</h1>\n";
 echo "<h2>$image_link</h2>\n";
 
-do_navigation(
+$navigation_text = get_navigation(
     $projectid,
     $image,
     $L_round_num,
@@ -200,7 +199,7 @@ if ($L_text != $R_text) {
  * Build up the text for the navigation bit, so we can repeat it
  * again at the bottom of the page
  */
-function do_navigation(
+function get_navigation(
     $projectid,
     $image,
     $L_round_num,
@@ -213,7 +212,7 @@ function do_navigation(
     $only_nonempty_diffs,
     $bb_diffs
 ) {
-    global $navigation_text;
+    $navigation_text = "";
     $jump_to_js = "this.form.image.value=this.form.jumpto[this.form.jumpto.selectedIndex].value; this.form.submit();";
 
     $navigation_text .= "\n<form method='get' action='diff.php'>";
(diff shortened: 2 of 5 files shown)


pinc/Round.inc
```diff
@@ -620,3 +623,13 @@ function get_round_param(array $arr, string $key, ?Round $default = null, bool $
...
     }
 }
+
+function get_OCR_pseudoround()
+{
+    $round = new stdClass();
+    $round->round_number = 0;
+    $round->id = "OCR";
+    $round->text_column_name = 'master_text';
+    $round->user_column_name = "'none'";  // string literal, not column name
```
srjfoo, line 633: I don't know if this applies here, but when we skip rounds or add blanks, the user column is assigned the value `[none]` including the square brackets (in case someone comes along and registers that username?). AFAIK, we don't have a list of reserved usernames, do we? Just acceptable characters.

cpeel: This value was taken from the existing code this replaced, although I'm pretty sure that this string is actually never shown in the Page Details page. My inclination is to keep this as-is.

70ray (commented):
This is not very helpful, but my first thought was that this may not be the best approach. 'Round' was defined as an activity that a user can work in. We are really dealing with a different sort of thing here. Like fence posts and rails. I'll have to think about it some more.

pinc/Round.inc
bpfoley: TIL about `stdClass`.

TBH, I'm a little bit worried that playing dynamic tricks like this is going to cause bugs further down the line where we extend/refactor `Round` and forget to update `get_OCR_pseudoround`. It also makes the return type of `get_round_param()` into a white lie.

Looking at the implementation of `Round` and `Stage`, I'm not sure what the best way to handle this is.

Maybe just the simplest thing is to have a unit test that checks that all the attributes of OCR are also attributes of a real Round object.

tools/project_manager/diff.php
```diff
@@ -13,8 +13,16 @@
...
-$R_round_num = get_integer_param($_GET, 'R_round_num', null, 0, Rounds::get_last()->round_number);
+// older code used round numbers, which we keep for backwards compatibility
+if (isset($_GET['L_round_num'])) {
+    $L_round_num = get_integer_param($_GET, 'L_round_num', null, 0, Rounds::get_last()->round_number);
+    $R_round_num = get_integer_param($_GET, 'R_round_num', null, 0, Rounds::get_last()->round_number);
+    $L_round = Rounds::get_by_number($L_round_num) ?? get_OCR_pseudoround();
+    $R_round = Rounds::get_by_number($R_round_num) ?? get_OCR_pseudoround();
+} else {
+    $L_round = get_round_param($_GET, "L_round", null, false, true);
+    $R_round = get_round_param($_GET, "R_round", null, false, true);
```
bpfoley, line 24: If you wanted to deduplicate the boilerplate, you could do something like
```
$rounds = [];
foreach(["L_round", "R_round"] as $arg) {
    if (isset($_GET["{$arg}_num"])) {
      $num = get_integer_param($_GET, "{$arg}_num", null, 0, Rounds::get_last()->round_number);
      $rounds[] = Rounds::get_by_number($num) ?? get_OCR_pseudoround();
   } else {
      $rounds[] = get_round_param($_GET, $arg, null, false, true);
  }
}
[$L_round, $R_round] = $rounds;
```

pinc/Round.inc
bpfoley: At a minimum could we add a comment to the `@return` of `get_round_param()` to warn about OCR having limited attributes?

cpeel: Yeah, it doesn't feel quite right to munge `get_round_param()` like this. We could instead create `get_round_or_ocr_param()` which basically just wraps the former and handles `OCR`. That would obviate the worry that this would pollute other users and ensure the burden of dealing with the OCR snowflake is on the script using it.

cpeel:
Added a new commit such that we don't abuse `get_round_param()` but allow getting a round-like object for OCR.

Does this seem better?

bpfoley:
> Added a new commit such that we don't abuse `get_round_param()` but allow getting a round-like object for OCR.
> 
> Does this seem better?

Yup. Definitely better.

70ray:
The problem seems to be in passing parameters to a function to compare two rounds (say R1 and R2). If instead the two parameters were to mean compare the text **before** R1 to the text **after** R2 then we would not need any 'virtual'  rounds. E.g. compare(P2, P2) would mean compare the text from before P2 (how P1 left it) to the text how P2 left it. This would help to circumvent the need to define a 'virtual' round for OCR.

cpeel:
> The problem seems to be in passing parameters to a function to compare two rounds (say R1 and R2). If instead the two parameters were to mean compare the text **before** R1 to the text **after** R2 then we would not need any 'virtual' rounds. E.g. compare(P2, P2) would mean compare the text from before P2 (how P1 left it) to the text how P2 left it. This would help to circumvent the need to define a 'virtual' round for OCR.

I see where you're going, but `compare(P2, P2)` without that context is rather nonsensical. That also still requires `compare()` to know about `master_text` for "the thing before P1" explicitly so we're just moving the special casing around.

70ray:
The problem seems to be in passing parameters to a function to compare two rounds (say R1 and R2). If instead the two parameters were to mean compare the text **before** R1 to the text **after** R2 then we would not need any 'virtual'  rounds. E.g. compare(P2, P2) would mean compare the text from before P2 (how P1 left it) to the text how P2 left it. This would help to circumvent the need to define a 'virtual' round for OCR.

> > The problem seems to be in passing parameters to a function to compare two rounds (say R1 and R2). If instead the two parameters were to mean compare the text **before** R1 to the text **after** R2 then we would not need any 'virtual' rounds. E.g. compare(P2, P2) would mean compare the text from before P2 (how P1 left it) to the text how P2 left it. This would help to circumvent the need to define a 'virtual' round for OCR.
> 
> I see where you're going, but `compare(P2, P2)` without that context is rather nonsensical. That also still requires `compare()` to know about `master_text` for "the thing before P1" explicitly so we're just moving the special casing around.

I wasn't suggesting changing what the user sees or introducing any incompatibilities, just throwing out ideas which may or may not be useful.

bpfoley (approved):
Even if we haven't figured out quite what the ideal form is yet, this seems like a strict improvement to me, so let's go with it.

srjfoo (approved):

PR #1100, opened 2024-02-21
New split button

70ray:
Change the split direction button to use a font-awesome icon so it changes size with the font.


70ray:
Here is the sandbox:
Sandbox at: https://www.pgdp.org/~rp31/c.branch/new_split_button

scripts/view_splitter.js
```diff
@@ -35,7 +34,8 @@ var viewSplitter = function(container, storageKey) {
     });
 
     function setSplitControls() {
-        splitImage.src = splitVertical ? proofIntData.buttonImages.imgHSplit : proofIntData.buttonImages.imgVSplit;
+        buttonContent.className = splitVertical ? 'fas fa-grip-lines' : 'fas fa-grip-lines-vertical';
```
cpeel, line 37: I wonder if we could use `fal fa-columns` for the vertical and use CSS to rotate it 90 degrees for the horizontal?

https://fontawesome.com/v5/icons/columns?f=classic&s=light

pinc/page_controls.inc
```diff
@@ -35,15 +35,10 @@ function get_proofreading_interface_data_js()
     ];
 
     global $code_url;
-    $button_images = [
-        "imgVSplit" => "$code_url/tools/proofers/gfx/bt4.png",
```
chrismiceli, line 39: we use these icons in 2 other places I think. Should we update all usages to be consistent throughout the site?

chrismiceli (commented):
Is it possible to get the alignment to be more consistent with the other icons?
image

pinc/page_controls.inc
cpeel: At least one of the other places is the enhanced interface where it's next to a bunch of other of these old button icons. I wouldn't change anything in that interface and just keep what we have.

scripts/view_splitter.js
70ray: That would look better. I'll try it.

70ray:
> Is it possible to get the alignment to be more consistent with the other icons?  ...

I'll see how they are defined.

70ray:
Sandbox updated.

chrismiceli (approved):

scripts/view_splitter.js
cpeel: Bummer the `fal` is in the pro version. Can we try the `fa-columns` and see how it looks in a sandbox? `fa-grip-lines` doesn't say "columns" & "rows" to me and I worry it won't be intuitive to some users.

70ray: Implemented this now.

70ray:
Icon is now 'fas fa-columns' with css rotation.

Sandbox updated.

chrismiceli (approved):

cpeel (approved):
I think this looks great. Curious what @srjfoo thinks of the columns / rotated columns icon.

srjfoo:
> I think this looks great. Curious what @srjfoo thinks of the columns / rotated columns icon.

Apologies for taking so long!

I have really mixed feelings. That doesn't look like columns to me, it looks like an open book (it's the heavy bar at the top). In its normal orientation, it doesn't make me think of "vertical", though it does give the feeling of side-by-side. In the horizontal mode, though, it's awfully narrow. It'll probably jar people and they'll have to get used to it.

I took a look in all of FF, Chrome, Opera and Safari. The first three looked okay as far as balance was concerned, but unfortunatey, this is what it looks like in Safari:
image
It looks about the same in Safari on the iPad. And this is what it looks like in Firefox on the iPad (Dark mode in FF on the iPad is weird, it actually does invert stuff. And sorry for the size difference. The early screenshot is from the computer, this one from the iPad): 
image

Is there something we can do about that, or are we stuck with it?

70ray:
> 
> Is there something we can do about that, or are we stuck with it?

I've put padding left and right to zero. This makes it look more centred in Chrome.

srjfoo:
Okay, that looks good on the iPad in both FF and Safari, and in Safari on the Mac. I still think that people will need to get used to it, but I've gotten used to it already, so it shouldn't take too long. 🙂

srjfoo (approved):

70ray:
I think there could be a better way. Looking into it, I'll change this to draft for now.

70ray:
This now uses an iconify icon as a background image. This has the advantage that the icon will always be accurately centred in the button whereas the font icons positioning varies wit different browsers. There are also many icons to choose from on [their web site](https://iconify.design/).  We may not wish to merge it in its present form but host the icon ourselves.

scripts/view_splitter.js
```diff
@@ -35,7 +32,8 @@ var viewSplitter = function(container, storageKey) {
     });
 
     function setSplitControls() {
-        splitImage.src = splitVertical ? proofIntData.buttonImages.imgHSplit : proofIntData.buttonImages.imgVSplit;
+        splitButton.style.backgroundImage = splitVertical ? `url("https://api.iconify.design/octicon:rows-24.svg")` : `url("https://api.iconify.design/octicon:columns-24.svg")`;
```
chrismiceli, line 35: can we avoid the 3rd party request? Our users are sensitive to these requests (why we got rid of momentjs)

cpeel: Yes, we should create a `pinc/3rdparty/iconify` directory with these two files if we want to use them to avoid the 3rd party API request. They are both licensed with the MIT license so we should include that in the directory and a `README.md` that calls out where we got them and that both files (by name) are covered by the MIT license. That lets us add more later with different (but still acceptable) licenses too.

Oh, and a `details.json` so we can give proper attribution.

cpeel (commented):
These look fine to me.

scripts/view_splitter.js
70ray: Done this.

70ray:
Sandbox updated.

pinc/3rdparty/iconify/octicon/README.md
```diff
@@ -0,0 +1,3 @@
+# Iconify
```
cpeel, line 1: I realize that Iconify has icons with different licenses, but I don't think we need to have a subdir for each class. Consider moving this contents up a level, renaming `LICENSE` `MIT-LICENSE` and the icons `octicon-<name>`.

70ray: I'll do that.

70ray:
Sandbox updated.

cpeel (approved):
This works beautifully. I love that we have a template for how to get access to a very wide range of icons, so this work -- your research and the infra here -- will pay dividends in the future. Thank you!

I'll wait to merge until @srjfoo can give some feedback on the icons but I know worst-case it might mean small changes to which Iconify icons we use.

srjfoo (approved):
Love this version!

I honestly don't think this is a problem, because it fixes itself if you refresh the page, but thought I'd show a screenshot:

image


What I did: I started out the page with the browser default at 32, then changed it live back to 16. When hovering over the buttons after the change, you can see that they change back to 16, but that the rest of the text doesn't. If I refresh the page, everything goes back to the browser defaut font size. In the use case for the page browser, I think that a forced refresh would be fine, but I'm not clear on whether that would work for other use-cases.

pinc/3rdparty/iconify/octicon-details.json
```diff
@@ -0,0 +1,7 @@
+{
```
cpeel, line 1: This file has to be named `details.json` for it to show up in the Credits page. I know this license only applies to the `octicon` ones for now, but let's punt that down the road and just rename the file for now.

70ray: Done this. I should have read the documentation.

srjfoo (approved):

70ray:
> What I did: I started out the page with the browser default at 32, then changed it live back to 16. When hovering over the buttons after the change, you can see that they change back to 16, but that the rest of the text doesn't. If I refresh the page, everything goes back to the browser defaut font size. In the use case for the page browser, I think that a forced refresh would be fine, but I'm not clear on whether that would work for other use-cases.

I've noticed this with Firefox. Chrome etc. seem to automatically refresh the page when the font size is changed so this doesn't happen.

srjfoo:
> > What I did: I started out the page with the browser default at 32, then changed it live back to 16. When hovering over the buttons after the change, you can see that they change back to 16, but that the rest of the text doesn't. If I refresh the page, everything goes back to the browser defaut font size. In the use case for the page browser, I think that a forced refresh would be fine, but I'm not clear on whether that would work for other use-cases.
> 
> I've noticed this with Firefox. Chrome etc. seem to automatically refresh the page when the font size is changed so this doesn't happen.

Unless someone else thinks it's worth worrying about, then, it's probably not. And good reminder to me to check other browsers! 😁

PR #1102, opened 2024-02-24
Use a ProjectStates class to avoid global variables. 

70ray:
The global variables are $project_states which is only used in one place and has been removed.
$PROJECT_STATES_IN_ORDER is used in many places but only those in project_states.inc have been replaced so far (by ProjectStates::get_states();).
There are some functions which can also be replaced later.

Best viewed on here with whitespace hidden although it gets a bit mixed up even then.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/project_states_class

### DIFF
--- a/pinc/project_states.inc
+++ b/pinc/project_states.inc
@@ -44,24 +44,9 @@ class ProjectStates
 {
     private static $states = [];
 
-    private static function declare_project_state(
-        $constant_name,
-        $constant_value,
-        $medium_label,
-        $long_label,
-        $forum,
-        $phase,
-        $star_metal
-    ) {
-        define($constant_name, $constant_value);
-
-        self::$states[$constant_value] = new ProjectState(
-            $medium_label,
-            $long_label,
-            $forum,
-            $phase,
-            $star_metal
-        );
+    public static function add_state($state_name, $state)
+    {
+        self::$states[$state_name] = $state;
     }
 
     public static function get_states()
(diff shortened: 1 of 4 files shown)


pinc/project_states.inc
```diff
@@ -40,262 +40,282 @@ class ProjectState
...
-
-function project_states_text($state)
-{
-    global $project_states;
-    return $project_states[$state]->long_label ?? '';
-}
+    ) {
+        define($constant_name, $constant_value);
+
+        self::$states[$constant_value] = new ProjectState(
```
cpeel, line 58: If we follow the existing Activity/Activities pattern of having the `ProjectState` constructor assign itself to `ProjectStates` we avoid a lot of this code churn while still getting all of the benefits of having the "container" `ProjectStates` class. And while I don't really love that pattern, it is used a fair amount in our code and I'd almost rather keep it for consistency.

70ray: Hmmm...

pinc/project_states.inc
```diff
@@ -40,262 +40,282 @@ class ProjectState
...
+    // is the order in which they will be displayed in various contexts
+    // (via ProjectStates::get_states()).
+
+    public static function declare_project_states($rounds, $pools)
+    {
+        global $waiting_projects_forum_idx, $projects_forum_idx, $pp_projects_forum_idx, $posted_projects_forum_idx,
+        $completed_projects_forum_idx, $deleted_projects_forum_idx;
+
+        // for the initial creation of a project
+        self::declare_project_state(
```
70ray, line 107: It would be clearer to define the constant explicitly and add the state rather than have it all inside a function. Maybe

```php
    define("PROJ_NEW", "project_new");
    ProjectStates::add_state(PROJ_NEW, new ProjectState(
        _("New Project"),
        _("New Project"),
        $waiting_projects_forum_idx,
        'NEW',
        ''
    ));
```
Admitedly this involves a repetition but only of a constant or member variable so the compiler would catch any mistakes.

70ray, line 107: It would be clearer to define the constant explicitly and add the state rather than have it all inside a function. Maybe

```php
    define("PROJ_NEW", "project_new");
    ProjectStates::add_state(PROJ_NEW, new ProjectState(
        _("New Project"),
        _("New Project"),
        $waiting_projects_forum_idx,
        'NEW',
        ''
    ));
```
Admitedly this involves a repetition but only of a constant or member variable so the compiler would catch any mistakes.

70ray:
Sandbox updated incorporating changes suggested by reviewers.

pinc/project_states.inc
```diff
@@ -40,76 +40,63 @@ class ProjectState
...
+    private static $states = [];
 
-function project_states_text($state)
-{
-    global $project_states;
-    return $project_states[$state]->long_label ?? '';
-}
+    public static function add_state($state_name, $state)
+    {
+        self::$states[$state_name] = $state;
```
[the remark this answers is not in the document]
70ray: This will work.

pinc/project_states.inc
```diff
@@ -40,76 +40,63 @@ class ProjectState
...
+        return self::$states;
+    }
 
-// -----------------------------------------------
+    public static function get_medium_label($state)
+    {
+        return self::$states[$state]->medium_label ?? '';
+    }
+
+    public static function text($state)
```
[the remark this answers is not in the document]
70ray: God idea.

70ray:
Sandbox updated.

pinc/project_states.inc
```diff
@@ -26,90 +26,80 @@ include_once($relPath.'site_vars.php');
...
-    $constant_name,
-    $constant_value,
-    $medium_label,
-    $long_label,
-    $forum,
-    $phase,
-    $star_metal
-) {
-    global $project_states;
+    public static function add_state($state)
```
cpeel, line 49: So we pass a reference to the state instead of a copy of it. We do this for Activities too.
```suggestion
    public static function add_state(&$state)
```
This is mostly important for when we pass this back to callers -- everyone is working with the same object.

cpeel (approved):
Love this - thank you for doing this work.

In a follow-up PR (not this one) we should move the `ProjectStates` and `ProjectState` class definitions and the functions at line 290 into a `ProjectState.inc` file so that we follow our desired pattern of separating files that define classes & broader functions and one that creates structure (ala `stages.inc`).

pinc/project_states.inc
70ray: I guess it can't hurt.
Or you could avoid using the add_state() function by making $states public and say
```php
    ProjectStates::$states[PROJ_NEW] = new ProjectState( ...
```
or keep $states private and define everything inside ProjectStates using a public function.

70ray: Done this.

70ray:
Sandbox updated.

srjfoo:
> Sandbox updated.

I just now tried refreshing an open window of the sandbox, and got A Very Large Number of
>  Notice: Only variables should be passed by reference in /home/rp31/public_html/c.branch/project_states_class/pinc/project_states.inc on line ...

At various line numbers

cpeel:
Ugh. Let's just revert that last commit then and call it good.

cpeel:
FYI: I reverted the commit.

cpeel:
@srjfoo - I have a sandbox for this at https://www.pgdp.org/~cpeel/c.branch/project_states_class/

If we can get this in tonight I can get a PR up for a holistic fix to the errors we're seeing regarding ordering.

srjfoo (approved):

PR #1107, opened 2024-02-25
tasks.php: Add permalinks to each tasks comment

bpfoley:
The comment IDs and anchors already existed, we just needed to use them!

Example in sandbox at https://www.pgdp.org/~bfoley/c.branch/task-comment-permalink/tasks.php?action=show&task_id=2288

### DIFF
--- a/tasks.php
+++ b/tasks.php
@@ -1486,11 +1486,14 @@ function TaskComments($tid, $action)
         $can_edit_comment = ($requester_u_id == $row['u_id'] && $now_sse - (int)$row['comment_date'] <= 86400) || user_is_a_sitemanager();
         echo "<div class='task-comment' id='$comment_id'>";
         $comment_username_link = private_message_link_for_uid($row['u_id']);
-        echo "<b>$comment_username_link - " . date("l d M Y @ g:ia", $row['comment_date']) . "</b>";
+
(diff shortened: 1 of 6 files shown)


tasks.php
```diff
@@ -1486,11 +1486,14 @@ function TaskComments($tid, $action)
         $can_edit_comment = ($requester_u_id == $row['u_id'] && $now_sse - (int)$row['comment_date'] <= 86400) || user_is_a_sitemanager();
         echo "<div class='task-comment' id='$comment_id'>";
         $comment_username_link = private_message_link_for_uid($row['u_id']);
+
+        echo "<div class='task-comment-title'>";
+        echo "<a class='perma' href='$tasks_url?action=show&task_id=$tid#$comment_id'><i class='fas fa-link'></i></a>&nbsp;";
```
cpeel, line 1491: I think we should put the link icon at the end of the row (or at least after the date/time) and not in the front. Also can we make it a little smaller? It feels larger than the text.

Love this idea though.

bpfoley: Sounds good. And I debolded the inconsistently bold dash too.

cpeel (approved):

chrismiceli (approved):

srjfoo (approved):

PR #1109, opened 2024-02-25
Allow users to select their preferred project detail level

cpeel:
Give project detail levels meaningful descriptors and allow users to select which one they want to see by default. This also removes all of the bits in the code that override the `detail_level` via URL.

Note that we change the default level back to 2 (it was changed from 2 to 3 in 2014). This present users with a smaller set of data to hopefully be less confusing. Subsequent PRs could further pair down the data shown at level 2 to improve the new user experience.

Testable in https://www.pgdp.org/~cpeel/c.branch/project-detail-level/

[Task 669](https://www.pgdp.net/c/tasks.php?task_id=669)

### DIFF
--- a/tools/proofers/PPage.inc
+++ b/tools/proofers/PPage.inc
@@ -129,8 +129,7 @@ class PPage
             . "id={$this->lpage->projectid}"
             . $amp
             . "expected_state={$this->proj_state}"
-            . $amp
-            . "detail_level=1#project-comments";
+            . "#project-comments";
     }
 
     public function urlencoded($escape_amp = false)
--- a/userprefs.php
+++ b/userprefs.php
@@ -11,6 +11,7 @@ include_once($relPath.'Settings.inc');
 include_once($relPath.'js_newpophelp.inc');
 include_once($relPath.'forum_interface.inc'); // get_forum_user_details(), get_url_to_edit_profile()
 include_once($relPath.'User.inc');
+include_once($relPath.'Project.inc');
 
 require_login();
 
@@ -379,7 +380,26 @@ function echo_general_tab($user)
         'credit_name_adhoc',
         null
     );
-    show_blank();
+
+    foreach (get_project_detail_levels() as $level => $label) {
+        // We don't allow users to select level 1 and 4 because:
+        // * Level 1 doesn't include Start Proofreading and will confuse people
+        // * Level 4 includes the page detail which is resource intensive
+        //   and can produce a Very Large Page (>35MB)
+        if ($level == 1 || $level == 4) {
+            continue;
+        }
+
+        $page_detail_options[$level] = sprintf("%s (%d)", $label, $level);
+    }
+    show_preference(
+        _('Default Project Detail Level'),
+        'project_detail',
+        'project_detail',
+        $userSettings->get_value('project_detail', 1),
+        'dropdown',
+        $page_detail_options
+    );
     echo "</tr>\n";
 
     echo_bottom_button_row();
@@ -422,6 +442,8 @@ function save_general_tab($user)
     }
 
     $userSettings->set_boolean('hide_special_colors', $_POST["show_special_colors"] == 'no');
+
+    $userSettings->set_value('project_detail', $_POST["project_detail"]);
 }
 
 function echo_proofreading_tab($user)
--- a/faq/pophelp/prefs/prefs_pophelp.inc
+++ b/faq/pophelp/prefs/prefs_pophelp.inc
@@ -127,6 +127,12 @@ _("<p>This setting controls who is able to see your statistics.</p>
             'content' => _("<p>This setting determines which site theme is used as your default.</p>"),
         ],
 
+    'set_project_detail' =>
+        [
+            'title' => _("Project Detail"),
+            'content' => _("<p>This setting determines the default detail level shown on a project page.</p>"),
+        ],
+
     // =====================================================================
     // PM Tab Settings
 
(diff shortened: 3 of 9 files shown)


tools/proofers/PPage.inc
```diff
@@ -129,8 +129,7 @@ class PPage
             . "id={$this->lpage->projectid}"
             . $amp
             . "expected_state={$this->proj_state}"
-            . $amp
-            . "detail_level=1#project-comments";
+            . "#project-comments";
```
cpeel, line 132: This is the only change I'm wary about. Having the page at level 1 gives much easier access to the project comments which is the goal here.

srjfoo: Fwiw, one of the complaints was from someone who didn't appreciate being sent to level 1 by clicking on the Project Comments link in the PI. I suspect that some people, anyway, have gotten used to thinking of it as an easy link back to the project page in general.

srjfoo: I see I wasn't quite clear enough. I have, multiple times while testing, clicked on the "Project Comments" link from the project page. I *always* switch away from detail level 1, because what I want is below the Project Comments. We actually put the internal `#project-comments` link in to focus directly on the PCs, because even at detail level 1, you had to scroll down to get to them. You've retained the internal link directly to the PCs, so I think it'll be okay.

srjfoo:
I'm getting

```
Notice: Undefined offset: 0 in /home/cpeel/public_html/c.branch/project-detail-level/project.php on line 604
```
When going to a new project to load pages. But not on something in P1 Unavailable. Compare:

[P1 Unavailable](https://www.pgdp.org/~cpeel/c.branch/project-detail-level/project.php?id=projectID5e0512c3ca72a)
[New Project](https://www.pgdp.org/~cpeel/c.branch/project-detail-level/project.php?id=projectID62151244d8b17)

cpeel:
> When going to a new project to load pages. But not on something in P1 Unavailable.

Good catch - fixed!

srjfoo (approved):

userprefs.php
```diff
@@ -379,7 +380,26 @@ function echo_general_tab($user)
...
+        //   and can produce a Very Large Page (>35MB)
+        if ($level == 1 || $level == 4) {
+            continue;
+        }
+
+        $page_detail_options[$level] = sprintf("%s (%d)", $label, $level);
+    }
+    show_preference(
+        _('Default Project Detail Level'),
+        'project_detail',
```
chrismiceli, line 397: Does it make sense to make the setting name a constant to be reference here and in the project page?

chrismiceli (approved):

70ray (commented):
Would it make sense for the setting selected on the project page to persist rather than having to change it in the settings?

srjfoo:
> Would it make sense for the setting selected on the project page to persist rather than having to change it in the settings?

The point of setting the default in the usersettings is to restrict which can be set as default. I don't think setting level 1 as a default would be that big a deal (though not very helpful), but if the detail level selected on the project page were used as the default, that would allow detail level 4 to be used as a default, which we do want to avoid. Granted, some code tweaking could probably be done to automatically downgrade any level 4 to level 2 or 3 when the most recent level chosen was stored for later retrieval, but I don't know what the trade-offs are.

cpeel:
> Would it make sense for the setting selected on the project page to persist rather than having to change it in the settings?

I think if we were allowing any of the levels to be a default it would, but like Sharon said the fact that we prevent levels 1 and 4 from being used would likely lead to surprises: "I left this page at level 4, why when I come back it's level 3?".

userprefs.php
cpeel: That's a good idea as that would find problems at "compile" time. We do this (poor) pattern all over the place for user prefs. I think I'd like to be consistent with the others for this PR but think moving to constants for all of these settings is a good idea -- we just need to establish where those constants should live.

70ray (approved):

windymilla (approved):
All seems to work well.
For a change like this, ss it worth a temporary news item on proofing pages when it is deployed, to point people to the Default Project Detail Level setting?

srjfoo:
> All seems to work well. For a change like this, is it worth a temporary news item on proofing pages when it is deployed, to point people to the Default Project Detail Level setting?

I don't know, but if we do, I'd point to the announcement in Recent Additions, not to the user prefs.

windymilla:
> > All seems to work well. For a change like this, is it worth a temporary news item on proofing pages when it is deployed, to point people to the Default Project Detail Level setting?
> 
> I don't know, but if we do, I'd point to the announcement in Recent Additions, not to the user prefs.

Yes, that would be best. I just wondered, since it is a change rather than a new feature, it might be something people notice but aren't sure what to do about it. Maybe not a problem though.

cpeel:
> > > All seems to work well. For a change like this, is it worth a temporary news item on proofing pages when it is deployed, to point people to the Default Project Detail Level setting?
> > 
> > 
> > I don't know, but if we do, I'd point to the announcement in Recent Additions, not to the user prefs.
> 
> Yes, that would be best. I just wondered, since it is a change rather than a new feature, it might be something people notice but aren't sure what to do about it. Maybe not a problem though.

I was thinking that beyond the announcement in the forums and perhaps a news item, we could update the preference for all PM, PF, and SAs upon roll-out to be 3 instead of 2. I suspect they are going to be the majority of users who want to see things at that level (although anyone can, of course).

windymilla:
Good idea, @cpeel

PR #1111, opened 2024-02-27
sitemap.php: Fix error caused by PROJECT_STATES_IN_ORDER

bpfoley:
2b97d677d2 caused the initialization of PROJECT_STATES_IN_ORDER to move from project_states.inc to stages.inc (it's done by the call to declare_project_states in stages.inc)

sql_collater_for_project_state() (defined in project_states.inc) requires PROJECT_STATES_IN_ORDER to be defined.

(I'm not confident this is the correct fix, but I thought I'd put it up to advance the discussion)

### DIFF
--- a/sitemap.php
+++ b/sitemap.php
@@ -1,7 +1,7 @@
 <?php
 $relPath = "pinc/";
 include_once($relPath."base.inc");
-include_once($relPath."project_states.inc");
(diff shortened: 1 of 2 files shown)


cpeel (approved):
Thank you for this. Getting `stages.inc` included at a higher level is on my TODO list to avoid this kind of thing since pretty much everything needs the site structure.

PR #1113, opened 2024-02-27
Define site structure once and only once

cpeel:
How the site is structured, stages/rounds/pools, are integral to virtually every page and the order matters. Create a file that
defines the structure (by including the structure-building files) and include that once and only once in `bootstrap.inc`. This avoids the whack-a-mole of trying to get the ordering right in each and every file (eg https://github.com/DistributedProofreaders/dproofreaders/pull/1111).

Furthermore, since every DP page must include `base.inc`, `site_vars.php` is already loaded thanks to bootstrap.inc.

Update the includes checker to validate the best practices for which files to include and not include.

Sandbox at https://www.pgdp.org/~cpeel/c.branch/define-site-structure/


chrismiceli (approved):
Looks fine to me, though surprising to see quizzes.inc is so vital

70ray:
> ... surprising to see quizzes.inc is so vital

I guess it is connected to the code in Activity.inc which checks if the user is qualified to do what he wants to do.

70ray (approved):
As Chris said it is surprizing what a lot of structure is created on every site visit. Often hardly any of it is actually needed but computer time and hardware is so cheap now there may be little value in trying to reduce it.

cpeel:
> > ... surprising to see quizzes.inc is so vital
> 
> I guess it is connected to the code in Activity.inc which checks if the user is qualified to do what he wants to do.

Yes, this is it exactly.

> Often hardly any of it is actually needed but computer time and hardware is so cheap now there may be little value in trying to reduce it.

Sort of. The navbar in the header uses most of it (what the rounds are, who has access to them (which involves the quizzes), etc) and that's on most every page except the proofreading interface. You're right that there's a lot more as part of that which isn't needed on every page, but I agree we're at a point in time where it's not useful to try and optimize it.

bpfoley (approved):

srjfoo (approved):

PR #1114, opened 2024-02-27
Link "skip empty diffs" with "without formatting"

cpeel:
When users are comparing diffs without formatting they expect "skip empty diffs" to skip pages that only differ by formatting. This links the two together so that it behaves the way that expect. [Task 2000](https://www.pgdp.net/c/tasks.php?action=show&task_id=2000).

Sandbox: https://www.pgdp.org/~cpeel/c.branch/navigate-without-formatting/

### DIFF
--- a/tools/project_manager/diff.php
+++ b/tools/project_manager/diff.php
@@ -172,6 +172,42 @@ if ($L_text != $R_text) {
     echo $navigation_text;
 }
 
+// Load an associative array for which pages in the project have diffs.
+function load_page_diff_array(Project $project, $L_round, $R_round, bool $include_no_format = false): array
+{
+    if ($include_no_format) {
+        $un_formatter = new PageUnformatter();
+        $text_columns = "
+            $L_round->text_column_name as L_text,
+            $R_round->text_column_name as R_text,
+        ";
+    } else {
+        $text_columns = "";
+    }
+    $sql = "
+        SELECT image,
+            $L_round->user_column_name as username,
+            $text_columns
+            CAST($L_round->text_column_name AS BINARY) = CAST($R_round->text_column_name AS BINARY) AS is_same
+        FROM $project->projectid
+        ORDER BY image ASC
+    ";
+    $res = DPDatabase::query($sql);
+    $result = [];
+    while ($row = mysqli_fetch_assoc($res)) {
+        $result[$row["image"]] = [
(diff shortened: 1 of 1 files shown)


tools/project_manager/diff.php
```diff
@@ -172,6 +172,42 @@ function copyToClip(textstring) {
     echo $navigation_text;
 }
 
+// Load an associative array for which pages in the project have diffs.
```
bpfoley, line 175: Since we're trying to get better about this could we add
```
/**
  * @return array<string, array<string, mixed>> 
  * Return an associative array indexed by page image name, where each page has an array of attributes.
*/
function load_page_diff_array(Project $project, Round $L_round, Round $R_round, bool $include_no_format = false): array
```

tools/project_manager/diff.php
```diff
@@ -200,27 +236,24 @@ function get_navigation(
...
     $prev_image = "";
     $next_image = "";
     $prev_from_proofer = "";
     $next_from_proofer = "";
     $got_there = false;
     // construct the dropdown; work out where previous and next buttons should take us
-    while ([$this_val, $this_user, $is_empty_diff] = mysqli_fetch_row($res)) {
+    foreach ($diff_array as $this_val => $diff_record) {
+        $this_user = $diff_record["username"];
+        $is_diff = $format == "remove" ? $diff_record["is_diff_without_formatting"] : $diff_record["is_diff"];
```
bpfoley, line 249: If you wanted to be slightly cleverer you could push the ternary inside the index:
```$is_diff == $diff_record[$format == "remove" ? "is_diff_without_formatting": "is_diff"];```
or
```
$idx = $format == "remove" ? "is_diff_without_formatting": "is_diff";
$is_diff = $diff_record[$idx];
```

bpfoley (approved):

srjfoo (approved):
Tested between P3 and F1, but probably a better time to use it is between F1 and F2, to see if F1 has made incorrect proofreading changes.

cpeel:
Just rebased this branch and updated the sandbox.

This PR is being held until until we iron out if it's going in or not.

cpeel:
We're still waiting on feedback from people and this probably needs to be rebased so I put it in Draft to indicate that.

chrismiceli (approved):

PR #1115, opened 2024-02-27
Replace project state strings with constants

70ray:
There are some more cases involving rounds. This might conflict with what @cpeel is doing on task #2000 so not done yet.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/use_defined

### DIFF
--- a/pinc/user_is.inc
+++ b/pinc/user_is.inc
@@ -139,9 +139,10 @@ function that_user_is_over_PP_checked_out_limit($username)
         SELECT COUNT(*)
         FROM projects
         WHERE checkedoutby LIKE '%s'
-            AND state = '" . PROJ_POST_FIRST_CHECKED_OUT . "'
+            AND state = '%s'
         ",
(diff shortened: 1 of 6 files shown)


chrismiceli (approved):

pinc/user_is.inc
```diff
@@ -139,7 +139,7 @@ function that_user_is_over_PP_checked_out_limit($username)
         SELECT COUNT(*)
         FROM projects
         WHERE checkedoutby LIKE '%s'
-            AND state = 'proj_post_first_checked_out'
+            AND state = '" . PROJ_POST_FIRST_CHECKED_OUT . "'
```
cpeel, line 142: We're already in an `sprintf()` here, let's just make this another param we pass it. Ditto some of the others here.

70ray: Done this.

cpeel (approved):

srjfoo (approved):

PR #1116, opened 2024-02-27
mediawiki: Update docs on how to checkout, check tags

bpfoley:


pinc/3rdparty/mediawiki/README.md
```diff
@@ -35,19 +35,20 @@ To use them, see ../DifferenceEngineWrapper.inc
...
+  `git clone --depth 1 https://gerrit.wikimedia.org/r/mediawiki/core mediawiki`
+  `git fetch --tags`
 2. Show list of release tags
-  `git tags`
+  `git tag -l`
 3. Change to the release you want to use
   `git checkout <tag_name>`
 4. Copy updated files from locations mentioned above to this directory
-5. Unless the code has been fully converted to UTF-8:
+6. Unless the code has been fully converted to UTF-8:
```
chrismiceli, line 45: should it be 5 again?

bpfoley: Woops. Good point. I've updated all the other numbers.

cpeel:
These may work get the latest code, but I think the code has drifted enough that we can no longer integrate it into DP. We should try and if that's true just reframe this as "we lifted it from X and this is as far as we can go".

bpfoley:
Fair. I didn't actually try. I was just looking at all our third party deps, and wanted to check out the mediawiki code to see how far behind we were, and noticed the URL had changed (and a couple of the commands weren't quite right). We can abandon PR this if you don't feel it's worthwhile.

cpeel:
I think we should either abandon the PR, or do a quick check and if I'm right revise it to say as much (eg this is the end of the line for this borrowed code getting updated).

bpfoley:
Fair enough. It's not really something I want to explore right now, so I'll close this out. I'm sure it'll come up again!

PR #1119, opened 2024-02-28
Replace project state strings by round member variables

70ray:


tools/proofers/for_mentors.php
```diff
@@ -185,7 +185,7 @@ function output_project_details($mentored_round, $projectid, $nameofwork, $autho
 
 function get_beginner_projects_in_state($state, $mentored_round)
```
cpeel, line 186: This is a good time to add types to this function def so phpstan can help us enforce this change.

tools/project_manager/page_compare.php
```diff
@@ -167,10 +167,10 @@ public function radio_string($value, $label)
         return "<input type='radio' name='page_set' value='$value'$checked>" . html_safe($label);
     }
 
-    public function has_project_started_round($round_id)
+    private function has_project_started_round($round)
```
cpeel, line 170: Ditto adding function type defs here to assist with this change.

70ray: Done.

70ray:
Sandbox at: https://www.pgdp.org/~rp31/c.branch/use_defined_2

cpeel (approved):

srjfoo (approved):

PR #1120, opened 2024-02-29
projectmgr.php: Add "Your Available Projects" tab

bpfoley:
As requested in task 2051.

Testable in https://www.pgdp.org/~bfoley/c.branch/t2051-pm-available/tools/project_manager/projectmgr.php?show=user_avail


srjfoo (approved):

cpeel (approved):
@srjfoo - do we want to get some PM eyes on this before merging it? I'm fine without, but thought I'd ask.

tools/project_manager/projectmgr.php
```diff
@@ -74,6 +74,14 @@
     // adjust $_GET so will work corectly with refine search and sort and navigate
     // keep "user_all" or we won't know it is user all
     $_GET = array_merge($_GET, ['project_manager' => $user->username]);
+} elseif ($show_view == "user_avail") {
```
chrismiceli, line 77: We have `user_avail` in 3 spots. Would it make sense to have a constant ss it is clear they all represent the same thing?

chrismiceli (approved):

tools/project_manager/projectmgr.php
bpfoley: Yeah, that would be nice, but I'm not sure where to put them. The problem is that `$show_view` can be any of `['search_form', 'search', 'user_all', 'user_active', 'user_avail', 'blank', 'set_columns', 'config']`.

`set_columns` and `config` have special meanings in `handle_set_cols` and `handle_config` in `pinc/ProjectSearchResultsConfig.inc`

`search_form` and `search`are used by `tools/project_manager/projectmgr.php` and `tools/search.php`.

The rest are specific to `tools/project_manager/projectmgr.php` (and the corresponding settings in `echo_pm_tab()` in `userprefs.php`

I could, I guess, define them at the top of `pinc/ProjectSearchResultsConfig.inc`, but that would mean I'd need to `include_once` that in `userprefs.php`, which I'm not sure I love, but maybe it's not that big a deal.

srjfoo:
> @srjfoo - do we want to get some PM eyes on this before merging it? I'm fine without, but thought I'd ask.

It wouldn't hurt, I guess, but I doubt any of them would object. At least two PMs suggested the same thing, and the only comment against it was against changing what "Active" meant. This doesn't do that, it creates another whole tab. So I'm fine without asking, too, in this case.

PR #1121, opened 2024-02-29
List states explicitly in get_project_status_descriptor

70ray:
Sandbox at: https://www.pgdp.org/~rp31/c.branch/use_defined_3

### DIFF
--- a/pinc/project_states.inc
+++ b/pinc/project_states.inc
@@ -344,8 +344,6 @@ $project_status_descriptors = ['posted', 'PPd', 'proofed', 'created'];
  */
 function get_project_status_descriptor(string $which): object
 {
-    global $testing;
-
     $obj = new stdClass();
 
     switch ($which) {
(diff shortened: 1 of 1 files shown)


pinc/project_states.inc
```diff
@@ -397,5 +404,9 @@ function get_project_status_descriptor($which)
             die(html_safe("bad value for 'which': '$which'"));
     }
 
+    if ($testing) {
```
cpeel, line 407: This is fine for the PR, but this function could be used in places where we aren't outputting HTML (either because we're going to do `metaredirect()` or because we're using it in a script somewhere) so let's remove this before we merge it.

cpeel (approved):

pinc/project_states.inc
70ray: Done this.

srjfoo (approved):

Issue #1005, opened 2024-01-01
page_compare.php should not hard-code round IDs


cpeel:
`tools/project_manager/page_compare.php` hard-codes round values in the Comparator:
```php
        $this->L_round_options = ['P1', 'P2', 'P3', 'F1'];
        $this->R_round_options = ['F1', 'F2'];
```
As a rule we do not hard-code round IDs outside of `stages.inc` to ensure that other sites that use our code don't need to slog through the code if they have a different number of rounds (also to help future proof the code for pgdp.net if we change the number of rounds).

The Comparator should dynamically determine these in the constructor.

70ray:
Possibly also consider project_states.inc.

cpeel:
> Possibly also consider project_states.inc.

Ah, yes. What's even more confusing is that `project_states.inc` needs to be included _before_ the rounds are created in `stages.inc` since the round creation coded uses the constants.

70ray:
> > Possibly also consider project_states.inc.
> 
> Ah, yes. What's even more confusing is that `project_states.inc` needs to be included _before_ the rounds are created in `stages.inc` since the round creation coded uses the constants.

I've made a PR to try to resolve this.

cpeel:
Fixed in #1124

PR #1124, opened 2024-03-02
Generalise round options

70ray:
This removes the hard-coded round ids, addressing issue #1005 

The compare without formatting link on the project page should comtinue to work as before.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/no_hard_code_rounds

### DIFF
--- a/tools/project_manager/page_compare.php
+++ b/tools/project_manager/page_compare.php
@@ -16,18 +16,25 @@ class Comparator
 {
     public function __construct()
     {
-        global $PROJECT_STATES_IN_ORDER;
+        $all_rounds = Rounds::get_all();
+        $all_ids = Rounds::get_ids();
 
-        $this->L_round_options = ['P1', 'P2', 'P3', 'F1'];
-        $this->R_round_options = ['F1', 'F2'];
-        $this->state_index = array_flip($PROJECT_STATES_IN_ORDER);
+        // L_round_options are all rounds except the last
+        $this->L_round_options =  array_slice($all_ids, 0, count($all_ids) - 1);
+        // R_round_options are all formatting rounds
+        $this->R_round_options = array_keys(array_filter($all_rounds, 'is_formatting_round'));
+        // default R_round is first R-round
+        $this->default_R_round_id = $this->R_round_options[0];
+        // default L_round is the preceding one
+        $this->default_L_round_id = $all_ids[array_search($this->default_R_round_id, $all_ids) -1];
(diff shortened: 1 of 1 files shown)


cpeel (approved):

tools/project_manager/page_compare.php
```diff
@@ -16,18 +16,25 @@ class Comparator
...
+        $all_ids = Rounds::get_ids();
+
+        // L_round_options are all rounds except the last
+        $this->L_round_options = array_slice($all_ids, 0, count($all_ids) - 1);
+        // R_round_options are all formatting rounds
+        $this->R_round_options = array_keys(array_filter($all_rounds, 'is_formatting_round'));
+        // default R_round is first R-round
+        $this->default_R_round_id = $this->R_round_options[0];
+        // default L_round is the preceding one
+        $this->default_L_round_id = $all_ids[array_search($this->default_R_round_id, $all_ids) - 1];
```
cpeel, line 29: This might be easier if you use the objects rather than the IDs for all of these and then just use `get_round_param()` to get the object. We probably don't really care if they manipulate the URL and get something they weren't expecting.
```php
       // L_round_options are all rounds except the last
       $this->L_round_options = array_slice($all_rounds, 0, count($all_rounds) - 1);
       // R_round_options are all formatting rounds
       $this->R_round_options = array_filter($all_rounds, 'is_formatting_round');
       // default R_round is first R-round
       $this->default_R_round = $this->R_round_options[0];
       // default L_round is the preceding one
       $this->default_L_round = Rounds::get_by_number($this->default_R_round->round_number - 1);

       $this->L_round_id = get_round_param($_GET, "L_round_id", $this->default_L_round)->id;
       $this->R_round_id = get_round_param($_GET, "R_round_id", $this->default_R_round)->id;
```
Although looking at the rest of the script it might be clearer if the whole script dealt with round objects instead of round IDs. Because ultimately we convert them into round objects anyway:
```php
        $L_round = get_Round_for_round_id($this->L_round_id);
        $R_round = get_Round_for_round_id($this->R_round_id);
```

70ray: I've done this and removed some of the member functions so the $L_round and $r_round do not have to be class members.
It would be possible to remove the class altogether if we put some parameters in the functions. It seems a bit OTT to use a class just to avoid a few parameters or global variables.

70ray:
Sandbox updated.

cpeel (approved):

srjfoo (approved):

PR #1125, opened 2024-03-02
Rename project_states to ProjectState

cpeel:
Now that `project_states.inc` only provides the structure to define project states instead of defining them itself we can move it to the same structure as the Activity/Stage/Round/Pool files that it mirrors.

https://www.pgdp.org/~cpeel/c.branch/massage-project-state-file/


pinc/ProjectState.inc
```diff
@@ -325,7 +303,11 @@ function sql_collator_for_project_state($state_column)
 }
 
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
-$project_status_descriptors = ['posted', 'PPd', 'proofed', 'created'];
+
+function get_project_status_descriptors()
+{
+    return ['posted', 'PPd', 'proofed', 'created'];
```
srjfoo, line 309: Just making sure this doesn't need to include `'deleted'`.

cpeel: None of the current consumers expect it to, so no.

srjfoo (approved):

70ray (approved):

PR #1127, opened 2024-03-02
page_detail.php: Expose UI to allow PMs to filter by user

bpfoley:
As requested in https://www.pgdp.net/c/tasks.php?task_id=2049

Sandbox and example page at https://www.pgdp.org/~bfoley/c.branch/t2049-pagedetail-by-user/tools/project_manager/page_detail.php?project=projectID5894b5a113ec9&show_image_size=0&select_by_user=srjfoo&select_by_round=P3

### DIFF
--- a/tools/project_manager/page_detail.php
+++ b/tools/project_manager/page_detail.php
@@ -7,6 +7,8 @@ include_once($relPath.'page_table.inc');
 
 require_login();
 
+$errors = [];
+
 $projectid = get_projectID_param($_GET, 'project');
 $show_image_size = get_integer_param($_GET, 'show_image_size', 0, 0, 1);
 $round_for_page_selection = get_round_param($_GET, 'select_by_round', null, true);
@@ -17,23 +19,31 @@ $round_for_page_selection = get_round_param($_GET, 'select_by_round', null, true
 //    <username> = show specific user
 $username_for_page_selection = array_get($_GET, 'select_by_user', null);
 if ($username_for_page_selection === '') {
-    $username_for_page_selection = $pguser;
+    $username_for_page_selection = User::current_username();
 }
+
+$project = new Project($projectid);
+
+// Only the project PM (and SA/PFs) can use select_by_user. This is to prevent
+// users from using this value to determine what user proofread what page in a
+// project. This is possible because while the table renderer will hide
+// usernames for users not privileged to access them, select_by_user and
+// select_by_round will still only present the requested rows allowing the
+// requester to determine what pages a user did in which round.
+$may_select_user = $project->can_be_managed_by_current_user;
 
 // Validate user selection
 if ($username_for_page_selection) {
-    // Only SAs and PFs can use select_by_user. This is to prevent users from
-    // tweaking this value and determining what user proofread what page in a
-    // project. This is possible because while the table renderer will hide
-    // usernames for users not privileged to access them, select_by_user and
-    // select_by_round will still only present the requested rows allowing the
-    // requester to determine what pages a user did in which round.
-    if (! (user_is_a_sitemanager() || user_is_proj_facilitator())) {
-        $username_for_page_selection = $pguser;
+    if (!$may_select_user) {
+        $username_for_page_selection = User::current_username();
     }
 
     // Validate the requested username
     if (!User::is_valid_user($username_for_page_selection)) {
+        $errors[] = sprintf(
+            _('Error: "%1$s" is not a valid user. Showing all pages.'),
+            $username_for_page_selection
+        );
         $username_for_page_selection = null;
     }
 }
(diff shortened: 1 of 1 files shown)


styles/layout.less
```diff
@@ -1497,6 +1497,11 @@ table.search-column {
     }
 }
 
+#user-filterbox{
+    #linkbox;
+    padding: 0.5em 0.5em;
```
cpeel, line 1502: Do we really need the padding tweak? This makes it a snowflake with all the other uses of this type of floating box.

tools/project_manager/page_detail.php
```diff
@@ -43,18 +47,55 @@
...
 }
 
-$project = new Project($projectid);
 $state = $project->state;
 $title = $project->nameofwork;
 $page_details_str = _('Page Detail');
 
 output_header("$page_details_str: $title", NO_STATSBAR);
 
+echo "<div style='display: flex; justify-content: space-between'>";
```
cpeel, line 56: Is this necessary? We don't seem to use anything like this the other places where we output a linkbox.

bpfoley: I didn't want to special case it, but the float gets positioned very badly because of the wide page details table if we do it the same way that linkbox is done on the other pages.

I went through a few iterations to end up with this solution. NB we do use flex boxes in a few other parts of the codebase.

cpeel: Ah yes, the "page is stupid wide with a lot of rounds" thing.

Let's add a comment here on why we're doing it this way for the next person that comes along.

styles/layout.less
bpfoley: The other linkboxes work around the spacing by having a `<p>` inside the `<form>` which introduces extra padding. It actually ends up being more padding than the `0.5em` I added here.

In early iterations of this PR that looked like a bit much, but since the box is now at the top right of the page, it doesn't really matter, so I've removed the special case `user-filterbox`.

tools/project_manager/page_detail.php
bpfoley: Done.

cpeel (approved):

tools/project_manager/page_detail.php
```diff
@@ -17,19 +17,23 @@
 //    <username> = show specific user
```
chrismiceli, line 17: if round is not valid, I get a fatal error. If we had the options, we would encounter this less often unless someone is manipulating query params.

tools/project_manager/page_detail.php
```diff
@@ -43,18 +47,61 @@
...
+    echo "<td class='bold'>", _("Username"), "</td>";
+    echo "<td><input type='text' name='select_by_user' value='" . attr_safe($username) . "' required></td>";
+    echo "</tr>";
+    echo "<tr>";
+    echo "<td class='bold'>", _("User in round"), "</td>";
+    echo "<td><input type='text' name='select_by_round' value='" . attr_safe($roundid) . "'></td>";
+    echo "</tr>";
+    echo "<tr>";
+    echo "<td colspan=2>";
+    echo "<input type='submit' value='" . attr_safe(_("Refresh")) . "'>";
```
chrismiceli, line 96: might be worth having a clear button for PMs that want to go back to seeing all pages.

tools/project_manager/page_detail.php
```diff
@@ -43,18 +47,61 @@
...
+    echo "<input type='hidden' name='show_image_size' value='$show_image_size'>";
+    echo "<table>";
+    echo "<tr><td colspan=2>", _("Filter pages by:"), "</td></tr>";
+    echo "<tr>";
+    echo "<td class='bold'>", _("Username"), "</td>";
+    echo "<td><input type='text' name='select_by_user' value='" . attr_safe($username) . "' required></td>";
+    echo "</tr>";
+    echo "<tr>";
+    echo "<td class='bold'>", _("User in round"), "</td>";
+    echo "<td><input type='text' name='select_by_round' value='" . attr_safe($roundid) . "'></td>";
```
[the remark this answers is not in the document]
bpfoley: Good idea. Will do.

tools/project_manager/page_detail.php
```diff
@@ -43,18 +47,61 @@
...
+// `float: right` like the other linkboxes in the codebase do because on
+// this page the inflexible and very wide page details table causes the
+// linkbox to display in ugly locations.
+echo "<div style='display: flex; justify-content: space-between'>";
+echo "<div>";
 echo "<h1>" . html_safe($title) . "</h1>\n";
 echo "<h2>$page_details_str</h2>\n";
+echo "</div>";
+if ($may_select_user) {
+    echo "<div>";
```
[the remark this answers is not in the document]
bpfoley: I'll see if I can remove some of the layers, but I needed the top level div-inside-div so that the linkbox didn't get stretched to the same vertical size as the h1/h2 div.

tools/project_manager/page_detail.php
bpfoley: I don't think it's possible to do that directly within the form, because form submits explicitly put all text inputs into the query even if they're empty.

Unfortunately, `page_detail.php` distinguishes between `$select_by_user` being the empty string (ie `...&select_by_user=&...` in the URL) and null (ie absent entirely from the URL). I did have a 'clever trick' to fake this with a button-link, but I decided against it in the end.

```
function echo_button_link(string $label, string $url, bool $enabled)
{
    if ($enabled) {
        // TODO(bfoley) Move to .less
        echo "<a style='color: black; text-decoration: none' href='", attr_safe($url), "'>";
        echo "<button type='button'>$label</button>";
        echo "</a>";
    } else {
        echo "<button type='button' disabled>$label</button>";
    }
}
echo_button_link(_("Show All Pages"), '?project=$projectid&show_image_size=$show_image_size' style=', !is_null($select_by_user));
```

We can bring it back if we decide it's worth it.

srjfoo: Yeah, Brian had the Show All Pages in the form initially. That didn't feel like the right place for it. We're only going to show this to PFs, Squirrels and PMs. Ordinary users won't see it, because they can't `select_by_user`, so they need the "Show just my pages" and "Show all pages instead" links  not to move from where they are, above the pages table. Hopefully, PMs are smart enough to remember to look there if they need to clear the form. That approach works well for me in the sandbox version.

tools/project_manager/page_detail.php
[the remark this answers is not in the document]
srjfoo: I like that idea, too!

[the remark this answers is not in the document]
bpfoley: Done.

tools/project_manager/page_detail.php
```diff
@@ -43,18 +47,61 @@
...
+    echo "<input type='hidden' name='project' value='$projectid'>";
+    echo "<input type='hidden' name='show_image_size' value='$show_image_size'>";
+    echo "<table>";
+    echo "<tr><td colspan=2>", _("Filter pages by:"), "</td></tr>";
+    echo "<tr>";
+    echo "<td class='bold'>", _("Username"), "</td>";
+    echo "<td><input type='text' name='select_by_user' value='" . attr_safe($username) . "' required></td>";
+    echo "</tr>";
+    echo "<tr>";
+    echo "<td class='bold'>", _("User in round"), "</td>";
```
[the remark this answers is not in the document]
bpfoley: I've added the `<label>`s, but kept the table for now because it's the easiest way to keep the inputs aligned. The received wisdom these days seems to be to use flexboxes, which is fair enough.

I'd rather not make this linkbox a special case for right now though, and I've opened Task #2074 to remind us that we should do this.

tools/project_manager/page_detail.php
bpfoley: ACK. It's thrown early in the input processing by the call to `get_round_param` which is good, that's what we want. And now, as you say, users are less likely to trip over it unless they're "hacking" URLs.

tools/project_manager/page_detail.php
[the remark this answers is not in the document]
bpfoley: OK. One layer of <div> removed.

srjfoo:
Should the existence of username be being validated?

1. If I check for all rounds for a known good username who hasn't proofed in the project, I just get an empty table.
2. If I check for a known user who has proofed in the project but not in that round
3. If I check for a nonexistent user, I get a cleared form and the full page table.

1 and 2 sound correct; shouldn't 3 be getting a nonexistent user error?

bpfoley:
> Should the existence of username be being validated?
> 
>     1. If I check for all rounds for a known good username who hasn't proofed in the project, I just get an empty table.
> 
>     2. If I check for a known user who has proofed in the project but not in that round
> 
>     3. If I check for a nonexistent user, I get a cleared form and the full page table.
> 
> 
> 1 and 2 sound correct; shouldn't 3 be getting a nonexistent user error?

Good point. Will do.

chrismiceli (approved):
Minor comment, but rewriting the history make it hard to review revisions of your changes. I'm ok with rebasing down to 1 commit before merge, but during a review I lose track of what was in each version.

bpfoley:
> Minor comment, but rewriting the history make it hard to review revisions of your changes. I'm ok with rebasing down to 1 commit before merge, but during a review I lose track of what was in each version.

That's a good point. I'll cut back on the squashing when there's an iterative review process and a mostly static patchset.

srjfoo (approved):
LGTM!

PR #1128, opened 2024-03-03
Update release tooling

cpeel:
This updates some of the pre-release tooling we use to confirm all database changes we made to the live DB on TEST are represented by the upgrade scripts. Every release I bang my head against subtle differences in the output due to collation (among others) and this time I decided to just fix it.

I've also updated the dump format so it is mostly just what `mysqldump` does rather than mirroring the phpMyAdmin dump format which we haven't used in well over a decade. There are no functional changes in `db_schema.sql`, just changes to comments (mostly format).

### DIFF
--- a/SETUP/dump_db_schema.template
+++ b/SETUP/dump_db_schema.template
@@ -7,7 +7,7 @@ db_password='<<DB_PASSWORD>>'
 db_name='<<DB_NAME>>'
 code_dir='<<CODE_DIR>>'
 
-setup_dir=`dirname $0`
+setup_dir=$(realpath $(dirname $0))
 
 export LC_ALL=POSIX
 # because on www.pgdp.net, none of the LC_* are defined,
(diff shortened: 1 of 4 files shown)


SETUP/dump_db_schema.template
```diff
@@ -18,6 +18,10 @@ export LC_ALL=POSIX
 # So how will we determine which ones to dump?
 if [ "$1" = '--table-list-from-db' ]; then
     table_list_source='db'
+    # If they've passed in another arg, assume it's the DB name
```
bpfoley, line 21: This might be an opportunity to put `set -euo pipefail` at the top of this template, but to do that we're going to need to change things that rely on expanding undefined variables. The `set -u` safe way of spelling `if [ ! -z "$foo" ]` is `if [ -n "${foo-}" ]`.
 `if [ -n "${2-}" ]; then` here, and `elif [ "${1-}" = "--table-list-from-code" ]; then` and similar above and below.

SETUP/dump_db_schema.template
```diff
@@ -5,7 +5,7 @@
 db_user='<<DB_USER>>'
 db_password='<<DB_PASSWORD>>'
 db_name='<<DB_NAME>>'
-code_dir=<<CODE_DIR>>
+code_dir='<<CODE_DIR>>'
 
 setup_dir=`dirname $0`
```
bpfoley: A minor thing, but backticks for expansion is regarded as not-great style, mainly because you can't nest them. Could you make this `setup_dir=$(dirname $0)`?

SETUP/dump_db_schema.template
bpfoley: Also perhaps make this whole if...elif..else chain into
```
case ${1-} in
  --table-list-from-db)
   ...;;
  --table-list-from-code)
   ...;;
   *)
   ...;;
esac
```

chrismiceli (approved):

SETUP/dump_db_schema.template
cpeel: I think something got crossed somewhere because that's what the current version does.

SETUP/dump_db_schema.template
cpeel: Got `set -euo pipefail` added to both scripts. Turns out the "get db list from code" was subtly broken so I fixed that too.

srjfoo (approved):

SETUP/dump_db_schema.template
bpfoley: I'm talking about the backticks in ``` setup_dir=`dirname $0` ```, but no matter, it's not worth worrying about.

bpfoley (approved):

SETUP/dump_db_schema.template
cpeel: Right, but [none of the current code in the PR](https://github.com/DistributedProofreaders/dproofreaders/pull/1128/files) has that?

PR #1133, opened 2024-03-07
Patch erusev/parsedown to avoid a PHP Notice on malformed input

bpfoley:
The notice is for indexing beyond the end of a string, and PHP's result (an empty string) has logically the right effect in the parser code, it just outputs an annoying notice which is visible on the web page.

This is far from exhaustive. It would need to be done in a lot of other places too, but serves as an interesting example of how to patch vendored dependencies if we need to. The `cwegans/composer-patches` composer plugin adds only 164kB to `vendor/` so is pretty lightweight too.

Sandbox here: https://www.pgdp.org/~bfoley/c.branch/parsedown-patch/tasks.php?action=show&task_id=2296

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,8 @@
         "erusev/parsedown": "1.8.0-beta-7",
         "erusev/parsedown-extra": "0.8.0",
         "voku/portable-utf8": "^6.0",
-        "phpmailer/phpmailer": "^6.8"
+        "phpmailer/phpmailer": "^6.8",
+        "cweagans/composer-patches": "^1.7.3"
     },
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^3.0",
@@ -29,5 +30,18 @@
         "symfony/polyfill-intl-grapheme": "*",
         "symfony/polyfill-intl-normalizer": "*",
         "symfony/polyfill-mbstring": "*"
+    },
+    "extra": {
+        "composer-exit-on-patch-failure": true,
+        "patches": {
+            "erusev/parsedown": {
+                "Parser PHP Notice on invalid HTML attribute": "patches/composer/parsedown-1.diff"
+            }
+        }
+    },
+    "config": {
+        "allow-plugins": {
+            "cweagans/composer-patches": true
+        }
     }
 }
--- a/composer.lock
+++ b/composer.lock
@@ -4,8 +4,56 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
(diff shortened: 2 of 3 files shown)


chrismiceli:
I worry about maintainability if we start patching packages. Mainly because:
1. It would make upgrades more difficult as the patch won't apply to newer versions in some cases
2. It requires us to understand the package internals which we may make false assumptions about as we patch
3. If there is a newer version and we can't patch for some reason, we get stuck on an old version that may have security vulnerabilities, etc.

bpfoley:
> I worry about maintainability if we start patching packages. Mainly because:
> 
>     1. It would make upgrades more difficult as the patch won't apply to newer versions in some cases
> 
>     2. It requires us to understand the package internals which we may make false assumptions about as we patch
> 
>     3. If there is a newer version and we can't patch for some reason, we get stuck on an old version that may have security vulnerabilities, etc.

Absolutely. Patching is a sharp tool, and gratuitous patching is something to be avoided if possible for all the reasons you mention.

However there is the flip side: when we have a package that we can't readily replace and that package has a bug that hasn't been/won't be fixed upstream and it's causing us problems, patching is a less-bad solution.

It's a common pattern in larger projects to have patches against upstream packages, and to simultaneously open issues/pull requests against those packages to fix the problem. Once the upstream package has a release with a fix in it, the downstream project can update its dependency and remove the patch. It's extra work, but it means that the downstream project's release schedule isn't tied to the release schedule of one, or potentially more upstream packages.

The most extreme instance of this is for distribution maintainers such as Debian, NixOS. A random example that comes to mind is https://github.com/NixOS/nixpkgs/tree/master/pkgs/servers/nosql/mongodb

Note, I'm not saying this this particular patch is the right way to solve this particular issue. In fact having slept on it, I suspect it would be better to just use the error control operator `@` on all our  `$Parsedown->text` calls. I didn't realize that `@` actually suppresses errors in (nested) function calls, not just simple expressions. Eg

```php
  function inner_produces_notice($str) { return $str[100] . "y"; }
  function outer($str) { return inner_produces_notice($str); }
  echo @outer("tooshort") . "\n";
```

cpeel:
A few thoughts:
1. It's great to know we have a tool to do patches against vendor code if we absolutely have to. I'd never looked into this before and appreciate you figuring it out, Brian.
2. I think in this instance we should use the `@` approach to quell notices generated from Parsedown.
3. Per my async chat with Brian we should probably look into replacing Parsedown with a better-supported tool if we can.

bpfoley:
I've opened a PR with the `@` approach as #1134. Closing this patch as it's no longer useful.

PR #1134, opened 2024-03-08
Centralize Parsedown code, and suppress Notices generated by the parser

bpfoley:
These were visible in comments on the task center such as "&type c:\windows\win.ini&"

They rendered OK, but also emitted PHP Notice messages into the web page.

Example in sandbox at https://www.pgdp.org/~bfoley/c.branch/quell-parsedown-notices/tasks.php?action=show&task_id=2296

### DIFF
--- a/pinc/misc.inc
+++ b/pinc/misc.inc
@@ -651,6 +651,27 @@ function sanitize_html($html, $parent_tag = 'div')
     return $purifier->purify($html);
 }
 
+/* Parse markdown text string using Parsedown library and output HTML
+ * @param $message Markdown text
+ * @return string HTML content
+ */
+function render_markdown_as_html(string $message): string
+{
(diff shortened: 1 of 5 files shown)


chrismiceli (approved):

srjfoo (approved):

pinc/misc.inc
```diff
@@ -1622,3 +1622,20 @@ if (!class_exists('\ValueError')) {
     {
     }
 }
+
+function render_markdown_as_html(string $message): string
```
cpeel, line 1626: This isn't a "useful exception" but it's in that section -- can we move it up somewhere else in the file? Perhaps under `sanitize_html()`. And let's add a docstring above it too please.

bpfoley: Done.

PR #1143, opened 2024-03-15
remote_file_manager.php: Case insensitive, natural sorting of dirs

bpfoley:
This makes it easier for site admins -- they don't have to remember if user dirs in ~uploads/Users are capitalized or not.


cpeel (approved):

srjfoo (approved):

cpeel:
This will get merged after I cut a release.

PR #1144, opened 2024-03-16
api/v1/projects/:id: Allow endpoint to return selected fields

bpfoley:
Clients can already access a subset of fields when listing projects with `api/v1/projects` so it makes sense to allow this for individual projects too. And in particular clients may want to avoid the `comments` field because it can be very large.

### DIFF
--- a/api/dp-openapi.yaml
+++ b/api/dp-openapi.yaml
@@ -393,6 +393,13 @@ paths:
         required: true
         schema:
           type: string
+      - name: field
+        description: Field to return, if not set all fields are returned.
+          To request multiple fields, append with [] and pass one per desired
+          field (e.g. field[]=projectid&field[]=author).
+        in: query
+        schema:
+          type: string
       responses:
         200:
           description: project response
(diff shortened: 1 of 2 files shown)


cpeel (approved):

api/dp-openapi.yaml
```diff
@@ -393,6 +393,13 @@ paths:
         required: true
         schema:
           type: string
+      - name: field
+        description: Field to return, if not set all fields are returned.
+          To request multiple fields, append with [] and pass one per desired
+          field (e.g. field[]=projectid&field[]=author).
+        in: query
+        schema:
+          type: string
```
chrismiceli, line 402: ideally we could identify the fields to be requested here as some enumeration

chrismiceli (approved):

api/dp-openapi.yaml
bpfoley: That's a good point. It makes the yaml even more verbose and redundant, but I guess that can't be helped.

cpeel: It's a good idea but I'd punt on this for now. We don't enumerate them for the other endpoints and someone can request the data and then pare down via fields.

srjfoo (approved):

PR #1145, opened 2024-03-17
Get info in LPage.inc

70ray:
Move the code for getting info. from PPage to Lpage leaving PPage to wrap the info. in html. The LPage function can then be used also for the proofreading API.
Sandbox at: https://www.pgdp.org/~rp31/c.branch/lp_data

### DIFF
--- a/pinc/LPage.inc
+++ b/pinc/LPage.inc
@@ -7,6 +7,7 @@ include_once($relPath.'Project.inc');
 include_once($relPath.'page_tally.inc');
 include_once($relPath.'send_mail.inc');
 include_once($relPath.'daily_page_limit.inc'); // get_dpl_count_for_user_in_round
+include_once($relPath.'forum_interface.inc'); // get_forum_user_id()
 
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 
@@ -260,13 +261,13 @@ class RoundInfo
 {
     public string $round_id;
     public string $username;
-    public string $user_id;
+    public string $forum_user_id;
 
-    public function __construct(string $round_id, string $username, string $user_id)
+    public function __construct(string $round_id, string $username, string $forum_user_id)
     {
         $this->round_id = $round_id;
         $this->username = $username;
-        $this->user_id = $user_id;
+        $this->forum_user_id = $forum_user_id;
     }
 }
 
@@ -339,8 +340,8 @@ class LPage
         foreach ($this->round->other_rounds_with_visible_usernames as $other_round_id) {
             $other_round = get_Round_for_round_id($other_round_id);
             $username = $this->get_username_for_round($other_round);
-            $user_id = get_forum_user_id($username);
-            $round_info_array[] = new RoundInfo($other_round->id, $username, $user_id);
+            $forum_user_id = get_forum_user_id($username);
+            $round_info_array[] = new RoundInfo($other_round->id, $username, $forum_user_id);
         }
 
         return $round_info_array;
--- a/pinc/LPage.inc
+++ b/pinc/LPage.inc
@@ -253,6 +253,20 @@ function project_continuity_test($project, $orig_state, $no_more_pages)
     }
 }
 
+class RoundInfo
+{
+    public $round_id;
+    public $username;
+    public $user_id;
+
+    public function __construct($round_id, $username, $user_id)
+    {
+        $this->round_id = $round_id;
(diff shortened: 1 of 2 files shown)


pinc/LPage.inc
```diff
@@ -310,6 +310,21 @@ class LPage
...
+    public function get_info()
+    {
+        foreach ($this->round->other_rounds_with_visible_usernames as $other_round_id) {
+            $other_round = get_Round_for_round_id($other_round_id);
+            $username = $this->get_username_for_round($other_round);
+            $user_id = get_forum_user_id($username);
+            $round_info_array[] = [$other_round->id, $username, $user_id];
+        }
+
+        return [
```
cpeel, line 322: I'd be inclined to return either an associative array with labeled keys or an adhoc stdClass object with named attributes. Right how this is a very opaque structure that requires a lot of mental context to trace:
```
[
    <imagename>,
    [
        [<roundid>, <username>, <userid>],
        ...
    ]
]
```

compared to:
```
[
    "imagename": <imagename>,
    "round_info": [
        [
            "round_id": <roundid>,
            "username": <username>,
            "user_id <user_id>,
        ],
        ...
    ]
]
```

I'm curious on @bpfoley's thoughts on this too.

cpeel: It would probably be clearer if we broke this out into two functions: one that returned the pagename and one that returned the round details.

cpeel (commented):
Think this is a great idea, good call.

pinc/LPage.inc
bpfoley: Yeah. If this was Python, I'd say "use a dataclass instead of a list". In PHP 8 we can approximate this with (from memory)
```php
class RoundInfo
{
  public function __construct(
    public string $round_id,
    public string $username,
    public int $user_id
  ) {}
}
```
but since it's PHP 7.4 we can't do this neatly yet, so I'd say using `stdClass` (there's an example in `get_OCR_pseudoround`) is as close as we can get for now.

70ray: Classes are probably more efficient than associative arays because the keys are not duplicated for every instance [see https://www.exakat.io/en/moving-from-array-to-class/](https://www.exakat.io/en/moving-from-array-to-class/). But as Casey says, clarity of the code is the most important thing. I'll try both ways and see how they look.

70ray:
Sandbox updated.

tools/proofers/PPage.inc
```diff
@@ -178,7 +178,7 @@ class PPage
         echo "</span>\n";
     }
 
-    private function make_round_user_string($info)
+    private function make_round_user_string(RoundInfo $info): String
```
bpfoley, line 181: `string` instead of `String`?

cpeel (approved):

pinc/LPage.inc
```diff
@@ -253,6 +253,20 @@ function project_continuity_test($project, $orig_state, $no_more_pages)
     }
 }
 
+class RoundInfo
+{
+    public $round_id;
+    public $username;
+    public $user_id;
```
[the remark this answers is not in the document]
70ray: I've added a comment about RoundInfo.
The comment after the include is not appropriate now and we can remove the include.
But should we have the include statement in Lpage.inc? LPage works without it because `forum_interface.inc` is included via at least one other path (through Project.inc). What do @cpeel  and @bpfoley  think about this? Use the minimum necessary includes, or show an include for where every function used in the file is defined, or don't worry about it.

chrismiceli (approved):

pinc/LPage.inc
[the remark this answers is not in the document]
cpeel: In general we should not rely on transient includes and if a function is used in a file we should have an include that pulls in the file with that function (exceptions are `misc.inc` and some of the site structure files which the github action will alert if those are included so there's some checks against that).

Our includes are getting better but it's still a bit of a mess.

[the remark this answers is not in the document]
srjfoo: > I've added a comment about RoundInfo.

My apologies for not being specific enough. When I saw `user_id`, I immediately thought of `users.u_id`, not the phpBB user_id. At least not until I saw the call to `get_forum_user_id()`. However, I'm fine with whatever others think works best. It's obvious once you read the rest of the code.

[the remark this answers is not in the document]
cpeel: That's a really good point Sharon. It would be clearer if we called the variable `forum_user_id`.

70ray:
Sandbox updated.

cpeel (approved):

srjfoo (approved):

PR #1130, opened 2024-03-06
Pageload smoketest

bpfoley:
A smoketest that sets up `site_vars.php` and a minimal set of database tables so that PGDP can be served from the `php -S` command line web server.

Then it loads many pages (soon to be every page) with representative parameters and checks if there are any PHP errors, warnings, or notices in the server logs.

Useful for, e.g., catching variables that happen not to be defined at runtime, and other issues that PHPStan and the PHP linter can't find.

This is deliberately not an exhaustive test of all possible inputs -- smoke tests are meant to be fast and to weed out obvious brokenness.  That said, we probably should check each of, e.g. `task.php`'s `action=` parameters so we get reasonable code coverage.

This is incomplete, but is good enough to be useful as is.

Next steps:
1. Add all the remaining pages, and any necessary database tables needed.

PR #1148, opened 2024-03-18
Add an include_once dependency graphing tool for dev purposes.

bpfoley:
Example use:
```
./SETUP/showdeps.py --mode=graph pinc/base.inc
```
and feed the output into a tool such as https://dreampuf.github.io/GraphvizOnline

### DIFF
--- a/SETUP/showdeps.py
+++ b/SETUP/showdeps.py
@@ -1,0 +1,77 @@
+#!/usr/bin/env python3
+"""Report the transitive include_once graph for a list of PHP files.
+
+For usage, run with --help
+
+Known to work with Python 3.8 or later on Linux and macOS, but should
+also work with any Python 3.x on any supported Python platform.
+"""
+
+import argparse
+import os
+import re
+import sys
+
+from collections import defaultdict
+from contextlib import suppress
+
+def get_deps(filename: str, get_all: bool, dep_graph: dict):
+    """Get the transitive include dependencies for filename, updating dep_graph"""
+    with suppress(FileNotFoundError):
+        with open(filename, 'r') as f:
+            for l in f:
+                # Assuming code is well indented, conditional includes
+                # won't be at the start of a line.
+                anchor = "" if get_all else "^"
+                m = re.search(anchor + r"(require|include)_once *\(([^)]+)\);", l)
+                if not m:
+                    continue
+                # Do some 'good enough' parsing of $relPath.'foo.inc'
+                p = m[2].replace('"', '\'')
+                m2 = re.match(r"\$relPath\.'([^']+)'", p)
+                if m2:
+                    p = 'pinc/' + m2[1]
+                if '$relPath' not in p:
+                    p = os.path.normpath(p)
+                    dep_graph[filename].append(p)
+            for dep in dep_graph[filename]:
+                if dep not in dep_graph:
+                    get_deps(dep, get_all, dep_graph)
+
+def main() -> int:
+    p = argparse.ArgumentParser()
+    p.add_argument('--mode',
+                   type=str,
+                   default='graph',
+                   help='Produce a graphviz graph of dependencies, or output a list of dependencies, one per line',
+                   choices=['graph', 'deps']
+    )
+    p.add_argument('--get-all',
+                   action='store_true',
+                   help='get all dependencies, including conditional ones.'
+    )
+    p.add_argument('files',
+                   nargs='+',
+                   help='PHP files to scan for dependencies'
+    )
+    args = p.parse_args()
+
+    dep_graph = defaultdict(list)
+    for fn in args.files:
+        get_deps(fn, args.get_all, dep_graph)
+
+    if args.mode == 'graph':
+        print("digraph deps {")
+        for fn, deps in dep_graph.items():
+            print("".join(f'    "{fn}" -> "{d}";\n' for d in deps), end='')
+        print("}")
+    elif args.mode == 'deps':
+        deps = set()
+        for d in dep_graph.values():
+            deps.update(d)
+        print("\n".join(sorted(deps)))
+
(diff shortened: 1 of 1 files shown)


chrismiceli (commented):
I think this is the first python script in this repo, right? Would it be an easier script to maintain in another language like a shell script or PHP that we already use copiously?

bpfoley:
Well, #1130 is now a Python script too -- I rewrote it after it went way past my 'shell script complexity threshold'. I'd definitely prefer to keep #1130 as a Python script because there are a few details relating to output buffering and HTTP request and cookie handling that you need to get right for the smoke tests to work reliably. I've figured out how to do them with Python's `subprocess` and `urllib` and I don't really trust PHP not to 'helpfully' do the wrong thing behind our backs.

I'd strongly prefer not to do this tool as a shell script for a few reasons
- the most immediate one is my gut instinct that cumulatively it's above the complexity limit of what I like to do in a shell script: recursion, hash tables, string processing, regexps. All possible, yes, but easy to make mistakes.
-  the string processing and regexp handling would get messier.
-  it relies on a dict to represent the dependencies and bash only got hash tables in v4.0, and macOS is never going to move past v3.2 because that's the last version under GPLv2.

I could rewrite it as PHP, although I think it would be a bit longer and a bit uglier.

Note also that this is just a tool for developer curiosity/convenience. It's not used in CI, installation, or production, so to a certain extent it doesn't really matter if it breaks and nobody can maintain it.

cpeel:
The question is totally valid. I'm ok with python code like this in ancillary places (like CI or dev tools, like here). I wouldn't want to introduce it as a runtime dependency though.

cpeel (approved):
I thought about seeing if we should move this into a `SETUP/tools/` dir but that whole directory is a hot mess of stuff anyway.

chrismiceli:
> Well, #1130 is now a Python script too -- I rewrote it after it went way past my 'shell script complexity threshold'. I'd definitely prefer to keep #1130 as a Python script because there are a few details relating to output buffering and HTTP request and cookie handling that you need to get right for the smoke tests to work reliably. I've figured out how to do them with Python's `subprocess` and `urllib` and I don't really trust PHP not to 'helpfully' do the wrong thing behind our backs.
> 
> I'd strongly prefer not to do this tool as a shell script for a few reasons
> 
>     * the most immediate one is my gut instinct that cumulatively it's above the complexity limit of what I like to do in a shell script: recursion, hash tables, string processing, regexps. All possible, yes, but easy to make mistakes.
> 
>     * the string processing and regexp handling would get messier.
> 
>     * it relies on a dict to represent the dependencies and bash only got hash tables in v4.0, and macOS is never going to move past v3.2 because that's the last version under GPLv2.
> 
> 
> I could rewrite it as PHP, although I think it would be a bit longer and a bit uglier.
> 
> Note also that this is just a tool for developer curiosity/convenience. It's not used in CI, installation, or production, so to a certain extent it doesn't really matter if it breaks and nobody can maintain it.

This all makes sense, just wanted to make sure we make a conscious choice here and don't just fall back to what is comfortable. In the case we stick with python, I'd vote to have some documentation on Python usage in the repo or at least in the comments of this file that could include
1. Versions of python supported
2. How to invoke the script
3. Which OSes are expected to work (if not all of them)
Does it make sense to add any python linting tools to our CI for these scripts, or is that not a priority for these scripts?

bpfoley:
> > Well, #1130 is now a Python script too -- I rewrote it after it went way past my 'shell script complexity threshold'. I'd definitely prefer to keep #1130 as a Python script because there are a few details relating to output buffering and HTTP request and cookie handling that you need to get right for the smoke tests to work reliably. I've figured out how to do them with Python's `subprocess` and `urllib` and I don't really trust PHP not to 'helpfully' do the wrong thing behind our backs.
> > I'd strongly prefer not to do this tool as a shell script for a few reasons
> > ```
> > * the most immediate one is my gut instinct that cumulatively it's above the complexity limit of what I like to do in a shell script: recursion, hash tables, string processing, regexps. All possible, yes, but easy to make mistakes.
> > 
> > * the string processing and regexp handling would get messier.
> > 
> > * it relies on a dict to represent the dependencies and bash only got hash tables in v4.0, and macOS is never going to move past v3.2 because that's the last version under GPLv2.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I could rewrite it as PHP, although I think it would be a bit longer and a bit uglier.
> > Note also that this is just a tool for developer curiosity/convenience. It's not used in CI, installation, or production, so to a certain extent it doesn't really matter if it breaks and nobody can maintain it.
> 
> This all makes sense, just wanted to make sure we make a conscious choice here and don't just fall back to what is comfortable. In the case we stick with python, I'd vote to have some documentation on Python usage in the repo or at least in the comments of this file that could include
> 
>     1. Versions of python supported
> 
>     2. How to invoke the script
> 
>     3. Which OSes are expected to work (if not all of them)
>        Does it make sense to add any python linting tools to our CI for these scripts, or is that not a priority for these scripts?

Thanks. I did make a conscious choice, but admittedly I didn't articulate it, so you couldn't have know.

I've updated the script docstring to point out that you can run it with `--help` to show the usage -- [argparse](https://docs.python.org/3/library/argparse.html) for the win! --  and added a comment on the supported Python versions.

It definitely works with Python 3.8.x and later (on Ubuntu 20.04 and macOS), and I specifically avoided using the "walrus operator" introduced in 3.8 and there's nothing special in the script so I expect it should work with any Python 3.x on any platform.

We can certainly add Python linting tools, but I had other higher impact linting work I wanted to get done first (see, e.g. #1150), and I didn't want to put additional effort into adding a Python linter until I knew the tool would be accepted.

PR #1149, opened 2024-03-18
Enforce minimum username length for new users

cpeel:
phpBB has a minimum username length of 3, but we're not checking for that. The GM has requested a longer minimum. I'm lobbying for 4 but will await what @lhamilton1 decides on as the final number.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/enforce-min-username-len/

### DIFF
--- a/pinc/User.inc
+++ b/pinc/User.inc
@@ -542,8 +542,10 @@ $valid_username_chars_statement_for_elsewhere = _("(Valid characters are: a-z A-
  */
 function check_username($username, $registering = false)
 {
-    $username_max_len = 25;
     // This is the length of the 'username' field in the 'users' table.
+    $username_max_len = 25;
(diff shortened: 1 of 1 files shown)


srjfoo (approved):

pinc/User.inc
```diff
@@ -555,6 +557,12 @@ function check_username($username, $registering = false)
         return $error;
     }
 
+    // For new registrations, prevent too-short usernames
+    if ($registering && strlen($username) < $username_min_len) {
+        $error = sprintf(_("Your User Name is too short.<br>(The minimum is %s characters.)"), $username_min_len);
```
bpfoley, line 562: Not that it makes a difference to non-strict PHP, but shouldn't the `%s` here (and above) be `%d`?

bpfoley (approved):

srjfoo (approved):

PR #1150, opened 2024-03-18
Add Shellcheck shell script linting in CI

bpfoley:


SETUP/locale/compile.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-for po_file in `find . -name '*.po'`; do
-    mo_file=`echo $po_file | sed 's/.po/.mo/'`
-    echo Compiling $po_file to $mo_file
-    msgfmt $po_file -o $mo_file
-done
+while IFS= read -rd '' po_file; do
```
cpeel, line 3: Can we not with the IFS and really obscure bashisms here please? For loops are simple to read and understand.

SETUP/generate_css_from_less.sh
```diff
@@ -21,42 +18,36 @@ else
...
         fi
-        rm $new_css_file
     fi
 }
 
 # run over every .less file in our style directory
-for file in $(find $STYLE_DIR -name "*.less"); do
-    run_less $file
-done
+while IFS= read -rd '' file; do
```
cpeel, line 49: Again no IFS please.

SETUP/tests/ci_configuration.sh
```diff
@@ -100,23 +104,52 @@ TAG=master
 GROUP=users
 SHIFT_TO_LIVE=yes
 
-_PHP_CLI_EXECUTABLE=`which php`
```
cpeel, line 103: For this file I say we just use `command -v php` directly. Ditto `_URL_DUMP_PROGRAM` -- we control the CI env so we should just set that to `command -v wget` (or `curl` -- whichever we have in the Action) and remove the conditional bits.

cpeel (commented):
We're getting fancy now -- shell linting!

SETUP/locale/compile.sh
bpfoley: Yeah, for loops are simple to read, but they're also unsafe, which is kinda the problem that [SC2044](https://www.shellcheck.net/wiki/SC2044) is warning us about.

They break if you have arg separators or any metacharacters in any of your paths. Also `for f in $(find ...)` breaks if `find` returns a lot of output and exceeds your OS's command line arg limits, which is completely academic in our case, given `ARG_MAX` is typically >1MB these days, but it can be a problem. 

See also the long discussion on https://unix.stackexchange.com/a/9499

bpfoley: For some of these cases we can get away with `find ... -exec sh -c 'shell commands'`: the overhead of starting a subshell is negligible, and we don't need to carry any variable/functions into the subshell. I'll go back and see the places I can do that simplification.

SETUP/tests/ci_configuration.sh
bpfoley: OK. Will do.

bpfoley:
> We're getting fancy now -- shell linting!

Chris's suggestion! And definitely a valuable one. These scripts are load bearing, and a bit fragile at the moment.

SETUP/locale/compile.sh
cpeel: Yes, but those are edgecases. The input to all of these scripts is largely well know. I'd rather kill SC2044 and keep the for loops because they are so much more readable. All of the fixes just make it so much more difficult to read and maintain.

bpfoley: OK, done. The latest patchset goes back to suppressing the warning and using `for` loops.

SETUP/mk_loc_lang_names.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
```
cpeel, line 1: We should just delete this. According to git log it's been in our codebase for 14 years and never used.

bpfoley: `rm`-inating!

cpeel (approved):

chrismiceli (approved):
Shell scripting is almost greek to me. I prefer PowerShell (ducks from tomatoes coming my way), but the changes seem logical to me.

PR #1152, opened 2024-03-19
Remove miscellaneous old image files not referenced in the source

bpfoley:
At least, as best I can tell.

I tried searching in a variety of ways for filenames that might be composed from PHP variables, but didn't find these...

Sandbox at https://www.pgdp.org/~bfoley/c.branch/image-ectomy/


cpeel (approved):
re: remaining `tools/project_manager/wbf_images/*` images -- looks like those are used in the RFM. Feels like we could replace those with FontAwesome images and be done with them.

Musings not for this PR.

srjfoo (approved):
`greek5.gif` and `greek6.gif` are referenced from the dpwiki from this page: Greek/The Greek Alphabet, and from the French and Spanish translations of same. Both should be moved into the wiki, and the relevant pages updated before this is deployed to PROD.

I didn't find any of the others in the wiki, and I'm not inclined to worry about whether or not someone referenced them in the forums.

bpfoley:
> `greek5.gif` and `greek6.gif` are referenced from the dpwiki from this page: Greek/The Greek Alphabet, and from the French and Spanish translations of same. Both should be moved into the wiki, and the relevant pages updated before this is deployed to PROD.
> 
> I didn't find any of the others in the wiki, and I'm not inclined to worry about whether or not someone referenced them in the forums.

Well noticed. I've uploaded these two files to the wiki and updated the links on the 3 pages that used them to point at the wiki images.

PR #1155, opened 2024-03-21
Replace hard-coded faq/*-referenced URLs

cpeel:
Replace the hard-coded `faq/*` URLs with the ones that take the user directly to the wiki rather than relying on the redirects inside the `faq/*` pages to redirect them there. This is the first step in removing the the `faq/*` files that have been moved to the wiki.

Sandbox at https://www.pgdp.org/~cpeel/c.branch/replace-hardcoded-faq-urls/

### DIFF
--- a/faq/prooffacehelp.php
+++ b/faq/prooffacehelp.php
@@ -402,6 +402,8 @@ if ($i_type == 0) {
     echo "</dl>\n";
     echo "</form>\n";
 } else {
+    $faq_url = get_faq_url("ProoferFAQ.php");
+    $guidelines_url = get_faq_url("proofreading_guidelines.php");
     ?>
 
 <DIV ALIGN="CENTER"><TABLE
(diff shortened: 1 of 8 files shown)


cpeel:
Force-pushed after resolving a php formatting issue found by the linter.

chrismiceli (approved):

faq/prooffacehelp.php
```diff
@@ -402,6 +402,8 @@ interface, click on the \"Quit\" button.
     echo "</dl>\n";
     echo "</form>\n";
 } else {
+    $faq_url = get_faq_url("ProoferFAQ.php");
+    $guidleines_url = get_faq_url("proofreading_guidelines.php");
```
bpfoley, line 406: should be `$guidelines_url` :)

bpfoley (commented):
Looks like a good move, but if I've understood correctly, you've missed some:

- send_welcome_mail in pinc/new_user_mails.inc
- ProofreadingToolbox::draw_tools in pinc/ProofreadingToolbox.inc
- pinc/simple_proof_text.inc
- tools/proofers/round.php
- tools/proofers/srchrep.php

cpeel:
> Looks like a good move, but if I've understood correctly, you've missed some:
>     * send_welcome_mail in pinc/new_user_mails.inc

The PDF file is still present where it's referenced as it's not in the wiki.

>     * pinc/simple_proof_text.inc
>     * tools/proofers/round.php

These are addressed in https://github.com/DistributedProofreaders/dproofreaders/pull/1151

>     * ProofreadingToolbox::draw_tools in pinc/ProofreadingToolbox.inc
>     * tools/proofers/srchrep.php

`prooffacehelp.php` is staying in `faq/` so these are ok as-is.

faq/prooffacehelp.php
cpeel: Oops - fixed!

cpeel:
Force-pushed a commit to fix the spelling error Brian found.

70ray (approved):

srjfoo (approved):

PR #1157, opened 2024-03-21
Delete unlinked FAQ files

cpeel:
These aren't linked from anywhere and don't appear to be used by the code itself. The Word macro I know can go and I suspect the PDF can too, but am hoping @srjfoo can help determine that.


chrismiceli (approved):

srjfoo (approved):
the pdf looks like a Dutch translation of the (combined P&F) guidelines dating from 2004. I don't think we need to keep it.

PR #1159, opened 2024-03-26
Disable autocapitalization of username input fields

bpfoley:
The autocapitalize attribute is standard HTML5, but the default behavior when it isn't explicitly set differs across browsers.

In particular we have some users using iPads, and in Safari, autocapitalize is on by default.

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autocapitalize#usage_notes

Support goes back to iOS Safari 5 (2012); Chrome 43 (2015); Opera 30 (2015); but not until Firefox 111 (2023). However the Firefox non-support is harmless as it defaults to autocapitalize off anyway.

### DIFF
--- a/accounts/activate.php
+++ b/accounts/activate.php
@@ -159,7 +159,7 @@ $csrf_token
 <table class='themed theme_striped' style='width: auto'>
 <tr>
   <th><label for='loginform-userNM'>" . _("ID") . "</label></th>
-  <td><input type='text' id='loginform-userNM' name='userNM' value='".attr_safe($user->username)."' required></td>
+  <td><input type='text' id='loginform-userNM' name='userNM' value='".attr_safe($user->username)."' autocapitalize='none' required></td>
 </tr>
 <tr>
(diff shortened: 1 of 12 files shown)


chrismiceli (approved):
any `textarea`s that could benefit as well?

cpeel (approved):

bpfoley:
> any `textarea`s that could benefit as well?

Good question. I think there are probably a few -- but the only one that sticks out as an obvious candidate is `echo_proofing_textarea`.

srjfoo (approved):
Tested on both iPad and computer. Even with autocap on on the computer, I didn't get any autocap problems.

On the iPad, however, the one place I saw autocap happening was when registering a new account. I tested the other scripts that were touched, but none of them did the autocap thing. The branch did fix the autocap for registration. 👍 

That doesn't mean it won't happen on other browsers/systems in the other places, just that I was unable to duplicate it being a problem except for registering a new account. Autocap stayed on for Real Name, but switched off for username.

lhamilton1:
I approve. Thanks!

PR #1162, opened 2024-03-27
Add smoketests for all the tools/proofers/*.php pages.

bpfoley:
Add a placeholder project directory to allow some tests to work and re-enable some other pages that also require a project dir.


SETUP/smoketests/pageload_smoketest.py
```diff
@@ -575,7 +609,7 @@ def main() -> int:
         if args.verbose:
             print(data.decode())
         if status not in expect_status or test_failed(logs):
-            print(f'Status: {status}')
+            print(f'Status: {status} (expected {epected_status})')
```
cpeel, line 612:
```suggestion
            print(f'Status: {status} (expected {expect_status})')
```

.github/workflows/ci.yml
```diff
@@ -120,6 +120,12 @@ jobs:
       - name: Add smoketest DB entries
         run: |
             mysql -uroot -proot < SETUP/smoketests/test_tables.sql
+      - name: Create smoketest project dirs
+        run: |
+            mkdir -p $HOME/projects/projectID5e23a810ef693
```
cpeel, line 125: Should we put this inside a "smoketest_prep.sh" script and call it here rather than put the logic in the CI itself?

SETUP/smoketests/pageload_smoketest.py
bpfoley: Time to get that linter :)

.github/workflows/ci.yml
bpfoley: Done.

cpeel (approved):

PR #1163, opened 2024-03-28
Use project object rather than projectid to construct LPage

70ray:
This should be useful for the API.

Another new project is not now made at line 450 but I think this should not cause any problem. The only reason to make another new project is if we have done something which changes the projects table so the project object could be invalid. The only thing that does this here is project_transition() but that occurs later.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/LP_project

### DIFF
--- a/pinc/LPage.inc
+++ b/pinc/LPage.inc
@@ -146,7 +146,7 @@ function get_available_page($projectid, $proj_state, $pguser, &$err)
         $err = $exception->getMessage();
         return null;
     }
-    $lpage = new LPage($project->projectid, $imagefile, $state, 0);
+    $lpage = new LPage($project, $imagefile, $state, 0);
 
     $lpage->checkout($pguser);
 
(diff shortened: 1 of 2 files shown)


pinc/LPage.inc
```diff
@@ -283,9 +283,10 @@ class RoundInfo
  */
 class LPage
 {
-    public function __construct($projectid, $imagefile, $page_state, $reverting_to_orig)
+    public function __construct($project, $imagefile, $page_state, $reverting_to_orig)
     {
-        $this->projectid = $projectid;
+        $this->project = $project;
+        $this->projectid = $project->projectid;
```
chrismiceli, line 289: would it make sense to just remove this property so we don't have to worry about keeping the `project` and `projectid` in sync?

chrismiceli (approved):

pinc/LPage.inc
70ray: I suppose it would.

70ray: The reasons for not doing that are that it could be slower (insignificant) and needs more changes, but there seems little risk of breaking anything without noticing so I'll do that.

70ray: PPage uses LPage->projectid. I don't want to change PPage now so don't do this now. If in the future we use proofreading API and don't use PPage we can do it then.

cpeel (approved):

srjfoo (approved):

PR #1165, opened 2024-03-28
api/v1_projects.inc: Return an error if client requests invalid field

bpfoley:
As suggested by srjfoo. I agree with her that it's better to fail loudly rather than silent ignore unrecognized fields. If nothing else, it helps when a dev makes typos in client code.

I needed to do some code reorg to avoid repeating a subset of SQL column names and API field names for a third time.

Sandbox at https://www.pgdp.org/~bfoley/c.branch/invalid-field but it doesn't touch any non-API code.

Example call:
```
curl -s -X GET "$ROOT/api/v1/projects&field[]=pages_total&field[]=no_such_field" \
    -H "X-API-KEY: $INSERT_YOUR_API_KEY_KEY"
```

### DIFF
--- a/api/v1_projects.inc
+++ b/api/v1_projects.inc
@@ -12,36 +12,39 @@ include_once("api_common.inc");
 //===========================================================================
 // projects/
 
+/**
+ * Get an array of the fields to be returned to the client.
+ * @param ?array<string,string> $query_params the HTTP query
+ * @param ?Project $project
+ *   If we're querying a single project, what is it? (Certain fields
+ *   aren't queryable or viewable by most users.)
+ * @return ?array<string>
+ * @throws InvalidValue if any of the filter arguments aren't recognized
+ */
+function get_return_fields(?array $query_params, ?Project $project)
+{
+    $return_fields = array_get_as_array($query_params, "field", null);
+    if (is_null($return_fields)) {
+        return null;
+    }
+    $valid_render_fields = array_keys(get_project_fields_with_attr(null, $project));
+    $invalid_fields = array_diff($return_fields, $valid_render_fields);
+    if (!empty($invalid_fields)) {
+        throw new InvalidValue("Invalid filter args: " . implode(", ", $invalid_fields));
+    }
+    return $return_fields;
+}
+
 function api_v1_projects($method, $data, $query_params)
 {
     // set which fields are queryable and their column names
-    $valid_fields = [
-        "projectid" => "projectid",
-        "state" => "state",
-        "title" => "nameofwork",
-        "author" => "authorsname",
-        "languages" => "language",
-        "genre" => "genre",
-        "difficulty" => "difficulty",
-        "special_day" => "special_day",
-        "project_manager" => "username",
-        "pg_ebook_number" => "postednum",
-        "pages_available" => "n_pages_available",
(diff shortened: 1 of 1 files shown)


cpeel (approved):
Code looks good to me but I haven't tested it.

api/v1_projects.inc
```diff
@@ -139,30 +146,56 @@ function api_v1_projects($method, $data, $query_params)
 //---------------------------------------------------------------------------
 // projects/:projectid
 
+function get_project_fields_with_attr(?string $attr = null): array
+{
+    // A list of all the fields that can be rendered by render_project_json
+    // and the corresponding name of the SQL column.
+    // In addition, fields with the "queryable" attribute can be used
+    // as query term for api/v1/projects, and fields with the "updatable"
+    // attribute can be set by a POST request.
```
srjfoo, line 155: Not 100% sure where to put this comment, but I wanted to ask before stopping (still need to do some testing).

Looking at the attributes below -- for example, clearance is queryable for those who have access to do so. `postednum` is queryable (or should be, since you can search on that field in project search (though maybe I'm missing the intent of for purposes of the API).

bpfoley: Currently on pgdp.org, you can query by `postedenum` (the query name is `pg_ebook_number`), and that remains the case.
 Eg:


```
curl -s -X GET 'https://www.pgdp.org/~bfoley/c.branch/invalid-field/api/v1/projects&pg_ebook_number=76543&field[]=title' -H 'Accept: application/json' -H 'X-API-KEY: REDACTED'
[
    {
        "title": "Test CTF generation"
    }
]
```

You're right that I missed the fact that we add  `clearance` to `valid_fields` in `api_v1_projects()` but neglected to do it in `api_v1_project()` though. I'll move that into `get_project_fields_with_attr`.

srjfoo: Did the sandbox get updated? Queries on `clearaince` fail (i.e. provide no data), even for a squirrel (tested for both squirrel and non-squirrel-non-PF access privileges).

Query with an invalid field correctly throws an error.

Should `api/v1/projects/<projectid>`also be throwing an error if invalid fields are specified?

bpfoley: OK. I've updated this now, and moved the (slightly odd at first sight) logic for handling the clearance field into one place with some extra comments.

I've tried the following 4 with users who do and don't have site admin/proj facilitator permissions. As always, these URLs are passed to curl with `curl -s -X GET $URL -H "Accept: application/json" -H "X-API-KEY: REDACTED"`
1. Search for projects and only returning title & clearance: https://www.pgdp.org/~bfoley/c.branch/invalid-field/api/v1/projects&pg_ebook_number=76543&field[]=title&field[]=title&field[]=clearance
2. Search by clearance:  https://www.pgdp.org/~bfoley/c.branch/invalid-field/api/v1/projects&clearance=okokok&field[]=title&field[]=clearance
3. Get project info: https://www.pgdp.org/~bfoley/c.branch/invalid-field/api/v1/projects/projectID62395e044016d
4. Get project info with clearance: https://www.pgdp.org/~bfoley/c.branch/invalid-field/api/v1/projects/projectID62395e044016d&field[]=clearance

srjfoo (approved):

PR #1173, opened 2024-04-04
Collapse Activities in navbar based on avail space (& preference)

cpeel:
Introduce the idea of drop-down menu options to group like things together and make better use of the top header space. Much to my chagrin we use the new menu based on a user preference; if the preference is set to "auto" (the default) dynamically use the menu based on the width of the window.

This is available for testing in https://www.pgdp.org/~cpeel/c.branch/refine-navbar/

See also the discussion in the [Proposed changes to the navigation bar - 2024 edition](https://www.pgdp.net/phpBB3/viewtopic.php?f=4&t=81413) forum.

Testing notes: In addition to confirming this works in all our minimum browsers, we should make sure this looks good in all themes -- including having an unread message in the forums which will add an "unread messages" number to the new Inbox icon -- and on various mobile devices.

### DIFF
--- a/pinc/theme.inc
+++ b/pinc/theme.inc
@@ -91,7 +91,7 @@ function output_footer($nameofpage, $show_statsbar = true)
     echo "<p>\n";
     echo "<small>\n";
 
-    echo "&copy;$site_name";
+    echo "&copy;" . html_safe($site_name);
 
     echo "<span id='buildtime'>";
     echo " &middot; ";
@@ -119,30 +119,96 @@ function html_logobar()
 {
     global $code_url, $site_name;
     global $maintenance, $testing, $alert_message;
+    global $blog_url, $wiki_url;
 
     echo "\n<div id='header'>\n";
-    echo "<a href='$code_url'>\n";
+
     echo "<div id='logo'>\n";
-    echo "<img src='$code_url/graphics/dp-logo.svg' alt='$site_name'>\n";
-    echo "</div>\n";
+    echo "<a href='$code_url'>\n";
+    echo "<img id='logo-img' src='$code_url/graphics/dp-logo.svg' alt='" . attr_safe($site_name) . "'>\n";
     echo "</a>\n";
+    echo "</div>\n"; // logo
 
-    echo "<div id='titles-preserved'><p>\n";
+    echo "<div id='logo-right'>";
+    $user = User::load_current();
+    if (!$user) {
+        // show number of books posted to PG
+        // (takes into account that some books are processed as several projects)
+        echo "<div id='titles-preserved'>\n";
+        echo "<span id='x-preserved'>";
+        echo sprintf(
+            _('%s titles preserved for the world!'),
+            number_format(total_completed_projects())
+        );
+        echo "</span>\n";
+        echo "<span id='month-preserved-summary'>";
+        show_completed_projects(true);
+        echo "</span>";
+        echo "</div>\n";
+    } else {
+        $icon_links = [];
+        if (get_url_for_forum()) {
+            $icon_links[] = [
+                "name" => _("Inbox"),
+                "class" => "fas fa-envelope fa-2x",
+                "url" => get_url_for_inbox(),
+            ];
 
-    // show number of books posted to PG
-    // (takes into account that some books are processed as several projects)
-    echo "<span id='x-preserved'>";
-    echo sprintf(
-        _('%s titles preserved for the world!'),
-        number_format(total_completed_projects())
-    );
-    echo "</span>\n";
-    echo "<span id='month-preserved-summary'>";
-    show_completed_projects(true);
-    echo "</span>";
+            $icon_links[] = [
+                "name" => _("Forums"),
+                "class" => "fas fa-comments fa-2x",
+                "url" => get_url_for_forum(),
+            ];
+        }
+        if (!empty($wiki_url)) {
+            $icon_links[] = [
+                "name" => _("Wiki"),
+                "class" => "fas fa-scroll fa-2x",
+                "url" => $wiki_url,
+            ];
+        }
+        if (!empty($blog_url)) {
+            $icon_links[] = [
+                "name" => _("Blog"),
+                "class" => "fab fa-wordpress fa-2x",
+                "url" => $blog_url,
+            ];
+        }
+        $icon_links[] = [
+            "name" => _("Help"),
+            "class" => "fas fa-question fa-2x",
+            "url" => get_faq_url("faq_central.php"),
+        ];
+        $links = [];
+        foreach ($icon_links as $data) {
+            $name = attr_safe($data["name"]);
+            $class = $data["class"];
+            $url = $data["url"] ?? "";
+            $text = "<div class='icon-menu-item'>";
+            if ($url) {
+                $text .= "<a href='$url' aria-label='$name'>";
+            }
+            $text .= "<i aria-hidden='true' class='$class' title='$name'></i>";
+            if ($url) {
+                $text .= "</a>";
+            }
+            if ($data["name"] == _("Inbox")) {
+                $numofPMs = get_number_of_unread_messages($user->username);
+                if ($numofPMs > 0) {
+                    $text .= "<span class='inbox-badge'>$numofPMs</span>";
+                }
+            }
+            $text .= "<br><span class='icon-menu-item-name'>$name</span>";
+            $text .= "</div>";
+            $links[] = $text;
+        }
 
-    echo "</p></div>\n";
-    echo "</div>\n";
+        echo "<div id='icon-menu'>";
+        echo join("", $links);
+        echo "</div>"; // icon-menu
+    }
+    echo "</div>"; // logo-right
+    echo "</div>\n"; // header
 
     html_navbar();
 
@@ -174,7 +240,7 @@ function html_logobar()
  */
 function html_navbar()
 {
-    global $code_url, $site_abbreviation, $blog_url, $wiki_url;
+    global $code_url, $blog_url, $wiki_url;
 
     echo "<div id='navbar-outer'>";
     echo "<div id='navbar'>";
@@ -183,13 +249,16 @@ function html_navbar()
     $user = User::load_current();
     if (!$user) {
         $links = [];
-        $links[] = ['text' => $site_abbreviation, 'url' => "default.php"];
         if (!empty($blog_url)) {
             $links[] = ['text' => _("Blog"), 'url' => $blog_url];
+        }
+        if (!empty($wiki_url)) {
+            $links[] = ['text' => _("Wiki"), 'url' => $wiki_url];
         }
         if (get_url_for_forum()) {
             $links[] = ['text' => _("Forums"), 'url' => get_url_for_forum()];
         }
+        $links[] = ['text' => _('Help'), 'url' => get_faq_url("faq_central.php")];
         echo "<span id='navbar-left'>";
         echo headerbar_text_array($links);
         echo "</span>\n";
@@ -230,7 +299,6 @@ function html_navbar()
         $links = [];
         $links[] = ['text' => $login_form];
         $links[] = ['text' => _('Register'), 'url' => "accounts/addproofer.php"];
-        $links[] = ['text' => _('Help'), 'url' => get_faq_url("faq_central.php")];
         echo headerbar_text_array($links);
         echo "</div>\n";
     } else {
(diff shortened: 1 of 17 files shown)


pinc/theme.inc
```diff
@@ -119,30 +119,96 @@ function html_logobar()
...
+    global $blog_url, $wiki_url;
 
     echo "\n<div id='header'>\n";
-    echo "<a href='$code_url'>\n";
+
     echo "<div id='logo'>\n";
-    echo "<img src='$code_url/graphics/dp-logo.svg' alt='$site_name'>\n";
-    echo "</div>\n";
+    echo "<a href='$code_url'>\n";
+    echo "<img id='logo-img' src='$code_url/graphics/dp-logo.svg' alt='$site_name'>\n";
```
bpfoley, line 128: I wonder do we want to escape $site_name? It could contain single quotes, e.g. "Brian's Proofing Emporium".

pinc/theme.inc
```diff
@@ -119,30 +119,96 @@ function html_logobar()
...
+    echo "</div>\n"; // logo
 
-    echo "<div id='titles-preserved'><p>\n";
-
-    // show number of books posted to PG
-    // (takes into account that some books are processed as several projects)
-    echo "<span id='x-preserved'>";
-    echo sprintf(
-        _('%s titles preserved for the world!'),
-        number_format(total_completed_projects())
```
bpfoley, line 137: Saves an extra trip to the DB. Woo.

pinc/theme.inc
```diff
@@ -119,30 +119,96 @@ function html_logobar()
...
+        echo "<span id='month-preserved-summary'>";
+        show_completed_projects(true);
+        echo "</span>";
+        echo "</div>\n";
+    } else {
+        $icon_links = [];
+        if (get_url_for_forum()) {
+            $icon_links[] = [
+                "name" => _("Inbox"),
+                "class" => "fas fa-envelope fa-2x",
```
bpfoley, line 153: Perhaps pull "fa-2x" out as something that's common to all the icons?

pinc/theme.inc
```diff
@@ -230,29 +299,59 @@ function html_navbar()
...
 
         $links = [];
         $links[] = ['text' => _("Activity Hub"), 'url' => "activity_hub.php"];
 
-        // use white-space: nowrap; to keep the rounds within []s together
-        $activities_string = "<span style='white-space: nowrap;'>[ ";
-        if (get_pages_proofed_maybe_simulated() < 100) {
-            $activities_string .= "<span class='text'>" . _("Activities") . ": </span>";
+        // We create 2 different ways of showing the activities, one with them
+        // all out in one string and one with with them collapsed behind a menu.
```
bpfoley, line 314: s/all out/all visible/?

pinc/theme.inc
```diff
@@ -230,29 +299,59 @@ function html_navbar()
...
+        // Furthermore, never collapse the menu for new users or users with
+        // fewer than 4 activities.
+        $activity_links = get_activity_links();
+
+        $is_new_user = get_pages_proofed_maybe_simulated() < 100;
+
+        // Calculate style-level CSS that will override the classes
+        // 2 - never collapse
+        if ($user->navbar_activity_menu == 2 || $is_new_user || count($activity_links) <= 3) {
+            $string_style = "style='display: inline;'";
```
bpfoley, line 328: You can drop all the trailing semicolons in your HTML `style='foo: bar;'` attributes.

pinc/theme.inc
```diff
@@ -270,43 +369,52 @@ function html_navbar()
...
-        if (get_url_for_forum()) {
-            $links[] = ['text' => _("Forums"), 'url' => get_url_for_forum()];
-        }
+        echo headerbar_text_array($links);
+        echo "  &#183; ";
 
-        if (!empty($blog_url)) {
-            $links[] = ['text' => _("Blog"), 'url' => $blog_url];
-        }
+        $account_links = [];
```
bpfoley, line 377: Maybe just make this a single assign rather than appending? 
```
$account_links = [ ..., ... , ... ];
```

pinc/theme.inc
```diff
@@ -270,43 +369,52 @@ function html_navbar()
...
+{
+    $menu_style = "";
+    $menu_contents_style = "";
+    if ($align_right) {
+        $menu_style = "style='padding-right: 0.5em;'";
+        $menu_contents_style = "style='right: 0;'";
+    }
+
+    $menu_id = 'menu' . md5($menu_title);
+    $output = [];
```
bpfoley, line 401: Again, perhaps single assign here rather than repeated append. Makes it a little less noisy.

scripts/navbar_menu.js
```diff
@@ -0,0 +1,33 @@
...
+document.addEventListener("click", function (event) {
+    let inMenu = false;
+    document.querySelectorAll(".menu").forEach(element => {
+        if (element.contains(event.target) || element == event.target) {
+            inMenu = true;
+        }
+    });
+
+    if (!inMenu) {
+        document.querySelectorAll(".menu-contents").forEach(element => {
```
bpfoley, line 25: This duplicates the actions of `toggleMenu`. Can we reword this as something like
```
document.querySelector(".menu").forEach(element => toggleMenu(element.id));
```

styles/layout.less
```diff
@@ -219,13 +220,43 @@ header, #header {
...
+        .center-align;
+        text-decoration: none;
+        .sans-serif;
+        font-size: 0.7em;
+        color: @heading-color;
+        .unicolor-link(@heading-color);
+        .icon-menu-item-name {
+            font-size: 0.9em;
+        }
+        .inbox-badge {
```
bpfoley, line 247: I feel that the badge is slightly lower than looks good. Any chance we can push it up a little bit?

styles/layout.less
```diff
@@ -219,13 +220,43 @@ header, #header {
...
+          background-color: @text-error;
+          border-radius: 2px;
+          color: @page-background;
+
+          padding: 1px 3px;
+          font-size: 10px;
```
bpfoley, line 253: Should `font-size` be expressed in `em`s rather than as an absolute `px` size?

pinc/theme.inc
```diff
@@ -230,29 +299,59 @@ function html_navbar()
...
         // left side
         echo "<span id='navbar-left'>";
-        echo headerbar_text_array([['text' => $site_abbreviation, 'url' => "default.php"]]);
-
-        echo " | ";
 
         $links = [];
         $links[] = ['text' => _("Activity Hub"), 'url' => "activity_hub.php"];
 
-        // use white-space: nowrap; to keep the rounds within []s together
```
bpfoley, line 248: Note to self: `white-space: nowrap` has moved into the CSS files.

bpfoley (approved):

scripts/navbar_menu.js
cpeel: Ah, but it doesn't actually. `toggleMenu()` just toggles. If we called `toggleMenu()` it would close any open menus and open any closed menus, it wouldn't close everything. If there were 1 and only 1 menu they would be the same, but with more than 1 we can't use it.

pinc/theme.inc
cpeel: Fair question, but to me the whole class attribute feels like a unit here and I'm wary of breaking that into two pieces, one of which is stashed out in a stylesheet.

pinc/theme.inc
cpeel: Noted, but I think I'll leave them (makes it easier to add new ones if necessary, move it to stylesheet if necessary, and I'm lazy 😁 ).

cpeel (commented):
Will address (eg make code changes) on your other items later today. Thanks for the review!

scripts/navbar_menu.js
bpfoley: Ah. Yeah, I didn't read closely enough. Good point!

scripts/navbar_menu.js
```diff
@@ -0,0 +1,33 @@
...
+            element.classList.add('invisible');
+            element.classList.add('transparent');
```
chrismiceli, line 27: does this work for things like preventing tabbing to these elements?

windymilla:
Tested in Chrome Version 123.0.6312.105 (Official Build) (64-bit) on Windows:
All behaved as expected, including with extremely small/large font sizes, and with all 3 prefs settings.

Love the horizontal drop-down, and the little arrow works for both opening & closing the dropdown.

windymilla (approved):
It seems odd that there is now a pref that will have no effect for a new user. If a new user is exploring and changes that pref to Always or Never, perhaps that should be respected. I think it should only have the "new user" override when it is set to auto. For a new user "Auto" means "Always". Just a small thing

Congrats on tracking through all the places that needed edits to implement this!

pinc/theme.inc
cpeel: Good catch - fixed this and the other instance in this file.

pinc/theme.inc
cpeel: Oh yeah, that's much better. And obvious in retrospect.

pinc/theme.inc
cpeel: Yup, done.

scripts/navbar_menu.js
```diff
@@ -0,0 +1,33 @@
+/* exported toggleMenu */
+
+// Toggle a menu contents to visible and rotate the menu icon
+function toggleMenu(menuId) {
+    let menu = document.getElementById(menuId);
+    [menu.getElementsByClassName("menu-contents")].forEach(element => {
```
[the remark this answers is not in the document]
cpeel: I admit I don't know the difference. What I have works but I was literally throwing shit at the wall. Would you mind showing me what you think this should look like? I'm not familiar with the `[...` syntax and don't know what it's called to look it up.

scripts/navbar_menu.js
cpeel: I don't know. In the browser how do you access something via tab that isn't an input? I'm using Firefox and clicking into the window and hitting tab isn't letting me navigate around.

cpeel (commented):
> It seems odd that there is now a pref that will have no effect for a new user. If a new user is exploring and changes that pref to Always or Never, perhaps that should be respected. I think it should only have the "new user" override when it is set to auto. For a new user "Auto" means "Always".

That's a really great point, thanks for bringing it up. I've updated so that we only override the collapsing in "auto" mode.

---

The last commit doesn't resolve all the comments, more to come.

scripts/navbar_menu.js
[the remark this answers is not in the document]
chrismiceli: I was actually slightly wrong, you don't even need the `...` in this case.

The return type from getElementsByClassName does not have a `forEach` method as it is not an Array type for some strange reason. That means if you want to go through each element returned, you traditionally have to do a silly old for loop:
```
const menuContents = menu.getElementsByClassName('menu-contents');
for (let i = 0; i < menuContents.length; i++)
{
  const myElement = menuContents.item(i);
  // ...
}
```

Obviously this is kind of verbose. You can however use its [iterator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/iterator) properties to loop through
```
const menuContentsIterator = menu.getElementsByClassName('menu-contents')[Symbol.iterator]().next();
while(menuContentsIterator.done === false)
{
  const myElement = menuContentsIterator.value;
  // ...
  menuContentsIterator = menuContentsIterator.next();
}
```
Now that is not much better, so there is some syntatic sugar to use the iterator with a for of loop:
```
for (const myElement of menu.getElementsByClassName('menu-contents'))
{
  // ...
}
```

scripts/navbar_menu.js
chrismiceli: I was referring to child elements (anchors), but it seems you set them to `visibility: hidden`, so that should make them not tabbable.

scripts/navbar_menu.js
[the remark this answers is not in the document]
cpeel: That's perfect, Chris - thank you!

cpeel:
Latest commits do the following:
* Updates how we're iterating over the elements in the JS (thanks Chris!)
* Changes the icon colors back to the navbar background (except for Charcoal) -- the text remains the header color which is every so slightly darker (in most cases) but isn't black.
* Addresses other PR requests, including moving the unread messages slightly up and to the right from where it was.

Finally, it includes a separate commit that will allow someone to use the `?numofpages=` parameter on TEST to set it to any value, not just lower like it does on PROD.

I'll rebase the fixup commits after we've finished iterating over the changes in the PR & testing.

windymilla:
> Finally, it includes a separate commit that will allow someone to use the `?numofpages=` parameter on TEST to set it to any value, not just lower like it does on PROD.

Thanks - just returned a page to the rounds so I'm back down to 99 P1 pages on TEST, and confirmed `numofpages` param worked correctly for me.

70ray (approved):

chrismiceli (approved):

cpeel:
Some browser testing results, all good.

# Our minimum supported versions
Chrome 57 on Windows 10
Firefox 52 on Windows 10
Opera 44 on Windows 10
Safari 11 on High Sierra

# Latest versions
Chrome: 123.0 on macOS 12.7
Firefox 124.0.1 on macOS 12.7
Safari 17.4.1 on macOS 12.7
Opera 109 on Windows 10
Edge 123 on Windows 10
Edge 109 on Windows 8 (latest supported on that OS)

# Mobile
Firefox on Pixel 4a (vert & horiz)
Chrome on Pixel 4a (vert & horiz)

srjfoo (approved):

srjfoo:
### iPad running 17.3.1
All three of Safari, Chrome and Firefox look good.

cpeel:
Last push merely squashed the PR commits into each other. Will merge after CI passes.

PR #1181, opened 2024-04-11
Use xtext from vertilia/text to extract strings

cpeel:
xgettext has a nasty bug where it doesn't understand the PHP 7.3+ heredocs where the closing string is indented. This causes xgettext to skip the entire rest of the document. Replace it with xtext.

Because we use xtext in prod, not just during development, we need to require nikic/php-parser explicitly as a non-dev package.

We're able to use xtext like this because we pushed several feature enhancements up stream and the xtext developer has been superbly responsive to our requests!

Sandbox: https://www.pgdp.org/~cpeel/c.branch/fix-xgettext/

### DIFF
--- a/SETUP/INSTALL.md
+++ b/SETUP/INSTALL.md
@@ -115,8 +115,8 @@ Increasing the number of open files (`open_files_limit`) and the table cache
 increase these more than the MySQL daemon can open based on kernel limits.
 Consult the MySQL documentation for more information.
 
-### Install gettext and xgettext
-If you want to localize the site messages, install gettext and xgettext.
+### Install gettext
+If you want to localize the site messages, install gettext.
 
 ### Install aspell
 The code uses aspell for the spellchecker in the proofreading interface.
--- a/SETUP/configuration.sh
+++ b/SETUP/configuration.sh
@@ -522,17 +522,6 @@ if [ ! -x "$_PHP_CLI_EXECUTABLE" ]; then
     fi
 fi
 
-# The location of xgettext executable.
-_XGETTEXT_EXECUTABLE=
-if [ ! -x "$_XGETTEXT_EXECUTABLE" ]; then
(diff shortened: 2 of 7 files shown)


chrismiceli (approved):

windymilla (approved):

srjfoo:
When I regenerate the PO Template using `gettext` on TEST, I get a formatted file of comments and messages that is themed according to my preferences; example: 
```
#: accounts/activate.php:36
msgid "Activate account"
msgstr ""

#: accounts/activate.php:43
msgid "It appears that the account has already been activated."
msgstr ""
```
When I regenerate it in the sandbox, I get one long string, unthemed:
>  #: ./accounts/activate.php:36 msgid "Activate account" msgstr "" #: ./accounts/activate.php:43 msgid "It appears that the account has already been activated." msgstr "" #: ./accounts/activate.php:45 msgid "(Probably you just clicked the activation link more than once.)" msgstr ""

Once I merge it with a PO file, though, it looks fine. Just checking to make sure this isn't a problem that needs to be addressed.

I just realized again that the `xtext` header is missing a lot of the metadata at the top of the file. It looks like it's there in the PO files, though, so maybe not an issue.

cpeel:
Oh, I see. If you "View" the file via the web page the xtext version is rendered all in one line rather than a text file (at least for me in Firefox). I'll see if I can fix that and if not just remove the View link -- the Download link downloads the correctly-formatted file.

cpeel:
"View" fixed.

I also added a commit to only keep the last 10 uploaded message files per language. We have files in there from 2012!

After this PR looks good I'll rebase to squash a commit before getting it in.

srjfoo (approved):

cpeel:
Instead of keeping just the last 10 backup files, keep the last 6 months.

srjfoo (approved):

PR #1183, opened 2024-04-14
Update composer dependencies

cpeel:
Updates all of our package dependencies with minor version updates. Most of these are dev dependencies and if our CI passes that's good enough. I've checked the changelogs for the two non-dev dependency updates -- ezyang/htmlpurifier and phpmailer/phpmailer -- and there is nothing that gives me pause. Both of them are primarily to support PHP 8.3.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -112,20 +112,20 @@
         },
         {
             "name": "ezyang/htmlpurifier",
-            "version": "v4.16.0",
+            "version": "v4.17.0",
             "source": {
                 "type": "git",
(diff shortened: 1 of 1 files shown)


bpfoley (approved):
SGTM. I've been keeping an eye on the PHPStan and php-cs-fix release notes to see if anything exciting has been added, but nothing has yet. In particular, I'm waiting for the parallel processing PR for php-cs-fix to get finished, but it's not 100% there yet.

srjfoo (approved):

chrismiceli (approved):

70ray (approved):

PR #1184, opened 2024-04-14
Add new "site search" page

cpeel:
This adds a new "site search" page. Users can either click on the search icon in the header or use the `/` hotkey in any non-input field on a site page. This has been available for public preview & discussion in the [New site search feature](https://www.pgdp.net/phpBB3/viewtopic.php?f=4&t=81458) forum.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/site-search/

### DIFF
--- a/pinc/theme.inc
+++ b/pinc/theme.inc
@@ -179,15 +179,19 @@ function html_logobar()
             "class" => "fas fa-question fa-2x",
             "url" => get_faq_url("faq_central.php"),
         ];
+        $icon_links[] = [
+            "name" => _("Search"),
+            "class" => "fas fa-search fa-2x",
+            "onclick" => "showSiteSearch(); return false;",
+        ];
         $links = [];
         foreach ($icon_links as $data) {
             $name = attr_safe($data["name"]);
             $class = $data["class"];
             $url = $data["url"] ?? "";
+            $onclick = isset($data["onclick"]) ? "onclick='" . attr_safe($data["onclick"]) . "'" : "";
             $text = "<div class='icon-menu-item'>";
-            if ($url) {
-                $text .= "<a href='$url' aria-label='$name'>";
-            }
+            $text .= "<a href='$url' aria-label='$name' $onclick>";
             $text .= "<i aria-hidden='true' class='$class' title='$name'></i>";
             if ($data["name"] == _("Inbox")) {
                 $numofPMs = get_number_of_unread_messages($user->username);
@@ -195,10 +199,8 @@ function html_logobar()
                     $text .= "<span class='inbox-badge'>$numofPMs</span>";
                 }
             }
-            $text .= "<br><span class='icon-menu-item-name'>$name</span>";
-            if ($url) {
-                $text .= "</a>";
-            }
+            $text .= "<br><span class='icon-menu-item-name' $onclick>$name</span>";
+            $text .= "</a>";
             $text .= "</div>";
             $links[] = $text;
         }
@@ -206,6 +208,13 @@ function html_logobar()
         echo "<div id='icon-menu'>";
         echo join("", $links);
         echo "</div>"; // icon-menu
+
+        echo "<div id='search-menu' class='nodisp'>";
+        $placeholder = attr_safe(sprintf(_("Search for '%s' to see options"), "help"));
+        echo "<form action='$code_url/tools/site_search.php'>";
+        echo "<input id='search-menu-input' type='text' placeholder='$placeholder' name='q' autocapitalize='none'>&nbsp;<i aria-hidden='true' class='far fa-window-close' onclick='hideSiteSearch()'></i>";
+        echo "</form>";
+        echo "</div>"; // search-menu
     }
     echo "</div>"; // logo-right
     echo "</div>\n"; // header
--- a/tools/site_search.php
+++ b/tools/site_search.php
@@ -1,0 +1,352 @@
+<?php
+$relPath = "./../pinc/";
+include_once($relPath.'base.inc');
+include_once($relPath.'theme.inc');
+include_once($relPath.'Project.inc');
+include_once($relPath.'User.inc');
+
+require_login();
+
+$query = trim(array_get($_GET, "q", "help"));
+
+// "help" is just a string to get us here, but we don't want to keep it
+if ($query == "help") {
+    $query = "";
+}
+
+if ($query !== "") {
+    // try a jump search first, as we want to prefer the built-in jump
+    // search strings over usernames that might conflict
+    jump_search($query);
+
+    // try a smart search -- usernames can't contain ':' so these will
+    // never conflict
+    smart_search($query);
+
+    try {
+        // try a prefix search
+        prefix_search($query);
+
+        $message = sprintf(_("Not sure what to do with '%s', maybe try one of the prefixes below."), $query);
+    } catch (Exception $exception) {
+        $message = sprintf(_("The input for your prefix search was invalid: %s"), $exception->getMessage());
+    }
+}
+
+$title = _("Site Search");
+output_header($title, NO_STATSBAR);
+
+echo "<h1>$title</h1>";
+
+if (@$message) {
+    echo "<p class='warning'>" . html_safe($message) . "</p>";
+}
+
+echo "<form>";
+echo "<input type='text' name='q' value='" . attr_safe($query) . "'> ";
+echo "<input type='submit' value='" . attr_safe(_("Search")) . "'>";
+echo "</form>";
+
+echo "<p>" . _("Site search attempts to figure out what you're looking for and get you to it quickly.") . "</p>";
+
+echo "<ul>";
+echo "<li>" . _("Enter a projectID to go to the project's page.") . "</li>";
+echo "<li>" . _("Enter a username to go to the user's statistics page.") . "</li>";
+echo "<li>" . sprintf(
+    _("Enter an activity (%s) to go to the activity's page."),
+    surround_and_join(array_keys(get_activity_jumps()), "<kbd>", "</kbd>", ", ")
+) . "</li>";
+echo "</ul>";
+
+echo "<p>" . _("The following prefixes can be used to search across various categories.") . "</p>";
+
+echo "<ul>";
+foreach (get_prefix_searches() as $prefix => $data) {
+    echo "<li>";
+    echo "<kbd>$prefix:</kbd> - " . $data["desc"];
+    if (@$data["alias"]) {
+        echo " (" . sprintf(_("alias: %s"), "<kbd>" . $data["alias"] . ":</kbd>") . ")";
+    }
+    echo "</li>";
+}
+echo "</ul>";
+
+echo "<p>" . _("In addition, the following words can be used to jump to specific site pages.") . "</p>";
+
+echo "<ul>";
+foreach (get_jump_words() as $prefix => $data) {
+    echo "<li>";
+    echo "<kbd>$prefix</kbd> - " . $data["desc"];
+    if (@$data["alias"]) {
+        echo " (" . sprintf(_("alias: %s"), "<kbd>" . $data["alias"] . "</kbd>") . ")";
+    }
+    echo "</li>";
+}
+echo "</ul>";
+
+echo "<hr>";
+
+echo "<p>" . _("You can access site search in three ways:") . "</p>";
+echo "<ul>";
+echo "<li>" . _("Use the search icon (the magnifying glass) in the header to start a site search. You can cancel it with the escape key.") . "</li>";
+echo "<li>" . _("When on any page with the logo in the header, typing the <kbd>/</kbd> key on your keyboard will bring up the search field as long as an input field doesn't already have focus.") . "</li>";
+echo "<li>" . sprintf(
+    _("Add this page to your browser's custom search feature (eg <a href='%s'>smart keywords</a> in Firefox, <a href='%s'>site shortcuts</a> in Chrome).<br>The URL to use in searching is <kbd>%s</kbd>"),
+    "https://support.mozilla.org/en-US/kb/how-search-from-address-bar",
+    "https://support.google.com/chrome/answer/95426",
+    "$code_url/tools/site_search.php?q=%s"
+) . "</li>";
+echo "</ul>";
+
+//----------------------------------------------------------------------------
+
+function get_prefix_searches(bool $flatten_aliases = false): array
+{
+    global $code_url, $wiki_url;
+
+    $prefix_searches = [
+        "title" => [
+            "url" => "$code_url/tools/search.php?show=search&title=%s",
+            "desc" => _("Project search for this in the title"),
+            "alias" => "t",
+        ],
+        "author" => [
+            "url" => "$code_url/tools/search.php?show=search&author=%s",
+            "desc" => _("Project search for this in the author"),
+            "alias" => "a",
+        ],
+        "pm" => [
+            "url" => "$code_url/tools/search.php?show=search&project_manager=%s",
+            "desc" => _("Project search for this project manager"),
+        ],
+        "user" => [
+            "url" => "$code_url/stats/members/mbr_list.php?uname=%s",
+            "desc" => _("Search for this username or part of a username"),
+            "alias" => "u",
+        ],
+        "team" => [
+            "url" => "$code_url/stats/teams/tlist.php?tname=%s",
+            "desc" => _("Search for this team name or part of a team name"),
+        ],
+        "task" => [
+            "url" => "$code_url/tasks.php?task_id=%s",
+            "desc" => _("Jump to this task ID"),
+            "validator" => "_validate_integer",
+        ],
+    ];
+
+    if (get_url_for_forum()) {
+        $prefix_searches["forum"] = [
+            "url" => get_url_for_search("%s", false),
+            "desc" => _("Search for this in forum posts"),
+            "alias" => "f",
+        ];
+    }
+
+    if (!empty($wiki_url)) {
+        $prefix_searches["wiki"] = [
+            "url" => "$wiki_url/Special:Search/Index.php?profile=all&search=%s",
+            "desc" => _("Search for this in the wiki"),
+            "alias" => "w",
+        ];
+    }
+
+    if ($flatten_aliases) {
+        foreach ($prefix_searches as $key => &$data) {
+            if (isset($data["alias"])) {
+                $alias = $data["alias"];
+                unset($data["alias"]);
+                $prefix_searches[$alias] = $data;
+            }
+        }
+    }
+
+    return $prefix_searches;
+}
+
+function get_jump_words(bool $flatten_aliases = false): array
+{
+    global $code_url, $wiki_url, $blog_url;
+
+    $jump_words = [
+        "activityhub" => [
+            "url" => "$code_url/activity_hub.php",
+            "desc" => _("Activity Hub"),
+            "alias" => "ah",
+        ],
+        "home" => [
+            "url" => "$code_url/default.php",
+            "desc" => _("Home Page"),
+        ],
+        "myprojects" => [
+            "url" => "$code_url/tools/proofers/my_projects.php",
+            "desc" => _("My Projects"),
+            "alias" => "myproj",
+        ],
+        "mysuggestions" => [
+            "url" => "$code_url/tools/proofers/my_suggestions.php",
+            "desc" => _("My Suggestions"),
+            "alias" => "mysugg",
+        ],
+        "preferences" => [
+            "url" => "$code_url/userprefs.php",
+            "desc" => _("Personal Preferences"),
+            "alias" => "prefs",
+        ],
+        "quizzes" => [
+            "url" => "$code_url/quiz/start.php",
+            "desc" => _("Interactive Quizzes and Tutorials"),
+            "alias" => "quiz",
+        ],
+        "statistics" => [
+            "url" => "$code_url/stats/stats_central.php",
+            "desc" => _("Statistics Central"),
+            "alias" => "stats",
+        ],
+        "tasks" => [
+            "url" => "$code_url/tasks.php",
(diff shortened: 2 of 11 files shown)


tools/site_search.php
```diff
@@ -0,0 +1,343 @@
...
+    if (!empty($wiki_url)) {
+        $prefix_searches["wiki"] = [
+            "url" => "$wiki_url/Special:Search/Index.php?search=%s",
+            "desc" => _("Search for this in the wiki"),
+        ];
+    }
+
+    if ($flatten_aliases) {
+        foreach ($prefix_searches as $key => $data) {
+            if (isset($prefix_searches[$key]["alias"])) {
```
bpfoley, line 153: Isn't `$prefix_searches[$key]` just `$data` for the various occurrences this loop?

I tried it out and I think you can write this a bit more effectively as the follow (note the use of a reference in the foreach)

```
  foreach($prefix_searches as $key => &$data) {
      if (($alias = @$data["alias"]) !== null) {
          unset($data["alias"]);
          $prefix_searches[$alias] = $data;
      }
  }
```

tools/site_search.php
```diff
@@ -0,0 +1,343 @@
...
+    }
+
+    if (user_is_a_sitemanager()) {
+        $jump_words["sa"] = [
+            "url" => "$code_url/tools/site_admin/index.php",
+            "desc" => _("Site Administration"),
+        ];
+    }
+
+    if ($flatten_aliases) {
```
bpfoley, line 251: as above...

tools/site_search.php
```diff
@@ -0,0 +1,343 @@
...
+    // is it an Activity?
+    $activity_jumps = get_activity_jumps();
+    if (isset($activity_jumps[strtoupper($query)])) {
+        metarefresh(0, $activity_jumps[strtoupper($query)]);
+    }
+}
+
+function prefix_search(string $query)
+{
+    $matches = [];
```
bpfoley, line 321: Again, I'm probably being fussy here, but we have a mild case of 'arrow code' here. We could avoid the nested indentation with 
```
if (preg_match('/^(\w+):\s*(.*)$/', $query, $matches) === false) return;
...
$data = ...
if ($data == null) return;
...
$query = ...
metarefresh(...);
```

tools/site_search.php
```diff
@@ -0,0 +1,343 @@
...
+    if (!empty($wiki_url)) {
+        $jump_words["wiki"] = [
+            "url" => $wiki_url,
+            "desc" => _("Wiki"),
+        ];
+    }
+
+    if (!empty($blog_url)) {
+        $jump_words["blog"] = [
+            "url" => $wiki_url,
```
srjfoo, line 228: Is the blog supposed to point to the wiki?

srjfoo:
I don't know if this has the potential for being a problem or not, but in testing, I've found myself using the `/` from the help page, which, of course, has the form input box highlighted and active. Which, of course, breaks whatever I'm searching for. Would it be reasonable to clear the form if a `/` is typed (including being cleared of the `/`? Or maybe the `/` should redirect to the form in the header?

For that matter, once the search term is entered, if it's not a valid search, the cursor moves to the beginning of the search string that's still in the form. Should it be being cleared if it's an invalid search term? I think what I'm getting at is, if the user does get an invalid search response, is there a way to get out of it without using the mouse other than editing the search field, or is that just one of the drawbacks?

Observations:
* The wiki search is just a basic wiki search, it looks like. If you search for a username, pages that have that username on them will come up, but the user's personal page will not (I tried adding `User` namespace search and making it persistent, but that didn't work). As I said, just an observation; I don't know if it's worth worrying about.
* Prefix searches are exact string, not substring (e.g. `forum: test` does not also give results for `tests` or `testing`).

tools/site_search.php
cpeel: No, I just caught this myself!

cpeel:
For failed searches, my inclination was to put the failed search terms in the box in case someone wanted to edit them and search again. If someone fat-fingers something like searching for `titl:mark twain` would they rather add the `e` or just type the whole thing over again? I'm not stuck on this and we could just show an empty input field, but that was my rationale.

We could prevent the search box on the page from having focus. That would make `/` work on the page again when it loads.

We could do a combination of either of the above.

I don't want to do anything special for `/` within that search box because that would be extremely unexpected -- when you're in an input field you expect text to be input, not have it not show up or have it move you outside the focus of the field.

> The wiki search is just a basic wiki search

Good call. I just made it do a full search which I think will be better.

> Prefix searches are exact string, not substring

Correct. This was by design -- the full string or the alias works, but not a subset.

---

I also meant to add a text to let users know they can use `/` to bring up the search field on any page. I'll see about getting that in there now.

cpeel:
I've addressed most of the PR comments (thanks @bpfoley) and added a tip on the use of `/` at the bottom.

After thinking about the auto-focus and failed search terms in the box, what I've decided is that we should think about 2 primary users:
* keyboard users probably want to hit `/` on the page and get back to searching without using the mouse. They're not going to want to correct the failed search anyways probably.
* mouse users will probably want to click in the box to edit their failed terms.

So I've removed the auto-focus from the page in hopes that it works for both users.

srjfoo:
> For failed searches, my inclination was to put the failed search terms in the box in case someone wanted to edit them and search again. If someone fat-fingers something like searching for `titl:mark twain` would they rather add the `e` or just type the whole thing over again? I'm not stuck on this and we could just show an empty input field, but that was my rationale.

Perfectly reasonable thinking.

> We could prevent the search box on the page from having focus. That would make `/` work on the page again when it loads.

I like the idea of removing focus on the search box on a failed search so that the `/` works. Would that make it harder for those who want to edit the input on the help page?

> We could do a combination of either of the above.

How would a combination work?

> I don't want to do anything special for `/` within that search box because that would be extremely unexpected -- when you're in an input field you expect text to be input, not have it not show up or have it move you outside the focus of the field.

I was really pretty much just tossing stuff at the wall in hopes that something might be helpful 😁 

> > The wiki search is just a basic wiki search
> 
> Good call. I just made it do a full search which I think will be better.

I really like that -- it should make it easier, for instance, to search for a template. Or a file. (Initially, when I tried the updated version, I did it without the `:`, and nearly said "butbutbut it's not working! 😆 )

> > Prefix searches are exact string, not substring
> 
> Correct. This was by design -- the full string or the alias works, but not a subset.

That's fine -- I just wanted to make sure it was intentional.
 
> I also meant to add a text to let users know they can use `/` to bring up the search field on any page. I'll see about getting that in there now.

👍

70ray (commented):
Noticed two minor things while testing:

1: If you put in something that it doesn't understand (like ab) you get a warning message. Then do a sensible search (like ah). Then go back to the previous page, you get to the page with the warning message. It seems like it would be better to get back to the raw page or have 'ah' filled in.

2: If you search for a language without capitalizing the first letter, the search understands the language correctly but if you then then do a refine search, the search does not remember the language.

cpeel:
> 1: If you put in something that it doesn't understand (like ab) you get a warning message. Then do a sensible search (like ah). Then go back to the previous page, you get to the page with the warning message. It seems like it would be better to get back to the raw page or have 'ah' filled in.

I agree this is not ideal. To do that we would need to not ever put the failed search term in the search box and at least right now I think that's the better trade-off. It's quite possible that I'm wrong and we should just never put the failed search term in the box and if the consensus (between the devs in this PR or users post-deploy) is that we don't then I'm happy to change that.

> 2: If you search for a language without capitalizing the first letter, the search understands the language correctly but if you then then do a refine search, the search does not remember the language.

Good find. I don't see an easy way to fix this without doing a fair amount of introspection on the data on the site search page itself. Right now you can search for parts of languages "eng", "english", "engli" and they all work to search, but will all fail if you refine it. I think we should either accept this limitation or remove `language` as something you can search on with the site search (which is my preference as a project language isn't going to be highly selective). @bpfoley you requested this feature, what do you think?

bpfoley:
Hmm. As you say, the selectivity of language on its own is pretty low and I'd imagine much of the time language specialists will also work in a specific round, so you really want to be able to search for, e.g. Hawaiian and F1. If refine search isn't working and is a lot of extra work to fix, I'd say it's probably not worth having language at all...

Sorry for creating the extra churn.

windymilla:
> Good find. I don't see an easy way to fix this without doing a fair amount of introspection on the data on the site search page itself. Right now you can search for parts of languages "eng", "english", "engli" and they all work to search, but will all fail if you refine it. I think we should either accept this limitation or remove `language` as something you can search on with the site search (which is my preference as a project language isn't going to be highly selective). @bpfoley you requested this feature, what do you think?

If we decide to keep language, I don't think that limitation is too serious, but I agree language is less useful than the others, so I'd vote to remove it. 
I suppose you could consider it a problem with the Project Search/Refine page that you can give it a URL that it can't reflect in the search form, e.g. `https://www.pgdp.org/~cpeel/c.branch/site-search/tools/search.php?show=search&language[]=f` lists all projects that have `f` in the language, but the search form can't reflect that. However, I doubt that it's something the users will notice or need. A workaround is to just re-select the language when using the form to refine.

windymilla (approved):
Excellent feature!

srjfoo:
> > 1: If you put in something that it doesn't understand (like ab) you get a warning message. Then do a sensible search (like ah). Then go back to the previous page, you get to the page with the warning message. It seems like it would be better to get back to the raw page or have 'ah' filled in.
> 
> I agree this is not ideal. To do that we would need to not ever put the failed search term in the search box and at least right now I think that's the better trade-off. It's quite possible that I'm wrong and we should just never put the failed search term in the box and if the consensus (between the devs in this PR or users post-deploy) is that we don't then I'm happy to change that.

The first thing that came to mind to test this on other sites was archive.org. I tried their advanced search (which I know uses keywords) and set up a search that I knew would work, then changed it slightly to get a null result. Then fixed it again. Going back to the previous page takes you back to the unsuccessful result, so I don't think that's the wrong approach.
> 
> > 2: If you search for a language without capitalizing the first letter, the search understands the language correctly but if you then then do a refine search, the search does not remember the language.
> 
> Good find. I don't see an easy way to fix this without doing a fair amount of introspection on the data on the site search page itself. Right now you can search for parts of languages "eng", "english", "engli" and they all work to search, but will all fail if you refine it. I think we should either accept this limitation or remove `language` as something you can search on with the site search (which is my preference as a project language isn't going to be highly selective). @bpfoley you requested this feature, what do you think?

I played with this, too. If you do a title search, and refine the search, the text you typed in *is* in the project search form. I assume that the difference is that title (also author) is a text field, and language is a multi-select?

That being said, I agree with you, Brian and Nigel. I'd just remove it. I would think that language would be best used in conjunction with something else (like all German projects available for PP, for instance) and that's just easiest done from the search form itself.

`<tb>`

Rambling thoughts: As I was doing this round of testing it occurred to me to ask if we should allow a search on PG number (which would point to the project page for that project). Unlike searching for a project by projectID, that would require a keyword. I don't feel strongly about it, though. I would not even expect a search on projectid to be generally useful unless someone is working with a list of bare projectIDs, and  (for those of us who might be working from a list of bare projectIDs, yes. But even then, you're better off copying the list of projectids and pasting it into a project search form, because you can give that form multiple projectIDs instead of doing them one at a time.

pinc/theme.inc
```diff
@@ -179,12 +179,18 @@ function html_logobar()
             "class" => "fas fa-question fa-2x",
             "url" => get_faq_url("faq_central.php"),
         ];
+        $icon_links[] = [
+            "name" => _("Search"),
+            "class" => "fas fa-search fa-2x",
```
chrismiceli, line 184: The current HTML output does not change the mouse on hover, so I wasn't aware this search icon was interact-able at first. Could we make the cursor a hand on hover link an anchor?

tools/site_search.php
```diff
@@ -0,0 +1,354 @@
...
+    if (isset($jump_words[$query])) {
+        metarefresh(0, $jump_words[$query]["url"], "", "", true);
+    }
+}
+
+function smart_search(string $query)
+{
+    global $code_url;
+
+    // is it a valid projectID?
```
chrismiceli, line 303: can a user name match a project id? Same for can a user name be one of our jumps?

tools/site_search.php
```diff
@@ -0,0 +1,354 @@
...
+    $activity_jumps = get_activity_jumps();
+    if (isset($activity_jumps[strtoupper($query)])) {
+        metarefresh(0, $activity_jumps[strtoupper($query)]);
+    }
+}
+
+function prefix_search(string $query)
+{
+    $matches = [];
+    if (! preg_match('/^(\w+):\s*(.*)$/', $query, $matches)) {
```
chrismiceli, line 329: *minor* is that space after `!` normal?

pinc/theme.inc
cpeel: Good catch - I'll fix this.

tools/site_search.php
cpeel: Technically a username could match a project ID and in that case the requester would be taken to the project, not the user. They could use the `user:` prefix to specify they want the user not the project.

Most all of our jump words would be valid usernames. The jump words are evaluated first for this reason. See also [this line](https://github.com/DistributedProofreaders/dproofreaders/pull/1184/files#diff-05373f93c16a82c90bebb30d9ec57edeaf43a5e3d8e35aae34d1f1ad2b59ab21R18).

tools/site_search.php
cpeel: Looks like most of our code uses no-space, but we have spaces elsewhere, eg:
```
user_is.inc:        if (!$shown) {
user_is.inc:    if (! is_numeric($limit)) {
```
Sad that `php-cs-fixer` isn't standardizing those. I'll fix this one to match the bulk of our code.

cpeel:
`language:` prefix removed, "Search" name is now also a link and both the icon and the name use pointers for cursors. Addressed other PR feedback.

srjfoo:
See the image below. The pointer seems to work for the search icon with no problem, and I tried both French and German, and while I was able to keep the pointer from working on the others occasionally, it's not so bad, but "Mitteilungen" is long enough that there's a lot of empty space within the rectangle that shows as a pointer, but doesn't actually bring up the link, both to the left, where the pointer in the image, below, is, and to the right between the number and the label (I assume that it would mirror the left if there weren't a number there.

image 

Looking more closely (literally, zooming in on the page header), and clicking around all of the icons (back in English), I'm seeing some inconsistency. I couldn't get the pointer with no link for the wiki or blog icons, but I could for Inbox, Forums and Help. The search icon is a bit different, and clicking anywhere in the enclosing rectangle triggers it just fine.

(I apologize for the image quality, but while I have great fun doing screenshots in general, I can't seem to make them show the cursor, even with that option turned on, so I took a pic with my phone to get this.)

70ray:
> I suppose you could consider it a problem with the Project Search/Refine page that you can give it a URL that it can't reflect in the search form, e.g. `https://www.pgdp.org/~cpeel/c.branch/site-search/tools/search.php?show=search&language[]=f` lists all projects that have `f` in the language, but the search form can't reflect that. However, I doubt that it's something the users will notice or need. A workaround is to just re-select the language when using the form to refine.

Yes, this is really a problem with the Project Search page, possibly fixable later.

chrismiceli:
Minor UX feedback, the error message doesn't stand out that much. With most of the page being red, it was difficult to see what went wrong
image

pinc/theme.inc
```diff
@@ -206,6 +212,13 @@ function html_logobar()
         echo "<div id='icon-menu'>";
         echo join("", $links);
         echo "</div>"; // icon-menu
+
+        echo "<div id='search-menu' class='nodisp'>";
+        $placeholder = attr_safe(_("Search for 'help' to see options"));
```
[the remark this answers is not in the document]
cpeel: Yeah, I will pull out the `help` into an `sprintf()` so that doesn't get translated.

cpeel:
> Minor UX feedback, the error message doesn't stand out that much. With most of the page being red, it was difficult to see what went wrong

These are using the existing `warning` style for paragraphs and our standard style for `<kbd>` tags to indicate keyboard input. We could style the "don't know what to do with the search" text differently (`error` type seems a little dramatic for this -- neither `warning` or `error` seem entirely accurate) and/or we could change the `<kbd>` style across the site. The former is probably easier to do as I don't really want to revisit the `<kbd>` style across the site. Thoughts?

cpeel:
> Looking more closely (literally, zooming in on the page header), and clicking around all of the icons (back in English), I'm seeing some inconsistency.

I've revised the "how things link" and styling once again and now the icons and texts are links and as links are the only thing that show the pointer icon.

I hope this also resolves @chrismiceli 's comment about breaking tabbing as now everything in the header -- including search -- is a link.

> Yes, this is really a problem with the Project Search page, possibly fixable later. So I think we should keep the language option.

We can re-add the language option if/when we fix the project search. Easier to add things we know work than release things that don't quite work and field questions.

pinc/theme.inc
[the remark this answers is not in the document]
srjfoo: I would expect all keywords to be keywords. It's true, I haven't checked any other sites that use keyword searches to see if they translate them, but, I think that it's reasonable not to have translated versions. The tooltips should be translated, though.

Which makes me wonder -- for the help message, should a TRANSLATORS note be added to suggest that a comment be added to the translated version that the keywords work only as given?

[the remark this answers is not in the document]
windymilla: I'm not suggesting it should be changed now, but just wondered about it. I did a quick trial in Slack. You can type `/open #general` for example to open a channel named general. If you change your language to French, you can type either `/open #general` or `/ouvrir #general`. So, it seems as though the French version of the keyword is added, but the English version is also retained.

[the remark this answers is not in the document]
cpeel: Interesting! I think we should start with just English. Then we can work with the translators on custom prefix & jump words. I don't think just wrapping them in `_()` is going to be the right approach here.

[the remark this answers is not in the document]
windymilla: Definitely agree!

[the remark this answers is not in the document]
srjfoo: > Interesting! I think we should start with just English. Then we can work with the translators on custom prefix & jump words. I don't think just wrapping them in `_()` is going to be the right approach here.

I agree. Somehow, you'd have to link the translation in the PO fle to the keyword, wouldn't you?

70ray (approved):

cpeel:
> I don't suppose there's a way to get rid of the latter? It's not really needed.

We're not doing anything "special", that's just the browser indicating that the input field has focus. I don't know of anything that could change it but my CSS-foo is not amazing.

srjfoo:
> > I don't suppose there's a way to get rid of the latter? It's not really needed.
> 
> We're not doing anything "special", that's just the browser indicating that the input field has focus. I don't know of anything that could change it but my CSS-foo is not amazing.

Based on what other sites do, there is definitely a way to get rid of it (for instance this text input box has a blue highlight while the field is active; phpbb on PMs of forum posts has no special highlighting at all. But like you, I have no idea how to do it. It's not the end of the world, just a bit more emphasis than I think the field really needs. And I just checked -- it's not nearly as much in-your-face on the lighter themes. 🤣

chrismiceli (approved):
It could be cool if we could autocomplete some of the jump options? Definitely a nice to have https://jsfiddle.net/7jdzntke/

bpfoley:
> It could be cool if we could autocomplete some of the jump options? Definitely a nice to have https://jsfiddle.net/7jdzntke/

Agreed. I was thinking this too, but forgot to bring it up.

70ray:
> > > I don't suppose there's a way to get rid of the latter? It's not really needed.
> > 
> > 
> > We're not doing anything "special", that's just the browser indicating that the input field has focus. I don't know of anything that could change it but my CSS-foo is not amazing.
> 
> Based on what other sites do, there is definitely a way to get rid of it (for instance this text input box has a blue highlight while the field is active; phpbb on PMs of forum posts has no special highlighting at all. But like you, I have no idea how to do it. It's not the end of the world, just a bit more emphasis than I think the field really needs. And I just checked -- it's not nearly as much in-your-face on the lighter themes. 🤣

You can remove it with something like "#search-menu-input:focus {
  outline: none;
}"
We can't use outline yet but can use its components.
[https://developer.mozilla.org/en-US/docs/Web/CSS/:focus](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus) explains why you shouldn't do this.

cpeel:
> You can remove it with something like

Oh nice, thanks Ray. My inclination is to leave it as-is and not mess with browser accessibility features.

srjfoo:
> > You can remove it with something like
> 
> Oh nice, thanks Ray. My inclination is to leave it as-is and not mess with browser accessibility features.

I'll just live with it, then.

bpfoley:
> > You can remove it with something like
> 
> Oh nice, thanks Ray. My inclination is to leave it as-is and not mess with browser accessibility features.

Agreed. The focus ring is intended as a cue for users who use keyboard navigation -- whether it's built-in OS/browser accessibility features, or site-specific Javascript. We should keep it as is unless we've a specific reason to remove it.

srjfoo:
> > > You can remove it with something like
> > 
> > 
> > Oh nice, thanks Ray. My inclination is to leave it as-is and not mess with browser accessibility features.
> 
> Agreed. The focus ring is intended as a cue for users who use keyboard navigation -- whether it's built-in OS/browser accessibility features, or site-specific Javascript. We should keep it as is unless we've a specific reason to remove it.

I will defer to wiser heads; I'll get used to it. I do want to mention that, at least for this search, I'm seeing a double ring in all but Safari (though I haven't looked at other browsers/OSs). Who knew that accent colors could be such a pain? Too brigh in this case, not bright enough when highlighting small pieces of text in some cases, with the same base color!

cpeel:
Rebased to squash the commits and will merge after the CI passes.

Thanks to everyone for your great feedback!

PR #1185, opened 2024-04-14
Add download button for charts

chrismiceli:
Sandbox here:
https://www.pgdp.org/~chrismiceli/c.branch/download-chart/stats/stats_central.php

### DIFF
--- a/scripts/svgGraphs.js
+++ b/scripts/svgGraphs.js
@@ -70,6 +70,65 @@ const {barLineGraph, stackedAreaGraph, pieGraph} = (function () {
         }
     }
 
+    const nodeIterator = (elem, fn) => {
+        fn(elem);
+        for (const child of elem.children) {
+            nodeIterator(child, fn);
+        }
+    };
+
+    const createSvgText = (svgElement) => {
+        const classNames = new Set(['graph-background']);
+
+        nodeIterator(svgElement, (elem) => {
+            elem.classList.forEach(className => {
+                classNames.add(className);
+            });
+        });
+
+        const dummyElement = document.createElement('a');
+        classNames.forEach(className => {
+            dummyElement.classList.add(className);
+        });
+        const cssStyleRules = new Set();
+        for (const sheet of document.styleSheets) {
+            for (const rule of sheet.cssRules) {
+                if (dummyElement.matches(rule.selectorText)) {
+                    cssStyleRules.add(rule.cssText);
+                }
+            }
+        }
+
+        let svgText = svgElement.outerHTML.replace('<svg ', '<svg xmlns="http://www.w3.org/2000/svg" width="450px" ');
+        const endOfFirstElement = svgText.indexOf('>');
+        svgText = `${svgText.slice(0, endOfFirstElement + 1)}
+            <style>
+                ${[...cssStyleRules].join('')}
+            </style>
+            <rect width="100%" height="100%" class="graph-background" />
+            ${svgText.slice(endOfFirstElement + 1)}`;
+        return svgText;
+    };
+
+    function addDownloadButton(svg, config) {
+        const svgElement = svg.nodes()[0];
+        const downloadButton = document.createElement('button');
+        downloadButton.textContent = 'Download Image';
+        downloadButton.style.margin = `5px ${margin.left}px 0 ${margin.right}px`;
+        downloadButton.addEventListener('click', () => {
+            let svgBlob = new Blob([createSvgText(svgElement)], { type: 'image/svg+xml' });
+            const downloadLink = document.createElement('a');
+            const downloadLinkHref = window.URL.createObjectURL(svgBlob);
+            downloadLink.href = downloadLinkHref;
+            downloadLink.download = config.title;
+            document.body.appendChild(downloadLink);
+            downloadLink.click();
+            document.body.removeChild(downloadLink);
+            window.URL.revokeObjectURL(downloadLinkHref);
+        });
+        svgElement.parentElement.appendChild(downloadButton);
+    }
+
     function stackedAreaGraph(id, config) {
         const height = 400;
         const width = 640;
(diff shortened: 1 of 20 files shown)


srjfoo (approved):
Downloaded svg charts look good to me.

scripts/svgGraphs.js
```diff
@@ -70,6 +70,65 @@ const {barLineGraph, stackedAreaGraph, pieGraph} = (function () {
...
+        const cssStyleRules = new Set();
+        for (const sheet of document.styleSheets) {
+            for (const rule of sheet.cssRules) {
+                if (dummyElement.matches(rule.selectorText)) {
+                    cssStyleRules.add(rule.cssText);
+                }
+            }
+        }
+
+        let svgText = svgElement.outerHTML.replace('<svg ', '<svg xmlns="http://www.w3.org/2000/svg" width="450px" ');
```
cpeel, line 102: I think users are going to expect the downloaded SVG to be the same size as the one they had on the screen, not smaller. I know this isn't Linda's use-case but this is now a button for everyone for (almost) every graph.

Is it possible to set the downloaded width to the _current_ width of the on-page SVG? Then Linda could narrow her screen size to the width she wants and just download them then.

scripts/svgGraphs.js
```diff
@@ -70,6 +70,65 @@ const {barLineGraph, stackedAreaGraph, pieGraph} = (function () {
...
+            <rect width="100%" height="100%" class="graph-background" />
+            ${svgText.slice(endOfFirstElement + 1)}`;
+        return svgText;
+    };
+
+    function addDownloadButton(svg, config) {
+        const svgElement = svg.nodes()[0];
+        const downloadButton = document.createElement('button');
+        downloadButton.textContent = 'Download Image';
+        downloadButton.style.margin = `5px ${margin.left}px 0 ${margin.right}px`;
```
cpeel, line 117: Let's move this styling into a class in the stylesheet. We might have more flexibility if the DOM of this was:
```
<div class='graph-download'><button ... ></div>
```
Then we could have the button be right-justified, for instance.

cpeel (commented):
This is incredibly cool.

The download button added to every graph of a specific type makes me wonder if and how we're going to run into a situation where someone wants a smaller graph where the download button doesn't fit (or fits but is really big in relation to the smaller graph). I think we should punt this issue until it becomes a problem though.

chrismiceli:
> This is incredibly cool.
> 
> The download button added to every graph of a specific type makes me wonder if and how we're going to run into a situation where someone wants a smaller graph where the download button doesn't fit (or fits but is really big in relation to the smaller graph). I think we should punt this issue until it becomes a problem though.

I made it configurable by added the ability to not provide the localized label. If there is no label, the button won't show. This is currently the case for stats_central chart since we cache the data and the cache doesn't have the `downloadLabel` config property for localization.

scripts/svgGraphs.js
chrismiceli: Good idea. I've implemented this.

scripts/svgGraphs.js
```diff
@@ -70,6 +70,68 @@ const {barLineGraph, stackedAreaGraph, pieGraph} = (function () {
...
+            ${svgText.slice(endOfFirstElement + 1)}`;
+        return svgText;
+    };
+
+    function addDownloadButton(svg, config) {
+        if (config.downloadLabel) {
+            const svgElement = svg.nodes()[0];
+            const downloadButton = document.createElement('button');
+            downloadButton.textContent = config.downloadLabel;
+            downloadButton.classList.add('grpah-download-button');
```
cpeel, line 119:
```suggestion
            downloadButton.classList.add('graph-download-button');
```
(and in `layout.less` -- at least they matched! :D)

scripts/svgGraphs.js
cpeel: You added a style to the button, but I'd like to have the button in side a div, so we could -- if we wanted -- control where the button is within the div (left, center right).

bpfoley:
This is definitely very useful. If we want to make the download button into a link/button inside the div, we could possibly use https://fontawesome.com/icons/file-export?f=classic&s=solid or https://fontawesome.com/icons/download?f=classic&s=solid

bpfoley:
A thought though -- and this might not match with other people's expectations/desires.

Pretending that I haven't read the button title, my mental model would have been that clicking on the SVG button might have opened a standalone SVG in a new browser window -- which might be useful if you just want to see the graph without the surrounding HTML. If you wanted to save the image, you could then just save that page (with cmd-S/ctrl-S/whatever).

That is to say, I wonder would 'open SVG in new tab' might be the more flexible functionality to have?

cpeel:
> That is to say, I wonder would 'open SVG in new tab' might be the more flexible functionality to have?

To me that is less functional / desirable than downloading it. Seems that wanting to view the graph without the other stuff on a page would not be as common as wanting to download it for use elsewhere.

70ray (commented):
This is not really related to this PR but some of the svg files seem rather large in relation to the appearance of the graph.

bpfoley:
They compress very well with [svgcleaner](https://github.com/RazrFalcon/svgcleaner) and [svgo](https://github.com/svg/svgo). For example one of the graphs that was 2.6MB compresses to ~350kB. I haven't looked in depth, but part of the problem is that many of the coordinates have far too many digits of precision, eg 100.8127534

chrismiceli:
> They compress very well with [svgcleaner](https://github.com/RazrFalcon/svgcleaner) and [svgo](https://github.com/svg/svgo). For example one of the graphs that was 2.6MB compresses to ~350kB. I haven't looked in depth, but part of the problem is that many of the coordinates have far too many digits of precision, eg 100.8127534

I think we can make sure that Linda is aware of those tools to make the files smaller. I don't think this compression is something we should handle client side at download time.

scripts/svgGraphs.js
cpeel: This isn't critical and we could add it later if we wanted.

scripts/svgGraphs.js
cpeel: I pushed a commit to your branch to fix this.

cpeel (approved):

scripts/svgGraphs.js
chrismiceli: Thanks, I forgot to come back to that!

srjfoo (approved):

PR #1187, opened 2024-04-16
Issue warning if the project CTF was generated pre-Unicode

cpeel:
_Note: this is going into `pgdp-production`._

PP artifacts that were generated before the Unicode conversion still have their contents in Latin-1. There aren't many of these but we periodically stumble over some that are reclaimed.

Luckily, we just so happen to have have a project in TEST in this state: https://www.pgdp.org/~cpeel/c.branch/pgdp-warn-old-pp-artifacts/project.php?id=projectID44dda73a86096&expected_state=proj_post_first_available

### DIFF
--- a/project.php
+++ b/project.php
@@ -1654,6 +1654,11 @@ function echo_download_zip($link_text, $discriminator)
     echo "</a>";
     echo_byte_size($filesize_b);
     echo_last_modified($last_modified);
+    // output a warning if the file was generated before the UTF-8 cutover.
+    // timestamp is midnight May 18 2020
+    if ($discriminator == "" && $last_modified && $last_modified < 1589785200) {
+        echo "<br><span class='warning'>" . _("This file was generated before the Unicode conversion. If you are just starting to PP this project, please ask a site administrator to regenerate the file.") . "</span>";
+    }
     echo "</li>";
     echo "\n";
 }


project.php
```diff
@@ -1654,6 +1654,10 @@ function echo_download_zip($link_text, $discriminator)
     echo "</a>";
     echo_byte_size($filesize_b);
     echo_last_modified($last_modified);
+    // output a warning if the file was generated before the UTF-8 cutover
+    if ($discriminator == "" && $last_modified && $last_modified < 1589698800) {
+        echo "<br><span class='warning'>" . _("This file was generated before the Unicode conversion. If you are just starting to PP this project, it's recommended you regenerate the file.") . "</span>";
```
srjfoo, line 1659: I recommend that
> it's recommended you regenerate the file.

be changed to
> please ask a site administrator to regenerate the file.

unless you want to get wordy and point out that they can generate a CTF for download that ("see 'Download Concatenated Text' below"), while it won't include GWL/BWL, will have a regenerated version of the text. Only site admins can regenerate the post files that are stored on the server.

windymilla (approved):
Thanks - I agree with Sharon's slight wording adjustment

project.php
```diff
@@ -1654,6 +1654,10 @@ function echo_download_zip($link_text, $discriminator)
     echo "</a>";
     echo_byte_size($filesize_b);
     echo_last_modified($last_modified);
+    // output a warning if the file was generated before the UTF-8 cutover
+    if ($discriminator == "" && $last_modified && $last_modified < 1589698800) {
```
chrismiceli, line 1658: would it make sense to translate the date in a comment?

cpeel: Indeed - added.

project.php
cpeel: Wording updated, thanks for the assist!

srjfoo (approved):

chrismiceli (approved):

Issue #1190, opened 2024-04-17
Change past_tallies to a sparse format


cpeel:
The `past_tallies` table stores historical round, user, and team statistics (`current_tallies` stores the current day's statistics) including the number of pages done that day (`tally_delta`). Once a day the site takes a snapshot of `current_tallies`, calculates the delta since the last snapshot, and adds a row in the `past_tallies` table for every `[tally_type, tally_name, holder_id]` tuple.

After a user has done a page in a round, they have an entry for that round in `current_tallies` and the site adds a new record in `past_tallies` for that user/round in every daily snapshot going forward. Said another way: we keep daily records of every user that has touched a round forever, even if they never proofread another page.

At pgdp.net, 24 years of this has caught up with us. The `past_tallies` table is 50GB in size and contains 598,611,362 rows. Of those 594,322,594 are records where `tally_delta`=0 -- 99%.

We need to convert the table to a sparse format where we only store rows where `tally_delta` <> 0 and adjust all of the code accordingly. This will drastically reduce the size of the table and improve our DB backup times considerably.

cpeel:
We use `past_tallies` in a few primary ways

1. To get the sum of pages done over a specific time span eg: last week, last month
   * A sparse table format where we do not store rows where `tally_delta` == 0 isn't a problem for this one. We just `sum(tally_delta)` over a time range anyway -- we don't need the zeros.
2. To get the number of pages done at a specific time (usually "yesterday")
   * This one we could calculate from a sparse `past_tallies` pretty easily.
3. To get the number of pages done every day over some span for a table or a chart
   * Without some sort of secondary table that records all of the snapshot times we can't accurately generate the data because we don't know what days are in-between the records where `tally_delta` <> 0. `past_tallies` itself isn't set up to query against `timestamp` and there's nothing that guarantees that every snapshot will have at least one non-zero `tally_delta`. (This might be practically true on pgdp.net but isn't guaranteed to be true for all sites.)
4. In concert with `current_tallies` to determine how many pages they did between yesterday and today to add new entries into `past_tallies
   * We can programmatically determine this from a sparse `past_tallies` but we add a whole lot of computation for every snapshot in ways that just don't make sense.

---

I propose introducing 2 new tables:
```sql
CREATE TABLE `latest_snapshot_tallies` (
  `tally_name` char(2) NOT NULL default '',
  `holder_type` char(1) NOT NULL default '',
[...]

PR #1191, opened 2024-04-17
Remove select_from_site_past_tallies_and_goals()

cpeel:
Part of solving https://github.com/DistributedProofreaders/dproofreaders/issues/1190 is going to be decreasing the number of instances where we directly query `past_tallies` where the caller assumes that it has entries for every single day. This PR removes one of the offenders used only in the statistics code. It pulls the DB-ness into a function. Right now that function just queries `past_tallies` directly, but in a future revision it will encode more logic to handle the transition to a sparse format.

This has two functional changes:
* `misc_stats1.php` (what a terrible name for a file, BTW) no longer outputs the top 10 proofreading days this year and it no longer outputs the days above certain thresholds. Supporting these two use-cases required more work for what I think is a pretty small return
* The tables in `misc_stats1.php` do not have a "This Month?" column as that seemed like a pretty poor UI choice to begin with.

Except for those two things, the stats pages and graphs should match what's on TEST.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/reduce-past-tallies-direct-access/stats/stats_central.php

Notes:
* The sandbox has a temporary hack such that graphs are not cached to ease testing.
* `misc_stats1.php` is the "Top Proofreading Days and Months, etc." link on the main statistics page.

### DIFF
--- a/pinc/graph_data.inc
+++ b/pinc/graph_data.inc
@@ -533,16 +533,10 @@ function total_pages_by_month_graph($tally_name)
     ///////////////////////////////////////////////////
     //Total pages by month since beginning of stats
 
-    $result = DPDatabase::query(select_from_site_past_tallies_and_goals(
-        $tally_name,
-        "SELECT {year_month}, SUM(tally_delta), SUM(goal)",
-        "",
-        "GROUP BY 1",
-        "ORDER BY 1",
-        ""
-    ));
-
-    [$datax, $datay1, $datay2] = dpsql_fetch_columns($result);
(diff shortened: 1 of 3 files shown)


chrismiceli (approved):

srjfoo (approved):

windymilla (commented):
Expected stats in branch look the same as on TEST. Completely agree that the "top 10 this year" and "> threshold" graphs are not worth the work or added complexity.

windymilla (approved):

PR #1193, opened 2024-04-20
include navbar_menu.js only if needed

70ray:
Avoids js exception on pages without a dp header.

I haven't made a sandbox. This follows on from Casey's search function and he might want to fix it another way.


cpeel:
Thanks for finding this Ray! I'd like to solve it slightly differently (https://github.com/DistributedProofreaders/dproofreaders/pull/1195).

PR #1195, opened 2024-04-20
Don't include navbar JS if there's no navbar

cpeel:
Don't include the navbar JS when there is no navbar. In addition add some checks to see if the search menu exists before adding listeners for it. A slightly different approach than Ray's https://github.com/DistributedProofreaders/dproofreaders/pull/1193. Diff best viewed ignoring whitespace.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/fix-navbar-js-error/

Testing notes: The new search menu should still function. On page without a header, like the Image Browser, there should be no error in the JS console.


70ray (approved):
LGTM

chrismiceli (approved):

srjfoo (approved):

PR #1196, opened 2024-04-20
Improve performance of calculating moving average

cpeel:
This is, frankly, embarrassing.

The thing that takes _so long_ to generate the stats graphs is calculating the moving average due to some complex SQL that joins `past_tallies` against itself. If we calculate the SMA in the code it's more than 10x faster on TEST -- probably even faster on PROD. It also removes another direct use of `past_tallies` which is needed for moving to the future sparse format.

Note that the old code only generated integer-based averages (I don't know why) whereas the new one will generate floating point averages, so the graphs will look slightly different.

Compare:
* https://www.pgdp.org/c/stats/pages_proofed_graphs.php?tally_name=P1
* https://www.pgdp.org/~cpeel/c.branch/improve-sma/stats/pages_proofed_graphs.php?tally_name=P1

Both of the above have graph caching disabled to show both the speed difference and to ensure we're seeing the actual results of the code and not a cached copy.

### DIFF
--- a/pinc/TallyBoard.inc
+++ b/pinc/TallyBoard.inc
@@ -624,15 +624,17 @@ class TallyBoard
 
     /**
      * Return all tally deltas recorded for the given holder
-     * since $min_timestamp.
+     * since $min_timestamp and optionally before $max_timestamp
      *
      * Specifically, the result is an array of key => value pairs,
      * where key is the ascribed time of a snapshot,
      * and value is the tally delta recorded for that time.
      * (The items are sorted by timestamp.)
      */
-    public function get_deltas($holder_id, $min_timestamp)
+    public function get_deltas($holder_id, $min_timestamp, $max_timestamp = null)
     {
+        $addl_where = $max_timestamp === null ? "" : sprintf("AND timestamp < %d", $max_timestamp);
+
         $sql = sprintf(
             "
             SELECT timestamp, tally_delta
@@ -642,6 +644,7 @@ class TallyBoard
                 AND holder_type = '%s'
                 AND holder_id   = %d
                 AND timestamp  >= $min_timestamp
+                $addl_where
             ORDER BY timestamp ASC
             ",
             DPDatabase::escape($this->tally_name),
--- a/pinc/graph_data.inc
+++ b/pinc/graph_data.inc
@@ -574,10 +574,8 @@ function array_accumulate($arr)
     return $result;
 }
 
-function pages_daily($tally_name, $c_or_i, $timeframe)
+function pages_daily(string $tally_name, string $c_or_i, string $timeframe)
 {
-    $site_tallyboard = new TallyBoard($tally_name, 'S');
-
     $now_timestamp = time();
     $now_assoc = getdate($now_timestamp);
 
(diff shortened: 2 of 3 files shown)


chrismiceli (approved):

srjfoo (approved):

windymilla (approved):
Much faster!

pinc/TallyBoard.inc
```diff
@@ -624,15 +624,17 @@ class TallyBoard
...
      *
      * Specifically, the result is an array of key => value pairs,
      * where key is the ascribed time of a snapshot,
      * and value is the tally delta recorded for that time.
      * (The items are sorted by timestamp.)
      */
-    public function get_deltas($holder_id, $min_timestamp)
+    public function get_deltas($holder_id, $min_timestamp, $max_timestamp = null)
     {
+        $max_where = $max_timestamp === null ? "AND 1" : sprintf("AND timestamp < %d", $max_timestamp);
```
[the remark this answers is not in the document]
cpeel: There is not and either would work. There are times we use a construct like:
```sql
SELECT blah
FROM blerg
WHERE $where
```
and `$where` needs to be a valid expression and so `1` is the right option for "all rows" -- and this is what I was probably channeling.

jmdyck:
> Note that the old code only generated integer-based averages (I don't know why) whereas the new one will generate floating point averages, so the graphs will look slightly different.

For the linked example, the averages are in the thousands, so the effect of rounding would be a small fraction of a pixel. So something else must be making them look different.

One point of comparison: in the top graph, after 2024-04-01's massive count, you'd expect the SMA to go up; in the status quo graph, it does, but in the sandbox graph, it stays constant, and doesn't go up until a day later (after a day of basically zero pages). That suggests an off-by-one error, but it can't be that simple, because the sandbox's SMA line isn't just shifted one-day-to-the-right.

Another thing I noticed: in the "Pages Done Per Day (since stats began)" graph, the SMA is basically zero for the whole graph, but it should be visibly non-zero once we get into late 2023. (Mind you, the status quo graph shows the same thing, so it's a pre-existing problem.)

jmdyck:
Further evidence that something is wrong: Consider the two 21-day windows:
- 2024-03-29 to 2024-04-18 inclusive
- 2024-03-30 to 2024-04-19 inclusive

In shifting from one to the next, we subtract the count for 03-29 and add the one for 04-19. But we can see that the counts for these two days were both zero. So the window totals, and thus the successive SMAs should be the same. But that's not what the graphs (status quo or sandbox) show.

Or am I misunderstanding what SMA is supposed to be?

70ray (commented):
I'd have to spend such a long time figuring out how this works that I'll skip this.

pinc/TallyBoard.inc
```diff
@@ -624,15 +624,17 @@ class TallyBoard
...
-     * since $min_timestamp.
+     * since $min_timestamp and optionally before $max_timestamp
      *
      * Specifically, the result is an array of key => value pairs,
      * where key is the ascribed time of a snapshot,
      * and value is the tally delta recorded for that time.
      * (The items are sorted by timestamp.)
      */
-    public function get_deltas($holder_id, $min_timestamp)
+    public function get_deltas($holder_id, $min_timestamp, $max_timestamp = null)
```
bpfoley, line 634: Let's sprinkle type annotations here:`(int $holder_id, int $min_timestamp, int? $max_timestamp = null)`

pinc/graph_data.inc
```diff
@@ -650,64 +650,45 @@ function pages_daily($tally_name, $c_or_i, $timeframe)
...
-        // Unless it's for all_time when start_timestamp==0. For this case we need
-        // $days_to_average days after the timestamp is zero to ensure the average
-        // is correct.
-        if ($start_timestamp == 0) {
-            $where_start_timestamp = $start_timestamp + ($days_to_average * 60 * 60 * 24);
+        // To calculate a simple moving average over $days_to_average we need
+        // to have at least that many days worth of data before $start_timestamp.
+        // This is simple for everything except right at the beginning when
+        // stats were collected. For that we need to jump through some hoops
+        // to 1) detect it and 2) pad the data with 0s so that we can calculate
```
bpfoley, line 657: It seems to me we might want to pad with the first value, not with 0s... That way, the SMA gets closer and closer to the first value the closer `$sma_start_timestamp` gets to `$start_timestamp`

pinc/graph_data.inc
```diff
@@ -650,64 +650,45 @@ function pages_daily($tally_name, $c_or_i, $timeframe)
...
-        while ($result = mysqli_fetch_assoc($res)) {
-            if ($earliest_timestamp === null) {
-                $earliest_timestamp = $result["timestamp"];
+        foreach ($deltas as $timestamp => $delta) {
+            if ($timestamp < $start_timestamp) {
+                continue;
             }
-
-            $average_lookup[date("Y-m-d", $result["timestamp"])] = $result["sma"];
+            $average_lookup[date("Y-m-d", $timestamp)] = array_shift($smas);
```
bpfoley, line 680: For what it's worth, I suspect this date formatting will be relatively costly to do in bulk. Could we just use `$day = floor($timestamp / 86400);` to calculate (integer) days since epoch instead?

pinc/graph_data.inc
```diff
@@ -857,3 +838,16 @@ function query_graph_cache($function, $args = [], $expiration = 3600)
...
     return memoize_function($function, $args, $expiration, get_desired_language());
 }
+
+/**
+ * Given an array of numbers and a window, calculate the moving average over
+ * the window.
+ */
+function calculate_sma(array $array, int $period): array
+{
+    $results = [];
```
bpfoley, line 848: This is quite an inefficient way of doing this: you're doing `N` `array_slice`s and `N*period` additions to get your SMA. A better way is to calculate the sum / $period to give you the first result, and then for each subsequent result subtract off the oldest element (divided by $period) and add the newest element (divided by $period).

(That said, we should probably profile the code before doing optimizations like this.)

pinc/graph_data.inc
cpeel: Maybe, but 4 calls to this function load in under a second on TEST, which is operating under the same quantity of data in PROD at this point in the script, so it doesn't feel like it's worth the optimization.

pinc/graph_data.inc
cpeel: See aforementioned "this is already stupid fast" comment.

cpeel:
> For the linked example, the averages are in the thousands, so the effect of rounding would be a small fraction of a pixel. So something else must be making them look different.

There is probably another bug, but the current charts are 100% only using integers for the SMA. Here's the live code with a window between 2007-01-01 and 2007-12-01 -- it's all either 0 or 1. Every timeframe I've looked at with the current code shows only integer SMA values.

Screen Shot 2024-04-21 at 5 30 48 PM

pinc/graph_data.inc
cpeel: But then we skew the SMAs for all the subsequent counts over the next 21 days because there weren't $first_value pages proofread before we started collecting stats. 0 isn't right, but $first_value isn't either (and 0 is easier). The "right" think is probably to not an SMA for the first 21 days, or to have a sliding window for the first 0 to 21 days.

srjfoo: > The "right" think is probably to not an SMA for the first 21 days, or to have a sliding window for the first 0 to 21 days.

That's what makes the most sense to me.

cpeel:
I've pulled some of the stats into a spreadsheet and at least for non-beginning-of-time my SMA numbers are off by 2 days somehow. Working on fixing this.

jmdyck:
> There is probably another bug, but the current charts are 100% only using integers for the SMA.

No argument there, the `round` in the current query does that. I'm just saying the effect is invisible in the linked graphs, so doesn't explain any of their differences. So there's *definitely* another bug, either in the current code or this PR's code or both. (And I think it's both.)

jmdyck:
> I've pulled some of the stats into a spreadsheet and at least for non-beginning-of-time my SMA numbers are off by 2 days somehow. Working on fixing this.

Note that snapshots are ascribed to the preceding midnight, so if you convert that to a date, it's the day *after* the day whose activity 'led to' that snapshot. (See `$SECONDS_TO_YESTERDAY` in page_tally.inc.) So that might account for 1 of your 2 days. Not sure about the other. Time is a pain.

pinc/graph_data.inc
jmdyck: Right, padding with first-value or with zero gives unwarranted weight to that value until the padding falls out of the window.

> or to have a sliding window for the first 0 to 21 days

(Well, we use a sliding window for *all* days.) I'm guessing what you mean is just average whatever real data exists in the window. (E.g., to get the SMA for day 5, just take the average of days 1 through 5.) That'd be my choice.

pinc/graph_data.inc
jmdyck: I think that `floor($timestamp / 86400)` could work for PGDP, because we're far enough from the prime meridian that variations due to DST don't change the GMT-day within which server midnight falls.

I don't think it would work for a DP site whose server time was British Time (with DST). E.g., consider the day they switch to BDST (spring forward, lose an hour): the midnight that starts that server-day occurs at 0:00 GMT, but the midnight that ends that server-day is at 23:00 GMT *of the same GMT day*, so `floor($timestamp / 86400)` would be the same for the two different midnights. Conversely, when they fall back (gain an hour), there'd be a GMT-day that appeared to be skipped by that formula.

pinc/graph_data.inc
```diff
@@ -857,3 +838,16 @@ function query_graph_cache($function, $args = [], $expiration = 3600)
 {
     return memoize_function($function, $args, $expiration, get_desired_language());
 }
+
+/**
+ * Given an array of numbers and a window, calculate the moving average over
+ * the window.
+ */
+function calculate_sma(array $array, int $period): array
```
jmdyck, line 846: Given the function-comment, it would maybe make more sense to rename `$period` to `$window`.

pinc/graph_data.inc
```diff
@@ -857,3 +838,16 @@ function query_graph_cache($function, $args = [], $expiration = 3600)
...
+{
+    $results = [];
+    for ($i = 0; $i < (count($array) - $period); $i++) {
```
jmdyck, line 849: I think that `<` should be `<=`, otherwise you're skipping the window that ends at the end of the array.

(This might be why the example graphs show no SMA for the last day. Mind you, they also show no Goal for the last day, which presumably can't be pinned on this bug.)

cpeel:
To track down the off-by-2(!) problem I pulled out the data gathering of the SMA-based graphs into a separate function. I've changed the SMA function such that at the beginning of a set of data when it doesn't have the full window to average it will average just the values it has. This will make the first 21 days of stats have a better value.

I confirmed via data dump and spreadsheet math that this now looks correct.

I would appreciate a re-review since the code got revamped a fair amount although the overall approach is the same.

windymilla (approved):

srjfoo (approved):
The visual difference is much more obvious if you compare the P2-F2 graphs. P1 is artificially elevated to the point where you can't see small differences.

jmdyck:
The P1 "Pages Done Per Day (since stats began)" graph in the sandbox is now showing a non-zero SMA toward the end, so that's good.

However, there's still something wrong with the P1 day-by-day graphs. E.g., consider again 2024-04-01's massive count (13,692), which should increase the SMA by about 652 (13692/21), specifically from ~2609 to ~3261. (There's also a count from 03-11 that *leaves* of the window, but it's negligible.)
In the status quo, we see that increase (or something like it) plotted over the width of the bar for 04-01. But in the sandbox, the 3261 point is plotted at the *left* edge of the bar, so we don't see the uptick over that day. (The uptick *would* appear over the last day in the March graph, except that the SMA line stops a day short.)

Given that the SMA is calculated as of the end of the day, I think it should be plotted at the *right* edge of the corresponding bar. (Or else in the middle of the bar, that could work too.)

cpeel:
I think the confusion is that the left side of the bar _is_ the day, not the middle or right, so the rise happens up to the left side of the bar -- that's just how our graphs work. Which means the increase on 04/01 is accounted for, it's just that you don't see the uptick because we don't show 03/31.

All of the math still looks correct to me. You're welcome to access the raw data from https://www.pgdp.org/~cpeel/c.branch/improve-sma/tally-validation.php (which outputs the return from the specified sources) and plug it into a spreadsheet and do the rolling average to confirm though, maybe I'm missing something.

jmdyck:
> I think the confusion is that the left side of the bar _is_ the day, not the middle or right,

Hm. The tick+label for each day is aligned with the middle of the bar for that day. To me, that strongly suggests, if any single x-value 'is' the day, that it's the middle of the bar.

> so the rise happens up to the left side of the bar -- that's just how our graphs work`*`. Which means the increase on 04/01 is accounted for, it's just that you don't see the uptick because we don't show 03/31.

More specifically, the March graph does plot the 03-31 SMA value, but doesn't show the line (across the 03-31 bar) going to the 04-01 SMA value.  (I.e., in a month of 31 days, you'd want to plot 32 data points for SMA.)

> All of the math still looks correct to me. You're welcome to access the raw data from https://www.pgdp.org/~cpeel/c.branch/improve-sma/tally-validation.php (which outputs the return from the specified sources) and plug it into a spreadsheet and do the rolling average to confirm though, maybe I'm missing something.

No, I think the math is probably fine, it sounds like I'm just complaining about weirdness in either the graphing library or how we use it.

> `*` and we could probably change this to have lines plotted in the middle of the bars or something, but that's outside the scope of this PR.

Yup, that's fair.

cpeel:
Fixed minor min/max error which just pulled more data than was necessary when calculating SMA.

PR #1197, opened 2024-04-25
Basic proofreading interface using proof_api_10

70ray:
For information only, showing how the endpoints can be used in a PI.

The only files different to proof_api_10 are:
scripts/text_view.js
tools/proofers/proof.js
tools/proofers/proof.php

Sandbox at: https://www.pgdp.org/~rp31/c.branch/ui_10a3

### DIFF
--- a/SETUP/tests/ApiTest.php
+++ b/SETUP/tests/ApiTest.php
@@ -4,6 +4,75 @@
 
 class ApiTest extends ProjectUtils
 {
+    //---------------------------------------------------------------------------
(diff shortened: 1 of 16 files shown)


cpeel:
This is a very cool technical proof of concept for how the new API endpoints can be used, thanks for setting it up.

I'm going to close the PR now that the API endpoint PR has been merged and we can start to work on the UX design that will front-end the endpoints at the user-level before we work our way down to the code level.

PR #1202, opened 2024-05-01
Reduce the overhead of Activity->user_access() calls

cpeel:
To investigate something else I added a DB tracing feature to optionally output SQL queries and their callers to a log file (commit 1). During that investigation I discovered that we're doing the same query against `quiz_passes` table and `current_tallies` multiple times in a page load. I traced that back to the `Activity->user_access()` call.

This PR removes a redundant call in `gradual.inc` (this is the second time we load that object in the function -- see line 335) and adds a small cache within `Activity->user_access()` for the user score. The cache removes 20 DB queries on each of the round pages and Activity Hub. The queries are fast but add unnecessary load to the DB for data we literally just asked for microseconds before.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/reduce-db-query-hits/

### DIFF
--- a/pinc/DPDatabase.inc
+++ b/pinc/DPDatabase.inc
@@ -11,11 +11,11 @@ class DBQueryError extends Exception
 final class DPDatabase
 {
     private static $_connection = false;
-    private static $_db_name = null;
-    private static $_default_db_charset = null;
-    private static $_default_db_collation = null;
-    public static $skip_encoding_check = false;
-    public static $tracing_log = null;
+    private static ?string $_db_name = null;
+    private static ?string $_default_db_charset = null;
+    private static ?string $_default_db_collation = null;
+    public static bool $skip_encoding_check = false;
+    public static ?string $tracing_log = null;
 
     public static function connect()
     {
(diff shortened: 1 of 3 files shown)


chrismiceli (approved):

srjfoo (approved):

pinc/DPDatabase.inc
```diff
@@ -15,6 +15,7 @@ final class DPDatabase
     private static $_default_db_charset = null;
     private static $_default_db_collation = null;
     public static $skip_encoding_check = false;
+    public static $tracing_log = null;
```
bpfoley, line 18: Nitpick, but could we get into the habit of adding type annotations?
```public static ?string $tracing_log = null;```

File: pinc/DPDatabase.inc [hunk 2]
```diff
@@ -189,6 +191,30 @@ final class DPDatabase
         return mysqli_fetch_assoc($result) !== null;
     }
 
+    private static function log_trace($sql)
```
bpfoley, line 194: ```private static function log_trace(string $sql)```

bpfoley (approved):

pinc/DPDatabase.inc
cpeel: I've added a chaser commit that adds type defs to all class variables and function definitions. Might as well do it all at once. Sandbox has been updated.

We use all of these all over the place so phpstan + smoketest + unit tests + "does loading pages work" should be sufficient.

srjfoo (approved):

chrismiceli (approved):

70ray (approved):

PR #1203, opened 2024-05-02
Replace dc.xml with metadata.json and drop marc_records

cpeel:
The reason why Dublin Core (`dc.xml`) files were originally created is unclear. Users (PMs, PPs, and squirrels) currently use them when managing project directory contents to make sure they are in the right place. Replace them with the `metadata.json` files we already generate during archiving.

This allows us to stop storing and managing MARC records in the database. We could have opted to just have the `dc.xml` file be updated from the project table rather than the MARC record (we only pull 2 fields from it), but we are already generating the `metadata.json` file during archiving and there doesn't seem to be a good reason to keep both.

This code does not proactively create `metadata.json` files for existing projects until they are edited (or archived), nor does it remove any existing `dc.xml` files.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/remove-marc-records/

Testing hints:
* The `metadata.json` file should be created and updated when editing projects.
* Creating a project from the external search (eg MARC record) should still work fine, we just don't save the MARC record itself.

### DIFF
--- a/SETUP/db_schema.sql
+++ b/SETUP/db_schema.sql
@@ -119,17 +119,6 @@ CREATE TABLE `job_logs` (
 );
 
 --
--- Table structure for table `marc_records`
---
-
-CREATE TABLE `marc_records` (
-  `projectid` varchar(22) NOT NULL default '',
-  `original_array` text NOT NULL,
-  `updated_array` text NOT NULL,
-  PRIMARY KEY (`projectid`)
-);
-
---
 -- Table structure for table `news_items`
 --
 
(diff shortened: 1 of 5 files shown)


chrismiceli (approved):

SETUP/upgrade/21/20250101_drop_marc_records.php
```diff
@@ -0,0 +1,18 @@
+<?php
```
srjfoo, line 1: If I understanding the naming convention correctly, should the script be `20240501_drop_marc_records.php`?

cpeel: Yes indeed - I'll get that fixed.

windymilla (approved):
Code looks good, and `metadata.json` file contained correct info after editing a project.

SETUP/upgrade/21/20250101_drop_marc_records.php
cpeel: Upgrade script renamed and force-pushed to update PR.

srjfoo (approved):

PR #1207, opened 2024-05-05
Ensure unittests are run against a test-only DB

cpeel:
When running unit tests, run them against a test DB instead of the real DB. This doesn't impact the CI at all, but does help when running unit tests on the TEST server. While the tests try to clean up after themselves, it's impossible to do so cleanly, for instance the crazy high round page counts.

One downside to this approach on TEST is that its possible the databases will drift. It's easy enough to drop and recreate the database however if the unit tests fail because of said drift.

### DIFF
--- a/.github/actions/setup-mysql-db/action.yml
+++ b/.github/actions/setup-mysql-db/action.yml
@@ -7,9 +7,9 @@ runs:
     - name: Configure mysql
       shell: bash
       run: |
-        mysql -uroot -proot -e "CREATE DATABASE dp_db CHARACTER SET utf8mb4;"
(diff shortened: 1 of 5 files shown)


srjfoo (approved):

chrismiceli (approved):

bpfoley (approved):
Very much approve of this. I didn't like the way I was relying on the the tables created in `test_tables.sql` being compatible with the `dp_db` tables when I was doing manual testing. It didn't occur to me that it was as simple as doing a `DPDatabase::close`/`DPDatabase::connect` in the test bootstrap.

PR #1208, opened 2024-05-05
Add proofreading API examples to the user's guide

cpeel:
Update the API User's Guide to include examples for the proofreading endpoints. These examples are from my comment on the PR https://github.com/DistributedProofreaders/dproofreaders/pull/975#pullrequestreview-2026258337

### DIFF
--- a/api/USERS_GUIDE.md
+++ b/api/USERS_GUIDE.md
@@ -282,3 +282,58 @@ current holds for the project:
     "P1.proj_avail"
 ]
 ```
+
+### Proofreading
+
+The proofreading APIs require retaining and passing in state information. The
+following examples use [jq](https://jqlang.github.io/jq/) to parse responses
+and store them as environment variables.
+
+Get a project's details and parse out the project state:
+```bash
+curl -s -H "Accept: application/json" -H "X-API-KEY: $API_KEY" \
+    -X PUT "https://www.pgdp.org/api/v1/projects/projectID44de3936807f1" \
+    > project.json
+
+PROJECT_STATE=$(cat project.json | jq -r '.state')
+```
+
+Check out a page and parse out the page name, state, and page text:
+```bash
+curl -s -H "Accept: application/json" -H "X-API-KEY: $API_KEY" \
+    -X PUT "https://www.pgdp.org/api/v1/projects/projectID44de3936807f1/checkout?state=$PROJECT_STATE" \
+    > proof_page.json
+
+PAGE_NAME=$(cat proof_page.json | jq -r .pagename)
+PAGE_STATE=$(cat proof_page.json | jq -r .pagestate)
+cat proof_page.json | jq '{text}' > page_text.json
+```
+
+Validate the page text doesn't have any weird characters:
+```bash
+curl -s -H "Accept: application/json" -H "X-API-KEY: $API_KEY" \
+    -X PUT "https://www.pgdp.org/api/v1/projects/projectID44de3936807f1/validatetext" \
+    -d @page_text.json
+```
+
+Save the page as in-progress:
+```bash
+curl -s -H "Accept: application/json" -H "X-API-KEY: $API_KEY" \
+    -X PUT "https://www.pgdp.org/api/v1/projects/projectID44de3936807f1/pages/$PAGE_NAME?state=$PROJECT_STATE&pagestate=$PAGE_STATE&pageaction=save" \
+    -d @page_text.json
+```
+
+Get the current page state:
+```bash
+curl -s -H "Accept: application/json" -H "X-API-KEY: $API_KEY" \
+    -X GET "https://www.pgdp.org/api/v1/projects/projectID44de3936807f1/pages/$PAGE_NAME" \
+    > page_info.json
+
+PAGE_STATE=$(cat page_info.json | jq -r .pagestate)
+```
+
+Save the page as done:
+```bash
+curl -s -H "Accept: application/json" -H "X-API-KEY: $API_KEY" \
+    -X PUT "https://www.pgdp.org/api/v1/projects/projectID44de3936807f1/pages/$PAGE_NAME?state=$PROJECT_STATE&pagestate=$PAGE_STATE&pageaction=checkin"
+```
--- a/api/USERS_GUIDE.md
+++ b/api/USERS_GUIDE.md
@@ -295,7 +295,7 @@ curl -s -H "Accept: application/json" -H "X-API-KEY: $API_KEY" \
     -X PUT "https://www.pgdp.org/api/v1/projects/projectID44de3936807f1" \
     > project.json
 
-PROJECT_STATE=$(cat project.json | jq -r '.state')
+PROJECT_STATE=$(cat project.json | jq -r .state)
 ```
 
 Check out a page and parse out the page name, state, and page text:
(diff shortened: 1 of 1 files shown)


api/USERS_GUIDE.md
```diff
@@ -281,4 +281,59 @@ current holds for the project:
...
+```
+
+Check out a page and parse out the page name, state, and page text:
+```bash
+curl -s -H "Accept: application/json" -H "X-API-KEY: $API_KEY" \
+    -X PUT "https://www.pgdp.org/api/v1/projects/projectID44de3936807f1/checkout?state=$PROJECT_STATE" \
+    > proof_page.json
+
+PAGE_NAME=$(cat proof_page.json | jq -r .pagename)
+PAGE_STATE=$(cat proof_page.json | jq -r .pagestate)
```
srjfoo, line 308: I took a look at "Invoking jq" in the manual (wow, that's a long file!), and one of the things it says there is

> Note: it is important to mind the shell´s quoting rules. As a general rule
       it´s best to always quote (with single-quote characters on Unix shells) the
       jq program, as too many characters with special meaning to jq are also
       shell meta-characters.

At line 298 above, you quote `'.state'`, and below, at line 309, you quote `'{text}'`. Is there a reason (or is it optional?) that `.pagename` and `.pagestate` aren't quoted?

To be fair, it's like this in the original example in PR #1208, and it worked (I just didn't notice, and didn't check the manual on jq at the time).

cpeel: The quotes are only required if there are special characters (like `{}` or `$`, etc) so the `'{text}'` requires the quotes and the other places are optional. I'll remove the ones around `'.state'` since those aren't required.

srjfoo (approved):

chrismiceli (approved):

api/USERS_GUIDE.md
bpfoley: This is a total nitpick, so feel free to ignore, but you can always replace things like 
```PAGE_STATE=$(cat proof_page.json | jq -r .pagestate)```
with
```PAGE_STATE=$(jq -r .pagestate < proof_page.json)```
or, if you prefer it
```PAGE_STATE=$(< proof_page.json jq -r .pagestate)```

File: api/USERS_GUIDE.md [hunk 2]
```diff
@@ -281,4 +281,59 @@ current holds for the project:
...
+### Proofreading
+
+The proofreading APIs require retaining and passing in state information. The
+following examples use [jq](https://jqlang.github.io/jq/) to parse responses
+and store them as environment variables.
+
+Get a project's details and parse out the project state:
+```bash
+curl -s -H "Accept: application/json" -H "X-API-KEY: $API_KEY" \
+    -X PUT "https://www.pgdp.org/api/v1/projects/projectID44de3936807f1" \
```
srjfoo, line 295: This should be `GET`, shouldn't it? (Didn't notice it on my first readthrough, but sure noticed it just now trying to test it.)

api/USERS_GUIDE.md
```diff
@@ -281,4 +281,59 @@ current holds for the project:
...
+    -X GET "https://www.pgdp.org/api/v1/projects/projectID44de3936807f1/pages/$PAGE_NAME" \
+    > page_info.json
+
+PAGE_STATE=$(cat page_info.json | jq -r .pagestate)
+```
+
+Save the page as done:
+```bash
+curl -s -H "Accept: application/json" -H "X-API-KEY: $API_KEY" \
+    -X PUT "https://www.pgdp.org/api/v1/projects/projectID44de3936807f1/pages/$PAGE_NAME?state=$PROJECT_STATE&pagestate=$PAGE_STATE&pageaction=checkin"
```
srjfoo, line 338: This appears to be missing the continuation `\` and the

```
   -d @page_text.json
```

Though not strictly necessary, if going by the testing set in #975, also missing the trailing page state query, if the intent is to double-check that the page state is what it's supposed to be:

```
# get the current page state
curl -s -H"Accept: application/json" -H "X-API-KEY: $API_KEY" \
    -X GET "https://www.pgdp.org/api/v1/projects/$PROJECT/pages/$PAGE_NAME" \
    > page_info.json

PAGE_STATE=$(cat page_info.json | jq -r .pagestate)
```

api/USERS_GUIDE.md
```diff
@@ -281,4 +281,59 @@ current holds for the project:
...
+curl -s -H "Accept: application/json" -H "X-API-KEY: $API_KEY" \
+    -X PUT "https://www.pgdp.org/api/v1/projects/projectID44de3936807f1/checkout?state=$PROJECT_STATE" \
+    > proof_page.json
+
+PAGE_NAME=$(cat proof_page.json | jq -r .pagename)
+PAGE_STATE=$(cat proof_page.json | jq -r .pagestate)
+cat proof_page.json | jq '{text}' > page_text.json
+```
+
+Validate the page text doesn't have any weird characters:
```
70ray, line 312: This is fine but I'd intended this to be used only if you had found there were weird characters when save (as in progress or as done) returned an error saying there were weird characters. It returns not only vlaid as true or false but the text marked up to show the weird characters.

api/USERS_GUIDE.md
cpeel: Yes it should, thanks for catching it.

api/USERS_GUIDE.md
cpeel: Also good catch on the missing `-d @page_text.json` -- I failed at the copy/paste/update from the PR.

I didn't include a follow-up GET because it's not strictly necessary and we'd already given an example of that prior in the guide.

api/USERS_GUIDE.md
srjfoo: For my own information, if the only extra piece of information that is returned by the separate `validate` function is the list of weird characters, why not do that as part of the save process?

70ray: There are really two questions here:
1. Why not return the weird character information when saving?
2. Why not just return a list of weird characters instead of a marked up text?

For the first question, you could equally well do it like that and some earlier versions did and it would be more efficient in the sense that you do not have to submit the text twice. The only reason not to is that if there are weird characters the function will fail to save the submitted text so it should arguably return a 400+ code. You can still return the information with a 400+ code but it seemed simpler not to. There are no good reasons for not doing it like that if you wanted to.

The second question I had not considered as such. The way the present system works is that the client gets a regular expression from the server to use to check for invalid characters which seems a bit messy because javascrpt regexs work a bit differently to php regexs so it seemed better to let the server do the work and send the result back to the client. If you want to display a text with the characters highlighted then if you just return a list of characters then the client would have to search for them which is possible but just a different way to do it. There is also the question of complicated situations involving multiple unicode codepoints and graphemes where the client and server would have to agree on exactly what constituted a grapheme. If it was a large text then it could be better to do it like that so as not to have to return so much data.

srjfoo: Yeah, predictability of regex engines can trip us up if we're not careful. Thanks for the explanation. I suspect we could get into an interesting philosophical discussion about the whole issue, but I don't think this is the place to. 😁

srjfoo (approved):

70ray (approved):

PR #1210, opened 2024-05-09
Don't die if metadata.json creation fails

cpeel:
Editing project information can happen even if the project directory doesn't exist (the project has been deleted or archived). In those cases, just don't generate the `metadata.json` file instead of failing.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/fix-edit-deleted-project/

### DIFF
--- a/pinc/Project.inc
+++ b/pinc/Project.inc
@@ -634,7 +634,12 @@ class Project
         }
 
         // Create or update the metadata file
-        $this->generate_metadata_json();
+        try {
+            $this->generate_metadata_json();
+        } catch (NoProjectDirectory $exception) {
+            // this is likely because the project has been deleted, either way
+            // continue instead of failing
+        }
     }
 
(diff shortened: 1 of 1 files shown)


pinc/Project.inc
```diff
@@ -634,7 +634,12 @@ class Project
         }
 
         // Create or update the metadata file
-        $this->generate_metadata_json();
+        try {
+            $this->generate_metadata_json();
```
chrismiceli, line 638: would it make more sense to check in the `generate_metadata_json` method?

chrismiceli (approved):

pinc/Project.inc
cpeel: We actually raise the exception in that function, I did it this way so that the function wouldn't silently fail.

srjfoo (approved):

PR #1211, opened 2024-05-11
Make "Available" tab mean "available to work in right now"

cpeel:
The original selection criteria for the "Available" tab on the My Projects page was very broad: the project was in some available state. That's not really useful for most users who are thinking of "available" as available for them to work on right now.

Change the tab to reflect this smaller set of available-to-them projects and make it the first tab as it's arguably the most useful now. Reminder: the My Projects page saves the last tab the user is on, so moving this first isn't going to impact most users but will help newbies.

Note that the My Suggestions page handles these cases already.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/my-projects-available/

### DIFF
--- a/tools/proofers/my_projects.php
+++ b/tools/proofers/my_projects.php
@@ -18,6 +18,7 @@ if (user_is_a_sitemanager() || user_is_proj_facilitator()) {
         die("Invalid username.");
     }
 }
+$focus_user = new User($username);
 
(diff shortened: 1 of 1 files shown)


chrismiceli (approved):

srjfoo (approved):
Note for other testers: I hacked the sandbox so that `page_tally_threshold` is 20 pages instead of 500.

PR #1212, opened 2024-05-12
Small changes to prepare for PHP 8.3

cpeel:
This PR includes some prep work for PHP 8.3 that will work with our current PHP version too.
* Update our CI to include tests against PHP 8.3 so we can detect issues with it before we formally upgrade to it later this year.
* Update some string functions to be UTF-8 aware. PHP 8.2 made a backwards incompatible change where [many string functions that depend on case are no longer locale-aware](https://www.php.net/manual/en/migration82.incompatible.php#migration82.incompatible.standard). I've combed through the code and found a few instances where we should really be using UTF-8 versions of these functions. We have other cases where those functions are used but only against ASCII strings so those instances were not changed.
* PHP 8.3 introduced the `mb_str_pad()` function but rather than conditionally defining it based on PHP version, just use the UTF-8 version instead. This only impacts `mktable.php`.

These are likely not the exhaustive set of required changes for PHP 8.3. We'll focus on finding those in a few months after we upgrade a clone of TEST to use it.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/prep-for-php-8.3/

Testing suggestions:
* The [Table Maker](https://www.pgdp.org/~cpeel/c.branch/prep-for-php-8.3/tools/proofers/mktable.php) accessible in the proofreading interface could use some testing with UTF-8 characters.
* All of the PM tools to manage WordCheck suggestions were updated but only to sort the list of words so minimal testing on these pages is necessary.


chrismiceli (approved):

quiz/generic/quiz_page.inc
```diff
@@ -215,8 +215,8 @@ function qp_text_contains_anticipated_error($text, $test)
                 $part = strstr($text, $test["starttext"]);
                 $part = substr($part, $len, strpos($part, $test["stoptext"]) - $len);
             } else {
-                $part = stristr($text, $test["starttext"]);
-                $part = substr($part, $len, stripos($part, $test["stoptext"]) - $len);
+                $part = voku\helper\UTF8::stristr($text, $test["starttext"]);
```
70ray, line 218: Why not use mb_stristr here? Wouldn't it be better to use a php function rather than a 3rd party one?
Simiarly stripos etc.

cpeel: We could, but we've standardized on using the UTF8 package for all our multibyte work rather than `mb_*` calls (or `iconv` calls) directly and needing to provide the string encoding (which is always UTF-8). This was an conscious decision during the Unicode conversion.

Under the covers the UTF8 package is doing the necessary `mb_*` or `iconv` or other calls but it provides a consistent interface.

srjfoo (approved):
I don't know that I found all the word list management places, but I tried several ways of adding UTF-8 words to the GWL (and in one case the BWL), and all sorted properly as far as I could tell. (And I guess for some tables, the table-maker is better than nothing, but it's really painful to use!)

theshriek (approved):
I tested the adding of an UTF-8 word to the good words list. 

1. I added it to the good words list when creating the project.
2. I went to the proofreading interface and requested it there, and then added to the list.

I am not sure if there is any difference between those two, but in both instances the word was correctly alphabetized.

I did not test the table maker.

70ray (approved):
Tested table maker

PR #1213, opened 2024-05-13
Replace die()s with exceptions and fix an assert

cpeel:
`die()`s in functions are a Very Bad Idea because there's no way for callers to capture and do something about the error. They also can cause silent failures by exiting the page without anyone else knowing. This replaces `die()`s in our include files with exceptions. This does not replace `die()`s in non-include files. This shouldn't need a lot of (any?) testing as `die()`s are effectively the same as thrown exceptions from a "stop the world" perspective.

Also replace an `assert()` that should always be checked. Asserts are only for dev & test and in "production" mode the code within them is not generated or evaluated at all. See https://www.php.net/manual/en/function.assert.php

Sandbox: https://www.pgdp.org/~cpeel/c.branch/die-die-and-asserts/

### DIFF
--- a/SETUP/smoketests/pageload_smoketest.py
+++ b/SETUP/smoketests/pageload_smoketest.py
@@ -83,7 +83,7 @@ MISC_TESTS = [
 
 QUIZ_TESTS = [
     {'path': 'quiz/start.php'},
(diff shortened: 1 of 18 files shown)


chrismiceli (approved):

70ray (approved):

srjfoo (approved):
If there is a way to test these, let me know; I couldn't really tell from reading the code. Code LGTM.

PR #1214, opened 2024-05-16
Update php-cs-fixer to 3.57.1

bpfoley:
After a looong review process and a couple of fast followup fixes, this introduces parallel linting.

On an M2 MacBook Pro (using 12 cores) lint times go from 16.5s to 2.1s.

On TEST (using 2 cores on a very underpowered AWS instance) lint times go from 16.8s to 15.6s.

This speeds up CI tests on GitHub somewhat (unscientifically a before and after CI run goes from ~85s to ~55s), and as long as it doesn't break anyone's development workflow, shouldn't need any additional testing.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -539,6 +539,70 @@
     ],
     "packages-dev": [
         {
+            "name": "clue/ndjson-react",
+            "version": "v1.3.0",
+            "source": {
+                "type": "git",
+                "url": "https://github.com/clue/reactphp-ndjson.git",
+                "reference": "392dc165fce93b5bb5c637b67e59619223c931b0"
+            },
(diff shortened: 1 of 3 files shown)


composer.lock
```diff
@@ -538,6 +538,70 @@
         }
     ],
     "packages-dev": [
+        {
```
chrismiceli, line 541: are we just updating transitive dependency versions or did you forget to include a change to `compose.json` in this review?

chrismiceli (approved):

chrismiceli (approved):

cpeel (approved):
Nice!

srjfoo (approved):

PR #1215, opened 2024-05-18
Reorganize code test layout

cpeel:
This tries to bring some coherence to our tests which have sprawled a bit. Most of this PR is just renames but there were some path updates in a couple of places. `qunit.html` had so many changes it looks like a delete/create in git vs a rename/update.

@chrismiceli - I'm not sure if what I've done with the JS tests (shoving all the JS files into a single directory) is a best-practice there or if there is a better way to do that, your thoughts on that would be appreciated.


chrismiceli (approved):
I think the way we structure the js tests are fine for now in my opinion. We can create multiple test suites etc. later if we want to break up the files and make CI failures easier to pinpoint.

bpfoley:
LGTM. Making the paths/filenames a bit more systematic can't hurt.

bpfoley (approved):

srjfoo (approved):

70ray (approved):

PR #1218, opened 2024-05-19
Refine site search based on user feedback

cpeel:
Refine the new site search experience after input from users. In brief, this:
* Removes the "jump to username" feature if the string is a username, now this needs to be prefixed with `u:`.
* Adds an icon to the search menu to get to the site search help page
* If no prefix was used the search page now lists suggestions to help the user (thanks @windymilla for this suggestion)
* Enables searching for a task by summary in addition to string (thanks @srjfoo for the suggestion)

Sandbox: https://www.pgdp.org/~cpeel/c.branch/refine-site-search/

Forum discussion: https://www.pgdp.net/phpBB3/viewtopic.php?t=81458

### DIFF
--- a/tools/site_search.php
+++ b/tools/site_search.php
@@ -3,7 +3,6 @@ $relPath = "./../pinc/";
 include_once($relPath.'base.inc');
 include_once($relPath.'theme.inc');
 include_once($relPath.'Project.inc');
-include_once($relPath.'User.inc');
 
 require_login();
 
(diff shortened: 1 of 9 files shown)


srjfoo (approved):
FWIW, the task search actually searches in both the summary and the details, not just the summary. I think the description is fine. It's already the longest one :D), but thought I'd clarify that.

windymilla (approved):
LGTM!

chrismiceli (approved):

tools/site_search.php
```diff
@@ -44,6 +42,17 @@
 echo "<input type='submit' value='" . attr_safe(_("Search")) . "'>";
 echo "</form>";
 
+if ($query && $query != "help") {
+    echo "<p class='warning'>" . sprintf(_("Not sure what to do with '%s', perhaps try one of these."), $query) . "</p>";
```
bpfoley, line 46: I think we want to `html_safe($query)` here and below

bpfoley (approved):

tools/site_search.php
cpeel: Oh yes - good catch! Done and squashed into the relevant (last) commit in the branch.

70ray (approved):

PR #1220, opened 2024-05-21
Embed garamond font

cpeel:
Bundle EB Garamond rather than relying on Google's `gstatic.com`. This removes yet another live 3rd-party dependency.

Back in 2020 when this was added it wasn't clear from Google's Fonts licensing if this is possible, but it's very clear now as:
1. they offer an option to download the fonts
2. provide docs on how to self-host it
3. This copy of EBGaramond is explicitly under the SIL according to its [Github account](https://github.com/octaviopardo/EBGaramond12).

https://www.pgdp.org/~cpeel/c.branch/embed-garamond-font/


srjfoo (approved):

cpeel:
After a Slack discussion I've changed the TTF to a WOFF2 which makes it smaller both in git and on the wire.

bpfoley (approved):
Definitely an improvement on every front: no/minimal additional download cost to the client; a heavily cached resource; no potential for Google cookie tracking; and the additional bandwidth cost on the server should be minimal, considering how many project page images we serve up to users anyway.

chrismiceli (approved):

70ray (approved):

PR #1225, opened 2024-06-04
Make polytonic greek accents visible in proofreading interface etc.

70ray:
Sandbox at: https://www.pgdp.org/~rp31/c.branch/pi_indent

### DIFF
--- a/tools/proofers/PPage.inc
+++ b/tools/proofers/PPage.inc
@@ -237,7 +237,7 @@ class PPage
         if ($font_size != '') {
             $font_size_string = "font-size: $font_size;";
         }
-        echo "style=\"font-family: $font_family; $font_size_string padding-left: 0.5em;\" ";
+        echo "style=\"font-family: $font_family; $font_size_string\" ";
 
         if (!$line_wrap) {
             echo "wrap='off' ";
--- a/tools/proofers/PPage.inc
+++ b/tools/proofers/PPage.inc
@@ -237,7 +237,7 @@ class PPage
         if ($font_size != '') {
             $font_size_string = "font-size: $font_size;";
         }
-        echo "style=\"font-family: $font_family; $font_size_string padding-left: 0.25em;\" ";
+        echo "style=\"font-family: $font_family; $font_size_string padding-left: 0.5em;\" ";
 
         if (!$line_wrap) {
             echo "wrap='off' ";
--- a/quiz/generic/proof.php
+++ b/quiz/generic/proof.php
@@ -42,7 +42,7 @@ slim_header("", $header_args);
 <form action="./returnfeed.php?quiz_page_id=<?php echo $quiz_page_id; ?>" target="right" method="post" name="editform" id="editform">
 <textarea rows="12" cols="60" name="text_data" id="text_data" wrap="off"
     autocorrect="off" autocomplete="off" autocapitalize="off"
-    style="width:100%; <?php echo $font_settings; ?>">
+    style="padding-left:0.5em; width:100%; <?php echo $font_settings; ?>">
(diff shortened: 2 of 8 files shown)


70ray:
padding works for textareas but not for inputs. I don't know why.
text-indent only works on first line in a textarea.

srjfoo:
> padding works for textareas but not for inputs. I don't know why. text-indent only works on first line in a textarea.

I think that `text-indent` works for the textareas as long as you have a return at the end of each line. You have to add `each-line` along with the indent value. I'm not sure that `text-indent` was intended to be used the way we used it, though 😁 . I can't swear to it, though, because I was only trying it out in the browser console -- I never got as far as setting up a branch to test that out.

srjfoo:
Bearing in mind that we don't do Greek transcription in the quizzes, and don't have anything that would need the extra indent, should the padding or indent be added to the textarea in teh quizzes for consistency's sake?

70ray:
0.5em padding added to quiz textarea and sandbox updated.

cpeel (approved):
This looks good to me.

chrismiceli (approved):

srjfoo:
> 0.5em padding added to quiz textarea and sandbox updated.

Sorry for not thinking of this earlier, but I'm thinking that Format Preview should be fixed, too. For consistency, if nothing else.

tools/proofers/PPage.inc
```diff
@@ -237,7 +237,7 @@ class PPage
         if ($font_size != '') {
             $font_size_string = "font-size: $font_size;";
         }
-        echo "style=\"font-family: $font_family; $font_size_string padding-left: 0.25em;\" ";
+        echo "style=\"font-family: $font_family; $font_size_string padding-left: 0.5em;\" ";
```
srjfoo, line 240: I understand why `font-family` specs are here, but is there a reason that `padding-left` is here instead of `page_interfaces.less`?

Or is this one of those "Meh, it's been there for forever, it doesn't affect the styling of the themes, there's no good reason to change it, and it may very well change once the new PI is implemented, so why waste time on it?" situations?

srjfoo (changes requested):
The only change I'm recommending is adding the padding (or text indent if padding doesn't work) to Format Preview. The other is just a comment.

---

Tested in Safari and FF on the iPad, and FF, Safari, Opera and Chrome on a MacBook Pro running Sonoma. I have not yet done any random testing in browserstack.

srjfoo:
One more question. We've added padding left for the Greek characters. I am finding that scrollbars of the non-permanent sort overlap text on the right and bottom of the text pane. Would it be reasonable to add .25em to .3em padding there, also, to reduce the effect? (Not specifically asking for it in the PR, just something to consider.)

tools/proofers/PPage.inc
70ray: I think there probably isn't a good reason why it's not in `page_interfaces.less`. Specifying it for  `id text_data` could be better then it would work for the quiz also. I'd have to check what else text_data is used for so I was just being lazy.

70ray: grepping `id='text_data'` revealed another textarea in `tools/project_manager/handle_bad_page.php` so it would perhaps be good to put the padding in page_interfaces and remove the padding from PPage.inc. and quiz/generic/proof.php.
(Note to self: I induced some other errors while running fix bad page so should look at them later)

70ray:
Padding now added in style rather than in-line. Affects PI, quiz and handle_bad_page. Padding also for format preview.
Sandbox updated.

windymilla (approved):

srjfoo (approved):
Good catch on handle_bad_page!

Issue #1231, opened 2024-06-09
Avoid javascript type errors in userprefs.php


70ray:
A type error is occuring in the function dropdown_select_values_and_labels() in _show_credit_name_adhoc() around line 987. This is ignored unless you happen to have the developer tools window open.

cpeel:
Fixed in #1234

PR #1234, opened 2024-06-10
revise function for selecting credit name

70ray:
This addresses issue #1231.
Select elements by id rather than from form.
Change credit name input to hidden rather than disabled if not applicable.
If this works as it appears to be intended the code could be simplified by removing comments which no longer apply  and revising the function `dropdown_select_values_and_labels()` which appears to be used in only one place.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/pref_credit


chrismiceli (approved):

70ray:
Sandbox updated.

userprefs.php
```diff
@@ -951,54 +945,18 @@ function _show_credit_name_adhoc()
...
-function dropdown_select_values_and_labels($field_name, $current_value, $values, $labels, $on_change = '')
-{
-    global $event_id, $window_onload_event;
-
-    $function_name = 'event' . ++$event_id;
-    $jscode = "var f=document.forms[0];\nvar t=f.$field_name;\n$on_change";
+    $credit_options = [
+        'real_name' => _('Real Name'),
+        'username' => _('Username'),
+        'other' => _('Other:'),
```
cpeel, line 951:
```suggestion
        'other' => _('Other') . ':',
```
We don't want the `:` inside the translated string.

userprefs.js
```diff
@@ -0,0 +1,14 @@
+window.addEventListener("DOMContentLoaded", () => {
```
cpeel, line 1: Let's move this to `scripts/` so it's not hanging out here in the root. `userprefs.php` probably shouldn't be here either but that ship has sailed.

cpeel (commented):
This is _much_ better than what was there before!

70ray:
Sandbox updated.

scripts/userprefs.js
```diff
@@ -0,0 +1,14 @@
+window.addEventListener("DOMContentLoaded", () => {
+
+    let creditNameSelector = document.getElementById('credit_name');
+    let creditOtherInput = document.getElementById('credit_other');
+
+    function maybeShowOther() {
+        creditOtherInput.hidden = (creditNameSelector.value != 'other');
```
chrismiceli, line 7: *minor* unnecessary parentheses. Also I think eslint may complain or warn about not using `!==` over `!=`.

chrismiceli (approved):

srjfoo (approved):

scripts/userprefs.js
70ray: I think the parentheses make it more readable. probably better to always use !==.

chrismiceli (approved):

cpeel (approved):

Issue #1233, opened 2024-06-10
Standardize cronjob scripts and surface errors more easily


cpeel:
We currently have a selection of various cronjob in `crontab/` that we tell admins to add to a user's crontab (see `SETUP/dp.cron`) for various background processing activities.

Each of these is their own special snowflake that outputs different information. The only way to determine if one of them has failed is to look at the cron email and parse the results. Moreover, the email is the only record that they ran and the results. All of the scripts run through Apache which requires us to set a long TimeOut value (upwards of 30 to 60 minutes) to ensure these long-running scripts complete.

We need to create a standard way to run these jobs such that:
* We have a non-cron-email record that they ran, when the ran, and a log of their output (see also #1232)
* We are notified via email if the script fails, but not if it succeeds
* The scripts are run outside of Apache (note that this may result in some file permissions issues to work through)

It probably makes sense to keeping using cron to manage which job runs at what time -- why recreate that wheel? However, we could have a wrapper script that runs a named job such that there's one entrypoint for these background activities.

cpeel:
One possible solution would be to create a new `BackgroundJob` class that individual cron job activities would inherit from. This parent class would handle logging the create time and end time and detecting failures.

The child classes would output any status messages to a logging file somewhere -- not stdout. Upon failure the child class would throw an exception with a useful error message.

There would then be single PHP script that accepts a class name as the first parameter. The script would check if the specified class exists and is inherited from `BackgroundJob` and if so it instantiates one and runs its `go()` function (or somesuch). If the function succeeds (no exceptions) it doesn't output anything. If it fails, it outputs information to stdout that the cronjob would then see and email to the user.

The entrypoint script would be called from the crontab.

jmdyck:
> Moreover, the email is the only record that they ran and the results.

`automodify.php` is one exception to this, as the results of each run are saved to a file in `d/stats/automodify_logs/`

Also, `automodify.php` and `take_tally_snapshots.php` add entries to the `job_logs` table to record that they ran. (There used to be a few other jobs that added entries to that table, but not since June 2005.)

>     * We are notified via email if the script fails, but not if it succeeds

Currently, I'd rather receive an empty email on success (for jobs that run daily or less often), because a lack of email could be for various reasons (cron failure, email failure, job is still running/is stuck), and I don't want those to look the same as success. However, I'm guessing that the new framework would include an easy way (dashboard?) to see the status/logs of recent cron jobs, so that would probably take care of my concern.

[...]

PR #1236, opened 2024-06-12
Add job log visibilitiy

cpeel:
Did you know we have a job log? Me neither! That's because we currently log to it in exactly two scripts and have zero way to view it. This gives SAs the ability to see the table and adds an index on the timestamp.

It removes a useless "MIDDLE" message from automodify.

https://www.pgdp.org/~cpeel/c.branch/job-log-visibilitiy/

This is related to https://github.com/DistributedProofreaders/dproofreaders/issues/1233

### DIFF
--- a/tools/site_admin/show_job_log.php
+++ b/tools/site_admin/show_job_log.php
@@ -45,13 +45,13 @@ echo "<hr>";
 echo "<table class='themed theme_striped'>";
 echo "<tr>";
 echo "<th>" . _("Timestamp") . "</th>";
-echo "<th>" . _("Filename") . "</th>";
+echo "<th>" . _("Job") . "</th>";
 echo "<th>" . _("Event") . "</th>";
 echo "<th>" . _("Comments") . "</th>";
 echo "</tr>";
 foreach (get_job_log_entries($start_timestamp, $filename, $event) as $row) {
     echo "<tr>";
-    echo "<td>" . date("Y-h-d H:i:s", $row["tracetime"]) . "</td>";
+    echo "<td>" . date("Y-m-d H:i:s", $row["tracetime"]) . "</td>";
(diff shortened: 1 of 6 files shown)


chrismiceli (approved):

srjfoo (approved):
I like it!

I don't know if this is the time or place to do it, but is it possible to have some documentation on valid filenames and events. I guessed that END would be. (😁)

tools/site_admin/show_job_log.php
```diff
@@ -0,0 +1,60 @@
...
+echo "<table class='themed theme_striped'>";
+echo "<tr>";
+echo "<th>" . _("Timestamp") . "</th>";
+echo "<th>" . _("Filename") . "</th>";
+echo "<th>" . _("Event") . "</th>";
+echo "<th>" . _("Comments") . "</th>";
+echo "</tr>";
+foreach (get_job_log_entries($start_timestamp, $filename, $event) as $row) {
+    echo "<tr>";
+    echo "<td>" . date("Y-h-d H:i:s", $row["tracetime"]) . "</td>";
```
70ray, line 54: Should be Y-m-d

70ray (changes requested):

tools/site_admin/show_job_log.php
cpeel: Good catch - fixed!

cpeel:
> I don't know if this is the time or place to do it, but is it possible to have some documentation on valid filenames and events. 

I've changed 'Filename' to 'Job' in the last commit as I anticipate us using it in a broader sense. The "event"s are very ad hoc right now. So "not yet but there's work to be done in this area".

srjfoo (approved):

70ray (approved):

PR #1242, opened 2024-06-21
Stop logging WordCheck corrections

cpeel:
Every time someone submits a correction in WordCheck, in the record of the event itself we write the corrected text to the `corrections` column in the `wordcheck_events` table. They look like this:
```
[-Laman-] {+Lamán+}
[-Komarov-] {+Komaróv+}
[-pubHcist-] {+publicist+}
[-Vyazemsky-] {+Vyázemsky+}
```
But we never use it. Ever. Nor have we ever used it. At some point (circa 2007) we thought that data might come in handy for some analysis later but here we are ~17 years later and nada.

This PR:
1. stops collecting the data
2. includes an upgrade script to drop the column

Before we drop the column from PROD I intend to do write a dump of this SQL query to a file in some structured format:
```sql
select projectid, image, round_id, corrections from wordcheck_events where corrections <> "";
```
That data might be useful to some researcher somewhere who is interested in common scannos -- the projectid could give them information on the language and if they really cared they could reference it back to the scan and page text. I'll do the same for the `wordcheck_events` in the archive table too (dump the data and then drop the column).

Sandbox to confirm I didn't break WordCheck itself from the corrections excising: https://www.pgdp.org/~cpeel/c.branch/stop-wc-correction-logging/

### DIFF
--- a/SETUP/db_schema.sql
+++ b/SETUP/db_schema.sql
@@ -707,7 +707,6 @@ CREATE TABLE `wordcheck_events` (
   `round_id` char(2) NOT NULL,
   `username` varchar(25) NOT NULL,
   `suggestions` text,
-  `corrections` text,
   PRIMARY KEY (`check_id`),
   KEY `pc_compound` (`projectid`,`timestamp`,`image`)
 );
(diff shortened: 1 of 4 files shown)


cpeel:
@lhamilton1 - please confirm this approach sounds good to you.

lhamilton1:
I approve

srjfoo (approved):

windymilla (approved):
More stuff removed - yay!

70ray (approved):

chrismiceli (approved):

PR #1245, opened 2024-06-23
Decrease list_etexts.php overhead

cpeel:
`list_etexts.php` is one of the few public pages (does not require login to access) and its worth trying to decrease the system overhead for it. This PR does the following:
1. optimizes the SQL used to generate the page by:
   * removing 1 of the 2 queries and using a 1-hour cached value for total number of projects per star metal
   * simplifying the query for the silver and bronze stars
2. removing the sorting options on the page -- this reduces the links crawlers can hit that just generate more load and removes a feature that I don't think is really all that useful to humans.

While we're here increase the default page size to 50 from 20. This is likely more useful to users without changing the impact to the query.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/decrease-list_etexts-overhead/list_etexts.php?x=g&sort=5

### DIFF
--- a/pinc/list_projects.inc
+++ b/pinc/list_projects.inc
@@ -4,8 +4,16 @@ include_once($relPath.'pg.inc');
 /**
  * Output a list of projects giving very brief information about each
  */
-function list_projects_tersely($where_condition, $order_clause, $limit_clause)
+function list_projects_tersely(string $metal, string $order_clause, string $limit_clause)
 {
+    if ($metal == "gold") {
+        $where_condition = SQL_CONDITION_GOLD;
+    } elseif ($metal == "silver") {
+        $where_condition = SQL_CONDITION_SILVER;
+    } elseif ($metal == "bronze") {
+        $where_condition = SQL_CONDITION_BRONZE;
+    }
+
     $sql = "
         SELECT nameofwork, authorsname, language, postednum
         FROM projects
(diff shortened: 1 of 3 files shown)


cpeel:
Tagging @lhamilton1 for her review and agreement on removing the sorting feature of this page.

chrismiceli (approved):

srjfoo (approved):

pinc/list_projects.inc
```diff
@@ -37,11 +45,39 @@ function list_projects_tersely($where_condition, $order_clause, $limit_clause)
  *
  * url_base is the URL with the beginning of a query string, eg "foo.php?x=10&amp;" or "foo.php?"
  */
-function list_projects($where_condition, $order_clause, $group_clause, $url_base, $per_page = 20, $offset = 0)
+function list_projects(string $metal, string $order_clause, string $url_base, int $per_page = 20, int $offset = 0)
```
srjfoo, line 48: I know `$per_page = 20` is overridden in the function call, but if you want the default to be 50 instead of 20, should it be changed here?

windymilla (approved):
I agree it seems fine to remove the sorting - this page isn't the best place for humans to find out about books that have been completed, nor ones that are in progress.

pinc/list_projects.inc
cpeel: It actually doesn't matter since the only thing calling it is setting it to a number that isn't 20.

PR #1248, opened 2024-06-27
Add User annotation for arguments in userprefs.php

jchaffraix:
In order to have proper PHPStan coverage, first
annotated the return value from User::load_current()
then added User type annotation to userprefs.php.

This uncovered 2 missing properties on Users (`profile`
and `navbar_activity_menu`) and some fields on `UserProfile`
that were typed as `bool` but should have been `int`.

Manually tested updating a user's prefs and changing
the layout during proofing.

### DIFF
--- a/tools/proofers/processtext.php
+++ b/tools/proofers/processtext.php
@@ -347,7 +347,7 @@ try {
 function switch_layout()
 {
     $user = User::load_current();
-    $user->profile->i_layout = $user->profile->i_layout == 1 ? 0 : 1;
+    $user->profile->i_layout = $user->profile->i_layout == true ? false : true;
     $user->profile->save();
 }
 
--- a/tools/proofers/processtext.php
+++ b/tools/proofers/processtext.php
@@ -347,7 +347,7 @@ try {
 function switch_layout()
 {
     $user = User::load_current();
-    $user->profile->i_layout = $user->profile->i_layout == true ? false : true;
+    $user->profile->i_layout = $user->profile->i_layout == 1 ? 0 : 1;
     $user->profile->save();
 }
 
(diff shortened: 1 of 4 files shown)


tools/proofers/processtext.php
```diff
@@ -347,7 +347,7 @@
 function switch_layout()
 {
     $user = User::load_current();
-    $user->profile->i_layout = $user->profile->i_layout == 1 ? 0 : 1;
+    $user->profile->i_layout = $user->profile->i_layout == true ? false : true;
```
cpeel, line 350: This is a good find, but I think the error is in the UserProfile type specification for this, I think the following types are wrong:
```
 * @property bool $i_res;  # should be an int, values are in pinc/resolution.inc
 * @property bool $i_type;  # should be an int -- yes the values are just 0 and 1 but that's really a 2-value enum
 * @property bool $i_layout;  # same about a 2-value enum
 
 * @property bool $v_fntf;  # should be an int -- multi-valued enum
 * @property bool $h_fntf;  # same as ^^

```

cpeel (commented):
Thanks for the PR! These kinds of improvements are much appreciated.

tools/proofers/processtext.php
jchaffraix: Sounds good, I have applied the fix to the suggested types and reverted the int-to-bool change.

cpeel:
@srjfoo at al, I spun up a test sandbox for this branch at https://www.pgdp.org/~cpeel/c.branch/julien_add_user_in_userprefs/

cpeel (approved):

srjfoo (approved):

70ray (approved):

PR #1250, opened 2024-06-28
Alert SAs on background job failures

cpeel:
Background job failures are pretty rare, but when they happen we want people to know about them. When background jobs fail, cron sends an email with the failure code. But we should also let SAs know in case the emails go awry. This adds a notice to the alert bar _only for SAs_ if there was a job failure in the last 24 hours. The query is fully covered by an index so it is both fast and efficient.

The downside is that there is no way to dismiss the bar, we just have to wait for it to age out.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/jobs-0-alert-sa-on-job-fail/

I've inserted a demo failed job so the alert will show for the next 24 hours.

### DIFF
--- a/pinc/job_log.inc
+++ b/pinc/job_log.inc
@@ -52,3 +52,19 @@ function prune_job_log_entries(int $days_to_keep)
     );
     DPDatabase::query($sql);
 }
+
+function count_job_log_failures(int $days_ago): int
+{
+    $cutoff = new DateTimeImmutable(sprintf("%d days ago", $days_ago));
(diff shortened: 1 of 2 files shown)


chrismiceli (approved):

chrismiceli (approved):

srjfoo (approved):

windymilla (approved):
Looks good, and my two TEST users showed & didn't show the banner appropriately according to their SA-setting.

70ray (approved):

PR #1252, opened 2024-06-28
Type manage_site_access_privileges.php

jchaffraix:
The manipulated types were pretty clear
so this should be safe.

Manually tested the change by adding,
removing and loading some privileges.

_Update by cpeel_: Sandbox avail at https://www.pgdp.org/~cpeel/c.branch/julien_type_site_admin_manage_privileges/

### DIFF
--- a/tools/site_admin/manage_site_access_privileges.php
+++ b/tools/site_admin/manage_site_access_privileges.php
@@ -78,7 +78,7 @@ if ($username) {
 
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 
-function show_username_form($username)
+function show_username_form(?string $username): void
 {
     echo "<form method='GET'>\n";
     echo "<p>" . _("Username") . ": ";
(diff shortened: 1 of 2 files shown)


tools/site_admin/manage_site_access_privileges.php
```diff
@@ -45,7 +45,7 @@
     'pp_limit_value' => _("Post-processing projects limit, -1 means unlimited"),
 ];
 
-$username = array_get($_POST, 'username', array_get($_GET, 'username', null));
+$username = array_get($_POST, 'username', array_get($_GET, 'username', ''));
```
jchaffraix, line 48: I changed the fallback from `null`  (which made the type `?string`) to the empty string. This seemed slightly better as it avoids having to think about handling `null` downstream.

jchaffraix: I ended up rescinding this change as it is common for `$username` to be `null` and it is better to match the current expectations rather than break from it.

cpeel (approved):

70ray (approved):

chrismiceli (approved):

srjfoo (approved):

PR #1253, opened 2024-06-28
Type pinc/gradual.inc

jchaffraix:
I avoided a few types that I couldn't trace
back enough so would prefer to send them in
a separate PR.

Manually tested the activity hub &
my_suggestions to confirm that the new
user's banner is shown as expected.

_Updated by cpeel_: Sandbox avail at https://www.pgdp.org/~cpeel/c.branch/julien_type_gradual_php/

### DIFF
--- a/tools/proofers/my_suggestions.php
+++ b/tools/proofers/my_suggestions.php
@@ -556,7 +556,7 @@ function compare_array_by_priority($a, $b)
     return ($a["priority"] < $b["priority"]) ? -1 : 1;
 }
 
-function get_user_suggestion_criteria($username, $flush_cache = false)
+function get_user_suggestion_criteria(?string $username, $flush_cache = false)
 {
     global $testing;
 
--- a/pinc/User.inc
+++ b/pinc/User.inc
@@ -411,7 +411,7 @@ class User
     /**
      * Return the username of the currently-logged-in user
      */
-    public static function current_username()
+    public static function current_username(): ?string
     {
         global $pguser;
         return $pguser;
--- a/pinc/gradual.inc
+++ b/pinc/gradual.inc
@@ -25,7 +25,7 @@ include_once($relPath.'graph_data.inc'); // get_round_backlog_stats()
  * to pretend to have a different number of pages. In production only allow
  * users to adjust down; during testing allow adjusting higher or lower.
  */
-function get_pages_proofed_maybe_simulated($username = null)
+function get_pages_proofed_maybe_simulated(?string $username = null): int
 {
     global $testing;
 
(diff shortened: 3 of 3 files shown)


tools/proofers/my_suggestions.php
```diff
@@ -556,7 +556,7 @@ function compare_array_by_priority($a, $b)
     return ($a["priority"] < $b["priority"]) ? -1 : 1;
 }
 
-function get_user_suggestion_criteria($username, $flush_cache = false)
+function get_user_suggestion_criteria(?string $username, $flush_cache = false)
```
cpeel, line 559: `$flush_cache` could be `bool` I think if you want to make this edit too:
```diff
diff --git a/tools/proofers/my_suggestions.php b/tools/proofers/my_suggestions.php
index 99667fe36..f5156540e 100644
--- a/tools/proofers/my_suggestions.php
+++ b/tools/proofers/my_suggestions.php
@@ -11,7 +11,7 @@ include_once($relPath.'graph_data.inc'); // get_round_backlog_stats()
 require_login();

 $verbose = get_integer_param($_GET, "verbose", 0, 0, 1);
-$flush_cache = get_integer_param($_GET, "flush_cache", 0, 0, 1);
+$flush_cache = get_integer_param($_GET, "flush_cache", 0, 0, 1) == true;

 $username = User::current_username();
 if (user_is_a_sitemanager() || user_is_proj_facilitator()) {
```

And looks like `$verbose` could get the same treatment, honestly. Not in this PR, but I think we probably need to create a `get_boolean_param()` function in `misc.inc` that returns true/false instead of faking it like we are doing now.

cpeel: While ^^ is true, on reflection I wonder if this change belongs in this PR which is focused on `gradual.inc`?

cpeel: And finally, I created an issue for adding a new boolean validation func in `misc.inc`: https://github.com/DistributedProofreaders/dproofreaders/issues/1254

jchaffraix: > While ^^ is true, on reflection I wonder if this change belongs in this PR which is focused on gradual.inc?

That is a fair question. I included this as this function is the only one that `get_pages_proofed_maybe_simulated` with `$username` and typing it ensured that PHPStan would catch a type mismatch.

I looked at the callers of the functions in `pinc/gradual.inc` and this one seemed like a simple change that increased our type safety. To be fair, I wasn't consistent as I didn't update [this call site](https://github.com/DistributedProofreaders/dproofreaders/blob/504652ea0bb4f4c4ede0fb2a34f7f7ffa1b2b490/tools/proofers/my_suggestions.php#L64) that would have required a much larger change. Happy to revert this if you prefer that we avoid those small, tangentially-related additions.

cpeel: Nah, what you have in the PR is good.

cpeel (approved):

srjfoo (approved):

PR #1257, opened 2024-06-28
Adjust background job web proxy error detection

cpeel:
Last night's run of CleanUploadsTrash reported success in `job_logs` but the CLI (and cron) thought it failed. I don't know exactly why but in looking at it I've added some things to help us diagnose it in the future.

Until an HTTP stream returns some byte of data, the connection hasn't succeeded. And until the stream connects we can't get any useful metadata about it, such as if it timed out. This changes the web proxy code to flush out the HTTP headers as quickly as possible while the job runs, allowing us to better detect timeouts and handle other error conditions.

As part of this, adjust the clean up job to more gracefully handle the case where the TRASH folder doesn't exist, as the cleanup job will delete it if the folder has no contents (and Remote File Manager will create it again as-needed).

Happy path:
```
$ php run_background_job.php CleanUploadsTrash true
Background job: CleanUploadsTrash
Succeeded: true
Message: No trash directory found, nothing to do.
```

Artificially-induced timeout path:
```
$ php run_background_job.php CleanUploadsTrash true
PHP Fatal error:  Uncaught RuntimeException: CleanUploadsTrash job timed out proxied through web server in /home/cpeel/u-cpeel-dp/crontab/run_background_job.php:91
Stack trace:
#0 {main}
  thrown in /home/cpeel/u-cpeel-dp/crontab/run_background_job.php on line 91
```

I made some minor field changes in the UI to make the job selection form match the table headers: https://www.pgdp.org/~cpeel/c.branch/fine-tune-job-web-proxy/tools/site_admin/show_job_log.php

### DIFF
--- a/tools/site_admin/show_job_log.php
+++ b/tools/site_admin/show_job_log.php
@@ -5,7 +5,7 @@ include_once($relPath.'theme.inc');
 include_once($relPath.'job_log.inc');
 
 $days_ago = get_integer_param($_GET, 'days_ago', 1, 0, 365);
-$filename = array_get($_GET, 'filename', null);
+$job = array_get($_GET, 'job', null);
 $event = array_get($_GET, 'event', '');
 
 $start_timestamp = time() - (60 * 60 * 24 * $days_ago);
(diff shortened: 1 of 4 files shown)


chrismiceli (approved):

srjfoo (approved):

tools/site_admin/show_job_log.php
```diff
@@ -50,7 +50,7 @@
 echo "<th>" . _("Event") . "</th>";
 echo "<th>" . _("Comments") . "</th>";
 echo "</tr>";
-foreach (get_job_log_entries($start_timestamp, $filename, $event) as $row) {
+foreach (get_job_log_entries($start_timestamp, $job, $event) as $row) {
```
70ray, line 53: Does the definition of this function still use $filename, could that be confusing?

cpeel: It does because the field in the table itself is "filename". I don't think this will be too confusing.

70ray (approved):

PR #1261, opened 2024-07-01
Beef up development documentation

cpeel:
It's a bit of a brain dump but here's my first pass at adding more useful development docs to the repo. A reminder that the markdown files can be viewed rendered in the [update-dev-docs branch directly](https://github.com/cpeel/dproofreaders/tree/update-dev-docs?tab=readme-ov-file).

To review this, my request is that you come at the updates with "fresh" eyes and think what other questions you might have as a new person interested in developing that this doesn't answer but should.

### DIFF
--- a/SETUP/DEVELOPMENT.md
+++ b/SETUP/DEVELOPMENT.md
@@ -13,6 +13,25 @@ for information on minimum middleware versions.
 The code does not use a PHP framework or an ORM. It was developed 20 years ago
 and has grown very organically over the years. There has been effort in the
 past decade to standardize things but you'll find many different patterns remain.
+We have a defined [coding style](CODE_STYLE.md) for the codebase and a set of
+[best practices](https://www.pgdp.net/wiki/DP_Code_Best_Practices) that cover
+some of our prescribed patterns.
+
+## Contributing
+
+DP code uses git for source control and is housed at GitHub in the
+[DProofreaders/dproofreaders](https://github.com/DistributedProofreaders/dproofreaders)
+repository -- likely where you got this code to begin with.
+
+We use the "fork, branch, develop, merge" model, where developers are expected to:
+1. fork the main repository
+2. create a development branch off master
+3. develop code on the branch
+4. generate a merge/pull request when ready to merge
+
+See also [DP Code Development Using git](https://www.pgdp.net/wiki/DP_Code_Development_Using_git)
+and a similar guide specific to the [dproofreaders](https://www.pgdp.net/wiki/DP_Code_Development_Using_git:_dproofreaders)
+repo.
 
 ## Development Environment
 
@@ -39,6 +58,38 @@ nodejs dependencies (runtime and development) run the following in the repo base
 npm install
 ```
 
+## Organizing principals
+
+The code is loosely organized around the following ideas:
+* PHP pages end in `.php` extensions. As site entry points, all `.php` pages
+  need to include `pinc/base.inc` which sets up common infrastructure like
+  database connections, gettext for localization, error handlers, and more.
+* PHP code that is included in PHP pages end in `.inc` extensions.
+* Strings are localized by using gettext. These are wrapped in a `_()` function
+  that the localization tooling will extract into `.po` files that volunteer
+  translators will translate. We do not have any official change control over
+  translated strings, just update them as necessary. See the
+  [translation best practices](https://www.pgdp.net/wiki/DP_Code_Best_Practices#String_localization).
+
+### Directories
+
+* `pinc/` - Includes many of the `.inc` files -- this is short for *P*hp *INC*lude.
+  * `pinc/3rdparty/` - code that we distribute but (generally) do not modify.
+    See the [3rdparty readme](../pinc/3rdparty/README.md).
+* `scripts/` - JS files used for various features.
+* `styles/` - CSS scripts which include `.less` source files and the rendered
+  `.css` files. See [CSS / style documentation](../style/README.md).
+* `SETUP/` - Non-runtime code, such as site admin docs, development docs, tests,
+  tooling and other miscellanea. The expectations is that this directory
+  will not, and should not be, accessible via the web context on a live site.
+  * `SETUP/tests/` - Tests, mostly automated but some manual. See the
+    [tests README](tests/README.md).
+  * `SETUP/upgrade/` - Upgrade scripts. Each release gets a new subdir
+    where we collect upgrade scripts for site admins to run when they install
+    a new release. See also [UPGRADE](UPGRADE.md).
+  * `SETUP/ci/` - CI scripts that are run as part of Github Actions -- and can
+    be run manually. See `.github/workflows/ci.yml` for the GHA workflows.
+
 ## Development-related docs
 
 * [coding standards](CODE_STYLE.md)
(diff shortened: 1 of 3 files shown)


cpeel:
@jchaffraix and @mrducky4 - I would value your feedback as new developers who have stumbled around our piss-poor dev docs: is this useful and/or what it is lacking?

SETUP/DEVELOPMENT.md
```diff
@@ -0,0 +1,79 @@
...
+## Introduction
+
+The DP code is almost entirely server-side PHP with a smattering of JS and
+CSS. It uses a MySQL backend database and requires a web server -- pgdp.net
+uses Apache but Nginx should work as well. See the [installation docs](INSTALL.md)
+for information on minimum middleware versions.
+
+The code does not use a PHP framework or an ORM. It was developed 20 years ago
+and has grown very organically over the years. There has been effort in the
+past decade to standardize things but you'll find many different patterns remain.
```
jchaffraix, line 15: I wonder if it would be a good idea to link to [the code style page](https://github.com/DistributedProofreaders/dproofreaders/blob/master/SETUP/CODE_STYLE.md) or [best practice](https://www.pgdp.net/wiki/DP_Code_Best_Practices) that outlines some of those patterns.

A a side note, if there are more expected patterns, it may be worth expanding on the list in a follow-up PR (coming to mind would be getting GET/POST parameters, avoiding boolean in DB and favoring `tinyint` enums (maybe?)). As a newcomer, those are not necessarily obvious due to a mix of new and older deprecated patterns mixing.

SETUP/DEVELOPMENT.md
```diff
@@ -0,0 +1,79 @@
...
+* [automated tests](tests/README.md)
+* [CSS / style documentation](../style/README.md)
+* [API documentation](../api/README.md)
+* [coding documentation](CODE_DOCS.md)
+* [graphs design](GRAPHS.md)
+* [Unicode support](UNICODE.md)
+
+## Organizing principals
+
+The code is loosely organized around the following ideas:
```
jchaffraix, line 55: This section is very useful to outline where things live in the repository :medal_sports: 

I wonder if it would be better to structure this a bit differently though as there seems to be two concepts being introduced concurrently in this section:
* The organizing principals, e.g. when to use `.php` vs `.inc`.
* An overview of the directories in the repo, e.g. `styles` contains all CSS files, `SETUP` the setup scripts + developer doc.

The latter may be slightly more readable as a list/tree of directories, followed by what they contain.

jchaffraix (approved):
I really liked how this centralizes a lot of the information that lives in different files right now. Some comments on potential improvements but those can be handled a follow-ups.

70ray:
Is it possible to say something about the smoketests?
It seems they find errors which would be found by a compiler and linker in a system which used that workflow. Is it possible to run them without making a pull request?

SETUP/DEVELOPMENT.md
```diff
@@ -0,0 +1,79 @@
...
+uses Apache but Nginx should work as well. See the [installation docs](INSTALL.md)
+for information on minimum middleware versions.
+
+The code does not use a PHP framework or an ORM. It was developed 20 years ago
+and has grown very organically over the years. There has been effort in the
+past decade to standardize things but you'll find many different patterns remain.
+
+## Development Environment
+
+The best current development environment is using the
```
70ray, line 19: Would it be useful to say something here about developing in a sandbox on the test server? I've found this to be very satisfactory. It is so responsive it is hard to believe I'm working on a machine 8000 miles away except on very rare occasions when there are server or internet problems. It seems to be quite easy to set up compared to a VM. Is there something I'm missing?

windymilla: I agree - it was relatively easy to get set up to develop using the test server, but it did require a bit of hand-holding, for me at least. A little information here would probably reduce the need for that, and maybe a direct link to https://www.pgdp.net/wiki/DP_Code_Development_Using_git which I found useful (unless I missed that in this doc).

windymilla (approved):
Great idea to beef up the docs, both for new devs and "forgetful" ones like me :)

SETUP/DEVELOPMENT.md
cpeel: The TEST server is super useful but also a bit tricky. It's a shared dev environment and also houses some handy not-PROD-but-not-far-from-it resources. We don't give access to every potential dev that asks, they need to be a bit of a known entity and, by some definition "proved" themselves first.

In short: it's not a good option for someone who just stumbles over the repo.

SETUP/DEVELOPMENT.md
cpeel: Good suggestion. I've broken this apart, how is it now?

cpeel:
> Is it possible to say something about the smoketests?
> It seems they find errors which would be found by a compiler and linker in a system which used that workflow. Is it possible to run them without making a pull request?

I'll have to figure this out separate from this PR -- it's a good idea. It's possible to run them locally but I'll need to figure out the database side of things and write up some docs.

SETUP/DEVELOPMENT.md
jchaffraix: That looks great, thanks!

jchaffraix (approved):

SETUP/DEVELOPMENT.md
```diff
@@ -0,0 +1,102 @@
...
+
+The code uses [Composer](https://getcomposer.org/) to manage PHP dependencies.
+Some of these dependencies are runtime dependencies, others are just used for
+development (phpstan, phpunit, etc). To install PHP dependencies (runtime
+and development) run the following in the repo base:
+```bash
+composer install
+```
+
+The code also uses [npm](https://www.npmjs.com/) for some _development only_
```
chrismiceli, line 54: weirdly I don't see a download for npm on that site. I've only ever managed to install npm through the nodejs installer. We should probably call that our here as well as which version of node. Currently I think our tasks use node 20, the current LTS.

chrismiceli (approved):

SETUP/DEVELOPMENT.md
cpeel: Are you able to provide some text for here on installing nodejs/npm? On TEST we have the Ubuntu 20.04-provided nodejs and npm version which are ancient.

chrismiceli: Since node/npm is only used in dev scenarios (tests and less) I don't think the installed version matters much. For development though, there are a few options depending on your OS/distro.
1. Ubuntu - I use nodesource's repositories/instructions outlined here: https://github.com/nodesource/distributions?tab=readme-ov-file#debian-and-ubuntu-based-distributions
2. Windows - download installer from https://nodejs.org

If you need to manage multiple node/npm versions, you can use tools like nvm on linux: https://github.com/nvm-sh/nvm

Note, nodejs installed npm with it. They pair versions but aren't the same. Currently node 20 pairs with npm 10.

cpeel: Thanks Chris, that was really helpful! I've updated the docs, let me know if they make more sense.

70ray (approved):

chrismiceli (approved):

SETUP/DEVELOPMENT.md
chrismiceli: Looks good to me!

mrducky4: I used the previous version of instructions which referred to (https://nodejs.org/en/download/package-manager), which worked flawlessly for me using the VirtualBox VM.

SETUP/DEVELOPMENT.md
```diff
@@ -0,0 +1,106 @@
...
+
+DP code uses git for source control and is housed at GitHub in the
+[DProofreaders/dproofreaders](https://github.com/DistributedProofreaders/dproofreaders)
+repository -- likely where you got this code to begin with.
+
+We use the "fork, branch, develop, merge" model, where developers are expected to:
+1. fork the main repository
+2. create a development branch off master
+3. develop code on the branch
+4. generate a merge/pull request when ready to merge
```
mrducky4, line 30: Is there a set of automated regression tests a developer should run on their own before submitting a PR? Anything in the SETUP/tests directory that should be run before a PR?

PR #1264, opened 2024-07-02
Add type annotations to manage_special_days.php

jchaffraix:
The file is self-contained so it should be pretty safe.

Manually tested the change by adding a new special day, setting it to a project and made sure that it showed up.

_Update by cpeel_: Sandbox avail at https://www.pgdp.org/~cpeel/c.branch/julien_type_special_days/

### DIFF
--- a/tools/site_admin/manage_special_days.php
+++ b/tools/site_admin/manage_special_days.php
@@ -127,19 +127,20 @@ if ($action == 'show_specials') {
 class SpecialDay
 {
     public bool $new_source;
+    // Note: spec_code cannot be null in the DB but this class makes it optional.
     public ?string $spec_code;
     public string $display_name;
     public int $enable;
-    public string $comment;
+    public ?string $comment;
     public string $color;
-    public int $open_day;
-    public int $open_month;
-    public int $close_day;
-    public int $close_month;
-    public string $date_changes;
-    public string $info_url;
-    public string $image_url;
-    public string $symbol;
+    public ?int $open_day;
+    public ?int $open_month;
+    public ?int $close_day;
+    public ?int $close_month;
+    public ?string $date_changes;
+    public ?string $info_url;
+    public ?string $image_url;
+    public ?string $symbol;
 
     public function __construct(?string $spec_code = null)
     {
@@ -168,7 +169,7 @@ class SpecialDay
         }
     }
 
-    public function show_listing_row(int $count, int $current_month): int
+    public function show_listing_row(int $count, int $current_month): ?int
     {
         global $page_url;
         $sid = html_safe($this->spec_code);
--- a/tools/site_admin/manage_special_days.php
+++ b/tools/site_admin/manage_special_days.php
@@ -169,7 +169,7 @@ class SpecialDay
         }
     }
 
-    public function show_listing_row(int $count, int $current_month): ?int
+    public function show_listing_row(int $count, ?int $current_month): ?int
     {
         global $page_url;
         $sid = html_safe($this->spec_code);
@@ -373,7 +373,7 @@ class SpecialDay
         DPDatabase::query($sql);
     }
 
-    private function _show_summary_row(string $label, string $value, bool $htmlspecialchars = true): void
+    private function _show_summary_row(string $label, ?string $value, bool $htmlspecialchars = true): void
     {
         echo "  <tr>" .
             "<th class='label'>$label</th>" .
@@ -381,7 +381,7 @@ class SpecialDay
             "</tr>\n";
     }
 
-    private function _get_status_cell(int $status, string $class = '')
+    private function _get_status_cell(int $status, string $class = ''): string
     {
         switch ($status) {
             case 1:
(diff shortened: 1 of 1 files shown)


chrismiceli (approved):

tools/site_admin/manage_special_days.php
```diff
@@ -372,30 +372,15 @@ public function save_from_post()
         DPDatabase::query($sql);
     }
 
-    public function _set_field($field, $value)
```
cpeel, line 375: Probably a copy/pasta from `manage_image_sources.php`. Good catch to delete it.

cpeel:
On the test server we have some, let's just call it "interesting", data in the database where there are nulls for some fields.

```
mysql> select count(*) from special_days where image_url is null or info_url is null or date_changes is null or symbol is null;
+----------+
| count(*) |
+----------+
|       27 |
+----------+
1 row in set (0.00 sec)
```

But we don't have this problem on PROD:
```
mysql> select count(*) from special_days where image_url is null or info_url is null or date_changes is null or symbol is null;
+----------+
| count(*) |
+----------+
|        0 |
+----------+
1 row in set (0.00 sec)
```

That said, the defaults for some of these columns in the database is NULL:
```sql
CREATE TABLE `special_days` (
  `spec_code` varchar(20) COLLATE utf8mb4_general_ci NOT NULL DEFAULT '',
  `display_name` varchar(80) COLLATE utf8mb4_general_ci NOT NULL DEFAULT '',
  `enable` tinyint(1) NOT NULL DEFAULT '1',
  `comment` varchar(255) COLLATE utf8mb4_general_ci DEFAULT NULL,
  `color` varchar(8) COLLATE utf8mb4_general_ci NOT NULL DEFAULT '',
  `open_day` tinyint DEFAULT NULL,
  `open_month` tinyint DEFAULT NULL,
  `close_day` tinyint DEFAULT NULL,
  `close_month` tinyint DEFAULT NULL,
  `date_changes` varchar(100) COLLATE utf8mb4_general_ci DEFAULT NULL,
  `info_url` varchar(255) COLLATE utf8mb4_general_ci DEFAULT NULL,
  `image_url` varchar(255) COLLATE utf8mb4_general_ci DEFAULT NULL,
  `symbol` varchar(2) COLLATE utf8mb4_general_ci DEFAULT '',
  UNIQUE KEY `spec_code` (`spec_code`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci COMMENT='definitions of SPECIAL days'
```

So we probably want to update the type definitions in the class to handle null from columns whose default is NULL so we don't break anything accidentally.

-or-

We could add an upgrade script that updates all null values to empty string and changes the schema.

The former would be my recommendation for now.

jchaffraix:
> So we probably want to update the type definitions in the class to handle null from columns whose default is NULL so we don't break anything accidentally.
> 
> -or-
> 
> We could add an upgrade script that updates all null values to empty string and changes the schema.
> 
> The former would be my recommendation for now.

Good catch. I have updated the type definitions to match the schema when null was allowed. 

Calling out that `symbol` was made nullable even if its default is '' (based on your SQL query it seemed like the test server has some data with NULL in it).

70ray (approved):

tools/site_admin/manage_special_days.php
```diff
@@ -417,7 +403,7 @@ public function _get_status_cell($status, $class = '')
 
 // ----------------------------------------------------------------------------
 
-function make_link($url, $label)
+function make_link(string $url, string $label): string
```
cpeel, line 406: Because `info_url` can be null, both of these need to accept null too (see line 224).

jchaffraix: Indeed, thanks. Fixed this in the next commit + a few extra potential issues caught by PHPStan level 9.

I am bit concerned about this not being caught by PHPStan in the first place so will look into why tomorrow.

tools/site_admin/manage_special_days.php
```diff
@@ -127,21 +127,22 @@
 class SpecialDay
 {
     public bool $new_source;
+    // Note: spec_code cannot be null in the DB but this class makes it optional.
```
srjfoo, line 130: By itself, I found this comment to be confusing. Since this isn't an autoincrement field, but requires `$spec_code` to be created by the person entering the new special day, the code has to make sure that the value is not null. Perhaps add something along the lines of
```php
// to ensure that the value in the form for the field is not null
```

jchaffraix: Sounds good. I have expanded the comment, which I think captured your suggestion. Let me know if that was what you had in mind:

```
// spec_code cannot be null in the DB. However to reuse the HTML form generation
// logic the first time the page is loaded, we do allow a null value in this class.
// Thus when reading the value from the form to store in the DB, we must ensure
// that it is not null.
```

cpeel:
Test sandbox updated.

cpeel (approved):

srjfoo (approved):

PR #1265, opened 2024-07-03
Convert del_teams.php to BackgroundJob and update README

cpeel:
One last PR to get all of our background jobs moved over to the new infra.

This one converts `del_teams.php` which works, but we don't actually run it or tell people to run it. Because we don't have foreign keys it's probably not a good idea to delete empty teams because in the past they might have completed pages and so there are records of them in `past_tallies` that would no longer have a team record associated with them. We could opt to remove the script instead of keeping it.

This also updates the crontab README with info on the new infra.


chrismiceli (approved):

70ray (approved):

srjfoo:
> Because we don't have foreign keys it's probably not a good idea to delete empty teams because in the past they might have completed pages and so there are records of them in `past_tallies` that would no longer have a team record associated with them. We could opt to remove the script instead of keeping it.

I went ahead and approved, but honestly, if it's not being run, and shouldn't be run, it's probably safer to delete it. I feel certain that teams have been deleted in the past, without due consideration of the issues you brought up.

cpeel:
> I went ahead and approved, but honestly, if it's not being run, and shouldn't be run, it's probably safer to delete it. 

That makes sense to me too -- we can always pull it back out of git history of needed. I force-pushed an update to do just that.

chrismiceli (approved):

PR #1267, opened 2024-07-05
Switch to PHPStan to level 5

jchaffraix:
This level gives us type coverage on arguments.

To ease the transition, the baseline that
corresponds to the current errors was added
to `phpstan.neon`. Due to our CI using PHP8.3
and prod/developers using PHP7.4, I have added
the baseline for both versions.

Some common, low-value exceptions were also
added to prevent having a large list of
individual errors.


cpeel (approved):
This is excellent, thank you for doing the legwork.

PR #1268, opened 2024-07-05
Give PMs info on languages without dictionaries

cpeel:
We give PMs a stupidly long list of languages when creating projects, but only a few of them have dictionaries installed for WordCheck. Give PMs information about which languages have dictionaries installed when creating and editing projects.

This also removes the duplicate entries in the `get_iso_language_list()` return and for the places where order matters (filters and project search form) we show the list of languages with dictionaries first then everything else. This means that English is no longer at the very top, but it's only a few down so that seems reasonable.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/warn-pms-about-wordcheck-dicts/

### DIFF
--- a/faq/wordcheck_data.php
+++ b/faq/wordcheck_data.php
@@ -1,0 +1,65 @@
+<?php
+$relPath = '../pinc/';
+include_once($relPath.'base.inc');
+include_once($relPath.'wordcheck_engine.inc');
+include_once($relPath.'theme.inc');
+
+require_login();
+
+$title = _("WordCheck Site Data");
+output_header($title, NO_STATSBAR);
+
+echo "<h1>$title</h1>";
+
+echo "<h2 id='site_word_lists'>" . _("Site Word Lists") . "</h2>";
+echo "<p>" . _("Site-level words are stored in language-specific files.") . "</p>";
+echo "<p>" . _("Site-level Good and Bad word lists are used when calculating Flagged words in a body of text. Here is the current set of such lists.") . "</p>";
+
+createWordListTable(get_site_good_bad_word_lists());
+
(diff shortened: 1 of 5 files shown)


srjfoo (approved):

cpeel:
Added a second commit that moves (well, copies) the `noncvs/wordcheck_data.php` script into the code and includes a link to it from the project creation page. After this gets rolled out to PROD we can update the wiki to point to the new page and remove it from `noncvs`.

srjfoo (approved):

70ray (commented):
It seems like Spanish is getting lost somehow.

windymilla:
> It seems like Spanish is getting lost somehow. It doesn't appear in the wordcheck list or in the list for selecting project languages.

I think Spanish (when it means the official language of Spain) is called "Castilian" in the dictionaries to distinguish it from other variants of Spanish, like Catalan

chrismiceli (approved):

windymilla (approved):

cpeel:
Thanks for the "Spanish" catch. Both are in the lang list but there's probably something happening when its doing the "is there a dict" mapping. I'll figure it out.
```
$ grep "spa" iso_lang_list.inc
        ["lang_name" => "Castilian", "lang_code" => "spa"],
        ["lang_name" => "Spanish", "lang_code" => "spa"],
```

cpeel:
I've added a commit that specifies a "preferred" language for the duplicate ISO-639-2 entries instead of just picking the first one in the list. I went through the languages used in projects at pgdp.net and picked the variant that has been used and/or used the most often. All languages will continue to be shown in the drop-downs and to users.

srjfoo (approved):

70ray (approved):

Issue #1224, opened 2024-06-03
Delete XML project feeds


cpeel:
The project XML feeds are no longer valuable and haven't been for years. An analysis of the feed requests from pgdp.net show that the RSS feeds are used frequently but only bots and crawlers are accessing the XML feeds.

We should:
* remove the XML-based code from [backend.php](https://github.com/DistributedProofreaders/dproofreaders/blob/master/feeds/backend.php) and make it only return RSS regardless of what `type` is passed in (if any).
* update appropriate pages that link to the XML-based project feeds.
* delete the [rss](https://github.com/DistributedProofreaders/dproofreaders/blob/master/graphics/rss.gif) image and use the [rss-square](https://fontawesome.com/v5/search?q=rss-square&o=r&m=free) FontAwesome icon.
* ideally do some cleanup in `backend.php` along the way.

We need to confirm that we do not break existing RSS links for the feeds.

The team and member XML feeds have been accessed non-zero times in the past month (7 and 38 respectively) and while my inclination is to delete them we should do that in a separate PR if we decide to do so.

mrducky4:
I'm interested in working on this issue.

cpeel:
> I'm interested in working on this issue.

@mrducky4 - sounds great!

One thing I thought of is that caching the feed in a file and serving it up from there is probably no longer necessary for our current database and production server. Moreover we now have a memcached instance for hot data that we could use via the [memoize()](https://github.com/DistributedProofreaders/dproofreaders/blob/master/pinc/misc.inc#L1447) function if we wanted to cache it.

cpeel:
Fixed by #1269 by @mrducky4  -- thanks Mark!

PR #1269, opened 2024-07-06
Remove xml feeds

mrducky4:
Address issue #1224
Remove the old XML feeds, keeping the RSS feeds.
Remove the feed type parameter from backend.php.
Replace the old feed icon with the fontawesome RSS icon.
Add backend.php to pageload_smoketest.py.

_Update from cpeel_: Sandbox avail at https://www.pgdp.org/~cpeel/c.branch/remove_xml_feeds


chrismiceli (approved):

feeds/backend.php
```diff
@@ -4,22 +4,43 @@
...
+ * @param string $content
+ *   The category for the feed: posted, postprocessing, proofing, or smoothreading
+ *
+ * @param string $site_name
+ *   The name of the site serving the feed
+ *
+ * @param string $code_url
+ *   The base URL for the site
+ *
+ * @param string $charset
```
cpeel, line 31: `$charset` is always going to be `UTF-8` since the conversion to Unicode in 2020. I have a branch ready to remove this global from the codebase but was waiting on this PR to land first so I didn't generate conflicts. Lets drop the variable here and just put `UTF-8` in the XML header as a literal.

feeds/backend.php
```diff
@@ -4,22 +4,43 @@
...
+ *   The character set of the feed, such as "UTF-8"
+ *
+ * @param string $site_manager_email_addr
+ *   Contact info for the feed <webMaster> element
+ *
+ * @return string
+ *   The XML document for the RSS feed
+ */
+function generate_rss_feed($content, $site_name, $code_url, $charset, $site_manager_email_addr)
```
cpeel, line 40: Please add type definitions here:
```suggestion
function generate_rss_feed(string $content, string $site_name, string $code_url, string $site_manager_email_addr): string
```

pinc/showavailablebooks.inc
```diff
@@ -118,16 +118,15 @@ function show_projects_for_smooth_reading()
     echo "\n<h2>$title</h2>";
 
     $content = 'smoothreading';
-    $xml_rss_images = "
+    $rss_image = "
         <a href='$code_url/feeds/backend.php?content=$content'>
-        <img src='$code_url/graphics/xml.gif' width='36' height='14' alt='[XML]'></a>
-        <a href='$code_url/feeds/backend.php?content=$content&amp;type=rss'>
-        <img src='$code_url/graphics/rss.gif' width='36' height='14' alt='[RSS]'></a>
```
cpeel, line 125: We can delete `graphics/rss.gif` in this PR too. Need to leave the XML one around for a bit though since the Teams and Members code uses it.

cpeel (commented):
This is great and I had just a few suggestions -- thank you for working on it!

feeds/backend.php
mrducky4: Done. I changed it in the HTTP response header too.

pinc/showavailablebooks.inc
mrducky4: Done.

feeds/backend.php
mrducky4: Done. Is this a new practice? I didn't see it in [DP Code Best Practices](https://www.pgdp.net/wiki/DP_Code_Best_Practices).

styles/themes/charcoal.css
```diff
@@ -1790,6 +1790,9 @@ hr.divider {
   text-align: center;
   text-shadow: 3px 2px 5px black;
 }
+.rss-icon {
+  color: #f26522;
+}
```
srjfoo, line 1795: After looking at this actual size on Charcoal, I like it less than I initially thought I would (it looks fine on the white-based themes) -- I've done a bit of research, and it looks like there is no one true color for the RSS icon as far as the internet is concerned, so, if I may suggest an alternative for Charcoal ... I tried #ff9a00 (another, somewhat lighter version), and then tried mixing the two. #fd9205 looks good to me, and is slightly darker but not as vibrant against the very dark grey background as the existing color.

IIRC, it means adding an override to `charcoal.less`. I don't feel like it has to happen in this PR, but wanted to mention it here first.

pinc/showstartexts.inc
```diff
@@ -38,9 +38,8 @@ function showstartexts($etext_limit, $type)
     // Display feed buttons
     echo "
         <a href='$code_url/feeds/backend.php?content=$content'>
-        <img src='$code_url/graphics/xml.gif' width='36' height='14' alt='[XML]'></a>
-        <a href='$code_url/feeds/backend.php?content=$content&amp;type=rss'>
-        <img src='$code_url/graphics/rss.gif' width='36' height='14' alt='[RSS]'></a>
+        <i class='fas fa-rss-square rss-icon'></i>
```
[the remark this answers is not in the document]
mrducky4: Good question. I followed [fontawesome documentation](https://fontawesome.com/v5/icons/rss-square?f=classic&s=solid), without digging into the underlying mechanism. I also saw other examples in the code that already use `<i>` for fontawesome. Just now I also found documentation that says you can use `<span>` instead of `<i>`, and maybe other elements too, but fontawesome recommends using one element type consistently.

feeds/backend.php
```diff
@@ -4,22 +4,40 @@
...
-    $xmlfile = "$xmlfeeds_dir/{$content}_rss.$intlang.xml";
-} else {
-    $xmlfile = "$xmlfeeds_dir/{$content}.$intlang.xml";
-}
+// Let the browser cache it for $cache_duration seconds
+$cache_duration = 30 * 60;
+$now = time();
+header("Content-Type: text/xml; charset=UTF-8");
+header("Expires: " . gmdate("D, d M Y H:i:s", $now + $cache_duration) . " GMT");
+header("Last-Modified: " . gmdate("D, d M Y H:i:s", $now) . " GMT");
```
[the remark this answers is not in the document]
mrducky4: I checked on the specification for the Expires header and found GMT is required. [RFC9111](https://www.rfc-editor.org/rfc/rfc9111#field.expires) refers to [RFC9110](https://www.rfc-editor.org/rfc/rfc9110#section-5.6.7), which says it must be GMT. I assume the same is true for Last-Modified, but I didn't check that one.

styles/themes/charcoal.css
mrducky4: I tried it with #fd9205, I have no preference of one color over the other. It is an easy change, so I'm happy to add this color for the charcoal theme in this PR if that's what the team prefers.

pinc/showstartexts.inc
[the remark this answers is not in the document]
srjfoo: 😁 -- shows you how alert I am, or I'd have seen it in earlier fa-related code. If I'd noticed then, I'd probably have suggested `<span>` just out of persnicketyness (since it doesn't imply a particular treatment of the enclosed character), but I agree. Whatever we use, we should be consistent.

styles/themes/charcoal.css
srjfoo: Screenshots for everyone to look at:

Current sandbox (#f26522), charcoal: 
image

Charcoal with #fd9205: 
image

default: 
image

default with #fd9205: 
image

(Wow ... I had no idea that the background for charcoal was so close to the background (in dark mode) for github 😁 )

feeds/backend.php
cpeel: It is a new practice. I'll get the best practices code updated - thanks for calling it out.

pinc/showstartexts.inc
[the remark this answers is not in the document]
cpeel: Yes, let's stick with `<i>` as that's what we use most other places.

styles/themes/charcoal.css
cpeel: I think the current color on white looks good and defer to @srjfoo on what color she thinks is best for charcoal. If we can get it into this PR that would be appreciated.

feeds/backend.php
cpeel: Added this (brief) section on typing here: https://www.pgdp.net/wiki/DP_Code_Best_Practices#Typing

styles/themes/charcoal.css
mrducky4: I updated the PR to use #fd9205 for charcoal.

cpeel (approved):

srjfoo (approved):

PR #1270, opened 2024-07-07
Remove $charset global

cpeel:
The site has been UTF-8-only since 2020, this just removes the `$charset` global variable. Should be zero functional change.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/prune-charset/

### DIFF
--- a/pinc/wordcheck_engine.inc
+++ b/pinc/wordcheck_engine.inc
@@ -312,7 +312,7 @@ class BadWordAccumulator
 function get_bad_words_via_external_checker($input_words_w_freq, $languages)
 {
     global $aspell_temp_dir;
-    global $aspell_executable, $aspell_prefix, $charset;
+    global $aspell_executable, $aspell_prefix;
 
     $messages = [];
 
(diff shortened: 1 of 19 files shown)


tools/authors/bioxml.php
```diff
@@ -33,8 +33,8 @@
     $wrap_in_big_tag = true;
 }
 
-header("Content-Type: text/xml; charset=$charset");
-echo "<?xml version=\"1.0\" encoding=\"$charset\" ?>\n";
+header("Content-Type: text/xml; charset=UTF-8");
+echo "<?xml version=\"1.0\" encoding=\"UTF-9\" ?>\n";
```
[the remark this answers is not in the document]
cpeel: Yes it sound, good catch!

chrismiceli (approved):

70ray (approved):

srjfoo (approved):

pinc/wordcheck_engine.inc
```diff
@@ -343,7 +343,7 @@ function get_bad_words_via_external_checker($input_words_w_freq, $languages)
                     "list",
                     "--prefix=" . escapeshellarg($aspell_prefix),
                     "-d " . escapeshellarg($dict_file),
-                    "--encoding " . escapeshellarg($charset),
+                    "--encoding=UTF-8",
```
mrducky4, line 346: `"--encoding=UTF-8"` looks correct. Was the old version with `"--encoding UTF-8"` wrong, but still working?

cpeel: Both work, aspell accepts args with or without `=`s: `--param value` or `--param=value`

PR #1271, opened 2024-07-07
Fix string|float-to-int convertion errors

jchaffraix:
There were 3 major errors:
- int fields from the DB that are returned as string by mysqli.
- date('s|d|y|U') that returns an int as a string.
- PHP float truncating functions that returns a float.

_Update by cpeel_: Sandbox avail at https://www.pgdp.org/~cpeel/c.branch/julien_fix_int_to_str_conv/

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -24,10 +24,8 @@ parameters:
           path: pinc/3rdparty/mediawiki/DairikiDiff.php
         - message: '#Comparison operation .* results in an error#'
           path: pinc/3rdparty/mediawiki/DiffEngine.php
-        - message: '#^Parameter \#2 \$num of function array_fill expects int, float given\.$#'
+        - message: '#^Parameter \#2 \$.* of function array_fill expects int, float given\.$#'
           path: pinc/3rdparty/mediawiki/DiffEngine.php
-          # This only applies to PHP7, remove this rule when we use PHP8.
-          reportUnmatched: false
         - message: '#Instantiated class DiffFormatter is abstract#'
           path: pinc/DifferenceEngineWrapper.inc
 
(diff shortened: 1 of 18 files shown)


phpstan.neon
```diff
@@ -24,7 +24,7 @@ parameters:
           path: pinc/3rdparty/mediawiki/DairikiDiff.php
         - message: '#Comparison operation .* results in an error#'
           path: pinc/3rdparty/mediawiki/DiffEngine.php
-        - message: '#^Parameter \#2 \$num of function array_fill expects int, float given\.$#'
+        - message: '#^Parameter \#2 \$.* of function array_fill expects int, float given\.$#'
```
jchaffraix, line 27: That's a small mistake I made in the previous PR that makes PHPStan fails: PHP 8 does actually see the error but with a different parameter name.

pinc/Pool.inc
```diff
@@ -35,17 +35,18 @@ class Pool extends Stage
...
-        $access_minima,
-        $after_satisfying_minima,
-        $evaluation_criteria,
-        $access_change_callback,
-        $description,
-        $document,
-        $foo_Header,
-        $foo_field_name,
-        $blather
+        string $id,
```
jchaffraix, line 38: I added this type to prove that that the `$id` parameter is a string and thus we need to manually ignore some issue in PHPStan on `pinc/stages.inc`.

The other types in the constructors were added bonuses.

All the types are copied from the class member's.

cpeel (approved):

pinc/stages.inc
```diff
@@ -307,7 +307,7 @@ new Stage(
 // -----------------------------------------------------------------------------
 
 new Pool(
-    'PPV',
+    'PPV', // @phpstan-ignore-line PHPStan 1.10 reports a string-to-int conv here, seems like a bug
```
chrismiceli, line 310: is it worth setting up a minimal example and creating an issue against phpstan?

chrismiceli (approved):

srjfoo (approved):

pinc/stages.inc
jchaffraix: Yes, I just did! Also the reproduction showed that the issue is that we reuse the name `Pool` that exists in the PHP standard library and confuses PHPStan:

https://github.com/phpstan/phpstan/issues/11303

chrismiceli: That's awesome, thanks.

Issue #1232, opened 2024-06-10
Implement standard logging interface


cpeel:
PHP has no standard logging interface (unlike Python's robust [logging module](https://docs.python.org/3/library/logging.html)). All it has is [error_log()](https://www.php.net/manual/en/function.error-log) which dumps a string out to the `php_errors` file. We use this feature in several places in the code to log things:
```
$ grep -R error_log *
accounts/activate.php:    error_log("activate.php - Error activating $reg_token: $create_user_status");
api/index.php:        error_log("api/index.php - Error setting $key:count=$count in memcache");
pinc/base.inc:        error_log("base.inc - $error_message");
pinc/DPDatabase.inc:        error_log("DPDatabase.inc - log_error from $caller: $error");
pinc/send_mail.inc:        error_log("send_email.inc - Message could not be sent. PHPMailer Error: {$mail->ErrorInfo}");
pinc/upload_file.inc:        error_log("upload_file.inc - zip error: " . $exception->getMessage());
pinc/upload_file.inc:        error_log($reporting_string);
tasks.php:            error_log("tasks.php - task $tid is linked to non-existent topic $topic_id");
tools/project_manager/add_files.php:                error_log("add_files.php - error running \"$cmd\"; exit code: $exit_status");
tools/upload_resumable_file.php:        error_log("upload_resumable_file.php - $error");
```

We also have various forms of adhoc logging like DPDatabase->log_trace().

What we need is a logging interface that we can use instead of `php_errors` for things that aren't actually errors (although they might be) that we want to keep track of. Like the Python logging library it needs to:
* honor logging levels (error, warning, info, debug)
* configure logging levels for individual classes or modules
* be simple for consumers to use (eg: `DPLogger::info("log this message")`)

There may be existing logging packages out there although I didn't find a lightweight one on packagist.

cpeel:
Correction, I think https://github.com/Seldaek/monolog is a good option here. It has a lot of optional requires but the default set is very small.

PR #1273, opened 2024-07-11
Install and use Symfony Process for system calls

cpeel:
Symfony's Process class handles shell escaping and other niceties like stdin/stdout redirection. This instance fixes the issue where we can't tell if the aspell process completed successfully and didn't find anything misspelled or if it errored out.

This was identified in the newly spun up `clone.pgdp.net` instance because `shell_exec()` returned NULL and we passed that into `explode()` and that's now deprecated in PHP 8.1. Moving to `Process()` is an overall Good Thing and we should do it across the codebase. I'll create an issue.

I've done ImportPGCatalog as an example which doesn't use stdin.

Sandbox available at https://www.pgdp.org/~cpeel/c.branch/use-symfony-process/

### DIFF
--- a/crontab/ImportPGCatalog.inc
+++ b/crontab/ImportPGCatalog.inc
@@ -14,6 +14,7 @@ use Symfony\Component\Process\Process;
 class ImportPGCatalog extends BackgroundJob
 {
     private string $local_catalog_dir;
+    private int $time_limit_seconds = 300;
 
     private array $display_mapping = [
         'application/epub+zip' => 'EPUB',
@@ -56,7 +57,7 @@ class ImportPGCatalog extends BackgroundJob
 
     public function work()
     {
-        set_time_limit(300);
+        set_time_limit($this->time_limit_seconds);
 
         // determine temporary directory for download and extraction
         $this->local_catalog_dir = sys_get_temp_dir() . "/pg-catalog-files";
@@ -101,7 +102,7 @@ class ImportPGCatalog extends BackgroundJob
             $this->local_catalog_dir,
             "--overwrite",
         ]);
-        $process->setTimeout(600);
+        $process->setTimeout($this->time_limit_seconds);
         $process->run();
 
         // Each file in the tar archive describes one ebook in the PG collection,
--- a/crontab/ImportPGCatalog.inc
+++ b/crontab/ImportPGCatalog.inc
@@ -1,6 +1,8 @@
 <?php
 include_once($relPath.'pg.inc');
 
+use Symfony\Component\Process\Process;
+
 // Timing requirements: This script can run anytime, but probably daily.
 
 /*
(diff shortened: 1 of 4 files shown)


crontab/ImportPGCatalog.inc
```diff
@@ -87,16 +89,21 @@ class ImportPGCatalog extends BackgroundJob
...
+            "--file",
+            $local_compressed_file,
+            "--strip-components",
+            3,
+            "--directory",
+            $this->local_catalog_dir,
             "--overwrite",
         ]);
-        system($cmd, $ret);
+        $process->setTimeout(600);
```
chrismiceli, line 104: May make sense to store in some variable and name it so it is clear why and how long it is, like:
```
const MAXIMUM_EXTRACT_TIMEOUT_IN_MS = 600
```
any reason we decided on 600? Was that the default for system before perhaps?

chrismiceli (approved):

srjfoo (approved):

pinc/wordcheck_engine.inc
```diff
@@ -311,21 +312,11 @@ class BadWordAccumulator
...
 
-    $messages = [];
-
-    $tmp_file_path = tempnam($aspell_temp_dir, "pagetext-");
-
     $tmp_file_text = join("\n", array_keys($input_words_w_freq));
-    $wasWritten = file_put_contents($tmp_file_path, $tmp_file_text);
-    unset($tmp_file_text);
-    if ($wasWritten === false) {
-        $messages[] = sprintf(_("Error: unable to write text to temp file: %s"), $tmp_file_path);
```
jchaffraix, line 325: Do we need to remove the orphaned translations for this string? Or is that going to be handled when we next update the translation files?

jchaffraix (approved):

pinc/wordcheck_engine.inc
srjfoo: If the string isn't in the code, the translation doesn't get displayed, because it doesn't map to anything that gets displayed.

Even once the POT is regenerated, the orphaned translation remains in the POT until the old strings are purged, but, again, there's nothing they map to, unless one of them gets re-added to the code.

Thanks for the reminder -- it's past time for me to regenerate the POT files, etc.

70ray (approved):

crontab/ImportPGCatalog.inc
```diff
@@ -105,7 +112,7 @@ class ImportPGCatalog extends BackgroundJob
         //     cache/epub/NNN
         // part, and --directory=$local_catalog_dir so that we end up with
         //     $local_catalog_dir/pgNNN.rdf
-        if ($ret) {
+        if (!$process->isSuccessful()) {
             throw new RuntimeException("Unable to extract files from $local_compressed_file to {$this->local_catalog_dir}");
```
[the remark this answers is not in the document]
cpeel: It would but we don't have a good place to do that right now beyond `php_errors`. After we get https://github.com/DistributedProofreaders/dproofreaders/issues/1232 in place we can start adding logging throughout the codebase in ways that make sense.

crontab/ImportPGCatalog.inc
cpeel: We already had a 300s timeout for the job and I centralized the value so its used in both places via a private class value, good suggestion.

srjfoo (approved):

chrismiceli (approved):

PR #1277, opened 2024-07-13
Wrangle special_days management page

cpeel:
Several fields are actually required and we should make them so on the input form. Also update the database schema to remove nulls and adjust the class to match -- it doesn't make any sense why the DB schema was like it was.

Sandbox at https://www.pgdp.org/~cpeel/c.branch/wrangle-special-days/

### DIFF
--- a/tools/site_admin/manage_special_days.php
+++ b/tools/site_admin/manage_special_days.php
@@ -134,16 +134,16 @@ class SpecialDay
     public ?string $spec_code;
     public string $display_name;
     public int $enable;
-    public ?string $comment;
+    public string $comment;
     public string $color;
-    public ?int $open_day;
-    public ?int $open_month;
-    public ?int $close_day;
-    public ?int $close_month;
-    public ?string $date_changes;
-    public ?string $info_url;
-    public ?string $image_url;
-    public ?string $symbol;
+    public int $open_day;
+    public int $open_month;
+    public int $close_day;
+    public int $close_month;
+    public string $date_changes;
+    public string $info_url;
+    public string $image_url;
+    public string $symbol;
 
     public function __construct(?string $spec_code = null)
     {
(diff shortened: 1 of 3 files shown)


chrismiceli (approved):

tools/site_admin/manage_special_days.php
```diff
@@ -257,14 +257,14 @@ public function show_edit_form(): void
...
+            $this->_show_edit_row('spec_code', _('Special Day ID'), 'text', 20, null, null, true);
         } else {
             echo "<input type='hidden' name='editing' value='true'>" .
                 "<input type='hidden' name='spec_code' value='" . attr_safe($this->spec_code) ."'>
                 <table class='edit_special_day'>";
             $this->_show_summary_row(_('Special Day ID'), $this->spec_code);
         }
-        $this->_show_edit_row('display_name', _('Display Name'), 'text', 80);
+        $this->_show_edit_row('display_name', _('Display Name'), 'text', 80, null, null, true);
         $this->_show_edit_row('symbol', _('Symbol'), 'text', 4);
```
srjfoo, line 268: Not entirely sure where to put this comment, but, in testing, I got a fatal error when the symbol I chose was processed, I'm assuming because the emoji I was trying to use was too complex.

jchaffraix (approved):

tools/site_admin/manage_special_days.php
cpeel: The db current column size is 2 which limits the unicode characters a single symbol can use. I've extended it to 4 on TEST, can you try your cat emoji again on TEST?

srjfoo: Confirmed that the problem was that the emoj I used was "too long". 

Currently, the field is `varchar(2)`. If we push it to `varchar(3)`, that potentially widens the column slightly if someone decides to just use text characters up to a max of 3. There are a lot of emoji that are composed of 4 or more characters, though.

In further experimenting, I discovered that the 4-character emoji weren't even displayable in the symbol field in the form. However, will they be if we change to varchar(3)? This is mostly of concern because I was allowed to actually enter a character that was one character longer than the field width.

I found [this](https://stackoverflow.com/questions/77225991/whats-the-maximum-number-of-code-points-for-a-utf-8-encoded-displayed-character) that talks about the possible lengths of available emoji. 😁

srjfoo: > The db current column size is 2 which limits the unicode characters a single symbol can use. I've extended it to 4 on TEST, can you try your cat emoji again on TEST?

Yes, with a column size of 4, I can add the emoji that caused this problem in the first place.

cpeel: I increased it to `varchar(4)` as this matches the length of the form and just going to punt the "how many characters can we cram into a single symbol" problem because frankly we haven't run into it yet in the real world.

srjfoo: Works for me.

srjfoo (approved):

PR #1278, opened 2024-07-14
Revert "Allow image to scroll beyond edges (#1010)"

70ray:
This reverts commit c09a0e4fc8ae5cf9461bdb10edfa5265a5b31597. until we can deal with iPad problems which outweigh advantages

### DIFF
--- a/scripts/control_bar.js
+++ b/scripts/control_bar.js
@@ -173,12 +173,12 @@ function makeControlDiv(container, content, controls) {
         menu.hide();
     });
 
-    let onChangeCompassPoint = new Set();
+    let onChange = new Set();
     function newPoint(newP) {
         compassPoint = newP;
         saveLocation();
         setCompassPoint();
-        onChangeCompassPoint.forEach(function (onChangeCallback) {
+        onChange.forEach(function (onChangeCallback) {
             onChangeCallback();
         });
         menu.hide();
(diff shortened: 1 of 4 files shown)


cpeel:
Thanks for getting this ready. I want to hold off merging it in until we find that it's a real problem for a few users. I know it's not great for iPads and I'd like to get it fixed, but we've also never said that we support proofreading on tablets so I don't feel this is currently urgent.

srjfoo:
I've been pondering this for quite a while now. We haven't seen much in the way of complaints in the forums, but that may be because those who don't like it have decided that there's no percentage any longer in complaining.

I would hate to disable it altogether, because I think it can be useful, and I think that at some point, if we sync image with text, we'll need it. It just occurred to me to wonder if it would be reasonable to add a checkbox to the image controls so that overscroll/panning (I'm not sure what the correct term is) can be turned off for those who hate it.

The two issues I'm aware of:

* The issue on the iPad has to do with what happens to the proofing image if the user has the PI open then tabs away temporarily to deal with something else, then comes back to the PI. In the horizontal format, the image has scooted all the way as far to the right as it can go. In both Standard and Enhanced. I have confirmed the behavior on my iPad. I was able to get it to happen in the vertical orientation, as well, but only if the image is fit to height. Perhaps, in this case, the problem is that the position of the image is not being remembered?
* The other issue is that some people just overcompensate with the scroll wheel when moving the image up and down. And they hate it because they then have to scroll back in the other direction.

cpeel:
> In the horizontal format, the image has scooted all the way as far to the right as it can go. In both Standard and Enhanced. I have confirmed the behavior on my iPad. I was able to get it to happen in the vertical orientation, as well, but only if the image is fit to height. Perhaps, in this case, the problem is that the position of the image is not being remembered?

While it's probably a pain, I'd rather see if we can figure out why this is happening on the iPad and fix it directly. Barring that, maybe disable it just for iPad (if that's somehow possible). Given the lack of complaints about this I don't think we should roll it back.

And unsurprisingly I'm against making this an optional feature.

cpeel:
At this point we're not going to revert this change. The issue is still a problem but apparently not a big / widespread one or we'd have people screaming about it.

I'd be interested in seeing if we can find a better solution for those users, but I don't see it as urgent.

PR #1279, opened 2024-07-14
Remove unneeded calls to DPDatabase::escape()

jchaffraix:
Those are covered by a general exception for int-to-string conversions in `phpstan.neon` so
were found by temporarily disabling it. That's also why there is no progression from PHPStan's perspective.

### DIFF
--- a/tools/authors/manage.php
+++ b/tools/authors/manage.php
@@ -132,7 +132,7 @@ if ($_POST) {
     $count = count($enable_author_values);
     for ($i = 0; $i < $count; $i++) {
         $sql = sprintf(
-            "UPDATE authors SET enabled = '%d' WHERE author_id = %d",
+            "UPDATE authors SET enabled = %d WHERE author_id = %d",
             $enable_author_values[$i],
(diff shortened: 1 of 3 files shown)


chrismiceli (approved):

srjfoo (approved):

tools/authors/manage.php
```diff
@@ -132,8 +132,8 @@
     $count = count($enable_author_values);
     for ($i = 0; $i < $count; $i++) {
         $sql = sprintf(
-            "UPDATE authors SET enabled = '%s' WHERE author_id = %d",
-            DPDatabase::escape($enable_author_values[$i]),
+            "UPDATE authors SET enabled = '%d' WHERE author_id = %d",
```
cpeel, line 135: We shouldn't surround numbers with quotes, even though MySQL will accept it.
```suggestion
            "UPDATE authors SET enabled = %d WHERE author_id = %d",
```

cpeel (commented):
Just one small change request.

cpeel (approved):

PR #1281, opened 2024-07-15
Add cronjobs and additional pages to smoketests

cpeel:
Add cronjobs and some more page test variations to the smoketests.

### DIFF
--- a/SETUP/tests/smoketests/pageload_smoketest.py
+++ b/SETUP/tests/smoketests/pageload_smoketest.py
@@ -1,13 +1,14 @@
 #!/usr/bin/env python3
 
 import argparse
+import os
 import re
 import signal
 import sys
 import time
 
 from http.cookiejar import CookieJar, DefaultCookiePolicy
-from subprocess import Popen, DEVNULL
+from subprocess import Popen, DEVNULL, run, STDOUT, PIPE
 from urllib.error import URLError
 from urllib.parse import urlencode, urlparse
 from urllib.request import build_opener, HTTPCookieProcessor, HTTPErrorProcessor, Request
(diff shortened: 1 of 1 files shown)


SETUP/tests/smoketests/pageload_smoketest.py
```diff
@@ -667,6 +704,26 @@ def main() -> int:
...
+                "php",
+                os.path.join(config['code_dir'], "crontab", "run_background_job.php"),
+                test,
+                "true"
+            ],
+            stdout=PIPE,
+            stderr=STDOUT,
+            encoding='utf-8'
+        )
+        if cmd_result.returncode != 0 or not 'Succeeded: true' in cmd_result.stdout or test_failed(logs):
```
jchaffraix, line 720: I don't think `logs` is updated in this loop (it is pulled from the PHP server in the web tests) so we can remove the check here and dumping it on error (l. 723).

srjfoo (approved):

SETUP/tests/smoketests/pageload_smoketest.py
cpeel: Indeed. I assumed (incorrectly) this was tied to the system `php_errors` but it isn't. I also discovered that we were relying on the default error level being E_ALL which it isn't. I've added a commit that explicitly sets the error level and display errors config. This resulted in PHP 8.3 showing deprecation errors that we'll need to resolve, so for now I've disabled the PHP 8.3 smoke tests.

jchaffraix (approved):

cpeel:
I pulled the changes to the error reporting level out of this PR. It fits better into the next one where I've resolved the deprecation issues.

PR #1284, opened 2024-07-17
Fix deprecation warnings for PHP 8.1 and 8.3

cpeel:
This enables all errors (errors, warnings, notices, and deprecation warnings) in our smoke tests which revealed several deprecation warnings for PHP 8.1 and 8.2 (yay smoke tests!). The other commits address these issues, which generally fell into:
* Passing `null` into functions that expect strings ([deprecated in 8.1](https://php.watch/versions/8.1/internal-func-non-nullable-null-deprecation))
* Old `${projectid}` vs `{$projectid}` syntax in string interpolation ([deprecated in 8.2](https://php.watch/versions/8.2/$%7Bvar%7D-string-interpolation-deprecated))
* Using dynamic properties in classes ([deprecated in 8.2](https://php.watch/versions/8.2/dynamic-properties-deprecated))

That last one revealed that we have two fields in the `projects` table that are never used. &lt;sigh&gt; This drops the field from the database.

Sandbox at https://www.pgdp.org/~cpeel/c.branch/fix-deprecation-warnings/


srjfoo:
> That last one revealed that we have two fields in the `projects` table that are never used. <sigh> This drops the field from the database.

I don't know what `year` was ever used for, but `updated` dates, I think, from almost DP's beginning of time, before PG and DP separated. There apparently used to be a method by which select individuals could update ebooks directly, and this was recorded in the database. Though, interestingly, I would have thought the default would have been `0` for `false`. The database reports the following for `updated`:
```
+---------+----------+
| updated | count(*) |
+---------+----------+
|       0 |      707 |
|       1 |    65925 |
+---------+----------+
```

I don't know if that's information that we want to preserve or not.

chrismiceli (approved):

pinc/ProjectSearchResults.inc
```diff
@@ -665,8 +671,7 @@ class ProjectSearchResults
         echo "</tr>\n";
 
         while ($project_assoc = mysqli_fetch_assoc($result)) {
-            $project = new Project($project_assoc);
-            $projectid = $project->projectid;
+            $project = new ProjectSearchResultEntry($project_assoc);
```
jchaffraix, line 674: Nit: Adding this new type is reasonable, though I have a hard finding if/where `$hold` is set...

Also the name is a bit confusing (it's not a `$project`, it's a `$project_result_entry`) so it may be worth renaming it so we are clear on what we are manipulating in this function.

jchaffraix (approved):

pinc/ProjectSearchResults.inc
cpeel: `hold` is loaded into `$project_assoc` array via the SQL at https://github.com/DistributedProofreaders/dproofreaders/blob/master/pinc/ProjectSearchResults.inc#L615 and set as variable in the `Project` [constructor](https://github.com/DistributedProofreaders/dproofreaders/blob/master/pinc/Project.inc#L240) from that associative array.

I'm going to add a comment about renaming this variable when we add type definitions to this file as it will change a ton of class function definitions as part of it and we might as well do it all at once.

tools/project_manager/autorelease.inc
```diff
@@ -352,6 +354,8 @@ function maybe_release_projects(
  */
 class ReleaseRestrictor
 {
+    public static array $restrictors = [];
```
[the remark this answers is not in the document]
cpeel: Done these as fixup commits that I'll squash before merging.

cpeel:
> I don't know if that's information that we want to preserve or not.

Thanks for that context. I don't think this data is worth preserving though.

pinc/ProjectSearchResults.inc
jchaffraix: Thanks for the context on where `hold` comes from and sounds great!

jchaffraix (approved):

srjfoo (approved):

PR #1285, opened 2024-07-17
Type pinc/ProjectState.inc

jchaffraix:
The types are inferred from the usage.

This showed some inconsistencies in `get_forum`'s return that was noted but not fixed to prevent regressions.

Manually tested some transitions and creating a post associated with a new project in PHPBB.

_Update by cpeel:_ sandbox avail at https://www.pgdp.org/~cpeel/c.branch/julien_type_ProjectState/

### DIFF
--- a/SETUP/configuration.sh
+++ b/SETUP/configuration.sh
@@ -404,7 +404,7 @@ _WIKI_URL=$base_url/wiki
 # If there's a wiki or a blog that you want your users to use, set the
 # above variables to their URLs and links will appear on the navigation
 # bar. If you don't have a blog or wiki, set either or both of them to
-# the empty string, and they won't appear in the navigation bar.
+# zero, and they won't appear in the navigation bar.
 
 _FORUMS_TEAMS_IDX=THE_INDEX_NUMBER_OF_THIS_FORUM
 # Each team gets an automatically-created discussion topic in this forum
--- a/SETUP/configuration.sh
+++ b/SETUP/configuration.sh
@@ -404,10 +404,11 @@ _WIKI_URL=$base_url/wiki
 # If there's a wiki or a blog that you want your users to use, set the
 # above variables to their URLs and links will appear on the navigation
 # bar. If you don't have a blog or wiki, set either or both of them to
-# zero, and they won't appear in the navigation bar.
+# the empty string, and they won't appear in the navigation bar.
 
 _FORUMS_TEAMS_IDX=THE_INDEX_NUMBER_OF_THIS_FORUM
 # Each team gets an automatically-created discussion topic in this forum
+# Setting the index to -1 will prevent posting.
 
 _FORUMS_PROJECT_WAITING_IDX=THE_INDEX_NUMBER_OF_THIS_FORUM
 _FORUMS_PROJECT_AVAIL_IDX=THE_INDEX_NUMBER_OF_THIS_FORUM
@@ -416,17 +417,20 @@ _FORUMS_PROJECT_POSTED_IDX=THE_INDEX_NUMBER_OF_THIS_FORUM
 # Each project gets an automatically-created discussion topic, which
 # moves around among these 4 forums, depending on its state.
 # (The exact correspondence is defined in pinc/project_states.inc.)
+# Setting any index to -1 will prevent posting to it.
 
 _FORUMS_PROJECT_DELETED_IDX=THE_INDEX_NUMBER_OF_THIS_FORUM
 _FORUMS_PROJECT_COMPLETED_IDX=THE_INDEX_NUMBER_OF_THIS_FORUM
 # Projects that are deleted or completed and not posted (rare)
 # have a forum of their own. Consider putting these in the same
 # hidden forum.
+# Setting any index to -1 will prevent posting to it.
 
 _FORUMS_BEGIN_SITE_IDX=THE_INDEX_NUMBER_OF_THIS_FORUM
 _FORUMS_POST_PROCESSORS_IDX=THE_INDEX_NUMBER_OF_THIS_FORUM
 # These forums aren't involved in the operation of the site,
 # they just show up in various links.
+# Setting any index to -1 will prevent posting to it.
 
 _AUTO_POST_TO_PROJECT_TOPIC=false
 # If you set this to `true`, the code will automatically add a post to a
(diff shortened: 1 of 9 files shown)


pinc/ProjectState.inc
```diff
@@ -272,22 +277,17 @@ function construct_star_metal_sql()
 
 // -----------------------------------------------------------------------------
 // these functions can be removed later
```
jchaffraix, line 279: This comments apply to the 4 functions following.

I only removed `get_forum_id_for_project_state` to prevent propagating the odd return type (and keep this PR somewhat contained as some of the wrappers have a large amount of usage).

pinc/ProjectState.inc
```diff
@@ -28,39 +28,48 @@ class ProjectState
...
-    public static function get_all()
+    /** @return array<string, ProjectState> */
+    public static function get_all(): array
     {
         return self::$states;
     }
 
-    public static function get_medium_label($state)
+    // TODO: We should ensure that `$state` is never null. For now, we allow it.
+    public static function get_medium_label(?string $state): string
```
jchaffraix, line 52: Some smoketest were sending null to this function (yay to smoke tests!), thus I have relaxed the string arguments I couldn't be sure of to allow the null.

pinc/ProjectState.inc
```diff
@@ -28,39 +28,48 @@ class ProjectState
...
 
-    public static function get_all()
+    /** @return array<string, ProjectState> */
+    public static function get_all(): array
     {
         return self::$states;
     }
 
-    public static function get_medium_label($state)
+    // TODO: We should ensure that `$state` is never null. For now, we allow it.
```
cpeel, line 51: I think this is a good goal but I'm not sure we'll get there and thus that the comment is useful. We've played fast-and-loose with data types -- particularly null / empty string -- for 20 years and if we started adding comments like this around it would be in 75% of all of the cases where we pass strings into functions.

With that sad and somewhat fatalistic view I don't think we should include these comments.

jchaffraix: Yeah, fair point about those comments.

I have seen cases of types being improved over time but it's a lot of work and we're nowhere close to having enough type coverage, at which point such changes are easier to do...

Removed those comments and kept the one about the return type as it seemed more of a one-off. Happy to remove it too and stick to documenting our current types rather than thinking too much ahead.

cpeel: The work you are doing is a great foundation for us making further type improvements. I'm hoping after we have more type coverage (and more code coverage via smoke tests?) we can make broader changes like `?string` to `string` in place like this.

One-offs like return comment types are great and welcome!

pinc/ProjectState.inc
```diff
@@ -10,12 +10,12 @@ class ProjectState
...
-        $name,
-        $medium_label,
-        $label,
-        $forum,
-        $phase,
-        $star_metal
+        string $name,
+        string $medium_label,
+        string $label,
+        string $forum,
```
cpeel, line 16: `$forum` is interesting because it really should be an integer. For some reason we're putting them in `''`s in `site_vars.php.template` (see [line 56](https://github.com/DistributedProofreaders/dproofreaders/blob/master/pinc/site_vars.php.template#L56)) but they're 100% integers because of phpBB.

I think the Right Thing To Do is to remove the `''`s from the `site_vars.php.template` file and add a comment in `SETUP/configuration.sh` before those about them needing to be integers, then having `get_forum()`'s return type be an int. I'm also amenable to that being a fast-follow PR but I feel like in this case we're passing up an opportunity to fix this -- and it seems pretty surgical?

pinc/ProjectState.inc
```diff
@@ -28,39 +28,44 @@ class ProjectState
...
-    public static function get_label($state)
+    public static function get_label(?string $state): string
     {
         return self::$states[$state]->label ?? '';
     }
 
-    public static function get_forum($state)
+    /** @return string|int */
+    // TODO: Switch the return to ?string but would need to inspect usage of the function.
+    public static function get_forum(?string $state)
```
srjfoo, line 63: Because I'm learning about this as we go, I just wanted to verify that this function is intentionally left without a return type (referencing the comment above it).

chrismiceli (approved):

SETUP/configuration.sh
```diff
@@ -404,7 +404,7 @@ _WIKI_URL=$base_url/wiki
 # If there's a wiki or a blog that you want your users to use, set the
 # above variables to their URLs and links will appear on the navigation
 # bar. If you don't have a blog or wiki, set either or both of them to
-# the empty string, and they won't appear in the navigation bar.
+# zero, and they won't appear in the navigation bar.
```
cpeel, line 407: This change is incorrect as it is referring to the blog and wiki, not the forum indexes. For reasons I don't fully understand Michael likes to add comments after things instead of before them.

cpeel: We should recommend `-1` as a value for forum indexes that don't exist. That's at least the value we return if there isn't a forum index set for a Pool.

jchaffraix: Oh wow, didn't realize the comment applied to the previous one...

Fixed and added a comment about `-1` disabling posting to all the forum indexes.

phpstan.neon
```diff
@@ -37,3 +37,14 @@ parameters:
         # Low value and frequent L5 errors that are ignored by default.
         - message: "#^Parameter \\#1 \\$.* of function array_multisort is passed by reference, so it expects variables only\\.$#"
         - message: '#^Parameter \#\d+ .* of (static )?(function|method) .* expects string(\|null)?, int.* given\.$#'
+
+        # TODO(jchaffraix): Temporarily disabling those as they are dynamically created (see comment in Pool's constructor).
```
jchaffraix, line 41: I am seeing those errors locally (which makes sense as we don't define those). As a lot can go wrong with this PR already, I would prefer to defer fixing those in a follow-up PR.

pinc/ProjectState.inc
```diff
@@ -69,7 +72,11 @@ class ProjectStates
 // Note that the order in which these project states are declared
 // is the order in which they will be displayed in various contexts
 // (via ProjectStates::get_states()).
-function declare_project_states($rounds, $pools)
+/**
```
cpeel, line 75: Let's add a blank line between the comment blocks to make it clearer they are separate.
```suggestion

/**
```

cpeel (approved):

srjfoo (approved):

PR #1286, opened 2024-07-18
Add get_bool_param() validation function

cpeel:
Add a validation function to return a bool from various sensible "truthy" and "falsey" values. This isn't used anywhere but I think it will come in handy based on recent type declaration PRs. It even comes with unit tests!

### DIFF
--- a/SETUP/tests/unittests/ParamValidatorTest.php
+++ b/SETUP/tests/unittests/ParamValidatorTest.php
@@ -330,7 +330,8 @@ class ParamValidatorTest extends PHPUnit\Framework\TestCase
         $this->assertEquals(false, $result);
     }
 
-    public function testBoolDefaultToNull() {
+    public function testBoolDefaultToNull()
+    {
         $default = null;
         $allow_null = true;
         $result = get_bool_param($this->GET, 'none', $default, $allow_null);
--- a/SETUP/tests/unittests/ParamValidatorTest.php
+++ b/SETUP/tests/unittests/ParamValidatorTest.php
@@ -323,18 +323,18 @@ class ParamValidatorTest extends PHPUnit\Framework\TestCase
     //------------------------------------------------------------------------
     // get_bool_param() tests
 
-    public function testBool()
-    {
-        $default = null;
-        $result = get_bool_param($this->GET, 'true_str', $default);
-        $this->assertEquals(true, $result);
-    }
-
     public function testBoolDefault()
     {
         $default = false;
         $result = get_bool_param($this->GET, 'none', $default);
         $this->assertEquals(false, $result);
+    }
+
+    public function testBoolDefaultToNull() {
+        $default = null;
+        $allow_null = true;
+        $result = get_bool_param($this->GET, 'none', $default, $allow_null);
+        $this->assertEquals(null, $result);
     }
 
     public function testBoolTrueVariants()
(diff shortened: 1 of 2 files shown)


srjfoo (approved):

SETUP/tests/unittests/ParamValidatorTest.php
```diff
@@ -309,6 +320,78 @@ public function testFloatMoreThanMax()
         get_float_param($this->GET, 'f10', $default, $min, $max);
     }
 
+    //------------------------------------------------------------------------
+    // get_bool_param() tests
+
+    public function testBool()
```
jchaffraix, line 326: This test is identical to the second test case in `testBoolTrueVariants`, is it worth the duplication?

pinc/misc.inc
```diff
@@ -419,6 +419,65 @@ function get_numeric_param(string $type, array $arr, string $key, $default, $min
...
+ * "false" and various capitalizations thereof.
+ *
+ * @param array<string, mixed> $arr
+ * @param string $key
+ * @param bool|null $default
+ * @param bool $allownull
+ *
+ * @return bool|null
+ *
+ * @throws InvalidArgumentException
```
jchaffraix, line 435: It would probably be better to either set the PHP types or PHPStan annotations, but not both. Having both could lead to inconsistencies between what PHPStan analyzes and the runtime behavior.

Here as we set the PHP types so we could simplify this to only the types that can't be represented by PHPStan:

```suggestion
 * @param array<string, mixed> $arr
 *
 * @throws InvalidArgumentException
```

Thoughts?

pinc/misc.inc
```diff
@@ -419,6 +419,65 @@ function get_numeric_param(string $type, array $arr, string $key, $default, $min
...
+    } else {
+        // not set, use default
+        if (is_null($default) && !$allownull) {
+            // There is no default. The parameter is required.
+            throw new InvalidArgumentException(sprintf(
+                _("Parameter '%1\$s' is required"),
+                $key
+            ));
+        } else {
+            return $default;
```
jchaffraix, line 476: Nit: This branch is untested, is it worth adding a test for it?

I was thinking something like:

```
public function testBoolDefaultToNull() {
    $default = null;
    $allow_null = true;
    $result = get_bool_param($this->GET, 'i10', $default, $allow_null);
    $this->assertEquals(null, $result);
}
```

jchaffraix (approved):

pinc/misc.inc
cpeel: Good catch, added.

pinc/misc.inc
cpeel: We originally added these for (not yet completed, obviously) documentation not for phpstan and your suggestion will work here because these params don't have descriptions, but elsewhere we add param descriptions not just types. Any thoughts on how to handle this more broadly?

SETUP/tests/unittests/ParamValidatorTest.php
cpeel: Yup, removed.

pinc/misc.inc
jchaffraix: I would say that the need for good documentation should trump the potential inconsistencies in PHPStan. Those are mostly theoretical at this point and we can revisit this if we find that they cause issues.

Here just adding the names+types of the parameters seems like it could be generated from the function signature (though I don't know if that's how PHPDoc work). Those with descriptions do add valuable documentation so we should keep those.

jchaffraix (approved):

PR #1287, opened 2024-07-21
Fix comments for PHPStan

jchaffraix:
I made some errors when adding comments for PHPStan.

The main one is that we must be using PHPDoc's format (/** @param type param_name */) or else those are
ignored by PHPStan.

This exposed a bad array syntax in
tools/proofers/spellcheck_text.inc that got fixed.

### DIFF
--- a/SETUP/ci/check_security.php
+++ b/SETUP/ci/check_security.php
@@ -63,6 +63,7 @@ foreach ($files as $file) {
 
 }
 
+/** @return string[] */
 function get_all_php_files(string $basedir): array
 {
     $php_files = [];
--- a/SETUP/ci/check_security.php
+++ b/SETUP/ci/check_security.php
@@ -63,7 +63,6 @@ foreach ($files as $file) {
 
 }
 
-// @param $basedir string[]
 function get_all_php_files(string $basedir): array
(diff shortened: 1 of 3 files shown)


tools/proofers/spellcheck_text.inc
```diff
@@ -386,8 +386,12 @@ $word_check_messages = [
     "rerun" => _("Check page against an additional language"),
 ];
 
-// @param $accepted_words string[]
-// @param $languages string[]
+/**
+ * output_wordcheck_interface outputs all the buttons used by WordCheck
```
cpeel, line 390:
```suggestion
 * Output all the buttons used by WordCheck.
```

SETUP/ci/check_security.php
```diff
@@ -63,7 +63,6 @@
 
 }
 
-// @param $basedir string[]
 function get_all_php_files(string $basedir): array
```
cpeel, line 66: Perhaps?
```php
/** @return string[] */
```

jchaffraix: Good catch, done.

cpeel (approved):

PR #1288, opened 2024-07-21
Better setlangcookie param validation

cpeel:
Last night's php_errors report showed some notices after someone was poking around in `tools/setlangcookie.php`. This addresses a few of them by tightening up param validation and enforcing param types in `metarefresh()`. I vetted the param types of every `metarefresh()` call in the code with this change and cleaned up a few optional parameters on some of the calls while I was there.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/better_param_vetting/

### DIFF
--- a/pinc/metarefresh.inc
+++ b/pinc/metarefresh.inc
@@ -10,7 +10,8 @@
  * client (determined if headers have been sent or if the output buffer
  * has contents in it).
  */
-function metarefresh($seconds, $url, $title = "", $body = "", $allow_external = false)
+/** @return never */
+function metarefresh(int $seconds, string $url, string $title = "", string $body = "", bool $allow_external = false)
 {
     global $testing, $code_url;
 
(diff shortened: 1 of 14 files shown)


chrismiceli (approved):

srjfoo (approved):

pinc/metarefresh.inc
```diff
@@ -10,7 +10,7 @@
  * client (determined if headers have been sent or if the output buffer
  * has contents in it).
  */
-function metarefresh($seconds, $url, $title = "", $body = "", $allow_external = false)
+function metarefresh(int $seconds, string $url, string $title = "", string $body = "", bool $allow_external = false)
```
jchaffraix, line 13: Nit: We may want to add a PHP return as this function should always `exit` for security reason.

```suggestion
/** @return never */
function metarefresh(int $seconds, string $url, string $title = "", string $body = "", bool $allow_external = false)
```

jchaffraix (approved):
This looks great. The suggestion can be a follow-up if you want to merge this sooner rather than later!

pinc/metarefresh.inc
cpeel: Great suggestion - thanks!

cpeel: Needed to remove some `exit;`s that were after calls to this function.

jchaffraix (approved):

srjfoo (approved):

PR #1292, opened 2024-07-28
Correct postednum & topic_id type definitions

cpeel:
These are both null or ints but we need to fix some places where they weren't assigned as such.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/project-field-types/

### DIFF
--- a/pinc/Project.inc
+++ b/pinc/Project.inc
@@ -200,8 +200,8 @@ class Project
     public string $checkedoutby;
     public string $difficulty;
     public string $username;
-    public ?string $postednum; // TODO: Should be int
-    public ?string $topic_id; // TODO: Should be int
+    public ?int $postednum;
+    public ?int $topic_id;
     public float $days_checkedout;
     public int $n_available_pages;
     public int $n_pages;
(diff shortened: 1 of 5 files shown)


srjfoo (approved):

tools/project_manager/editproject.php
```diff
@@ -275,12 +275,12 @@ public function set_from_post()
...
         }
 
-        // postednum needs to either be a valid number or null, not an empty string
+        // postednum needs to either be a valid int or null, not an empty string
         if (isset($_POST["postednum"])) {
             if ($_POST["postednum"] == "") {
                 $this->project->postednum = null;
             } else {
-                $this->project->postednum = sprintf("%d", $_POST["postednum"]);
+                $this->project->postednum = (int)$_POST["postednum"];
```
jchaffraix, line 283: Is it worth using `get_integer_param` here? I think this would be equivalent:

```suggestion
                $this->project->postednum = get_integer_param($_POST, "postednum");
```

pinc/Project.inc
```diff
@@ -452,19 +452,12 @@ class Project
...
-                // the etext, however, does not include the 'C', nor should
-                // the DP link. If this changes, update the pattern here.
-            }
-        }
+        // $this->postednum is either null or an integer by type definition.
+        // Occasionally, there will be a PG ebook that is still
+        // under U.S. copyright. This is indicated in their system
+        // by appending a 'C' to the etext number. The link to
+        // the etext, however, does not include the 'C', nor should
+        // the DP link.
```
jchaffraix, line 460: Would it worth hoisting this comment next to the declaration of `$postednum` above now that we don't do any validation? This seems like a worthwhile gotcha about the field.

jchaffraix (approved):

tools/project_manager/editproject.php
cpeel: It isn't quite that straightforward as `get_integer_param()` requires a default, a min, and a max. I think in this particular case its better to take the number provided and just make it an int as we don't control the range of numbers PG provides.

pinc/Project.inc
cpeel: I tried that and it just felt out of place in the definition. It feels a little more intuitive here where we are doing validation.

PR #1293, opened 2024-07-29
Fully type pinc/UserProfile.inc

jchaffraix:
The types were inferred from usage.

Also added an explicit integer cast on the
`profile` cookie as the comment mentioned
it was an ID.

_Edit by srjfoo_: Sandbox at https://www.pgdp.org/~srjfoo/c.branch/julien_fully_type_UserProfile/

### DIFF
--- a/pinc/User.inc
+++ b/pinc/User.inc
@@ -203,7 +203,7 @@ class User
                 // different browsers/devices.
                 if ($this->username == User::current_username() && @$_COOKIE["profile"]) {
                     try {
-                        $profile = new UserProfile($_COOKIE["profile"]);
+                        $profile = new UserProfile((int)$_COOKIE["profile"]);
                         if ($profile->u_ref != $this->u_id) {
                             throw new NonexistentUserProfileException(
                                 "Profile does not belong to user"
--- a/pinc/UserProfile.inc
+++ b/pinc/UserProfile.inc
@@ -48,7 +48,7 @@ class UserProfile
         "h_fntf_other",
     ];
 
-    public function __construct($id = null)
+    public function __construct(?int $id = null)
     {
         if ($id !== null) {
             $this->load($id);
@@ -58,7 +58,9 @@ class UserProfile
     // The __set() and __get() methods allow access to user fields without
     // creating accessors for them all individually.
     // See the PHP docs for "magic methods".
-    public function __set($name, $value)
+
+    /** @param mixed $value */
+    public function __set(string $name, $value)
     {
         if (isset($this->$name) && in_array($name, $this->immutable_fields)) {
             throw new DomainException(
@@ -72,17 +74,18 @@ class UserProfile
         $this->table_row[$name] = $value;
     }
 
-    public function __get($name)
+    /** @return mixed */
+    public function __get(string $name)
     {
         return $this->table_row[$name];
     }
 
-    public function __isset($name)
+    public function __isset(string $name): bool
     {
         return isset($this->table_row[$name]);
     }
 
-    private function load($id)
+    private function load(int $id): void
     {
         $sql = sprintf(
             "
@@ -107,7 +110,7 @@ class UserProfile
         mysqli_free_result($result);
     }
 
-    public function save()
+    public function save(): void
     {
         if (!isset($this->id)) {
             // validate that we have a u_ref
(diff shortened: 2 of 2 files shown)


chrismiceli (approved):

srjfoo:
I'm not entirely sure exactly how I got this:
> Fatal error: Uncaught InvalidArgumentException: Parameter 'c_profile' is required in /home/srjfoo/public_html/c.branch/julien_fully_type_UserProfile/pinc/misc.inc:412 Stack trace: #0 /home/srjfoo/public_html/c.branch/julien_fully_type_UserProfile/pinc/misc.inc(262): get_numeric_param() #1 /home/srjfoo/public_html/c.branch/julien_fully_type_UserProfile/userprefs.php(60): get_integer_param() #2 {main} thrown in /home/srjfoo/public_html/c.branch/julien_fully_type_UserProfile/pinc/misc.inc on line 412

I think it has something to do with the maximum number of profiles (at least, when I was testing, I had (without realizing it because I hadn't looked at the pophelp) tried to exceed the number of profiles. So, question: as far as I can tell, this PR is good, and I'll go ahead and approve it on that basis, but in a different PR, perhaps, should the text entry field for a new profile name be disabled if a user has 10 profiles already?

srjfoo (approved):

cpeel:
> I'm not entirely sure exactly how I got this:
> 
> > Fatal error: Uncaught InvalidArgumentException: Parameter 'c_profile' is required in /home/srjfoo/public_html/c.branch/julien_fully_type_UserProfile/pinc/misc.inc:412 Stack trace: #0 /home/srjfoo/public_html/c.branch/julien_fully_type_UserProfile/pinc/misc.inc(262): get_numeric_param() #1 /home/srjfoo/public_html/c.branch/julien_fully_type_UserProfile/userprefs.php(60): get_integer_param() #2 {main} thrown in /home/srjfoo/public_html/c.branch/julien_fully_type_UserProfile/pinc/misc.inc on line 412
> 
> I think it has something to do with the maximum number of profiles (at least, when I was testing, I had (without realizing it because I hadn't looked at the pophelp) tried to exceed the number of profiles. So, question: as far as I can tell, this PR is good, and I'll go ahead and approve it on that basis, but in a different PR, perhaps, should the text entry field for a new profile name be disabled if a user has 10 profiles already?

I don't know how you got that either, actually. We already enforce a limit of 10 by not providing a "Save as New Profile" button if they have reach the limit already. We can't disable the text box as that's how users rename the profile that they are on. I'm not certain that the error was from the number of profiles but I don't have a better idea of how you got it either.

srjfoo:
> We can't disable the text box as that's how users rename the profile that they are on.

Good point.

And I did manage to rename a profile accidentally while I was still unaware of the max. (Can you tell I never used the profiles much?)

jchaffraix:
> I'm not entirely sure exactly how I got this:
> 
> > Fatal error: Uncaught InvalidArgumentException: Parameter 'c_profile' is required in /home/srjfoo/public_html/c.branch/julien_fully_type_UserProfile/pinc/misc.inc:412 Stack trace: #0 /home/srjfoo/public_html/c.branch/julien_fully_type_UserProfile/pinc/misc.inc(262): get_numeric_param() #1 /home/srjfoo/public_html/c.branch/julien_fully_type_UserProfile/userprefs.php(60): get_integer_param() #2 {main} thrown in /home/srjfoo/public_html/c.branch/julien_fully_type_UserProfile/pinc/misc.inc on line 412
> 
> I think it has something to do with the maximum number of profiles (at least, when I was testing, I had (without realizing it because I hadn't looked at the pophelp) tried to exceed the number of profiles. So, question: as far as I can tell, this PR is good, and I'll go ahead and approve it on that basis, but in a different PR, perhaps, should the text entry field for a new profile name be disabled if a user has 10 profiles already?

I think the issue is actually unrelated to the number of profiles.

It is possible to click on the "Switch profiles" button without having selected a profile to switch to (what `c_profile` stores in the error above). The current profile is the default option but it is explicitly marked as non-selectable through the HTML5 `disabled` attribute. This means that opening the user's Proofreading preference page and clicking on the "Switch profiles" button - without changing the selected profile - triggers the issue.

It seems to me like we should either disable the "Switch profiles" button when the default profile is selected or allow switching to the current profile (which is a no-op, except it does a round-trip through the server).

cpeel:
> It seems to me like we should either disable the "Switch profiles" button when the default profile is selected or allow switching to the current profile (which is a no-op, except it does a round-trip through the server).

Oh good catch. I can get a PR up for that.

PR #1295, opened 2024-08-05
Relax get_(integer|float|bool)_param() definitions

cpeel:
Sometimes we just want to do basic type validation and return a param of that type. We already accept null values for these, so make them optional. This implements what @jchaffraix suggested in https://github.com/DistributedProofreaders/dproofreaders/pull/1292 -- which is obvious in retrospect but it took me thinking about it post-merge to see it.

Note that this means that `get_{type}_param($array, $key)` must return a value of that type and not null since `$default=null` and `$allownull=false`. This makes sense to me.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/relax-type-getters/

### DIFF
--- a/pinc/misc.inc
+++ b/pinc/misc.inc
@@ -257,7 +257,7 @@ function get_enumerated_param(array $arr, string $key, $default, array $choices,
  *
  * @see get_float_param()
  */
-function get_integer_param(array $arr, string $key, ?int $default, ?int $min, ?int $max, bool $allownull = false): ?int
+function get_integer_param(array $arr, string $key, ?int $default = null, ?int $min = null, ?int $max = null, bool $allownull = false): ?int
 {
     return get_numeric_param("integer", $arr, $key, $default, $min, $max, $allownull);
 }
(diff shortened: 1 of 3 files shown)


srjfoo:
I tried adding a negative number as a postednum. Main test site, I got

image

IN the `relax-type-getters` sandbox, I got
> *Fatal error*: Uncaught DBQueryError: An error occurred during a database query and has been logged. in /home/cpeel/public_html/c.branch/relax-type-getters/pinc/DPDatabase.inc:106 Stack trace: `#`0 /home/cpeel/public_html/c.branch/relax-type-getters/pinc/Project.inc(565): DPDatabase::query() `#`1 /home/cpeel/public_html/c.branch/relax-type-getters/tools/project_manager/editproject.php(364): Project->save() `#`2 /home/cpeel/public_html/c.branch/relax-type-getters/tools/project_manager/editproject.php(33): ProjectInfoHolder->save_to_db() `#`3 {main} thrown in */home/cpeel/public_html/c.branch/relax-type-getters/pinc/DPDatabase.inc on line 106*

Trying to add an alpha character, or a decimal number, got the same responses in both sandbox and on TEST

cpeel:
The postednum field on the project edit form now enforces positive numbers.

> Tangentially, and "not in this PR", should we allow a posted project to have a null value for the `postednum` field, and should we allow the `postednum` field to contain a value if the project has not been marked posted?)

Probably not. Wonder how many (if any) projects on PROD fall into these two categories.

srjfoo (approved):

70ray (approved):

PR #1296, opened 2024-08-06
Add an explicit Content Security Policy (CSP)

cpeel:
This adds an explicit [content security policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy) beyond preventing other sites from embedding our code in frames/iframes. This isn't a very high bar since we still allow inline JS and CSS but it is a small step in the right direction since we now disallow unsafe-evals.

The new CSP should let everything else work, including:
* the use of MathJAX in Format Preview
* the use of JS to manipulate CSS (used in Format Preview)
* the use of images from anywhere in project comments

The following pages were updated to remove unsafe evals:
* https://www.pgdp.org/~cpeel/c.branch/explicit-csp/tools/authors/manage.php
* https://www.pgdp.org/~cpeel/c.branch/explicit-csp/userprefs.php
* Leaving the proofreading interface

The kicker is going to be finding what other edgecases the new policy doesn't cover. Testing is best done with the developer tools console open.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/explicit-csp/

### DIFF
--- a/tools/authors/manage.php
+++ b/tools/authors/manage.php
@@ -183,12 +183,12 @@ function deleteAuthor(check, author) {
             check.checked = false;
             return;
         }
-        eval("form.new_enabled_author_"+author+".disabled=true;");
+        document.getElementsByName("new_enabled_author_" + author)[0].disabled = true;
         setMoveToAuthorDisabled(form, author, true);
         setAuthorsBiosChecksDisabled(form, author, true);
     }
     else {
-        eval("form.new_enabled_author_"+author+".disabled=false;");
+        document.getElementsByName("new_enabled_author_" + author)[0].disabled = false;
         setMoveToAuthorDisabled(form, author, false);
         setAuthorsBiosChecksDisabled(form, author, false);
     }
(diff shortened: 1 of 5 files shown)


tools/authors/manage.php
```diff
@@ -183,12 +183,12 @@ function deleteAuthor(check, author) {
             check.checked = false;
             return;
         }
-        eval("form.new_enabled_author_"+author+".disabled=true;");
+        document.getElementsByName("new_enabled_author_" + author)[0].disabled = true;
```
cpeel, line 186: These `document.getElementsByName()[0]` are safe in this context because the names are unique in the page.

tools/proofers/processtext.php
```diff
@@ -346,6 +346,6 @@ function leave_proofing_interface(string $title): void
     $text = _("You will be returned to the <a href='%s' target='_top'>project page</a> in one second.");
     echo sprintf($text, $safe_url);
     echo "<script><!--\n";
-    echo "setTimeout(\"top.location.href='$raw_url';\", 1000);\n";
```
chrismiceli, line 349: reading these evals and setTimeout with strings really taking me back in time!

chrismiceli (approved):

srjfoo (approved):

userprefs.php
```diff
@@ -122,13 +122,9 @@
...
     // First parameter: true/false.
     // Following parameters: The name of the checkboxes.
-    // The code checks that a checkbox really exists
-    // before accessing it.
     function check_boxes(value) {
-        var f = document.forms[0];
         for (var i = 1; i < arguments.length; i++) {
-            var name = arguments[i];
-            eval('if (f.'+name+') f.'+name+'.checked=value');
+            document.getElementsByName(arguments[i])[0].checked = value;
```
70ray, line 127: We should check that the box exists. This function is used in _show_credits_wanted_adhoc() where pm_credit might not exist and causes a TypeError exception when changing all the check boxes.

cpeel: Ah, good catch - I missed that the PM checkbox was conditional. I've fixed it.

70ray (approved):

srjfoo (approved):

Issue #1274, opened 2024-07-11
Convert system/exec usages to Symfony's Process class


cpeel:
Replace all applicable uses of [exec](https://secure.php.net/manual/en/function.exec.php), [passthru](https://secure.php.net/manual/en/function.passthru.php), [shell_exec](https://secure.php.net/manual/en/function.shell-exec.php) and [system](https://secure.php.net/manual/en/function.system.php) with Symfony's [Process](https://symfony.com/doc/5.x/components/process.html) class. The package/class was introduced into the codebase in https://github.com/DistributedProofreaders/dproofreaders/pull/1273

Add a CI script that checks these functions aren't used in the code with an approve-list for files we've allowed as an exceptions. Since this is security related, we might change `check_security.sh` into a PHP file similar to `check_includes.php` and put it in there but other approaches are fine.

cpeel:
Note: iterative work on this issue is appreciated, it doesn't need to be solved all at once.

mrducky4:
I will start working on this.

PR #1297, opened 2024-08-06
Partial fix for issue #1274, converting exec() to Process()

mrducky4:
I tested in the development VM.
I tested the translators change from the Translation Center page, as an admin, deleted the locale is_IS and verified the files and directories were actually deleted.
I tested the clean uploads trash job with `run_background_job.php CleanUploadsTrash` and verified it deletes TRASH/... files older than 30 days and deletes empty folders.

_Update by cpeel_: Sandbox avail at https://www.pgdp.org/~cpeel/c.branch/process_class/

### DIFF
--- a/crontab/CleanUploadsTrash.inc
+++ b/crontab/CleanUploadsTrash.inc
@@ -19,38 +19,34 @@ class CleanUploadsTrash extends BackgroundJob
         // remove files from TRASH subdirectory older than 30 days
         $process = new Process([
             "/usr/bin/find",
-            "$trash_dir",
+            $trash_dir,
             "-type", "f",
             "-mtime", "+30",
             "-delete",
         ]);
         $process->run();
-        $output = $process->getOutput();
-        $exitcode = $process->getExitCode();
         if (!$process->isSuccessful()) {
             echo "An error occurred while cleaning up files.\n";
-            echo "Return value: {$exitcode}\n";
+            echo "Return value: {$process->getExitCode()}\n";
             echo "Command output:\n";
-            echo $output;
+            echo $process->getOutput();
             throw new RuntimeException("An error occurred while removing files");
         }
 
         // remove empty directories
         $process = new Process([
             "/usr/bin/find",
-            "$trash_dir",
+            $trash_dir,
             "-type", "d",
             "-empty",
             "-delete",
         ]);
         $process->run();
-        $output = $process->getOutput();
-        $exitcode = $process->getExitCode();
         if (!$process->isSuccessful()) {
             echo "An error occurred while cleaning up files.\n";
-            echo "Return value: {$exitcode}\n";
+            echo "Return value: {$process->getExitCode()}\n";
             echo "Command output:\n";
-            echo $output;
+            echo $process->getOutput();
             throw new RuntimeException("An error occurred while removing empty directories");
         }
     }
--- a/crontab/CleanUploadsTrash.inc
+++ b/crontab/CleanUploadsTrash.inc
@@ -1,4 +1,5 @@
 <?php
+use Symfony\Component\Process\Process;
 
 // Clean the Uploads/TRASH subdirectory.
 class CleanUploadsTrash extends BackgroundJob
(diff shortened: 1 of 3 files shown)


crontab/CleanUploadsTrash.inc
```diff
@@ -16,42 +17,40 @@ class CleanUploadsTrash extends BackgroundJob
         }
 
         // remove files from TRASH subdirectory older than 30 days
-        $cmd = join(" ", [
+        $process = new Process([
             "/usr/bin/find",
-            escapeshellarg($trash_dir),
-            "-type f",
-            "-mtime +30",
+            "$trash_dir",
```
cpeel, line 22: Minor -- no need for quotes around the variable.
```suggestion
            $trash_dir,
```

crontab/CleanUploadsTrash.inc
```diff
@@ -16,42 +17,40 @@ class CleanUploadsTrash extends BackgroundJob
...
         }
 
         // remove empty directories
-        $output = $return = null;
-        $cmd = join(" ", [
+        $process = new Process([
             "/usr/bin/find",
-            escapeshellarg($trash_dir),
-            "-type d",
+            "$trash_dir",
```
cpeel, line 41:
```suggestion
            $trash_dir,
```

cpeel (approved):
Looks good!

crontab/CleanUploadsTrash.inc
```diff
@@ -16,42 +17,40 @@ class CleanUploadsTrash extends BackgroundJob
...
+            "-type", "f",
+            "-mtime", "+30",
             "-delete",
         ]);
-        $output = $return = null;
-        exec($cmd, $output, $return);
-        if ($return != 0) {
+        $process->run();
+        $output = $process->getOutput();
+        $exitcode = $process->getExitCode();
```
jchaffraix, line 29: Nit: We should move `$exitcode` into the `$process->isSuccessful()` branch as it is only needed there.

crontab/CleanUploadsTrash.inc
```diff
@@ -16,42 +17,40 @@ class CleanUploadsTrash extends BackgroundJob
...
+            "$trash_dir",
+            "-type", "d",
             "-empty",
             "-delete",
         ]);
-        exec($cmd, $output, $return);
-        if ($return != 0) {
+        $process->run();
+        $output = $process->getOutput();
+        $exitcode = $process->getExitCode();
```
jchaffraix, line 48: Nit: Ditto about moving $exitcode into the branch.

locale/translators/index.php
```diff
@@ -143,7 +145,12 @@
 elseif ($func == "delete") {
     $locale = validate_locale($_REQUEST['locale']);
     assert(is_dir("$dyn_locales_dir/$locale"));
-    exec("rm -r " . escapeshellarg("$dyn_locales_dir/$locale"));
+    $process = new Process([
+        "rm",
+        "-r",
+        "$dyn_locales_dir/$locale",
+    ]);
+    $process->run();
```
jchaffraix, line 154: This seems to be an existing error that we don't check if the run was successful. Is it worth adding a TODO about it?

jchaffraix (approved):

locale/translators/index.php
mrducky4: What would be the appropriate way to handle such an error? Just log it? Or report it back to the UI?

mrducky4:
Thanks for the comments. I can make the minor updates to address the feedback.

locale/translators/index.php
cpeel: I would probably throw a `RuntimeException()` in this instance. Unfortunately we don't yet have a standardized logging mechanism.

mrducky4: Ok, that's easy. It looks like the exception message strings should be localized, although most of them I see in the code are not. What's the policy for this?

cpeel: Either is fine. It looks like in general we haven't localized `RuntimeExceptions` probably either because we expect those to be things that the user hasn't done / can't control and rare or, more likely, we just weren't thinking about it too much.
```bash
cpeel@test:~/u-cpeel-dp $ grep -r RuntimeException * | grep -v vendor | grep -c '_('
1
cpeel@test:~/u-cpeel-dp $ grep -r RuntimeException * | grep -v vendor | grep -cv '_('
20
```

cpeel (approved):

jchaffraix (approved):
Thanks for fixing the lack of error handling!

Issue #1298, opened 2024-08-09
Whittle away at inline `<script>` tags


cpeel:
Ideally we would have no inline JS which would allow us to tighten down our Content Security Policy. That's going to be hard to remove entirely, but if we can remove all of the pages that specify `<script>` we can use a `nonce` to cover the ones included in `output_header()` and `slim_header()` via `$extra_args["js_data"]` (see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src#unsafe_inline_script).

So we should:
* Move JS code to `.js` files when possible
* Any code we can't move to `.js`. files needs to be lifted into `$extra_args["js_data"]` so it can be covered under a (to-be-developed) nonce mechanism.

Closing this ticket would be having no `<script>` content in PHP code outside that created in `html_page_common.inc` and a nonce mechanism cover the header-inserted code.

chrismiceli:
If we do ever get to disable inline scripts, be aware that I believe it may break people's extensions, which is kind of the point, but we may get some feedback in that regard.

cpeel:
@chrismiceli - do you have thoughts or suggestions on how to reduce the number of individual `.js` files that we create as part of this effort? I know that it's a best practice to reduce the number of round-trips to the server and this feels like we'll be creating a bunch of individual `.js` files.

chrismiceli:
> @chrismiceli - do you have thoughts or suggestions on how to reduce the number of individual `.js` files that we create as part of this effort? I know that it's a best practice to reduce the number of round-trips to the server and this feels like we'll be creating a bunch of individual `.js` files.

We definitely could create 1 or more minified bundles based on usage patterns or roles, but really I'm not 100% sure it is worth the effort. With http1.1 with keep alive or http2/3, there is only 1 TCP connection per server so there isn't a huge penalty to not bundling. Even more modern sites are kind of going away with it and instead having tons of es modules with various imports, etc.

Minifying would probably be trivial to do and reduce page load size, but our client side performance according to lighthouse is already really good: https://pagespeed.web.dev/analysis/https-www-pgdp-net-c/wf00h7vgv8?form_factor=desktop. Minifying does make debugging harder, but not really that much to be honest. 

If you want me to investigate what these strategies could look like for pgdp, I could, but want to confirm the goals clearly.

cpeel:
tl;dr: Hold off on any investigation until I see if we can support http/2.

[...]

PR #1299, opened 2024-08-12
remove <script> from page_table.inc

70ray:
contribution to issue #1298 
Sandbox at: https://www.pgdp.org/~rp31/c.branch/unscript_project

### DIFF
--- a/pinc/page_table.inc
+++ b/pinc/page_table.inc
@@ -348,20 +348,20 @@ function echo_page_table(
 
     echo "</tr>";
 
-    $avail_pages = [];
-
     $row_count = 0;
     foreach (fetch_page_table_data($project, $page_selector, $proofed_in_round) as $page_res) {
         $row_count += 1;
 
         $imagename = $page_res['image'];
+        $page_state = $page_res['state'];
 
         echo "<tr>";
 
         // --------------------------------------------
         // Selector
         if ($can_edit) {
-            echo "<td><input type='checkbox' name='selected_pages[$imagename]'></td>\n";
+            $avail_class = page_state_is_a($page_state, "page_avail_state") ? " availbox" : "";
+            echo "<td><input type='checkbox' class='selectorbox$avail_class' name='selected_pages[$imagename]'></td>\n";
         }
 
         // --------------------------------------------
@@ -395,12 +395,7 @@ function echo_page_table(
         // --------------------------------------------
 
         // Page State
-        $page_state = $page_res['state'];
         echo "<td>$page_state</td>\n";
-
-        if (page_state_is_a($page_state, "page_avail_state")) {
-            $avail_pages[] = $row_count;
-        }
 
         // --------------------------------------------
         // Per-Round Info
@@ -476,38 +471,8 @@ function echo_page_table(
     echo "</table>";
 
     if ($can_edit) {
-        ?>
-<script><!--
-let avail_pages = <?php echo json_encode($avail_pages); ?>;
-let num_rows = <?php echo $row_count; ?>;
-
-function changeSelection(sel) {
-    switch(sel) {
-        case "all":
-            for(i=1; i<=num_rows; i++)
-                document.pagesform.elements[i].checked=true;
-            break;
-        case "unproofed":
-            changeSelection("clear");
-            for(i in avail_pages)
-                document.pagesform.elements[avail_pages[i]].checked=true;
-            break;
-        case "invert":
-            for(i=1; i<=num_rows; i++)
-                document.pagesform.elements[i].checked=!document.pagesform.elements[i].checked;
-            break;
-        case "clear":
-            for(i=1; i<=num_rows; i++)
-                document.pagesform.elements[i].checked=false;
-            break;
-        default:
-            break;
-    }
-}
-// --></script>
-<?php
         echo "<br>\n";
-        echo "<select name='select' onChange='changeSelection(this.options[this.selectedIndex].value)'>\n";
+        echo "<select name='select' id='option_selector'>\n";
         echo "<option value='0'>"._("Select...")."</option>";
         echo "<option value='all'>"._("Select all")."</option>";
         echo "<option value='unproofed'>"._("Select all unproofread")."</option>";
--- a/pinc/page_table.inc
+++ b/pinc/page_table.inc
@@ -360,8 +360,8 @@ function echo_page_table(
         // --------------------------------------------
         // Selector
         if ($can_edit) {
-            $avail_class = page_state_is_a($page_state, "page_avail_state") ? " availbox" : "";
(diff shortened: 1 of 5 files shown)


pinc/page_table.inc
```diff
@@ -348,20 +348,20 @@ function echo_page_table(
...
+        $page_state = $page_res['state'];
 
         echo "<tr>";
 
         // --------------------------------------------
         // Selector
         if ($can_edit) {
-            echo "<td><input type='checkbox' name='selected_pages[$imagename]'></td>\n";
+            $avail_class = page_state_is_a($page_state, "page_avail_state") ? " availbox" : "";
+            echo "<td><input type='checkbox' class='selectorbox$avail_class' name='selected_pages[$imagename]'></td>\n";
```
cpeel, line 364: Slight preference for keeping the class name just the name and not space + name -- the HTML won't care if there's an errant space in the `class` attr if `$avail_class` is empty -- and this makes it clearer that it's an additional class and not modifying the existing class.

```suggestion
            $avail_class = page_state_is_a($page_state, "page_avail_state") ? "availbox" : "";
            echo "<td><input type='checkbox' class='selectorbox $avail_class' name='selected_pages[$imagename]'></td>\n";
```

project.php
```diff
@@ -69,6 +69,8 @@
     '</script>',
 ];
 
+$extra_args["js_files"] = ["$code_url/scripts/page_table.js"];
+
```
cpeel, line 73:
```suggestion
    "js_files" => ["$code_url/scripts/page_table.js"],
];
```

cpeel (commented):
Thanks for working on this! Can you spin up a sandbox on TEST too please?

70ray:
Sandbox updated.

scripts/page_table.js
```diff
@@ -0,0 +1,36 @@
+window.addEventListener("DOMContentLoaded", () => {
+
+    let optionSelector = document.getElementById('option_selector');
+
+    function changeSelection(option) {
+        switch(option) {
+        case "all":
```
chrismiceli, line 7: Not sure if what I came up is cleaner, but just sharing it in case you did:

```
window.addEventListener("DOMContentLoaded", () => {

    let optionSelector = document.getElementById('option_selector');

    function changeSelection(option) {
      if (option === 'unproofed') {
        changeSelection('clear');
      }
      for (let box of document.getElementsByClassName('selectorbox')) {
        if (option === 'invert') {
          box.checked = option === 'all' || option === 'unproofed';
        } else if (option === 'all' || option === 'unproofed') {
          box.checked = true;
        } else {
          box.checked = false;
        }
      }
    };

    if(optionSelector) {
        optionSelector.addEventListener("change", function() {
            changeSelection(this.value);
        });
    }
});
```

chrismiceli (approved):

scripts/page_table.js
70ray: That is shorter, but maybe not so easy to understand (at least for me).

srjfoo (approved):

scripts/page_table.js
chrismiceli: Yeah I wasn't sure either. What you have is perfectly valid so I approved.

project.php
```diff
@@ -67,6 +67,7 @@
     "head_data" => '<script type="application/ld+json">' .
         json_encode($ld_json_object, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES) .
     '</script>',
+    "js_files" => ["$code_url/scripts/page_table.js"],
```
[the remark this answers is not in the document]
70ray: Good point. Done.

70ray:
> I think this PR didn't update [this call to `echo_page_table`](https://github.com/DistributedProofreaders/dproofreaders/blob/2e67b263cd13a22aaaff75014ab236061b10904b/project.php#L2056) that would need to include the new JS file, no?
> 
> I am slightly concerned about the hidden assumption that calling `echo_page_table` now requires to import the script in the caller or else the selection will not work. It would probably be better to have it output a `<script defer src="...">` inside `echo_page_table` but I could be missing something.

I'm not sure if I have understood your concern correctly but do_page_table() is only called in project.php so the js file will be loaded in the page head by output_html_header() in html_page_common.inc.
If I understand it correctly, issue #1298 requires removing all embedded js from php files, not because the js is bad in itself but because it makes it easier to enhance security by eventually disallowing it.

70ray:
Sandbox updated.

jchaffraix:
First, sorry if my previous comment was unclear. Also I am still fairly new to DP development so feel free to point out if this doesn't make sense.

> > I think this PR didn't update [this call to `echo_page_table`](https://github.com/DistributedProofreaders/dproofreaders/blob/2e67b263cd13a22aaaff75014ab236061b10904b/project.php#L2056) that would need to include the new JS file, no?
> > I am slightly concerned about the hidden assumption that calling `echo_page_table` now requires to import the script in the caller or else the selection will not work. It would probably be better to have it output a `<script defer src="...">` inside `echo_page_table` but I could be missing something.
> 
> I'm not sure if I have understood your concern correctly but do_page_table() is only called in project.php so the js file will be loaded in the page head by output_html_header() in html_page_common.inc. If I understand it correctly, issue #1298 requires removing all embedded js from php files, not because the js is bad in itself but because it makes it easier to enhance security by eventually disallowing it.

The concern is that the code prior to this PR would automatically register an `onchange` listener for the table generated by `echo_page_table` (on [this line](https://github.com/DistributedProofreaders/dproofreaders/pull/1299/files#diff-e7c7f0e626d9032fb2ad179f2459b03b328ce5d5b98bd07966f2a7c757b24668L510)). The new code puts this logic into a separate JS file that needs to be explicitly loaded. This means that a caller needs to be aware that any calls to `echo_page_table` down the stack requires some special handling.
The suggestion was to let `echo_page_table` automatically generate a deferred `<script>` tag that would pull this needed dependency.

jchaffraix (approved):

cpeel:
> The concern is that the code prior to this PR would automatically register an `onchange` listener for the table generated by `echo_page_table` (on [this line](https://github.com/DistributedProofreaders/dproofreaders/pull/1299/files#diff-e7c7f0e626d9032fb2ad179f2459b03b328ce5d5b98bd07966f2a7c757b24668L510)). The new code puts this logic into a separate JS file that needs to be explicitly loaded. This means that a caller needs to be aware that any calls to `echo_page_table` down the stack requires some special handling. The suggestion was to let `echo_page_table` automatically generate a deferred `<script>` tag that would pull this needed dependency.

Your comment makes sense to me and this two-step process does introduce the possibility of an error. I think in this case its worth it, as having the JS file included as part of our header code allows the introduction of a nonce for these in one place rather than spread around the code. What would probably be ideal is for an `.inc` file to "register" the need for a JS file if it is included such that the JS is automatically pulled in (or just bundle all of our JS together somehow into a single file). I think that's overkill for this situation though.

PR #1300, opened 2024-08-14
Remove script from theme.inc

70ray:
Removes embedded javascript from theme.inc.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/theme_script

### DIFF
--- a/pinc/theme.inc
+++ b/pinc/theme.inc
@@ -632,19 +632,13 @@ function maybe_show_language_selector()
     $options = get_locale_translation_selection_options();
     if (count($options) <= 1) {
         return;
-    } ?>
-    <script><!--
-        function submitLang(i) {
-            top.document.forms.langform.submit();
-        }
-    // --></script>
-    <?php
+    }
 
     $intlang = get_desired_language();
     echo "<form id='langform' action='$code_url/tools/setlangcookie.php' method='POST'>";
     echo "<div>\n";
     echo "<label for='langform-lang'>" . _("Set language") . ": </label>";
-    echo "<select id='langform-lang' name='lang' onChange='submitLang(this)'>\n";
+    echo "<select id='langform-lang' name='lang'>\n";
     echo "<option value=''>" . _("Browser default") . "</option>";
     foreach ($options as $locale => $label) {
         echo "<option value='$locale'".(($locale == $intlang) ? " selected" : "").">$label</option>\n";
--- a/pinc/theme.inc
+++ b/pinc/theme.inc
@@ -144,7 +144,7 @@ abstract class actionIcon
         echo "<br><span class='icon-menu-item-name'>$this->name</span>";
     }
 
-    abstract function draw_action();
+    abstract public function draw_action();
 
     public function draw_extra()
     {
--- a/pinc/theme.inc
+++ b/pinc/theme.inc
@@ -169,7 +169,7 @@ class searchIcon extends actionIcon
 {
     public function draw_action()
     {
-        echo "<div id='site_search' aria-label='$this->name'>";
+        echo "<div tabindex='0' id='site_search' aria-label='$this->name'>";
         $this->draw_icon();
         echo "</div>\n";
     }
--- a/pinc/theme.inc
+++ b/pinc/theme.inc
@@ -244,7 +244,7 @@ function html_logobar()
         echo "<form action='$code_url/tools/site_search.php'>";
         echo "<input id='search-menu-input' type='text' placeholder='$placeholder' name='q' autocapitalize='none'>";
         echo "&nbsp;<a href='$code_url/tools/site_search.php'><i aria-hidden='true' class='fas fa-info-circle fa-lg'></i></a>";
-        echo "&nbsp;<i aria-hidden='true' class='far fa-window-close fa-lg' onclick='hideSiteSearch()'></i>";
+        echo "&nbsp;<i aria-hidden='true' class='far fa-window-close fa-lg' id='hide_search'></i>";
         echo "</form>";
         echo "</div>"; // search-menu
     }
--- a/pinc/theme.inc
+++ b/pinc/theme.inc
@@ -119,7 +119,7 @@ function output_footer($nameofpage, $show_statsbar = true)
     output_html_footer();
 }
 
-class actionIcon
+abstract class actionIcon
 {
     protected string $name;
     protected string $icon;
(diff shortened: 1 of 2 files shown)


scripts/navbar_menu.js
```diff
@@ -46,6 +43,13 @@ function hideSiteSearch() {
 }
 
 window.addEventListener('DOMContentLoaded', function() {
+    const menuSelectors = document.getElementsByClassName('menu-selector');
+    for(const menuSelector of menuSelectors) {
```
cpeel, line 47:
```suggestion
    for (const menuSelector of menuSelectors) {
```

scripts/navbar_menu.js
```diff
@@ -54,11 +58,24 @@ window.addEventListener('DOMContentLoaded', function() {
...
                 event.preventDefault();
                 showSiteSearch();
             }
         });
     }
+
+    // if the page has a language selector, listen for a change
+    let langSelector = document.getElementById('langform-lang');
+
+    if(langSelector) {
```
cpeel, line 75:
```suggestion
    if (langSelector) {
```

pinc/theme.inc
```diff
@@ -119,6 +119,78 @@ function output_footer($nameofpage, $show_statsbar = true)
     output_html_footer();
 }
 
+abstract class actionIcon
```
cpeel, line 122: PHP class names should be PascalCase to match all of our other class names.

pinc/theme.inc
```diff
@@ -119,6 +119,78 @@ function output_footer($nameofpage, $show_statsbar = true)
...
+    }
+
+    public function draw()
+    {
+        echo "<div class='icon-menu-item'>";
+        $this->draw_action();
+        echo "</div>";
+    }
+
+    public function draw_icon()
```
cpeel, line 140:
```suggestion
    protected function draw_icon()
```

pinc/theme.inc
```diff
@@ -119,6 +119,78 @@ function output_footer($nameofpage, $show_statsbar = true)
...
+    {
+        echo "<i aria-hidden='true' class='$this->icon' title='$this->name'></i>";
```
cpeel, line 142: `this->name` needs an `attr_safe()`

pinc/theme.inc
```diff
@@ -119,6 +119,78 @@ function output_footer($nameofpage, $show_statsbar = true)
...
+
+    public function __construct(string $name, string $icon, string $url)
+    {
+        $this->url = $url;
+        parent::__construct($name, $icon);
+    }
+
+    public function draw_action()
+    {
+        echo "<a href='$this->url' aria-label='$this->name'>";
```
cpeel, line 166: Ditto `attr_safe()` for both of these.

pinc/theme.inc
```diff
@@ -119,6 +119,78 @@ function output_footer($nameofpage, $show_statsbar = true)
...
+        $this->draw_icon();
+        echo "</a>";
+    }
+}
+
+class searchIcon extends actionIcon
+{
+    public function draw_action()
+    {
+        echo "<div tabindex='0' id='site_search' aria-label='$this->name'>";
```
cpeel, line 176: `attr_safe()` here too.

pinc/theme.inc
```diff
@@ -119,6 +119,78 @@ function output_footer($nameofpage, $show_statsbar = true)
...
+    public function draw_icon()
+    {
+        echo "<i aria-hidden='true' class='$this->icon' title='$this->name'></i>";
+        $this->draw_extra();
+        echo "<br><span class='icon-menu-item-name'>$this->name</span>";
+    }
+
+    abstract public function draw_action();
+
+    public function draw_extra()
```
cpeel, line 149:
```suggestion
    abstract protected function draw_action();

    protected function draw_extra()
```

pinc/theme.inc
cpeel: Also `MenuIcon` might be a better name since it's in a `icon-menu`-wrapped div.

pinc/theme.inc
```diff
@@ -119,6 +119,78 @@ function output_footer($nameofpage, $show_statsbar = true)
...
+    {
+    }
+}
+
+class linkIcon extends actionIcon
```
cpeel, line 154: `MenuLinkIcon`?

pinc/theme.inc
```diff
@@ -119,6 +119,78 @@ function output_footer($nameofpage, $show_statsbar = true)
...
+
+    public function draw_action()
+    {
+        echo "<a href='$this->url' aria-label='$this->name'>";
+        $this->draw_icon();
+        echo "</a>";
+    }
+}
+
+class searchIcon extends actionIcon
```
cpeel, line 172: `MenuSearchIcon`?

pinc/theme.inc
```diff
@@ -119,6 +119,78 @@ function output_footer($nameofpage, $show_statsbar = true)
...
+{
+    public function draw_action()
+    {
+        echo "<div tabindex='0' id='site_search' aria-label='$this->name'>";
+        $this->draw_icon();
+        echo "</div>\n";
+    }
+}
+
+class inboxIcon extends linkIcon
```
cpeel, line 182: `MenuInboxIcon`?

cpeel (commented):
That's a lot of abstraction but I think it does make things clearer.

scripts/navbar_menu.js
chrismiceli: we have a linter which should enforce some of these suggestions I believe:
https://github.com/DistributedProofreaders/dproofreaders/blob/master/.eslintrc.json
If not, we can definitely codify our desired format.

cpeel: It appears to have [passed the eslint linter](https://github.com/DistributedProofreaders/dproofreaders/actions/runs/10391438980/job/28774349190?pr=1300)? Wonder if something's not configured quite right.

chrismiceli: Let me investigate. We may just not have too many format rules configured.

70ray:
> That's a lot of abstraction but I think it does make things clearer.

I did this because it didn't seem very clear where to put the id in the original but, looking at it again, I think it should be fairly easy. So it would perhaps be better to keep closer to the original for now. I'll make a different PR.

cpeel:
I really don't mind the abstraction you have in place if we can update some of the things I identified in the PR. Whatever you'd like to do.

70ray:
New revised PR opened.

PR #1301, opened 2024-08-15
Js linting

chrismiceli:
Best to review this by commits. They are distinct.

The first commit upgrades packages and node versions, we should probably accept this commit.

The second commits removes the basic format checks our current linter did and replaces them with prettier, an opinionated (although slightly configurable formatter) that is common in the industry. I've done some basic config to minimize the changes, but it does touch quite a few of our js files. This commit is up for discussion if this is something we want.


package.json
```diff
@@ -23,9 +25,18 @@
...
+    "geckodriver": "^4.4.2",
+    "globals": "^15.9.0",
     "less": "^4.2.0",
-    "selenium-webdriver": "^4.12.0"
+    "prettier": "^3.3.3",
+    "selenium-webdriver": "^4.23.0"
+  },
+  "prettier": {
+    "tabWidth": 4,
+    "printWidth":160
```
cpeel, line 40:
```suggestion
    "printWidth": 160
```

cpeel (approved):
Personally I'm rather fond of opinionated formatters (like `black` for python) but I know not everyone feels that way.

I'm great with this change but would like @70ray's input as you two are our most prolific JS devs.

70ray (approved):
I'm fine with this.
Can you also write something in CODE_STYLE.md about prettier?

chrismiceli:
> I'm fine with this. Can you also write something in SETUP/CODE_STYLE.md about prettier?

Great reminder. I've updated that documentation.

package.json
chrismiceli: Good catch.

SETUP/CODE_STYLE.md
```diff
@@ -29,9 +29,24 @@ guide in the wiki.
 
 ## Javascript
 
-We use [eslint](https://eslint.org/) for basic JS code styling and linting.
+We use [prettier](https://prettier.io/) for JS
+code formatting.code styling and linting.
```
srjfoo, line 33: `formatting.code`? Should that be a comma space instead of a period?

chrismiceli: oof, yeah. Let me update that tonight.

cpeel:
@chrismiceli - can you please squash the branch so it has two commits:
1. update node, deps (current first commit)
2. everything with prettier

Then I'll get it merged in.

chrismiceli:
> @chrismiceli - can you please squash the branch so it has two commits:
> 
>     1. update node, deps (current first commit)
> 
>     2. everything with prettier
> 
> 
> Then I'll get it merged in.

Good idea. Rebased off latest master and combined commits.

PR #1303, opened 2024-08-16
remove scripts from theme.inc (simpified version)

70ray:
Sandbox at: https://www.pgdp.org/~rp31/c.branch/alt_theme_script

There are some considerations regarding the icon links: It was necessary to use the url "#" for the search to prevent the page from refreshing and immediately removing the search box. This apparently does not happen if "onclick" is present. This is mentioned in https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#onclick_events. Using an anchor in this way it is possible to right-click and attempt to open the search in a new tab which does not work (same for present version).
As discussed on the mozilla page, ideally a button should be used for this purpose. I tried this earlier but it presents a different appearance which could probably be fixed by styling. I also tried (in the earlier PR) using a 'div'. This does not change the appearance but behaves slightly differently: it is not tabbable (unless it is given a tab_order) and if selected (by tabbing) does not actuate on pressing 'enter' (possibly also fixable by adding an event listener).
It then seemed best for this PR just to remove embedded scripts and leave the rest unchanged.
It would also be possible in future to re-arrange the icon link code to use classes which could make it clearer and make it easier to revise how search option works.

### DIFF
--- a/pinc/theme.inc
+++ b/pinc/theme.inc
@@ -186,16 +186,16 @@ function html_logobar()
         $icon_links[] = [
             "name" => _("Search"),
             "class" => "fas fa-search fa-2x",
-            "onclick" => "showSiteSearch(); return false;",
+            "identifier" => "id='site_search'",
         ];
         $links = [];
         foreach ($icon_links as $data) {
             $name = attr_safe($data["name"]);
             $class = $data["class"];
-            $url = $data["url"] ?? "";
-            $onclick = isset($data["onclick"]) ? "onclick='" . attr_safe($data["onclick"]) . "'" : "";
+            $url = $data["url"] ?? "#";
+            $identifier = $data["identifier"] ?? "";
             $text = "<div class='icon-menu-item'>";
-            $text .= "<a href='$url' aria-label='$name' $onclick>";
+            $text .= "<a href='$url' aria-label='$name' $identifier>";
             $text .= "<i aria-hidden='true' class='$class' title='$name'></i>";
             if ($data["name"] == _("Inbox")) {
                 $numofPMs = get_number_of_unread_messages($user->username);
@@ -203,7 +203,7 @@ function html_logobar()
                     $text .= "<span class='inbox-badge'>$numofPMs</span>";
                 }
             }
-            $text .= "<br><span class='icon-menu-item-name' $onclick>$name</span>";
+            $text .= "<br><span class='icon-menu-item-name'>$name</span>";
             $text .= "</a>";
             $text .= "</div>";
             $links[] = $text;
--- a/pinc/theme.inc
+++ b/pinc/theme.inc
@@ -632,19 +632,13 @@ function maybe_show_language_selector()
     $options = get_locale_translation_selection_options();
     if (count($options) <= 1) {
         return;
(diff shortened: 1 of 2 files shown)


scripts/navbar_menu.js
```diff
@@ -54,11 +58,22 @@ window.addEventListener('DOMContentLoaded', function() {
             }
         });
 
+        document.getElementById('site_search').addEventListener('click', showSiteSearch);
```
jchaffraix, line 61: I think the new behavior triggers the default handler that tries to navigate to the `<a>` tag's `href` attribute. The inline `onclick` handler would prevent this default handling through the use of `return false`.

It seems to me that we could remove the need for setting `href='#'` by changing this to:
```
document.getElementById('site_search').addEventListener('click', function (ev) {
  ev.preventDefault(); // Prevents clicking on the anchor.
  showSiteSearch();
});
```

This would be equivalent to the current code and seems slightly better but could be easily convinced it's better to use the current approach.

pinc/theme.inc
```diff
@@ -431,10 +432,9 @@ function headerbar_menu(string $menu_title, array $links, bool $align_right = fa
         $menu_contents_style = "style='right: 0;'";
     }
 
-    $menu_id = 'menu' . md5($menu_title);
```
jchaffraix, line 434: Nice that you remove a call to MD5 :trophy:

scripts/navbar_menu.js
70ray: > I think the new behavior triggers the default handler that tries to navigate to the `<a>` tag's `href` attribute. The inline `onclick` handler would prevent this default handling through the use of `return false`.
> 
> It seems to me that we could remove the need for setting `href='#'` by changing this to:
> 
> ```
> document.getElementById('site_search').addEventListener('click', function (ev) {
>   ev.preventDefault(); // Prevents clicking on the anchor.
>   showSiteSearch();
> });
> ```
> 
> This would be equivalent to the current code and seems slightly better but could be easily convinced it's better to use the current approach.

I tried this now with `href = ''`. It does work and is closer to the old behaviour in that if the page is scrolled down slightly it does not jump up to the top as it does with `href='#'`.

70ray: > Note, if you remove the `href="#"` then you cannot tab to the link (unless you get funky with a `tabindex` attribute). To get open in new tab feature working, we could give the fragment a name like `href="#search"` or something, and have some `window.hashchange` event logic. Just some ideas.

I'm inclined to leave it like it is. We could open a new PR to address the issue of what happens if you try to open the search in a new tab.

srjfoo: > I'm inclined to leave it like it is. We could open a new PR to address the issue of what happens if you try to open the search in a new tab, or to prevent this option from being available (e.g by using a button instead of an anchor).

Just verifying (though looking at the code I'm pretty sure) that this is the site search, not the project search?

70ray: You are right, the site search is the only one which is not really a link.

cpeel (approved):

srjfoo (approved):

PR #1304, opened 2024-08-16
Type pinc/faq.inc

jchaffraix:
The types were easy to infer from the usage.

As `faq/faq_central.php` was dead, removed it and any associated code (e.g. the function `maybe_redirect_to_external_faq`).

_Update from cpeel:_ Sandbox avail at https://www.pgdp.org/~cpeel/c.branch/julien_type_faq/


pinc/faq.inc
```diff
@@ -97,7 +97,8 @@ $Stealth_Scannos_URL = make_forum_url("t", $Stealth_Scannos_id, 1563);
 $Proofing_Interface_Help_URL = make_forum_url("t", $Proofing_Interface_Help_id, 2739);
 $XHTML_suggestion_URL = make_forum_url("t", $XHTML_suggestion_id, 4405);
 
-function make_forum_url($type, $id, $default)
+/** @param string|int $id */
```
jchaffraix, line 100: We don't have typing for all types of URL we generate below so I sticked to this type for now, though we may be able to narrow this down to `int` down the road :crossed_fingers:

pinc/faq.inc
```diff
@@ -204,14 +206,11 @@ function get_external_faq($faq, $langcode)
     return null;
 }
 
-function maybe_redirect_to_external_faq($langcode = null)
+/** @return never */
+function redirect_to_external_faq(): void
 {
-    $faq = basename($_SERVER['PHP_SELF']);
```
jchaffraix, line 209: This seemed convoluted that we rely on the script name when there is only one caller.

Hard-coding the value of `$faq` makes the logic clearer IMHO, in particular that we always will have a URL.

pinc/faq.inc
```diff
@@ -220,7 +219,5 @@ function maybe_redirect_to_external_faq($langcode = null)
         $url = get_external_faq($faq, "en");
     }
 
-    if ($url) {
-        metarefresh(0, $url, '', '', true);
-    }
+    metarefresh(0, $url, '', '', true);
```
jchaffraix, line 222: I didn't add an `assert($url !== null)` as those could be prod impacting, depending on our config.

pinc/faq.inc
```diff
@@ -35,7 +35,7 @@ $external_faq_overrides = [
         "fr" => "https://www.pgdp.net/wiki/DP_Official_Documentation:CP_and_PM/French/FAQ_fourniture_de_contenu",
         "pt" => "https://www.pgdp.net/wiki/DP_Official_Documentation:CP_and_PM/Portuguese/Fornecimento_de_Conte%C3%BAdos",
     ],
-    "faq_central.php" => [
+    "faq_central" => [
```
cpeel, line 38: This needs to remain `faq_central.php` because that key is used elsewhere (see `pinc/theme.inc`). Also so it's parallel to the other uses.

faq/faq_central.php
```diff
@@ -2,323 +2,5 @@
 $relPath = '../pinc/';
```
cpeel, line 2: We should just delete this file entirely. Not surprisingly, it hasn't shown up in our web logs in a Very Long Time because FAQ / help links for this go directly to the wiki and not here.

pinc/faq.inc
```diff
@@ -204,14 +206,11 @@ function get_external_faq($faq, $langcode)
     return null;
 }
 
-function maybe_redirect_to_external_faq($langcode = null)
```
cpeel, line 207: We can drop this function entirely now.

cpeel (commented):
Some history: we used to have a lot of other files in `faq/` that were recently(ish) deleted. We told other users about this cleanup in our last release in the [release notes](https://github.com/DistributedProofreaders/dproofreaders/releases/tag/R202403):
> This is the last release to include most `faq/*` content. These documents have not been maintained for quite some time as pgdp.net has moved them to their wiki.

At this point pgdp.net has moved entirely to the wiki and all links point directly there and not to `faq_central.php` (and haven't since at least 2021).

pinc/faq.inc
jchaffraix: Good catch, reverted the key change.

pinc/faq.inc
jchaffraix: Removed this function in the updated PR.

SETUP/ci/check_require_login.php
```diff
@@ -26,9 +26,7 @@
     // API
     "api/index.php",
     // FAQs
-    "faq/default.php",
```
jchaffraix, line 29: Removed this as it redirected to the now defunct `faq_central.php`. This means that the `faq/` URL will now return some PHP error.

pinc/faq.inc
```diff
@@ -158,7 +159,8 @@ function get_faq_url($faq, $langcode = null)
     return "";
 }
 
-function get_all_urls_for_faq($faq)
+/** @return array<string, string> */
+function get_all_urls_for_faq(string $faq): array
```
cpeel, line 163: This function appears unused now -- it was only used in `faq_central.php` -- let's delete it.

pinc/faq.inc
```diff
@@ -97,7 +97,8 @@ $Stealth_Scannos_URL = make_forum_url("t", $Stealth_Scannos_id, 1563);
 $Proofing_Interface_Help_URL = make_forum_url("t", $Proofing_Interface_Help_id, 2739);
 $XHTML_suggestion_URL = make_forum_url("t", $XHTML_suggestion_id, 4405);
 
-function make_forum_url($type, $id, $default)
+/** @param string|int $id */
+function make_forum_url(string $type, $id, string $default): string
```
cpeel, line 101: After doing some inspection of the code, let's just make this:
```suggestion
function make_forum_url(string $type, int $id, string $default): string
```
That's accurate with how the function is used. I'll do a fast-follow PR that will do a lot of cleanup of the variables above which we no longer need because they are not used.

pinc/faq.inc
cpeel: I pushed up a commit to delete this.

cpeel:
I updated the sandbox on TEST.

cpeel (approved):
Love how much red is in this PR 😍

70ray (approved):

srjfoo (approved):

PR #1305, opened 2024-08-18
Resolve duplicate -- and different -- `make_link()` functions

cpeel:
We have 3 `make_link()` functions in the codebase, one each in:
* `ProjectSearchResults.inc`
* `manage_special_days.php`
* `manage_image_sources.php`

For the last two scripts, one of them was very likely created as a copy of the other in the distant past.

Recently type definitions were added for the one in `manage_special_days.php` and for some users phpstan is getting confused with having duplicate names that accept different types. Even though these functions are never defined concurrently in a single page scope, phpstan doesn't know this and having duplicate function names (without namespaces) is a bad practice anyway.

This:
1. removes the two functions in the `manage_*.php` functions as they really aren't needed.
2. renames some functions in `ProjectSearchResults.inc` to be less generic since they are specific to search results.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/remove_dup_make_links/

### DIFF
--- a/pinc/ProjectSearchResults.inc
+++ b/pinc/ProjectSearchResults.inc
@@ -117,11 +117,11 @@ class Column
                     $switch_sort = ['sec_sort' => $sort_column_id];
                 } else {
                     $sec_sort_array = ['sec_sort' => $this->id, 'results_offset' => 0];
-                    $sec_sorter = "&nbsp;" . make_link($sec_sort_array, "^", "#head");
+                    $sec_sorter = "&nbsp;" . make_search_page_link($sec_sort_array, "^", "#head");
                 }
             }
             $pri_sort_array = array_merge(['sort' => "$this->id$dir_code", 'results_offset' => 0], $switch_sort);
-            $label = make_link($pri_sort_array, $label, "#head") . "$marker$sec_sorter";
+            $label = make_search_page_link($pri_sort_array, $label, "#head") . "$marker$sec_sorter";
         }
         echo "<th class='$this->css_class' title='", attr_safe($this->get_tooltip()), "'>$label</th>\n";
     }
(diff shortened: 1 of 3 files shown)


srjfoo (approved):

pinc/ProjectSearchResults.inc
```diff
@@ -717,23 +717,23 @@ function get_search_configure_link()
...
 }
 
-function make_url($new_array, $anchor = '')
+function make_search_page_url(array $new_array, string $anchor = ''): string
 {
     return "?" . http_build_query(array_replace($_GET, $new_array)) . "$anchor";
 }
 
-function make_link($new_array, $text, $anchor = '')
+function make_search_page_link(array $new_array, string $text, string $anchor = ''): string
```
jchaffraix, line 725: Ditto about narrowing the type of `$new_array`:

```suggestion
/** @param array<string, int> $new_array */
function make_search_page_link(array $new_array, string $text, string $anchor = ''): string
```

pinc/ProjectSearchResults.inc
```diff
@@ -717,23 +717,23 @@ function get_search_configure_link()
     return "<a href='?show=config&amp;origin=$origin'>" . _("Configure Result") . "</a>";
 }
 
-function make_url($new_array, $anchor = '')
+function make_search_page_url(array $new_array, string $anchor = ''): string
```
jchaffraix, line 720: I *think* we could narrow the type of `$new_array` here:

```suggestion
/** @param array<string, mixed> $new_array */
function make_search_page_url(array $new_array, string $anchor = ''): string
```

jchaffraix (approved):
Thanks for fixing those inconsistencies!

chrismiceli (approved):

pinc/ProjectSearchResults.inc
cpeel: They both need to be `/** @param array<string, mixed> $new_array */` -- I've added a new commit.

PR #1306, opened 2024-08-18
Issue #1274 convert exec to Symfony Process

mrducky4:
cleaning up file archiving.inc, use Process instead of exec. Add DocBlock function header comments.
Add type declarations for functions.
Remove unused functions.

I did not test this change in my developer VM. cpeel suggested testing it in a TEST sandbox.

### DIFF
--- a/pinc/archiving.inc
+++ b/pinc/archiving.inc
@@ -17,7 +17,7 @@ include($relPath.'udb_user.php'); // $archive_db_name
  *   write out project's metadata to a JSON file.
  *   (for later off-site migration).
  * - Mark the project as having been archived.
- * 
+ *
  * @param Project $project
  *   the Project to archive
  * @param bool $dry_run
--- a/pinc/archiving.inc
+++ b/pinc/archiving.inc
@@ -75,11 +75,14 @@ function archive_project(Project $project, bool $dry_run = false)
             echo "    Move $project_dir to $new_dir.\n";
         } else {
             // Remove uncompressed versions of whole-project texts, leaving zips.
-            $process = new Process(["rm", "$project_dir/projectID*.txt"]);
-            $process->run();
-            if (!$process->isSuccessful()) {
-                throw new RuntimeException(sprintf(_("Unable to delete %s"), "$project_dir/projectID*.txt"));
+            if (glob("$project_dir/projectID*.txt")) {
+                $process = new Process(["rm", "$project_dir/projectID*.txt"]);
+                $process->run();
+                if (!$process->isSuccessful()) {
+                    throw new RuntimeException(sprintf(_("Unable to delete %s"), "$project_dir/projectID*.txt"));
+                }
             }
+
             if (!rename($project_dir, $new_dir)) {
                 throw new RuntimeException(sprintf(_('Unable to move %1$s to %2$s', $project_dir, $new_dir)));
             }
--- a/pinc/archiving.inc
+++ b/pinc/archiving.inc
@@ -2,6 +2,8 @@
 include_once($relPath.'Stopwatch.inc');
 include_once($relPath.'Project.inc'); // does_project_page_table_exist()
 
+use Symfony\Component\Process\Process;
+
 // include udb_user.php to place $archive_db_name in this file's scope.
 global $archive_projects_dir, $archive_db_name;
 include($relPath.'udb_user.php'); // $archive_db_name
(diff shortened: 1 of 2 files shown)


mrducky4:
I don't know how to interpret the linting check failure. I ran `make lint` in my developer VM and there were no errors.

pinc/archiving.inc
```diff
@@ -15,13 +17,18 @@ include($relPath.'udb_user.php'); // $archive_db_name
  *   write out project's metadata to a JSON file.
  *   (for later off-site migration).
  * - Mark the project as having been archived.
+ *
```
cpeel, line 20: The linter is catching this errant space I believe.
```suggestion
 *
```

pinc/archiving.inc
```diff
@@ -223,30 +264,3 @@ $archival_recovery_multiplier = 1;
 // If you're running the script when the site is down (e.g., when upgrading),
 // you probably don't need to insert sleeps, so you can eliminate them
 // by setting this variable to zero.
-
-// -----------------------------------------------------------------------------
-
-function archival_skip_logs_init()
```
cpeel, line 229: Thanks for catching and deleting these!

mrducky4:
I removed that errant space, now all 12 checks pass. Should I have seen this error in a VM test before submitting a PR?

cpeel:
> I removed that errant space, now all 12 checks pass. Should I have seen this error in a VM test before submitting a PR?

Probably if you ran `make lint` or `make lint_code` (or just `make` which I think runs most everything except the tests and shellcheck which isn't on the VM).

pinc/archiving.inc
mrducky4: It's easier to delete than to add DocBlock comments :)

mrducky4:
I did run `make lint`, it does not complain about the extra space.

pinc/archiving.inc
```diff
@@ -64,9 +75,13 @@ function archive_project($project, $dry_run = false)
             echo "    Move $project_dir to $new_dir.\n";
         } else {
             // Remove uncompressed versions of whole-project texts, leaving zips.
-            exec("rm $project_dir/projectID*.txt");
+            $process = new Process(["rm", "$project_dir/projectID*.txt"]);
+            $process->run();
+            if (!$process->isSuccessful()) {
+                throw new RuntimeException(sprintf(_("Unable to delete %s"), "$project_dir/projectID*.txt"));
+            }
```
cpeel, line 82: This change is slightly problematic because there might not be files matching that format to delete which throws an error. We need to add a `glob()` to see if there are any files before trying to delete them.

```suggestion
            if (glob("$project_dir/projectID*.txt")) {
                $process = new Process(["rm", "$project_dir/projectID*.txt"]);
                $process->run();
                if (!$process->isSuccessful()) {
                    throw new RuntimeException(sprintf(_("Unable to delete %s"), "$project_dir/projectID*.txt"));
                }
            }

```

cpeel (commented):
Tested this on TEST and except for the one suggestion it worked great.

cpeel:
> I did run `make lint`, it does not complain about the extra space.

I was wrong -- php-cs-fixer isn't included as part of our makefile checks. I'll get that fixed just after Chris gets the updated prettier formatter in for JS.

mrducky4:
Sorry for the delay. I've been camping with limited Internet. I will get back to this.

cpeel (approved):
This looks & works great for me!

cpeel:
I found two small bugs when helping @srjfoo test this and pushed up a commit to fix them:
1. `Process` will escape the `*` in the `rm` and so it won't actually find the files we think it will, so change that to an iterator over what glob returns, which means we can just `unlink()` it rather than use `Process` and `rm`.
2. `archive_ancillary_data_for_project_etc()` takes a Project as the first art so load and pass it a real Project instead of the stdClass stub.

srjfoo (approved):

cpeel:
Pushed up a commit to try and resolve the merge conflict.

mrducky4:
Thanks for fixing this. I should have made a tiny test script to check it myself.

cpeel:
> Thanks for fixing this. I should have made a tiny test script to check it myself.

Not a problem -- the inability for you to readily test this is a DevEx failure and I'm glad I could help. Appreciate you working on this!

PR #1309, opened 2024-08-19
use old german dictionary if possible

70ray:
This addresses [task 573](https://www.pgdp.net/c/tasks.php?action=show&task_id=573)
This is a hack for this special case. Looking at the list of .multi files in /usr/lib/aspell there are many variants which could be useful. It would require more extensive code changes to allow for more general selection of options.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/lang_var
It can be tested by using some old and new german spellings with wordcheck. e.g. 'daß dass'.

### DIFF
--- a/pinc/wordcheck_engine.inc
+++ b/pinc/wordcheck_engine.inc
@@ -295,6 +295,8 @@ class BadWordAccumulator
 
 function get_dictionary_for_langcode(string $langcode): ?string
 {
+    global $aspell_prefix;
+
     $dict_file = "$aspell_prefix/lib/aspell/$langcode.multi";
     return is_file($dict_file) ? $dict_file : null;
 }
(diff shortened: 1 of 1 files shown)


cpeel:
I appreciate you working on this, but using the alt instead of the main dictionary for German is already a solved problem -- we just need to change a symlink on the server. The question is if we _should_ do this.

There was some initial discussions in 2018 in the [Team Germany](https://www.pgdp.net/phpBB3/viewtopic.php?p=1136516#p1136516) thread which seemed to be yes but there was squirrel discussion that it's not a good generalized solution since some books would be better by the "modern" German dictionary.

Then there was a follow-up convo between @wf49670 and I about this and it would be best if there was a way for PMs to select which dictionary variant they would want for a project but that was a rather involved change (and I still think it is). That's why nothing has been done on this ticket.

If there's been renewed discussion/consensus on this global change please point @wf49670 and I to it and we can see about that symlink change.

PR #1311, opened 2024-08-23
Allow users to bookmark projects

cpeel:
Allow users to bookmark arbitrary projects and see them on the My Projects page. [Task 2084](https://www.pgdp.net/c/tasks.php?action=show&task_id=2084)

Sandbox: https://www.pgdp.org/~cpeel/c.branch/project-bookmarks/

### DIFF
--- a/project.php
+++ b/project.php
@@ -77,7 +77,7 @@ $extra_args = [
 
 output_header($title_for_theme, NO_STATSBAR, $extra_args);
 
-echo "<h1>" . get_bookmark_form() . html_safe($title)  . "</h1>\n";
+echo "<h1>" . get_bookmark_form() . html_safe($title) . "</h1>\n";
 
 maybe_output_new_proofer_project_message($project);
 
--- a/project.php
+++ b/project.php
@@ -36,6 +36,7 @@ $projectid = get_projectID_param($_GET, 'id');
 $expected_state = get_enumerated_param($_GET, 'expected_state', null, ProjectStates::get_states(), true);
 $detail_levels = get_project_detail_levels();
 $detail_level = get_enumerated_param($_GET, 'detail_level', $userSettings->get_value('project_detail', 2), array_keys($detail_levels));
+$mark_bookmark = get_bool_param($_POST, 'bookmark', null, true);
 
 $project = new Project($projectid);
 
@@ -46,7 +47,11 @@ $project = new Project($projectid);
 // project) first.
 $title_for_theme = sprintf(_('"%s" project page'), $project->nameofwork);
 
-$title = sprintf(_("Project Page for '%s'"), $project->nameofwork);
+$title = $project->nameofwork;
+
+if ($user_is_logged_in && $mark_bookmark !== null) {
+    upi_set_bookmark($pguser, $project->projectid, $mark_bookmark);
+}
 
 // -----------------------------------------------------------------------------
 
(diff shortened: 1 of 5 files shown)


project.php
```diff
@@ -72,7 +77,7 @@
 
 output_header($title_for_theme, NO_STATSBAR, $extra_args);
 
-echo "<h1>" . html_safe($title) . "</h1>\n";
+echo "<h1>" . get_bookmark_form() . html_safe($title)  . "</h1>\n";
```
srjfoo, line 80: Extra space after `($title)`?

srjfoo (commented):
I really like the feature, but have a couple of minor questions.

Deleted projects can be bookmarked, but don't display in the "Bookmarks" list. I don't see a whole lot of percentage in displaying them, but if not, should the bookmark icon be removed from deleted projects?

Should project clones retain the bookmark (they don't currently)? I'm unsure about this. On one hand someone who's bookmarked a project might want to know if an MP project was run for it, on the other hand, they don't get notifications of anything that happens to a bookmarked project, and most MP projects are short enough that they'd be gone (i.e. deleted) before the user knew they'd existed.

chrismiceli (commented):
Technically this looks great. I am usually cautious about implementing bookmarks since the browser has bookmarks, but I'll defer on those matters to the community.

windymilla (approved):

cpeel:
> Technically this looks great. I am usually cautious about implementing bookmarks since the browser has bookmarks, but I'll defer on those matters to the community.

That's a valid point. This is probably better named something like "custom list" but this seemed clearer at the time. The value I think this has over a browser-based bookmark is that it lets you see status information about the bookmarked projects together instead of needing to go to each one. Well, that and cross-browser/cross-device functionality but browsers have other ways to handle that these days.

project.php
cpeel: Fixed, thanks!

cpeel (commented):
> Deleted projects can be bookmarked, but don't display in the "Bookmarks" list.

This was an oversight and I've fixed it. IMHO we should display all projects that someone has bookmarked and not do any special filtering on them. I also don't think we should change a user's bookmarks in any way (adding them to new projects, removing them from old ones). I view it as an entirely manual list they control.

Clones are separate projects and we don't do anything special on their creation -- we don't clone notifications, etc -- so we shouldn't treat them differently for bookmarks either.

70ray (approved):

srjfoo:
> > Deleted projects can be bookmarked, but don't display in the "Bookmarks" list.
> 
> This was an oversight and I've fixed it. IMHO we should display all projects that someone has bookmarked and not do any special filtering on them. I also don't think we should change a user's bookmarks in any way (adding them to new projects, removing them from old ones). I view it as an entirely manual list they control.

Works for me. 

> Clones are separate projects and we don't do anything special on their creation -- we don't clone notifications, etc -- so we shouldn't treat them differently for bookmarks either.

I'm also okay with this. I was thinking that there are times it would be good to clone notifications (from a squirrel POV), but then remembered that `copy_pages.php` handles that.

> > Technically this looks great. I am usually cautious about implementing bookmarks since the browser has bookmarks, but I'll defer on those matters to the community.

>The value I think this has over a browser-based bookmark is that it lets you see status information about the bookmarked projects together instead of needing to go to each one. ...

I agree. I think, too, that browser bookmarks are a bit of a black hole. I have a habit of losing track of bookmarks, and I'm sure that I have tons of them that are totally irrelevant now.

srjfoo (approved):

PR #1319, opened 2024-09-03
Some style cleanups

srjfoo:
Found when investigating what it would take to do a dark themed version of the proofreading interfaces for the Charcoal theme:

* Update PM WC management colors for charcoal
* Make tdbottom background color same as tdtop for Enhanced interface.

Sandbox: [misc-style-fixes](https://www.pgdp.org/~srjfoo/c.branch/misc-style-fixes/)

### DIFF
--- a/styles/page_interfaces.less
+++ b/styles/page_interfaces.less
@@ -2,10 +2,13 @@
 /* Page Interface Structures */
 
 /*
- * The following classes all relate to various page interfaces, such as the
- * proofreading interfaces (standard and enhanced), WordCheck, the proofreading
- * toolbox, code point validator, display image interface, etc.
+ * The following variables and classes all relate to various page
+ * interfaces, such as the proofreading interfaces (standard and
+ * enhanced), WordCheck, the proofreading toolbox, code point
+ * validator, display image interface, etc.
  */
+
+/* variables */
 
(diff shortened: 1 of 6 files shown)


chrismiceli (approved):

styles/page_interfaces.less
```diff
@@ -25,6 +28,16 @@
 @page-interface-visited-color: #990099;
 @page-interface-active-color: #ff0000;
 
+/* WordCheck*/
+@preWC: #c2c2c2;
+@noWC: #ffb321;
+@WC: #97fc93;
```
cpeel, line 34: Was it intentional to change this from `#97fc9e`?  I don't have strong feelings about this subtle change but wanted to confirm you mean to make it.

srjfoo: > Was it intentional to change this from `#97fc9e`? I don't have strong feelings about this subtle change but wanted to confirm you mean to make it.

Um, no, not intentional at all. Me fat-fingering it, I suppose (the 3 is above the e on the keyboard, and my finger gets them mixed up more often than I'd like 😁). I'll get that changed back.

cpeel (approved):

70ray (approved):

PR #1321, opened 2024-09-08
Type most *Exception::_construct

jchaffraix:
The arguments from the derived classes are passed
to the base constructor and thus must match the
definition from PHP:

https://www.php.net/manual/en/exception.construct.php

### DIFF
--- a/api/exceptions.inc
+++ b/api/exceptions.inc
@@ -7,7 +7,7 @@
  */
 class ApiException extends Exception
 {
-    public function __construct($message = "API exception", $code = 1)
+    public function __construct(string $message = "API exception", int $code = 1)
     {
         parent::__construct($message, $code);
     }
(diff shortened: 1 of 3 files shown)


api/exceptions.inc
```diff
@@ -100,15 +100,15 @@ class ForbiddenError extends ApiException
 
 class UnexpectedError extends ApiException
 {
-    public function __construct($message = "Unexpected error", $code = 7)
+    public function __construct(string $message = "Unexpected error", int $code = 7)
```
jchaffraix, line 103: I didn't fix this class using the same `$code` as the `ForbiddenError` (nor the general fact that `$code` is rarely updated and could be a class constant).

chrismiceli (approved):

cpeel (approved):

srjfoo (approved):

api/exceptions.inc
srjfoo: Is this something that should be fixed, or does it not matter?

jchaffraix: Both issues are very minor and I don't believe they cause any issue.

Reusing `$code` would be a bother if we did have some dashboard for the exception thrown in prod/test grouped by `$code` as we would misclassify some of them.

Not having `$code` be a constant is more of a design/code quality point so pretty low on the totem pole.

70ray (approved):

PR #1322, opened 2024-09-09
Documentation updates for upcoming R202409 release

cpeel:
This updates the usual doc suspects for our upcoming R202409 release. As a reminder we try to highlight things in the changelog that we think most site admins would be interested in`*` and tell them to look in the git history for the full details. The order is roughly what I think is the most important first which is very subjective.

Feedback on what was included, the order of the things listed, and any other thoughts are appreciated!

`*` I also ensure that every developer who contributed to the release has an item in the `CHANGELOG.md` file too.

### DIFF
--- a/SETUP/CHANGELOG.md
+++ b/SETUP/CHANGELOG.md
@@ -7,13 +7,30 @@ see the git history.
 [R202009](https://github.com/DistributedProofreaders/dproofreaders/releases/tag/R202009)
 first before upgrading to R202102 or later releases.**
 
-## R??????
+## R202409
 Scripts supporting this upgrade are in `SETUP/upgrade/21`
 
-**This is the last release to support PHP 7.4. Future releases will only
-support PHP 8.0 and later.**
+### Notices & Deprecations
+
+This is the last release to support PHP 7.4. Future releases will only
+support PHP 8.1 and later.
+
+This is the last release to support the Hieroglyphs tool. WikiHiero which backs
+the tool has been unsupported for 20+ years and was written for PHP 4.3.3.
+
+### Changes
 
 * Updated minimum middleware to MySQL 8.0 (cpeel)
+* The deprecated globals mentioned in the last release notes were removed (cpeel)
+* Most `faq/*` content has been deleted (cpeel)
+* A `metadata.json` file has replaced `dc.xml` in the project directory (cpeel)
+* New API endpoints for proofreading (70ray)
+* Cronjobs now run through common interface with SA UI to show job status (cpeel)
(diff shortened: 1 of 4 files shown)


SETUP/CHANGELOG.md
```diff
@@ -7,13 +7,23 @@ see the git history.
 [R202009](https://github.com/DistributedProofreaders/dproofreaders/releases/tag/R202009)
 first before upgrading to R202102 or later releases.**
 
-## R??????
+## R202409
 Scripts supporting this upgrade are in `SETUP/upgrade/21`
 
 **This is the last release to support PHP 7.4. Future releases will only
-support PHP 8.0 and later.**
+support PHP 8.1 and later.**
```
cpeel, line 14: Rationale behind this:
* PHP 8.0 is already out of support.
* PHP 8.1 is in Ubuntu 22.04 which is still supported. It also includes some useful [type features](https://www.php.net/manual/en/language.types.declarations.php). We tested with PHP 8.1 on the clone system earlier this year and know it works.
* PHP 8.3 is what will be running on PROD after the upgrade later this year.

chrismiceli (approved):

jchaffraix (approved):

SETUP/CHANGELOG.md
```diff
@@ -7,13 +7,23 @@ see the git history.
...
+* The deprecated globals mentioned in the last release notes were removed (cpeel)
+* Most `faq/*` content has been deleted (cpeel)
+* A `metadata.json` file has replaced `dc.xml` in the project directory (cpeel)
+* New API endpoints for proofreading (70ray)
+* Cronjobs now run through common interface with SA UI to show job status (cpeel)
+* Typing added to numerous functions and classes (bpfoley, jchaffraix)
+* XML feed types removed and feed backend improved (mrducky4)
+* Navigation bar updates for smaller screens (cpeel)
+* New site search page (cpeel)
+* `past_tallies` is now a sparse table (cpeel)
```
srjfoo, line 26: Should the new mentor page be included?

windymilla (approved):

SETUP/CHANGELOG.md
cpeel: I don't think so, it's a pretty minor change for a pretty small subset of users.

srjfoo (approved):

srjfoo:
Reminder to add item about removing the wikihiero code.

cpeel:
> Reminder to add item about removing the wikihiero code.

Added a commit about this being the last release to support WikiHiero.

srjfoo (approved):

PR #1324, opened 2024-09-09
Type pinc/CharSuites.inc

jchaffraix:
This exposed that CharSuite::description was not initialized in the constructor and was never set for several charsuite, Basic Cyrillic among them.

TESTS=Enabled/disabled a charsuite for the SA,
made sure all charsuites could be displayed in SA, added/removed a charsuite on a project.

_Update from cpeel_: sandbox at https://www.pgdp.org/~cpeel/c.branch/julien_type_CharSuite/

### DIFF
--- a/pinc/CharSuites.inc
+++ b/pinc/CharSuites.inc
@@ -7,11 +7,13 @@ class CharSuiteNotEnabledException extends Exception
 
 class PickerSet
 {
-    public $name;
+    public string $name;
     private array $subsets;
+    /** @var string[] */
     private array $titles;
 
-    public function add_subset($name, $codepoints, $title = null)
+    /** @param $codepoints string[] */
+    public function add_subset(string $name, array $codepoints, ?string $title = null)
     {
         $this->subsets[$name] = $codepoints;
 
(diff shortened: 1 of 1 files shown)


chrismiceli (approved):

cpeel (approved):

chrismiceli (approved):

srjfoo (approved):
> ... This exposed that CharSuite::description was not initialized in the constructor and was never set for Basic Cyrillic.

Unless I'm missing something, it was also never set for Basic Greek, Basic Latin and Polytonic Greek. (At least on my review of the code.) The difference between them and Basic Cyrillic is that they were added with PR #302, even before rollout, and Charsuite::description wasn't added until later, with PR #379 (so they couldn't have had a description added at the time they were created).

jchaffraix:
> > ... This exposed that CharSuite::description was not initialized in the constructor and was never set for Basic Cyrillic.
> 
> Unless I'm missing something, it was also never set for Basic Greek, Basic Latin and Polytonic Greek. (At least on my review of the code.) The difference between them and Basic Cyrillic is that they were added with PR #302, even before rollout, and Charsuite::description wasn't added until later, with PR #379 (so they couldn't have had a description added at the time they were created).

You are 100% correct that Basic Greek, Basic Latin and Polytonic Greek are also missing the description. I detected the issue on Basic Cyrillic and thought this was the only charset following this pattern (though I tested rendering all the charsuite to be safe). Broadened the PR description so it's less confusing.

70ray (approved):

PR #1326, opened 2024-09-14
Add missing types in Settings.inc

jchaffraix:
The file has a lot of PHPStan type annotations that were left as is (ie still not enforced by the PHP runtime).

For the new types, I have made them enforced.

TEST=Manipulated the settings page for different profile.

_Update by cpeel_: Sandbox at https://www.pgdp.org/~cpeel/c.branch/julien_add_missing_type_Settings/

### DIFF
--- a/pinc/Settings.inc
+++ b/pinc/Settings.inc
@@ -10,12 +10,13 @@ class Settings
     // Multi-dimensional array containing all user settings. Each setting
     // may have one or more value associated with it. Format is:
     //     $settings_array[$setting] = array($value1, ...);
-    private $settings_array = [];
+    /** @var array<string, mixed[]> */
+    private array $settings_array = [];
 
     // username of subject user.
-    private $username = null;
+    private ?string $username = null;
 
-    public function __construct($name)
+    public function __construct(?string $name)
     {
         if ($name == '') {
             // Note that that's a "loose equals", so $name could be:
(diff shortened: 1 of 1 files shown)


pinc/Settings.inc
```diff
@@ -10,12 +10,13 @@ class Settings
...
-    private $settings_array = [];
+    /** @var array<string, mixed[]> */
+    private array $settings_array = [];
 
     // username of subject user.
-    private $username = null;
+    private ?string $username = null;
 
-    public function __construct($name)
+    public function __construct(?string $name)
```
cpeel, line 19: The comment below this says that `$name` could be an integer or a boolean too (which honestly I think is very unlikely). If we feel confident this type annotation we should update the comment below.

pinc/Settings.inc
```diff
@@ -107,17 +108,22 @@ class Settings
...
 
-    private function _clear_setting($settingCode)
+    private function _clear_setting(string $settingCode): void
     {
-        $sql = "
+        $sql = sprintf(
+            "
             DELETE FROM usersettings 
-            WHERE username = '$this->username'
-                AND setting = '$settingCode'
```
cpeel, line 115: Thank you for this change!

chrismiceli (approved):

pinc/Settings.inc
jchaffraix: Good catch, I also feel fairly confident about this type so amended the comment below.

cpeel (approved):

srjfoo (approved):

70ray (approved):

PR #1327, opened 2024-09-14
Correctly load UTF-8 system locales

cpeel:
The list of system locales is only used in the translation center for managing translations. The list being returned from `get_installed_system_locales()` incorrectly included the `.utf8` extension. Sharon found this during the PHP 8.3 testing.

This fixes that bug and also changes the `exec()` to a `Process`. Because this bug prevents creating new translations I'd like to get this merged before the upcoming release -- even if we don't get it rolled out to PROD until next month.

https://www.pgdp.org/~cpeel/c.branch/fix-translation-center/

### DIFF
--- a/SETUP/ci/check_security.php
+++ b/SETUP/ci/check_security.php
@@ -13,7 +13,6 @@ $basedir = $argv[1] ?? "../../";
 //       others are added from the current set until they get updated.
 $ok_system_calls = [
     "pinc/forum_interface_phpbb3.inc",
-    "pinc/languages.inc",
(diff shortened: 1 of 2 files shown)


srjfoo (approved):

chrismiceli (approved):

70ray (approved):

cpeel:
Updated commit to remove `languages.inc` from `SETUP/ci/check_security.php` since it no longer uses `exec()`. Will merge this in after the CI successfully passes.

FYI @mrducky4 since you're working in a similar area.

PR #1329, opened 2024-09-15
Fix code_style error in configuration.sh which flows into site_vars.php

mrducky4:
Fix spaces that php-cs-fixer complains about, when incorporated into site_vars.php.

The error appears if you use configuration.sh as the input to SETUP/configure,
which creates site_vars.php, and then you run `make -C SETUP/ci code_style`


srjfoo (approved):

cpeel (approved):
Thanks for this fix!

PR #1330, opened 2024-09-15
Make ImageSource class & DB match

cpeel:
86ab2156db set the types for the ImageSource class but some of the strings can be null in the DB. Reconcile the two. While we're here make the url and credit fields the max 255 varchar size (like the URLs are in other tables).

That commit was added right after the R202403 release so I'd like to get this one in before the upcoming one. Turns out there is a single row in PROD with a `null` value in it and it's a wonder we didn't catch this prior in the past 6 months.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/fix-manage-image-sources-types/

### DIFF
--- a/tools/project_manager/manage_image_sources.php
+++ b/tools/project_manager/manage_image_sources.php
@@ -152,7 +152,7 @@ class ImageSource
     public int $ok_keep_images; // TODO: bool
     public int $ok_show_images; // TODO: bool
     public string $public_comment;
-    public string $internal_comment;
+    public ?string $internal_comment;
     public int $is_active;
 
     public function __construct($code_name = null)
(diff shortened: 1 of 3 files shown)


tools/project_manager/manage_image_sources.php
```diff
@@ -152,7 +152,7 @@ class ImageSource
     public int $ok_keep_images; // TODO: bool
     public int $ok_show_images; // TODO: bool
     public string $public_comment;
-    public string $internal_comment;
+    public ?string $internal_comment;
```
cpeel, line 155: `text` fields can't have default values so this ensures `null` is a valid entry from the DB, despite this form always setting it to `''` on creation.

srjfoo (approved):

chrismiceli (approved):

jchaffraix (approved):

70ray (approved):

PR #1331, opened 2024-09-16
Type pinc/unicode.pinc

jchaffraix:
The collection of functions are leaves so it
should help type its callers.

While touching it, changed the return type for
guess_string_encoding to ?string instead of using
the existing return (string|false).

TEST=PM: created project + upload, some P1 review, SA charsuite display.

_Update by cpeel_: sandbox at https://www.pgdp.org/~cpeel/c.branch/julien_type_unicode/

### DIFF
--- a/pinc/unicode.inc
+++ b/pinc/unicode.inc
@@ -373,9 +373,10 @@ function convert_text_file_to_utf8(string $filename)
  * * Windows-1252
  * * ISO-8859-1
  *
- * These strings works similarly to mb_detect_encoding() except that it returns
- * null rather than false. The function returns null if it's unable to guess,
- * although it will readily return ISO-8859-1 in many circumstances.
+ * This function works similarly to mb_detect_encoding() -- returning the same
+ * encoding strings, except that it returns null rather than false.
+ * The function returns null if it's unable to guess, although it will readily
+ * return ISO-8859-1 in many circumstances.
  */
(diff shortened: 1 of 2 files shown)


pinc/unicode.inc
```diff
@@ -103,8 +107,11 @@ function utf8_character_name($characters)
  * Split a string into chunks on Unicode script boundaries.
  *
  * Inherited characters are included in the chunk with the base character
+ *
+ * @param string $string
```
jchaffraix, line 111: I could have enforced this type (and several others in this file) but thought it would be better to lower the risk of this PR to just add PHPStan type annotation.

cpeel: For this one we should enforce it in the function def.

jchaffraix: Fair, added it to the function def. As an aside, `UTF8::str_split` accepts `int|string`, which is why I allowed the ambiguity for now.

pinc/unicode.inc
```diff
@@ -359,10 +369,10 @@ function convert_text_file_to_utf8($filename)
  * * ISO-8859-1
  *
  * These strings match what mb_detect_encoding() would return. The function
```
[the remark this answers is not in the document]
jchaffraix: Updated the comment.

chrismiceli (approved):

srjfoo (approved):

PR #1332, opened 2024-09-17
Fix two PHP 8.3 deprecation warnings

cpeel:
Somewhat shockingly, these are the only two outstanding issues that were found in the PHP 8.3 testing. These address two deprecation warnings.

### DIFF
--- a/tools/upload_text.php
+++ b/tools/upload_text.php
@@ -22,7 +22,7 @@ $stage = get_enumerated_param($_REQUEST, 'stage', null, $valid_stages, true);
 // days not given (defaults to 0): replace file only
 $days = get_integer_param($_REQUEST, 'days', 0, 0, 56);
 $action = @$_REQUEST['action'];
-$postcomments = @$_POST['postcomments'];
+$postcomments = array_get($_POST, 'postcomments', "");
 
 $project = new Project($projectid);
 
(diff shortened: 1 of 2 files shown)


srjfoo (approved):

chrismiceli (approved):

tools/upload_text.php
```diff
@@ -22,7 +22,7 @@
 // days not given (defaults to 0): replace file only
 $days = get_integer_param($_REQUEST, 'days', 0, 0, 56);
 $action = @$_REQUEST['action'];
-$postcomments = @$_POST['postcomments'];
+$postcomments = array_get($_POST, 'postcomments', "");
```
jchaffraix, line 25: Side question: do we have any recommendations around using `array_get` vs the null coalescing operator (`$_POST['postcomment'] ?? null`)?

jchaffraix (approved):

70ray (approved):

tools/upload_text.php
cpeel: We don't. The rule of thumb I've been using is that if we're getting user attributes (like this) using `array_get()` alongside other `get_*()` functions seems most intuitive. Elsewhere I've used and see both. I don't have strong feelings about this one way or the other.

We've been using `array_get()` before the null coalsecer operator was available to us so we'll have a mix of that for quite a while I expect.

jchaffraix: Thanks for the context!

As we add more type, I think the null coalescing operator will end up being favored: we can't easily type `array_get` without splitting it by the expected return type, which seems heavy.

PR #1333, opened 2024-09-19
Type tools/project_manager/remote_file_manager.php

jchaffraix:
TEST=Interacted with the remote file manager (uploaded/deleted files, check directories)

_Edit by srjfoo_: Sandbox at [~srjfoo/c.branch/julien_type_tools_remote_file_manager/](https://www.pgdp.org/~srjfoo/c.branch/julien_type_tools_remote_file_manager/)

### DIFF
--- a/tools/project_manager/remote_file_manager.php
+++ b/tools/project_manager/remote_file_manager.php
@@ -204,14 +204,13 @@ switch ($action) {
     default:
         // no matching $action in input
         fatal_error(sprintf(_("Invalid action: '%s'"), html_safe($action)));
-        exit;
 }
 
 do_showdir($action_message);
 
 //---------------------------------------------------------------------------
 
-function do_showdir($action_message)
+function do_showdir(string $action_message): void
 {
     global $curr_relpath, $hce_curr_displaypath, $home_dirname, $commons_rel_dir;
     global $pguser, $home_dir_created, $autoprefix_message;
@@ -270,7 +269,7 @@ function do_showdir($action_message)
     show_caveats();
 }
 
-function do_showupload()
+function do_showupload(): void
 {
     global $curr_relpath, $hce_curr_displaypath;
     global $pguser, $autoprefix_message;
@@ -292,7 +291,7 @@ function do_showupload()
     show_return_link();
 }
 
-function do_upload()
+function do_upload(): void
 {
     global $curr_abspath, $hce_curr_displaypath;
     global $pguser, $despecialed_username;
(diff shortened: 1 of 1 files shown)


tools/project_manager/remote_file_manager.php
```diff
@@ -737,14 +737,14 @@ function is_valid_move_destination($dir)
  * removing empty components, rejecting parent traversal,
  * and re-joining them.
  *
- * Returns False is path contains invalid components.
+ * Returns null is path contains invalid components.
  */
-function canonicalize_path($relpath)
+function canonicalize_path(string $relpath): ?string
```
jchaffraix, line 741: This function and `get_directory_items_sorted` were switched from returning `false` on error to returning `null` as it yields to be a better type. I updated the callers, which were all in this file, to check against `null`.

tools/project_manager/remote_file_manager.php
```diff
@@ -1075,33 +1078,31 @@ function show_message($type, $message)
 /**
  * Display a return link (to the 'showdir' view)
  */
-function show_return_link($relpath = null)
+function show_return_link(): void
```
jchaffraix, line 1080: None of the callers passed the only argument so I removed it and simplified the code below.

tools/project_manager/remote_file_manager.php
```diff
@@ -204,14 +204,13 @@
     default:
         // no matching $action in input
         fatal_error(sprintf(_("Invalid action: '%s'"), html_safe($action)));
-        exit;
```
jchaffraix, line 207: This `exit` is unreachable as `fatal_error` doesn't return.

srjfoo (approved):

tools/project_manager/remote_file_manager.php
```diff
@@ -737,14 +736,14 @@ function is_valid_move_destination($dir)
  * removing empty components, rejecting parent traversal,
  * and re-joining them.
  *
- * Returns False is path contains invalid components.
+ * Returns null is path contains invalid components.
```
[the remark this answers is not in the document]
jchaffraix: I don't know if I understood your question so I rephrased this sentence as it had a typo (`is` should have been `if`). Let me know if the updated phrasing clears up your confusion:

> Returns null if `$relpath` contains any invalid component.

chrismiceli (approved):

tools/project_manager/remote_file_manager.php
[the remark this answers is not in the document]
srjfoo: Yes, that clarifies it. It should have occurred to me that it might be an is/if typo. Thanks!

cpeel (approved):

70ray (approved):

PR #1336, opened 2024-09-22
Type ProjectTransition

jchaffraix:
The types are inferred from the
callers and the stored values.

Manually tested doing some transition,
either from a PM or a site-admin role.

_Update from cpeel_: sandbox at https://www.pgdp.org/~cpeel/c.branch/julien_type_ProjectTransition/

### DIFF
--- a/pinc/ProjectTransition.inc
+++ b/pinc/ProjectTransition.inc
@@ -190,7 +190,7 @@ class ProjectTransition
      * It should perhaps also be the only place that modifies these columns:
      *     modifieddate checkedoutby postproofer postcomments
      */
-    public function do_state_change(Project $project, string $who, $extras): string
+    public function do_state_change(Project $project, string $who, array $extras): string
     {
         global $testing;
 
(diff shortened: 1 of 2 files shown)


chrismiceli (approved):

cpeel (approved):
Code lgtm but I'd appreciate @srjfoo running it through its paces via things like normal manual transitions, automatic transitions via automodify, and esoteric SA transitions via SA tooling and noncvs scripts (if applicable).

srjfoo (approved):

pinc/ProjectTransition.inc
```diff
@@ -195,7 +190,7 @@ class ProjectTransition
      * It should perhaps also be the only place that modifies these columns:
      *     modifieddate checkedoutby postproofer postcomments
      */
-    public function do_state_change($project, $who, $extras)
+    public function do_state_change(Project $project, string $who, $extras): string
```
70ray, line 193: Can we say that $extras is an array?

jchaffraix: Indeed as all callers pass an array. Updated in the latest patch.

chrismiceli (approved):

70ray (approved):

PR #1337, opened 2024-09-22
Type tools/charsuites.php

jchaffraix:
This uncovered a wrong type in CharSuite.
The type only impacted PHPStan though.

Tested by opening tools/charsuites.php.

### DIFF
--- a/tools/charsuites.php
+++ b/tools/charsuites.php
@@ -88,7 +88,7 @@ if ($charsuite) {
 
 //----------------------------------------------------------------------------
 
-function output_charsuite($charsuite, $show_detail = false)
+function output_charsuite(CharSuite $charsuite, bool $show_detail = false): void
 {
     $slug = utf8_url_slug($charsuite->title);
     echo "<h2 id='$slug'>" . html_safe($charsuite->title) . "</h2>";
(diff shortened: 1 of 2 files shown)


tools/charsuites.php
```diff
@@ -198,7 +201,7 @@ function output_characters_slice($slice)
             echo "<span class='gs-char'>$char</span><br>";
             echo "<span class='gs-codepoint'>$codepoint</span>";
             echo "</td>";
-        } else {
+        } else { /** @phpstan-ignore-line */
```
jchaffraix, line 204: This is needed to neuter a PHPStan warning: this `else` is never reached based on the type as `string` can never be `null`.

We can probably remove this branch but for now, I am sticking with disabling it.

chrismiceli (approved):

tools/charsuites.php
srjfoo: Maybe I'm not understanding what you mean here? Several of the picker sets do include `null`, and this function is used to display the picker sets.

jchaffraix: Indeed, I misread the code and mistyped a set as `string[]` (ie not allowing for the possibility of `null` in the array, which made the `else` branch unreachable). Updated several types to `string|null` removes the need for `@phpstan-ignore-line` on this line.

As a side note, do you know what `null` means in a Picker's range? Some kind of placeholder char?

cpeel: IIRC we use null for layout (an empty placeholder).

cpeel (approved):

tools/charsuites.php
srjfoo: > IIRC we use null for layout (an empty placeholder).

Yes, several of the character suites use null in picker sets.

srjfoo (approved):

70ray (approved):

PR #1338, opened 2024-09-23
Type pinc/wordcheck_engine.pinc

jchaffraix:
There were some missing return types and params.

When possible, I have narrowed down the PHPStan
annotations so they are more useful (the runtime
behavior in this case is unchanged).

TEST=Tested adding/removing words to the site-wide and the project-wide lists.

_Update by cpeel_: sandbox at https://www.pgdp.org/~cpeel/c.branch/julien_type_wordcheck_engine/

### DIFF
--- a/pinc/wordcheck_engine.inc
+++ b/pinc/wordcheck_engine.inc
@@ -251,7 +251,7 @@ function get_bad_words_for_text($text, $languages, $word_lists = [])
 
 class BadWordAccumulator
 {
-    /** @var string[] */
+    /** @var array<string, int> */
     public array $words;
     /** @var string[] */
     public array $messages;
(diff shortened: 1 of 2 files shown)


pinc/wordcheck_engine.inc
```diff
@@ -251,7 +251,7 @@ function get_bad_words_for_text($text, $languages, $word_lists = [])
 
 class BadWordAccumulator
 {
-    /** @var string[] */
+    /** @var array<string, int> */
```
jchaffraix, line 254: This was mistyped. To convince oneself, just look at how `add_bad_words` inserts words:

```
$this->words[$word] = $val;
```

chrismiceli (approved):

cpeel (approved):

srjfoo (approved):

70ray (approved):

PR #1339, opened 2024-09-25
Type pinc/DPDatabase.inc

jchaffraix:
The function is core to all the code so this
PR adds some integration tests to make sure
that the basic operations still work.

The tests uncovered a few issues that were fixed,
like `close` wouldn't clear the connection or
`query` could return an unset variable.

The code in DPDatabase.pinc was switched to the
object-oriented API for easier typing but the
caller are mostly using the imperative API. The
mysqli documentation confirmed that this should
be fine, albeit ugly.

### DIFF
--- a/SETUP/tests/unittests/DPDatabaseTest.php
+++ b/SETUP/tests/unittests/DPDatabaseTest.php
@@ -18,12 +18,12 @@ class DPDatabaseTest extends PHPUnit\Framework\TestCase
         }
     }
 
-    public function test_does_table_exit_existing_table(): void
+    public function test_does_table_exist_existing_table(): void
     {
         $this->assertSame(true, DPDatabase::does_table_exist("dpdb_tests"));
     }
 
-    public function test_does_table_exit_inexistent_table(): void
+    public function test_does_table_exist_inexistent_table(): void
     {
         $this->assertSame(false, DPDatabase::does_table_exist("inexistent_for_sure"));
     }
--- a/SETUP/tests/unittests/DPDatabaseTest.php
+++ b/SETUP/tests/unittests/DPDatabaseTest.php
@@ -1,0 +1,83 @@
+<?php
+
+// DPDatabase basic tests
+
+class DPDatabaseTest extends PHPUnit\Framework\TestCase
+{
+    public static function setUpBeforeClass(): void
+    {
+        // Sanity check: Remove the existing table so we start from a clean state.
+        $r = DPDatabase::query("DROP TABLE IF EXISTS `dpdb_tests`");
+        if ($r === false) {
+            throw new Exception("Failed to clean-up test table");
+        }
+
+        $r = DPDatabase::query("CREATE TABLE `dpdb_tests` (`id` int(20) NOT NULL, `comment` varchar(25) NOT NULL default '', PRIMARY KEY (`id`))");
+        if ($r === false) {
+            throw new Exception("Failed to create test table");
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


SETUP/tests/unittests/DPDatabaseTest.php
```diff
@@ -0,0 +1,83 @@
...
+            throw new Exception("Failed to clean-up test table");
+        }
+
+        $r = DPDatabase::query("CREATE TABLE `dpdb_tests` (`id` int(20) NOT NULL, `comment` varchar(25) NOT NULL default '', PRIMARY KEY (`id`))");
+        if ($r === false) {
+            throw new Exception("Failed to create test table");
+        }
+    }
+
+    public function test_does_table_exit_existing_table(): void
```
cpeel, line 21:
```suggestion
    public function test_does_table_exist_existing_table(): void
```

SETUP/tests/unittests/DPDatabaseTest.php
```diff
@@ -0,0 +1,83 @@
...
+    {
+        $this->assertSame(true, DPDatabase::does_table_exist("dpdb_tests"));
+    }
+
+    public function test_does_table_exit_inexistent_table(): void
```
cpeel, line 26:
```suggestion
    public function test_does_table_exist_inexistent_table(): void
```

SETUP/tests/unittests/DPDatabaseTest.php
```diff
@@ -0,0 +1,83 @@
...
+        $this->assertTrue($r);
+
+        // Fetch the row back.
+        $r = DPDatabase::query("SELECT * FROM `dpdb_tests` where id=0");
+        if (is_bool($r)) {
+            throw new Exception("Got a boolean from select on DPDatabase::query: {$r}");
+        }
+        $this->assertSame(1, $r->num_rows);
+        $this->assertSame(1, DPDatabase::affected_rows());
+        $row = $r->fetch_array();
```
cpeel, line 43: I have a slight preference for using `mysqli_fetch_array()` here as additional validation that the procedural methods work since that's what we use everywhere else.

SETUP/tests/unittests/phpunit_bootstrap.php
```diff
@@ -8,6 +8,8 @@
 
 // Reconnect to the test DB
 DPDatabase::close();
+// Note: The connect arguments need to be kept in sync with DPDatabaseTest.php
```
cpeel, line 11: This is true but it's true also for `smoketests/test_tables.sql` and `ci_configuration.sh` -- we should either call out all the places or none of them so we don't mislead readers. Could just say:
```
// Note: These values need to be kept in sync with other files in SETUP/tests/
```

cpeel (commented):
Thanks for adding unit tests!

SETUP/tests/unittests/DPDatabaseTest.php
jchaffraix: Good point, switched to `mysqli_fetch_array`.

SETUP/tests/unittests/phpunit_bootstrap.php
jchaffraix: Sure, updated the comment to your suggestion.

cpeel (approved):

chrismiceli (approved):

70ray (approved):

srjfoo (approved):

PR #1342, opened 2024-09-29
Issue #1274 remove false positive security violation in phpbb3.inc

mrducky4:
The `check_security` script complained about comments in `phpbb3.inc`, so I changed the comments
and removed it from the `ok_system_calls` whitelist.

### DIFF
--- a/pinc/phpbb3.inc
+++ b/pinc/phpbb3.inc
@@ -20,7 +20,7 @@ $this_is_top_level_cli_invocation = (
 );
 // i.e., this file is being executed because it was the
 // script argument to a command-line invocation of php,
-// probably invoked by DP code via Process() or exec () or system ().
+// probably invoked by DP code via Process().
 
 if ($this_is_top_level_cli_invocation) {
     // We assume it was invoked from the pinc/ directory.
@@ -356,6 +356,6 @@ if ($this_is_top_level_cli_invocation) {
     $result = call_user_func_array($func_name, $args);
     echo $result;
     // So that $result will show up as the last line of input,
-    // which is captured & returned by Process(), exec () and system (),
+    // which is captured & returned by Process(),
     // assuming that's how this is being invoked.
 }
--- a/pinc/phpbb3.inc
+++ b/pinc/phpbb3.inc
@@ -20,7 +20,7 @@ $this_is_top_level_cli_invocation = (
 );
 // i.e., this file is being executed because it was the
 // script argument to a command-line invocation of php,
-// probably invoked by DP code via exec() or system().
+// probably invoked by DP code via Process() or exec () or system ().
 
 if ($this_is_top_level_cli_invocation) {
     // We assume it was invoked from the pinc/ directory.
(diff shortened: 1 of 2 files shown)


pinc/phpbb3.inc
```diff
@@ -20,7 +20,7 @@ $this_is_top_level_cli_invocation = (
 );
 // i.e., this file is being executed because it was the
 // script argument to a command-line invocation of php,
-// probably invoked by DP code via exec() or system().
+// probably invoked by DP code via Process() or exec () or system ().
```
cpeel, line 23: Putting a space in here feels fragile. It would be better if we stripped comments before checking for the functions (well, it would _really_ better if we were reading a parsed AST but that seems overkill).

Perhaps:
```diff
diff --git a/SETUP/ci/check_security.php b/SETUP/ci/check_security.php
index f687f1e7c..c46057360 100755
--- a/SETUP/ci/check_security.php
+++ b/SETUP/ci/check_security.php
@@ -13,7 +13,6 @@ $basedir = $argv[1] ?? "../../";
 //       others are added from the current set until they get updated.
 $ok_system_calls = [
     "pinc/forum_interface_phpbb3.inc",
-    "pinc/phpbb3.inc",
     "pinc/POFile.inc",
     "pinc/Project.inc",
     "pinc/project_quick_check.inc",
@@ -98,8 +97,15 @@ function file_includes_system_call(string $filename): bool
     // escapeshellcmd() but it's a sign that there's probably some system/shell
     // magic happening that we want to catch.

-    $contents = file_get_contents($filename);
-    return preg_match("/\b(?:exec|system|passthru|shell_exec|escapeshellcmd)\(/", $contents, $matches);
+    foreach (file($filename) as $line) {
+        // strip comments
+        $line = preg_replace("#//.*$#", "", $line);
+        $line = preg_replace("/\#.*$/", "", $line);
+        if (preg_match("/\b(?:exec|system|passthru|shell_exec|escapeshellcmd)\(/", $line, $matches)) {
+            return true;
+        }
+    }
+    return false;
 }

 /** @return never */
```

cpeel: I attempted, and failed, to fix the regex with a negative lookback for a start-of-comment so this seemed like the next-easiest thing and correctly detects the current uses of those functions in the code.

mrducky4: Or we could just completely remove `exec()` and `system()` from comments, and not worry about any potential problem with stripping comments using regular expressions.

mrducky4: For example, with simplistic comment stripping, something like this goes unnoticed:
`if ($s != "//") exec($something);`

mrducky4: Also, that simplistic comment stripping approach doesn't take into account comments using `/* */`.
So I updated the PR with my suggestion.

cpeel: Yup, this change is the simplest.

cpeel (approved):

Issue #1334, opened 2024-09-19
Don't show duplicate picker sets in proofreading toolbox


srjfoo:
If there are character suites equipped for a project that contain identical picker sets, the duplicates stack (see screenshot).

Currently, this only affects the Greek picker sets, because Basic Greek is a proper subset of Polytonic Greek, and the picker set in Basic Greek is identical to the first picker set in Polytonic Greek.

<img width="313" alt="image" src="https://github.com/user-attachments/assets/2193913c-f421-42a6-bdd9-6d9dfbba8bd2">

---

Historical note: We have both Basic Greek and Polytonic Greek because there are occasions where only the basic alphabet is needed, and so it seemed excessive to force the PM to have a large number of unnecessary picker sets for those projects. We do not currently have any other character suites where there are identical picker sets.

Position of characters in the picker sets does not matter. If exactly the same characters are in two picker sets, they will stack in the toolbox.

PR #1344, opened 2024-10-01
use unique identifier for character pickers

70ray:
This addresses issue #1334.
The problem was that an identifier is used to link the selector button with the appropriate picker set. This was based on the contents of the selector button which was the same for the two greek sets. Any unique identifier would serve. There are  some issues with uniquid() but it appears that it sleeps for a microsecond at each invocation so it works for our purpose. A serial number would also work.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/uniq_picker

### DIFF
--- a/pinc/CharacterSelector.inc
+++ b/pinc/CharacterSelector.inc
@@ -21,6 +21,7 @@ class CharacterSelector
         $selector_string = "<div id='selector_row'>";
         $row_string = "";
 
+        $pickerid = 1;
         foreach ($this->picker_sets as $picker_set) {
             if (!$picker_set) {
                 continue;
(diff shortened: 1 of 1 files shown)


pinc/CharacterSelector.inc
```diff
@@ -33,7 +33,7 @@ class CharacterSelector
             }
 
             foreach ($picker_set->get_subsets() as $code => $picker) {
-                $safe_code = bin2hex($code);
+                $safe_code = uniqid();
```
chrismiceli, line 36: if it just needs to be unique, why not do code + index or something. So if there are 2 with the same code, then you could append 1, 2, etc.? That would avoid any sleep issues.

70ray: Using sertal number now. Interestingly putting "picker_{$pickerid++}" causes an error.

pinc/CharacterSelector.inc
```diff
@@ -33,10 +34,11 @@ class CharacterSelector
             }
 
             foreach ($picker_set->get_subsets() as $code => $picker) {
```
chrismiceli, line 36: maybe a regular for loop would be more appropriate now that we are managing a counter variable.

chrismiceli (approved):

pinc/CharacterSelector.inc
cpeel: I think a foreach is still the right loop type because its iterating over a dictionary -- not an integer-indexed array. If PHP had `enumerate()` like python I'd say use that but it doesn't.

cpeel (approved):

srjfoo (approved):

PR #1346, opened 2024-10-04
Type project.php

jchaffraix:
The change uncovered some mistyped synthetic rows from the project events table in `fill_gaps_in_event`. Those were fixed to use the correct type. Also added a TODO as this underlines a missing class for the type.

TEST=Manipulated a project as PM (state change, add pages) and confirmed that it renders correctly.

_Update from cpeel_: sandbox at https://www.pgdp.org/~cpeel/c.branch/julien_type_projectphp/

### DIFF
--- a/project.php
+++ b/project.php
@@ -150,7 +150,7 @@ if ($detail_level == 1) {
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 
-function do_detail_level_switch()
+function do_detail_level_switch(): void
 {
     global $project, $detail_level, $detail_levels;
 
@@ -203,7 +203,7 @@ function get_bookmark_form(): string
 
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
 
-function do_expected_state()
+function do_expected_state(): void
 {
     global $project, $expected_state;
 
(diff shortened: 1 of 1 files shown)


chrismiceli (approved):

project.php
```diff
@@ -1313,15 +1314,18 @@ function do_history()
...
-function fill_gaps_in_events($in_events)
+// TODO(jchaffraix): Add a class for ProjectEvent and switch to this function to it.
+function fill_gaps_in_events(array $in_events): array
 {
     $out_events = [];
 
     // Creation at the start
     if (count($in_events) == 0 || $in_events[0]['event_type'] != 'creation') {
         $pseudo_event = [
-            'timestamp' => '?',
```
cpeel, line 1324: Background: `project_events` was added sometime after the site was live and so not all projects have records in the table. This creates pseudo records and for unknown timestamps we set it to `?` instead and output that value. This special value is used up on [line 1196](https://github.com/DistributedProofreaders/dproofreaders/pull/1346/files#diff-7b684f769c53e42bad88ffd69e373b9df6082ae4a6410d661aaf653a3865880eR1196) to output this value. For an example of this on PROD, see [projectID40bfb08353e74](https://www.pgdp.net/c/project.php?id=projectID40bfb08353e74).

If we want to make this a little more type friendly we could make `timestamp` an int-or-null and set it to null here and update line 1196.

cpeel: It would make more sense to me if this function pulled from the database, filled in any gaps, and return the final set instead of having the caller do the query and then pass it into here. Not that it needs addressing in this PR.

project.php
```diff
@@ -1595,7 +1599,7 @@ function do_post_files()
 }
 
 // -----------------------------------------------------------------------------
-function echo_upload_backup($projectid, $text, $stage)
+function echo_upload_backup(?string $projectid, string $text, string $stage): void
```
cpeel, line 1602: `$projectid` should not be null and if it that's probably a bug.

jchaffraix: Makes sense, I don't remember why I allowed it but reverted this.

project.php
jchaffraix: Thanks for the context, that's very useful.

> If we want to make this a little more type friendly we could make timestamp an int-or-null and set it to null here and update line 1196.

I have made this change. An unfortunate side effect is that we allow `null` which will never be returned from the DB (the field is marked as `NOT NULL default 0`). This seemed like minor diff though.

cpeel (approved):

srjfoo (approved):

70ray (approved):

PR #1347, opened 2024-10-04
Type Activity/Stage/Round/Pool

jchaffraix:
Those 4 are closely coupled so it made
sense to handle them as a single PR.

Unfortunately due to the coupling, the return of
some key getters were not added, instead I added
a comment explaning the situation. Properly typing will likely require some refactoring to split up
the per-type lists, which is better done in a follow-up.

TEST=Checked a user's available rounds, updated a project's state to confirm that the rounds/pools were still listed.

_Update from cpeel_: sandbox at https://www.pgdp.org/~cpeel/c.branch/julien_type_Activity/

### DIFF
--- a/pinc/Activity.inc
+++ b/pinc/Activity.inc
@@ -396,7 +396,7 @@ function get_user_scores(string $username, array $criteria): array
     $user_obj = $username == User::current_username() ? User::load_current() : new User($username);
 
     $user_scores = [];
-    foreach ($criteria as $criterion_code => $_) {
+    foreach (array_keys($criteria) as $criterion_code) {
         $user_scores[$criterion_code] = get_user_score($user_obj, $criterion_code);
     }
     return $user_scores;
--- a/pinc/Activity.inc
+++ b/pinc/Activity.inc
@@ -29,7 +29,26 @@ $ACCESS_CRITERIA = [
  */
 class Activities
 {
-    private static $_activities = [];
+    /*
+     * `$_activities` contains a mapping from a specific type of Activity (e.g. an
+     * Activity, a Stage, a Round or a Pool) to the instances of those. They are mapped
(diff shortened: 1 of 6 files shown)


pinc/Activity.inc
```diff
@@ -41,31 +60,40 @@ class Activities
...
     {
         foreach (array_merge([get_class($activity)], class_parents($activity)) as $class) {
             assert(!isset(self::$_activities[$class][$activity->id]), "Duplicate activity ID; IDs must be unique");
             self::$_activities[$class][$activity->id] = $activity;
         }
     }
 
-    public static function get_all()
+    /**
+     * The return of this function is untyped for now (see the comment above `$_activities` for why).
```
cpeel, line 77: This comment is likely to be used as the function description in the phpdocs that get generated. I think if we add these as a `//` style comment they won't. (also below)

pinc/Activity.inc
```diff
@@ -360,13 +396,13 @@ function get_user_scores($username, $criteria)
     $user_obj = $username == User::current_username() ? User::load_current() : new User($username);
 
     $user_scores = [];
-    foreach ($criteria as $criterion_code => $criterion_descr) {
+    foreach ($criteria as $criterion_code => $_) {
```
cpeel, line 399: Would this be clearer as
```suggestion
    foreach (array_keys($criteria) as $criterion_code) {
```
?

jchaffraix: I am not super opinionated on those so included the change as it made sense.

pinc/Activity.inc
jchaffraix: Done.

cpeel (approved):

70ray (approved):

srjfoo (approved):

PR #1348, opened 2024-10-10
Type the files in tools/authors/

jchaffraix:
Those are inter-related so it made sense to
handle them in one PR.

TEST=Added/updated several authors with a mix
of unknown fields and AC/BC dates.

_Update by cpeel_: sandbox at https://www.pgdp.org/~cpeel/c.branch/julien_type_tools_charsuite/

### DIFF
--- a/tools/authors/add.php
+++ b/tools/authors/add.php
@@ -188,7 +188,7 @@ echo "<h1>$title</h1>";
 
 echo_menu();
 
-function write_months_list($bd, $selected)
+function write_months_list(string $bd, int $selected): void
 {
     global $months;
     echo "<select name=\"$bd" . "month\" size=\"1\">\n";
@@ -201,7 +201,7 @@ function write_months_list($bd, $selected)
     }
     echo "</select>\n";
 }
-function write_days_list($bd, $selected)
+function write_days_list(string $bd, int $selected): void
 {
     echo "<select name=\"$bd" . "day\" size=\"1\">\n";
     echo "<option value=\"0\"" . ($selected == '0' ? ' SELECTED' : '') . ">" . _('Unknown');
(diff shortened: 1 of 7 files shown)


tools/authors/add.php
```diff
@@ -354,20 +354,25 @@ function isValidDate(year, month, day) {
...
     return $$var;
 }
-function echo_date_fields($bd)
+
+function echo_date_fields(string $bd): void
 {
     ?>
 <table class='no-border'>
 <tr><td><?php echo _('Month'); ?>:</td><td>
-<?php echo write_months_list($bd, _var($bd, 'month')); ?></td></tr>
```
jchaffraix, line 368: This needed to be removed as `write_month_list` (and `write_days_list` below) is void and already `echo` their content.

tools/authors/add.php
```diff
@@ -201,7 +201,7 @@ function write_months_list($bd, $selected)
     }
     echo "</select>\n";
 }
-function write_days_list($bd, $selected)
+function write_days_list(string $bd, int $selected): void
 {
     echo "<select name=\"$bd" . "day\" size=\"1\">\n";
     echo "<option value=\"0\"" . ($selected == '0' ? ' SELECTED' : '') . ">" . _('Unknown');
```
jchaffraix, line 207: Note that we check against the string '0' using a non-strict equality on this line.

We could switch to a strict equality with 0 though. I punted on this as it seemed small (and PHP may already optimize this).

chrismiceli (approved):

cpeel (approved):
The `tools/authors/*` code is such a hot mess and I keep hoping we'll just purge it altogether.

70ray (approved):

srjfoo (approved):

PR #1349, opened 2024-10-10
check if timespan exists

70ray:
it will not when search refine or result configure.

### DIFF
--- a/tools/project_manager/projectmgr.js
+++ b/tools/project_manager/projectmgr.js
@@ -13,7 +13,9 @@ window.addEventListener("DOMContentLoaded", () => {
     });
 
     function showTime() {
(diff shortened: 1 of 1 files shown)


tools/project_manager/projectmgr.js
```diff
@@ -13,7 +13,9 @@ window.addEventListener("DOMContentLoaded", () => {
     });
 
     function showTime() {
-        timeSpan.textContent = dateTimeFormat.format(Date.now());
+        if (timeSpan) {
```
chrismiceli, line 16: minor, but you could put the if before even setting up the interval or anything to save CPU cycles every minute, but yeah...not a big deal.

chrismiceli (approved):

cpeel (approved):

srjfoo (approved):

PR #1352, opened 2024-10-11
Add checks for corrupt illustrations in PQC

cpeel:
Project Quick Check (PQC) already checked for corrupt page images (pngs) but not for corrupt illustrations (usually but not always jpegs). This adds illustration file checking by centralizing image checking functionality into a new `ImageUtils` class. This class can later be expanded to include using `pngcrush`/`optipng` on images on project load too, hence the more generic name for the class.

Sandbox at https://www.pgdp.org/~cpeel/c.branch/add-jpegs-pqc/

Two projects with "problem" images on TEST:
* Bad JPG illus: `projectID5e6821d744beb`
* Bad page PNG: `projectID4c49f39aa829c`

### DIFF
--- a/pinc/ImageUtils.inc
+++ b/pinc/ImageUtils.inc
@@ -1,0 +1,76 @@
+<?php
+
+use Symfony\Component\Process\Process;
+
+class ImageUtils
+{
+    public const IMAGE_OK = 1;
+    public const IMAGE_CORRUPT = 2;
+    public const IMAGE_SKIPPED = 3;
+
+    protected array $external_tools = [
+        "pngcheck" => false,
+        "jpeginfo" => false,
+    ];
+
+    public function __construct()
+    {
+        $this->check_external_tools();
+    }
+
+    protected function check_external_tools(): void
+    {
+        foreach (array_keys($this->external_tools) as $tool) {
+            $process = new Process([$tool, "-h"]);
+            $process->run();
+            if ($process->isSuccessful()) {
+                $this->external_tools[$tool] = true;
+            }
+        }
+    }
+
+    /**
+     * Checks if this class can validate image files given an extension.
+     */
+    public function can_validate(string $extension): bool
+    {
+        $extension = strtolower($extension);
+        if ($extension == "png" && $this->external_tools["pngcheck"]) {
+            return true;
+        } elseif (($extension == "jpeg" || $extension == "jpg") && $this->external_tools["jpeginfo"]) {
+            return true;
+        }
+
+        return false;
+    }
+
+    /**
+     * Validates integrity of the specified image using external tools.
+     *
+     * @returns array<int, string>
+     */
+    public function validate_integrity(string $filename): array
+    {
+        $extension = strtolower(pathinfo($filename, PATHINFO_EXTENSION));
+
+        if (!$this->can_validate($extension)) {
+            return [ImageUtils::IMAGE_SKIPPED, ""];
+        }
+
+        if ($extension == "png") {
+            $checker = ["pngcheck"];
+        } elseif ($extension == "jpeg" || $extension == "jpg") {
+            $checker = ["jpeginfo", "--check"];
+        }
+
+        $checker[] = $filename;
+        $process = new Process($checker);
+        $process->run();
+        if (!$process->isSuccessful()) {
+            $output = explode("\n", $process->getOutput());
+            return [ImageUtils::IMAGE_CORRUPT, $output[0]];
+        }
+
+        return [ImageUtils::IMAGE_OK, ""];
+    }
+}
--- a/pinc/ImageUtils.inc
+++ b/pinc/ImageUtils.inc
@@ -62,6 +62,8 @@ class ImageUtils
             $checker = ["pngcheck"];
         } elseif ($extension == "jpeg" || $extension == "jpg") {
             $checker = ["jpeginfo", "--check"];
+        } else {
+            throw new LogicException("Unsupported extension found, update can_validate()");
         }
 
         $checker[] = $filename;
--- a/pinc/ImageUtils.inc
+++ b/pinc/ImageUtils.inc
@@ -8,6 +8,7 @@ class ImageUtils
     public const IMAGE_CORRUPT = 2;
     public const IMAGE_SKIPPED = 3;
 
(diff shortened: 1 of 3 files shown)


chrismiceli (approved):

srjfoo (approved):

pinc/ImageUtils.inc
```diff
@@ -0,0 +1,76 @@
...
+
+use Symfony\Component\Process\Process;
+
+class ImageUtils
+{
+    public const IMAGE_OK = 1;
+    public const IMAGE_CORRUPT = 2;
+    public const IMAGE_SKIPPED = 3;
+
+    protected array $external_tools = [
```
jchaffraix, line 11: Nit: Worth adding the type for PHPStan for good measure:

```suggestion
    /** @var array<string, bool> */
    protected array $external_tools = [
```

pinc/ImageUtils.inc
```diff
@@ -0,0 +1,76 @@
...
+
+        if (!$this->can_validate($extension)) {
+            return [ImageUtils::IMAGE_SKIPPED, ""];
+        }
+
+        if ($extension == "png") {
+            $checker = ["pngcheck"];
+        } elseif ($extension == "jpeg" || $extension == "jpg") {
+            $checker = ["jpeginfo", "--check"];
+        }
```
jchaffraix, line 64: Nit: If we pass an unknown extension, we will be using the undefined variable `$checker` below:

```suggestion
        } else {
            throw new Exception("Unsupported extension {$extension}, did you forget to add it to ImageUtils?");
        }
```

pinc/project_quick_check.inc
```diff
@@ -223,19 +222,14 @@ function _test_project_for_corrupt_pngs($projectid)
...
-
-            if (strtolower($path_info["extension"]) == "png") {
-                $output = "";
-                exec('pngcheck -q ' . escapeshellarg($file_info->abs_path), $output, $return_code);
-                if ($return_code != 0) {
-                    $details .= "<tr>";
-                    $details .= "<td><a href='{$file_info->abs_url}'>" . html_safe($image) . "</a></td>";
-                    $details .= "<td>" . str_replace($file_info->abs_path, "", $output[0]) . "</td>";
-                    $details .= "</tr>";
+            [$image_status, $image_message] = $checker->validate_integrity($file_info->abs_path);
```
jchaffraix, line 225: It _seems_ like we could be validating `jpeg` images on this line as we removed the extension check on l. 228. Just pointing this out in case that was not intentional.

jchaffraix (approved):
The logic seems fine, some comments that be handled in follow-up PRs.

pinc/ImageUtils.inc
cpeel: The `$this->can_validate()` above it confirms that we can validate it already. This does require that those be kept in sync and I'll add something similar as a straps-and-suspenders.

pinc/project_quick_check.inc
cpeel: It was intentional because I didn't want to refactor the function that is very png-focused. Basically I punted it for now.

srjfoo:
Hm ... second round of testing, I went in and intentionally copied a `png` illustration and renamed it as a `gif`. I get no errors, and all PQC says for Illustration Image Files is "Success 51 valid illustration files found". With the `gif`, there should be 52. projectID5e267334297df

So, I guess I need some clarification -- will that sort of thing just be ignored, or is it supposed to cause an error?

cpeel:
> Hm ... second round of testing, I went in and intentionally copied a `png` illustration and renamed it as a `gif`. I get no errors, and all PQC says for Illustration Image Files is "Success 51 valid illustration files found". With the `gif`, there should be 52. projectID5e267334297df

For whatever reason we don't consider `gif`s to be either pages or illustrations. For instance, gifs don't show up in the [Image Index](https://www.pgdp.org/~cpeel/c.branch/add-jpegs-pqc/tools/proofers/images_index.php?project=projectID5e267334297df) for the project. The `gif` in question shows up as "Extra Files" on the [project page](https://www.pgdp.org/~cpeel/c.branch/add-jpegs-pqc/project.php?id=projectID5e267334297df).

It seems we're pretty consistent on illustrations just being pngs or jpegs:
* https://github.com/DistributedProofreaders/dproofreaders/blob/master/project.php#L1431
* https://github.com/DistributedProofreaders/dproofreaders/blob/master/pinc/Project.inc#L1308
* https://github.com/DistributedProofreaders/dproofreaders/blob/master/tools/project_manager/update_illos.php#L23

srjfoo:
> > Hm ... second round of testing, I went in and intentionally copied a `png` illustration and renamed it as a `gif`. I get no errors, and all PQC says for Illustration Image Files is "Success 51 valid illustration files found". With the `gif`, there should be 52. projectID5e267334297df
> 
> For whatever reason we don't consider `gif`s to be either pages or illustrations. For instance, gifs don't show up in the [Image Index](https://www.pgdp.org/~cpeel/c.branch/add-jpegs-pqc/tools/proofers/images_index.php?project=projectID5e267334297df) for the project. The `gif` in question shows up as "Extra Files" on the [project page](https://www.pgdp.org/~cpeel/c.branch/add-jpegs-pqc/project.php?id=projectID5e267334297df).
> 
> It seems we're pretty consistent on illustrations just being pngs or jpegs: 
> * https://github.com/DistributedProofreaders/dproofreaders/blob/master/project.php#L1431
> * https://github.com/DistributedProofreaders/dproofreaders/blob/master/pinc/Project.inc#L1308
> * https://github.com/DistributedProofreaders/dproofreaders/blob/master/tools/project_manager/update_illos.php#L23

Yes -- and in fact, I didn't think we allowed the `jpeg` extension, just `jpg`, though I may be wrong.

And maybe I'm thinking, too, of project loads. `add_files` will, for instance, identify a tiff file as an unrecognized format, and refuse to load it, and I guess I was expecting PQC to identify anything in the directory that shouldn't be there as an error.

From what you say, though, it's just ignored. And thanks for mentioning the "Extra Files" list. I didn't even think to look there.

srjfoo (approved):

jchaffraix (approved):

70ray (approved):

PR #1353, opened 2024-10-12
Remove shell commands from add_files.php

cpeel:
I went to add the corrupt page & illustration checks to the project load and decided it would be best to start with removing the shell commands. More work towards #1274.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/remove-shell-add-files/

### DIFF
--- a/tools/project_manager/add_files.php
+++ b/tools/project_manager/add_files.php
@@ -485,8 +485,7 @@ class Loader
 
         // Filenames starting with a hyphen are excluded because
         // they may be mistaken with command line options of
-        // shell utilities such zip
-        // (when generating zips to download image files)
+        // shell utilities.
         if (preg_match('/^-/', $filename)) {
             return _('filename starts with a hyphen');
         }
--- a/tools/project_manager/add_files.php
+++ b/tools/project_manager/add_files.php
@@ -828,9 +828,7 @@ class Loader
                         $now
                     );
                 } catch (DBQueryError $error) {
-                    echo "<p class='error'>";
-                    echo "for base=$base, project_add_page raised a DB error";
-                    echo "</p>";
+                    $this->error("for base=$base, project_add_page raised a DB error");
                 }
 
                 if (!copy($src_image_file_name, "$this->dest_project_dir/$src_image_file_name")) {
--- a/tools/project_manager/add_files.php
+++ b/tools/project_manager/add_files.php
@@ -103,20 +103,10 @@ if (substr($abs_source, -4) == ".zip") {
     $source_project_dir = $abs_source;
 }
 
-// Attempt to make everything all-readable.
-// (This will probably only succeed if we have just unpacked
-// a zip file [above], but no harm in trying in all cases.)
-exec("chmod -R a+r " . escapeshellarg($source_project_dir));
-
-
 // Rather than performing commands from an arbitrary location,
-// using absolute paths, e.g.
-//     system("cp $source_project_dir/*.png $dest_project_dir");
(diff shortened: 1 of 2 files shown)


tools/project_manager/add_files.php
```diff
@@ -103,20 +103,10 @@
     $source_project_dir = $abs_source;
 }
 
-// Attempt to make everything all-readable.
```
cpeel, line 106: I don't know why we do this. The script needs read permissions from the source files but if the `www-data` webserver user created the files (which would be the case from an extracted zip or something uploaded through Remote File Manager) then it has read permissions on it already.

This might be an artifact from when we allowed ftp uploads but that is a long time gone.

chrismiceli (approved):

srjfoo (approved):

tools/project_manager/add_files.php
cpeel: To change the permissions on the files we have to have write access, and if we have write access we probably have read access. And the only thing we actually need is read access to copy it from this location to the new location and when its copied it will be owned by the `www-data` user, so I don't think this is needed at all.

tools/project_manager/add_files.php
```diff
@@ -496,7 +485,7 @@ public function _check_file(string $filename): string
 
         // Filenames starting with a hyphen are excluded because
         // they may be mistaken with command line options of
-        // shell utilities such as cmp (in this file, below) and zip
+        // shell utilities such zip
```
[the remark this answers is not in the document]
cpeel: Actually, the `such as zip` isn't true either since we use zip library calls to create zips now. I've updated the comment.

mrducky4 (approved):

tools/project_manager/add_files.php
```diff
@@ -621,8 +609,7 @@ public function _get_action(string $base, string $toi, array $db_exts, array $sr
             $same = (
                 is_file($p_file)
                 &&
-                shell_exec("cmp " . escapeshellarg($src_file)
-                    . " " . escapeshellarg($p_file)) == ''
+                md5_file($src_file) == md5_file($p_file)
```
jchaffraix, line 612: For my understanding, we are not concerned about collisions on those files as uploading require the PM permission, which is only for a few selected users, right?

tools/project_manager/add_files.php
```diff
@@ -835,108 +818,61 @@ public function execute(): void
...
+                        $base,
+                        $src_image_file_name,
+                        $src_text_file_path,
+                        $pguser,
+                        $now
+                    );
+                } catch (DBQueryError $error) {
+                    echo "<p class='error'>";
+                    echo "for base=$base, project_add_page raised a DB error";
+                    echo "</p>";
```
jchaffraix, line 833: Nit: This could use the new `error` helper:

```suggestion
                    $this->error("for base=$base, project_add_page raised a DB error");
```

jchaffraix (approved):

tools/project_manager/add_files.php
cpeel: Exactly correct -- we don't need a bytelevel comparison and low probability of collisions in the md5 space are acceptable. Also, the failure scenario is that we determine they are not the same and the project file is replaced by a user who has permissions to do that anyway.

srjfoo (approved):

chrismiceli (approved):

cpeel:
Going to get this merged in and a PR up that builds off of this to check for corrupt images at project load time.

PR #1355, opened 2024-10-12
Move misc.inc include above exception handlers

cpeel:
Our exception handlers use `html_safe()` so we need to include `misc.inc` before connecting to the database in case it raises an exception. Discovered this on the clone test system.

### DIFF
--- a/pinc/bootstrap.inc
+++ b/pinc/bootstrap.inc
@@ -30,10 +30,6 @@ spl_autoload_register('dp_class_autoloader');
 // Load composer autoloader
 require_once($relPath.'../vendor/autoload.php');
 
-// Include misc.inc which contains useful functions that almost 90% of
-// the pages use in some way, including our exception handlers.
-include_once($relPath.'misc.inc');
-
 // Exception handlers are defined differently for HTML and the API
 set_exception_handler($testing ? 'test_exception_handler' : 'production_exception_handler');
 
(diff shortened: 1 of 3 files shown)


srjfoo (approved):

chrismiceli (approved):

pinc/bootstrap.inc
```diff
@@ -30,6 +30,10 @@ spl_autoload_register('dp_class_autoloader');
 // Load composer autoloader
 require_once($relPath.'../vendor/autoload.php');
 
+// Include misc.inc which contains useful functions that almost 90% of
+// the pages use in some way, including our exception handlers.
+include_once($relPath.'misc.inc');
```
70ray, line 35: There is a "DPDatabase::escape" in misc.inc. Could that cause a problem?

cpeel: Good catch. It isn't technically a problem because `set_col_str()` isn't called from the handlers so PHP won't care. It's a little unfortunate that there's a function in `misc` that relies on other DP code though. Ah look, there's `dp_setcookie()` used in the csrf_token code too.

So while technically this works I think I'll update the exception handlers to escape the HTML differently.

cpeel:
I've changed the approach to remove `html_safe()` from the exception handlers and added a comment about not using "outside" functions in them for the next person who touches them.

chrismiceli (approved):

srjfoo (approved):

70ray (approved):

PR #1356, opened 2024-10-12
Remove array_get

jchaffraix:
The function cannot be properly typed as it
is and fixing this would take a while, removing
it is shorter and better.

### DIFF
--- a/pinc/misc.inc
+++ b/pinc/misc.inc
@@ -11,18 +11,6 @@
 use voku\helper\UTF8;
 
 /**
- * Return $arr[$key], or if it's not defined, $default.
- */
-function array_get($arr, $key, $default)
-{
-    if (isset($arr[$key])) {
-        return $arr[$key];
-    } else {
-        return $default;
-    }
-}
-
-/**
  * Given an array of arrays/objects, return the values of the desired field.
  */
 function array_extract_field($array, $field)
@@ -50,7 +38,7 @@ function array_extract_field($array, $field)
  */
 function array_get_as_array($array, $key, $default)
 {
-    $value = array_get($array, $key, $default);
+    $value = $array[$key] ?? $default;
     if (!is_null($value) && !is_array($value)) {
         $value = [$value];
     }
--- a/pinc/misc.inc
+++ b/pinc/misc.inc
@@ -11,6 +11,22 @@
 use voku\helper\UTF8;
 
 /**
+ * Return $arr[$key], or if it's not defined, $default
+ *
+ * @deprecated
+ * @param string|int $key
+ * @return mixed
+ */
+function array_get(array $arr, $key, $default)
+{
+    if (isset($arr[$key])) {
+        return $arr[$key];
+    } else {
+        return $default;
+    }
+}
+
+/**
  * Given an array of arrays/objects, return the values of the desired field.
  */
 function array_extract_field($array, $field)
--- a/SETUP/tests/manual_web/page_compare/manual_page_compare_test.php
+++ b/SETUP/tests/manual_web/page_compare/manual_page_compare_test.php
@@ -10,8 +10,8 @@ include_once($relPath."PageUnformatter.inc"); // PageUnformatter()
 
 require_login();
 
-$L_input = array_get($_POST, "Ltex", "");
-$R_input = array_get($_POST, "Rtex", "");
+$L_input = $_POST["Ltex"] ?? "";
+$R_input = $_POST["Rtex"] ?? "";
 $unwrap = isset($_POST['Unwrap']); // can only be 'on'
 $ignore_case = isset($_POST['IgnoreCase']);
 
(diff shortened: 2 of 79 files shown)


pinc/misc.inc
```diff
@@ -10,18 +10,6 @@
 
 use voku\helper\UTF8;
 
-/**
- * Return $arr[$key], or if it's not defined, $default.
- */
-function array_get($arr, $key, $default)
```
cpeel, line 16: We should not remove this function yet because it's used by code that isn't checked into git (ie: `noncvs` code) and perhaps others who use our code.

Instead, we should add typing for the bits we know (`array $arr`), set it as `@deprecated`, and do a `trigger_error("array_get() is deprecated, use the null coalescing operator instead", E_USER_DEPRECATED);`.

cpeel (commented):
This is a lot and I will need to work my way through it.

I'm not against this change, but I do want to ask: does doing this help PHPStan check our code better than using `array_get()` without typing? I want to make sure there's some value-add and that we're not doing it just to do it. My comments on the `array_get()` function assume the answer is "yes, there is value".

jchaffraix:
> This is a lot and I will need to work my way through it.
> 
> I'm not against this change, but I do want to ask: does doing this help PHPStan check our code better than using `array_get()` without typing? I want to make sure there's some value-add and that we're not doing it just to do it. My comments on the `array_get()` function assume the answer is "yes, there is value".

Yes, there is value as PHPStan can infer the narrow type of the return/args when using `??` but won't with `array_get`. Even if we add types, those will have to be the super-types (`array` for the arg and `mixed` for the return).

Let's take those 2 equivalent statements: `$test_var = array_get($_POST, 'test', null)` and `$test_var = _$POST['test'] ?? null`.

In the first case, PHPStan doesn't know the return of `array_get` so it can't do any kind of validation on the usage of `$test_var`. If we added `mixed` as the return, PHPStan will only do minimal checks (`mixed` is a super-type for all types) and will have to ask us to cast the narrow type when needed (e.g. when doing a function call).
In the second case, PHPStan can infer that the variable is of type `?string` and warn/error accordingly.

cpeel:
> Let's take those 2 equivalent statements: `$test_var = array_get($_POST, 'test', null)` and `$test_var = _$POST['test'] ?? null`.
> 
> In the first case, PHPStan doesn't know the return of `array_get` so it can't do any kind of validation on the usage of `$test_var`. If we added `mixed` as the return, PHPStan will only do minimal checks (`mixed` is a super-type for all types) and will have to ask us to cast the narrow type when needed (e.g. when doing a function call). In the second case, PHPStan can infer that the variable is of type `?string` and warn/error accordingly.

`$_POST` by definition only contains strings so let's broaden it up a bit. If I have `$test_var = $some_random_array[$some_var] ?? 42` is PHPStan able to infer the value type of that element of that array from other code introspection? Because if not it's no more useful than `array_get($some_random_array, $some_var, 42)` and that's what I'm trying to tease apart.

jchaffraix:
> > Let's take those 2 equivalent statements: `$test_var = array_get($_POST, 'test', null)` and `$test_var = _$POST['test'] ?? null`.
> > In the first case, PHPStan doesn't know the return of `array_get` so it can't do any kind of validation on the usage of `$test_var`. If we added `mixed` as the return, PHPStan will only do minimal checks (`mixed` is a super-type for all types) and will have to ask us to cast the narrow type when needed (e.g. when doing a function call). In the second case, PHPStan can infer that the variable is of type `?string` and warn/error accordingly.
> 
> `$_POST` by definition only contains strings so let's broaden it up a bit. If I have `$test_var = $some_random_array[$some_var] ?? 42` is PHPStan able to infer the value type of that element of that array from other code introspection? Because if not it's no more useful than `array_get($some_random_array, $some_var, 42)` and that's what I'm trying to tease apart.

Yes, it should be able to with a caveat that it needs to know the narrow type of `$some_random_array` somehow. Unfortunately I think those will have to come through PHPStan annotations (like `@param array<string, int|string>` or `@return array<int, ?Round>`).

cpeel:
@jchaffraix - are you intending on splitting this up into pieces to review separately or would you like help to do that? There's no rush, I just realized it wasn't clear who was taking point on next steps.

cpeel:
> > @jchaffraix - are you intending on splitting this up into pieces to review separately or would you like help to do that? There's no rush, I just realized it wasn't clear who was taking point on next steps.
> 
> @cpeel Thanks for the ping! I have been distracted lately (thus the slow updates) but I have it on my plate to split this. I will make time this week to handle it.

Sound good! There really is no rush, I just didn't want you waiting on me and me not realizing it. 😄

cpeel:
Closing this as it has been implemented in 7 separate PRs, the last of which just merged.

Thank you @jchaffraix for all of this work!

PR #1357, opened 2024-10-12
Check for corrupt images at project load

cpeel:
Check for corrupt images at project load -- both page images and illustrations. We've found that there are some warnings reported by the image check tools that don't prevent the image from loading. We special-case these as warnings, not errors, in add projects.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/validate-images-add-files/

### DIFF
--- a/pinc/ImageUtils.inc
+++ b/pinc/ImageUtils.inc
@@ -5,13 +5,25 @@ use Symfony\Component\Process\Process;
 class ImageUtils
 {
     public const IMAGE_OK = 1;
-    public const IMAGE_CORRUPT = 2;
-    public const IMAGE_SKIPPED = 3;
+    public const IMAGE_WARNING = 2;
+    public const IMAGE_CORRUPT = 3;
+    public const IMAGE_SKIPPED = 4;
 
     /** @var array<string, bool> */
     protected array $external_tools = [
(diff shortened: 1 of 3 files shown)


pinc/ImageUtils.inc
```diff
@@ -5,15 +5,27 @@ use Symfony\Component\Process\Process;
...
+    // These strings are returned from the validation tools but don't
+    // prevent the image from being loaded/used. Let's treat these substrings
+    // as warnings, not errors.
+    /** @var string[] */
+    public array $warning_strings = [
+        "illegal (unless recently approved) unknown, public chunk eXIf",
+        "invalid gAMA value (0.0000)",
+        "invalid tIME year (1969)",
+        "tEXt text contains NULL character(s)",
+    ];
```
srjfoo, line 27: I pulled an image for each of the last two for one of my test loads. They were identified as errors, not warnings, and wouldn't load.

cpeel: Ah! Yes, I had the `stripos()` params backwards. Good catch!

srjfoo (approved):

chrismiceli (approved):

70ray (approved):

PR #1358, opened 2024-10-13
Remove exec() from forum_interface_phpbb3.inc

mrducky4:
Continuing Issue #1274 , replace exec() with Symfony Process() in forum_interface_phpbb3.inc.
The only test I did was click "Start a discussion about this project" and verify it creates a new forum topic.

_Update from cpeel_: Sandbox at https://www.pgdp.org/~cpeel/c.branch/remove_exec_forum_interface/


pinc/forum_interface_phpbb3.inc
```diff
@@ -776,47 +778,29 @@ function call_phpbb_function($func_name, $args)
...
-
-    $output = [];
-    $last_line = exec($command, $output, $exit_status);
-
-    if ($exit_status != 0 || $DEBUG) { /** @phpstan-ignore-line */
+    $cwd_saved = getcwd();
+    chdir("$code_dir/pinc");
+    $process = new Process(array_merge([$php_cli_executable, "-f", $phpbb_include_file, $func_name], $args));
+    $process->run();
+    chdir($cwd_saved);
```
mrducky4, line 789: I'm not sure if restoring the cwd is necessary.

mrducky4:
I'm not sure what to do about the phpstan error for `$process->getExitCode()`. I used it before, in `crontab/CleanUploadsTrash.inc`. But I think I had to look in the Symfony source code to find it, because I couldn't find it in the [documentation](https://symfony.com/doc/current/components/process.html).

One option is just remove it from this file, but it seems like getting the exit code is a common thing to do after running a process, so maybe there is another way to do it which I haven't found yet.

pinc/forum_interface_phpbb3.inc
```diff
@@ -776,47 +778,29 @@ function call_phpbb_function($func_name, $args)
         return;
     }
 
-    $escaped_args_str = implode(' ', array_map('my_escapeshellarg', $args));
-
     // (cd into the pinc directory, so the script doesn't
     // have to guess where it's being invoked from.)
     global $code_dir, $php_cli_executable;
```
cpeel, line 783: This is the only file that uses `$php_cli_executable;` and Symfony has [PhpSubprocess](https://symfony.com/doc/current/components/process.html#executing-a-php-child-process-with-the-same-configuration) that automatically finds the PHP executable. It would be nice to see if that works for us and if so remove `$php_cli_executable` from here, `site_vars.php.template`, and the whole section in `configuration.sh` that tries to find the executable.

chrismiceli (approved):

pinc/forum_interface_phpbb3.inc
```diff
@@ -776,47 +778,29 @@ function call_phpbb_function($func_name, $args)
...
+    if (!$process->isSuccessful() || $DEBUG) { /** @phpstan-ignore-line */
         echo "<pre>\n";
-        echo "$func_name command exited with status=$exit_status\n";
-        if (count($output) > 0) {
-            echo "Here's its output:\n";
-            foreach ($output as $line) {
-                echo "$line\n";
-            }
-        }
+        echo "$func_name command exited with status=$process->getExitCode()\n";
```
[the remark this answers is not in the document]
mrducky4: Ah, that was a silly mistake. I'm still thinking in javascript.

pinc/forum_interface_phpbb3.inc
mrducky4: Good idea, I'll try it.

mrducky4: We are using symfony/process version 5.4, which does not include the PhpSubprocess class. It looks like this was introduced in version 6. But version 6 requires php 8. My dev VM has php 7.4.3, is that the version I should have?

mrducky4:
I updated this to change the comment, simplify setting cwd, and fixing usage of the exit code. I did not try to upgrade Symfony to use `PhpSubprocess()`.  It looks like the related class `PhpProcess` is not suitable for this.

pinc/forum_interface_phpbb3.inc
cpeel: Ah, I missed that. Something for us to consider down the road when we're on PHP 8. Your VM has the right version of PHP on it 👍

cpeel (approved):

70ray (approved):

srjfoo (approved):

PR #1360, opened 2024-10-15
Remove remaining shell calls

### LINKED ISSUE #1274: Convert system/exec usages to Symfony's Process class

cpeel:
This removes the remainder of the exec/shell/system calls in the code and closes #1274. Updated script areas:
* translation center
* project delete
* AV scans on file uploads
* stealth scanno WordCheck tool

Sandbox: https://www.pgdp.org/~cpeel/c.branch/remove-remaining-shell/

### DIFF
--- a/pinc/POFile.inc
+++ b/pinc/POFile.inc
@@ -1,5 +1,7 @@
 <?php
 
+use Symfony\Component\Process\Process;
+
 /**
  * @property-read bool $exists
  * @property-read int|bool $last_modified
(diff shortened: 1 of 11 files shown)


chrismiceli (approved):

70ray (approved):

pinc/POFile.inc
```diff
@@ -126,76 +128,68 @@ class POFile
...
+            "-c", "TRANSLATORS",
             ".",
-            "2>&1",
-            ">",
-            escapeshellarg($pot_filename),
-        ]);
-        exec($cmd, $exec_out, $return_value);
-        if ($return_value != 0) {
-            throw new Exception(implode("\n", $exec_out));
+            $pot_filename,
```
mrducky4, line 141: Does xtext use this as an output file name, so you don't have to use output redirection?

pinc/Project.inc
```diff
@@ -647,7 +649,11 @@ class Project
         validate_projectID($this->projectid);
 
         if (is_dir($this->dir)) {
-            exec("rm -rf " . escapeshellarg($this->dir));
+            $process = new Process(["rm", "-rf", $this->dir]);
+            $process->run();
+            if (!$process->isSuccessful()) {
+                throw new RuntimeException(sprintf(_("Unable to delete %s"), $this->dir));
```
mrducky4, line 655: Nice that this is the same text string as in archiving.inc

mrducky4 (approved):

pinc/POFile.inc
cpeel: Oh, no, I messed this one up. Thanks for catching that!

cpeel: Fixed this and made the new logic nicer in case `xtext` fails. Requires jumping through a few more hoops but its worth it.

chrismiceli (approved):

pinc/POFile.inc
```diff
@@ -126,76 +128,85 @@ class POFile
 
     public function create_template($basedir, $pot_filename)
     {
-        $cmd = join(" ", [
-            // change into the base directory so the filenames in the POT
-            // don't have it prefixed to them
-            "cd " . escapeshellarg($basedir),
-            "&&",
+        // create the new template in a temporary file in case xtext fails
```
srjfoo, line 131: Perhaps I'm misunderstanding what I'm supposed to be testing, so I figured I'd better ask. Comparing against the same action on TEST, I regenerated the POT. Two things I noticed:
* It took a  _**lot**_ longer for the template to be regenerated in the [remove-remaining-shell/](https://www.pgdp.org/~cpeel/c.branch/remove-remaining-shell/) sandbox.
* The "last modified" timestamp was exactly the same. I just tried regenerating it again in the sandbox, and got the same results (It's giving me the message that xtext ran successfully).
* I switched back to TEST, and got an updated timestamp; switched back to the sandbox, and the timestamp matched TEST, once again. Same extra-long wait time for the template to be regenerated, same message saying that the file was successfully generated.

I only downloaded and compared the POT files for the first time around -- as expected, they matched.

I still need to do other testing, but it's time for me to stop for the night, and I wanted to go ahead and mention this.

cpeel: This is my mistake. I neglected to update the sandbox after fixing the logic error @mrducky4 found in the PR. I've updated it now. `POFile.inc` was the only file changed in that update.

srjfoo (approved):

PR #1361, opened 2024-10-15
Type DPage.inc

jchaffraix:
TEST=Manipulated a project (adding pages, doing some P1 reviews, transitioning).

_Update from cpeel:_ Sandbox at https://www.pgdp.org/~cpeel/c.branch/julien_type_DPage/

### DIFF
--- a/pinc/DPage.inc
+++ b/pinc/DPage.inc
@@ -64,7 +64,7 @@ function project_drop_pages(string $projectid): void
 
 // -----------------------------------------------------------------------------
 
-function Pages_prepForRound(string $projectid, $round_number): void
+function Pages_prepForRound(string $projectid, int $round_number): void
 {
     validate_projectID($projectid);
     $round = get_Round_for_round_number($round_number);
@@ -382,7 +382,7 @@ function Page_reopen(string $projectid, string $image, Round $round, string $use
 
 // -----------------------------------------------------------------------------
 
-function Page_returnToRound(string $projectid, string $image, Round $round, string $user)
+function Page_returnToRound(string $projectid, string $image, Round $round, string $user): string
 {
     _Page_require(
         $projectid,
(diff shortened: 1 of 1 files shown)


pinc/DPage.inc
```diff
@@ -64,7 +64,7 @@ function project_drop_pages($projectid)
 
 // -----------------------------------------------------------------------------
 
-function Pages_prepForRound($projectid, $round_number)
+function Pages_prepForRound(string $projectid, $round_number): void
```
cpeel, line 67:
```suggestion
function Pages_prepForRound(string $projectid, int $round_number): void
```

pinc/DPage.inc
```diff
@@ -382,7 +382,7 @@ function Page_reopen($projectid, $image, $round, $user)
 
 // -----------------------------------------------------------------------------
 
-function Page_returnToRound($projectid, $image, $round, $user)
+function Page_returnToRound(string $projectid, string $image, Round $round, string $user)
```
cpeel, line 385:
```suggestion
function Page_returnToRound(string $projectid, string $image, Round $round, string $user): string
```

chrismiceli (approved):

cpeel (approved):

70ray (approved):

pinc/DPage.inc
```diff
@@ -816,13 +822,14 @@ function _project_adjust_n_pages($projectid, $adjustment)
         SET n_pages = %s
```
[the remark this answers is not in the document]
jchaffraix: This is done so we can let MySQL do the arithmetic operation for us. For example, one of the caller passes "+1"  for `$adjustment`. This will translate to the following valid SQL statement:

```
UPDATE projects SET n_pages = n_pages + 1 where projectid=...
```

If we didn't do this, we would need to do a SELECT (to get the current value of `n_pages`), calculate the new value then sent the UPDATE.... but this would open us to a race condition if there are concurrent edits on the same row. The solution would be to lock the row during the read-modify-write but I don't think that's worth it as it brings its own issues.

PR #1363, opened 2024-10-15
Remove onclick from project.php

70ray:
(The js file is out of sync in the 3 commits but there didn't seem any point in re-doing it as it ends up correct)

The boomark button is now black and the cursor does not change over it. It would be possible to add some effect to indicate that it does something when clicked.

### DIFF
--- a/project.php
+++ b/project.php
@@ -600,7 +600,7 @@ function do_project_info_table(): void
             if ($topic_id == "") {
                 $blurb = html_safe(_("Start a discussion about this project"));
                 $url = "$code_url/tools/proofers/project_topic.php?project=$projectid";
-                echo_row_a(_("Forum"), "<a href='$url' onclick='this.outerHTML=\"$blurb\";'>$blurb</a>");
+                echo_row_a(_("Forum"), "<a href='$url' id='no_topic_yet'>$blurb</a>");
             } else {
                 $details = get_topic_details($topic_id);
                 if ($details) {
--- a/project.php
+++ b/project.php
@@ -68,11 +68,16 @@ $ld_json_object = [
     "genre" => $project->genre,
 ];
 
+$js_files = ["$code_url/scripts/project.js"];
+if ($detail_level >= 4) {
+    $js_files[] = "$code_url/scripts/page_table.js";
+}
+
 $extra_args = [
     "head_data" => '<script type="application/ld+json">' .
         json_encode($ld_json_object, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES) .
     '</script>',
-    "js_files" => $detail_level >= 4 ? ["$code_url/scripts/page_table.js"] : [],
+    "js_files" => $js_files,
 ];
 
 output_header($title_for_theme, NO_STATSBAR, $extra_args);
@@ -2028,22 +2033,15 @@ function do_change_state(): void
             $text_next_to_btn = "<span class='transition-disabled'>$text_next_to_btn</span> [$reason]";
         }
 
-        echo "<form method='POST' action='$code_url/tools/changestate.php'>";
+        $question = attr_safe($transition->confirmation_question);
+        echo "<form method='POST' class='frm_chg_stt' data-confirmation-question='$question' action='$code_url/tools/changestate.php'>";
         echo "<input type='hidden' name='projectid'  value='{$project->projectid}'>\n";
         echo "<input type='hidden' name='curr_state' value='{$project->state}'>\n";
         echo "<input type='hidden' name='next_state' value='{$transition->to_state}'>\n";
         echo "<input type='hidden' name='confirmed'  value='yes'>\n";
         echo "<input type='hidden' name='return_uri' value='", attr_safe($here), "'>\n";
 
-        $question = $transition->confirmation_question;
-        if (is_null($question)) {
-            $onClick_condition = "return true;";
-        } else {
-            $onClick_condition = "return confirm(\""
-                . javascript_safe($question) . "\");";
-        }
-        $onclick_attr = "onClick='$onClick_condition'";
-        echo "<input type='submit' value='", attr_safe($transition->action_name), "' $onclick_attr $optional_btn_attr>";
+        echo "<input type='submit' value='", attr_safe($transition->action_name), "'>";
         echo $text_next_to_btn;
 
         echo "</form>\n";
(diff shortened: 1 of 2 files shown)


project.php
```diff
@@ -190,13 +195,11 @@ function get_bookmark_form(): string
...
         <input type='hidden' name='bookmark' value='$bookmark_value'>
-        <a
-            href='#'
-            onClick='this.closest("form").submit();'
+        <button
             aria-label='$label'
             title='$tooltip'
-            style='font-size: 60%'
-        ><i class='$icon' aria-hidden='true'></i></a>
+            style='font-size: 60%; border: none; background: none;'
```
cpeel, line 201: I don't know how much of this is that it's different, but I like that the current version looks like a link (color, and cursor change) to indicate that it's something users can interact with. I don't think it has to be a link, but the link-like behavior seems advantageous for discoverability.

I'm curious what other reviewers think.

scripts/project.js
```diff
@@ -0,0 +1,24 @@
...
+
+    function confirmSubmit(event) {
+        if (this.dataset.confirmationQuestion && !confirm(this.dataset.confirmationQuestion)) {
+            event.preventDefault();
+        }
+    }
+
+    for (const changeStateForm of changeStateForms) {
+        changeStateForm.addEventListener("submit", confirmSubmit);
+    }
```
cpeel, line 12: Can we group the variables with the loops/ifs -- I think this makes it easier to read.
```suggestion
    function confirmSubmit(event) {
        if (this.dataset.confirmationQuestion && !confirm(this.dataset.confirmationQuestion)) {
            event.preventDefault();
        }
    }

    const changeStateForms = document.getElementsByClassName("frm_chg_stt");
    for (const changeStateForm of changeStateForms) {
        changeStateForm.addEventListener("submit", confirmSubmit);
    }
```

Ditto the block below with `newTopicLink`.

project.php
70ray: We can make the cursor change to a pointing hand over it and change its colour.

chrismiceli: Yeah, we need something. I would never have clicked it unless I knew I was supposed to.

chrismiceli (approved):

scripts/project.js
70ray: Done this.

70ray:
Cursor over bookmark changed to pointer.
Sandbox updated.

chrismiceli (approved):

project.php
cpeel: I think we missed changing the color -- I'd make it `@link-color`. I'm not worried about it always having the same color even after clicked, but I think it needs to be more obvious that it's something you can click.

70ray: We could also make the backgound change colour when the cursor approaches it like Slack does. I'll do that as well then we can choose which options look best.

70ray:
Added blue colour and hover background.
The background has a rather drastic effect in the charcoal theme but we could easily fix it by using the @form-text-background colour and defining this also for the other themes.

Sandbox updated.

chrismiceli (approved):

project.php
```diff
@@ -2028,22 +2031,15 @@ function do_change_state(): void
             $text_next_to_btn = "<span class='transition-disabled'>$text_next_to_btn</span> [$reason]";
         }
 
-        echo "<form method='POST' action='$code_url/tools/changestate.php'>";
+        $question = attr_safe($transition->confirmation_question);
+        echo "<form method='POST' class='frm_chg_stt' data-confirmation-question='$question' action='$code_url/tools/changestate.php'>";
```
cpeel, line 2035: I'd prefer the more verbose wording of the class -- a few extra characters to be clearer won't hurt.
```suggestion
        echo "<form method='POST' class='form_change_state' data-confirmation-question='$question' action='$code_url/tools/changestate.php'>";
```

scripts/project.js
```diff
@@ -0,0 +1,22 @@
+window.addEventListener("DOMContentLoaded", function () {
+    function confirmSubmit(event) {
+        if (this.dataset.confirmationQuestion && !confirm(this.dataset.confirmationQuestion)) {
+            event.preventDefault();
+        }
+    }
+
+    const changeStateForms = document.getElementsByClassName("frm_chg_stt");
```
cpeel, line 8:
```suggestion
    const changeStateForms = document.getElementsByClassName("form_change_state");
```

styles/layout.less
```diff
@@ -117,6 +117,18 @@ span.button {
...
 
+.icon_button {
+    font-size: 60%;
+    border: none;
+    background: none;
+    cursor: pointer;
+    color: @link-color;
+}
+
+.icon_button:hover {
```
cpeel, line 128: I kinda like this but also wonder if it would just be a snowflake and that it creates an implicit dependency for a theme between the link color-and the table-cell-base-lowc color.

@chrismiceli / @srjfoo - what do y'all think?

cpeel (approved):
Couple of minor comments but overall this LGTM.

styles/layout.less
chrismiceli: I guess it depends on the themes. Does this work in all themes for good contrast and is it limiting us? I could see maybe splitting it up in a future theme? Maybe for now it is ok?

However, looking throughout the site, it is weird. There isn't much hover so it feels strange that we need something new for this. Would it be better to make it a normal button like most of the other icon buttons?
image

project.php
70ray: Done this.

styles/layout.less
cpeel: > Would it be better to make it a normal button like most of the other icon buttons?

I think the only reason it needs all this complexity is because we want it to POST instead of GET so the "bookmark this page" URL isn't bookmarkable -- if we could use `<a>` to do a POST not a GET we wouldn't need the JS stuff at all. But alas, here we are. Feels like making this act like just another link is _probably_ the most intuitive thing UI-wise?

70ray: It seems to me that the anchor element is really intended to navigate to a new page and its right-click menu has  items "Open link in new tab" etc. If this is not meaningful in the context then the element should not be an anchor but a button of some sort. But we can make it have any appearance we like: a button with text or with an icon, or just an icon or a piece of text with some distinctive style. And we can show that is is clickable by changing the cursor or the appearance of the 'button' or both.

srjfoo: > I kinda like this but also wonder if it would just be a snowflake and that it creates an implicit dependency for a theme between the link color-and the table-cell-base-lowc color.
> 
> @chrismiceli / @srjfoo - what do y'all think?

Works for me!

srjfoo (approved):

PR #1364, opened 2024-10-20
Move change event handler to javascript file

70ray:
Submit form when the 'Number of word context results' selector changes revised.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/wordcontsel

### DIFF
--- a/tools/project_manager/show_word_context.js
+++ b/tools/project_manager/show_word_context.js
@@ -70,4 +70,11 @@ window.addEventListener("DOMContentLoaded", function () {
             showImage(this.dataset.value);
         });
     });
+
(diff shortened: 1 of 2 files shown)


cpeel (approved):

chrismiceli (approved):

srjfoo (approved):
Side issue regarding:
```php
echo "<select id='wordInstancesSelect' name='wordInstances' style='margin-left: 2px;'>";
```

I don't see anything in the Code Best Practices about inline styles, and I admit that every time I see an inline style, I cringe. However, I know that there are appropriate places to use them. Any thoughts on a rule of thumb for this?

chrismiceli (approved):

PR #1365, opened 2024-10-21
Move javascript in word_freq_table into a new file

70ray:
And simplify it using classes instead of identifiers. Use implode to make cutoff links.
Use button instead of link for select and unselct all and cutoff numbers to avoid jump, style button like link and name it lean_button.
Remove parameter $instances from printTableFrequencies(), remove unused checkbox id, remove unused parameter from echo_checkbox_selects()

### DIFF
--- a/scripts/word_freq_table.js
+++ b/scripts/word_freq_table.js
@@ -1,0 +1,50 @@
+window.addEventListener("DOMContentLoaded", function () {
+    function checkAllVisible() {
+        const checkBoxes = document.getElementsByClassName(`cb_${this.dataset.instance}`);
+        for (const checkBox of checkBoxes) {
+            const tBody = checkBox.closest("tbody");
+            if (getComputedStyle(tBody).display !== "none") {
+                checkBox.checked = true;
+            }
+        }
+    }
+
+    const checkAllVisLinks = document.getElementsByClassName("check_all");
+    for (const checkAllVisLink of checkAllVisLinks) {
+        checkAllVisLink.addEventListener("click", checkAllVisible);
+    }
+
+    function unCheckAll() {
+        const checkBoxes = document.getElementsByClassName(`cb_${this.dataset.instance}`);
+        for (const checkBox of checkBoxes) {
+            checkBox.checked = false;
+        }
+    }
+
+    const unCheckAllLinks = document.getElementsByClassName("un_check_all");
+    for (const unCheckAllLink of unCheckAllLinks) {
+        unCheckAllLink.addEventListener("click", unCheckAll);
+    }
+
+    function setCutoff() {
+        const newCutoff = this.dataset.cutoff;
+        const tableBodies = document.getElementsByClassName("table_body");
+        for (tableBody of tableBodies) {
+            if (Number(tableBody.dataset.freqCutoff) < Number(newCutoff)) {
+                tableBody.style.display = "none";
+            } else {
+                tableBody.style.display = "";
+            }
+        }
+        document.getElementById("current_cutoff").innerHTML = newCutoff;
+        // persist cutoff after submit for show_good_word_suggestions.php
+        if (document.getElementById("freqCutoffValue")) {
+            document.getElementById("freqCutoffValue").value = newCutoff;
+        }
+    }
+
+    const cutoffLinks = document.getElementsByClassName("cut_off_link");
+    for (const cutoffLink of cutoffLinks) {
+        cutoffLink.addEventListener("click", setCutoff);
+    }
+});
--- a/tools/project_manager/show_all_good_word_suggestions.php
+++ b/tools/project_manager/show_all_good_word_suggestions.php
@@ -50,8 +50,6 @@ $projects = _get_projects_for_pm($pm);
 
 $submitLabel = _("Add selected words to Good Words List");
 
-// how many instances (ie: frequency sections) are there?
-$instances = count($projects) + 1;
 // what are the cutoff options?
 $cutoffOptions = [1, 2, 3, 4, 5, 10, 25, 50];
 // what is the initial cutoff frequency?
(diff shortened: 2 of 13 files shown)


scripts/word_freq_table.js
```diff
@@ -0,0 +1,50 @@
+window.addEventListener("DOMContentLoaded", function () {
+    function checkAllVisible() {
+        const checkBoxes = document.getElementsByClassName(`cb_${this.dataset.instance}`);
+        for (const checkBox of checkBoxes) {
+            const tBody = checkBox.closest("tbody");
+            if (getComputedStyle(tBody).display !== "none") {
```
chrismiceli, line 6: do we need the computed style or could we use presence of some class or data- attribute instead? From memory computing styles blocks a lot of parallel execution in the browser since it needs css and javascripts to execute fully.

chrismiceli (approved):

scripts/word_freq_table.js
70ray: I'll look into it.

70ray:
Sandbox updated.

scripts/word_freq_table.js
```diff
@@ -0,0 +1,50 @@
+window.addEventListener("DOMContentLoaded", function () {
+    function checkAllVisible() {
+        const checkBoxes = document.getElementsByClassName(`cb_${this.dataset.instance}`);
+        for (const checkBox of checkBoxes) {
+            const tBody = checkBox.closest("tbody");
+            if (tBody.style.display !== "none") {
```
jchaffraix, line 6: Note that `style` returns a `CSSStyleDeclaration` on Chrome (the same return as `getComputedStyle`) so this may still trigger a synchronous style resolution. The reason is that to answer this comparison, the browser has to run the  CSS cascade to know which style applies to the element.

The way around this is to add some "invisible" class. Classes are set in the DOM and don't cascade so they can be queried without forcing a style resolution.

TBH, our pages are fairly static (no animation/transition that would quickly updates the style) so it's not clear if we should be concerned about any synchronous style resolution.

jchaffraix (approved):

chrismiceli (approved):

scripts/word_freq_table.js
70ray: The way it is working seems fast enough in the examples I've tried but there could be thousands of checkboxes with only one word frequency which the user did not want to select so it could be better to first select the tbody elements and then for any that are vissible check its child checkboxes. I'll try that out.

70ray:
When selecting all visible checkboxes, first identify the table, then the visible table bodies in it, then the checkboxes in them. Rather than selecting all the checkboxes with the class of the appropriate table and then for each see if its parent table body is visible.

chrismiceli (approved):

cpeel (commented):
Overall looks good but I haven't tested it yet myself.

scripts/word_freq_table.js
```diff
@@ -0,0 +1,54 @@
+window.addEventListener("DOMContentLoaded", function () {
+    function checkAllVisible() {
+        const table = document.getElementById(`table_${this.dataset.instance}`);
+        const tableBodies = table.getElementsByClassName("table_body");
```
jchaffraix, line 4: FYI the class `table_body` seems pretty synthetic at first glace. I think this is needed because we don't set a `<thead>` for the header row here:

https://github.com/DistributedProofreaders/dproofreaders/blob/0abe4844c429ae04cb8eb27184e56dae6d7b5e0d/tools/project_manager/word_freq_table.inc#L183

Per the [html 5 spec](https://html.spec.whatwg.org/multipage/parsing.html#parsing-main-intable), this means that the browser inserts a `<tbody>` for it to keep the table structure.

Not sure if that's really worth it, but _I think_ we could probably remove the `table_body` class and use `table.getElementsByTagName("tbody")` if we fixed this but curious about your thoughts.

jchaffraix (approved):

tools/project_manager/word_freq_table.inc
```diff
@@ -170,14 +88,14 @@ function getInitialCutoff($idealCutoff, $cutoffOptions, $wordCount = null)
  * @param array $checkbox
  *   Array containing checkbox form elements for each word
  */
-function printTableFrequencies($initialFreq, $cutoffOptions, $wordCount, $instance, $auxData = null, $checkbox = null)
+function printTableFrequencies($initialFreq, $cutoffOptions, $wordCount, $auxData = null, $checkbox = null, $instance = 0)
```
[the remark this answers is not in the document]
70ray: Good point, I've done this.

scripts/word_freq_table.js
70ray: I had wondered if getElementsByTagName("tbody") would be better but was a bit wary of it. The way the table is constructed seems a bit obscure to me so I didn't want to change it. I'll do what you suggest.

70ray:
Sandbox updated.

jchaffraix (approved):

srjfoo (approved):

tools/project_manager/show_all_good_word_suggestions.php
```diff
@@ -61,9 +59,8 @@
     "js_files" => [
         "$code_url/scripts/splitControl.js",
         "$code_url/tools/project_manager/show_all_good_word_suggestions.js",
```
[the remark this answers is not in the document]
70ray: It wouldn't hurt, but there are several other similar cases so we could make a new PR to move them.

cpeel (approved):

cpeel:
@70ray - looks like we have some merge conflicts in the CSS. If you want to resolve those I'll get this merged in. Let me know if you'd like some help.

70ray:
> @70ray - looks like we have some merge conflicts in the CSS. If you want to resolve those I'll get this merged in. Let me know if you'd like some help.

Iv'e resolved the conflicts and tested that the buttons look right and work correctly. I'll look into making another PR later to combine the common parts of the icon_button and lean_button.

PR #1368, opened 2024-10-30
Move javascript from tasks.php to a new file

70ray:
Sandbox at: https://www.pgdp.org/~rp31/c.branch/taskjs

### DIFF
--- a/scripts/tasks.js
+++ b/scripts/tasks.js
@@ -1,0 +1,19 @@
+window.addEventListener("DOMContentLoaded", function () {
+    const meTooButton = document.getElementById("me_too_button");
+    const meTooMainDiv = document.getElementById("MeTooMain");
+    const cancelMeTooButton = document.getElementById("cancel_me_too");
+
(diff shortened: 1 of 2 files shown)


jchaffraix (approved):

srjfoo (approved):

scripts/tasks.js
```diff
@@ -0,0 +1,19 @@
+window.addEventListener("DOMContentLoaded", function () {
+    const meTooButton = document.getElementById("me_too_button");
+    const meTooMainDiv = document.getElementById("MeTooMain");
```
chrismiceli, line 3: very minor, but the casing of the ids are inconsistent. I recognize MeTooMain was first so maybe we align on that style? Really doesn't matter.

chrismiceli (approved):

scripts/tasks.js
70ray: OTOH snake case seems generally more common in php files.

PR #1372, opened 2024-11-14
Move javascript for credit checkboxes to the js file

70ray:
Sandbox at: https://www.pgdp.org/~rp31/c.branch/userpref_js

### DIFF
--- a/userprefs.php
+++ b/userprefs.php
@@ -118,21 +118,6 @@ $theme_extra_args["js_files"] = ["$code_url/scripts/userprefs.js"];
 $theme_extra_args["js_data"] =
     get_newHelpWin_javascript("$code_url/pophelp.php?category=prefs&name=set_") . "
 
-    // function that can be used to check/uncheck a lot
-    // of checkboxes at a time.
-    // First parameter: true/false.
-    // Following parameters: The name of the checkboxes.
-    // The code checks that a checkbox really exists
-    // before accessing it.
-    function check_boxes(value) {
-        for (var i = 1; i < arguments.length; i++) {
-            var elements = document.getElementsByName(arguments[i]);
-            if (elements.length > 0) {
-                elements[0].checked = value;
-            }
-        }
-    }
-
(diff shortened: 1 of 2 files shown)


userprefs.php
```diff
@@ -932,18 +917,18 @@ function _show_credits_wanted_adhoc(): void
     $pm_credit_checked = $userSettings->get_boolean('pm_anonymous') ? '' : 'checked ';
     $pp_credit_checked = $userSettings->get_boolean('pp_anonymous') ? '' : 'checked ';
 
-    echo "<input type='checkbox' name='cp_credit' value='yes' $cp_credit_checked> CP\n";
-    echo "<input type='checkbox' name='ip_credit' value='yes' $ip_credit_checked> IP\n";
-    echo "<input type='checkbox' name='tp_credit' value='yes' $tp_credit_checked> TP\n";
+    echo "<input type='checkbox' name='cp_credit' class='credit_box' value='yes' $cp_credit_checked> CP\n";
```
chrismiceli, line 920: minor suggestion, but instead of adding a class to every checkbox, maybe just tag the table cell and then query all inputs under that cell?
```
tableCell.querySelectorAll('input[type=checkbox]')
```

chrismiceli (approved):

userprefs.php
70ray: Good idea. I'll look into it.

srjfoo: Is @chrismiceli's suggestion still being considered?

70ray: Thanks for reminding me. @chrismiceli's suggestion would make it more compact but looking at structure of userprefs.php, the function _show_credits_wanted_adhoc() is far removed from the table structure so I think it makes it clearer how it is intended to work to leave it like it is. (Leaving aside the unusual way the show_ functions are called, see comment on line 874).

srjfoo (approved):

cpeel (approved):

PR #1374, opened 2024-11-20
[1/n] Remove array_get

jchaffraix:
The function cannot be properly typed as it
is and fixing this would take a while, removing
it is shorter and better.

This PR only covers SETUP/* and api/*.
It also deprecates array_get so users know that
they should migrate.

TESTS=None.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/julien_remove_array_get_1/


chrismiceli (approved):

70ray (approved):

srjfoo (approved):
I haven't done any testing on these, because the testing isn't as straighforward, but all the code changes look good to me.

PR #1379, opened 2024-11-30
Wordcheck API returning bad words with levels

70ray:
Sandbox at: https://www.pgdp.org/~rp31/c.branch/api_wc3

There are two endpoints:
- PUT v1/projects/:projectid/wordcheck This does not require the user to own the page. Needs accepted words and languages. Returns an array of bad words with their levels. 
-  PUT v1/projects/:projectid/pages/:pagename/wordcheck. [edited] This requires the user to own the page and executes save_wordcheck_event(). The logic of when to call this is left to the client. It can be as follows: When the user does any kind of save, if he has previously entered wordcheck mode and a wordcheck report has not previously been submitted then submit a report before the save even if the user has not suggested any words to be accepted. On any subsequent saves submit a report only if the user has suggested some words since the previous report and include only these words. In this way the same word will not be reported more than once for each page.

### DIFF
--- a/SETUP/tests/unittests/ApiTest.php
+++ b/SETUP/tests/unittests/ApiTest.php
@@ -81,7 +81,7 @@ class ApiTest extends ProjectUtils
         return $router->route($path, []);
     }
 
-    protected function wordcheck_report(string $projectid, string $project_state, string $page_name, string $page_state, array $accepted_words = [])
+    protected function wordcheck_report(string $projectid, string $page_name, array $accepted_words = [])
     {
         global $request_body;
 
@@ -89,7 +89,7 @@ class ApiTest extends ProjectUtils
         $request_body = ["accepted_words" => $accepted_words];
         $path = "v1/projects/$projectid/pages/$page_name/wordcheck";
         $router = ApiRouter::get_router();
-        return $router->route($path, ['state' => $project_state, 'pagestate' => $page_state]);
+        return $router->route($path, []);
     }
 
     //---------------------------------------------------------------------------
@@ -748,7 +748,7 @@ class ApiTest extends ProjectUtils
 
         // report wordcheck results
         $accepted_words = ["Snelling", "b[oe]uf"];
-        $this->wordcheck_report($project->projectid, "P1.proj_avail", "001.png", "P1.page_out", $accepted_words);
+        $this->wordcheck_report($project->projectid, "001.png", $accepted_words);
         [$result] = load_wordcheck_events($project->projectid);
         $this->assertEquals("P1", $result[1]);
         $this->assertEquals("001.png", $result[2]);
--- a/SETUP/tests/unittests/ApiTest.php
+++ b/SETUP/tests/unittests/ApiTest.php
@@ -81,6 +81,17 @@ class ApiTest extends ProjectUtils
         return $router->route($path, []);
     }
 
+    protected function wordcheck_report(string $projectid, string $project_state, string $page_name, string $page_state, array $accepted_words = [])
+    {
+        global $request_body;
+
+        $_SERVER["REQUEST_METHOD"] = "PUT";
+        $request_body = ["accepted_words" => $accepted_words];
+        $path = "v1/projects/$projectid/pages/$page_name/wordcheck";
+        $router = ApiRouter::get_router();
+        return $router->route($path, ['state' => $project_state, 'pagestate' => $page_state]);
+    }
+
     //---------------------------------------------------------------------------
     // tests
 
@@ -723,6 +734,26 @@ class ApiTest extends ProjectUtils
             "messages" => [],
         ];
         $this->assertEquals($expected, $response);
+    }
+
+    public function test_wordcheck_report()
+    {
+        global $pguser;
+
+        $pguser = $this->TEST_USERNAME;
+        $project = $this->_create_available_project();
+
+        // check out a page
+        $this->checkout($project->projectid, "P1.proj_avail");
+
+        // report wordcheck results
+        $accepted_words = ["Snelling", "b[oe]uf"];
+        $this->wordcheck_report($project->projectid, "P1.proj_avail", "001.png", "P1.page_out", $accepted_words);
+        [$result] = load_wordcheck_events($project->projectid);
+        $this->assertEquals("P1", $result[1]);
+        $this->assertEquals("001.png", $result[2]);
+        $this->assertEquals("ProjectTest_php", $result[3]);
+        $this->assertEquals($accepted_words, $result[4]);
     }
 }
 
--- a/SETUP/tests/unittests/ApiTest.php
+++ b/SETUP/tests/unittests/ApiTest.php
@@ -70,6 +70,17 @@ class ApiTest extends ProjectUtils
         return $router->route($path, []);
     }
 
+    protected function wordcheck(string $projectid, string $text, $languages = [], $accepted_words = []): array
+    {
+        global $request_body;
+
+        $request_body = ["text" => $text, "languages" => $languages, "acceptedwords" => $accepted_words];
+        $_SERVER["REQUEST_METHOD"] = "PUT";
+        $path = "v1/projects/$projectid/wordcheck";
+        $router = ApiRouter::get_router();
+        return $router->route($path, []);
+    }
+
     //---------------------------------------------------------------------------
     // tests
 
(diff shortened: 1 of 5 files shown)


api/dp-openapi.yaml
```diff
@@ -979,6 +986,61 @@ paths:
...
+        description: text, accepted words and languages
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                text:
+                  type: string
+                acceptedwords:
```
cpeel, line 1006:
```suggestion
                accepted_words:
```

api/dp-openapi.yaml
```diff
@@ -1002,11 +1064,25 @@ paths:
...
+        description: text, +accepted words in report case
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                text:
+                  type: string
+                acceptedwords:
```
cpeel, line 1082:
```suggestion
                accepted_words:
```

SETUP/tests/unittests/ApiTest.php
```diff
@@ -70,6 +70,17 @@ protected function validate_text(string $projectid, string $text): array
         return $router->route($path, []);
     }
 
+    protected function wordcheck(string $projectid, string $text, $languages = [], $accepted_words = []): array
+    {
+        global $request_body;
+
+        $request_body = ["text" => $text, "languages" => $languages, "acceptedwords" => $accepted_words];
```
cpeel, line 77:
```suggestion
        $request_body = ["text" => $text, "languages" => $languages, "accepted_words" => $accepted_words];
```

api/v1_projects.inc
```diff
@@ -848,6 +865,9 @@ function api_v1_project_page(string $method, array $data, array $query_params)
                 return $proof_project_page->save_and_revert(receive_project_text_from_request_body('text'));
             case 'abandon':
                 return $proof_project_page->returnToRound($pguser);
+            case 'report':
```
cpeel, line 868: I don't think this is a good fit for this endpoint. All other actions operate on the page text, perform an action on the page text, and return a page text. Adding a new endpoint specifically for the page's wordcheck would be cleaner and not require the schema juggling, eg:
```
PUT v1/projects/:projectid/pages/:pagename/wordcheck
```

api/dp-openapi.yaml
70ray: done this

api/v1_projects.inc
70ray: Seems reasonable.

70ray:
Sandbox updated

api/dp-openapi.yaml
```diff
@@ -1175,17 +1283,6 @@ components:
       schema:
         type: string
 
-  requestBodies:
```
cpeel, line 1178: Can't we restore this?

api/dp-openapi.yaml
```diff
@@ -1057,6 +1127,44 @@ paths:
...
           $ref: '#/components/responses/UnexpectedError'
 
+  /projects/{projectid}/pages/{pagename}/wordcheck:
+    put:
+      tags:
+      - project
+      description: Report wordcheck reults
+      parameters:
+      - $ref: '#/components/parameters/projectid'
+      - $ref: '#/components/parameters/state_query'
```
cpeel, line 1137: I still don't understand why the caller needs to pass in the project or page state. It either is in a state where the user can make this report -- and the API call passes -- or it isn't and the API call fails.

api/dp-openapi.yaml
70ray: Do you mean just the heading 'requestBodies'?

api/dp-openapi.yaml
70ray: I think this applies to all the proofreading stuff. How it works now is that it first checks that the project and page are in the state the caller thinks they are in and then checks that the action can be carried out in the given states. Should we change it?

cpeel: I know it currently applies to the checking/checkout endpoint and lets not change that in this PR at least but we should remove the requirements from this endpoint.

Think about it this way: why are we asking a function call to validate our idea of the object state before doing something we think we can do. Either we can do it because the object is a state that allows it, or we can't, and the API should enforce that. Any object could change state in an async system while we weren't paying attention. This has never made sense to me in our current codebase.

api/dp-openapi.yaml
cpeel: No, the whole `page_text`. You removed this schema object because you needed to change it in one instance, but we're no longer doing that -- you reverted that change -- so why can't we restore this schema object and its use in the two places?

api/dp-openapi.yaml
70ray: That seems a good way to look at it. I'll change this endpoint.

cpeel:
@70ray - can you please update your sandbox?

cpeel:
The `v1/projects/:projectid/wordcheck` endpoint is looking great.

Some samples for other reviewers.

Example files:
```
cpeel@test:~$ cat data.json
{
    "text": "one two uno four, arid"
}
cpeel@test:~$ cat data_with_language.json
{
    "text": "one two uno four, arid",
    "languages": ["Spanish", "English"]
}
cpeel@test:~$ cat data_with_accepted.json
{
    "text": "one two uno four, arid",
    "accepted_words": ["uno"]
}
```

API calls:
```bash
export API_KEY=review
     -d @data.json \
     -X PUT "https://www.pgdp.org/~rp31/c.branch/api_wc3/api/v1/projects/projectID45c225f598e32/wordcheck"

     -d @data_with_language.json \
     -X PUT "https://www.pgdp.org/~rp31/c.branch/api_wc3/api/v1/projects/projectID45c225f598e32/wordcheck"

     -d @data_with_accepted.json \
     -X PUT "https://www.pgdp.org/~rp31/c.branch/api_wc3/api/v1/projects/projectID45c225f598e32/wordcheck"
```

70ray:
> @70ray - can you please update your sandbox?

Sandbox updated now.

cpeel (approved):
The `v1/projects/{projectid}/pages/{pagename}/wordcheck` also looks good. I tested it with a page I wasn't allowed to change, one that I was but in the wrong state, and one that I was and had checked out.

Sample data files:
```
cpeel@test:~$ cat wc_report.json
{
}
cpeel@test:~$ cat wc_report_accepted.json
{
    "accepted_words": ["uno"]
}
```

And API calls I used -- note that you have to have the page saved as in-progress by the user with the API key for this to work or you'll (correctly) get errors.
```bash
export API_KEY=something
     -d @wc_report.json \
     -X PUT "https://www.pgdp.org/~rp31/c.branch/api_wc3/api/v1/projects/projectID45c225f598e32/pages/123.png/wordcheck"

     -d @wc_report_accepted.json \
     -X PUT "https://www.pgdp.org/~rp31/c.branch/api_wc3/api/v1/projects/projectID45c225f598e32/pages/123.png/wordcheck"
```

srjfoo:
> The `v1/projects/{projectid}/pages/{pagename}/wordcheck` also looks good. I tested it with a page I wasn't allowed to change, one that I was but in the wrong state, and one that I was and had checked out.

I used your examples, customized for myself, but I'm not entirely sure what sort of data I should have in my file, or what results I should be getting on success.

cpeel:
> > The `v1/projects/{projectid}/pages/{pagename}/wordcheck` also looks good. I tested it with a page I wasn't allowed to change, one that I was but in the wrong state, and one that I was and had checked out.
> 
> I used your examples, customized for myself, but I'm not entirely sure what sort of data I should have in my file, or what results I should be getting on success.

This endpoint is inserting records into the `wordcheck_events` table. You can query the table for the most recent events for your user with:
```sql
select * from wordcheck_events where username='srjfoo' order by timestamp desc limit 10;
```

The `accepted_words` in the API call should show up in the `suggestions` column.

srjfoo:
> This endpoint is inserting records into the `wordcheck_events` table. You can query the table for the most recent events for your user with:
> 
> ```sql
> select * from wordcheck_events where username='srjfoo' order by timestamp desc limit 10;
> ```
> 
> The `accepted_words` in the API call should show up in the `suggestions` column.

Thanks, this was exactly what I needed!

srjfoo (approved):

PR #1381, opened 2024-12-01
Add an endpoint for get pickersets

70ray:
We could avoid the expand_pickersets() function by creating the pickersets with character names to start with but then we would have to change some other stuff to use the data in this form.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/api_pickers

### DIFF
--- a/api/v1_projects.inc
+++ b/api/v1_projects.inc
@@ -817,6 +817,19 @@ function api_v1_project_wordcheck($method, $data, array $query_params)
     ];
 }
 
+function api_v1_project_pickersets(string $method, array $data, array $query_params): array
+{
+    $project = $data[":projectid"];
+    $verbose_pickersets = [];
+    foreach ($project->get_pickersets() as $pickerset) {
+        $verbose_pickersets[] = [
+            "name" => $pickerset->name,
+            "subsets" => $pickerset->get_verbose_subsets(),
+        ];
+    }
+    return $verbose_pickersets;
+}
+
 function api_v1_project_page(string $method, array $data, array $query_params)
 {
     global $pguser;
--- a/pinc/CharSuites.inc
+++ b/pinc/CharSuites.inc
@@ -31,6 +31,24 @@ class PickerSet
         return $this->subsets;
     }
 
+    public function get_verbose_subsets(): array
+    {
+        $verbose_subsets = [];
+        foreach ($this->subsets as $subset_name => $codepoint_rows) {
+            $verbose_rows = [];
+            foreach ($codepoint_rows as $row) {
+                $chars = convert_codepoint_ranges_to_characters($row);
+                $char_names = [];
+                foreach ($chars as $char) {
+                    $char_names[] = [$char, (null !== $char) ? utf8_character_name($char) : null];
+                }
+                $verbose_rows[] = $char_names;
+            }
+            $verbose_subsets[] = ["name" => (string)$subset_name, "title" => $this->get_title($subset_name), "rows" => $verbose_rows];
+        }
+        return $verbose_subsets;
+    }
+
     public function get_title(string $name): string
     {
         return $this->titles[$name];
--- a/SETUP/tests/unittests/ApiTest.php
+++ b/SETUP/tests/unittests/ApiTest.php
@@ -753,6 +753,21 @@ class ApiTest extends ProjectUtils
         $this->assertEquals("ProjectTest_php", $result[3]);
         $this->assertEquals($accepted_words, $result[4]);
     }
+
+    public function test_pickersets()
+    {
+        $project = $this->_create_project();
+        $path = "v1/projects/$project->projectid/pickersets";
+        $router = ApiRouter::get_router();
+        $_SERVER["REQUEST_METHOD"] = "GET";
+        $response = $router->route($path, []);
+        $pickerset = $response[0];
+        $this->assertEquals("basic-latin", $pickerset["name"]);
+        $this->assertEquals("!", $pickerset["subsets"][3]["name"]);
+        $this->assertEquals("Punctuation", $pickerset["subsets"][3]["title"]);
+        $this->assertEquals(["!", "EXCLAMATION MARK"], $pickerset["subsets"][3]["rows"][0][0]);
+        $this->assertEquals(["¿", "INVERTED QUESTION MARK"], $pickerset["subsets"][3]["rows"][1][1]);
+    }
 }
 
 // this mocks the function in index.php
--- a/api/dp-openapi.yaml
+++ b/api/dp-openapi.yaml
@@ -1041,6 +1041,63 @@ paths:
         default:
           $ref: '#/components/responses/UnexpectedError'
 
+  /projects/{projectid}/pickersets:
+    get:
+      tags:
+      - project
+      description: Gets pickersets for the project including utf8 character names
+      parameters:
+      - $ref: '#/components/parameters/projectid'
+      responses:
+        200:
+          description: picker data
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  type: object
+                  properties:
+                    name:
+                      type: string
+                      example: basic-latin
+                    subsets:
+                      type: array
+                      items:
+                        type: object
+                        properties:
+                          name:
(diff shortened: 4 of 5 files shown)


api/v1_projects.inc
```diff
@@ -799,6 +799,40 @@ function api_v1_project_validatetext($method, $data, array $query_params)
...
 }
 
+function api_v1_project_pickersets($method, $data, $query_params)
+{
+    // expand the character ranges so that each character can have its name
+    $project = $data[":projectid"];
+    return expand_pickersets($project->get_pickersets());
+}
+
+function expand_pickersets($pickersets)
```
cpeel, line 809: Lets put most of this functionality in the PickerSet class, maybe a `get_verbose_subsets()` (or a better-named) function, and then move the API-specific bits like mapping to an array into `api_v1_project_pickersets()` directly.

70ray: Done something like this.

pinc/CharSuites.inc
```diff
@@ -31,6 +31,33 @@ class PickerSet
...
+                    }
+                }
+                $verbose_rows[] = $char_names;
+            }
+            $verbose_subsets[$subset_name] = $verbose_rows;
+        }
+        return $verbose_subsets;
+    }
+
+    public function get_titles(): array
```
jchaffraix, line 56: Thanks for adding a return type to this function!

To help PHPStan, we should add the precise type when it is easy to infer/simple:

```suggestion
    /** @return string[] */
    public function get_titles(): array
```

jchaffraix (approved):

api/v1_projects.inc
```diff
@@ -799,6 +799,21 @@ function api_v1_project_validatetext($method, $data, array $query_params)
     return $text_checker->analyse(receive_project_text_from_request_body('text'), $data[":projectid"]);
 }
 
+function api_v1_project_pickersets($method, $data, $query_params)
+{
+    $project = $data[":projectid"];
+    $verbose_pickersets = [];
+    $pickersets = $project->get_pickersets();
+    foreach ($pickersets as $pickerset) {
```
cpeel, line 807:
```suggestion
    foreach ($project->get_pickersets() as $pickerset) {
```

api/v1_projects.inc
```diff
@@ -799,6 +799,21 @@ function api_v1_project_validatetext($method, $data, array $query_params)
...
+        $verbose_pickersets[] = [
+            "name" => $pickerset->name,
+            "subsets" => $pickerset->get_verbose_subsets(),
+            "titles" => $pickerset->get_titles(),
+        ];
```
cpeel, line 812: Would the return value for this be more useful to the client if it looked like:
```
$verbose_pickersets[] = [
    "name" => $pickerset->name,
    "subsets" =
        [
            "title" => "string",
            "characters" => [...],
        ],
        [
            "title" => "string2",
            "characters" => [...],
        ],
];
```
Or something similar? That way the client doesn't have to map the titles back to the subsets?

cpeel (commented):
Having this in the PickerSet class makes a lot more sense, thanks for moving it in there.

api/v1_projects.inc
```diff
@@ -799,6 +799,21 @@ function api_v1_project_validatetext($method, $data, array $query_params)
     return $text_checker->analyse(receive_project_text_from_request_body('text'), $data[":projectid"]);
 }
 
+function api_v1_project_pickersets($method, $data, $query_params)
```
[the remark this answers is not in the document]
70ray: Yes indeed.

api/v1_projects.inc
70ray: Yes, thta would be better

pinc/CharSuites.inc
70ray: removed this function now.

jchaffraix (approved):

70ray:
Sandbox updated.

cpeel (approved):
Brilliant.

For other reviewers, this can be tested with:
```
export API_KEY=review
     -X GET "https://www.pgdp.org/~rp31/c.branch/api_pickers/api/v1/projects/projectID45b5cfaa0ff81/pickersets"
```

srjfoo:
> Brilliant.
> 
> For other reviewers, this can be tested with:
> 
> ```
> export API_KEY=review
> curl -i -H "Accept: application/json" -H "X-API-KEY: $API_KEY" \
>      -X GET "https://www.pgdp.org/~rp31/c.branch/api_pickers/api/v1/projects/projectID45b5cfaa0ff81/pickersets"
> ```

Thanks, Casey, that helps immensely.

@70ray, in looking at the output, the thing that stood out to me is that the numbers don't show the actual character. They're also bracketed by square brackets instead of curly, as with the others. I have a memory of you fixing something similar to that in the code -- is that what's going on here? (see screenshot below of picker set labeled `1`.)

image

cpeel:
> @70ray, in looking at the output, the thing that stood out to me is that the numbers don't show the actual character. They're also bracketed by square brackets instead of curly, as with the others. I have a memory of you fixing something similar to that in the code -- is that what's going on here? (see screenshot below of picker set labeled `1`.)

Oh, very good catch Sharon! We need to massage those into a string type instead of an int type. I forget exactly how we do that but you are correct we do that elsewhere.

70ray:
> > @70ray, in looking at the output, the thing that stood out to me is that the numbers don't show the actual character. They're also bracketed by square brackets instead of curly, as with the others. I have a memory of you fixing something similar to that in the code -- is that what's going on here? (see screenshot below of picker set labeled `1`.)
> 
> Oh, very good catch Sharon! We need to massage those into a string type instead of an int type. I forget exactly how we do that but you are correct we do that elsewhere.

An easy work-around would be to return pairs [character, name] rather than an assciative array. This would be just as easy for the client to use. So the result would be:
image

cpeel:
> An easy work-around would be to return pairs [character, name] rather than an assciative array. This would be just as easy for the client to use.

We could, but we're doing something to coerce some array key numbers into strings already if you look at the very top of Sharon's screenshot: 
image

What are we doing that coerces the `1`?

70ray:
> > An easy work-around would be to return pairs [character, name] rather than an assciative array. This would be just as easy for the client to use.
> 
> We could, but we're doing something to coerce some array key numbers into strings already if you look at the very top of Sharon's screenshot: image
> 
> What are we doing that coerces the `1`?

It seems that "1" is because the keys are a mixture or numbers and other characters, but if you use consecutive numbers starting from 0 it gets turned into an ordinary array. But we can coerce it into an object with (object).

70ray:
But it seems an object is not the same sort of thing as an associative array.

70ray:
Test has been adjusted so it works with an object. Is this a good solution or would returning the [char, name] pairs be better? Or if we went back to the state of [93e73a3] I think it would work ok since the keys of the simple array would be the correct numbers and javascript would tun them back into strings.

70ray:
Sandbox updated.

srjfoo (approved):

cpeel:
> Is this a good solution or would returning the [char, name] pairs be better? I like the pairs better.

I think this solution works well. I'm partial to the hash as I think it lets clients easily get the keys via built-in functions. I'm not against the pairs though if you'd rather do that -- both seem acceptable to me. Whichever you prefer is fine.

70ray:
I looked at how this appears when it gets to the javascript. Unfortunately the "1: numbers and math" key appears as an integer so the subsets get in the wrong order. [mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in) says:
"Within each component of the prototype chain, all non-negative integer keys (those that can be array indices) will be traversed first in ascending order by value, then other string keys in ascending chronological order of property creation."
I think the best solution would be to treat the components that we want to keep in the right order as simple arrays. I'll made it like this.

70ray:
Github doesn't seem to be running the checks

70ray:
Conflicts resolved, snadbox updated.

cpeel (approved):
This LGTM!

srjfoo (approved):

PR #1386, opened 2024-12-08
Type pinc/misc.inc

jchaffraix:
This exposed 2 issues:

1. In Page_MarkAsBad, `$reason` was typed as a string but it should have been an int. The function and its callers were updated so we have better typing moving forward.

2. `array_remove_invalid_keys` was not removing any keys if the list of valid keys was empty. Fixed this as this is a potential security issue (returning unexpected data in our API) and added a test for this case.

TESTS=Uploaded several files, marked a page as bad, proofread a page.

Sandbox at: https://www.pgdp.org/~cpeel/c.branch/julien_type_misc/

### DIFF
--- a/SETUP/tests/unittests/MiscUtilsTest.php
+++ b/SETUP/tests/unittests/MiscUtilsTest.php
@@ -52,13 +52,13 @@ class MiscUtilsTest extends PHPUnit\Framework\TestCase
     {
         $array = ["f1" => false, "f2" => true];
         array_remove_invalid_fields($array, []);
-        $this->assertEquals($array, []);
+        $this->assertEquals([], $array);
     }
 
     public function test_remove_no_invalid_fields()
     {
         $array = ["f1" => false, "f2" => true];
         array_remove_invalid_fields($array, null);
-        $this->assertEquals($array, ["f1" => false, "f2" => true]);
+        $this->assertEquals(["f1" => false, "f2" => true], $array);
     }
 }
--- a/SETUP/tests/unittests/MiscUtilsTest.php
+++ b/SETUP/tests/unittests/MiscUtilsTest.php
@@ -47,4 +47,18 @@ class MiscUtilsTest extends PHPUnit\Framework\TestCase
         $this->assertTrue(any(explode(",", "6502,Z80"), fn ($cpu) => is_numeric($cpu)));
         $this->assertFalse(all(explode(",", "6502,Z80"), 'is_numeric'));
     }
+
+    public function test_remove_all_invalid_fields()
+    {
+        $array = ["f1" => false, "f2" => true];
+        array_remove_invalid_fields($array, []);
+        $this->assertEquals($array, []);
+    }
+
+    public function test_remove_no_invalid_fields()
+    {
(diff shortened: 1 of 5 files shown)


pinc/misc.inc
```diff
@@ -72,10 +77,12 @@ function array_get_as_array($array, $key, $default)
...
  *
  * Note: This is an in-place operation that mutates `$array`!
+ *
+ * @param string[] $valid_fields
  **/
-function array_remove_invalid_fields(&$array, $valid_fields)
+function array_remove_invalid_fields(array &$array, ?array $valid_fields): void
 {
-    if ($valid_fields) {
+    if (!is_null($valid_fields)) {
```
jchaffraix, line 85: @70ray FYI this is a slight change in behavior as we will remove all keys if `$valid_fields` is set to `[]`. This seems reasonable to avoid returning unwanted keys but let me know what you think.

cpeel: Your change reflects the desired behavior, This makes sense to me.

pinc/misc.inc
```diff
@@ -947,7 +962,7 @@ function set_col_str($colname, $s)
  *
  * If `$i` is non-numeric, an `InvalidArgumentException` is thrown.
  */
-function set_col_num($colname, $i)
+function set_col_num(string $colname, int $i)
```
cpeel, line 965:
```suggestion
function set_col_num(string $colname, int $i): string
```

pinc/misc.inc
```diff
@@ -1347,7 +1362,7 @@ function get_upload_err_msg($upload_error_code)
 /**
  * Represent bytes in the largest reasonable units
  */
-function humanize_bytes($n)
+function humanize_bytes(float $n): string
```
cpeel, line 1365: Is `float` a superset of `int`? Because in project.php we call it with an `int`:
```php
function echo_byte_size(int $size): void
{
    echo " (", humanize_bytes($size), ")";
}
```

jchaffraix: `float` is not a superset of `int` from the type system's perspective. However `int` can be promoted to a `float` without loss.

The change is done because we do a float division on l. 1371 that involves `$n` (and we return potentially 2 decimal point) so it makes sense to use the wider type in the signature.

SETUP/tests/unittests/MiscUtilsTest.php
```diff
@@ -47,4 +47,18 @@ public function test_any_all()
         $this->assertTrue(any(explode(",", "6502,Z80"), fn ($cpu) => is_numeric($cpu)));
         $this->assertFalse(all(explode(",", "6502,Z80"), 'is_numeric'));
     }
+
+    public function test_remove_all_invalid_fields()
+    {
+        $array = ["f1" => false, "f2" => true];
+        array_remove_invalid_fields($array, []);
+        $this->assertEquals($array, []);
```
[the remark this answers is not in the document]
jchaffraix: Fixed, thanks for pointing this out!

70ray (approved):

cpeel (approved):

srjfoo (approved):

PR #1389, opened 2024-12-11
Type tasks.php

jchaffraix:
This uncovered a few float -> int conversions
that were fixed using get_integer_param, which
allowed some simplification in argument checks.

TESTS=Created a task and manipulated it:
add/remove related task/topic, close, add "me too"

Sandbox at: https://www.pgdp.org/~cpeel/c.branch/julien_type_tasks/

### DIFF
--- a/tasks.php
+++ b/tasks.php
@@ -901,7 +901,11 @@ function process_related_task(object $pre_task, string $action, int $related_tas
 {
     global $pguser, $tasks_url;
     assert($action == 'add' || $action == 'remove');
-    assert($related_task_id >= 1);
+
+    if ($related_task_id < 1) {
+        ShowError(_("You must supply a valid related task ID."), true);
+        return;
+    }
 
     $adding = ($action == 'add');
     $pre_task_id = $pre_task->task_id;
@@ -934,7 +938,11 @@ function process_related_topic(object $pre_task, string $action, int $related_to
 {
     global $pguser, $tasks_url;
     assert($action == 'add' || $action == 'remove');
-    assert($related_topic_id >= 1);
+
+    if ($related_topic_id < 1) {
+        ShowError(_("You must supply a valid related task ID."), true);
+        return;
+    }
 
     $adding = ($action == 'add');
     $pre_task_id = $pre_task->task_id;
--- a/tasks.php
+++ b/tasks.php
@@ -322,7 +322,7 @@ $SearchParams_choices = [
  * initializing it with any (valid) value that the current request
  * has supplied for the parameter.
  */
-function SearchParams_echo_controls()
+function SearchParams_echo_controls(): void
 {
     global $SearchParams_choices, $tasks_url;
 
@@ -356,7 +356,7 @@ function SearchParams_echo_controls()
  * implied by the values (if any) supplied for the search parameters
  * by the current request.
  */
-function SearchParams_get_sql_condition($request_params)
+function SearchParams_get_sql_condition(array $request_params): string
 {
     global $testing, $SearchParams_choices;
 
@@ -406,7 +406,7 @@ function SearchParams_get_sql_condition($request_params)
  * that represents (and possibly just reiterates) the values (if any)
  * supplied for the search parameters by the current request.
  */
-function SearchParams_get_url_query_string()
+function SearchParams_get_url_query_string(): string
 {
     global $SearchParams_choices;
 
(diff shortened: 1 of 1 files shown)


tasks.php
```diff
@@ -649,7 +650,7 @@ function handle_action_on_a_specified_task()
         $action = 'show';
     }
 
-    $task_id = (int)get_float_param($_REQUEST, 'task_id', null, 1, null);
+    $task_id = get_integer_param($_REQUEST, 'task_id', null, 1, null);
```
cpeel, line 653: The original was weird and I went back to do a `git blame` and this was me with:
```
commit f43d4da633fdc66a665683d673f401f9d6b02b43
Author: Casey Peel <cpeel@users.noreply.github.com>
Date:   Mon Dec 24 05:19:59 2018 +0000

    Use type=number inputs for numeric form fields

    For a better user experience, use number field types in forms where the
    input is only numeric. This isn't perfect as some browsers will allow
    the user to input an 'e' (expotenetial notation) and floating point
    numbers are also allowed, but it's a step up from the wide-open
    anything-goes that it was prior.
```

So in 2018 when we converted the input field to a numeric some browsers were sending (or allowed?) a float instead of an int? I guess let's evaluate what, if anything, we need to do about this. It could have been that I was just trying to accept as many "valid but unusual" inputs for these fields, but lets confirm that.

My gut says your change is fine though.

tasks.php
```diff
@@ -896,17 +897,11 @@ function handle_action_on_a_specified_task()
...
     global $pguser, $tasks_url;
     assert($action == 'add' || $action == 'remove');
-
-    // Validate task_id. It must be an integer >= 1
-    $related_task_id = trim($related_task_id);
-    if (!is_numeric($related_task_id) || $related_task_id < 1) {
-        ShowError(_("You must supply a valid related task ID."), true);
-        return;
-    }
+    assert($related_task_id >= 1);
```
cpeel, line 904: This isn't going to do what you think it will. Asserts are only valid for development, we can't rely on them for non-development purposes (see https://www.php.net/manual/en/function.assert.php -- `zend.assertions` != 1).

Since we want to validate that the id is >= 1 we have to do this in a non-assert check.

tasks.php
jchaffraix: So I tested and we could still send a float value from the form right now. As you probably found out in 2018,  input allows any float value and doesn't prevent form submission if the `<input type="number">` is invalid.

The top answer on [this StackOverflow thread](https://stackoverflow.com/questions/37043867/how-to-avoid-decimal-values-in-input-type-number) would prevent this: some JS that blocks any non-digit input.

tasks.php
jchaffraix: This makes sense (and yeah, I missed the fact that asserts are disabled unless we set them up in our ini file).

One question here: we already validate that the task id is >= 1 when we call `get_{integer|float}_param`. Is it worth keeping the assert then? How about the extra validation on every user request?

tasks.php
jchaffraix: Sounds good. I haven't added some of the solutions from the thread. We can always follow-up if we find that this is an issue with folks.

tasks.php
```diff
@@ -935,15 +934,13 @@ function process_related_task($pre_task, $action, $related_task_id)
...
 {
     global $pguser, $tasks_url;
     assert($action == 'add' || $action == 'remove');
 
-    // Validate related_topic_id. It must be an integer >= 1
-    $related_topic_id = trim($related_topic_id);
-    if (!is_numeric($related_topic_id) || $related_topic_id < 1) {
-        ShowError(_("You must supply a valid related topic ID."), true);
+    if ($related_topic_id < 1) {
+        ShowError(_("You must supply a valid related task ID."), true);
```
cpeel, line 943:
```suggestion
        ShowError(_("You must supply a valid related topic ID."), true);
```

cpeel (commented):
Code LGTM - going to get a sandbox spun up for us to poke at in just a moment.

tasks.php
srjfoo: I take the attitude that if someone is playing around with trying out things that aren't integers for task IDs, they deserve what they get. (Bearing in mind that I do that sort of thing to try to break various things in testing.)

srjfoo (approved):

70ray (approved):

cpeel (approved):

PR #1392, opened 2024-12-12
Request only project title field in page browser.

70ray:
Previously the field was not specified and all fields were returned.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/browse_title

### DIFF
--- a/scripts/page_browse.js
+++ b/scripts/page_browse.js
@@ -352,7 +352,7 @@ function pageBrowse(params, storageKey, replaceUrl, mentorMode = false, setShowF
     }
 
     getProjectData = function () {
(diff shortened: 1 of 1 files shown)


scripts/page_browse.js
```diff
@@ -352,7 +352,7 @@ function pageBrowse(params, storageKey, replaceUrl, mentorMode = false, setShowF
     }
 
     getProjectData = function () {
```
cpeel, line 354: Should this be renamed `getProjectTitle` since it's no longer returning the whole project data?

70ray: That seems logical.

chrismiceli (approved):

srjfoo (approved):

PR #1394, opened 2024-12-13
Encode array in query parameter

70ray:
This is one of the possible ways of sending an array.

### DIFF
--- a/scripts/api.js
+++ b/scripts/api.js
@@ -14,7 +14,14 @@ function ajax(method, apiUrl, queryParams = {}, data = {}, fetchPromise = fetch)
     let searchParams = new URLSearchParams();
     searchParams.append("url", apiUrl);
     for (const key in queryParams) {
-        searchParams.append(key, queryParams[key]);
+        const value = queryParams[key];
+        if (Array.isArray(value)) {
+            for (const element of value) {
+                searchParams.append(`${key}[]`, element);
+            }
+        } else {
+            searchParams.append(key, value);
+        }
     }
     url.search = searchParams;
     let upperCaseMethod = method.toUpperCase();
--- a/scripts/api.js
+++ b/scripts/api.js
@@ -16,8 +16,8 @@ function ajax(method, apiUrl, queryParams = {}, data = {}, fetchPromise = fetch)
     for (const key in queryParams) {
         const value = queryParams[key];
         if (Array.isArray(value)) {
-            for (const element of value) {
-                searchParams.append(`${key}[]`, element);
+            for (const item of value) {
+                searchParams.append(`${key}[]`, item);
             }
         } else {
             searchParams.append(key, value);
--- a/SETUP/tests/jsTests/ajaxTests.js
+++ b/SETUP/tests/jsTests/ajaxTests.js
@@ -89,4 +89,15 @@ QUnit.module("Ajax test", function () {
 
         return ajax("POST", "myUrl", { querykey: "queryvalue" }, { datakey: "datavalue" }, fetchPromise).then();
     });
+
+    QUnit.test("Check array parameter", function (assert) {
+        function fetchPromise(url, options) {
+            const blob = new Blob([JSON.stringify({ key: "value" })], { type: "application/json" });
+            assert.strictEqual(url.href, "https://www.dummy.org/api/index.php?url=myUrl&key%5B%5D=value1&key%5B%5D=value2");
+            const init = { status: 200, headers: { "content-type": "application/json" } };
+            return Promise.resolve(new Response(blob, init));
+        }
+
+        return ajax("GET", "myUrl", { key: ["value1", "value2"] }, {}, fetchPromise).then();
+    });
 });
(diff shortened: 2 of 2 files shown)


70ray:
> This method already fits into our published API specifications and is my preferred approach.

ok, I'll just change element to item.

70ray:
Sandbox at: https://www.pgdp.org/~rp31/c.branch/ajax_array

cpeel (approved):

scripts/api.js
```diff
@@ -14,7 +14,14 @@ function ajax(method, apiUrl, queryParams = {}, data = {}, fetchPromise = fetch)
     let searchParams = new URLSearchParams();
     searchParams.append("url", apiUrl);
     for (const key in queryParams) {
-        searchParams.append(key, queryParams[key]);
+        const value = queryParams[key];
+        if (Array.isArray(value)) {
+            for (const item of value) {
+                searchParams.append(`${key}[]`, item);
```
jchaffraix, line 20: Adding `[]` seems odd as any key can be repeated in a query and the PHP code has to handle this inconsistency anyway (a single key will be a string but multiples are an array). This also increases the length of URL as we have to encode `[]`.

Maybe it would make sense to have a convention of using plural nouns (e.g. `keys` rather than `key`) when we expect some repeated field?

It's not a huge issue though.

jchaffraix (approved):

chrismiceli (approved):

scripts/api.js
cpeel: These are good points.

The project search API was one of the first one we implemented and the only one currently using this multi-key design. It's been in the code for over 4 years at this point and used by several external callers so we can't make breaking changes in the endpoint (using comma-separated names for keys, etc).

That doesn't mean we can't make non-breaking changes though (adding a new `fields` entry, making a new `v2` endpoint, etc) and use that instead.

I think this PR is the right fix for this immediate problem -- updating our internal ajax code to support the endpoint as it exists. I'm open to other/better ways of handling this in the API but we need to do so in ways that isn't going to break backwards compatibility.

jchaffraix: @cpeel your answer makes a lot of sense in the context of keeping our API backwards compatible.

I was confused with how `field[]` and `fields` would be an issue as they are both new parameters. Digging into our API, I think the concern is about parameters that are _already plurals_. I could only find `languages` but that's enough as the plural-rule-means-array would change how it works, so this is not really an option. Sorry for the noise...

srjfoo:
I realized after bringing it up in Swagger that this really can't be tested in Swagger (since we can't specify multiple fields in Swagger unless they've fixed that since the last time I tried it), but the listing in the "new" editor, which I have never gotten to work, listed some warnings. They may not be relevant, but I thought I'd mention them just in case:

```
          Line      Code              Message
Warning    916    5260101   All other properties in a "$ref" object are ignored
Warning   1539      10023   maxItems has no effect on non arrays
Warning   1538      10025   minItems has no effect on non arrays
```

70ray:
> I realized after bringing it up in Swagger that this really can't be tested in Swagger (since we can't specify multiple fields in Swagger unless they've fixed that since the last time I tried it), but the listing in the "new" editor, which I have never gotten to work, listed some warnings. They may not be relevant, but I thought I'd mention them just in case:
> 
> ```
>           Line      Code              Message
> Warning    916    5260101   All other properties in a "$ref" object are ignored
> Warning   1539      10023   maxItems has no effect on non arrays
> Warning   1538      10025   minItems has no effect on non arrays
> ```

Interesting, I'd never tried the new editor before. These warnings are not related to this MR but we should address them sometime.

srjfoo:
> > I realized after bringing it up in Swagger that this really can't be tested in Swagger (since we can't specify multiple fields in Swagger unless they've fixed that since the last time I tried it), but the listing in the "new" editor, which I have never gotten to work, listed some warnings. They may not be relevant, but I thought I'd mention them just in case:
> > ```
> >           Line      Code              Message
> > Warning    916    5260101   All other properties in a "$ref" object are ignored
> > Warning   1539      10023   maxItems has no effect on non arrays
> > Warning   1538      10025   minItems has no effect on non arrays
> > ```
> 
> Interesting, I'd never tried the new editor before. These warnings are not related to this MR but we should address them sometime.

Works for me.

Another thing I noticed, though this is just a stylistic nit, I think. Should there be a blank line in the following preceding `/queues:`? (I'm assuming that it's not necessary, but all the similar places in the code have them.
image

srjfoo:
> Interesting, I'd never tried the new editor before. These warnings are not related to this MR but we should address them sometime.

I'm probably missing something that's necessary to get it to work, but I've never been able to get the "new" editor to work, even when the same exact yaml file works for the old editor. I periodically try again, to see if something has changed, but mostly just use the old editor when I'm testing in Swagger. I just happened to notice that the yaml file listing showed those warnings at the bottom.

scripts/api.js
cpeel: It wasn't noise and there is no apology needed -- please continue asking questions and poking at code!

srjfoo (approved):

PR #1398, opened 2024-12-15
Add api endpoints for documents

70ray:
This works only with the documents referred to in $external_faq_overrides in pinc/faq.inc but this is sufficient for the present purpose.
Sandbox at: https://www.pgdp.org/~rp31/c.branch/api_docs

### DIFF
--- a/api/dp-openapi.yaml
+++ b/api/dp-openapi.yaml
@@ -1294,11 +1294,11 @@ paths:
     get:
       tags:
       - documents
-      description: Gets an array of document names, if lang_code is given then only those available in that language
+      description: Gets an array of available documents. If language_code is given, only documents available in that language are returned.
       parameters:
-      - name: lang_code
+      - name: language_code
         in: query
-        description: language code
+        description: Two-letter (ISO 639-1) language code.
         schema:
           type: string
       responses:
@@ -1311,21 +1311,26 @@ paths:
                 items:
                   type: string
 
-  /documents/{document_name}:
+  /documents/{document}:
     get:
       tags:
       - documents
-      description: Gets url for the given document in users language if possible, else english
+      description: Gets a URL for the given document in language_code (if specified and available) or English.
       parameters:
-      - name: document_name
+      - name: document
         in: path
-        description: name of document
+        description: filename of the document
         required: true
+        schema:
+          type: string
+      - name: language_code
+        in: query
+        description: Two-letter (ISO 639-1) language code.
         schema:
           type: string
       responses:
         200:
-          description: Url for the document
+          description: URL for the document
           content:
             application/json:
               schema:
--- a/api/dp-openapi.yaml
+++ b/api/dp-openapi.yaml
@@ -1303,7 +1303,7 @@ paths:
           type: string
       responses:
         200:
-          description: Array of document names
+          description: Array of document filenames
           content:
             application/json:
               schema:
--- a/api/dp-openapi.yaml
+++ b/api/dp-openapi.yaml
@@ -1290,6 +1290,49 @@ paths:
               schema:
                 $ref: '#/components/schemas/round_stats'
 
+  /documents:
+    get:
+      tags:
+      - documents
+      description: Gets an array of document names, if lang_code is given then only those available in that language
+      parameters:
+      - name: lang_code
+        in: query
+        description: language code
+        schema:
+          type: string
+      responses:
+        200:
+          description: Array of document names
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  type: string
+
+  /documents/{document_name}:
+    get:
+      tags:
+      - documents
+      description: Gets url for the given document in users language if possible, else english
+      parameters:
+      - name: document_name
+        in: path
+        description: name of document
+        required: true
+        schema:
+          type: string
+      responses:
+        200:
+          description: Url for the document
+          content:
+            application/json:
+              schema:
+                type: string
+        404:
+          $ref: '#/components/responses/NotFound'
+
 components:
   securitySchemes:
     ApiKeyAuth:
--- a/api/v1_docs.inc
+++ b/api/v1_docs.inc
@@ -5,7 +5,7 @@ function api_v1_documents(string $method, array $data, array $query_params): arr
 {
     global $external_faq_overrides;
 
-    $lang_code = $query_params["lang_code"] ?? null;
+    $lang_code = $query_params["language_code"] ?? null;
     if ($lang_code) {
         $docs_with_lang = [];
         foreach ($external_faq_overrides as $doc => $doc_array) {
@@ -21,6 +21,6 @@ function api_v1_documents(string $method, array $data, array $query_params): arr
 
 function api_v1_document(string $method, array $data, array $query_params): string
 {
-    $lang_code = $query_params["lang_code"] ?? "en_US";
+    $lang_code = $query_params["language_code"] ?? "en_US";
     return get_faq_url($data[":document_name"], $lang_code);
 }
--- a/api/v1_docs.inc
+++ b/api/v1_docs.inc
@@ -1,0 +1,25 @@
+<?php
+include_once($relPath.'faq.inc');
+
+function api_v1_documents(string $method, array $data, array $query_params): array
+{
+    global $external_faq_overrides;
+
+    $lang_code = $query_params["lang_code"] ?? null;
+    if($lang_code) {
+        $docs_with_lang = [];
+        foreach($external_faq_overrides as $doc => $doc_array) {
+            if (isset($doc_array[$lang_code])) {
+                $docs_with_lang[] = $doc;
+            }
+        }
+        return $docs_with_lang;
+    } else {
+        return array_keys($external_faq_overrides);
+    }
+}
+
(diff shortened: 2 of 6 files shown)


chrismiceli (approved):

api/dp-openapi.yaml
```diff
@@ -1290,6 +1290,49 @@ paths:
...
+        200:
+          description: Array of document names
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  type: string
+
+  /documents/{document_name}:
```
srjfoo, line 1314: Perhaps it should have been obvious to me, and will be to those who actually use the API to pull the documentation, but I was unsure just what string to use to pull the URL to the documentation when I first started testing. Would it be a good idea to document that the document names should be drawn from the list of documents in the results from the `/documents` query?

Everything else looks good.
---

Specifying a 2-letter language code worked just fine for `/documents`, inasmuch as the list that was returned was limited to only the docs available in that language.

If I had my language pref set for French for `/documents/{document_name}`, `curl` correctly gave me the link to the French version of the proofreading guidelines 🎉 . However, when testing in Swagger, `/documents/{document_name}` does not return the language result--I assume because Swagger doesn't have access to my prefs on TEST.

api/v1_docs.inc
```diff
@@ -0,0 +1,25 @@
...
+                $docs_with_lang[] = $doc;
+            }
+        }
+        return $docs_with_lang;
+    } else {
+        return array_keys($external_faq_overrides);
+    }
+}
+
+function api_v1_document(string $method, array $data, array $query_params): string
```
cpeel, line 22: This function needs to accept a locale so it knows which URL to return. And `get_faq_url()` below _must_ be passed a locale -- either the one specified by the user or `en_US` so that it returns English if none was specified. We should not have generalized endpoints rely on the user's preferences like this one does (and @srjfoo discovered with testing).

api/dp-openapi.yaml
cpeel, line 1314: I would make this `/documents/{document}:` vs `document_name` since we return a filename and not the name of the document itself.

api/dp-openapi.yaml
```diff
@@ -1290,6 +1290,49 @@ paths:
...
+    get:
+      tags:
+      - documents
+      description: Gets an array of document names, if lang_code is given then only those available in that language
+      parameters:
+      - name: lang_code
+        in: query
+        description: language code
+        schema:
+          type: string
```
cpeel, line 1303: This name and description isn't going to be useful, we can help them out:
```suggestion
      - name: language_code
        in: query
        description: Two-letter (ISO 639-1) language code. 
        schema:
          type: string
```

api/dp-openapi.yaml
```diff
@@ -1290,6 +1290,49 @@ paths:
               schema:
                 $ref: '#/components/schemas/round_stats'
 
+  /documents:
+    get:
+      tags:
+      - documents
+      description: Gets an array of document names, if lang_code is given then only those available in that language
```
cpeel, line 1297:
```suggestion
      description: Gets an array of available documents. If language_code is given, only documents available in that language is returned.
```

api/dp-openapi.yaml
```diff
@@ -1290,6 +1290,49 @@ paths:
...
+              schema:
+                type: array
+                items:
+                  type: string
+
+  /documents/{document_name}:
+    get:
+      tags:
+      - documents
+      description: Gets url for the given document in users language if possible, else english
```
cpeel, line 1318:
```suggestion
      description: Gets a URL for the given document in language_code (if specified and available) or English.
```

api/dp-openapi.yaml
```diff
@@ -1290,6 +1290,49 @@ paths:
...
+      parameters:
+      - name: document_name
+        in: path
+        description: name of document
+        required: true
+        schema:
+          type: string
+      responses:
+        200:
+          description: Url for the document
```
cpeel, line 1328:
```suggestion
          description: URL for the document
```

api/v1_validators.inc
```diff
@@ -67,3 +67,13 @@ function validate_release_queue($queueid, $data)
     }
     return $queue_data;
 }
+
+function validate_document_name(string $document_name): string
```
cpeel, line 71: s/document_name/document/ in whole function

api/dp-openapi.yaml
70ray: good point.

api/v1_docs.inc
70ray: Anyhow, I've changed it as you suggested. It also seems rather unsatisfactory that _GET should be referenced inside the function get_desired_language() but with the language code specified we won't ever get there.

api/dp-openapi.yaml
```diff
@@ -1290,6 +1290,54 @@ paths:
...
+      description: Gets an array of available documents. If language_code is given, only documents available in that language are returned.
+      parameters:
+      - name: language_code
+        in: query
+        description: Two-letter (ISO 639-1) language code.
+        schema:
+          type: string
+      responses:
+        200:
+          description: Array of document names
```
cpeel, line 1306:
```suggestion
          description: Array of document filenames
```

cpeel (approved):

jchaffraix (approved):

api/dp-openapi.yaml
70ray: Oops, missed that one.

srjfoo (approved):

srjfoo:
I need to rescind my approval. There's a disconnect between `lang_code` and `language_code`. 

Last time I tested, both `fr` and `de` gave a subset list of the document filenames  for `/documents`. Now, both queries are giving the full list. In the curl query from the command line, I changed the `language_code` that Swagger uses to `lang_code` and got the correct list. And `/documents/{document}` is still giving me the English  result for a non-English 2-letter language code (also for "language code"_"Country Code").

srjfoo (changes requested):
Please see https://github.com/DistributedProofreaders/dproofreaders/pull/1398#issuecomment-2552747727 -- between the last time I tested, and this time, something broke.

70ray:
> I need to rescind my approval. There's a disconnect between `lang_code` and `language_code`.
> 
> Last time I tested, both `fr` and `de` gave a subset list of the document filenames for `/documents`. Now, both queries are giving the full list. In the curl query from the command line, I changed the `language_code` that Swagger uses to `lang_code` and got the correct list. And `/documents/{document}` is still giving me the English result for a non-English 2-letter language code (also for "language code"_"Country Code").

Yes, this is inconsistent, I'll change it. There remains the question of what happens if an invalid language code or a valid but unavailable one is given. Should we throw exceptions with messages like "xx is not a valid language code" or "the document is not available in Welsh" or silently return the English one?

srjfoo:
> Yes, this is inconsistent, I'll change it. There remains the question of what happens if an invalid language code or a valid but unavailable one is given. Should we throw exceptions with messages like "xx is not a valid language code" or "the document is not available in Welsh" or silently return the English one?

Right now, English is returned when it doesn't recognize something. I don't really like that as a default, but I'll defer to @cpeel on that.

cpeel:
> > Yes, this is inconsistent, I'll change it. There remains the question of what happens if an invalid language code or a valid but unavailable one is given. Should we throw exceptions with messages like "xx is not a valid language code" or "the document is not available in Welsh" or silently return the English one?
> 
> Right now, English is returned when it doesn't recognize something. I don't really like that as a default, but I'll defer to @cpeel on that.

Yes, throwing an error is probably the right thing to do in the API. Perhaps a change like this (untested):
```diff
diff --git a/pinc/faq.inc b/pinc/faq.inc
index 739d184cd..b3b73b948 100644
--- a/pinc/faq.inc
+++ b/pinc/faq.inc
@@ -63,7 +63,7 @@ $external_faq_overrides = [
 ];


-function get_faq_url(string $faq, ?string $langcode = null): string
+function get_faq_url(string $faq, ?string $langcode = null, bool $english_fallback = True): string
 {
     global $code_dir, $code_url;

@@ -72,8 +72,13 @@ function get_faq_url(string $faq, ?string $langcode = null): string
         $langcode = substr(get_desired_language(), 0, 2);
     }

+    $check_lang_codes = [$langcode];
+    if ($english_fallback) {
+        array_push($check_lang_codes, "en");
+    }
+
     // first try $langcode, then English
-    foreach ([$langcode, "en"] as $langcode) {
+    foreach ($check_lang_codes as $langcode) {
         // Any overrides?
         $external_faq = get_external_faq($faq, $langcode);
         if ($external_faq) {
@@ -87,7 +92,7 @@ function get_faq_url(string $faq, ?string $langcode = null): string
     }

     // fall back to the English version in the code if it exists
-    if (is_file("$code_dir/faq/$faq")) {
+    if ($english_fallback && is_file("$code_dir/faq/$faq")) {
         return "$code_url/faq/$faq";
     }

```

And having the caller in the API check and if `""` is returned raise a `RuntimeException()`.

70ray:
Changes implemented, Sandbox updated.

cpeel (approved):

srjfoo (approved):

PR #1402, opened 2024-12-19
Type release queues

jchaffraix:
Enforced a few types in pinc/release_queue.inc
that were annotated as it seemed like a low
risk change and better for us.

TEST=Tried the queue API and the stats endpoint.

Sandbox: https://www.pgdp.org/~jchaffraix/c.branch/julien_type_release_queue/

### DIFF
--- a/stats/release_queue.php
+++ b/stats/release_queue.php
@@ -23,8 +23,7 @@ $listing_view_modes = [
 $round = get_round_param($_GET, 'round_id', null, true);
 $name = @$_GET['name'];
 $listing_view_mode = get_enumerated_param($_GET, "show", "populated", array_keys($listing_view_modes));
-// We should probably switch to get_bool_param but that would change our behavior.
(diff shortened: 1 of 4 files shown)


stats/release_queue.php
```diff
@@ -23,7 +23,8 @@
 $round = get_round_param($_GET, 'round_id', null, true);
 $name = @$_GET['name'];
 $listing_view_mode = get_enumerated_param($_GET, "show", "populated", array_keys($listing_view_modes));
-$unheld_only = get_integer_param($_GET, 'unheld_only', 0, 0, 1);
+// We should probably switch to get_bool_param but that would change our behavior.
```
cpeel, line 26: I don't think it's a problem to use `get_bool_param()` here.

jchaffraix: OK, updated.

chrismiceli (approved):

cpeel (approved):

srjfoo (approved):

PR #1404, opened 2024-12-21
Type LPage and users

jchaffraix:
This uncovered an inconstency in our API where we used the result of a void function. Explicitly added a `null` return to keep our current behavior.

TEST=Tested all the proofing buttons (save, mark as bad, ...).

Sandbox: https://www.pgdp.org/~jchaffraix/c.branch/julien_type_LPage_all/

### DIFF
--- a/api/v1_projects.inc
+++ b/api/v1_projects.inc
@@ -866,7 +866,8 @@ function api_v1_project_page(string $method, array $data, array $query_params)
             case 'revert':
                 return $proof_project_page->save_and_revert(receive_project_text_from_request_body('text'));
             case 'abandon':
-                return $proof_project_page->returnToRound($pguser);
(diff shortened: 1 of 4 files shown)


cpeel (approved):

srjfoo (approved):

api/v1_projects.inc
```diff
@@ -866,7 +866,8 @@ function api_v1_project_page(string $method, array $data, array $query_params)
             case 'revert':
                 return $proof_project_page->save_and_revert(receive_project_text_from_request_body('text'));
```
70ray, line 867: (I spotted my earlier mistake here with the text parameter. I'll make a PR to fix it after this so it doesn't conflict.)

70ray (approved):

chrismiceli (approved):

PR #1405, opened 2024-12-21
Add client JSON storage API endpoint

cpeel:
We want the new proofreading interface to be able to persist client configuration details across browsers and devices. This requires that some configuration values are stored server-side. To enable this, allow clients to store opaque JSON blobs on the server, up to one blob per client per user.

Please read the updates to `SETUP/API.md` on important caveats about this endpoint and ensure we're OK with this from a server-side perspective as well as the client side.

This PR introduces a new generalized JsonStorage class and database table with an ApiClientStorage layer on top of that wired into the API. When JSON blobs are updated in the JsonStorage class their timestamp field in the database is also updated. This field isn't currently made available via the code right now but seems like it would be useful if, for instance, we want to clean up client blobs older than some period of time. I'm open to thoughts on this.

`curl` examples to test this:
```bash
export API_KEY=review

# Save a json blob
    -X PUT -d '{"key": 1}' "https://www.pgdp.org/~cpeel/c.branch/client-storage/api/v1/storage/clients/newpi"

# Fetch a json blob
    -X GET "https://www.pgdp.org/~cpeel/c.branch/client-storage/api/v1/storage/clients/newpi"

# Delete a json blob
    -X DELETE "https://www.pgdp.org/~cpeel/c.branch/client-storage/api/v1/storage/clients/newpi"
```

This is likely to have minor merge conflicts with https://github.com/DistributedProofreaders/dproofreaders/pull/1398 and I'll resolve those after that gets merged.


pinc/JsonStorage.inc
```diff
@@ -0,0 +1,91 @@
...
+     */
+    public function set(string $setting, string $value): void
+    {
+        // It's possible $value could be very large and including it twice
+        // in the query will cause it to be too big. The default max query
+        // size is 64MB which means this will fail if the JSON blob is
+        // just shy of 32MB. That seems plenty big.
+        $sql = sprintf(
+            "
+            INSERT INTO json_storage
```
70ray, line 25: Does this amount to the same thing as 'REPLACE'? Sorry if this is a silly question but I have very little experience of databases.

api/v1_storage.inc
```diff
@@ -0,0 +1,31 @@
...
+
+function api_v1_storage_clients(string $method, array $data, array $query_params)
+{
+    global $pguser;
+
+    $clientid = $data[":clientid"];
+
+    $storage = new ApiClientStorage($clientid, $pguser);
+    if ($method == "GET") {
+        return json_decode($storage->get());
```
70ray, line 16: Should not the client decode and enode the JSON data so we are just dealing with strings here?

pinc/ApiClientStorage.inc
```diff
@@ -0,0 +1,38 @@
...
+class ApiClientStorage
+{
+    private JsonStorage $storage;
+    private string $setting;
+
+    public function __construct(string $client, string $username)
+    {
+        global $api_client_storage_keys;
+
+        if (!in_array($client, $api_client_storage_keys)) {
```
70ray, line 12: Would the client not already have been validated before we get here?

70ray (commented):
This looks like a good overall structure, just a few comments about details.

SETUP/API.md
```diff
@@ -66,3 +66,24 @@ Three settings in `configuration.sh` control limiting:
...
 * `_API_RATE_LIMIT_SECONDS_IN_WINDOW` - the number of seconds within a given
   window.
+
+## Client Storage
+
+To facilitate javascript UI clients persisting data across browsers and devices,
+the API includes an optional endpoint for clients to store and fetch JSON blobs.
+To enable this feature, add a string for the client to the
+`_API_CLIENT_STORAGE_KEYS` configuration setting and have the client use that
+string with the endpoint as the `clientid`.
```
jchaffraix, line 76: I am having a hard time understanding the purpose of the API client storage keys. If the end goal is to allow any client to store this configuration down the road (which means that a user could trivially get one of our keys), is this some temporary measure to prevent arbitrary storage from storing JSON in our DB?

api/v1_storage.inc
cpeel: The API automatically converts returns to JSON values so if we don't decode this into an object we stringify it twice.

pinc/JsonStorage.inc
cpeel: It does, yes. This is a MySQL-ism and not part of an SQL standard and we use it pretty frequently in the code.

SETUP/API.md
cpeel: You have the gist of it, but it's not temporary.

Keep in mind that every user has de facto API access via PHP sessions -- that's how the Page Browser users the API and how the new proofreading UI will too. If we don't have some restriction, any logged in user could fill up our server with random JSON blobs. The `clientid` means that each user can have only at most `len(API_CLIENT_STORAGE_KEYS)` JSON blobs and those have a maximum size.

api/dp-openapi.yaml
```diff
@@ -1290,6 +1290,58 @@ paths:
...
+        required: true
+        schema:
+          type: string
+      responses:
+        200:
+          description: JSON blob that was persisted
+          content:
+            application/json:
+              schema:
+                type: object
```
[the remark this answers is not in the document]
cpeel: The timestamp is arguably just as useless as the JSON blob because it's always `now()`. I was uncertain what a good return value would be so I modeled what we do for PUTs for word lists -- and what the product at my job does for PUTs and PATCHes -- which is to return the object that was actually serialized.

Open to ideas / thoughts on what API clients should "expect" from a successful PUT / PATCH / etc beyond the HTTP status code.

pinc/ApiClientStorage.inc
cpeel: It would. I did it here too to ensure it's enforced at the data layer not just the user input layer if this is used elsewhere for some reason.

pinc/ApiClientStorage.inc
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    // -------------------------------------------------------------------------
+
+    public function set(string $value): void
+    {
+        $this->storage->set($this->setting, $value);
+    }
+
+    public function get(): ?string
```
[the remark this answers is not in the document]
cpeel: Thanks for catching this. The function originally returned `null` and I changed it and missed this.

cpeel (commented):
Thanks for the questions and updates. I'll get the (excellent) code suggestions done later today.

SETUP/tests/ci_configuration.sh
```diff
@@ -47,6 +47,8 @@ _API_ENABLED=true
 _API_RATE_LIMIT=false
 _API_RATE_LIMIT_REQUESTS_PER_WINDOW=3600
 _API_RATE_LIMIT_SECONDS_IN_WINDOW=3600
+_API_RATE_LIMIT_SECONDS_IN_WINDOW=3600
+_API_CLIENT_STORAGE_KEYS='[]'
```
cpeel, line 51:
```suggestion
_API_CLIENT_STORAGE_KEYS='[]'
```
@srjfoo caught this - copy/pasta error on my part and I'll get it fixed.

SETUP/API.md
cpeel: Perhaps "client" is the wrong word for this? I am using it in the "JS client that uses the APIs". Would "app" or some other word be clearer?

SETUP/API.md
```diff
@@ -66,3 +66,24 @@ Three settings in `configuration.sh` control limiting:
...
+
+Some important notes about this feature:
+* Client storage is one blob per user per client. Said another way: clients are
+  only able to store one blob per `clientid` and that blob is only for the user
```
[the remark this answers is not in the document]
cpeel: It's both. It's one per `clientid` and by the user that is authenticated. I'll change it slightly to maybe make it clearer.

cpeel:
PR feedback so far incorporated and resolved the merge conflicts with the documents API PR.

api/v1_storage.inc
70ray: Yes, of course, you are right.

api/v1_storage.inc
```diff
@@ -0,0 +1,32 @@
...
+    } elseif ($method == "PUT") {
+        $storage->set(json_encode(api_get_request_body()));
```
70ray, line 19: In this case it is already encoded by the client so we don't want to encode it again.

api/v1_storage.inc
70ray: But I'm still concerned about this. As I've discovered elsewhere, converting between javascript objects and arrays and php arrays and associative arrays via JSON can cause some screw-ups. Let me think about this some more.

api/v1_storage.inc
cpeel: > In fact, given the existing API structure, we do want to encode it twice when returning the string to the client. The client encodes its javascript object into a JSON string and expects to get back the same string.

No, that's not it works.

The Client isn't sending us an opaque string, it's sending us a JSON object. That comes through as a string and the ApiRouter decodes that string into an object -- it does this for all request bodies and it how all of the functions work. Our ApiClientStorage `get()` and `set()`s take strings, so we have to encode that back into a string to store it.

Client sends JSON in request -> API converts to object -> `v1_storage.inc` converts back to string -> ApiClientStorage persists string

When we `get()` it from ApiClientStorage the class returns a string. But returns from API router functions needs to be an object, because the API will encode that into a string, so we need to decode it into an object first.

ApiClientStorage returns string -> `v1_storage.inc` converts back to object -> API converts object to string -> Client receives JSON in return

cpeel: > p.s. I think the api_get_request_body() is a bit confusing because when it says return $json, the thing it is returning generally isn't JSON but what was originally provided as data to the ajax function. (though in my scenario it now is).

Yes, that's totally fair -- `api_get_request_body()` returns a decoded object/array, not a string - `$json_object` would be clearer.

And yes, the client is encoding the object into JSON -- we do that already in the ajax function:
```js
    if (upperCaseMethod !== "GET") {
        // POST or PUT
        options.headers["Content-Type"] = "application/json";
        options.body = JSON.stringify(data);
    }
```
The client just needs to pass in the object it wants persisted as `$data` to `ajax()` and it'll all just work.

SETUP/API.md
cpeel: The JS client (ie: new proofreading interface) will need to figure out which settings it should store in the local browser cache (and only available to that browser -- like window size) and which one it should store on the server (and available to any browser device -- like font family). All this endpoint does is handle saving and returning whatever the JS client sends it. We have to figure out the other parts but that's all done within the JS client.

api/v1_storage.inc
70ray: I see what you mean. But I think there could be some problems with that approach. The original javascript object will get encoded into JSON and will get decoded into a PHP object and then re-encoded into JSON and the reverse on the way back. PHP objects don't correspond exactly to javascript objects. As we have seen, a php associative array which happens to have consecutive numerical keys or is empty is encoded as an ordinary array. What we get back may not be the same as what we started with.
The other approach of having the client make a JSON object with JSONstringify and then sending that (it would get doubly encoded) would, I imagine, be safer since we are only json-encoding and -decoding strings in the server.  v1_storage would then not need to do any encoding or decoding but the client would have to encode and decode so the number of codings would end up much the same overall.

cpeel: If the client wants to send an encoded JSON string as part of a JSON object they're more than welcome to do so (`options.body = {data = JSON.stringify(data)};` or however the JS syntax is) although I think it's a bad idea, interoperability issues or not.

It's unsafe and unwise for us to store completely unstructured blobs which is what you're asking for.

70ray: If what the client sends is not valid JSON then the database will reject it.

cpeel: So somehow bypass the API infra? I'm willing to entertain it if you want to fork my branch and show me what you have in mind.

70ray: No, I don't want to do that. When we save a page from the PI we send a string from the client and the same string gets put into the db. What I'm proposing is sending a JSON string in the same way and this same JSON string gets put into the db as  JSON.

cpeel: The page is a string inside a JSON object -- it's not a naked string. So what you're proposing is having a JSON string as a value of a JSON object:
```json
{
    "data": "{\"key\": \"value\"}"
}
```
And then the API would extract `data` out of the payload and pass the string into the ApiClientStorage class.

That's a weird and unexpected interface. No one does that -- just pass the JSON object 😁

70ray: It's getting late here now. I'll consider it later but have a happy Christmas.

cpeel: I've put up a commit that allows a routed function to get the non-deserialized input and output non-serialized output. This is less hacky than I thought it would be. It's in a separate commit for evaluation / discussion.

70ray: Essentially yes but without the wrapping.

api/ApiRouter.inc
```diff
@@ -55,15 +57,16 @@ class ApiRouter
             throw new MethodNotAllowed();
         }
         $function = $url_map["endpoint"][$method];
-        return $function($method, $data, $query_params);
+        $this->_response = $function($method, $data, $query_params);
```
70ray, line 60: Couldn't you just say:
```php
$response = $function($method, $data, $query_params);
if ($this->_raw_response) {
            return $response;
        } else {
            return json_encode($response, JSON_PRETTY_PRINT |
                JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES);
        }
    }

File: api/ApiRouter.inc [hunk 11]
```diff
@@ -73,7 +76,35 @@ class ApiRouter
         throw new InvalidAPI();
     }
 
-    public static function get_router()
+    public function request(bool $raw = false)
```
70ray, line 79: This function could stay in index.php

api/ApiRouter.inc
```diff
@@ -73,7 +76,35 @@ class ApiRouter
...
+        } else {
+            $json_object = json_decode(file_get_contents('php://input'), true);
+            if ($json_object === null) {
+                throw new InvalidValue("Content was not valid JSON");
+            }
+            return $json_object;
+        }
+    }
+
+    public function response(bool $raw = false): string
```
70ray, line 92: We wouldn't need this if it was incorporated as above.

api/ApiRouter.inc
```diff
@@ -8,8 +8,10 @@ class ApiRouter
 {
     private $_url_map = [];
     private $_validators = [];
+    private $_response;
```
70ray, line 11:
```php
 private $_response;
```
 would not be needed with changes below.

api/index.php
```diff
@@ -39,8 +39,8 @@ function api()
     unset($query_params["url"]);
 
     $router = ApiRouter::get_router();
-
-    api_output_response($router->route($path, $query_params));
+    $router->route($path, $query_params);
```
70ray, line 42: This could remain unchanged.

70ray (commented):
This is great. I think it could be slightly simplified.

SETUP/API.md
jchaffraix: > The clientid means that each user can have only at most len(API_CLIENT_STORAGE_KEYS) JSON blobs and those have a maximum size.

I am not seeing any validation on the number of stored blobs, we only check that the client id is allow-listed but besides that the constraints on the DB would allow unbounded storage. Those are trusted clients so it doesn't seem to be a concern, just curious about this.

> Perhaps "client" is the wrong word for this? I am using it in the "JS client that uses the APIs". Would "app" or some other word be clearer?

Mhh, is the idea to separate our browser users from other "app" that programmatically call our API?

> You have the gist of it, but it's not temporary.

OK, as a side question since we need a logged in users to store the information, would it make sense to use the user rather than an extra key? Or are we thinking that some "apps" will backfill/mutate a lot of users' keys? I am probably missing something rather obvious here as I don't have much context on the new UI.

api/dp-openapi.yaml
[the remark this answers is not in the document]
jchaffraix: > The timestamp is arguably just as useless as the JSON blob because it's always now(). 

Sure as it represents the last modified timestamp from the server's perspective.

However I do think there is a lot more value to returning this compared to the blob that was sent and is known to the client. The timestamp would allow the client to implement some caching (e.g. don't check with the server if the timestamp is fresh enough), lower the bandwidth needed (by return the JSON object on a `get` request if the timestamp doesn't match, useful for mobile devices) or potentially even allow for delta (this would need to switch to some event based logic). Arguably, neither of those scenarios applies here, but those are enabled by returning the timestamp :smile:

> Open to ideas / thoughts on what API clients should "expect" from a successful PUT / PATCH / etc beyond the HTTP status code.

I don't think we need to return anything from the API on success: the return code says if it was successful and we can return an optional error field to return the cause of errors to the client. Those 2 are what a minimal client would expect. We can always add more as we understand our needs more.

jchaffraix (commented):
Sorry for delay in the reply and merry Christmas! :christmas_tree:

api/ApiRouter.inc
cpeel: I moved this here to keep the encoding and decoding together. And because I had to move the encoding here this came along with it.

api/ApiRouter.inc
cpeel: Because I wanted to keep the encoding and decoding together and they got moved in here, breaking them out into clear functions makes it clearer what's going on IMHO.

api/ApiRouter.inc
70ray: That seems reasonable. I've had some further thoughts about this and deleted my other ananswered comments which had overlooked a change you made in index.php. It's difficult to see the whole picture in this github view so I dowloaded the files to look at them more carefully.

70ray (commented):
It seems to me that to receive and return raw data we should really intervene in index.php in the functions api_get_request_body() and api_output_response() rather than make any changes in ApiRouter.
The function api_get_request_body() is easy by adding a 'raw' parameter as you have done but api_output_response() is more difficult. We could set a global variable to tell api_output_response() to send raw data but adding more global variables isn't something we want to do. Perhaps make a static API class to put the variable in. Would this be possible?

cpeel:
> We could set a global variable to tell api_output_response() to send raw data but adding more global variables isn't something we want to do. Perhaps make a static API class to put the variable in. Would this be possible?

That's exactly why I put it in ApiRouter -- to avoid a global variable or adding a new artificial class with static variables. To me it made sense to put it in the router because the router is accepting a request and returning a response.

70ray:
> > We could set a global variable to tell api_output_response() to send raw data but adding more global variables isn't something we want to do. Perhaps make a static API class to put the variable in. Would this be possible?
> 
> That's exactly why I put it in ApiRouter -- to avoid a global variable or adding a new artificial class with static variables. To me it made sense to put it in the router because the router is accepting a request and returning a response.

Fair enough.

api/ApiRouter.inc
```diff
@@ -55,15 +57,16 @@ class ApiRouter
             throw new MethodNotAllowed();
         }
         $function = $url_map["endpoint"][$method];
-        return $function($method, $data, $query_params);
+        $this->_response = $function($method, $data, $query_params);
+        return $this->_response;
```
70ray, line 61: Do you need this line now?

cpeel:
> I am not seeing any validation on the number of stored blobs, we only check that the client id is allow-listed but besides that the constraints on the DB would allow unbounded storage. Those are trusted clients so it doesn't seem to be a concern, just curious about this.

I'm really not doing a good job of describing this which means that I've named things incredibly poorly in the code and I need to figure out how to fix this, my apologies.

Let me zoom out just a little: to use the DP API you must be authenticated with an API key (or a valid PHP session which is a valid proxy)  -- we do not allow unauthenticated access. So every request to the API is mapped to a user / username.

We expect the API to be used by two primary consumers:
1. JS code like the existing Page Browser and upcoming new proofreading interface
2. tech-savvy project managers, post-processors, and perhaps others who want to use it for their own use or for automated systems

The endpoint in question is geared to enabling the new proofreading interface (that is: client) with the ability to store arbitrary (but well-structured) configuration information to persist across different browsers and devices. Fundamentally we want to allow a client to store one and only one blob per user. Because all endpoints are authenticated when they PUT to `v1/storage/clients/:client` we have that 1:1 mapping between the user (API key or PHP session) and the client (`clientid`).

We have an allow-list for valid `clientid`s to prevent a user from storing blobs for arbitrary `clientid` strings.

We can't rely on `clientid` to be a secret and known only to the JS code because it's available in the browser so we need to assume these values are effectively public. This means that any authenticated user can store any structurally-valid value at `v1/storage/clients/:clientid` outside of the JS code that it was intended for. That's just something the client is going to have to deal with.

So in summary, right now the maximum number of blobs that could be stored in this interface is `num_users * len(API_CLIENT_STORAGE_KEYS)`. The maximum size of a blob is not well defined but probably in the 16MB range, although I expect clients will store a lot less than that. If a user wanted to store a lot more information for their `clientid` record they _could_ but it would only impact them -- they can't store information for other users.

cpeel:
I'm going to do some slight renaming of some things in this to remove `client` as that's an unnecessary source of confusion. I'll work on that over the next couple of days.

cpeel:
Much later than I planned, but here's a small refactor. I've removed "client" everywhere and just went with "storage key". Hopefully that removes some confusion around the naming.

Functionally this hasn't changed:
* Admins can configure a storage key that API users can use to set, get, and delete valid JSON blobs.
* API users can only store one blob per storage key and users can only get, set, and delete their own blobs.

The new JS proofreading interface can use this to store per-user configuration information across browsers / computers. It could also be used for another client (such as the Page Browser, for instance) for the same purpose with a second storage key.

After https://github.com/DistributedProofreaders/dproofreaders/pull/1411 goes in (and maybe https://github.com/DistributedProofreaders/dproofreaders/pull/1407) I'll rebase this to fix any conflicts.

70ray (approved):

chrismiceli (approved):

jchaffraix (approved):

cpeel:
This has been rebased and the commits all squashed down into one. The sandbox has been updated and can be tested with:
```bash
export API_KEY=review

# Save a json blob
    -X PUT -d '{"key": 1}' "https://www.pgdp.org/~cpeel/c.branch/client-storage/api/v1/storage/newpi"

# Fetch a json blob
    -X GET "https://www.pgdp.org/~cpeel/c.branch/client-storage/api/v1/storage/newpi"

# Delete a json blob
    -X DELETE "https://www.pgdp.org/~cpeel/c.branch/client-storage/api/v1/storage/newpi"
```

srjfoo (approved):

PR #1406, opened 2024-12-24
Add endpoint for languages with dictionaries

70ray:
The code has been put in api_docs.inc. Perhaps we should give it a more generic name like api_site_data.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/api_dics

### DIFF
--- a/api/v1_docs.inc
+++ b/api/v1_docs.inc
@@ -31,3 +31,11 @@ function api_v1_document(string $method, array $data, array $query_params): stri
     }
     return $faq_url;
 }
+
+/** @param array<string, string|string[]> $query_params */
+function api_v1_dictionaries(string $method, array $data, array $query_params): array
+{
+    $dict_list = get_languages_with_dictionaries();
+    asort($dict_list);
+    return $dict_list;
+}
--- a/SETUP/tests/unittests/ApiTest.php
+++ b/SETUP/tests/unittests/ApiTest.php
@@ -835,7 +835,7 @@ class ApiTest extends ProjectUtils
     }
 
     //---------------------------------------------------------------------------
-    // tests for documents
+    // tests for documents & dictionaries
 
     public function test_available_italian_documents(): void
     {
@@ -872,6 +872,15 @@ class ApiTest extends ProjectUtils
         $router = ApiRouter::get_router();
         $_SERVER["REQUEST_METHOD"] = "GET";
         $router->route($path, ['language_code' => 'de']);
+    }
+
+    public function test_dictionaries(): void
+    {
+        $path = "v1/dictionaries";
+        $router = ApiRouter::get_router();
+        $_SERVER["REQUEST_METHOD"] = "GET";
+        $response = $router->route($path, []);
+        $this->assertEquals("English", $response["en"]);
     }
 
     //---------------------------------------------------------------------------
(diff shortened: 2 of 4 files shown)


api/v1_docs.inc
```diff
@@ -29,3 +29,10 @@ function api_v1_document(string $method, array $data, array $query_params): stri
     }
     return $faq_url;
 }
+
+function api_v1_dictionaries(string $method, array $data, array $query_params): array
+{
+    $dict_list = array_values(get_languages_with_dictionaries());
+    asort($dict_list);
```
70ray, line 36: Should be sort not asort.

chrismiceli (approved):

70ray:
Sandbox updated.

jchaffraix (approved):
Nit: typo in the description "dictioaries"

cpeel:
I rebased this to resolve the merge conflict.

api/v1_docs.inc
```diff
@@ -31,3 +31,10 @@ function api_v1_document(string $method, array $data, array $query_params): stri
     }
     return $faq_url;
 }
+
+function api_v1_dictionaries(string $method, array $data, array $query_params): array
```
cpeel, line 35:
```suggestion
/** @param array<string, string|string[]> $query_params */
function api_v1_dictionaries(string $method, array $data, array $query_params): array
```

api/v1_docs.inc
```diff
@@ -31,3 +31,10 @@ function api_v1_document(string $method, array $data, array $query_params): stri
...
+{
+    $dict_list = array_values(get_languages_with_dictionaries());
```
cpeel, line 37: I think if we returned the associated array directly from `get_languages_with_dictionaries()` instead of just the language this would be a more flexible endpoint.

cpeel (changes requested):
I've waffled on this for a bit as this will be a bit of a loner endpoint. Ultimately I agree that grouping it with docs makes sense rather than creating a single file just for this one function.

It needs to be added to `dp-openapi.yaml` and I had a couple of other small comments.

api/v1_docs.inc
70ray: Done this and changed to asort because sort would re-assign keys. These keys could be useful to send back to simplify finding the aspell dictionaries although the code currently has to re-discover them from the language names.

cpeel (approved):

api/v1_docs.inc
cpeel: > Isn't this premature optimization? in the sense of sending extra information that we don't need but could possibly be useful in the future.

It is, but when designing APIs we sometimes need to do that, because after we publish them we can't change them in ways that break. So having a little extra data here is going to mean we don't have to create a v2 of the endpoint to add it later.

cpeel:
Ray's sandbox isn't up-to-date so I published it under mine. This is testable with:
```bash
export API_KEY=review
    -X GET "https://www.pgdp.org/~cpeel/c.branch/api_dics/api/v1/dictionaries"
```

jchaffraix (approved):

70ray:
> Ray's sandbox isn't up-to-date so I published it under mine. ...
Thanks Casey, I forgot this. I've updated it now in case it's useful.

srjfoo (approved):

cpeel:
Rebased again to resolve the merge conflict. Will merge this in once it passes CI.

PR #1407, opened 2024-12-25
Type api/*

jchaffraix:
The first step was to rework the ApiRouter to
clearly state that it was using a trie and add
an appropriate TrieNode. This helped clean up
the types a bit.

Typed the handlers/validators though their
return value is anything that `json_encode`
can consume (so anything effectively :-)).

### DIFF
--- a/api/v1_projects.inc
+++ b/api/v1_projects.inc
@@ -36,7 +36,7 @@ function get_return_fields(?array $query_params, ?Project $project)
     return $return_fields;
 }
 
-function api_v1_projects($method, $data, $query_params)
+function api_v1_projects(string $method, array $data, array $query_params)
 {
     // set which fields are queryable and their column names
     $valid_fields = get_project_fields_with_attr("queryable", null);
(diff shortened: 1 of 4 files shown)


chrismiceli (approved):

70ray (approved):

cpeel:
@jchaffraix - Can you please break out the TrieNode changes into a separate PR we can get in first? This change makes sense but I'd like to review and merge it in separately than the typing changes.

jchaffraix:
> @jchaffraix - Can you please break out the TrieNode changes into a separate PR we can get in first? This change makes sense but I'd like to review and merge it in separately than the typing changes.

I have split the changes in api/ApiRouter.inc into #1411. This may be a bit more than what you asked as it includes some of the added typing in the class... Those seemed small enough they shouldn't impact the review, however let me know if you prefer that we break this further.

cpeel:
@jchaffraix - I just rebased this to resolve the merge conflict from https://github.com/DistributedProofreaders/dproofreaders/pull/1411

cpeel (approved):

chrismiceli (approved):

PR #1410, opened 2025-01-08
Use async and await for ajax

70ray:
Our current supported browsers can use async and await which need FF 52, Chrome 55, Safari 10.3.
This should make the code more comprehensible.
Sandbox at: https://www.pgdp.org/~rp31/c.branch/async_await
It should only affect pages using the page browser.

### DIFF
--- a/scripts/api.js
+++ b/scripts/api.js
@@ -6,6 +6,7 @@ const AJAX_ERROR_CODES = {
     UNKNOWN_ERROR: 999,
     INCORRECT_RESPONSE_TYPE: 998,
     NETWORK_ERROR: 997,
+    JSON_ERROR: 996,
 };
 
 class AjaxError extends Error {
@@ -52,7 +53,12 @@ async function ajax(method, apiUrl, queryParams = {}, data = {}, fetchPromise = 
     if (!contentType || !contentType.includes("application/json")) {
         throw new AjaxError("Incorrect response type", AJAX_ERROR_CODES.INCORRECT_RESPONSE_TYPE);
     }
-    const responseData = await response.json();
+    let responseData;
+    try {
+        responseData = await response.json();
+    } catch (error) {
+        throw new AjaxError(error, AJAX_ERROR_CODES.JSON_ERROR);
+    }
     if (!response.ok) {
         if (!responseData) {
             throw new AjaxError("Unknown error", AJAX_ERROR_CODES.UNKNOWN_ERROR);
--- a/scripts/api.js
+++ b/scripts/api.js
@@ -1,6 +1,5 @@
 /*global codeUrl */
 /* exported ajax */
-/* exported ajaxAlert */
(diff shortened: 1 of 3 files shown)


scripts/api.js
```diff
@@ -35,29 +42,22 @@ function ajax(method, apiUrl, queryParams = {}, data = {}, fetchPromise = fetch)
...
+    try {
+        response = await fetchPromise(url, options);
+    } catch (error) {
+        throw new AjaxError(error, AJAX_ERROR_CODES.NETWORK_ERROR);
+    }
+    const contentType = response.headers.get("content-type");
+    if (!contentType || !contentType.includes("application/json")) {
+        throw new AjaxError("Incorrect response type", AJAX_ERROR_CODES.INCORRECT_RESPONSE_TYPE);
+    }
+    const responseData = await response.json();
```
chrismiceli, line 55: we probably shouldn't call `.json()` until after we confirm the response is `ok`.

chrismiceli, line 55: Note .json can throw, not sure if we want to handle that.

chrismiceli, line 55: I'm not sure if it is the case yet, but in my experience, there are some API calls that return no body (204 response for example) and calling .json will fail there. I usually have a property or argument to the method configuration that allows the caller to say "don't decode the response body".

jchaffraix (approved):
Seconding the suggestions by @chrismiceli around checking for HTTP errors and potential exceptions in `.json`.

scripts/api.js
70ray: We need to decode the result even if it is not ok so this avoids doing it twice.

70ray: Good point. I'll handle it and add a test.

70ray: At the moment we only respond with JSON so we needn't worry about it yet.

scripts/page_browse.js
```diff
@@ -113,30 +113,23 @@ function pageBrowse(params, storageKey, replaceUrl, mentorMode = false, setShowF
...
-                        resolve();
-                    })
-                    .catch(function (data) {
-                        ajaxAlert(data);
-                        reject();
-                    });
-            }
-        });
-        return gotRoundSelector;
+    async function getRoundSelector() {
```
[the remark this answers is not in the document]
70ray: I'll do that.

70ray:
Sandbox updated.

chrismiceli (approved):

70ray:
It's not doing quite what it should in the real application.

70ray:
Sandbox updated again.

jchaffraix (approved):

cpeel (approved):
This is entirely outside my wheelhouse and I am happy to defer to @chrismiceli and @jchaffraix who have approved it.

srjfoo (approved):

PR #1411, opened 2025-01-09
Type api/ApiRouter.inc

jchaffraix:
This PR reworks the ApiRouter to clearly state that it was using a trie and add an appropriate TrieNode. This helped clarify what the code does and clean up the types in the class.

Sandbox: https://www.pgdp.org/~jchaffraix/c.branch/julien_trie_in_apirouter/

TESTS=Tried calling some APIs after the change to confirm that the routing still works.

### DIFF
--- a/api/ApiRouter.inc
+++ b/api/ApiRouter.inc
@@ -4,17 +4,34 @@ include_once("exceptions.inc");
 // Raise exceptions on assert failures
 ini_set("assert.exception", 1);
 
+/**
+ * We use a trie to match the path to its handler(s).
+ *
+ * Handlers contains the individual handlers, keyed by the method.
+ */
+class TrieNode
+{
+    /** @var array<string, TrieNode> */
+    public array $children;
+    /** @var array<string, callable> */
+    public array $handlers;
+}
+
 class ApiRouter
 {
-    private $_url_map = [];
-    private $_validators = [];
+    private TrieNode $root;
+    /** @var array<string, callable> */
+    private $_validators;
 
-    public function add_route($method, $url, $function)
+    public function __construct()
     {
-        // Confirm the function is defined or raise an assert exception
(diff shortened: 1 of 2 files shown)


api/ApiRouter.inc
```diff
@@ -4,17 +4,34 @@ include_once("exceptions.inc");
...
-    private $_validators = [];
+    private TrieNode $root;
+    /** @var array<string, callable> */
+    private $_validators;
 
-    public function add_route($method, $url, $function)
+    public function __construct()
     {
-        // Confirm the function is defined or raise an assert exception
-        assert(function_exists($function), "$function not defined");
```
cpeel, line 15: We should keep this in `add_route()` to help devs confirm that when they add new routes with validators that the validators are defined.

cpeel (commented):
I'd like to sanity-check this in your sandbox when I'm able, but the code LGTM sans this one dev-centric comment.

Thanks for breaking this out into a separate PR 🙏

api/ApiRouter.inc
jchaffraix: This is unneeded: the PHP runtime checks that a `callable` argument points to a valid function.

Let me show with an example:
```
function test(callable $f): void {}

test('print_r'); // Work on PHP 8.1/7.4.
test(print_r(...)); // Works on PHP 8.1 (first class callable function syntax).
test('not_a_function'); // Throws exception on PHP 8.1/7.4: Argument 1 must be of type callable but string given.
```

cpeel: Awesome, TIL! Thanks for explaining that 👍

cpeel (approved):

chrismiceli (approved):

70ray (approved):

srjfoo (approved):

PR #1414, opened 2025-01-19
Multiscript API

70ray:
This finds words with more than one script in the same word. We could instead find words with any characters in non-latin scripts as the present system does.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/multiscript_api

### DIFF
--- a/SETUP/tests/unittests/ApiTest.php
+++ b/SETUP/tests/unittests/ApiTest.php
@@ -787,6 +787,38 @@ class ApiTest extends ProjectUtils
         $this->assertEquals(["¿", "INVERTED QUESTION MARK"], $pickerset["subsets"][3]["rows"][1][1]);
     }
 
+    public function test_different_scripts()
+    {
+        global $request_body;
+
+        $request_body = ["text" => "Αlpha Вeta Gamma"];
+        $_SERVER["REQUEST_METHOD"] = "PUT";
+        $path = "v1/multiscript_words";
+        $router = ApiRouter::get_router();
+        $response = $router->route($path, []);
+        $expected = [
+            "words" => ["Αlpha" => [["Α", "Greek"], ["lpha", "Latin"]], "Вeta" => [["В", "Cyrillic"], ["eta", "Latin"]]],
+            "scripts" => ["Cyrillic", "Greek", "Latin"],
+        ];
+        $this->assertEquals($expected, $response);
+    }
+
+    public function test_different_scripts_in_word()
+    {
+        global $request_body;
+
+        $request_body = ["text" => "appᾔeaШrs"];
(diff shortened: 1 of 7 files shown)


srjfoo:
We've got two basic cases, don't we?
* Words that are all in a single non-Latin script
* Words that have mixed scripts

Both cases are covered by detecting any non-Latin characters, but this doesn't differentiate which case is which.

Can you explain more about what problem this is meant to solve, why it's better than what we're doing, and how the case of words in a single non-Latin should be handled? (Or am I over-thinking this? :D)

cpeel:
I'm not sure we need to solve the general case at all, much less as an API endpoint. If the client really wants to do something special with all the words in different scripts it can do that without any help from the site -- that can be done entirely in the client code (JS or otherwise).

If, however, the client wants to ask the site which of these words has characters in "uncommon" scripts as defined by the site, then it's asking the site for data it can't otherwise have and it makes sense to do that, and use the existing functions that use the already-defined `$common_unicode_scripts`.

70ray:
> We've got two basic cases, don't we?
> 
>     * Words that are all in a single non-Latin script
> 
>     * Words that have mixed scripts
> 
> 
> Both cases are covered by detecting any non-Latin characters, but this doesn't differentiate which case is which.
> 
> Can you explain more about what problem this is meant to solve, why it's better than what we're doing, and how the case of words in a single non-Latin should be handled? (Or am I over-thinking this? :D)

There may have been some discussion of this before but I couldn't find it. I presume the purpose of highlighting uncommon scripts is to detect where a wrong script has been used when the characters are identical such as latin A and capital alpha. For the case of a page mostly in Greek script there does not seem to be much value in highlighting it all. But this is just a proposal. I'm quite happy to code something else if you think it is better.

70ray:
> I'm not sure we need to solve the general case at all, much less as an API endpoint. If the client really wants to do something special with all the words in different scripts it can do that without any help from the site -- that can be done entirely in the client code (JS or otherwise).
> 
> If, however, the client wants to ask the site which of these words has characters in "uncommon" scripts as defined by the site, then it's asking the site for data it can't otherwise have and it makes sense to do that, and use the existing functions that use the already-defined `$common_unicode_scripts`.

I thought at first we would not need to involve the server in highlighting uncommon scripts, but javascript does not have the ability to find the unicode script of a character without a third-party package such as [](https://www.npmjs.com/package/unicode-properties).
Highlighting all non-latin characters on a page seemed to me to add more clutter than value but if that is what proofreaders would like then I am more than happy to do it like that.

70ray:
As @cpeel said, "uncommon" is defined by the site. We could consider making this configuarble per project or page.

70ray:
Closing this while investigating client side solutions.

PR #1415, opened 2025-01-20
Add the url to the invalid api message

70ray:
So the client can tell which request failed.

Sandbox at: https://www.pgdp.org/~rp31/c.branch/improve_error

### DIFF
--- a/api/ApiRouter.inc
+++ b/api/ApiRouter.inc
@@ -59,18 +59,22 @@ class ApiRouter
         $node = $this->root;
         $data = [];
         $parts = explode("/", $url);
-        foreach ($parts as $part) {
-            $next_node = $node->children[$part] ?? null;
-            if ($next_node) {
-                $node = $next_node;
-            } else {
-                [$param_name, $validator] = $this->get_validator($node);
-                $node = $node->children[$param_name];
-                $data[$param_name] = $validator($part, $data);
+        try {
+            foreach ($parts as $part) {
+                $next_node = $node->children[$part] ?? null;
+                if ($next_node) {
+                    $node = $next_node;
+                } else {
+                    [$param_name, $validator] = $this->get_validator($node);
+                    $node = $node->children[$param_name];
+                    $data[$param_name] = $validator($part, $data);
+                }
             }
-        }
-        if (empty($node->handlers)) {
-            throw new InvalidAPI();
+            if (empty($node->handlers)) {
+                throw new InvalidAPI();
+            }
+        } catch (InvalidAPI $exception) {
+            throw new InvalidAPI($exception->getMessage() . ": $url");
         }
         $method = $_SERVER["REQUEST_METHOD"];
         $handler = $node->handlers[$method] ?? null;


chrismiceli (approved):

cpeel:
The caller was the one who sent the invalid URL, why doesn't it know what URL it sent?

70ray:
> The caller was the one who sent the invalid URL, why doesn't it know what URL it sent?

Good question. I can delete this PR since it it not needed, but, although it may not be relevant now, I'll give a detailed explanation of how this arose:
I was re-making my prototype PI but had left out some of the endpoints so was getting this error. When I've come across this situation before, the network tab in the developer tools shows what has happened. Owing to a circumstance which I did not understand at the time the API requests were not appearing in the developer tools so I made this modification to discover what I had left out (since there was a chain of api requests in the code).
I later discovered that the reason the API requests were not showing up in the tools (in Firefox, haven't tried others) was because I had added the 'keepalive' flag to the ajax request for a reason I can go into later.
I still later found that the 'keepalive' flag needs Firefox 133 so I could not use it anyway so am now looking at another way to solve the problem it was intended to fix.

cpeel:
>  I later discovered that the reason the API requests were not showing up in the tools (in Firefox, haven't tried others) was because I had added the 'keepalive' flag to the ajax request for a reason I can go into later.

Interesting. KeepAlive requests don't show up in the Network section for developer tools? That is ... unexpected.

api/ApiRouter.inc
```diff
@@ -59,18 +59,22 @@ class ApiRouter
...
+                }
             }
-        }
-        if (empty($node->handlers)) {
-            throw new InvalidAPI();
+            if (empty($node->handlers)) {
+                throw new InvalidAPI();
+            }
+        } catch (InvalidAPI $exception) {
+            throw new InvalidAPI($exception->getMessage() . ": $url");
```
jchaffraix, line 77: While I agree with Casey that this is PR is unneeded (and seems like something off with Firefox, maybe you need to catch the socket when it is created and before Firefox starts reusing it (a bit like how websockets need to be handled)), I wonder if we couldn't help with the issue in a way that's more useful in general.

Something that may have helped clear some confusion is logging the URL in our error logs when we're throwing an exception in the API layer. We could handle this by improving our existing global exception handler to call `error_log` with some API specific information (like actual api called, ...). Thoughts?

70ray:
> > I later discovered that the reason the API requests were not showing up in the tools (in Firefox, haven't tried others) was because I had added the 'keepalive' flag to the ajax request for a reason I can go into later.
> 
> Interesting. KeepAlive requests don't show up in the Network section for developer tools? That is ... unexpected.

I don't know much about this high-level stuff but it seems that there is more than one sort of keep-alive. The one I was using was [this](https://developer.mozilla.org/en-US/docs/Web/API/Request/keepalive). Since it is only available in FF 133 and FF is now on 134 perhaps not too surprising the developer tools were not quite up-to-speed.

api/ApiRouter.inc
cpeel: `error_log` is going to be putting the data in the wrong place to be useful. The API user is the one who might benefit from this knowledge, but it'll be logged server-side which virtually none of them will be able to access -- Ray on the TEST being the exception because he's a developer.

Doing this temporarily in a sandbox to help debug things during development is a great idea though.

70ray: (I didn't know I could do that)

cpeel: Oh sure, you're welcome to use error_log during development. There's a sandbox-specific php_errors file that sandboxes dump into.

PR #1418, opened 2025-01-24
Add API endpoint to report a bad page.

70ray:
Sandbox at: https://www.pgdp.org/~rp31/c.branch/report_bad

### DIFF
--- a/api/dp-openapi.yaml
+++ b/api/dp-openapi.yaml
@@ -1212,6 +1212,41 @@ paths:
         default:
           $ref: '#/components/responses/UnexpectedError'
 
+  /projects/{projectid}/pages/{pagename}/reportbad:
+    put:
+      tags:
+      - project
+      description: Mark a page as bad
+      parameters:
+      - $ref: '#/components/parameters/projectid'
+      - $ref: '#/components/parameters/pagename'
+      requestBody:
+        description: reason for badness
+        required: true
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                reason:
+                  type: string
+                  enum: [missing_image, missing_text, image_text_mismatch, corrupted_image, other]
+      responses:
+        200:
+          description: OK
+        '400':
+          $ref: '#/components/responses/InvalidValue'
+        '401':
+          $ref: '#/components/responses/Unauthorized'
+        '403':
+          $ref: '#/components/responses/Forbidden'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '429':
+          $ref: '#/components/responses/RateLimitExceeded'
+        default:
+          $ref: '#/components/responses/UnexpectedError'
+
   /stats/site:
     get:
       tags:
--- a/api/v1.inc
+++ b/api/v1.inc
@@ -45,6 +45,7 @@ $router->add_route("GET", "v1/projects/:projectid/pagedetails", "api_v1_project_
 $router->add_route("GET", "v1/projects/:projectid/pages", "api_v1_project_pages");
 $router->add_route("PUT", "v1/projects/:projectid/pages/:pagename", "api_v1_project_page");
 $router->add_route("GET", "v1/projects/:projectid/pages/:pagename", "api_v1_project_page");
+$router->add_route("PUT", "v1/projects/:projectid/pages/:pagename/reportbad", "api_v1_project_page_report_bad");
 $router->add_route("PUT", "v1/projects/:projectid/pages/:pagename/wordcheck", "api_v1_project_page_wordcheck");
 $router->add_route("GET", "v1/projects/:projectid/pages/:pagename/pagerounds/:pageroundid", "api_v1_project_page_round");
 $router->add_route("GET", "v1/projects/:projectid/pickersets", "api_v1_project_pickersets");
--- a/api/v1_projects.inc
+++ b/api/v1_projects.inc
@@ -908,6 +908,35 @@ function api_v1_project_page(string $method, array $data, array $query_params)
     }
 }
 
+/** @param array<string, string|string[]> $query_params */
+function api_v1_project_page_report_bad(string $method, array $data, array $query_params): void
+{
+    global $pguser, $PAGE_BADNESS_REASONS;
+
+    try {
+        $project = $data[":projectid"];
+        $proof_project = new ProofProject($project);
+
+        $project_page = $data[":pagename"];
+        $proof_project_page = new ProofProjectPage($proof_project, $project_page);
+
+        $reason_string = receive_data_from_request_body("reason");
+        $reason_key = array_find_key($PAGE_BADNESS_REASONS, function (array $value) use ($reason_string) {
+            return $value["name"] === $reason_string;
+        });
+        if (null === $reason_key) {
+            throw new InvalidValue("$reason_string is not a valid badness reason");
+        }
+        $proof_project_page->markAsBad($pguser, $reason_key);
+    } catch (ProjectTransitionException $exception) {
+        throw new UnexpectedError($exception->getMessage(), $exception->getCode());
+    } catch (ProjectException $exception) {
+        throw new BadRequest($exception->getMessage(), $exception->getCode());
+    } catch (UserAccessException $exception) {
+        throw new ForbiddenError($exception->getMessage(), $exception->getCode());
+    }
+}
+
 /** @param array<string,string|string[]> $query_params */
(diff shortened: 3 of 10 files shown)


pinc/LPage.inc
```diff
@@ -467,7 +467,7 @@ class LPage
 
         $error_msg = project_transition($this->projectid, $this->round->project_bad_state, PT_AUTO);
         if ($error_msg) {
-            echo "$error_msg<br>\n";
+            throw new ProjectException($error_msg);
```
70ray, line 470: This will not get caught in report_bad_page.php which only catches 'ProjectPageException', should we change this to catch it?

cpeel: Yes, we should add `ProjectException` to the list of exceptions that `report_bad_page.php` will accept.

api/dp-openapi.yaml
```diff
@@ -1121,6 +1121,8 @@ paths:
             revert - Checks the page for invalid characters and if valid, saves it as in progress. Response:- page_text_state object.
             <br><br>
             abandon - Returns the page to the round, response:- none.
+            <br><br>
+            report_bad - Makes a bad page report, response:- none.
```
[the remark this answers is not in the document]
70ray: Regarding the request body, this also applies to the resume and abandon cases which don't need any request body. This seems like a documentation issue rather than a protocol design issue in the sense that there doesn't seem to be any reason why the same endpoint cannot do different things depending on a request parameter and each can be associated with a different sort of request body, but Swagger does not enable the association to be explicitly stated.
Since we are using Swagger to make the documentation it would make some sense to divide this endpoint up into endpoints which have the same request body or even into six different endpoints for each of the functions which we are now distinguising by the pageaction parameter.

pinc/LPage.inc
70ray: We should also catch PageNotOwnedException which can be thrown in page_require.

api/dp-openapi.yaml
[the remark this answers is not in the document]
cpeel: But at least resume and abandon would accept a `{"text": "blah"}` schema and still do the right thing (by ignoring it). This one wouldn't.

Protocol design *is* documentation. Swagger will let us say an endpoint accepts multiple schemas (https://swagger.io/docs/specification/v3_0/data-models/oneof-anyof-allof-not/) we just have to tell it.

pinc/DPage.inc
```diff
@@ -461,13 +461,33 @@ function Page_clearRound(string $projectid, string $image, Round $round, string
...
+// The numeric values of the following array must not change as they are
+// saved in the database to indicate the bad page reason.
 $PAGE_BADNESS_REASONS = [
-    '',
-    _('Missing Image'),
-    _('Missing Text'),
-    _('Image/Text Mismatch'),
-    _('Corrupted Image'),
-    _('Other'),
+    0 => [
```
70ray, line 467: Maybe remove "none" since it is not a valid reason. And make necessary adjusments elsewhere.

70ray:
Sandbox updated.

api/v1_projects.inc
```diff
@@ -880,6 +880,37 @@ function api_v1_project_page(string $method, array $data, array $query_params)
...
+
+    try {
+        $project = $data[":projectid"];
+        $proof_project = new ProofProject($project);
+
+        $project_page = $data[":pagename"];
+        $proof_project_page = new ProofProjectPage($proof_project, $project_page);
+
+        $reason_string = receive_data_from_request_body("reason");
+        // can't use array_find_key() yet
```
70ray, line 895: Should we update the symfony polyfil so we can use this function? then when we move to php8.* it will be ready.

pinc/DPage.inc
70ray: Done this.

70ray:
Sandbox updated again.

api/v1_projects.inc
cpeel: This function is in 8.4 so that's a ways out from our minimum version (8.2 going forward IIRC). But yes, I'm fine updating our polyfills to include the 8.x ones so we can use this. Let's do that in a separate PR.

tools/project_manager/handle_bad_page.php
```diff
@@ -193,7 +193,7 @@ function show_resolution_form($projectid, $image, $state, $project_round, $is_a_
         }
 
         if (!empty($b_code)) {
-            echo "<b>" . _("Reason") . ":</b> {$PAGE_BADNESS_REASONS[$b_code]}</br>";
+            echo "<b>" . _("Reason") . ":</b> {$PAGE_BADNESS_REASONS[$b_code]["string"]}</br>";
```
cpeel, line 196: This is getting a little convoluted and I'm surprised PHP isn't complaining about using `"` inside a string delimited by `"`s. Perhaps bypass it altogether with:

```suggestion
            echo "<b>" . _("Reason") . ":</b> " . $PAGE_BADNESS_REASONS[$b_code]["string"] . "</br>";
```

api/v1.inc
```diff
@@ -45,6 +45,7 @@ $router->add_route("GET", "v1/projects/:projectid/pagedetails", "api_v1_project_
 $router->add_route("GET", "v1/projects/:projectid/pages", "api_v1_project_pages");
 $router->add_route("PUT", "v1/projects/:projectid/pages/:pagename", "api_v1_project_page");
 $router->add_route("GET", "v1/projects/:projectid/pages/:pagename", "api_v1_project_page");
+$router->add_route("PUT", "v1/projects/:projectid/pages/:pagename/report_bad", "api_v1_project_page_report_bad");
```
cpeel, line 48: We haven't used `_`s in any of the endpoints so far, despite things like "imagesources" and "specialdays" so let's not start now:

```suggestion
$router->add_route("PUT", "v1/projects/:projectid/pages/:pagename/reportbad", "api_v1_project_page_report_bad");
```

(with corresponding update in `dp-openapi.yaml`)

api/v1_projects.inc
```diff
@@ -880,6 +880,37 @@ function api_v1_project_page(string $method, array $data, array $query_params)
...
+                $reason_key = $key;
+                break;
+            }
+        }
+        if (null === $reason_key) {
+            throw new InvalidValue("$reason_string is not a valid badness reason");
+        }
+        $proof_project_page->markAsBad($pguser, $reason_key);
+    } catch (ProjectException $exception) {
+        throw new NotFoundError($exception->getMessage(), $exception->getCode());
```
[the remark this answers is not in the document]
70ray: BadRequest should cover all all cases.

70ray:
Sandbox updated.

api/v1_projects.inc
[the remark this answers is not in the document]
70ray: Project or page not found will have been trapped earlier by the validators and throw notFoundError. Then, leaving aside the user exceptions which are trapped separately, the possible exceptions are:

- project not in a round
- project requires maintenance
- project not available
- page is in wrong state
Can we call these bad requests?
Project transition can return an error "Transition is illegal." This should not ever happen for our case so it would be more appropriate to return an unexpected error.

api/v1_projects.inc
70ray: Presumably I can't do that because it is not under git control.

cpeel: It's under git: https://github.com/DistributedProofreaders/dproofreaders/pull/1423

70ray:
Sandbox updated with array_find_key()

api/v1_projects.inc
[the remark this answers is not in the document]
cpeel: > These possibly could happen in some circumstance. can we call these bad requests?

I think that makes sense to me, yes.

api/v1_projects.inc
```diff
@@ -880,6 +880,32 @@ function api_v1_project_page(string $method, array $data, array $query_params)
     }
 }
 
+function api_v1_project_page_report_bad(string $method, array $data, array $query_params): void
```
cpeel, line 883:
```suggestion
/** @param array<string, string|string[]> $query_params */
function api_v1_project_page_report_bad(string $method, array $data, array $query_params): void
```

To get ahead of @jchaffraix's PR that this will miss.

cpeel (approved):

70ray:
Sandbox updated.

srjfoo (approved):

jchaffraix (approved):

cpeel:
Rebased to fix merge conflict. Will merge this in after CI completes successfully.

PR #1419, opened 2025-01-26
Try to speed up manage proofers suggestions

cpeel:
This attempts to speed up the "Manage proofreaders' Suggestions" page from [Task 2098](https://www.pgdp.net/c/tasks.php?action=show&task_id=2098) after some analysis in xdebug. Unfortunately there's not a lot we can do to make it faster because the bulk of the time is spent in getting all the distinct words from each project and ultimately that boils down to a `preg_match_all()`. I tried various iterations on ways to speed that part up but none of it panned out.

These changes are small code improvements that I tested and discovered which may help the page load faster on a loaded system but it's likely only to be a marginal improvement.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/faster-word-suggestions/

### DIFF
--- a/pinc/wordcheck_engine.inc
+++ b/pinc/wordcheck_engine.inc
@@ -16,12 +16,9 @@ $bracketed_character_pattern = "\\[(?:oe|OE|$mark$base|$base$mark)\\]";
 
 // This is used when splitting a text into words.
 $char_pattern = "(?:\w\pM*|$bracketed_character_pattern)";
-$word_pattern = "!$char_pattern+(?:'$char_pattern+)*!uS";
+$word_pattern = "!$char_pattern+(?:'$char_pattern+)*!u";
 // (The pattern is delimited by exclamation marks rather than the usual slashes,
 // because slash is a character within the pattern.)
-// (Supposedly, the 'S' modifier makes the regex engine spend more time
-// analyzing the pattern in order to speed up the time taken for matching,
-// but the effect seems pretty close to negligible.)
 
 
 // The following two variables control text highlighting in the interface.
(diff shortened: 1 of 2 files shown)


70ray (approved):

chrismiceli (approved):

pinc/wordcheck_engine.inc
```diff
@@ -755,6 +752,29 @@ function load_wordcheck_events($projectid, $start_time = 0)
     return $eventArray;
 }
 
+function load_project_good_word_suggestions_flat(string $projectid, int $start_time = 0): array
```
jchaffraix, line 755:
```suggestion
/** @return string[] */
function load_project_good_word_suggestions_flat(string $projectid, int $start_time = 0): array
```

jchaffraix (approved):
You probably thought about using memcache for caching some of the result(s), what is the project stance on it?

Also reading the description, it's unclear if the core slowness is caused by some repeated SQL queries (ie doing too many of those and/or hitting some tail latency due to the sheer number) vs some slow operations (regexp matching or some repeated transformations) vs a mix of both.

cpeel:
> You probably thought about using memcache for caching some of the result(s), what is the project stance on it?
> 
> Also reading the description, it's unclear if the core slowness is caused by some repeated SQL queries (ie doing too many of those and/or hitting some tail latency due to the sheer number) vs some slow operations (regexp matching or some repeated transformations) vs a mix of both.

The SQL to get the text from every page of a project isn't that slow but it does add up. But the bulk of the time per project is spent taking the text of every page and getting all the distinct words in it. That's a `preg_match_all()` which we can't get any faster. The combination of "get every page, get distinct words in the page" for a hundred projects just adds up, and if there are a lot of suggestions it's a lot of data to shove over the wire too.

Caching the results of the distinct words wouldn't be useful because the hit rate would be almost 0 based on how many projects we have in the site and the frequency of this being used -- it's a good question though!

srjfoo (approved):

Issue #1141, opened 2024-03-12
Remove reference to Internet Explorer in report_bad_page.php


70ray:
Since we no longer support IE.

What should we replace this list item with? Since I've never encountered this problem I don't know what to suggest.

bpfoley:
To me, it sounds like the example in text is referring to intermittent issue that might occur due to an unreliable/slow internet connection -- the browser gave up on loading the image because of a timeout. It doesn't sound like a browser-specific problem at all.

"Sometimes, the image may not show up due to technical problems with your browser.  Depending upon your browser there   are many ways to try to reload that image.  For example, in Internet Explorer you can right click on the image & left click Show Image   or Refresh.  This 90% of the time causes the image to then display." 

Since we're talking about IE:

- [faq/feeds_sdk/index.php](https://github.com/DistributedProofreaders/dproofreaders/blob/master/faq/feeds_sdk/index.php#L218) has an out of date reference to IE which is irrelevant to us since we don't support it.

- [tasks.php](https://github.com/DistributedProofreaders/dproofreaders/blob/master/tasks.php#L200) has a very arbitrary `$browser_array` and `$os_array` with ridiculously out of date entries. We really need a schema change to turn these into free-form text fields (even if we keep a list of suggested OSes/Browsers to autocomplete (or autodetect from the User-Agent) when creating new tasks.)

srjfoo:
> Since we no longer support IE.
> 
> What should we replace this list item with? Since I've never encountered this problem I don't know what to suggest.

Several years ago, a couple of the db-req squirrels looked at `report_bad_page.php` and thought that we should rewrite those instructions completely. We got distracted by more immediate tasks and never got it done. Judging from the fact that some people apparently think every intentional blank page is a bad page, I'm not sure how many people actually read those instructions. And that first paragraph, I think needs to be totally rewritten. What it's describing is not what we expect of proofreaders nowadays.

I agree with Brian about removing IE from other places on the site, but I think that's a separate issue.

PR #1420, opened 2025-01-28
Update instructions for reporting a bad page

cpeel:
This was nominally to resolve https://github.com/DistributedProofreaders/dproofreaders/issues/1141 but really the instructions were quite bad so I just rewrote them. I'm very open to changes, but let's settle on something that is better than what we have now vs letting perfect be the enemy of good enough.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/revise-report-bad-page/

### DIFF
--- a/tools/proofers/report_bad_page.php
+++ b/tools/proofers/report_bad_page.php
@@ -31,17 +31,28 @@ if (!isset($_POST['submitted']) || $_POST['submitted'] != 'true') {
 
     echo "<h1>$header</h1>";
 
-    echo "<h2>"._("Common Fixes for Bad Pages. Try these first!")."</h2>";
+    echo "<p>" . _("If you are unable to proofread the page you were presented, you can mark it bad with this form to let the Project Manager know it requires attention. Before doing so, let's review what constitutes a bad page and some possible fixes you can try first.") . "</p>";
+
+    echo "<h2>" . _("Commonly Misidentified Bad Pages") . "</h2>";
+    echo "<p>" . _("The following scenarios are commonly reported as bad pages, but they are not. If either of the below are true, return and continue proofreading the page.") . "</p>";
     echo "<ul>";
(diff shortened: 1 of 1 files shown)


srjfoo (approved):
Thank you!

tools/proofers/report_bad_page.php
```diff
@@ -31,17 +31,28 @@
 
     echo "<h1>$header</h1>";
 
-    echo "<h2>"._("Common Fixes for Bad Pages. Try these first!")."</h2>";
+    echo "<p>" . _("If you are unable to proofread the page you were presented, you can mark it bad with this form to let the Project Manager know it requires attention. Before doing so, let's review what constitutes a bad page and some possible fixes you can try first.") . "</p>";
+
+    echo "<h2>" . _("Commonly Misidentified Bad Pages") . "</h2>";
+    echo "<p>" . _("The following scenarios are commonly reported as bad pages, but they are not. If either of the below are true, return and continue proofreading the page.") . "</p>";
```
70ray, line 37: Maybe "If either of the below are true, hit 'Cancel' below and continue proofreading the page.

srjfoo: That sounds like a good idea.

---

Not something for this PR, but @70ray's comment reminded me of [Task #1972: Cancel Bad Page report fails if page was saved as in progress](https://www.pgdp.net/c/tasks.php?action=show&task_id=1972). I just verified that it's still happening. This will happen for anything that causes a save in progress (though I seem to be getting different error messages, depending on whether I cancel report bad page after report bad page after WC, or after changing from vertical to horizontal mode).

I don't know how important it is to get fixed -- I discovered it by accident, and nobody else seems to have stumbled over it, because generally they mark them bad before trying to proofread.

70ray (approved):

jchaffraix (approved):

PR #1421, opened 2025-01-28
Refine `$query_params` in API

jchaffraix:
_GET is an array<string, string> so we can refine the query_param in all API functions. This removes some violations in PHPStan level 6.

This uncovered 2 issues in api_v1_projects where we assumed the array could hold a nested array. The code was simplified to account for this.


cpeel:
> This uncovered 2 issues in api_v1_projects where we assumed the array could hold a nested array. The code was simplified to account for this.

But it can because PHP is special and we rely on this, eg:
```
    -X GET "https://www.pgdp.org/api/v1/projects?field[]=projectid&field[]=title"
```

`$query_params` will be something like:
```
$query_params = [
    "field" => [
        "projectid",
        "title",
    ],
];
```

chrismiceli (approved):

chrismiceli (approved):

jchaffraix:
> > This uncovered 2 issues in api_v1_projects where we assumed the array could hold a nested array. The code was simplified to account for this.
> 
> But it can because PHP is special and we rely on this, eg:
> 
> ```
> curl -i -H "Accept: application/json" -H "X-API-KEY: $API_KEY" \
>     -X GET "https://www.pgdp.org/api/v1/projects?field[]=projectid&field[]=title"
> ```
> 
> `$query_params` will be something like:
> 
> ```
> $query_params = [
>     "field" => [
>         "projectid",
>         "title",
>     ],
> ];
> ```

Thanks :bow: , I tested it out but didn't know that the trick was the `[]` on the field to allow an array in a query (without it, the last one wins). I will update the type everywhere to be `array<string, string|string[]>` and revert the changes related to array handling.

cpeel (approved):

70ray (approved):

srjfoo (approved):

PR #1427, opened 2025-02-03
Use SessionHandlerInterface object for sessions

cpeel:
Using the individual callback form of `session_set_save_handler()` is deprecated in PHP 8.4. Use the SessionHandlerInterface form instead which works with PHP > 5.4.

Diff best viewed ignoring whitespace, probably. No functional changes -- users should still be able to log in / log out / stay logged in just as before.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/update-session-handler/

### DIFF
--- a/pinc/dpsession.inc
+++ b/pinc/dpsession.inc
@@ -44,8 +44,6 @@ function dpsession_resume()
 
     if (isset($_SESSION['pguser']) && !empty($_SESSION['pguser'])) {
         // Refresh the cookie
-        // (session_start() used to do this for us,
-        // but they changed that in PHP 4.3.9)
         dp_setcookie(
             ini_get('session.name'),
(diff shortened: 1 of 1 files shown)


cpeel:
I tried running the smoketests against PHP 8.4 in the CI which is where I found this. Something's still not right for 8.4 though as it appears to be failing to log in the user -- loading the Activity Hub is returning 302 as if the user is not logged in -- but the same code is passing for 7.4, 8.1, and 8.3. I don't have a 8.4 system to debug this on at the moment but this PR is useful to get us one step closer to working with 8.4 so its worth doing.

chrismiceli (approved):

srjfoo (approved):

70ray (approved):

PR #1430, opened 2025-02-04
make function get_verbose_pickersets() in trait CharSuiteSet

70ray:
Use the function in v1_projects.inc. This will make it possible to use the toolbox of the new proofreading interface in the quiz (without any api calls).
Sandbox at: https://www.pgdp.org/~rp31/c.branch/move_verbose

### DIFF
--- a/api/v1_projects.inc
+++ b/api/v1_projects.inc
@@ -845,14 +845,7 @@ function api_v1_project_wordcheck(string $method, array $data, array $query_para
 function api_v1_project_pickersets(string $method, array $data, array $query_params): array
 {
     $project = $data[":projectid"];
-    $verbose_pickersets = [];
-    foreach ($project->get_pickersets() as $pickerset) {
-        $verbose_pickersets[] = [
-            "name" => $pickerset->name,
-            "subsets" => $pickerset->get_verbose_subsets(),
-        ];
-    }
-    return $verbose_pickersets;
+    return $project->get_verbose_pickersets();
 }
 
 /**
--- a/pinc/CharSuites.inc
+++ b/pinc/CharSuites.inc
@@ -247,7 +247,7 @@ class CharSuites
 
 trait CharSuiteSet
 {
-    /** @return array{name:string, subsets:array} */
+    /** @return array<int<0, max>, array{name: string, subsets: array}> */
     public function get_verbose_pickersets(): array
     {
         $verbose_pickersets = [];
--- a/pinc/CharSuites.inc
+++ b/pinc/CharSuites.inc
@@ -31,6 +31,7 @@ class PickerSet
         return $this->subsets;
     }
 
+    /** @return array{name: string, title: string, rows: (?string[])[][]}[] */
     public function get_verbose_subsets(): array
     {
         $verbose_subsets = [];
@@ -247,7 +248,7 @@ class CharSuites
 
 trait CharSuiteSet
 {
-    /** @return array<int<0, max>, array{name: string, subsets: array}> */
+    /** @return array{name: string, subsets: array{name: string, title: string, rows: (?string[])[][]}[]}[] */
     public function get_verbose_pickersets(): array
     {
         $verbose_pickersets = [];
--- a/pinc/CharSuites.inc
+++ b/pinc/CharSuites.inc
@@ -247,6 +247,7 @@ class CharSuites
 
 trait CharSuiteSet
 {
+    /** @return array{name:string, subsets:array} */
     public function get_verbose_pickersets(): array
     {
         $verbose_pickersets = [];
--- a/pinc/CharSuites.inc
+++ b/pinc/CharSuites.inc
@@ -247,6 +247,18 @@ class CharSuites
 
 trait CharSuiteSet
 {
+    public function get_verbose_pickersets(): array
+    {
+        $verbose_pickersets = [];
+        foreach ($this->get_pickersets() as $pickerset) {
+            $verbose_pickersets[] = [
+                "name" => $pickerset->name,
+                "subsets" => $pickerset->get_verbose_subsets(),
+            ];
+        }
+        return $verbose_pickersets;
+    }
+
     /** @return PickerSet[] */
     public function get_pickersets(): array
     {


pinc/CharSuites.inc
```diff
@@ -247,6 +247,18 @@ class CharSuites
 
 trait CharSuiteSet
 {
+    public function get_verbose_pickersets(): array
```
cpeel, line 250:
```suggestion
    /** @return array{0:string, 1:PickerSet[]} */
    public function get_verbose_pickersets(): array
```
I *think*. If PHPStan doesn't like this we'll need to consult with @jchaffraix on the right form.

@jchaffraix - it would be helpful for us n00bs if we updated the [DP Code Best Practices](https://www.pgdp.net/wiki/DP_Code_Best_Practices) section with some examples of how to properly annotate the return values of the various array forms we use. Is that something you could do?

70ray: perhaps it should be
```php
/** @return array{name:string, subsets:array} */

Comment (70ray) [reply pinc/CharSuites.inc:251]: That didn't work. The error messages says it should be:
array<int<0, max>, array{name: string, subsets: array}>

Comment (70ray) [reply pinc/CharSuites.inc:251]: That worked but it should probably specify what subsets is an array of. The function get_verbose_subsets() doesn't have anything specific, but it would probably be something complicated.

Comment (cpeel) [reply pinc/CharSuites.inc:251]: We can both simplify that and make it more specific -- I just tested this:
```
    /** @return array<array{name: string, subsets: PickerSet[]}> */
```

Comment (70ray) [reply pinc/CharSuites.inc:252]: Added a return type to get_verbose_subsets() also so it has to agree.

File: pinc/CharSuites.inc [hunk 2]
```diff
@@ -31,6 +31,7 @@ class PickerSet
         return $this->subsets;
     }
 
+    /** @return array<int, array{name: string, title: string, rows: array<int, array<int, array{string|null, string|null}>>}> */
```
cpeel, line 34: I don't think you actually need the `<int` part of this -- `array<array{` should be sufficient and makes it clearer that it's just a list vs an associative array. Several places in here and in the other one.

70ray: I'll try that

jchaffraix: Sorry about the delayed reply, work has been pretty harsh lately.

> @jchaffraix - it would be helpful for us n00bs if we updated the [DP Code Best Practices](https://www.pgdp.net/wiki/DP_Code_Best_Practices) section with some examples of how to properly annotate the return values of the various array forms we use. Is that something you could do?

Sure, happy to! I will update the typing section with some syntax.

> /** @return array<int, array{name: string, title: string, rows: array<int, array<int, array{string|null, string|null}>>}> */

There are 4 cases with arrays:
1. arbitrary string key -> value mapping (aka dictionnary) in which case you can use `array<key_type, value_type>`.
2. string key -> value mapping with known key/values, you can use `array{key1:type1, key2: type2}`
3. arbitrary index keys to values (aka vectors or lists) can use the syntax `type[]` (where `type` is the superset of the values' types)
4. vectors with known indexes can use the syntax `array{0:type1,1:type2}` (it's effectively 2. but with integer keys)

Here it seems like we got everything... This would probably benefit from defining some type to avoid the complicated output. I think we can simplify this to:

` /** @return array{name: string, title: string, rows: ?string[]}[] */`

70ray:
Now using classes for VerbosePickerSet and VerboseSubset which simplifies php_stan comments. This looks different but json_encodes to the same thing.

pinc/CharSuites.inc
```diff
@@ -243,10 +225,60 @@ class CharSuites
     }
 }
 
+class VerboseSubset
```
cpeel, line 228: Without context this name doesn't make a lot of sense.

This class also makes it non-parallel with PickerSet (No PickerSetSubset class). I think it would make more sense to remove this class and keep VerbosePickerSet to keep the two similar and bypass the nonsensical-without-context class name.

api/v1_projects.inc
```diff
@@ -822,14 +822,7 @@ function api_v1_project_wordcheck(string $method, array $data, array $query_para
...
     $project = $data[":projectid"];
-    $verbose_pickersets = [];
-    foreach ($project->get_pickersets() as $pickerset) {
-        $verbose_pickersets[] = [
-            "name" => $pickerset->name,
-            "subsets" => $pickerset->get_verbose_subsets(),
-        ];
-    }
-    return $verbose_pickersets;
+    return $project->get_verbose_pickersets();
```
cpeel, line 825: This works by [implicit casting of an object as an array](https://www.php.net/manual/en/language.types.array.php#language.types.array.casting) which is confusing (it confused the hell out of me how this worked to render the same JSON). We should either:
1. explicitly cast it as an array here so it's clearer what's going on.
2. add a comment what's happening here.
3. both

pinc/CharSuites.inc
cpeel: Thank you @jchaffraix for ^^ and the update to the DP Code Best Practices! I really appreciate you taking the type to write both of those up. 🙏

Sorry to hear work has been harsh recently and I hope it improves soon!

api/v1_projects.inc
70ray: When it gets into the javascript file it looks like this: (whichever way it was constructed in php) (only some of the componenets are expanded here)

0	Object { name: "projectID5690905894029", subsets: […] }
1	Object { name: "basic-latin", subsets: […] }
name	"basic-latin"
subsets	[ {…}, {…}, {…}, {…}, {…}, {…} ]
0	Object { name: "A-Z", title: "A-Z", rows: […] }
1	Object { name: "ˆ", title: "Letters with acute, grave, and circumflex", rows: […] }
2	Object { name: "~", title: "Letters with tilde and diaeresis; miscellaneous", rows: […] }
3	Object { name: "!", title: "Punctuation", rows: […] }
4	Object { name: "¶", title: "Symbols and Currency", rows: […] }
5	Object { name: "1", title: "Numbers and Math", rows: […] }
name	"1"
title	"Numbers and Math"
rows	[ […], […] ]
0	[ […], […], […], […], […], […], […], […], […], […] ]
1	[ […], […], […], […], […], […], […], […], […], […] ]

0	[ "#", "NUMBER SIGN" ]
1	[ "%", "PERCENT SIGN" ]
2	[ "°", "DEGREE SIGN" ]
3	[ "+", "PLUS SIGN" ]
4	[ "-", "HYPHEN-MINUS" ]
5	[ "=", "EQUALS SIGN" ]
6	[ "×", "MULTIPLICATION SIGN" ]
7	[ "÷", "DIVISION SIGN" ]
8	[ "±", "PLUS-MINUS SIGN" ]
9	[ "¬", "NOT SIGN" ]

Making Php objects more closely matches what the javascript file gets although it is quite likely that php encodes it in the way you suggested by first converting them into arrays.

70ray: It seems preferable to use classes, which are definitions of data structures, rather than php arrays which are sometimes maps. PhpStan seems to understand what the class data structure means which simplifies the situation.

pinc/CharSuites.inc
70ray: ... considering.

api/v1_projects.inc
cpeel: > It seems preferable to use classes

I'm fine with using classes for data structures. When we do so we need to think about good names for them and making sure we have parallel code with similar structures.

Why don't you get a new PR up that just does the conversion into classes for both the PickerSet and the VerbosePickerSet (and subsets). Let's review that, merge that, and then you can rev this.

Or drop the classes in this PR, get it in, and do a follow-up with the class conversion.

But lets please stop trying to iteratively boil the ocean in a PR. I'm not sure how we went from updating a PHPStan variable annotation to a class structure rewrite in a single PR.

pinc/CharSuites.inc
70ray: @jchaffraix , ehoing Casey's thanks for your explanation. While considering Casey's comment I noticed something that didn't seem quite right: Does phpStan just consider each file individually? or the rlation between them? e.g CharSuites.inc defines the function add_subset() which is used in charsuite-basic-latin.inc but the data $codepoints there doesn't seem to agree with the declaration in CharSuites.inc.

api/v1_projects.inc
70ray: I think it's called mission creep. I'll do what you suggest.

pinc/CharSuites.inc
jchaffraix: > @jchaffraix , ehoing Casey's thanks for your explanation. While considering Casey's comment I noticed something that didn't seem quite right: Does phpStan just consider each file individually? or the rlation between them?

PHPStan looks at the entire project, not individual files, as any PHP file can `include` any other.

> e.g CharSuites.inc defines the function add_subset() which is used in charsuite-basic-latin.inc but the data $codepoints there doesn't seem to agree with the declaration in CharSuites.inc.

WoW, this is embarrassing (and a good catch): due to a syntax error, those types were not enforced... The good news is that those were PHPStan annotations so it shouldn't have any impact on prod.

FYI this is what is wrong:

```
/** @param $codepoints (string|null)[] */
```

It should be written like some PHPDoc string:

```
/** @param (string|null)[] $codepoints */
```

I will whip up a PR to fix the errors I can find.

api/v1_projects.inc
70ray: Waiting till after #1431 is merged.

70ray:
Sandbox updated.

cpeel (approved):

srjfoo (approved):

jchaffraix (approved):

PR #1431, opened 2025-02-08
Fix syntax errors in PHPDocs

jchaffraix:
/** @param $var type */ is not a valid PHPDoc
and thus is not recognized by PHPStan.

Fixing those uncovered some mistyped values in
CharSuites.inc. Kudos to Ray for finding those
in the first place on PR #1430!

There should be no prod impact as those are
just PHPStan annotations.

Sandbox: https://www.pgdp.org/~jchaffraix/c.branch/julien_fix_phpdoc_param_syntax_errors

### DIFF
--- a/pinc/CharSuites.inc
+++ b/pinc/CharSuites.inc
@@ -8,12 +8,12 @@ class CharSuiteNotEnabledException extends Exception
 class PickerSet
 {
     public string $name;
-    /** @var array<string, (string|null)[]> */
+    /** @var (?string[])[] */
     private array $subsets;
     /** @var string[] */
     private array $titles;
 
-    /** @param $codepoints (string|null)[] */
+    /** @param (?string[])[] $codepoints */
     public function add_subset(string $name, array $codepoints, ?string $title = null)
     {
         $this->subsets[$name] = $codepoints;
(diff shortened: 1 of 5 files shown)


cpeel (approved):

chrismiceli (approved):

pinc/CharSuites.inc
```diff
@@ -8,12 +8,12 @@ class CharSuiteNotEnabledException extends Exception
 class PickerSet
 {
     public string $name;
-    /** @var array<string, (string|null)[]> */
+    /** @var (?string[])[] */
```
70ray, line 11: The first string should not be null (as it was). But there should be another level of array since below
```php
$this->subsets[$name] = $codepoints;
```
Although it doesn't seem to make any difference

pinc/CharSuites.inc
```diff
@@ -25,7 +25,7 @@ class PickerSet
         }
     }
 
-    /** @return array<string, string[]> */
+    /** @return (?string[])[] */
```
70ray, line 28: As above.

pinc/CharSuites.inc
jchaffraix: I don't think I understand this comment. There are `null` entries in the inner array, see e.g. https://github.com/DistributedProofreaders/dproofreaders/blob/master/pinc/charsuite-basic-greek.inc#L20

Did you mean something else I missed?

70ray: Sorry, I was getting confused. Approving it now.

70ray (approved):

PR #1432, opened 2025-02-09
Fix profile link in header

cpeel:
If the user is logged in and on a profile that that isn't their own the Profile link in the header should still point to the user's profile. [Task 2099](https://www.pgdp.net/c/tasks.php?action=show&task_id=2099)

Sandbox: https://www.pgdp.org/~cpeel/c.branch/fix-profile-link/

### DIFF
--- a/pinc/theme.inc
+++ b/pinc/theme.inc
@@ -409,7 +409,7 @@ function html_navbar()
         echo "  &#183; ";
 
         $account_links = [
-            ['text' => _("Profile"), 'url' => "stats/members/mdetail.php"],
+            ['text' => _("Profile"), 'url' => "stats/members/mdetail.php?id=" . $user->u_id],
             ['text' => _("Preferences"), 'url' => "userprefs.php"],
             ['text' => _("Log Out"), 'url' => "accounts/logout.php"],
         ];


pinc/theme.inc
```diff
@@ -409,7 +409,7 @@ function html_navbar()
         echo "  &#183; ";
 
         $account_links = [
-            ['text' => _("Profile"), 'url' => "stats/members/mdetail.php"],
+            ['text' => _("Profile"), 'url' => "stats/members/mdetail.php?id=" . $user->u_id],
```
jchaffraix, line 412: That seems fine. I guess a question here is whether `stats/member/mdetails.php` should point to the logged in user by default? This would remove the need to set the userID through the URL's query.

jchaffraix (approved):

pinc/theme.inc
cpeel: It does already, which is oddly the problem. The [`is_url_for_current_page()`](https://github.com/DistributedProofreaders/dproofreaders/blob/master/pinc/misc.inc#L1609) logic detects that the id-less page is a subset of the link and decides that it is the same page when it's actually not. We use that subset logic to make sure that the Round links correctly reflect that the user is on them.

jchaffraix: Sounds good, thanks for the explanation.

srjfoo (approved):

PR #1434, opened 2025-02-10
Lift automodify into a BackgroundJob

cpeel:
This lifts automodify into a normal BackgroundJob. In doing so it moves logic that was in `automodify.php` into `pinc/automodify.inc` so it can be called from within the new BackgroundJob. Unlike the prior version, the new job will save the text output directly to a log file so we no longer need the redirect in the crontab.

The `automodify.php` file is still there and can still be run by PMs for a single project (needed for some project transitions) and by SAs.

The crontab on TEST has been updated to use the job in the sandbox below for testing.

https://www.pgdp.org/~cpeel/c.branch/automodify-as-background-job/

### DIFF
--- a/SETUP/ci/check_require_login.php
+++ b/SETUP/ci/check_require_login.php
@@ -59,7 +59,6 @@ $ok_files = [
     // Tools
     "tools/setlangcookie.php",  // allows unauth users to set their language
     "tools/post_proofers/smooth_reading.php",
-    "tools/project_manager/automodify.php",  // requires localhost or login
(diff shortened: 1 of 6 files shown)


srjfoo (approved):

70ray (approved):

chrismiceli (approved):

cpeel:
Fixed a minor bug and incorporated PR feedback.

70ray (approved):

srjfoo (approved):

PR #1436, opened 2025-02-11
Remove jQuery from split control

70ray:
In preparation for re-styling the splitter bar

### DIFF
--- a/SETUP/tests/manual_web/split_test/switchable_split.js
+++ b/SETUP/tests/manual_web/split_test/switchable_split.js
@@ -1,5 +1,5 @@
 /* global $ splitControl */
-$(function () {
+window.addEventListener("DOMContentLoaded", function () {
     function appendControlButton(controlDiv, theSplit, splitVert) {
         let vSwitchButton = $("<input>", { type: "button", value: "Switch to Vertical Split" });
         let hSwitchButton = $("<input>", { type: "button", value: "Switch to Horizontal Split" });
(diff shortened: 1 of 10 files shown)


SETUP/tests/manual_web/split_test/switchable_split.js
```diff
@@ -41,7 +41,8 @@ $(function () {
     }
 
     let initialSplitVertical = false;
-    let mainSplit = splitControl("#container", { splitVertical: initialSplitVertical });
+    const container = document.getElementById("container");
```
chrismiceli, line 44: *minor*, I would just inline it with `splitControl(document.getElementById('...`. This happens the other instances too if you decide you want to make this change.

chrismiceli (approved):
This is great! I'll review again once you fix the js test issues too.

SETUP/tests/manual_web/split_test/switchable_split.js
70ray: Yes, good idea.

70ray:
Sandbox at: https://www.pgdp.org/~rp31/c.branch/unjq_split

chrismiceli (approved):

cpeel (approved):
Yay for fewer jQuery uses!

srjfoo (approved):

tools/project_manager/show_all_good_word_suggestions.js
```diff
@@ -1,7 +1,7 @@
 /*global splitControl */
 
 window.addEventListener("DOMContentLoaded", function () {
-    let mainSplit = splitControl("#suggestions_container", {
+    let mainSplit = splitControl(document.getElementById("suggestions_container"), {
```
jchaffraix, line 4: Just thinking out loud: we may want to add some helper function to shorten all those `document.getElementById` (I find myself adding it whenever I have to do some web work).

jchaffraix (approved):
That looks great, thanks!

tools/project_manager/show_all_good_word_suggestions.js
cpeel: That would make sense to me. jQuery's selection tool is really nice but we're trying to remove it from the codebase since it's not really needed for the browsers we support anymore. Doesn't mean we can't / shouldn't have some helpers to simplify common things like this.

PR #1440, opened 2025-02-21
Better handle banned account logins

cpeel:
We use phpBB for authentication by calling its `$auth->login()` function. Except that if the account is banned in the forums, that function never returns because `trigger_error()` is called which kills the page. We can install our own error handler to catch that and throw an exception which we _can_ handle more gracefully.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/handle-banned-accounts/

Testing notes: User authentication should continue to work as-is, and users with banned accounts who try to log in will now get a more reasonable error messages instead of an styled phpBB error page.

### DIFF
--- a/accounts/login.php
+++ b/accounts/login.php
@@ -51,6 +51,8 @@ if (!$success) {
         login_failure('unknown_failure', $destination);
     } elseif ($reason == 'too_many_attempts') {
         login_failure('too_many_attempts', $destination);
+    } elseif ($reason == 'banned') {
+        login_failure('banned', $destination);
     } else {
         login_failure('auth_failure', $destination);
     }
--- a/accounts/login_failure.php
+++ b/accounts/login_failure.php
@@ -32,7 +32,8 @@ $login_failures = [
     'form_timeout' => _("Form submission timeout, go back and try again."),
 ];
 
-$error = @$login_failures[@$_GET['error_code']];
+$error_code = @$_GET['error_code'];
+$error = @$login_failures[$error_code];
 if (!$error) {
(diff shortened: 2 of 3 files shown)


chrismiceli (approved):

70ray (approved):
Login works. Presumably one has to be a forum administrator to ban someone so I couldn't test how logging in as a banned user works.

srjfoo:
Tested twice to make sure I didn't do something wrong -- I banned one of my test accounts. 

My steps:
1. Logged in via a private window in Firefox, and a standard window in Chrome
2. Banned from my admin account
3. Went back to the page where the test account was logged in.
4. I was not logged out of the test site. I was even able (in both browsers) to proofread and save.
5. I was, however, required to re-login to the forums and was denied with the message that I was banned.
6. If I logged out, I was not permitted to log back in again.

cpeel:
> Tested twice to make sure I didn't do something wrong -- I banned one of my test accounts.
>     4. I was not logged out of the test site. I was even able (in both browsers) to proofread and save.
>     5. I was, however, required to re-login to the forums and was denied with the message that I was banned.
>     6. If I logged out, I was not permitted to log back in again.

These are all correct. Being banned from the forums logs the user out of the forums and prevents them from logging in to both the forums and the DP site but does not log them out of the DP site. This PR does not fix the last part (that will need to come after the OS upgrade) and right now still requires a manual DB command to drop the sessions.

What this PR does is improve the log in experience to the DP site for a user who has been banned.

srjfoo (approved):

PR #1441, opened 2025-02-22
Map sessions to users and allow admins to manage them

cpeel:
_Note: This will not be merged until after the code thaw in March._

In very rare cases admins may need to log a user out of the site. This makes that easier by including the username in the session table and a very simple UI for seeing the current sessions for a user and the option of logging them out.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/track-sessions-to-users/

Testing notes: when testing this, be sure to use the sandbox for the user you want to log out. Sessions created outside of this sandbox will not have the username associated with them and won't be logged out. When this code is deployed any site activity will add the username to the user's session thus solving the problem after deployment.

### DIFF
--- a/tools/site_admin/manage_user_sessions.php
+++ b/tools/site_admin/manage_user_sessions.php
@@ -1,0 +1,72 @@
+<?php
+$relPath = "./../../pinc/";
+include_once($relPath.'base.inc');
+include_once($relPath.'theme.inc');
+include_once($relPath.'User.inc');
+include_once('sa_common.inc');
+
+require_login();
+
+if (!user_is_a_sitemanager()) {
+    die(_("You are not authorized to invoke this script."));
+}
+
+// --------------------------------------
+
+$username = $_POST['username'] ?? ($_GET['username'] ?? null);
+$action = get_enumerated_param($_POST, 'action', null, ['logout'], true);
+
+$title = _("Manage User Sessions");
+output_header($title, NO_STATSBAR);
+
+echo "<h1>$title</h1>\n";
+echo "<p>" . _("This page shows how many separate sessions (logins) a user has and the option of logging them out.") . "</p>";
+
+show_username_form($username);
+
+if ($username) {
+    try {
+        $user = new User($username);
+    } catch (NonexistentUserException $exception) {
+        echo "<p class='error'>" . _("Invalid username") . "</p>";
+        exit;
+    }
+
+    echo "<hr>";
+    echo "<h2>$username ($user->real_name)</h2>";
+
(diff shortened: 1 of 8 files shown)


srjfoo (approved):

tools/site_admin/manage_user_sessions.php
```diff
@@ -0,0 +1,109 @@
...
+    }
+}
+
+// XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
+
+function show_username_form(?string $username): void
+{
+    echo "<form method='GET'>\n";
+    echo "<p>" . _("Username") . ": ";
+    echo "<input name='username' type='text' value='$username' size='20' autocapitalize='none' required>\n";
```
chrismiceli, line 50: do we need to encode username here?

chrismiceli (approved):

tools/site_admin/manage_user_sessions.php
```diff
@@ -0,0 +1,109 @@
...
+function delete_sessions(string $username): void
+{
+    $sql = sprintf(
+        "
+        DELETE FROM sessions
+        WHERE username = '%s'
+        ",
+        DPDatabase::escape($username)
+    );
+    DPDatabase::query($sql);
```
jchaffraix, line 106: Nit: Should we do error checking here? It should be a matter of checking if the result is `false`.

jchaffraix (approved):

tools/site_admin/manage_user_sessions.php
cpeel: Yes, we should. And this was a copy/pasta from another script. I've pushed up two commits (one will be squashed in a rebase later) that correctly escapes this and pulls the function into a separate file we can use here, in the original, and one other place in the `site_admin/*`.

tools/site_admin/manage_user_sessions.php
cpeel: Fair question and I don't think so. If the SQL ended up being invalid for some reason, the `query()` function will raise an exception (and log it). If there are no sessions this will still pass and just delete 0 rows. Pretty much all the code in the repo assumes we can talk to the database (ie: no transient network/socket failures) so that is a possible failure scenario we don't generally worry about.

70ray (approved):

srjfoo (approved):

cpeel:
Force-pushed an update to squash a fixup commit. No change to the code in the PR.

Issue #1256, opened 2024-06-28
Implement a new site configuration approach


cpeel:
To configure the DProofreaders code, right now one edits a shell script -- an example of which is in `./SETUP/configuration.sh` -- and then runs the `./SETUP/configure` script. That script searches the code for files ending in `.template` which contain variable placholders. It then creates new files without the `.template` extension with the values filled in. We used to have many `.template` files but now we have just these:
```
./pinc/udb_user.php.template
./pinc/site_vars.php.template
./SETUP/check_db_schema.template
./SETUP/dp.cron.template
./SETUP/dump_db_schema.template
```
You'll note that they are a combination of PHP scripts and shell scripts.

This approach has its pros and cons. For instance, on TEST we have a single configuration file that all developers can use for their sandbox and the shell script determines the right paths based on the user.

A downside is that it makes upgrades harder as one must inspect the new `./SETUP/configuration.sh` files for any new (or deleted) variables and manually update the system copy. Trying to use code with a new config variable that isn't in the old site will very likely result in errors because there is no way of setting up default values for them. The `SETUP/import_old_configuration.php` attempts to assist with the upgrade but it's a rough hack.

Another downside is that it means that it isn't dynamically extensible. Because we can't provide defaults we can't add new overrides easily. On TEST we'd like to decrease the number of pages which makes a user a "new" user but that value is hard-coded. We could make it a variable but that then requires us to add it to the configuration file and then becomes an upgrade issue. If we could set it to be a default value in the code and override it just for TEST that would solve that problem.

Other PHP projects have solved this problem, and we can too. MediaWiki, for instance, introduces new config variables all the time and you just need to set/override them in `LocalSettings.php`. One could also imagine having a configuration class that would load a PHP-based configuration file for settings and callers would ask that class for config values (bonus: fewer global variables!).

Something to keep in mind is that we do not want to introduce a new file read on every page load. That is: if we have a new non-PHP configuration file that some script has to load and parse that's probably not good. If it's a PHP file it will get parsed and loaded into the in-memory opcache which would make it fast. I think phpBB does something clever by creating PHP files dynamically in a cache to get the opcache benefit.

cpeel:
I took a look at [Symfony Config](https://symfony.com/doc/current/components/config.html) and it seems very heavy for what we want.

PR #1445, opened 2025-03-04
Be explicit about site globals

cpeel:
Don't assume because something is declared outside a function that it is a global. Be explicit about globals being global when we define them.

This is a forward-looking PR in preparation for upgrading to phpunit 10 and won't be merged until after the PROD upgrade and we lift the code freeze.

### DIFF
--- a/pinc/DPage.inc
+++ b/pinc/DPage.inc
@@ -463,6 +463,7 @@ function Page_clearRound(string $projectid, string $image, Round $round, string 
 
 // The numeric values of the following array must not change as they are
 // saved in the database to indicate the bad page reason.
+global $PAGE_BADNESS_REASONS;
 $PAGE_BADNESS_REASONS = [
     1 => [
         "name" => "missing_image",
--- a/pinc/faq.inc
+++ b/pinc/faq.inc
@@ -6,6 +6,7 @@ include_once($relPath.'metarefresh.inc');
 // that is used in FAQs. Also information about forums and wiki used in FAQs.
 
 // SITE-SPECIFIC
+global $external_faq_overrides;
 $external_faq_overrides = [
     "proofreading_guidelines.php" => [
         "en" => "https://www.pgdp.net/wiki/DP_Official_Documentation:Proofreading/Proofreading_Guidelines",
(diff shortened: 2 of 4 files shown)


srjfoo (approved):

pinc/DPage.inc
```diff
@@ -463,6 +463,7 @@ function Page_clearRound(string $projectid, string $image, Round $round, string
 
 // The numeric values of the following array must not change as they are
 // saved in the database to indicate the bad page reason.
+global $PAGE_BADNESS_REASONS;
```
jchaffraix, line 466: If I read [the php page on global](https://www.php.net/manual/en/language.variables.scope.php), this usage of `global` in the global scope is only useful if we want to include `pinc/DPPage.inc` inside a function. Is that something we want to support?

chrismiceli (approved):

pinc/DPage.inc
cpeel: Your reading is correct. The trick is how phpunit pulls in files is not always at the global context so tests fail because we didn't actually define the globals as globals.

cpeel: Because what I _think_ happens is that during some test the autoloader is being invoked from within a function, and the autoloader loads the file within the function, and that file may load other files (like `DPage.inc`) which is then loading the variable in the function's context rather than the global context.

jchaffraix: OK, that's a bit sad that we need to build this work-around for our tests when it's not something we wouldn't want to support. Thanks for updating those!

jchaffraix (approved):

pinc/DPage.inc
cpeel: Yeah. The solution is to stop using globals because it's a bad practice, but that's a ways out.

cpeel:
I'm going to scope this down just a bit (mostly remove the changes to `site_vars.php.template`) because I have some ideas on https://github.com/DistributedProofreaders/dproofreaders/issues/1256 that might address some of this more systemically. PR update to follow later.

srjfoo (approved):

PR #1448, opened 2025-03-08
Remove authors / biographies code

cpeel:
Of the 45k projects done at pgdp.net, only 80 have ever used this in the last 20 years and the effort to maintain it isn't worth the value.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/post-R202503-release-remove-authors-code/


70ray (approved):

chrismiceli (approved):
No img type resources or css rules we can cleanup? I tried to quickly glance but nothing stood out.

cpeel:
> No img type resources or css rules we can cleanup? I tried to quickly glance but nothing stood out.

Oh, very good question! I went back and looked through the code and I didn't find any custom classes.

Thanks for bringing it up as I never even considered it.

srjfoo (approved):
Note that until the upgrade script is run, the sandbox will show biographies like so:
image

After the script is run, they should show up like so:
image

Screenshots taken from TEST project projectID5899f94438377 (which I set up for another purpose long ago, and was actually able to find pretty quickly).

srjfoo:
I just realized, `authors_db_manager` needs to come out of [tools/site_admin/manage_site_access_privileges.php](https://github.com/DistributedProofreaders/dproofreaders/blob/926100b388643c48a897d764cf6a9e1262a6c31f/tools/site_admin/manage_site_access_privileges.php#L30). Should that be added here or taken out in a different PR?

cpeel:
> I just realized, `authors_db_manager` needs to come out of [tools/site_admin/manage_site_access_privileges.php](https://github.com/DistributedProofreaders/dproofreaders/blob/926100b388643c48a897d764cf6a9e1262a6c31f/tools/site_admin/manage_site_access_privileges.php#L30). Should that be added here or taken out in a different PR?

Good catch - it should be in this PR. Additional commit incoming.

srjfoo (approved):

PR #1449, opened 2025-03-08
Remove FTP-based file uploads support

cpeel:
pgdp.net hasn't allowed FTP file uploads for almost a decade and even sharing credentials for something like SCP is not a best practice, so remove it altogether.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/remove-ftp-uploads/

### DIFF
--- a/tools/project_manager/remote_file_manager.php
+++ b/tools/project_manager/remote_file_manager.php
@@ -122,15 +122,9 @@ $curr_relpath = get_current_dir_relative_path($home_dirname);
 $curr_abspath = "$uploads_dir/$curr_relpath";
 // The absolute path of the current directory.
 
-// $uploads_account is not required and may not be defined if they have not
-// set up FTP uploads. If it's blank, lets use a different string instead.
-// The point is to provide a "nice" representation of the absolute path
-// that doesn't reveal details of the upper reaches of the filesystem.
-if ($uploads_account) {
-    $curr_displaypath = "~$uploads_account/$curr_relpath";
-} else {
-    $curr_displaypath = "~uploads/$curr_relpath";
-}
+// Provide a "nice" representation of the absolute path that doesn't reveal
+// details of the upper reaches of the filesystem.
+$curr_displaypath = "~uploads/$curr_relpath";
 
 // For convenience, here are a couple of encoded forms:
 $hae_curr_relpath = attr_safe($curr_relpath);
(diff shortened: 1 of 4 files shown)


chrismiceli (approved):

jchaffraix (approved):
We should put a note to remove the account for `uploads_account` on the dev server (and probably prod). Pretty sure, you already have an item for this but better safe than sorry!

cpeel:
> We should put a note to remove the account for `uploads_account` on the dev server (and probably prod). Pretty sure, you already have an item for this but better safe than sorry!

The account on TEST and PROD that were originally used for FTP way back in the day are where the uploads for `remote_file_manager.php` go today, so we actually don't want to remove it. The account hasn't been enabled for login since we disabled FTP access though. But good check!

srjfoo:
> > We should put a note to remove the account for `uploads_account` on the dev server (and probably prod). Pretty sure, you already have an item for this but better safe than sorry!
> 
> The account on TEST and PROD that were originally used for FTP way back in the day are where the uploads for `remote_file_manager.php` go today, so we actually don't want to remove it. The account hasn't been enabled for login since we disabled FTP access though. But good check!

Should we harmonize TEST and PROD? Currently their uploads paths are different.

tools/project_manager/remote_file_manager.php
```diff
@@ -122,15 +122,9 @@
...
-// The point is to provide a "nice" representation of the absolute path
-// that doesn't reveal details of the upper reaches of the filesystem.
-if ($uploads_account) {
-    $curr_displaypath = "~$uploads_account/$curr_relpath";
-} else {
-    $curr_displaypath = "~uploads/$curr_relpath";
-}
+// Provide a "nice" representation of the absolute path that doesn't reveal
+// details of the upper reaches of the filesystem.
+$curr_displaypath = "~uploads/$curr_relpath";
```
srjfoo, line 127: I just want to make sure that I'm understanding the code correctly. `$uploads_dir` is still `/home/dpscans` (at least on PROD), but the form on the project page will display `~uploads/` instead of `~dpscans/` after this change goes in?

cpeel: Correct, it's just a name -- nothing functionally changes because the code isn't using that name for anything except showing it to the user.

srjfoo: Will need to be documented for squirrels not to pay any attention to what it says on the project page :D -- which on TEST is already the case, anyway.

srjfoo (approved):

tools/project_manager/remote_file_manager.php
cpeel: If it ends up being too confusing we an always hotfix the label in `pgdp-production`.

PR #1450, opened 2025-03-09
Prototype for new site configuration infrastructure

cpeel:
This PR is a prototype for a new site configuration infrastructure (see https://github.com/DistributedProofreaders/dproofreaders/issues/1256) that will allow us to:
1. stop using global variables -- beyond not using globals being a best practice, this makes phpunit much happier
2. have the configuration variables be typed (yay!)
3. allow variables to have default values, allowing us to add new ones more easily

Now, instead of the config values being in the global context, they are only accessible in the SiteConfig static object, so:
```php
global $preceding_proofer_restriction;
echo $preceding_proofer_restriction;
```
becomes
```php
echo SiteConfig::$preceding_proofer_restriction;
```

The new method re-uses the existing `site_vars.php` file which means there are no changes required to how the code is configured, just how we use those configuration variables. If we went forward with this approach, a second PR would later move away from the bash-based configuration model to just using the PHP file directly.

Keeping the configurations in a PHP file was intentional because PHP files are cached in the opcache and do not need to go to disk or be processed upon access, unlike a yaml, json, or ini-based file.

I have a fully-converted branch ([site-config](https://github.com/cpeel/dproofreaders/tree/site-config)) that updates all config variables uses (with the exception of the 4 mentioned at the end of `SiteConfig::load()`) passes all unit tests and smoke tests but it's huge and I wanted to focus the initial review on the proposed system rather than all of the individual changes. If we agree that this approach makes sense I'll close this PR and open a new one with the full set of changes.

Specific questions I'd like your feedback on:
* Thoughts on the `SiteConfig` class, both as a structure, its status as a static (ie: singleton).
* The verbosity of needing `SiteConfig::` prefixed to use the configs -- this explicitness seems like a good thing to me, but it is wordier.

I evaluated using an existing package like Symfony's [Config](https://symfony.com/doc/current/components/config.html) package but it's huge and very complex and I don't think we need that complexity.

### DIFF
--- a/pinc/SiteConfig.inc
+++ b/pinc/SiteConfig.inc
@@ -1,0 +1,144 @@
+<?php
+
+class SiteConfig
+{
+    private static bool $_is_loaded = false;
+
+    // alert messages
+    public static bool $maintenance = false;
+    public static string $maintenance_message = "";
+    public static string $alert_message = "";
+
+    // database configuration
+    public static string $db_server = "localhost";
+    public static string $db_user;
+    public static string $db_password;
+    public static string $db_name;
+
+    // archiving configuration
+    public static ?string $archive_db_name;
+    public static ?string $archive_projects_dir;
+
+    // directories & URLs
+    public static string $code_dir;
+    public static string $code_url;
+    public static string $projects_dir;
+    public static string $projects_url;
+    public static string $dyn_dir;
+    public static string $dyn_url;
+    public static string $dyn_locales_dir;
+    public static ?string $blog_url;
+    public static ?string $wiki_url;
+
+    // site identification
+    public static string $site_name;
+    public static string $site_abbreviation;
+    public static string $site_url;
+
+    // forums
+    public static string $forum_type = "phpbb3";
+    public static ?string $forums_phpbb_table_prefix;
+    public static ?string $forums_phpbb_dir;
+    public static ?string $forums_phpbb_url;
+    public static ?string $forums_json_users;
+    public static ?string $forums_json_posts = null;
+
+    // topic forums
+    public static ?int $beginners_site_forum_idx;
+    public static ?int $waiting_projects_forum_idx;
+    public static ?int $projects_forum_idx;
+    public static ?int $pp_projects_forum_idx;
+    public static ?int $posted_projects_forum_idx;
+    public static ?int $deleted_projects_forum_idx;
+    public static ?int $completed_projects_forum_idx;
+    public static ?int $post_processing_forum_idx;
+    public static ?int $teams_forum_idx;
+
+    // proofreading controls
+    public static string $preceding_proofer_restriction = 'not_immediately_preceding';
+    public static bool $public_page_details = false;
+
+    // uploads
+    public static string $uploads_dir;
+    public static string $uploads_subdir_trash = "TRASH";
+    public static string $uploads_subdir_commons = "Commons";
+    public static string $uploads_subdir_users = "Users";
+    public static string $antivirus_executable;
+
+    // WordCheck & locales
+    public static string $aspell_executable = '/usr/bin/aspell';
+    public static string $aspell_prefix = "/usr";
+    public static string $system_locales_dir = '/usr/share/locale';
+
+    // API
+    public static bool $api_enabled = true;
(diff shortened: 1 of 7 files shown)


pinc/SiteConfig.inc
```diff
@@ -0,0 +1,144 @@
+<?php
+
+class SiteConfig
+{
+    private static bool $_is_loaded = false;
+
+    // alert messages
+    public static bool $maintenance = false;
```
jchaffraix, line 8: I wonder if we could make the `SiteConfig` object immutable.

Unfortunately, I don't think this is possible with all the fields `static` as we would need a specific instance. This could be done through a singleton (e.g. `SiteConfig::get()`) and making the fields `readonly`, this would be a slightly longer than the current scheme but it would enforce that the fields can't be changed mid-request.

pinc/SiteConfig.inc
```diff
@@ -0,0 +1,144 @@
...
+            return;
+        }
+
+        try {
+            require $filename;
+        } catch (Error $error) {
+            throw new RuntimeException("Configuration file $filename not found");
+        }
+
+        $ro = new ReflectionObject(new SiteConfig());
```
jchaffraix, line 114: Most languages warn against using reflection as it tends to be unoptimized. Not sure where PHP stands on this and how much we care about it in practice... Thought I will bring it so we're aware of this potential issue.

The alternatives have other downsides:
1. an explicit list (probably with a test to make sure it stays up-to-date)
2. some autogenerator (which is probably going down the Symphony's `Config` route)

pinc/bootstrap.inc
```diff
@@ -49,7 +49,7 @@ if (!defined('SKIP_DB_CONNECT')) {
     } catch (Exception $e) {
         // If we're in maintenance mode, don't die here - we'll more gracefully
         // error out later
-        if (!$maintenance) {
+        if (!SiteConfig::$maintenance) {
```
jchaffraix, line 52: The prefix is fine. As you pointed it, it is worth being explicit and the overhead seems small to me.

jchaffraix (commented):
I think it's a step in the right direction!

One high-level comment would be whether we want to include all the variables under a single object: we used to have the DB configuration by itself and could use N objects grouped based on some criteria as we move those variables out of the global scope.

pinc/SiteConfig.inc
cpeel: I originally started with an object-based scheme which meant accessing things was `SiteConfig::get()->value` which seemed more unwieldy than `SiteConfig::$value` when I started using it.

re: immutability - we'd need to think about how to make default values work since we can't set them and override them. I can probably figure that out. We'll also need to figure out how to handle some unit test that do need to override these (which is decidedly an anti-pattern but also where we are right now).

It's a good suggestion, let me noodle over it.

pinc/SiteConfig.inc
cpeel: Some rough performance benchmarking says we don't need to worry about this use of reflection:
```
Reflection: 100000 iterations completed in 8650342746 nanoseconds, or 86503.42746 ns per iteration
Reflection: 100000 iterations completed in 8850424277 nanoseconds, or 88504.24277 ns per iteration

No reflection: 100000 iterations completed in 8419315452 nanoseconds, or 84193.15452 ns per iteration
No reflection: 100000 iterations completed in 8783086792 nanoseconds, or 87830.86792 ns per iteration
```

pinc/SiteConfig.inc
cpeel: I think I've figured out how to make default values work with `readonly`, but as I think more about our unit tests and wanting to test variations I'm unclear how we do that without either making the config modifiable at runtime or allowing the creation, and somehow use, of a new SiteConfig object.

I'm not sure enforcing immutability is important enough for the effort. Thoughts?

cpeel:
A second prototype that uses a singleton class with `readonly` properties is available at https://github.com/DistributedProofreaders/dproofreaders/pull/1465

I don't love how the default values are separate from the definitions and phpstan is unhappy because it can't figure out (justifiably so) that we are setting the values of the `readonly` value in the constructor, but those might be worthwhile.

PR #1456, opened 2025-03-13
Prototype proofreading interface

70ray:
This is incomplete. There is minimal styling. The styles in the file struct.css are important for the structure of the pages but do not specify any colours (except to avoid transparent which would allow what is underneath to show through). There are several unused files remaining.

### DIFF
--- a/quiz/generic/main.php
+++ b/quiz/generic/main.php
@@ -2,6 +2,8 @@
 $relPath = '../../pinc/';
 include_once($relPath.'base.inc');
 include_once($relPath.'slim_header.inc');
+include_once($relPath.'draw_toolbox.inc');
+include_once($relPath.'widget_text.inc'); // $widget_text
 
 require_login();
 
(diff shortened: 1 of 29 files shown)


quiz/generic/main.php
```diff
@@ -10,14 +12,65 @@
 include "./quiz_page.inc"; // qp_round_id_for_pi_toolbox
 
 $header_args = [
+    "css_files" => [
+        "$code_url/styles/struct.css",
+        "https://cdn.jsdelivr.net/npm/quill@2.0.3/dist/quill.core.css",
```
chrismiceli, line 17: should we just serve js/css from our domain since our user base is very sensitive to external resources?

chrismiceli (commented):
Going to review more closely, but not getting too critical until we align on design etc.

quiz/generic/main.php
70ray: Yes we should, It didn't seem very clear to me how to download the files so I took the easy way and had to change pinc/base.inc to make this work.

chrismiceli: Looks like the css nor js have any imports or external references, so should just be able to paste into our external resources folder (/pinc/3rdparty) with appropriate license and details.json file.

cpeel:
This branch laid the *extensive* groundwork and all the heavy lifting for the new proofreading interface that we're currently iterating over -- thank you @70ray!

Closing this draft as the work continues in other branches on top of this code.

PR #1462, opened 2025-03-16
Further type pinc/Project.inc

cpeel:
This adds more typing to `pinc/Project.inc`.

The biggest functional change is removing `$project->check_pages_table_exists($detail)` with the pass-by-reference to get the reason and having callers explicitly call `$project->pages_table_exists` to check and if they want the reason why to call `$project->pages_table_missing_reason()`.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/post-R202503-release-type-project-inc

### DIFF
--- a/pinc/Project.inc
+++ b/pinc/Project.inc
@@ -239,7 +239,8 @@ class Project
     public string $extra_credits;
     public int $int_level;
 
-    public function __construct($arg = null)
+    /** @param string|array<string, mixed>|null $arg */
+    public function __construct(mixed $arg = null)
     {
         if (is_string($arg)) {
             // $arg is the projectid.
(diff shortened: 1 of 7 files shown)


chrismiceli (approved):

70ray (approved):

pinc/Project.inc
```diff
@@ -1917,7 +1936,7 @@ function project_get_auto_PPer($projectid)
  * @param string $login_name Username
  * @param string $activity Should be one of 'cp', 'pm', 'pp', 'ip' and 'tp'.
```
jchaffraix, line 1937: We can restrict the type based on the comment using some PHPStan magic:

```suggestion
 * @param 'cp'|'pm'|'pp'|'ip'|'tp' $activity
```

pinc/Project.inc
```diff
@@ -1917,7 +1936,7 @@ function project_get_auto_PPer($projectid)
  * @param string $login_name Username
```
jchaffraix, line 1936: Since we pass the type now, we can remove this PHPStan annotation:

```suggestion
```

jchaffraix (approved):

srjfoo (approved):

PR #1470, opened 2025-04-13
Suppress the image flash when opening the page image viewer

mrducky4:
Is it still worthwhile to try to fix this image flash issue? I saw the discussion about this, so thought I might have an idea for another solution. I tested only with Chrome on Windows, I don't have a Mac or iPad.

The idea of this fix is to calculate the image styling immediately when adding the image source URL, so by the time the browser has a chance to render the image the styling is already calculated.

_Update from cpeel_: sandbox at https://www.pgdp.org/~cpeel/c.branch/no_image_flash/

### DIFF
--- a/scripts/control_bar.js
+++ b/scripts/control_bar.js
@@ -548,6 +548,7 @@ function makeImageWidget(container, align = "C") {
             sine = 0;
             cosine = 1;
             image.src = src;
+            setImageStyle();
         },
 
         reScroll: reScroll,


cpeel:
I defer to @70ray and @chrismiceli on the fix as I don't grok the JS.

I can't remember if someone from the dev team was able to reproduce the flashing but we should get someone who was able to reproduce the weird Safari behavior to test this if it seems to work for us.

mrducky4:
I can reproduce the flash that jjz reported [here](https://www.pgdp.net/phpBB3/viewtopic.php?p=1360699#p1360699). I don't see it any more after my change.
I bet IsolationBooth would be willing to test with Safari.

srjfoo:
> I can reproduce the flash that jjz reported [here](https://www.pgdp.net/phpBB3/viewtopic.php?p=1360699#p1360699). I don't see it any more after my change. I bet IsolationBooth or Chapka would be willing to test with Safari.

I'm in the process of writing a PM to several who experienced the problem -- the more the better.

chrismiceli (approved):
This looks ok to me in theory, but I am not 100% familiar enough with the page browser flows to confirm this would handle all scenarios well.

70ray (commented):
I can't reproduce the problem now. I think I did once but I can't remember how. This change should not have any bad effects.

srjfoo (approved):
I've heard back from 4 of the 5 that I recruited that this takes care of the image flashing problem, without the weird image effects they saw after last Thursday's deployment, before we rolled back the earlier fix for the image flashing.

The first three confirmed that it fixed things as far as they were concerned.

The fourth confirmed that the image flash was fixed, but also did a more extensive test, which just confirmed that the image moving to the right is still present on the iPad if you switch apps or windows and then switch back. Though oddly, only to the bottom or the right, depending on which edge of the image is larger than the viewport. Since this rougly matches the existing behavior, I don't think it's a show-stopper, because that wasn't the problem we were trying to fix.

cpeel (approved):

Issue #843, opened 2022-10-02
Ability for squirrels to change project images via script


srjfoo:
One of the most common tasks a squirrel does is replace proofing images, or add, delete or replace illustration (non-proofing) images. Sometimes just a handful, but often in bulk. Currently, this must be done via command line.

After discussing the idea with @cpeel , it seems that the best place to delete non-proofing images would be to extend the existing functionality in `tools/proofers/images_index.php` to allow squirrels to delete non-proofing images either selected or in bulk in any project state (obviously other than deleted or archived). Or to replace images of the same name. (This should be restricted just to the non-proofing images.)

The other functionality that is needed is the ability to replace existing proofing images in bulk. This could mean replacing all proofing images, or just a partial replacement, but more than is easily replaced using `handle_bad_page` one at a time. This can probably also be done by adding admin functionality to an existing script, but I'm not sure where.

cpeel:
> This can probably also be done by adding admin functionality to an existing script, but I'm not sure where.

Yeah, there's not a great existing place for these to go, just a handful of not-great-but-maybe-not-terrible places:
* `add_files.php` is what does the initial load and does some handling of bulk images, but this also deals a lot with text and I'm dubious that this is the right place.
* `remote_file_manager.php` is perhaps a logical place as it deals with file management, but right now it's all outside of the project directory (by design) and also only deals with individual files.

It is probably worth investigating `images_index.php` for the illustrations case and seeing if there's an opportunity for the page images case as well.

srjfoo:
> > This can probably also be done by adding admin functionality to an existing script, but I'm not sure where.
> 
> Yeah, there's not a great existing place for these to go, just a handful of not-great-but-maybe-not-terrible places:
> 
>  * `add_files.php` is what does the initial load and does some handling of bulk images, but this also deals a lot with text and I'm dubious that this is the right place.

Yeah, that's the problem with using `add_files.php`

>  * `remote_file_manager.php` is perhaps a logical place as it deals with file management, but right now it's all outside of the project directory (by design) and also only deals with individual files.

Right, so far, `remote_file_manager.php` deals just with the `dpscans` uploads directory.

> It is probably worth investigating `images_index.php` for the illustrations case and seeing if there's an opportunity for the page images case as well.

Sounds reasonable.

srjfoo:
> > This can probably also be done by adding admin functionality to an existing script, but I'm not sure where.
> 
[...]

PR #1472, opened 2025-04-15
Allow SAs to change project images in any state

cpeel:
Allow SAs to change project page images (and illustrations) regardless of the state of the project. Prevent them from changing project page texts however. Fixes https://github.com/DistributedProofreaders/dproofreaders/issues/843

Sandbox: https://www.pgdp.org/~cpeel/c.branch/sa-manage-images/

### DIFF
--- a/pinc/project_edit.inc
+++ b/pinc/project_edit.inc
@@ -7,6 +7,11 @@ function user_can_add_project_pages($projectid)
     $project = new Project($projectid);
     $state = $project->state;
 
+    // SAs can futz with project pages in any state
+    if (user_is_a_sitemanager()) {
+        return true;
+    }
+
     // Load text+images from uploads area into project.
     // Can do this if it's a new project (as measured by the state it's in)
     // If the user is disabled from uploading new projects, they can only
--- a/tools/project_manager/add_files.php
+++ b/tools/project_manager/add_files.php
@@ -117,7 +117,8 @@ if (!$r) {
     return;
 }
 
-$loader = new Loader($source_project_dir, $project->dir, $projectid);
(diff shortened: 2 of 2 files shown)


cpeel:
@srjfoo - can you confirm that this really does fix all of the things identified in the issue?

70ray (approved):

chrismiceli (approved):

srjfoo:
> @srjfoo - can you confirm that this really does fix all of the things identified in the issue?

As far as I can tell, yes, thank you!

I've tossed a couple of links to the other squirrels, in case they want to go play in the sandbox.

cpeel:
Update to move the warning into `display()` so it doesn't show up at the top of `execute()`.

susanskinner:
I've tested this and am happy that it's working as I would expect. I've tested projects in the rounds and in PP (available, checked out or unavailable). I also tried adding an image to a "posted" project, [https://www.pgdp.org/~cpeel/c.branch/sa-manage-images/project.php?id=projectID4afbd088a2e1b](https://www.pgdp.org/~cpeel/c.branch/sa-manage-images/project.php?id=projectID4afbd088a2e1b), and it failed (as I would expect) because the project directory no longer existed.

This is going to making fixing projects much easier - thank you!

In an ideal world, this would do similar checks for illustration files as it does for proofing images and give warnings when it's going to over-write an existing file or when you trying to replace a file with an identical file but that's definitely a potential future enhancement and not a reason to not roll this out in its current form.

srjfoo (approved):
Definitely very happy with it.

I tested a few different things from Susan, but I think between us, everything is pretty well covered.

Issue #1474, opened 2025-04-16
Formalize `$dyn_dir/download_tmp/`, including BackgroudJob to clean it up


cpeel:
`tools/download_images.php` creates temporary zips of image files in `$dyn_dir/download_tmp/` as a cache. The `download_images.php` script creates the directory if it doesn't exist and nothing manages it (ie: cleans it up). On PROD, there is the following cronjob as root:
```
# Clean up download tmp directory
35 * * * * /usr/bin/find /data/htdocs/d/download_tmp -cmin +720 -exec rm {} \;
```

This directory needs to be fully managed by the code including a BackgroundJob to clean it up.

This directory could possibly be used as a staging area for other downloads (possibly as a place to generate transient zipped text files that contain proofreader names for https://github.com/DistributedProofreaders/dproofreaders/issues/1471 -- although with unguessable filenames) pending what the PPers and squirrels need on that front.

Issue #1471, opened 2025-04-13
Download project text / images via API


tangledhelix:
I have scripts to set up a new PP project on my computer, using the `projects` API to fetch metadata (`title`, `author`, `forum_url`, ...) and I would like to fetch the project text and images this way as well. I had been doing this by faking a web login session and using the resulting cookie, but my script is broken now that CSRF token was added to the login process.

Is it possible to add API endpoints to fetch the text and images? These would be the "Download Zipped Images" and "Download Zipped Text" links on the project page.

cpeel:
We probably want to implement this as returning URLs that the client can download (and handling whatever it takes to generate the content and access to those URLs) rather than returning the data directly to adhere to our JSON input/output model.

If we do decide to return the data directly we should consider if we want one central endpoint that is used for downloads so we have one exit point for non-JSON responses.

cpeel:
Implementation notes:
* "Download Zipped Text" points to a static content URL, so nothing we need to do in the API beyond returning the URL.
* "Download Zipped Images" points to `tools/download_images.php` which ultimately redirects the user to a static content URL after ensuring that the images zip file exists and is up-to-date. The API just needs to do this prep work and return the URL.

cpeel:
@tangledhelix - can you tell me how PPers interact with the "Download Zipped Text" download? Do they want the auto-generated file that doesn't contain usernames or do they want one with usernames? I _think_ that PPers can't regenerate a version that is saved on disk that contains the username and only SAs can do that (which seems like a flaw I need to fix).

cpeel:
@srjfoo - what is the squirrel expectations of the saved zipped text file having usernames? SAs seem to be the only people who can generate a zipped file that contains the usernames but after it's there anyone can download it if they know where to look. I think the solution is to prevent saving the file to disk if it has usernames in it.

cpeel:
And tagging @windymilla for his thoughts re: PP expectation / usage of usernames in the PP zipped download file.

srjfoo:
Usernames used to be included in thd PP text long ago. For privacy reasons, this was changed so that the stored CTF does not include usernames, but the PP can still generate one if they need it.

srjfoo:
[...]

PR #1475, opened 2025-04-19
Adds new API endpoint to get PP artifacts

cpeel:
Adds a new API endpoint to get PP artifacts. This massages some code around to support this. Notable changes (in separate commits):
* Prevents SAs from saving CTF files to disk with usernames. Doing so leaks proofreader usernames in an unauthenticated space. This is the sole functional change to the existing code.
* Moves `post_files.inc` to `/pinc` and moves the logic that generates project image zips to that file for re-use.
* Adds a new BackgroundJob to clean up the `download_tmp` directory instead of relying on a cronjob. Project image zips are cached to this location and the lack of an in-code cleanup is a 20-year oversight. Fixes https://github.com/DistributedProofreaders/dproofreaders/issues/1474
* Finally, add a new API endpoint to get project artifacts. This was designed to allow fetching of artifacts from various project stages but is currently limited to PP. This can be extended to support the other PP artifacts, PPV artifacts, SR, and is how we should make CTF downloads accessible in the rounds as well. Fixes https://github.com/DistributedProofreaders/dproofreaders/issues/1471

The API can be tested on TEST (using an `API_KEY` that has the right privs) with:
```bash
export API_KEY=some-key
    -X GET "https://www.pgdp.org/~cpeel/c.branch/api-pp-artifacts/api/v1/projects/projectID4aeffc955cc31/artifacts/PP"
```
Returning
```json
[
    {
        "stage": "PP",
        "name": "images_zip",
        "url": "https://www.pgdp.org/d/download_tmp/projectID4aeffc955cc31_images.zip"
    },
    {
        "stage": "PP",
        "name": "text_zip",
        "url": "https://www.pgdp.org/projects/projectID4aeffc955cc31/projectID4aeffc955cc31.zip"
    }
]
```

Sandbox at: https://www.pgdp.org/~cpeel/c.branch/api-pp-artifacts/


cpeel:
FYI @windymilla in case this is at all useful in GG2 (or gives you ideas for other DP things that you might want API access to in GG2).

api/v1_projects.inc
```diff
@@ -372,6 +372,48 @@ function render_project_json(Project $project, ?array $return_fields = null)
...
+function api_v1_project_artifacts(string $method, array $data, array $query_params)
+{
+    // get the project and the stage
+    $project = $data[":projectid"];
+    $stage = $data[":stageid"];
+
+    // TODO: This check might be too strict -- do we really care if a user
+    // is able to access artifacts from a stage they can't *work* in?
+    // We only show PP artifacts for users who can access in the stage, but
+    // that doesn't seem strictly necessary.
```
srjfoo, line 388: Based on a quick look, probably we don't care, unless we want to tighten up on what we allow already. I just checked using one of my alternate TEST IDs that doesn't have PP access. I can't see the PP downloads on a project in the PP pool, but:
* I can see the PP pool page
* I can pull a copy of the CTF with the latest text, which is equivalent to the PP file (except missing the download of the PCs, the GWL and the BWL)
* I can view images online, and download them from the images index.

tangledhelix (approved):
I did not review the code itself, since I'm not familiar enough with this codebase to have any real opinions. I reviewed only the API access from a user's POV. It worked well, returning the links needed, and when those were further fetched with the same API key (and `Accept: application/zip`) everything worked as I'd expect. I used Python `requests` instead of the curl commands noted in the comments.

Minor note just for your information: some of the projects in the TEST PP pool have CTF with usernames. I regenerated one of them without usernames, for my test. I didn't check or regenerate any others. (I tested with projectID44ddaafe94131.)

cpeel:
Thanks for testing it, Dan!

> when those were further fetched with the same API key (and `Accept: application/zip`) everything worked as I'd expect.

Note that the URLs being returned are static content served directly by Apache and should not require any API keys (clearly doesn't hurt to have them in the header, but they are not required).

> Minor note just for your information: some of the projects in the TEST PP pool have CTF with usernames. I regenerated one of them without usernames, for my test. I didn't check or regenerate any others. (I tested with projectID44ddaafe94131.)

Thanks. Yeah, TEST is a hot mess of various pieces of data and it does not surprise me that there are CTFs with usernames in them since we use to allow that and this VM's contents have been around for around a decade.

srjfoo (approved):

windymilla:
> FYI @windymilla in case this is at all useful in GG2 (or gives you ideas for other DP things that you might want API access to in GG2).

tl;dr - nothing in the current phase of development.

Long version:

Thanks - possibly a future enhancement. Sharon mentioned about GG2 being able to take a project ID, and then setting up your project folder, including downloading the text zip file, and the images zip file, and unpacking them. 

If I understand correctly, the user would also have to enter their `API_KEY`  into GG2, so it could use the new API calls.

Another idea, but definitely not for now:

On the project page when the user has checked out a project for PPing, they can 
["Upload a partially post-processed file as a backup"](https://www.pgdp.net/c/tools/upload_text.php?project=projectID63b4e1e68a96d&stage=in_prog_1). It would presumably be possible to do that via a new API call and a menu entry in GG2, something like "Backup PP work to DP server". This would put a backup copy of the PPer's work-in-progress onto the DP server for safe-keeping.

Advantage: User is more likely to do it, and not lose work, if it's a single click, than if they need to Save the file in GG2, go to their File manager and zip the correct files & images dir, then go to their browser and visit the upload page, select the zip file for uploading, etc.

Disadvantage: User is more likely to do it frequently, potentially creating lots of backups of what can be quite large zip files.

cpeel:
> If I understand correctly, the user would also have to enter their `API_KEY` into GG2, so it could use the new API calls.

Correct. If and when we get there we'll need to do some education for users on the importance of keeping their API keys private and storing them in ways that treat them as such too.

> It would presumably be possible to do that via a new API call and a menu entry in GG2, something like "Backup PP work to DP server". This would put a backup copy of the PPer's work-in-progress onto the DP server for safe-keeping.

Yes, we could also expose this capability via API. We'd need to create a new mechanism for uploading binary files via the API which we don't currently have, so this is less turnkey than the API added in the current PR. Totally doable though.

mrducky4:
> > when those were further fetched with the same API key (and `Accept: application/zip`) everything worked as I'd expect.
> 
> Note that the URLs being returned are static content served directly by Apache and should not require any API keys (clearly doesn't hurt to have them in the header, but they are not required).

So it is ok that these zip files are available to anybody, with no authentication?

cpeel:
> > > when those were further fetched with the same API key (and `Accept: application/zip`) everything worked as I'd expect.
> > 
> > 
> > Note that the URLs being returned are static content served directly by Apache and should not require any API keys (clearly doesn't hurt to have them in the header, but they are not required).
> 
> So it is ok that these zip files are available to anybody, with no authentication?

Yes. All of this is data that all proofreaders have access to via the web interface already (Sharon outlines those bits in https://github.com/DistributedProofreaders/dproofreaders/pull/1475#discussion_r2051606399).

tangledhelix:
> > So it is ok that these zip files are available to anybody, with no authentication?
> 
> Yes. All of this is data that all proofreaders have access to via the web interface already (Sharon outlines those bits in [#1475 (comment)](https://github.com/DistributedProofreaders/dproofreaders/pull/1475#discussion_r2051606399)).

Well if we want to split hairs, the scenario she talks about is a user without PP permission, not an anonymous visitor not logged in. I'm not sure how much a difference that should make.

srjfoo:
I've just done some quick tests. An unprivileged user who doesn't have PP access has to be logged in to download CTF or images. A PPer who has the URLs for project files that they have checked out does not have to be logged in.

So, we're assuming ethical behavior. That is, the PPer who retrieves the URLs via the API will not share them with all and sundry. Except for PFs and site admins, PPers can download these files from the site only for projects they have checked out, and an API key will also give them access only to those projects they have checked out.

cpeel:
For clarity, the contents of `/projects` has never required authentication. There is nothing that prevents a random internet person from looking at the code in this repo for how these URLs are formed, accessing the projectIDs from [obscured for Reasons], creating the URL, and downloading the generated PP artifacts. This has been true for almost 25 years.

We could argue if this, and all files in `/projects` including the page images, should require authentication in today's internet but nothing in this PR changes what we have now.

mrducky4:
Should this PR include an update to `api/USERS_GUIDE.md` to add this new API path?

cpeel:
> Should this PR include an update to `api/USERS_GUIDE.md` to add this new API path?

The `USERS_GUIDE.md` includes some _examples_ to help the user out, but the `dp-openapi.yml` is the authoritative source for the API. We don't include all endpoints in the user's guide. If you (or others) feel this one is important to include we can.

cpeel:
> This is my first time delving into this API, so I apologize in advance for asking newbie questions. Now I know the user guide is just examples.

These are good questions - keep asking them!

> Next newbie question: Is it practical and desirable to add a test for this in `ApiTest.php`?

Maybe? Well, yes and lets see how much of our test infra I need to update to support this since it uses the filesystem in ways that prior tests don't. Let me look into that.

PR #1479, opened 2025-05-04
Localization tweaks

srjfoo:
Two corrections for English strings, and adding gettext() markup for another.

These are all tweaks that the translators notified me of. Specifically, I'm curious about the markup for the "Search" button in `review_work`. Does it need the `attr_safe`? I used the "Search" button from the project search form as a model.

Sandbox at [more-translation-strings/](https://www.pgdp.org/~srjfoo/c.branch/more-translation-strings/).


cpeel (approved):
These are all great changes, thanks for the PR!

And yes, we want the `attr_safe()` just like you added it in case the translated string includes a `'` in it.

chrismiceli (approved):

PR #1480, opened 2025-05-05
Handle edgecase of adding files to deleted projects

cpeel:
This addresses an edgecase: If a project has been deleted it doesn't show up in information_schema and thus doesn't have a TABLE_COLLATION. If the user gets to add_files with an old project page it would raise a warning and deprecation message. The page correctly tells the user that they can't load the project but gives an incorrect reason why (the project table isn't UTF-8) rather than the real reason (the project table doesn't exist).

This PR also prevents SAs getting to add_files if the project table doesn't exist.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/minor-db-warning/

### DIFF
--- a/pinc/DPDatabase.inc
+++ b/pinc/DPDatabase.inc
@@ -187,7 +187,7 @@ final class DPDatabase
         );
         $result = self::query($sql);
         $row = $result->fetch_assoc();
-        $table_encoding = $row['TABLE_COLLATION'];
+        $table_encoding = $row['TABLE_COLLATION'] ?? "";
         $result->free();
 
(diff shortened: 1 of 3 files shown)


srjfoo (approved):
If anyone wants to test the "not UTF-8", [projectID459698b8b63ac](https://www.pgdp.org/~cpeel/c.branch/minor-db-warning/project.php?id=projectID459698b8b63ac) has been kept intentionally "not UTF-8".

And oops! we forgot about testing archived and deleted projects in the original PR 😳

chrismiceli (approved):

70ray (approved):

Issue #1255, opened 2024-06-28
Enable a better local dev experience with vargrant and/or docker/rancher


cpeel:
Doing local development on the DProofreaders codebase is hard. The only prescribed method right now is the [development VM](https://www.pgdp.net/wiki/DP_Code_Development_VM) which is big and bulky and requires VirtualBox or VMWorkstation / Fusion (the last two are now free for personal use, though).

It would be easier if we could use vagrant or docker containers (via Docker Desktop (paid) or Rancher Desktop (free)). With the addition of the new "JSON" forum backend type (see https://github.com/DistributedProofreaders/dproofreaders/pull/953) we no longer need phpBB for the vast majority of feature development too. That means that our bear minimum set of requirements are MySQL and PHP because _technically_ PHP has a built-in web browser (`php -S`) so we don't even need Apache. This is used in the smoke tests so it works to some degree.

One of the challenges is going to be bootstrapping data into MySQL. The smoketests use a `test_tables.sql` file which loads some data and a `users.json` for the forums. A starting point might be to flesh those out to look more like the data in the developer VM and use those both for smoke tests and for local dev. Ideally we'd have some way of programmatically creating some of this without a DB dump but that shouldn't block this effort.

This ticket is a bit open-ended but success would be an initial way of doing development locally without needing the VM.

cpeel:
As a datapoint, I would lean towards a container-based approach where a `docker-compose.yml` would bring up a MySQL and PHP/Apache container with local volume mounts pointing to the code and configs and to persist the database. This will still require us figuring out how to install & use composer although there is a docker container for that too.

chrismiceli:
I started a branch here a while ago:
https://github.com/chrismiceli/dproofreaders/tree/docker
with this being the important commit: https://github.com/chrismiceli/dproofreaders/commit/e932b86dde7794fa2c2b0ff8beea9b81063cdad1
but I didn't get the database populated sufficiently.

I found someone else had a docker repo that used curl to initialize the DB you can see here:
https://github.com/fhs/dproofreaders-docker
particularly in this script
https://github.com/fhs/dproofreaders-docker/blob/master/app/bin/create-admin.bash

I think we' could use the unit test script or API to populate the database more reliably.

tangledhelix:
> It would be easier if we could use vagrant or docker containers (via Docker Desktop (paid) or Rancher Desktop (free)).

[...]

Issue #1477, opened 2025-04-19
Create a new development VM (or other acceptable development model)


cpeel:
The latest [DP Development VM](https://www.pgdp.net/wiki/DP_Code_Development_VM) runs Ubuntu 20.04 and PHP 7.4. This will not allow development on code after R202503 because of the new PHP 8.1 requirement. Moreover, due to Apple Silicon, many developers cannot use the x86-based VM anyway.

We need to either create a new VM in a method that we (or developers) can easily spin up x86- and Apple Silicon-based VMs or find other method to enable development outside of the TEST server.

One option is that rather than create a VM, create an [Ansible](https://docs.ansible.com/) playbook that can be used to create the VM. Developers would spin up an Ubuntu-based VM of their choice (VMware, VirtualBox, KVM, etc) and then run the playbook to create the development environment. We would maintain the Ansible playbook and not the whole VM.

Another approach might be to create a Docker- or Vagrant-based development environment (see https://github.com/DistributedProofreaders/dproofreaders/issues/1255).

gbnewby:
The new pglaf.org server has plenty of capacity, if you'd like to set up an Ansible-based VM there. I cannot tell from your question whether finding a suitable host is part of the problem, or if you are just looking for input on the flavour and mechanism of the development environments.

mrducky4:
I don't have good suggestions for how to proceed, but I can give some input about what I'd like to be able to do with a development environment.

I'm familiar with the current VM, and just getting familiar with the TEST server. I use two different development PCs (home and travel), both have Windows 10 Home. I am familiar with Docker, but not Docker Desktop, Ansible, Rancher or Vagrant. When I checked several years ago, I could not run Docker on Windows 10 Home. That might have changed by now.

One thing I'm looking for in a dev environment is the ability to use an IDE (or at least a text editor I like) for editing source code. I do this with the VM by putting the dproofreaders repo in the Windows host filesystem, then configuring the VM to mount it within the VM filesystem. So I can run VSCode on Windows to edit files, and they are the exact same files used in the VM. I haven't figured out yet how to do this with TEST, is it possible?

The next step up from that is the ability to run PHP code in a debugger, to step through code as it is responding to an HTTP request. I haven't spent any time yet to figure how to do this, is it possible?

[...]

PR #1485, opened 2025-05-17
Docker-based local development environment

cpeel:
This is work on https://github.com/DistributedProofreaders/dproofreaders/issues/1477 (and if accepted/merged it will fix https://github.com/DistributedProofreaders/dproofreaders/issues/1255) to create a docker-based local development environment.

I'm very interested in developers kicking the tires on this. This was developed on an Intel macOS using Docker Desktop and I'm particularly interested in users of other platforms trying it out.

I haven't tested it, but I am assuming this will work within a Linux VM as well, although I _suspect_ that we'll need to deal with some docker container permissions stuff which macOS appears to handle transparently for us. Should also work on Windows for those versions of Windows that will allow running containers as I'm told some don't.

The most important part of this PR is: will this work for you? Do you have other ideas / thoughts based on this? Is this structured like you would want to use it?


cpeel:
Notes:
* we'll want to update `SETUP/DEVELOPMENT.md` in this PR
* if https://github.com/DistributedProofreaders/dproofreaders/pull/1483 goes in first we should update this PR to use `install_dependencies.sh`. If this goes in first we should have that one update the code in this one to do the same.

SETUP/devex/README.md
```diff
@@ -0,0 +1,101 @@
...
+These will persist across container reboots.
+
+### Set up
+
+First, install either [Docker Desktop](https://www.docker.com/products/docker-desktop/)
+-- free for personal use -- or [Rancher Desktop](https://rancherdesktop.io/).
+Note that if you use Rancher Desktop the commands are slightly different,
+consult their manual for details.
+
+_Note: All commands should be run relative to this directory._
```
windymilla, line 20: I guessed that "this directory" meant the one containing the README, and everything worked OK. It did cause me some concern in my naivety, that maybe it meant the directory at the top of this tree, i.e. SETUP's parent directory. This was because some later commands, such as `docker compose exec dp-dev SETUP/devex/dpdev-docker/init_dyndir.sh` appeared to be referencing files relative to that parent directory. All went well, but I don't know if the wording "this directory" could be adjusted slightly to ease the tension for novices such as me :)

windymilla (approved):
Wow! From the viewpoint of a complete novice, I can confirm that all works well on Windows 11 Home, version 24H2.

It took about 30 minutes, by which time I had installed Docker Desktop, configured everything using the README, accessed the site locally in my browser, populated the database, registered my username, logged in, proofed a page, made a trival edit to the code, and refreshed my browser to see the change.

Only one thing surprised me (though it probably shouldn't have) - when I clicked on Help (top left of http://localhost:8080/c/) it took me to the main site wiki rather than something local.

Overall, a very smooth experience. Although I'm not actively developing the site code at the moment, this will/would be an excellent environment for me to do so. Thank you.

SETUP/devex/README.md
cpeel: This is excellent feedback and something I pondered too. If we moved the `docker-compose.yml` file to the top-level directory (and updated the relative paths within it) the commands could be run from there instead which might feel more natural anyway.

Thank you very much for testing it out on Windows!

SETUP/devex/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+If you want to interact with the database you can do so through the dp-dev
+container:
+```bash
+docker compose exec dp-dev mysql -hmysql -udp_user -pdp_password dp_db
+```
+
+You can run php-cs-fixer and phpstan directly as well:
+```bash
+docker compose exec dp-dev ./vendor/bin/php-cs-fixer fix
```
chrismiceli, line 107: should we link to the https://github.com/DistributedProofreaders/dproofreaders/blob/master/SETUP/CODE_STYLE.md here for other commands that could be available?

SETUP/devex/dpdev-docker/Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM php:8.1-apache
+
+# Install dependency packages
+RUN apt-get update
+# DP deps: mariadb-client gettext aspell aspell-fr pngcheck
+# Handy dev deps: unzip git
+RUN apt-get install -y mariadb-client gettext aspell aspell-fr pngcheck git unzip
```
chrismiceli, line 7: reading https://docs.docker.com/build/building/best-practices/#run, it seems we should combine the update in this command as well as a few other tweaks:
```
RUN apt-get update &&  apt-get install -y --no-install-recommends \
mariadb-client \
gettext \
aspell \
aspell-fr \
pngcheck \
git \
unzip \
&& rm -rf /var/lib/apt/lists/*
```

SETUP/devex/dpdev-docker/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+# Install PHP extensions
+ADD --chmod=0755 https://github.com/mlocati/docker-php-extension-installer/releases/latest/download/install-php-extensions /usr/local/bin/
+RUN install-php-extensions intl mbstring memcached gettext mysqli xml zip
+
+# And composer
+COPY install_composer.sh /tmp
+RUN bash /tmp/install_composer.sh && mv composer.phar /usr/local/bin/composer && rm /tmp/install_composer.sh
+
+# Install nodejs/npm courtesy of https://deb.nodesource.com/
+RUN curl -fsSL https://deb.nodesource.com/setup_22.x | bash - && apt-get install -y nodejs
```
chrismiceli, line 18: should we use add here? https://docs.docker.com/reference/dockerfile/#add not sure why it would be better but maybe more idiomatic? idk

sitemap.php
```diff
@@ -5,7 +5,7 @@
 // fixed, non-project pages to include in the listing with their
 // change frequency
 $fixed_pages = [
-    "default.php" => "monthly",
+    "index.php" => "monthly",
```
chrismiceli, line 8: Would it be easier to configure the apache to default in the container to default.php to avoid any risk to production?

SETUP/devex/dpdev-docker/Dockerfile
chrismiceli, line 18: if we put this earlier, we could combine the apt-get installs together to avoid layers in the image.

chrismiceli (approved):
This is AMAZING! Verified on windows with Docker deskop + WSL engine. This is incredible. It will definitely make contributions easier and keeping up with distros etc. much easier than the VM.

70ray (commented):
I'm missing something. I installed Docker desktop in Linux KDE neon (based on Ubuntu) then typed "docker compose up -d" but it replied "no configuration file provided: not found". I'm obviously missing some important step.

cpeel:
> I'm missing something. I installed Docker desktop in Linux KDE neon (based on Ubuntu) then typed "docker compose up -d" but it replied "no configuration file provided: not found". I'm obviously missing some important step.

In the code as-is, you must be in the `./SETUP/devex` directory to run the command. Although based on this and Nigel's comment I'm inclined to fix that and move the `docker-compose.yml` file to the base dir but I'd like more feedback before I start changing things.

70ray:
> > I'm missing something. I installed Docker desktop in Linux KDE neon (based on Ubuntu) then typed "docker compose up -d" but it replied "no configuration file provided: not found". I'm obviously missing some important step.
> 
> In the code as-is, you must be in the `./SETUP/devex` directory to run the command. Although based on this and Nigel's comment I'm inclined to fix that and move the `docker-compose.yml` file to the base dir but I'd like more feedback before I start changing things.

So I should have cloned this repository onto my local machine first?

cpeel:
> So I should have cloned this repository onto my local machine first?

Yes, you must be using the branch in the PR. Either fetching my repo into your existing clone and checking the branch out, or cloning my repo and checking the branch out.

70ray:
> Yes, you must be using the branch in the PR. Either fetching my repo into your existing clone and checking the branch out, or cloning my repo and checking the branch out.
I had always previously worked on the test server rather than my local machine so I missed that step.

SETUP/devex/dpdev-docker/Dockerfile
tangledhelix: Clearing the apt cache can be important in small images, however this isn't one. My build clocks in a bit under 1GB, whereas cleaning up the apt-cache would trim off only 19MB. There's no reason to **not** do this, but it isn't a significant difference IMHO.

SETUP/devex/dpdev-docker/Dockerfile
tangledhelix: ADD can copy a file into place, but we'd need to put a RUN after it to run the script (and then to install nodejs). Doing that here is adding another command (but I'm guessing you meant this to be in conjunction with the other suggestion to combine this apt install into the earlier command).

tangledhelix (approved):
I'm not a DP site dev, so I didn't do a whole lot of testing as far as playing with site code; instead I focused on whether the Docker setup seems to work, and adding any remarks about that specifically.

macOS 15.4.1 on Apple Silicon / Docker Desktop: works as-is, no problems.

Ubuntu 24.04 LTS arm64 (running as a VM on Apple Silicon) / Docker 26.1.3: works as-is, no problems, no security changes needed.

Looks great! I left a couple of remarks but they're minor things.

SETUP/devex/dpdev-docker/Dockerfile
tangledhelix: To find out, I looked at the install as-is vs. adding `--no-install-recommends` and the difference seems to be about 24MB.

SETUP/devex/README.md
```diff
@@ -0,0 +1,125 @@
...
+```
+
+On first run it will download some docker images and build the dp-dev image.
+When the containers start it will create an initial (empty) database and
+provide a web interface on port 8080.
+
+Install the PHP & JS packages:
+```bash
+docker compose exec dp-dev composer install
+docker compose exec dp-dev npm install
```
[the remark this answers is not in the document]
cpeel: It isn't quite that simple because we actually want the container to run its existing `ENTRYPOINT` and `CMD` (see [its layer details](https://hub.docker.com/layers/library/php/8.1-apache/images/sha256-3247f78bc84313834a06ff5a0f807fcfa65f5db4a60ce0850cc64750fa6e7a39)) that starts Apache. So to have it do something else I need to create a shell script that does everything we want to do and still do what it's doing already.

Separate from that, I'm a little wary of having starting the docker container have a side effect of doing something that will alter the contents of the user's code checkout.

sitemap.php
cpeel: It's a good question. I think I'd rather finally get the codebase using the standard `index.php` everywhere instead of just most places with these two outliers. We can add redirects in the prod Apache config to redirect to `index.php` if someone actually uses them though to prevent possible issues on prod.

cpeel:
@tangledhelix, as you tested the solution inside a VM, I don't suppose you have a rough-in of the commands you used that we can provide to folks who would prefer a VM solution? I assume it's something like:
1. spin up Ubuntu 24.04 VM
2. install core dev utils bundle (that includes git I think?)
3. [install docker on Ubuntu](https://docs.docker.com/engine/install/ubuntu/)
4. add your user to the `docker` group so they can run the docker commands
5. clone DProofreaders repo
6. follow the provided instructions for docker deployment

?

tangledhelix:
> @tangledhelix, as you tested the solution inside a VM, I don't suppose you have a rough-in of the commands you used that we can provide to folks who would prefer a VM solution? I assume it's something like:
> 
> 1. spin up Ubuntu 24.04 VM
> 2. install core dev utils bundle (that includes git I think?)
> 3. [install docker on Ubuntu](https://docs.docker.com/engine/install/ubuntu/)
> 4. add your user to the `docker` group so they can run the docker commands
> 5. clone DProofreaders repo
> 6. follow the provided instructions for docker deployment

It's mostly the above.

I used an existing VM I use for Guiguts testing, so git and such was already installed. So I was already past steps 1-2.

Your link uses Docker's apt repo, which is probably a better idea, but I just installed from the Ubuntu 24 repo and that also worked.

```
apt-get update && apt-get install -y docker.io docker-compose-v2
```

Of course, an Ubuntu desktop user could use the Docker Desktop stuff as well, but I didn't.

Added to the group: `sudo usermod -a -G docker dan` and had to either log out of the desktop UI or reboot to take effect; I forget which one. One could skip this step by running docker commands with sudo, though.

And then I did steps 5-6.

If you want, after this is merged, I can spin a new VM and PR some changes to the README.

70ray (approved):
It all worked smoothly once I had got the code onto my machine.

SETUP/devex/README.md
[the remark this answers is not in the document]
cpeel: I've tried to thread the needle and added a custom container startup script that does things we always want to do (like initialize the dyn_dirs) and allow users to add a local script that, if it exists, will run to do things like always update the Composer and NPM packages.

This follows my philosophy of not having the container modify any contents in the checkout by default.

SETUP/devex/README.md
cpeel: Good idea. I added the `npm run format` example as well a links to those.

SETUP/devex/dpdev-docker/Dockerfile
cpeel: It's a good suggestion but I'm going to punt it for now. Note also that `install-php-extension` is a whole script that also does a `apt-get update` and then removes the caches (and itself) at the end of the run. So there's a few things going on here that could be optimized.

PRs to further optimize the Dockerfile after this merges in are welcome!

cpeel:
Thanks everyone for the testing and great comments.

I've incorporated some of the review comments and made a few small quality of life updates:
* The `docker-compose.yml` is now at the base of the repo so `docker compose` commands are run from there. This will be more natural since the `WORKDIR` in the container is the base of the repo.
* The dyn_dir creation step has been moved into a container startup script, so one fewer thing to do by hand.
* Bringing the container up will not change any contents of your checkout, by design. Users can, however, create an optional local container startup script that if present will be run when the container starts. This could be useful for having it auto-update your composer / npm dependencies.

This feels like it's at a place to be merged and useful and we can iterate further on it from there.

cpeel:
Instructions added for installing a VM and using the docker solution (thanks @tangledhelix for the instructions).

While testing that I saw that as I feared the `vendor` and `node_modules` are being created by root on Linux whereas Docker Desktop on macOS appears to magically make them owned by the user. I've added a warning about this and instructions for how to fix it.

srjfoo (approved):
Sucess! Notes to self:
* It helps not to just have a local install, but to remember to check out the correct branch
* Yes, Sharon -- you do need to restart the app before anything works

Edit: This install was on the M1 MBP running 15.4.1

PR #1486, opened 2025-05-22
Add gettext support for JS & introduce ES6 modules

cpeel:
This adds "first-class" support for gettext in the javascript code just like we have for PHP. The localization script pulls the strings out of the javascript code and the PHP code and combines them together into a single `.pot` for the translators. When translations are enabled/disabled or updated, a new `translations.js` file is created that contains the translated strings¹ and is made available to the JS code.

The big advantage of this change is that we no longer need to have strings translated in PHP scripts shoved into JS code via JSON and passed around various functions, we can just use them directly. This PR implements this for the Remote File Manager upload strings as a working demonstration.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/js-gettext/

This code is the first instance of using [JS modules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) and that adds a few positives, some changes, and a big complication to how we've used JS in the past.
* Modules let us do some very sensible things like having JS files include the other files they need directly instead of indirectly through the HTML page (see how `scripts/file_resume.js` includes the new `gettext.js` file). This is a huge boon because it is much more intuitive to have JS files manage their own dependencies. I have an in-progress branch that updates the page browser code and its dependencies to a module.
* The use of modules will allow us to develop self-contained "applications", which we have one of already: the page browser. The upcoming new proofreading interface is another example. Both of these "applications" use the API directly and modules can allow for a better local development experience than continually futzing with the HTML to manage transitive dependencies. 
* It slightly increases our minimum supported browser versions (see the `SETUP/INSTALL.md` diff for details).
* We have to tell browsers to treat files that are included in the HTML as modules, thus the new `js_modules` to indicate these files. We also have to tell eslint that they are modules, not scripts, thus the changes in `eslint.config.js` (thanks to @chrismiceli for help with this part!).
* Perhaps most critically, and a downside, because of how ES6 modules are included in other files this breaks how we manage browser caching. Using modules with our current JS caching infra would quickly lead to old JS code being used after new code is deployed because browsers wouldn't realize the code referenced in the modules had changed.

That last one is the biggest challenge as it requires us to use a JS bundler to bundle our JS code as part of a new deployment step we haven't had in the past. I have some working prototypes to show this is possible and will require some iterations to have something ready for review and merging.

To make iterative progress on our JS infra, this PR is targeted to a new temporary branch `js-infra` which will contain the JS code changes until they are at the point where they could be merged to `master` and deployed safely. Having a temp branch lets us have smaller, reviewable PRs. I expect this temp branch will exist for maybe 4-6 weeks.

¹ right now this file is really big because it contains all the strings for the entire site for each language. We can further optimize this by just including the strings used in the JS code. I'm not really worried about it, however, since this file will change very infrequently and we let browsers aggressively cache our JS code so the user impact of the large file is very low.

### DIFF
--- a/locale/translators/index.php
+++ b/locale/translators/index.php
@@ -170,6 +170,8 @@ elseif ($func == "changeenable") {
     $enable = isset($_REQUEST['enable_locale']);
     set_locale_translation_enabled($locale, $enable);
 
+    build_translation_js();
+
     if ($enable) {
         $enable_string = _("Enabled");
     } else {
@@ -485,6 +487,7 @@ function do_upload($locale)
     $po_file = new POFile("$dyn_locales_dir/$locale/LC_MESSAGES/messages.po");
     try {
         $po_file->compile();
+        build_translation_js();
 
         echo "<p>" . _("File successfully uploaded and compiled.") . "</p>";
     } catch (Exception $exception) {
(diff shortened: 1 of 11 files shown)


locale/translators/index.php
```diff
@@ -561,3 +564,41 @@ function validate_locale($locale, $check_dir_exists = true)
...
+        // the strings used in *.js files (unless the files doesn't have any
+        // JS files in which case fall back to loading them all).
+        $pofile = new POFile($po_filename);
+        $pofile->convert_to_json($tempfile);
+        $locale_json[str_replace("_", "-", $locale)] = json_decode(file_get_contents($tempfile));
+        unlink($tempfile);
+    }
+
+    // use "var" not "const" because Safari <12 can't see those within modules
+    // due to some scoping weirdness
```
chrismiceli, line 592: can we just bump safari version? Also curious about the bug, but understand if you don't have a webkit bug number or anything.

scripts/gettext.js
```diff
@@ -0,0 +1,23 @@
...
+            var headers = value[""];
+            delete value[""];
+            translate.setMessages("messages", key, value, headers["plural-forms"]);
+        });
+    } else {
+        console.debug("No translations found.");
+    }
+};
+
+window.addEventListener("DOMContentLoaded", () => {
```
chrismiceli, line 19: do we need to wait this long?

chrismiceli (approved):

locale/translators/index.php
cpeel: The comment is wrong and I need to update it, `const` didn't work until Safari 14, not Safari 12, which is a pretty big jump for this change. I don't have a bug for it -- just tracked it to a Safari behavior change via BrowserStack.

cpeel:
Testing tips:
* Generating a POT should pull in strings from the `scripts/file_resume.js` script (eg "File name extension must be '.zip'") that used to be in `pinc/upload_file.inc`.
* Messages when uploading a file using RFM should still match the user's language if that translation is enabled (French is a good one to try).
* Enabling and disabling a translation should regenerate the `d/locale/translations.js` file.

70ray (approved):

cpeel:
I rebased `js-infra` from `master`, then rebased `js-gettext` off of that and after the CI passes I will merge this into the temp branch and get up step 2.

PR #1491, opened 2025-05-29
Enable explicit download of a pdf file for smooth reading.

70ray:
This addresses task 2109.
Also offer to download all formats only if there is more than one.

### DIFF
--- a/project.php
+++ b/project.php
@@ -1886,6 +1886,7 @@ function echo_smoothreading_options(Project $project): void
 
     // no space after bracket commas
     echo_file_downloads(glob("$smooth_dir/*.{epub,mobi}", GLOB_BRACE), $smooth_url);
+    echo_file_downloads(glob("$smooth_dir/*.pdf"), $smooth_url);
 
     // original uploaded zip file
     $file_base_name = $project->projectid . "_smooth_avail.zip";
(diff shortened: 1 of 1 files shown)


cpeel (approved):

srjfoo:
I threw together a quick sandbox here: [https://www.pgdp.org/~srjfoo/c.branch/sr-pdf/](https://www.pgdp.org/~srjfoo/c.branch/sr-pdf/)

I tested in all of Chrome, Firefox, Opera and Safari. Of the four, only Firefox opened the pdf in the browser; I think that FF does that by default, and leaves the user to save a pdf to the computer if they want to keep it.

However, all three of the other browsers I tried are able to display pdfs directly in the browser. Is it worth adding a link for reading the pdf online, to go along with the ability to read text and HTML online?

cpeel:
> However, all three of the other browsers I tried are able to display pdfs directly in the browser. Is it worth adding a link for reading the pdf online, to go along with the ability to read text and HTML online?

We have very little control what happens to a file (PDF or otherwise) when a user clicks on a link to one. HTML5 includes a new [download attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/a#download) but even that is merely a suggestion based on the docs.

There's also the question of if we should be trying to second-guess how the user expects a certain file type (eg PDF) to work when they click on a link to one.

In short: I'd probably just keep it as-is and let the user's browser do whatever it is going to do.

70ray:
Sorry @srjfoo et al. there was a sandbox but I forgot to put in the link to it.
Sandbox at: https://www.pgdp.org/~rp31/c.branch/srpdf

PR #1492, opened 2025-05-31
Make page browser a JS module

cpeel:
This makes the page browser and its dependencies (including the image controls, splitter, and API function) into ES modules and makes page browser use `gettext()` rather than passing in translated strings in via JSON objects. There should be no user-visible functional changes. This, too, is getting merged into the short-lived `js-infra` branch.

Testing hints:
* Page browser works
* The image pane in the proofreading interfaces continues to work (in and out of WordCheck)
* The image pane and splitters in the PM project word management tools continues to work.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/js-infra-01-page-browser-module


scripts/page_table.js
chrismiceli, line 1: Are these file permission changes?

chrismiceli (approved):

scripts/page_table.js
cpeel: Yeah, this is changing them to not be executable. Not sure why they were to begin with.

tools/project_manager/show_word_context.js
```diff
@@ -27,7 +30,7 @@ window.addEventListener("DOMContentLoaded", function () {
     mainSplit.reLayout();
 
     function setSplitLink() {
-        switchLink.textContent = splitVertical ? proofIntData.strings.layoutHorizontal : proofIntData.strings.layoutVertical;
+        switchLink.textContent = splitVertical ? translate.gettext("Change to vertical layout") : translate.gettext("Change to horizontal layout");
```
srjfoo, line 33: This is backwards -- if the split is vertical, shouldn't the link be "switch to horizontal"?

scripts/page_table.js
70ray: Showing empty file. see https://github.com/orgs/community/discussions/6814

70ray (approved):
This is great.

tools/project_manager/show_word_context.js
cpeel: Good catch! This was a lot of copy/paste and  got that one backwards.

srjfoo (approved):

PR #1493, opened 2025-05-31
Make Format Preview into an ES6 module

cpeel:
This makes Format Preview into a module. As part of this I moved it into `scripts/analyse_format.js` which is where @70ray has it in his `pi_2025` branch so this helps as a bridging step as well. This also moves QUnit as a dev dependency instead of relying on the CDN copy.

Testing notes:
* There should be no functional changes to Format Preview.
* https://www.pgdp.org/~cpeel/c.branch/js-infra-02-js-format-preview-module/SETUP/tests/jsTests/qunit.html should all pass

Sandbox: https://www.pgdp.org/~cpeel/c.branch/js-infra-02-js-format-preview-module/


tools/proofers/previewControl.js
```diff
@@ -299,7 +326,8 @@ window.addEventListener("DOMContentLoaded", () => {
         localStorage.preview_data = JSON.stringify(previewStyles);
     }
 
-    previewControl = {
+    // This is a hack to make previewControl global outside this module
+    window.previewControl = {
```
cpeel, line 330: `previewControl` is effectively an export of this module, but it's used by in-page JS which can't `import` it. Since how we use this module will change with the new proofreading interface this seems good enough for now.

chrismiceli: You could use the dynamic import: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import
This allows importing modules from non module environments. You'd have to make it a promise.

chrismiceli (approved):

70ray (commented):
Looks good.

tools/proofers/previewControl.js
```diff
@@ -312,7 +340,7 @@ window.addEventListener("DOMContentLoaded", () => {
             }
             var msie = document.documentMode;
```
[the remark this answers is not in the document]
cpeel: I thought about that too. Let's do that in a follow-up PR.

tools/proofers/previewControl.js
70ray: It wouldn't help at all, so let's not bother.

70ray (approved):

srjfoo (approved):

PR #1496, opened 2025-06-01
Remove dead code

cpeel:
This removes dead code and outdated comments & docs. This was checked against this repo and `noncvs`. This fixes 1482.


70ray (approved):

SETUP/INSTALL.md
```diff
@@ -319,12 +319,6 @@ hierarchy:
     d/teams
     d/teams/avatar
     d/teams/icon
-    d/pg
```
srjfoo, line 322: Looking on both PROD and TEST, the `d/pg` directory still has up-to-date information in it. On PROD:
```
 6824667 Jun  1 06:03 catalog.txt
12809218 Jun  1 08:45 davids_list.html
12666517 Jun  1 08:45 davids_list_new.txt
```
So, there's something, somewhere, that's still using it.

cpeel: That's from `noncvs`. It is no longer required as part of the main code so it doesn't need to be in `INSTALL.md`.

srjfoo (approved):

PR #1501, opened 2025-06-11
tasks.php: Convert several fixed arrays to const

bpfoley:
This lets PHP guarantee they're never mutated, and means we can remove a lot of global statements in task.php functions because consts are scoped to the current namespace.

(Rename them to be ALL_CAPS as is the convention, and change the names to more useful ones.)

Note that the expressions assigned to a const are restricted to relatively simple things such as literals, arrays and primitive operations on literals. They cannot include more complicated expressions, and cannot be the result of evaluating a function call. Because PHP has to be different to other languages and treats const specially instead of having it mean 'assign once, and don't allow to be modified after that'. Because of this, we manually sort some of the arrays instead of calling `asort`.

Note that we can also use const in other files such as `pinc/bad_bytes.inc` to remove more uses of `global`.

### DIFF
--- a/tasks.php
+++ b/tasks.php
@@ -99,13 +99,13 @@ if ($request_method == 'GET') {
 
 // This section sets up all the "pick from a list" properties of a task.
 
-$tasks_array = [
+const TASK_TYPE = [
     1 => "Bug Report",
     2 => "Feature Request",
     3 => "Support Request",
     4 => "Site Administrator Request",
 ];
-$severity_array = [
+const TASK_SEVERITY = [
     1 => "Catastrophic",
     2 => "Critical",
     3 => "Major",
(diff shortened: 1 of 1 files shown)


cpeel (commented):
> ... and means we can remove a lot of global statements in task.php functions because consts are scoped to the current namespace.

We chatted about this in Slack, but unlike python, in PHP each file isn't its own namespace -- you have to declare a namespace. So using `const`s instead of globals doesn't limit the use to the file they are defined in for our purposes. Still other benefits in many cases though!

chrismiceli (approved):

tasks.php
```diff
@@ -114,49 +114,49 @@
...
-    25 => "My Projects",
-    26 => "Page Details, Diffs",
     27 => "Quizzes",
-    28 => "Rounds",
-    29 => "Release Queues",
     30 => "Teams",
-    31 => "Project Notifications",
-    32 => "Format Preview",
-    99 => "Other",
+    // 32 max
```
srjfoo, line 158: Not sure what the comment is for -- there are actually 33 with `"99 => Other"`

tasks.php
```diff
@@ -166,117 +166,117 @@
...
-    36 => "iPadOS",
-    37 => "Android",
-    99 => "Other",
+    2 => "Windows 95",
+    3 => "Windows 98",
+    6 => "Windows NT",
+    4 => "Windows ME",
+    31 => "Windows Vista",
+    7 => "Windows XP",
+    // max 37
```
srjfoo, line 209: 39 with `"0 => All"` and `"99 => Other"`

tasks.php
```diff
@@ -166,117 +166,117 @@
...
     6 => "Deferred",
+    5 => "Duplicate",
     7 => "Fixed",
     8 => "Implemented",
     9 => "Resolved",
+    1 => "Not a Bug",
+    2 => "Won't Fix",
+    3 => "Won't Implement",
+    4 => "Works for Me",
+    // max 8
```
srjfoo, line 277: There are 9 entries.

srjfoo (commented):
One thing that jumped out at me is that there's still a lot of cruft in the OS and browser arrays. Granted, clearing out the cruft is not the main focus of the PR, but I thought I'd mention it, given that we're actually keeping track of supported browsers now (though browsers aren't the only crufty spots). Is there even a modern enough browser that will run on Macs running Systems 7, 8 or 9?

70ray (approved):

tasks.php
cpeel: This is to help the developer know which is the "highest that's not 99" key. I grok what Brian is doing here.

tasks.php
cpeel: But the highest key value is 37, this makes sense to me.

tasks.php
cpeel, line 277: Highest key value is 9 here
```suggestion
    // max 9
```

cpeel (approved):

tasks.php
```diff
@@ -114,49 +114,49 @@
...
+    16 => "Translation",
     22 => "WordCheck",
-    23 => "Project Search",
-    24 => "Mentoring",
-    25 => "My Projects",
-    26 => "Page Details, Diffs",
     27 => "Quizzes",
-    28 => "Rounds",
-    29 => "Release Queues",
     30 => "Teams",
```
srjfoo, line 157: Nit: Everything else seems to be in alpha order other than Quizzes and Teams -- is that intentional?

srjfoo (approved):
Other than the couple of nits (which are not critical), everything looks good!

tasks.php
cpeel: Fixed this.

cpeel:
I fixed two minor things found by @srjfoo and rebased this to resolve conflicts after the SiteConfig PR went in.

PR #1502, opened 2025-06-19
Bundle ES6 modules with webpack

cpeel:
This is the final PR into `js-infra` and after this merges in the branch is ready for merging back into `master`.

This PR adds the requisite bundling infra to ensure we can get proper cache control for ES6 modules. It uses [webpack](https://webpack.js.org/) to do this. Webpack builds a `manifest.json` file that the PHP code can read. There's a deployment script used for production deployments that converts the `manifest.json` into `manifest.php` for efficiently caching this in the opcache and prevents parsing the JSON file on every page load.

Please see the extensive updates to `SETUP/DEVELOPMENT.md` in this PR for an explanation in how this works and what this means for doing JS development.

I evaluated a few different JS bundlers, including:
* esbuild
* rollup
* parcel

The challenge I had was either the bundler expected a single entrypoint or they didn't provide adequate manifest generation / control for us to integrate it into PHP. 

Sandbox: https://www.pgdp.org/~cpeel/c.branch/js-infra-05-bundling-webpack/

Testing tips: A sanity check that the page browser works as do the WordCheck PM tools that show the page image for context.

### DIFF
--- a/SETUP/DEVELOPMENT.md
+++ b/SETUP/DEVELOPMENT.md
@@ -72,6 +72,80 @@ To install nodejs dependencies run the following in the repo base:
 npm install
 ```
 
+### JS development & deployment
+
+tl;dr: To develop JS: Delete `dist/manifest.php` if it exists. Either delete
+`dist/*` _or_ run `npm run build` whenever you change JS code.
+
+The site uses JS code only in the browser. This code uses ES6 modules which
+work directly in the browser (see [INSTALL.md](INSTALL.md) for the list of
+supported browsers). However, effectively using ES6 modules in production
+requires bundling the code such that all files have content-based hashes in
+the filenames to thwart browser caching when new code is released.
+
+The code uses [webpack](https://webpack.js.org/) to bundle the JS code and
+create a JSON manifest file that can be consumed by the PHP code. The JSON
+manifest file can then optionally be converted into a PHP file which is cached
+in the PHP opcache making it very fast and efficient.
+
+That means there are three possible ways the JS code can be served to a client:
+* as an ES6 modules directly in the code (works for development, does not work
+  for production due to browser caching)
+* as a webpack-built `dist/` code with a JSON manifest (works for development,
+  not ideal for production)
+* as a webpack-built `dist/` code with manifest-as-PHP-file (not ideal for
+  development, best for production)
+
+The code will happily work with any configuration automatically and as a
+developer working with the JS it's important to know the following logic:
+1. If the `dist/manifest.php` file exists it will be used
+2. If not, and the `dist/manifest.json` file exists, it will be used
+3. If neither exist, the ES6 modules will be presented to the browser as-is
+
+For JS developers, you either want to be working with the ES6 module code being
+served directly to the browser (delete `dist/*`) or you want the `dist/*js` files
+being auto-generated upon file change change (still delete `dist/manifest.php`).
+
+webpack can watch the JS files and automatically rebuild `dist/*` with:
+```bash
+npx webpack --watch --progress
+```
+
+To generate the manifest files -- JSON or PHP -- directly, the flow is:
+1. `npm run build` runs webpack and creates `dist/*` contents with JSON manifest
+2. `php ./SETUP/generate_manifest_cache.php` generates a PHP file from the JSON
+   manifest
+
+### Browser caching & ES6 modules
+
+The DP code assumes browsers are going to aggressively cache content like
+JS and images. Indeed, the Apache config for pgdp.net explicitly tells the
+browser to cache images, CSS, JS, and web fonts for a month to reduce the
+amount of content served up to users. To make this work in practice, it falls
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -45,5 +51,12 @@
...
     "jquery": "^3.5.1",
     "resumablejs": "^1.1.0"
-  }
+  },
+  "browserslist": [
+    "chrome 61",
+    "firefox 60",
+    "edge 16",
+    "opera 48",
+    "safari 11"
```
srjfoo, line 60: The browser versions that were documented in #1497:

> * Chrome 64
> * Firefox 78
> * Microsoft Edge 79
> * Opera 51
> * Safari 11.1

I need to do some more testing if we're going with the older versions; I only went back to the newer versions quoted in this comment.

SETUP/DEVELOPMENT.md
```diff
@@ -72,6 +72,50 @@ To install nodejs dependencies run the following in the repo base:
...
+requires bundling the code such that all files have content-based hashes in
+the filenames to thwart browser caching when new code is released.
+
+The code uses [webpack](https://webpack.js.org/) to bundle the JS code and
+create a JSON manifest file that can be consumed by the PHP code. The JSON
+manifest file can then optionally be converted into a PHP file which is cached
+in the PHP opcache making it very fast and efficient.
+
+That means there are three possible ways the JS code can be served to a client:
+* as an ES6 modules directly in the code (works for development, does not work
```
chrismiceli, line 92: Can we elaborate on the caching issue? Modules are cached on name, but can be forced updated with query param maybe?

SETUP/DEVELOPMENT.md
```diff
@@ -72,6 +72,50 @@ To install nodejs dependencies run the following in the repo base:
...
+
+The site uses JS code only in the browser. This code uses ES6 modules which
+work directly in the browser (see [INSTALL.md](INSTALL.md) for the list of
+supported browsers). However, effectively using ES6 modules in production
+requires bundling the code such that all files have content-based hashes in
+the filenames to thwart browser caching when new code is released.
+
+The code uses [webpack](https://webpack.js.org/) to bundle the JS code and
+create a JSON manifest file that can be consumed by the PHP code. The JSON
+manifest file can then optionally be converted into a PHP file which is cached
```
chrismiceli, line 88: to avoid additional scenarios (json vs php), would it make sense to just use .js and cache the file in redis? Probably too much since redis isn't required and invalidating it for updates is probably a pain. Maybe we just don't support the .json flow? Just trying to reduce scenarios to test/support.

chrismiceli (approved):

SETUP/DEVELOPMENT.md
cpeel: They can be force-updated with query param if they're source from the HTML, but not if they're loaded from a module.

So consider the following three files:

```js
// module.js
export var variable = 1;
```

```js
// main.js
import { variable } from `./module.js`;
```

```html
<!-- main.html -->
<script src='main.js?20250607031317' type='module'></script>
```

In `main.html` we can use a query param to force the browser to bypass its cache and load a new `main.js` file. But if we change `module.js` there's no way to force the browser to load a new one except to change the filename, which requires changing it in `main.js`. This is why ES6 modules are not useful natively in the browser in production without something like a bundler that adds a hash to the filename based on its contents.

cpeel: I'll see if I can elaborate more on this "fun" caching edgecase in the document.

SETUP/DEVELOPMENT.md
cpeel: As I went to take the JSON bit out I realized there's a development use-case for it we might want to consider. 

Without the JSON bit a JS developer is going to need to force-reload the page on every code update (or have the browser tools open to always have it always reload content). With the JSON bit, a JS developer can use `npx webpack --watch --progress` to auto-detect changes in the JS code and regenerate the JSON which is automatically picked up on the next PHP page load, so no need to force-refresh anything.

It seems like the latter is a more useful devex that removes the possibility that JS code is updated by the dev but not picked up by the browser. Thoughts?

chrismiceli: I could see that, but my development workflows for pgdp so far have not included the webpack server so I'd probably just continue that way. If you don't feel it is too much to carry all 3 scenarios, no need to remove since it is already present. Just an idea and me trying to understand each one.

cpeel: @70ray - as our most active JS developer, I'm very interested in your thoughts on this PR and the development workflow.

chrismiceli (approved):

SETUP/DEVELOPMENT.md
chrismiceli: Makes a ton of sense now to me, thanks for clarifying those scenarios!

srjfoo (approved):

PR #1504, opened 2025-06-21
Correct hair space Unicode value

### LINKED ISSUE #1503: Incorrect Unicode codepoint in `pinc/unicode.inc`
I believe the value for "hairspace" should be `U+200a` not `U+20a0` in the following line:

https://github.com/DistributedProofreaders/dproofreaders/blob/501a8d329deecde8cc1670e72530b60032f64c9f/pinc/unicode.inc#L464

It's also in [this table in the wiki](https://www.pgdp.net/wiki/Site_conversion_to_Unicode#Unicode_to_ASCII_mapping)

Context:

I asked ChatGPT to convert the table to a section of python code for me to include (out of laziness), and it left a comment
```
("\u20A0", " "),     # hair space (corrected from original, should be U+200A)
```

cpeel:
The correct Unicode value for the hair space is `U+200A` (see https://www.compart.com/en/unicode/U+200A) not `U+20A0` (see https://www.compart.com/en/unicode/U+20A0) which is the Euro-Currency sign ₠.

After this merges I'll update [this wiki page](https://www.pgdp.net/wiki/Site_conversion_to_Unicode#Unicode_to_ASCII_mapping).

Fixes #1503.

Thanks for finding this, @windymilla!

### DIFF
--- a/pinc/unicode.inc
+++ b/pinc/unicode.inc
@@ -461,7 +461,7 @@ function get_utf8_to_ascii_codepoints(): array
             'U+2007', // figure space
             'U+2008', // punctuation space
             'U+2009', // thin space
-            'U+20a0', // hair space
+            'U+200a', // hair space
             'U+202f', // narrow no-break space
             'U+205f', // medium mathematical space
             'U+3000', // ideographic space


windymilla:
> The one character substitution that GG2 will now have that we don't substitute on project load is the `not sign`: `U00ac`.

Another, possibly better, alternative is for me to align GG2 with the DP code. As discussed elsewhere, some math books may contain the `not sign`, so I think it would be better not to remove it in the CP functions in GG2. If a CP notices their OCR has done a lot of hyphen-->not sign misreads, and they know there are no actual not signs in the book, they can easily do a global search/replace operation to remove them.

PR #1520, opened 2025-08-19
Remove further authenticated SR bits

cpeel:
Remove unauthenticated SR information from the project page.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/remove-more-unauth-sr/

Separately, we need to update the DPOD for the SR Access Requirements: https://www.pgdp.net/wiki/DP_Official_Documentation:General/Access_Requirements#Smooth_Reading

### DIFF
--- a/project.php
+++ b/project.php
@@ -94,8 +94,6 @@ if (!User::is_logged_in()) {
     do_project_info_table();
     do_blurb_box($bottom_blurb);
 
-    do_smooth_reading();
-
     echo "<br>\n";
     return;
 } else {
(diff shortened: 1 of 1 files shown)


cpeel:
> It looks like the smooth read books are still available even when not logged in, for example `https://www.pgdp.org/projects/projectID4882418345032/smooth/test-text.txt` for the text version.

That's because everything under `projects/` is static content served directly from Apache and is open-access. It doesn't go through PHP and we have no way to enforce it for authenticated users. The best we can do is remove public access to the URLs.

mrducky4:
> That's because everything under `projects/` is static content served directly from Apache and is open-access. It doesn't go through PHP and we have no way to enforce it for authenticated users. The best we can do is remove public access to the URLs.

Ok, is that the way it is supposed to remain then? Or is there a desire to require authentication/authorization to fetch stuff under `projects/`? I have done authentication for stuff like that before with nginx and a separate process to verify a session token, but I don't know about apache and PHP.

cpeel:
> Ok, is that the way it is supposed to remain then? Or is there a desire to require authentication/authorization to fetch stuff under `projects/`?

Correct, we do not intend to change the behavior of how any of the static content is served because it's not a server resource problem even with the scrapers -- they have to know the URL.

> What is the reason to show a stripped down project page to a client that is not logged in?

Funny you should ask this, because @lhamilton1, @srjfoo, and I were talking about that this morning.

In the past we've said having some form of our project page visible makes us more discoverable if people are looking for specific books -- that is: it's marketing. Not only do we make the project pages public, we actively include them in our sitemap so search engines can easily discover them.

The project pages are also linked to from the `list_etexts` page (eg: https://www.pgdp.net/c/list_etexts.php?x=b&sort=5).

There was no clear outcome this morning from the discussion with the three of us on to keep or change this.

You can see a list of all pages that do not require authentication [in this CI validation script](https://github.com/DistributedProofreaders/dproofreaders/blob/master/SETUP/ci/check_require_login.php#L9).

mrducky4 (approved):

PR #1522, opened 2025-08-24
Fix cursor jumping when doing character replacement

cpeel:
The JS to detect and automatically replace characters that aren't allowed on the site needs to adjust the cursor based on the grapheme length, not the JS internal UTF-16 string length. This solves [this issue reported in the forums](https://www.pgdp.net/phpBB3/viewtopic.php?t=84347).

I'm going to rely on @chrismiceli to tell me if there's a better way to solve this 😁 

Sandbox: https://www.pgdp.org/~cpeel/c.branch/fix-text-processsor-jumping/

### DIFF
--- a/tools/proofers/text_data_processor.js
+++ b/tools/proofers/text_data_processor.js
@@ -1,11 +1,29 @@
 window.addEventListener("DOMContentLoaded", function () {
+    // By specifying the string adjustment value in "adjust" we don't need to
+    // calculate the character (grapheme) length which is challenging until
+    // we can use Intl.Segmenter in very modern browsers.
+    // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter
     const conversionMap = {
-        "‘": "'",
-        "“": '"',
-        "’": "'",
-        "”": '"',
-        "—": "--",
-        "…": "...",
+        "‘": {
+            replace: "'",
+        },
+        "“": {
+            replace: '"',
+        },
+        "’": {
+            replace: "'",
+        },
+        "”": {
+            replace: '"',
+        },
+        "—": {
+            replace: "--",
+            adjust: 1,
+        },
+        "…": {
+            replace: "...",
+            adjust: 2,
+        },
     };
 
     document.getElementById("text_data").addEventListener("input", (event) => {
@@ -13,9 +31,13 @@ window.addEventListener("DOMContentLoaded", function () {
         let selectionEnd = textDataElement.selectionEnd;
         textDataElement.value = [...textDataElement.value]
             .map((character) => {
-                const conversion = conversionMap[character] || character;
-                selectionEnd += conversion.length - 1;
-                return conversion;
+                if (conversionMap[character]) {
+                    const replace = conversionMap[character].replace;
+                    selectionEnd += conversionMap[character].adjust || 0;
+                    return replace;
+                } else {
+                    return character;
+                }
             })
             .join("");
         textDataElement.selectionStart = selectionEnd;


srjfoo (approved):
Works for me.

P2 project [Special Character Test Project with music](https://www.pgdp.org/~cpeel/c.branch/fix-text-processsor-jumping/project.php?id=projectID5deed201ccd06) has 4 astral plane characters at the end of the project custom chars.

chrismiceli:
This looks fine to me, but I don't see any behavior change. Based on the forum, it seems the issue is with the `'𝑓'` character having `.length === 2`. I don't see the replacement char there with an adjust set to 0. Would it make sense to swap `conversion.length - 1;` with `[...conversion].length - 1;`? That should get the most natural 'length' of a string based on graphemes and not utf-16 points. You can see @70ray doing this with the `[...textDataElement.value].map((character)` logic to get individual graphemes.

cpeel:
> This looks fine to me, but I don't see any behavior change. Based on the forum, it seems the issue is with the `'𝑓'` character having `.length === 2`. I don't see the replacement char there with an adjust set to 0.

It's not unique to '𝑓', it's that generally the `.length` is getting UTF-16 character length. If you open a random project on TEST, open a page, and paste the character the cursor jumps around.

> Would it make sense to swap `conversion.length - 1;` with `[...conversion].length - 1;`?

Maybe, what the heck does `[..conversion]` *do*? I don't even know what to look up to understand this JS syntax.

chrismiceli:
> > This looks fine to me, but I don't see any behavior change. Based on the forum, it seems the issue is with the `'𝑓'` character having `.length === 2`. I don't see the replacement char there with an adjust set to 0.
> 
> It's not unique to '𝑓', it's that generally the `.length` is getting UTF-16 character length. If you open a random project on TEST, open a page, and paste the character the cursor jumps around.
> 
> > Would it make sense to swap `conversion.length - 1;` with `[...conversion].length - 1;`?
> 
> Maybe, what the heck does `[..conversion]` _do_? I don't even know what to look up to understand this JS syntax.

:D `[...iterator]` essentially iterates an iterable and stores it in an array. The iterator for a string is unicode aware. So calling `[...string].length` returns a different value than `string.length`. Namely it returns the number of unicode code points handling surrogate pairs. It is mentioned here: https://dmitripavlutin.com/what-every-javascript-developer-should-know-about-unicode/#length-and-surrogate-pairs and is a common strategy for handling these kinds of `.length` issues.

And to my first point, I am not sure what change in this PR fixes the issue with `𝑓`, but I would not be surprised if it just went right over my head.

srjfoo:
> And to my first point, I am not sure what change in this PR fixes the issue with `𝑓`, but I would not be surprised if it just went right over my head.

The test project I mentioned above has some characters in it (at the end of the custom chars) that are from non-BMP characters. If you open a page in the main test sandbox (i.e. not the link above, but the same project in a regular TEST sandbox) and insert one or more of those characters and try deleting a regular character, the cursor should jump by the number non-BMP characters that you inserted. This doesn't happen in Casey's sandbox.

70ray (approved):
Fixes odd behaviour around 𝑓

chrismiceli:
> > This looks fine to me, but I don't see any behavior change. Based on the forum, it seems the issue is with the `'𝑓'` character having `.length === 2`. I don't see the replacement char there with an adjust set to 0. Would it make sense to swap `conversion.length - 1;` with `[...conversion].length - 1;`? That should get the most natural 'length' of a string based on graphemes and not utf-16 points. You can see @70ray doing this with the `[...textDataElement.value].map((character)` logic to get individual graphemes.
> 
> I don't remember writing any of this. git blame thinks @chrismiceli wrote it.

It was me! Look at that.

PR #1523, opened 2025-08-28
F1 daily page limit reduction

srjfoo:
GM request -- to hopefully avoid running out of pages to format in F1.

### DIFF
--- a/pinc/stages.inc
+++ b/pinc/stages.inc
@@ -163,7 +163,7 @@ new Round(
     _("The page-texts have already been proofread, but now need to be formatted with markup which may be specific to the project."),
     'formatting_guidelines.php',
     $pi_tools_for_F,
-    35, // daily_page_limit
+    25, // daily_page_limit
     [],
     [
         0 => _('Formatting Neophyte'),


cpeel (approved):

cpeel:
> Is there any documentation about the `pgdp-production` branch?

None outside of the sawiki. That branch holds code that is specific to pgdp.net and not applicable / useful to other sites. We (Sharon and I) rebase it off of `master` when we do a deploy to TEST and that code gets tagged as `production` when we do a push to PROD.

> Why does this PR include 28 commits? Is this a rebase of all those old commits? Do those old commits need to be checked to make sure they were merged correctly?

@srjfoo needs to rebase this branch on the updated `pgdp-production` since she updated TEST yesterday.

mrducky4:
What's an "sawiki"? I searched DPWiki for the branch name, couldn't find anything.

cpeel:
> What's a "sawiki"? I searched DPWiki for the branch name, couldn't find anything.

The SAwiki is a private wiki for system admins that cover administrative topics.

srjfoo:
> @srjfoo needs to rebase this branch on the updated `pgdp-production` since she updated TEST yesterday.

Oops! I totally overlooked that! After all, when I opened the PR, it was just fine with just the one change. Thanks for the heads-up!

srjfoo:
Also, hot-fixed on TEST and confirmed that the limit works correctly.

mrducky4:
Why is this page limit configuration managed on a branch like this, rather than as a regular configurable item managed through `site_vars.php`?

mrducky4:
Thanks, I think I get it. I don't have a good understanding of what is specific to pgdp.net and what is more general for other users of the code.

srjfoo:
> > Thanks, I think I get it. I don't have a good understanding of what is specific to pgdp.net and what is more general for other users of the code.
> 
> We _try_ (and not always succeed) of keeping pgdp.net-specific stuff either in `pgdp-production` or in the (private) `noncvs` repo. We've been slowly pulling some parts out of the main codebase into the `pgdp-production` branch that is specific to pgdp.net (like the history, branding, etc).

And the `pgdp-production` branch is a definite improvement over the way it used to be done.

Issue #1518, opened 2025-08-12
Remove Greek transliteration tool and quizzes


mrducky4:
This task is to remove a few things related to Greek transliteration.

1. Remove the "Greek Transliterator" link in the proofing interfaces.
2. Remove the pop-up tool itself, `greek2ascii.php`.
3. Remove the `[Greek: ]` button from the proofing interfaces. See also `stages.inc`, `ProofreadingToolbox.inc`.
4. Remove Greek transliteration from `prooffacehelp.php`, the help file for the proofing interfaces.
5. Remove the "Greek Transliteration Quiz and Tutorial" section from the quiz start page `quiz/start.php`.
6. Remove the Greek transliteration quiz tutorial pages, `tut_p_greek_[1-5].php`.
7. Remove the Greek transliteration quiz data files `qd_p_greek_[1-5].inc`, and any other quiz data that is used only for greek transliteration.
8. Remove mention of Greek transliteration from the Fraktur quiz tutorial.
9. Remove the greek quiz pages and transliteration pop-up from `pageload_smoketest.py`.
10. Create an upgrade script to remove greek quiz entries in the `quiz_passes` database table

How to coordinate on wiki documentation to be updated to be consistent with these changes?

cpeel:
> Is there anything about how quiz results per user are stored and retrieved that needs to change?

Quiz results are stored in `quiz_passes`. It would be nice to have an upgrade script that removed the Greek quiz entries from it.

cpeel:
Ideally we would solve this with at least two separate PRs so we can break it up into smaller pieces. Probably the PI work (1-3) and then the quiz work (4-7), and probably in that order?

mrducky4:
Thanks for the suggestions, I updated the issue. I agree two separate PRs for 1-4 and 5-10 would be good.

It wasn't clear to me from slack when you want to merge these changes. What would be a good deadline for submitting PRs, to give time for review before merging?

mrducky4:
When I return from travel, I'd like to do the first set of work (1-4) and submit the first PR the first week of Sept.

cpeel:
Code-wise this is complete with the linked PRs (thanks @mrducky4!). Going to leave this open until we update the wiki documentation, then we can close it.

cpeel:
The wiki updates have been made and the code has been merged and deployed to pgdp.net -- thank you @mrducky4!

PR #1531, opened 2025-09-27
Issue #1518 upgrade script to remove Greek quiz results from the database

mrducky4:
This is the final (that I know of) PR for Issue #1518.

### DIFF
--- a/SETUP/upgrade/24/20250927_delete_greek_quiz.php
+++ b/SETUP/upgrade/24/20250927_delete_greek_quiz.php
@@ -1,0 +1,19 @@
+<?php
+$relPath = '../../../pinc/';
+include_once($relPath.'base.inc');
+
+header('Content-type: text/plain');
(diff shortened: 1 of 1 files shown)


mrducky4:
The automated checks are failing on `sudo apt --fix-broken install aspell aspell-en aspell-fr`, which seems unrelated to changes in this PR. @cpeel , do you have any suggestion?

cpeel:
> > The automated checks are failing on `sudo apt --fix-broken install aspell aspell-en aspell-fr`, which seems unrelated to changes in this PR. @cpeel , do you have any suggestion?
> 
> We might be missing an apt-update in that workflow that something changed upstream and we need. I'll look at it later this weekend. Agreed it's completely unrelated to your PR.

Fixed in https://github.com/DistributedProofreaders/dproofreaders/pull/1532

Issue #1529, opened 2025-09-15
Upgrade phpstan to 2.x


cpeel:
The phpstan lead developer [posted](https://phpc.social/@OndrejMirtes/115209145594560670) that he plans to stop updating phpstan 1.x "in a release or two". We need to upgrade to 2.x.

PR #1535, opened 2025-10-18
Upgrade phpstan to 2.x

cpeel:
Upgrades phpstan to 2.x and fixes https://github.com/DistributedProofreaders/dproofreaders/issues/1529. These are mostly minor code-quality things where we were doing checks that didn't need to be done and the CI passing should be sufficient for testing.

That said, phpstan did catch a bug that has been in the code since at least 2017 when we moved to the mysqli extension. In some of the stats code where we render tables directly from SQL statements in the old dpsql code, integer columns were suppose to be right-aligned but weren't. The fix resulted in the columns being right-aligned and if we decide we don't actually want that behavior I can fix it a different way.

Compare:
* https://www.pgdp.org/c/stats/pm_stats.php
* https://www.pgdp.org/~cpeel/c.branch/upgrade-phpstan/stats/pm_stats.php

Sandbox: https://www.pgdp.org/~cpeel/c.branch/upgrade-phpstan/

### DIFF
--- a/pinc/CharSuites.inc
+++ b/pinc/CharSuites.inc
@@ -8,12 +8,12 @@ class CharSuiteNotEnabledException extends Exception
 class PickerSet
 {
     public string $name;
-    /** @var (?string[])[] */
+    /** @var list<list<?string>> */
     private array $subsets;
     /** @var string[] */
     private array $titles;
 
-    /** @param (?string[])[] $codepoints */
+    /** @param list<list<?string>> $codepoints */
     public function add_subset(string $name, array $codepoints, ?string $title = null)
     {
         $this->subsets[$name] = $codepoints;
@@ -25,13 +25,13 @@ class PickerSet
         }
     }
 
-    /** @return (?string[])[] */
+    /** @return list<list<?string>> */
     public function get_subsets(): array
     {
         return $this->subsets;
     }
 
-    /** @return array{name: string, title: string, rows: (?string[])[][]}[] */
+    /** @return array{name: string, title: string, rows: list<list<list<?string>>>}[] */
     public function get_verbose_subsets(): array
     {
         $verbose_subsets = [];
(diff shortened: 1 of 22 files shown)


srjfoo (approved):

cpeel:
> My main thoughts about the alignment are ...
> 
>     * The Project Created column got right aligned, but
> 
>     * The Rank column did not (visually, it's numeric -- I didn't verify that by inspecting the code, though.

There's a little more code nuance to what is right-aligned and what isn't. What it's doing now is the original intent.

> If one is right aligned, both probably ought to be, but perhaps the table should be forced much narrower and there ought to be appropriately placed padding, at least between the first two columns, so that they don't run into each other.

Based on ^^ I'm more inclined just to rip out the stuff that was broken before -- and thus not change what the code is doing -- rather than fixing it to work as it was intended to. The dpsql stuff is all deprecated and just used in a few places like this.

pinc/CharSuites.inc
```diff
@@ -13,7 +13,7 @@ class PickerSet
     /** @var string[] */
     private array $titles;
 
-    /** @param (?string[])[] $codepoints */
+    /** @param array<int, list<string|null>> $codepoints */
```
jchaffraix: Interesting, the parse for the previous syntax changed. This used to mean `list<list<?string>>` but it is now `list<list<string>|null>`.

We should be careful with using `array<int, ...>` as it could represent either a map or a vector/list. While the difference is minimal, it is important. Thus this suggestion:

```suggestion
    /** @param list<list<?string>> $codepoints */
```

Feel free to use `string | null` if you don't like the shorthand `?string` (though fair warning, we use it across the codebase).

jchaffraix (commented):
This seems pretty safe to merge. Commented on a potential inconsistency and a small type improvement.

pinc/CharSuites.inc
cpeel: Good suggestion - made this and the other set of changes you suggested. Thanks for the review!

jchaffraix (approved):
Thanks for updating PHPStan to v2!

chrismiceli (approved):

PR #1537, opened 2025-10-18
Centralize data retention policy & prune page data

cpeel:
Centralize the values for our data retention policy into one place and add a new cron job to prune old page data per the new policy.

I have manually tested the existing crons on TEST. I have tested a dry-run (everything except the `DELETE`) of the new cron on TEST. After this passes code review -- and before it merges -- I will update TEST to use the modified crons in this branch before merging it in.


cpeel:
New job worked great on TEST.

PR #1538, opened 2025-10-19
Remove right align dpsql

cpeel:
This removes the right-aligned dpsql usage that was broken in 2017 and fixed in https://github.com/DistributedProofreaders/dproofreaders/pull/1535 but turns out we really don't want after all. I checked and noncvs isn't using this option so no changes are needed there.

This also adds typing to dpsql because why not.

Compare:
* https://www.pgdp.org/c/stats/pm_stats.php
* https://www.pgdp.org/~cpeel/c.branch/remove-right-align-dpsql/stats/pm_stats.php

### DIFF
--- a/pinc/dpsql.inc
+++ b/pinc/dpsql.inc
@@ -2,19 +2,19 @@
 define("DPSQL_NO_RANK", false);
 define("DPSQL_SHOW_RANK", true);
 
-function dpsql_dump_query($query, $start_at = 0, $show_rank = DPSQL_NO_RANK)
+function dpsql_dump_query(string $query, int $start_at = 0, bool $show_rank = DPSQL_NO_RANK): void
 {
     $result = DPDatabase::query($query);
     dpsql_dump_query_result($result, $start_at, $show_rank);
 }
 
-function dpsql_dump_themed_query($query, $start_at = 0, $show_rank = DPSQL_NO_RANK)
+function dpsql_dump_themed_query(string $query, int $start_at = 0, bool $show_rank = DPSQL_NO_RANK): void
 {
     $result = DPDatabase::query($query);
     dpsql_dump_themed_query_result($result, $start_at, $show_rank);
 }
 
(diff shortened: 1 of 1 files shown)


pinc/dpsql.inc
```diff
@@ -2,22 +2,20 @@
...
     $result = DPDatabase::query($query);
-    dpsql_dump_themed_query_result($result, $start_at, $show_rank, $ralign_ints);
+    dpsql_dump_themed_query_result($result, $start_at, $show_rank);
 }
 
-function dpsql_dump_query_result($result, $start_at = 0, $show_rank = DPSQL_NO_RANK)
+/** @param bool|mysqli_result $result */
+function dpsql_dump_query_result($result, int $start_at = 0, bool $show_rank = DPSQL_NO_RANK): void
 {
     $n_cols = mysqli_num_fields($result);
```
jchaffraix: While the input type is correct as it's the return value from `DPDatabase::query`, I don't think this line will work with a `bool`.

pinc/dpsql.inc
```diff
@@ -59,7 +57,8 @@ function dpsql_dump_query_result($result, $start_at = 0, $show_rank = DPSQL_NO_R
     echo "</table>\n";
 }
 
-function dpsql_dump_themed_query_result($result, $start_at = 0, $show_rank = DPSQL_NO_RANK, $ralign_ints = DPSQL_RIGHT_ALIGN_INTS)
+/** @param bool|mysqli_result $result */
+function dpsql_dump_themed_query_result($result, int $start_at = 0, bool $show_rank = DPSQL_NO_RANK): void
```
jchaffraix: I think this would work as we're not on PHP 8.1, which supports union types:

```suggestion
function dpsql_dump_themed_query_result(bool|mysqli_result $result, int $start_at = 0, bool $show_rank = DPSQL_NO_RANK): void
```

PR #1543, opened 2025-10-25
Proposed optimization of logo SVG

Rossolson:
The export function from Inkscape left multiple copies of the filter function and overly detailed floating point values. Discussions about the site have noted occasional rendering glitches for the logo. This should mitigate or eliminate any glitches that stem from the file internally.


cpeel:
Also your version appears to have removed the drop shadow, if the GitHub browser rendering is accurate (and I don't have a reason to believe it isn't).

Rossolson:
> @Rossolson - what transformations did you do to the original Inkscape files to improve it? Assuming the new version is better we'll want to do it for all the SVGs used for [DP branding](https://www.pgdp.net/noncvs/dp_branding/).

I did nothing to the look of the file, I simply took a swipe at simplifying it as a way of dealing with the reported rendering glitches. It seemed to have a lot of unnecessary bloat in the XML.

Rossolson:
> Sure, but what did you _do_ to generate the new version? Are you hand-editing the SVG file in a text editor? Are you running it through a tool we can run against all the SVGs directly?

Hand editing. There are SVG optimizers (https://svgomg.net) that can be used instead.

I didn't know about the DP branding page, so I didn't know this file was being used in multiple locations. Using this version may not fulfill all the other needs.

Rossolson:
Any of these values on lines 437 or 449 (and others) can be altered to
taste. Specifically the flood-opacity moving from .1 to .2 (or even the
.28) would make the drop shadow much more prominent. I really don't want to
try and diagnose the exact shade of gray from my machine to look like you
expect on your end. But if you need me to, I can give it a try.

[image: image.png]

PR #1544, opened 2025-10-27
Track Format Preview usage

cpeel:
This implements [Task 2117](https://www.pgdp.net/c/tasks.php?action=show&task_id=2117) which requests that we keep track of when a user runs Format Preview just like we do for when they run WordCheck.

Unlike WordCheck which is all server-side PHP, Format Preview is all JS, so we build on top of Ray's new proofreading API framework to have Format Preview log when a user has _entered_ Format Preview. Also unlike WordCheck, we only log a single DB row for the last time a user has run Format Preview on a page, rather than every usage. Like WordCheck, we keep track of when a user runs Format Preview in any round, even though we only show status for the Formatting rounds (for WordCheck we only show status in Proofreading rounds).

In JS-land we're reusing the existing `imageData` object which is a bit of a kludge but seems reasonable until we get the new PI in place.

Of note, we wire handling of `format_preview_events` into these important places alongside `wordcheck_events`:
* Archiving
* Project deletion
* Project rename
* Page data pruning

This PR (and the sandbox) contains a second commit that will only be checked into `pgdp-production`. This commit will include a timestamp for when the code is deployed to PROD and alert users (and PMs) if the page was saved before Format Preview tracking was enabled, rather than reporting that Format Preview wasn't run on the page which would be inaccurate as we don't know if it was or not.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/format-preview-tracking/

Big thanks to @70ray for laying the groundwork that made the API side of this super easy.

### DIFF
--- a/SETUP/ARCHIVING.md
+++ b/SETUP/ARCHIVING.md
@@ -2,8 +2,8 @@
 
 After projects are completed, they can be archived. This includes moving the
 project files into a separate archive directory, moving the project tables into
-a separate archive database, and moving a project's page_events and
-wordcheck_events into the archive database.
+a separate archive database, and moving a project's `page_events`,
+`wordcheck_events`, and `format_preview_events` into the archive database.
 
 The database instructions in this file resemble those for the main database. See
 [INSTALL.md](INSTALL.md) for more information on how to run database commands.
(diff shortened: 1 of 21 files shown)


cpeel:
The fixup for `copy_pages` needs some work to just copy specific pages. Going to put this into a draft until I have time to fix it all up. Sorry for the noise.

PR #1548, opened 2025-11-12
Wordcheck handles old texts markup

mrducky4:
Reference [Proofing old texts](https://www.pgdp.net/wiki/DP_Official_Documentation:Proofreading/Proofing_old_texts#Nasal_abbreviations).
This adds support for [n] [m] and [que] markup within a word, which allows words such as cou[n]trey to be in the good word list, and not flagged by wordcheck. Similarly, the word pla[n]ting will be flagged as a whole word if it is not in the good word list.

I tested this with the project Legend Land Volume 4 (in P1), to which I added a few words to the GWL.
Sandbox at https://www.pgdp.org/~mrducky/c.branch/wordcheck

### DIFF
--- a/pinc/wordcheck_engine.inc
+++ b/pinc/wordcheck_engine.inc
@@ -9,10 +9,12 @@ use voku\helper\UTF8;
 // Handle dp-specific notation for non-Latin-1 characters
 // See faq/proofreading_guidelines.php#d_chars
 // and tools/proofers/process_diacritcal_markup.js
+// And handle [n] [m] and [que] as parts of a word as described
+// in the Proofing old texts wiki page.
 // $base and $mark are only used to define $bracketed_character_pattern
 $base = "\w{1,2}";    // pattern for: markable base character
 $mark = '[=:.`\\\\\'/v)(~,^\\\\\*]'; // pattern for: diacritical mark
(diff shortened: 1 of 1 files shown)


cpeel (approved):

srjfoo (approved):
Should [this section](https://www.pgdp.net/wiki/DP_Official_Documentation:Proofreading/Proofing_old_texts#-que_symbol) of the Proofing old texts wiki page be edited to remove `[*que]` as valid, or just leave it alone and not worry about it not being recognized by WC?

mrducky4:
> Should [this section](https://www.pgdp.net/wiki/DP_Official_Documentation:Proofreading/Proofing_old_texts#-que_symbol) of the Proofing old texts wiki page be edited to remove `[*que]` as valid, or just leave it alone and not worry about it not being recognized by WC?

I think the decision to remove `[*que]` or not should not be based on how it is treated by WC. If there is a good reason to keep `[*que]` then I think we should improve WC to handle it. I don't know the original reasoning why `[*que]` was included, but it seems now that it is redundant because `[que]` does the same job. If that is true, then simplifying the instructions by removing `[*que]` seems like a good idea to me.

PR #1553, opened 2025-12-19
Delete My Suggestions cache if the user has no suggestions

cpeel:
This PR seeks to solve two problem:
* if a user does not have access to a round and accesses My Suggestions, then gets access to the round and come back, they are served incorrect content (no projects). This is because their original access caches their lack of round access that won't timeout for an hour . This is the cause of [task 2124](https://www.pgdp.net/c/tasks.php?action=show&task_id=2124)
* correctly use the user's interface language as their preferred language if they haven't proofread any projects.

To solve the first one without making multiple functions aware of the cache, move the user suggestions logic into a class that can encapsulate the criteria, the cache, and the suggestions logic. The first commit in this PR does that and adds typing to the functions. The second two commits solve the two bugs mentioned above.

Sandbox at https://www.pgdp.org/~cpeel/c.branch/fix-my-suggestions-cache/

### DIFF
--- a/pinc/UserSuggestions.inc
+++ b/pinc/UserSuggestions.inc
@@ -104,10 +104,12 @@ class UserSuggestions
             }
         }
         // if the user hasn't proofread any projects, lets assume they'll want to
-        // proofread in the language the UI is in and fall back to English
+        // proofread in the language the UI is in and fall back to English.
+        // Project languages are in English, so we need to return the English
+        // name of the language.
         if (!$user_languages) {
             $user_languages = [
-                lang_name(short_lang_code(get_user_language($this->_username))) => 5,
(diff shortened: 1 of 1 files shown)


srjfoo:
> > I registered a new user account to test this. Upon registering, I was given access to the list of projects, and was allowed to proofread a page without having passed the quiz. I don't think that's quite what we want.
> 
> The code to force P1 quiz is in `pgdp-production`, not in `master`, so it isn't in this branch.
> 
> I've created a temp branch of this one and `pgdp-production` at: https://www.pgdp.org/~cpeel/c.branch/fix-my-suggestions-cache-pgdp-prod/

And my apologies for forgetting this when testing!

PR #1555, opened 2025-12-24
Remove unnecessary page_controls.inc usage

cpeel:
While working on cleaning up the new PI branch I discovered that we use `page_controls.inc` here but we don't need to. We only need `get_proofreading_interface_data_js()` if we're using the text pane of the Page Browser, but the context windows only show the image.

Testing will be confirming that the WordCheck PM tools, like showing the context of a proofreader suggestion, all still work and there are no errors on the browser console.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/remove-extraneous-page-controls


70ray (approved):

chrismiceli (approved):

srjfoo (approved):
Ran into some weird stuff in the console in FF, but neither Chrome not Safari showed any problems at all. I've determined to my satisfaction that the FF complaints have nothing to do with the test server or this branch, so, LGTM.

PR #1556, opened 2025-12-24
Replace JS popup help with tooltips

cpeel:
As a stepping stone to get rid of the new window JS code which has long outlived its usefulness, replace the user preferences pop-ups with tooltips. This makes the help more accessible and avoids needing to open a new window which might be blocked by pop-up blockers.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/better-pophelp/


userprefs.php
```diff
@@ -181,7 +184,7 @@ function do_font_sample_update(font_index, size_index, layout) {
 
 output_tab_bar($tabs, $selected_tab, "tab", "origin=" . urlencode($origin));
 
-echo "<p>" . _("Click the ? for help on that specific preference.") . "</p>";
+echo "<p>" . _("Hover over a setting to get more information about it.") . "</p>";
```
chrismiceli: hover may be hard to understand in the context of a touch device. Also, to me at least, the hover behavior was a little obtrusive navigating around my user preferences. Probably wouldn't think twice about it unless observing the behavioral differences though.

chrismiceli (approved):

userprefs.php
srjfoo: I tried it on my (Android) phone. Press-and-hold worked just fine. On my iPads (2017 and 2024), press-and-hold does not work. Turned out that just a quick tap did work for the iPads. I don't have an iPhone to be able to tell what works on them.

cpeel: Yeah, it's not ideal but I couldn't come up with a better way that made them accessible without being obtrusive. One of the reasons I put them over the labels instead of over the fields themselves.

I'm open to other ideas for how to present this info!

cpeel: I tell you what. I'm going to merge this in with the knowledge that we might need to iterate over how to best display this on the page -- but that we'll display it on the page somehow rather than in a new page, so it's a matter of adjusting the JS display code. This unblocks me from putting up a second PR that relies on this code being in.

PR #1558, opened 2025-12-25
Add missing JS files to eslint

cpeel:
Not sure how we missed having all the JS files in eslint, but this sweeps the codebase to add them all.

No sandbox, the CI passing is all we need.


srjfoo (approved):

eslint.config.js
```diff
@@ -33,20 +33,33 @@ module.exports = [
             "scripts/api.js",
             "scripts/character_test.js",
             "scripts/control_bar.js",
+            "scripts/dropdown.js",
```
chrismiceli: I can't remember why we didn't do *.js, because we should be able to, with appropriate excludes if necessary.

scripts/svgGraphs.js
```diff
@@ -1,6 +1,7 @@
 /* global d3 */
 /* exported barLineGraph, stackedAreaGraph, pieGraph */
 
+// eslint-disable-next-line no-unused-vars
 const { barLineGraph, stackedAreaGraph, pieGraph } = (function () {
```
chrismiceli: interesting the exported from above doesn't resolve this.

tools/proofers/toolbox.js
```diff
@@ -1,4 +1,4 @@
-/*global */
+/* eslint-disable no-unused-vars */
```
chrismiceli: hmm, I wonder if more exported are needed? Since we switched to modules though, we should rely on exports or window property assignments to resolve this?

eslint.config.js
cpeel: You didn't, this is me being dumb. We don't need this PR.

If I look closely this is the _modules_ section which should only have the modules in it -- not the scripts which are covered under the wildcard above. And that's why the other changes were needed -- because they are being checked as modules even though they weren't.

Sorry for the kerfuffle and thanks for calling this out Chris.

PR #1561, opened 2026-01-01
French projects bypass the daily page limit

cpeel:
Per GM request, let French projects bypass the daily page limit in any round. These projects will include a message next to the Start Proofreading button to make it clear to the user.

This commit is going into `pgdp-production`.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/pgdp-production-dpl-for-french/

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -33,8 +33,6 @@ parameters:
           path: pinc/DifferenceEngineWrapper.inc
         - message: '#Call to function is_array\(\) with array<string> will always evaluate to true.#'
           path: pinc/3rdparty/mediawiki/WordLevelDiff.php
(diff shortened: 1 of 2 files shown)


srjfoo (approved):

srjfoo:
I've fixed the DPL to 5 pages in P1, P2 and F1 for this sandbox, so that one doesn't have to proof quite so many pages to confirm by testing that the limits are indeed in place.

P1 has several projects with One form or another of French as a primary language, and one with French as a secondary language. Everything tests out fine. That is,
* Only French as a primary language is exempt
* French pages proofed/formatted first do count against the DPL, so if you proof 5 pages (In this sandbox) in French first, you can't go back and proof English pages.

F1 has a couple of projects with French as the language, but none with a secondary language.

PR #1572, opened 2026-04-01
Remove leading/trailing whitespace from credit parts

### LINKED ISSUE #1570: Unwanted space character in Credit line
See the credit line for these projects: https://www.pgdp.net/c/project.php?id=projectID61d685d04ab41 and https://www.pgdp.net/c/project.php?id=projectID61010c7e863e7

After the PPer's name is a space before the comma.

I'm not certain of the cause, but it seems likely the PPer has set their credit name to "Other" on their [preferences page](https://www.pgdp.net/c/userprefs.php) and inadvertently added a space to the end of their name.

Since PPVers copy/paste this credit line into the PG upload form, that could go unnoticed and appear in the final book.

Stripping leading/trailing spaces from "Other" credit names would fix this, if that is the cause.

cpeel:
Ensure that all components used to build the credits field do not have leading or trailing whitespace. This could come from a user's real name field or "other" value, scanner credit, or extra credit.

Note we don't have to do this for usernames as leading and trailing whitespace characters are not valid in usernames.

Fixes https://github.com/DistributedProofreaders/dproofreaders/issues/1570

Sandbox: https://www.pgdp.org/~cpeel/c.branch/trim-credit-components/

### DIFF
--- a/pinc/Project.inc
+++ b/pinc/Project.inc
@@ -1367,8 +1367,8 @@ class Project
             'pp' => $this->postproofer,     // username
             'ip' => $this->image_preparer,  // username
             'tp' => $this->text_preparer,   // username
-            'ec' => $this->extra_credits,   // arbitrary text
-            'cp' => $this->scannercredit,    // username or arbitrary text
+            'ec' => trim($this->extra_credits), // arbitrary text
+            'cp' => trim($this->scannercredit), // username or arbitrary text
         ];
 
         foreach ($creditables as $role => $name) {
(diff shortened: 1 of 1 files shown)


srjfoo (approved):
😁 -- this was challenging to test. What I wound up having to do ultimately was register a new TEST user with spaces trailing the real_name field. Updating your profile, you can't add trailing spaces to the real name, though you can to the credit name.

What turned out to be the challenge was that, as far as I could tell, the credit name field was already being trimmed. And the PPer field was already being trimmed. What wasn't being trimmed was the PM field.

To save others the trouble, compare [the main test server](https://www.pgdp.org/c/project.php?id=projectID57220f53deedb) with [the sandbox](https://www.pgdp.org/~cpeel/c.branch/trim-credit-components/project.php?id=projectID57220f53deedb)

windymilla (approved):
Thanks - that looks great!

chrismiceli (approved):

PR #1573, opened 2026-04-03
Add 'no-dup-check' warning to catalog search

srjfoo:
Some PMs may not realize that the catalog search does not check for duplicates. Add warning, and link to the pglaf in-progress search, and to the DP project search.

Sandbox [here](https://www.pgdp.org/~srjfoo/c.branch/warn-search-for-dups/tools/project_manager/external_catalog_search.php). Should only need a check to verify that the explanatory text is visible (assuming no code changes are needed).

### DIFF
--- a/tools/project_manager/external_catalog_search.php
+++ b/tools/project_manager/external_catalog_search.php
@@ -98,7 +98,6 @@ if ($action == 'show_query_form') {
 
 function show_query_form()
 {
-    global $code_url;
     global $search_params;
     global $PGLAF_inprogress_url;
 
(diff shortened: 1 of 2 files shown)


tools/project_manager/external_catalog_search.php
```diff
@@ -124,6 +126,12 @@ function show_query_form()
 
         echo "<p>";
         echo _("Please put in as much information as possible to search for your project.  The more information the better but if not accurate enough may rule out results.");
+        echo "</p><p>";
+        echo sprintf(
+            _("This catalog search does not check for duplicates. Before creating a project, <strong>please check for duplicates</strong> using <a href='%1\$s'>PG's In Progress search</a> and <a href='%2\$s'>DP's Project Search</a>!"),
```
cpeel: Remind me why we want to point to the DP project search? I know the PG in-progress page does API queries to the DP server to look for in-progress projects, but maybe it doesn't check for all project types?

srjfoo: The PG in-progress page can only search for projects that have a PG clearance, I think? Possibly I'm wrong, though. We discourage creating projects without a clearance, but we currently don't enforce that for the New Project state; it just has to be added before moving a project out of that state.

cpeel: Looks like [the code](https://github.com/gutenbergtools/pglaf-inprogress/blob/main/dpapi.py) is searching for just the title and author regardless of clearance, PG tool should cover all the DP cases too.

I'm fine to leave in the DP search of you want, but the PG search seems to cover that too.

srjfoo: Verified by testing, thanks -- I should have done that earlier.

Second commit added to remove the unnecessary link to the DP project search.

cpeel (approved):

chrismiceli (approved):

PR #1579, opened 2026-05-10
Add api/v1/stats/user/:username/rounds endpoints

bpfoley:
Code borrowed from theme.inc show_tally_specific_stats

---

Test hint edit from @cpeel courtesy of @bpfoley:
```
API_KEY=<api key here>
ROOT=https://www.pgdp.org/~bfoley/c.branch/user-stats-api
curl -X GET "$ROOT/api/v1/stats/user/bfoley/rounds" \
   -H "X-API-KEY: $API_KEY"

curl -X GET "$ROOT/api/v1/stats/user/bfoley/rounds/P1" \
    -H "X-API-KEY: $API_KEY"
```

### DIFF
--- a/api/v1_stats.inc
+++ b/api/v1_stats.inc
@@ -160,3 +160,56 @@ function api_v1_stats_site_round(string $method, array $data, array $query_param
 
     return render_round_stats($round->id);
 }
+
+//---------------------------------------------------------------------------
+// stats/user/:username/rounds
+
+function render_round_user_stats(User $user, Round $round)
+{
+    if (!user_is_a_sitemanager()) {
+        if ($user->username != User::current_username()) {
+            throw new UnauthorizedError("You do not have permission to view stats for another user");
+        }
+    }
+
+    $tallyboard = new TallyBoard($round->id, 'U');
+
+    $current_tally = $tallyboard->get_current_tally($user->u_id);
+    $current_rank = $tallyboard->get_rank($user->u_id);
+
+    $snapshot_info = $tallyboard->get_info_from_latest_snapshot($user->u_id);
+    $yesterday_page_delta = $snapshot_info['tally_delta'];
+    $today_page_delta = $current_tally - $snapshot_info['tally_value'];
+
+    return [
+        "pages_today" => (int)$today_page_delta,
+        "pages_yesterday" => (int)$yesterday_page_delta,
+        "pages_total" => (int)$current_tally,
+        "current_rank" => (int)$current_rank,
+    ];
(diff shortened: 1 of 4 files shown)


api/v1_stats.inc
```diff
@@ -160,3 +160,56 @@ function api_v1_stats_site_round(string $method, array $data, array $query_param
 
     return render_round_stats($round->id);
 }
+
+//---------------------------------------------------------------------------
+// stats/user/:username/rounds
+
+function render_round_user_stats(User $user, Round $round)
+{
+    if (user_is_a_sitemanager()) {
```
cpeel:
```suggestion
    if (! user_is_a_sitemanager()) {
```

cpeel (approved):

srjfoo (approved):
At first I was confused, because ordinary users can see a stats page for all users except those that have specifically set their stats to private. Then I realized that this is not just the plain vanilla stats page that everyone sees, but the table that only squirrels and some PFs see (access-request reviewers, I think?).

cpeel:
> At first I was confused, because ordinary users can see a stats page for all users except those that have specifically set their stats to private. Then I realized that this is not just the plain vanilla stats page that everyone sees, but the table that only squirrels and some PFs see (access-request reviewers, I think?).

That's a good point when thinking about the future of this endpoint. Right now it's more restrictive than it might need to be -- which is fine. If/when in the future we add more user stats and allow users to see other's stats as long as they aren't marked as private, should the ones we're showing here be handled differently?

My gut says no -- that we could just make what's in this PR fall into the same access constraints.

srjfoo:
> > At first I was confused, because ordinary users can see a stats page for all users except those that have specifically set their stats to private. Then I realized that this is not just the plain vanilla stats page that everyone sees, but the table that only squirrels and some PFs see (access-request reviewers, I think?).
> 
> That's a good point when thinking about the future of this endpoint. Right now it's more restrictive than it might need to be -- which is fine. If/when in the future we add more user stats and allow users to see other's stats as long as they aren't marked as private, should the ones we're showing here be handled differently?
> 
> My gut says no -- that we could just make what's in this PR fall into the same access constraints.

I agree. I think the only place that normal users can see actual numbers is in the neighborhood listing. Otherwise, they just see the basic user information (DP Profile, Forums Profile and list of teams, and links to P1-F2 charts. Which you can get numbers from by hovering over the bar, but it's kind of painful.

At this point, I don't think any of the access-request  reviewers is at all interested in pulling an app together using the API 😁

PR #1581, opened 2026-05-13
Fix api/v1/projects?languages=X query which failed with DB error

bpfoley:


bpfoley:
Reproducer to find a project with English as the primary language. Fails without this patch

```
  #!/bin/bash
  set -euxo pipefail

  API_KEY=bfoley
  ROOT=https://www.pgdp.org/~bfoley/c.branch/api-languages-fix

  curl -X GET "$ROOT/api/v1/projects?per_page=1&languages=English+with" \
      -H "X-API-KEY: $API_KEY"
```

api/v1_projects.inc
```diff
@@ -70,8 +70,13 @@ function api_v1_projects(string $method, array $data, array $query_params)
...
         $column_name = $valid_fields[$field];
-
-        if (in_array($field, ["author", "title", "languages", "clearance"])) {
+        // Languages occur as one language or two and are stored as
+        // "Lang1" or "Lang1 with Lang2" in projects.language.
+        // See Project.inc:encode_languages
+        if ($field == "languages") {
+            $likes_str = surround_and_join($values, "language LIKE '%", "%'", ' OR ');
+            $where .= " AND ($likes_str)";
+        } elseif (in_array($field, ["author", "title", "clearance"])) {
```
cpeel: It took me way to long to spot the difference here, this might be more straightforward:
```php
// The API exposes "languages" but the table column is "language" and
// are stored as "Lang1" or "Lang1 with Lang2".
// See Project.inc:encode_languages
if ($field == "languages") {
    $column_name = "language";
}
```

cpeel (changes requested):

api/v1_projects.inc
bpfoley: Yes! Less duplication, and more directly explains the transformation needed.

cpeel (approved):

PR #1586, opened 2026-06-03
page_info_fetch: Fix PHP 8.5 'cannot use bool as array' warns

bpfoley:
`page_info_fetch()` returns a list of 3 items, or false depending on if there's more information. false cannot be used in a destructuring assignment, and PHP 8.5 warns about this.

(Spotted while running smoke tests in an Ubuntu 26.04/PHP 8.5 VM)

### DIFF
--- a/pinc/post_files.inc
+++ b/pinc/post_files.inc
@@ -142,7 +142,8 @@ function write_project_comments($project, $fp)
 function get_page_texts($pages_res)
 {
     $page_texts = [];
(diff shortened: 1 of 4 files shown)


cpeel:
Oh, thanks! If you wanted to add PHP 8.5 to our CI test matrix (in another PR) I wouldn't be sad about that -- although I don't know what else lurks in there so it might be a can of worms.

cpeel (approved):

PR #1589, opened 2026-06-11
Add smoketests using JSON HTTP body in PUT

bpfoley:


SETUP/tests/smoketests/smoketest_setup.sh
```diff
@@ -3,9 +3,15 @@
 # Populate some DB tables used by the smoketests
 mysql -uroot -proot < SETUP/tests/smoketests/test_tables.sql
 
+# Create wordcheck files
+mkdir words
```
cpeel: Can we qualify where this directory should be rather than in the CWD of the script? I don't remember where this is run and am wary of it creating this dir (and the files) in unexpected places on a developer's system who is using it.

bpfoley: Good idea. The script relied on being run directly from the checkout dir (which `SETUP/tests/ci_configuration.sh` sets `_CODE_DIR` to), and it assumed `_PROJECTS_DIR` was `$HOME`. I've added an explicit check and made the script actually use `ci_configuration.sh` to make this more explicit.

cpeel: This is much clearer, thanks!

cpeel (approved):

PR #1591, opened 2026-06-30
Fix #1576 to allow clearing of pages with [none] user in the last round

### LINKED ISSUE #1576: Trying to clear effects of a round where username does not exist throws an exception.
Specific example:

When clearing pages in a project, I accidentally selected a page that was a blank inserted while the project had pages in F1.page_avail, so the proofer name inserted was `[none]`. Since the username didn't exist, it couldn't have any data associated with it cleared, so the following exception was thrown:

```
No user found with username = [none]

Fatal error:  Uncaught NonexistentUserException: No user found with username = [none] in /data/htdocs/c/pinc/User.inc:259
Stack trace:
#0 /data/htdocs/c/pinc/User.inc(141): User->load()
#1 /data/htdocs/c/pinc/page_tally.inc(58): User->__construct()
#2 /data/htdocs/c/tools/project_manager/page_operations.inc(73): page_tallies_add()
#3 /data/htdocs/c/tools/project_manager/edit_pages.php(62): page_clear()
#4 {main}
  thrown in /data/htdocs/c/pinc/User.inc on line 259
```

bpfoley:
While here, do some assorted code cleanup in `page_tally.inc`. This is old code and we can take advantage of some new PHP features now.

This can be checked on test on <https://www.pgdp.org/~bfoley/c.branch/clear-round-nouser/tools/project_manager/page_detail.php?project=projectID5e2676b51011e>. This project contains a lot of pages with `[none]` as the proofreader.

### DIFF
--- a/pinc/page_tally.inc
+++ b/pinc/page_tally.inc
@@ -214,7 +214,7 @@ class PageTally_Neighbor
 /**
  * Should we anonymize information about the given user?
  */
-function should_anonymize($username, $user_privacy_setting)
+function should_anonymize(?string $username, int $user_privacy_setting)
 {
     return !can_reveal_details_about($username, $user_privacy_setting);
 }
@@ -261,7 +261,7 @@ $SECONDS_TO_YESTERDAY = 1 + 3 * 60 * 60;
  * - prev_day_{goal,actual}
  * - curr_month_{goal,actual}
  */
-function get_site_page_tally_summary($tally_name)
+function get_site_page_tally_summary(string $tally_name)
 {
     $site_stats = new StdClass();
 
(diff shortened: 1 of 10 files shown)


pinc/page_tally.inc
```diff
@@ -603,7 +566,7 @@ function calculate_tally_sma(array $array, int $window): array
  *
  * In the returned array, each key is a date-string of the form "YYYY-MM-DD".
  */
-function get_pages_per_day_for_past_n_days($tally_name, $holder_type, $holder_id, $days_back)
+function get_pages_per_day_for_past_n_days(string $tally_name, string $holder_type, int $holder_id, $days_back)
```
cpeel: Can't we use union types for `$days_back`?
```suggestion
function get_pages_per_day_for_past_n_days(string $tally_name, string $holder_type, int $holder_id, string|int $days_back)
```

cpeel (commented):
Nice use of the new constructor ability and thanks for the added typing!

cpeel (approved):

srjfoo (approved):

PR #1594, opened 2026-07-07
Allow projects to be "parked"

cpeel:
A new "parked" state allows projects to be moved outside the usual project workflow to get fixed or worked on.

Projects can be parked from the project page by a site admin. To move projects out of a parked state, the project jump page needs to be used.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/proj_parked/

This PR will need to be rebased after https://github.com/DistributedProofreaders/dproofreaders/pull/1592 goes in.

### DIFF
--- a/pinc/ProjectState.inc
+++ b/pinc/ProjectState.inc
@@ -6,7 +6,7 @@ class ProjectState
         public readonly string $name,
         public readonly string $medium_label,
         public readonly string $label,
-        public readonly int $forum,
+        public readonly ?int $forum,
         public readonly string $phase,
         public readonly string $star_metal
     ) {
@@ -228,6 +228,18 @@ function declare_project_states(array $rounds, array $pools): void
         _("Project Complete"),
         SiteConfig::get()->completed_projects_forum_idx,
         'COMPLETE',
+        ''
+    ));
+
+    // for "parking" a project
+
+    define("PROJ_PARKED", "project_parked");
+    ProjectStates::add_state(new ProjectState(
+        PROJ_PARKED,
+        _("Project Parked"),
+        _("Project Parked"),
+        SiteConfig::get()->parked_projects_forum_idx,
+        'PARKED',
         ''
     ));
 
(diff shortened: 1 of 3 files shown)


pinc/ProjectTransition.inc
```diff
@@ -1166,6 +1166,29 @@ function create_project_transitions(): array
         ]
     );
 
+    // -----------------------------------------------------------------------------
+    // Transitions to PROJ_PARKED.
+
+    foreach (ProjectStates::get_states() as $from_state) {
+        if (in_array($from_state, [PROJ_NEW, PROJ_DELETE, PROJ_PARKED, PROJ_COMPLETE, PROJ_SUBMIT_PG_POSTED])) {
```
cpeel, line 1165: This line dictates the project states that cannot be parked.

windymilla (approved):
It's a sign of the good state of the code that implementing this didn't involve many edits, and they are easily understood by someone who doesn't know the code that well. Thanks!

chrismiceli (approved):

srjfoo (approved):

pinc/ProjectState.inc
```diff
@@ -5,15 +5,15 @@ class ProjectState
...
+    public ?int $forum;
     public string $phase;
     public string $star_metal;
 
     public function __construct(
         string $name,
         string $medium_label,
         string $label,
-        int $forum,
+        ?int $forum,
```
bpfoley: Hmm. It's mildly disturbing that PHPStan didn't catch this mismatch in all the callers. I'm guessing it couldn't properly determine that the type of `SiteConfig::get()->posted_projects_forum_idx` (etc) is `public ?int`

bpfoley:
Looks great to me.
I see you can search for parked projects (and Project Parked state is right before Project Deleted in the sort order and in the search state list).
And moving a project from F1 Available to Parked from the project page Just Works, as does moving from parked to F1 Unavailable using the jump page.
And the transitions listed in the project's Project History table make perfect sense.
Looks like all the ProjectState and ProjectTransition machinery works just as intended!

PR #1596, opened 2026-07-11
ProjectTransition: Use properties in constructor, not array.

bpfoley:
This way is a bit denser, but more importantly, allows phpstan to statically check the types of all the arguments.

While here, tighten up the types:
* disabled_during_SR = null doesn't diff from false.
* Make all the properties readonly.

### DIFF
--- a/pinc/ProjectTransition.inc
+++ b/pinc/ProjectTransition.inc
@@ -11,14 +11,8 @@ define('PT_AUTO', '[AUTO]');
 
 class ProjectTransition
 {
-    // Populated from $options
-    private ?string $project_restriction; // callable
-    public ?string $action_name;
-    public ?string $confirmation_question;
-    public ?string $detour;
-    public ?string $settings_template;
-    private ?string $collateral_actions; // callable
-    private ?bool $disabled_during_SR;
+    public readonly string $action_name;
+    public readonly ?string $confirmation_question;
 
     /**
      * Project Transition constructor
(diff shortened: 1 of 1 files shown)


bpfoley:
Test environment at https://www.pgdp.org/~bfoley/c.branch/pt-default

pinc/ProjectTransition.inc
```diff
@@ -474,14 +463,12 @@ function create_project_transitions(): array
...
         'unrestricted_manager',
-        [
-            'project_restriction' => null,
-            'action_name' => '[default]',
-            'confirmation_question' => null,
-            'detour' => null,
-            'settings_template' => '',
-            'collateral_actions' => null,
-        ]
+        project_restriction: null,
```
cpeel, line 468: Thoughts on not including the `null` ones in the definitions since that's the default value? It would make it more denser and perhaps make it easier to see what's unique about each of the definitions.

I'm on the fence about the idea but thought I'd toss it out there.

pinc/ProjectTransition.inc
```diff
@@ -11,54 +11,44 @@ define('PT_AUTO', '[AUTO]');
...
     public function __construct(
         public readonly string $from_state,
         public readonly string $to_state,
         public readonly string $who_restriction,
-        array $options
+        private readonly ?string $project_restriction = null, // callable
+        ?string $action_name = null,
+        ?string $confirmation_question = null,
+        public readonly ?string $detour = null,
+        public readonly ?string $settings_template = null,
```
cpeel: Feels like this should default to `''` and not take a null value given how it's used (realizing that you're just copying what was here before).

pinc/ProjectTransition.inc
bpfoley: ACK. I thought about removing all the defaults, but I figured you had them in there for a reason, and they might be explicitly there to make it easier for the skim-reader. I'm happy to go either way.

pinc/ProjectTransition.inc
bpfoley: ACK. It make sense to do that with at least `$action_name` and `$settings_template`. You could make the empty string have a special meaning for `$detour` as well, but it's slightly unnatural. And `$project_restriction` and `$collateral_actions` really should be `?Closure` (you can't use callables for properties).

pinc/ProjectTransition.inc
cpeel: This code predates my time at DP so it's been there for a while. Let's leave it like you have it.

cpeel (approved):

srjfoo (approved):

PR #1597, opened 2026-07-11
SiteConfig: Make get()'s return type the singleton.

bpfoley:
So that phpstan can statically check SiteConfig::get()->foo, and we get a more comprehensible runtime error if SiteConfig somehow isn't loaded.


bpfoley:
Test install at https://www.pgdp.org/~bfoley/c.branch/siteconfig-tycheck

cpeel (approved):

PR #1527, opened 2025-09-02
Configure MathJax to support unicode markup.

mrducky4:
For format preview to support the same unicode markup that was recently added to the m2svg utility.

Sandbox: [https://www.pgdp.org/~mrducky/c.branch/mathjax](https://www.pgdp.org/~mrducky/c.branch/mathjax)

I tested using format preview, with the following text taken from [Principia mathematica (Vol. 1 of 3) {LaTeX}](https://www.pgdp.net/c/project.php?id=projectID5fb6b1641fc59). I had to replace a few special characters because I haven't learned yet how to add characters to a project.

```
From 328.png
Thus we are led to three general definitions for descriptive double
functions, namely (if \(x\unicode{x2640} y\) be any such function)
\[
\begin{aligned}
x\unicode{x2640} \text{ is the relation of }x\unicode{x2640} y\text{ to }&y\text{ for any }y,\\
\unicode{x2640} y\quad "\quad "\quad "\quad "\quad " &x\quad "\quad x,\\
\alpha \unicode{x2640}_{,,} y\text{ is the class of values of }&x\unicode{x2640} y\text{ when }x\text{ is an }\alpha .
\end{aligned}
\]

From 579.png
\(\ast \)<b>88\(\cdot \)14</b>\(\ldotp ~~\vdash\colon\ldotp \kappa \subset A'P\ldotp \supset :P\upharpoonright \kappa \epsilon \text{Rel Mult}\ldotp \equiv \ldotp \exists !P_{\Delta }'\kappa \)

<i>Dem</i>.
\[
\begin{aligned}
&\vdash\ldotp \ast 80\cdot 23\ldotp \supset \vdash:\exists !P_{\Delta }'\kappa \ldotp \equiv \ldotp \exists !(P\upharpoonright \kappa )_{\Delta }'\kappa \quad (1)\\
&\vdash\ldotp \ast 35\cdot 65\ldotp \supset \vdash:\kappa \subset A'P\ldotp \supset \ldotp A'(P\upharpoonright \kappa )=\kappa \quad (2)\\
&\vdash\ldotp (1)\ldotp (2)\ldotp \supset \vdash\colon\ldotp \text{Hp}\ldotp \supset :\exists !P_{\Delta }'\kappa \ldotp \equiv \ldotp \exists !(P\upharpoonright \kappa )_{\Delta }'A'(P\upharpoonright \kappa )\ldotp\\
[...]

PR #1599, opened 2026-07-26
Update MathJax to version 4 for format preview.

mrducky4:
Use npm to install mathjax locally, and serve it locally, no longer using the jsdelivr CDN.
Adjust async behavior to await for MathJax.
Update Content-Security-Policy to allow MathJax to dynamically load a worker and fonts.
Fix old bug that showed a mathjax error (this exists in production).

The motivation for this is to use the same version of mathjax in both Format Preview and also the separate m2svg tool. This helps to avoid inconsistencies between the two tools, which are often used together for the same project.

To reproduce the old bug with the mathjax error:
1. Initial condition, Preview Math is disabled in format preview configuration.
2. Start proofing a page.
3. Click Preview to enter the format preview view.
4. Click Configure, then enable Preview Math, then click OK.
5. Expected: you see the preview view, with no error messages.
6. Actual: you see a pop-up error about MathJax.typeset is not a function.

I tested with math markup from PR #1527, which also has the expected output in preview mode.

To verify new consistency with m2svg 3.4.0, use this markup:
`Math with text inside: \(a + b = c + d + \text{\&c.}\)`
Expected output is `a + b = c + d + &c.` where a b c d are italic, &c is not italic. (The production version, with mathjax 3, shows it as `a + b = c + d + \&c.`)

I also tried various sequences of enabling/disabling Preview Math in the format preview configuration.

Sandbox at: https://www.pgdp.org/~mrducky/c.branch/mathjax4

### DIFF
--- a/tools/proofers/previewControl.js
+++ b/tools/proofers/previewControl.js
@@ -15,7 +15,7 @@ import { ajax } from "../../scripts/api.js";
 import { makePreview, defaultStyles } from "../../scripts/analyse_format.js";
 import { validateText } from "../../scripts/text_validator.js";
 
-window.addEventListener("DOMContentLoaded", () => {
+window.addEventListener("DOMContentLoaded", async () => {
     "use strict";
     var supp_set = ["charBeforeStart", "sideNoteBlank"];
     // this is a wrapper round text_preview which enables the padding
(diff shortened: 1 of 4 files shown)


cpeel:
Can we simplify the CSP by bundling mathjax with the code rather than pulling from the CDN? See commit b0033804af99d552fe8df9b60a84b71c33ccfcac in branch https://github.com/cpeel/dproofreaders/commits/mathjax4/ for how we could do that pretty easily. We do this for other JS code like d3.

Then the file is in `node_modules/mathjax/tex-svg.js`.

mrducky4:
I thought that might be an option, I'll take a closer look.

mrducky4:
Just checking - I think I should commit the corresponding update to `package-lock.json` too, right?

cpeel (approved):
Code looks good to me but I would appreciate @chrismiceli's eyes on it as he's our resident JS guru.

srjfoo (approved):

tools/proofers/previewControl.js
```diff
@@ -404,8 +405,9 @@ window.addEventListener("DOMContentLoaded", () => {
             saveStyle();
             // if loading MathJax, wait for it to finish
             initView().then(function () {
-                writePreviewText();
-                hideConfig();
+                writePreviewText().then(function () {
```
chrismiceli, line 408: Very minor, but if we are using async await above, we could here too I imagine.

chrismiceli (approved):

tools/proofers/previewControl.js
mrducky4: Yes, I think so, but I didn't change it since it was already using `.then()`. It should be equivalent either way.

PR #1604, opened 2026-08-25
Update to PHPStan 2.2.9

bpfoley:
A lot of our legacy code leans heavily on associative arrays to pass information between different modules instead of classes with properties with well defined types.

PHPStan 2.2 has a feature called 'sealed array shapes' which allows us to define the only keys allowed in an array (and the types of their values), which give an effect which looks much like classes with properties, but without all the code churn required to implement it.

This is the first step in the process of updating PHPStan's warning level to 6 which requires us to add type hints to all function arguments, function return values, array keys/values, and iterator values.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1456,11 +1456,11 @@
         },
         {
             "name": "phpstan/phpstan",
-            "version": "2.1.42",
+            "version": "2.2.9",
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/phpstan/phpstan/zipball/1279e1ce86ba768f0780c9d889852b4e02ff40d0",
-                "reference": "1279e1ce86ba768f0780c9d889852b4e02ff40d0",
+                "url": "https://api.github.com/repos/phpstan/phpstan/zipball/13d6b4f347bad222da436580c8304fa6f83e6bd0",
+                "reference": "13d6b4f347bad222da436580c8304fa6f83e6bd0",
                 "shasum": ""
             },
             "require": {
@@ -1482,6 +1482,17 @@
             "notification-url": "https://packagist.org/downloads/",
             "license": [
                 "MIT"
+            ],
+            "authors": [
+                {
+                    "name": "Ondřej Mirtes"
+                },
+                {
+                    "name": "Markus Staab"
+                },
+                {
+                    "name": "Vincent Langlet"
+                }
             ],
             "description": "PHPStan - PHP Static Analysis Tool",
             "keywords": [
@@ -1505,7 +1516,7 @@
                     "type": "github"
                 }
             ],
-            "time": "2026-03-17T14:58:32+00:00"
+            "time": "2026-08-22T07:38:16+00:00"
         },
         {
             "name": "phpunit/php-code-coverage",
--- a/pinc/bootstrap.inc
+++ b/pinc/bootstrap.inc
@@ -31,7 +31,7 @@ spl_autoload_register('dp_class_autoloader');
 require_once($relPath.'../vendor/autoload.php');
 
 // Exception handlers are defined differently for HTML and the API
-set_exception_handler(SiteConfig::get()->testing ? 'test_exception_handler' : 'production_exception_handler');
+set_exception_handler(SiteConfig::get()->testing ? test_exception_handler(...) : production_exception_handler(...));
 
 if (!headers_sent()) {
     // Tell proxies to vary the caching based on the Accept-Language header
@@ -66,7 +66,7 @@ include_once($relPath.'misc.inc');
 //----------------------------------------------------------------------------
 
 // Autoloading functions for DP classes in pinc/
-function dp_class_autoloader($class)
+function dp_class_autoloader(string $class): void
 {
     global $relPath;
     if (is_file($relPath.$class.".inc")) {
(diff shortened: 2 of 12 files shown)


pinc/bootstrap.inc
```diff
@@ -66,14 +66,27 @@ include_once($relPath.'misc.inc');
...
+ * @param array{
+ *            expires?: int,
+ *            path?: string,
+ *            secure?: bool,
+ *            httponly?: bool,
+ *            samesite?: 'Lax'|'None'|'Strict'
+ *        } $options
+ * @return bool
+ */
 function dp_setcookie($name, $value = "", $expires = 0, $options = [])
```
cpeel: Are we unable to update the function signature?

```suggestion
function dp_setcookie(string $name,string $value = "", int $expires = 0, array $options = []): bool
```

pinc/forum_interface_phpbb3.inc
```diff
@@ -46,6 +46,9 @@ function phpbb_table($table)
  * installed in the forum interface.
  *
  * Fall back to English if $langcode is not installed on the system.
+ *
+ * @param string|false $locale
+ * @return string
  */
 function phpbb_lang($locale = false)
```
cpeel:
```suggestion
function phpbb_lang(string|bool $locale = false): string
```

pinc/forum_interface_phpbb3.inc
```diff
@@ -107,6 +110,11 @@ function forum_password_hash($password)
  *
  * Note: Use a strict comparison operator (=== or !==) against TRUE
  * to evaluate the pass/fail status of this function.
+ * @param string $username
+ * @param string $password
+ * @param string $email
+ * @param bool $password_is_digested
+ * @return string|true
  */
 function create_forum_user($username, $password, $email, $password_is_digested = false)
```
cpeel:
```suggestion
function create_forum_user(string $username, string $password, string $email, bool $password_is_digested = false): string|bool
```

pinc/forum_interface_phpbb3.inc
```diff
@@ -155,21 +163,23 @@ function phpbb_user_error_handler($errno, $errstr, $errfile, $errline)
...
+ * @return list{bool, string}
+ *     Returns:
+ *     - [true, '']  - user successfully logged in
+ *     - [false, $reason] - login failed and the reason why, valid $reasons:
+ *       - incorrect_username
+ *       - incorrect_password
+ *       - too_many_attempts
+ *       - unknown
  */
 function login_forum_user($username, $password)
```
cpeel:
```suggestion
function login_forum_user(string $username, string $password): array
```

pinc/forum_interface_phpbb3.inc
```diff
@@ -704,8 +710,9 @@ function get_last_post_time_in_topic($topic_id)
  * - forum_name       - name of the forum the topic is in
  * - forum_id         - the id of the forum the topic is in
  * - creator_username - the username of the topic creator
+ * @return array<string, mixed>
  */
-function get_topic_details($topic_id)
+function get_topic_details(int $topic_id)
```
cpeel:
```suggestion
function get_topic_details(int $topic_id): array
```

pinc/languages.inc
```diff
@@ -9,6 +9,10 @@ include_once($relPath.'iso_lang_list.inc');
 
 use Symfony\Component\Process\Process;
 
+/**
+ * @param string $language_code
+ * @return bool|string
+ */
 function lang_code($language_code)
```
cpeel:
```suggestion
function lang_code(string $language_code): bool|string
```

cpeel (commented):
I didn't do all of these, but seems like we can encode some of these PHPStan param definitions directly into the function declaration -- yes?.

Obviously we need the PHPStan details on the array types.

bpfoley:
> I didn't do all of these, but seems like we can encode some of these PHPStan param definitions directly into the function declaration -- yes?.
> 
> Obviously we need the PHPStan details on the array types.

Something I didn't mention in the commit message, but probably should have, is that in cases where PHP can't natively express the types, I'd prefer to have all of types for a function either in the PHPDocs or in the function signature, but not scattered between both. The rationale for this is that it's a potential source of confusion for the human reader, and it's also a source of error if the types in one of the locations is out of date. So the way I decided to do this is where PHP can name all the argument types and the return type, I keep it in the code; otherwise I move them all to PHPDocs.

What do you think?

cpeel:
I'd rather have the types in the PHP functions for variables that PHP directly supports and not have them in the PHPDocs. Because PHP will enforce those at runtime.

pinc/forum_interface_phpbb3.inc
bpfoley: So... teeeechnically `string|true` (which PHP 8.0 can't natively express) is different to `string|bool`. Support for `string|true` was added in PHP 8.2

cpeel: That's ok, 8.3 is our minimum supported version: https://github.com/DistributedProofreaders/dproofreaders/blob/master/SETUP/INSTALL.md#php

pinc/forum_interface_phpbb3.inc
```diff
@@ -46,8 +46,10 @@ function phpbb_table($table)
  * installed in the forum interface.
  *
  * Fall back to English if $langcode is not installed on the system.
+ *
+ * @param string|false $locale
  */
-function phpbb_lang($locale = false)
+function phpbb_lang($locale = false): string
```
cpeel: Yes?
```suggestion
function phpbb_lang(string|false $locale = false): string
```

composer.lock
```diff
@@ -1456,11 +1456,11 @@
         },
         {
             "name": "phpstan/phpstan",
-            "version": "2.1.42",
+            "version": "2.2.9",
```
cpeel, line 1459: Just confirming this doesn't pair with an updated `composer.json`. I assume it falls under the `^2.1.31` that's already there but want to check.

cpeel (approved):

composer.lock
bpfoley: Oh, that's an excellent point. I did a `composer update phpstan/phpstan` without thinking about it, but yes, we do want to set the minimum to 2.2 to get sealed array shapes. I'll update the `composer.json` now.

cpeel (approved):

PR #1605, opened 2026-08-25
Upgrade d3 to latest

cpeel:
This updates d3, the graphing library used for all of our graphs, to the latest version. The 6.x version has a dependency on d3-color which has a ReDoS security vulnerability. v7 is a major change as they refactored how the code is distributed.

This shouldn't change any of our minimum browser supported versions. I tested it on Chrome 64 and Firefox 78 without any problems.

Sandbox: https://www.pgdp.org/~cpeel/c.branch/update-d3

### DIFF
--- a/SETUP/INSTALL.md
+++ b/SETUP/INSTALL.md
@@ -69,7 +69,7 @@ The following are the lowest known supported browser versions for the code:
 * Firefox 78
 * Microsoft Edge 79
 * Opera 51
-* Safari 11.1
+* Safari 14.1
 
 Internet Explorer is not supported.
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "SEE LICENSE IN license.txt",
       "dependencies": {
         "@fortawesome/fontawesome-free": "^5.15.4",
-        "d3": "^6.6.2",
+        "d3": "^7.9.0",
         "gettext.js": "^2.0.3",
         "jquery": "^3.5.1",
         "mathjax": "^4.1.3",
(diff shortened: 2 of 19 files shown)


chrismiceli (approved):
Thanks for taking care of this!

srjfoo:
> This updates d3, the graphing library used for all of our graphs, to the latest version. The 6.x version has a dependency on d3-color which has a ReDoS security vulnerability. v7 is a major change as they refactored how the code is distributed.
> 
> This shouldn't change any of our minimum browser supported versions. I tested it on Chrome 64 and Firefox 78 without any problems.
> 
> Sandbox: https://www.pgdp.org/~cpeel/c.branch/update-d3

Just making sure that [this](https://www.pgdp.net/wiki/DP_Official_Documentation:General/Supported_Browsers) is the correct list of supported browsers -- since you tested FF and Chrome @cpeel, I looked at Opera and Safari. Opera 51 was fine, but Safara 11.1 was not. Safari 11, 12 and 13 would show the graphs in the news items, but not the ones on the stats pages. They only came back with Safari 14.1.

Edit: Does it need to, or do we just need to move the supported version forward to 14.1?

cpeel:
> Just making sure that [this](https://www.pgdp.net/wiki/DP_Official_Documentation:General/Supported_Browsers) is the correct list of supported browsers -- since you tested FF and Chrome @cpeel, I looked at Opera and Safari. Opera 51 was fine, but Safara 11.1 was not.

Gah. Well, that's unfortunate.

> Safari 11, 12 and 13 would show the graphs in the news items, but not the ones on the stats pages. They only came back with Safari 14.1.

The news items are iframes which use URLs to the main TEST page so those would use the currently-live JS.

> Edit: Does it need to, or do we just need to move the supported version forward to 14.1?

I think we should bump up our supported Safari version to 14. I'll start a convo with Linda about it in Slack.

cpeel:
Confirmed with Linda that bumping Safari is ok. I added a new commit that updates `INSTALL.md` with this change.

mrducky4 (approved):
Thanks for inviting me to review. It helps me learn more about the code, even if I can't always meaningfully contribute to the review process.

srjfoo (approved):

PR #1606, opened 2026-08-27
Add initial function type hints, for PHPStan level 6

bpfoley:
This is a first sweep of adding function argument and return type hints to files in `pinc/` with the intention of making the entire codebase pass PHPStan's level 6 warnings.

This patchset adds a large number of hopefully straightforward to review type hints and almost nothing else.

### DIFF
--- a/pinc/CharacterSelector.inc
+++ b/pinc/CharacterSelector.inc
@@ -4,15 +4,16 @@ include_once($relPath."unicode.inc"); // convert_codepoint_ranges_to_characters(
 
 class CharacterSelector
 {
+    /** @var (?PickerSet)[] */
     private array $picker_sets;
 
-    public function __construct($projectid)
+    public function __construct(string $projectid)
     {
         $project = get_project_or_quiz($projectid);
         $this->picker_sets = $project->get_pickersets();
     }
 
-    public function draw()
+    public function draw(): void
     {
         // The code in tools/proofers/character_selector.js uses these ids and
         // classes and adds the MRU buttons if this code is changed, that code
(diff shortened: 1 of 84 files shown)


pinc/CharacterSelector.inc
```diff
@@ -52,7 +53,8 @@ class CharacterSelector
         echo "<div id='char-selector' class='nowrap'>$selector_string$row_string</div>\n";
     }
 
-    public function draw_row($chars)
+    /** @param array<?string> $chars */
+    private function render_row(array $chars): string
```
cpeel: Just curious: why the rename?

pinc/dpsession.inc
```diff
@@ -89,7 +89,7 @@ function dpsession_end()
 // -----------------------------------------------------------------------------
 // User-centric session management functions
 
-/** @return array<array{0: string, 1: int}> */
+/** @return list{string, int}[] */
```
cpeel, line 92: Related: can you please update https://www.pgdp.net/wiki/DP_Code_Best_Practices#Typing_arrays with what looks like a new/better recommendation for these types of arrays

cpeel (commented):
Part 1

pinc/CharacterSelector.inc
bpfoley: It's a little arbitrary, but draw() actually `echo`s (and has a return type of `void`), whereas `draw_row()` returns a `string`. And converting a data structure to a string is called usually claled rendering or formatting. Actually, perhaps I should call it `format_row` instead...

cpeel: That makes sense - I'm fine with either `render_row` (what you have) or `format_row`.

pinc/Settings.inc
```diff
@@ -190,12 +170,11 @@ class Settings
      * Value column. Note: if setting is really boolean, this will NOT return
      * True, but 'yes' (a string).
      *
-     * @param string $settingCode
      * @param mixed $default
      *
      * @return mixed
      */
-    public function get_value($settingCode, $default = null)
+    public function get_value(string $settingCode, $default = null)
```
cpeel: Is there a reason you aren't using the `mixed` type directly in the function definition and leaving that in the PHPDocs? I've see that elsewhere but it's really prominent in this file.

cpeel (commented):
Second / final part.

pinc/Settings.inc
bpfoley: No reason at all. I just missed them. I've gone back and moved all instances of `@param mixed` and `@return mixed` in the files I touched.

cpeel (approved):

PR #1608, opened 2026-08-28
Add more function type hints and tidyups for PHPStan level 6

bpfoley:
This is a second sweep of cleanups  to `pinc/` with the intention of making the entire codebase pass PHPStan's level 6 warnings.

This patchset requires a little more human judgement than the previous one, either because it
- makes API changes to (internal) functions, such as using `bool` instead of `int`, or using nullable arguments more appropriate
- it removes manual runtime argument checks for cases now caught by static analysis
- it uses some trickier PHPStan types
- it add some missing error handling, or removes some uses of `@`

### DIFF
--- a/pinc/archiving.inc
+++ b/pinc/archiving.inc
@@ -257,7 +257,7 @@ function move_project_rows_to_archive_db(string $projectid, string $table_name, 
     // this sleep gives the db server some time to recover
     // in between queries.
     global $archival_recovery_multiplier;
-    sleep($t_elapsed * $archival_recovery_multiplier);
+    sleep((int)ceil($t_elapsed * $archival_recovery_multiplier));
 }
 
 $archival_recovery_multiplier = 1;
--- a/pinc/misc.inc
+++ b/pinc/misc.inc
@@ -16,13 +16,12 @@ use voku\helper\UTF8;
  * You should be using the ?? operator moving forward as this function is deprecated:
  *    $arr[$key] ?? $default
  *
- * @param array $arr
- * @param string|int $key
+ * @param array<string, mixed> $arr
  * @param mixed $default
  * @return mixed
  * @deprecated
  */
-function array_get($arr, $key, $default)
+function array_get(array $arr, string|int $key, $default)
 {
     trigger_error("array_get() is deprecated, use the null coalescing operator instead", E_USER_DEPRECATED);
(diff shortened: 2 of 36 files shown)


pinc/misc.inc
```diff
@@ -137,20 +137,15 @@ function get_changed_fields_for_objects(object $new, object $old, bool $remove_n
...
- * @param string $key
  * @param mixed $default
  * @param array<string|int> $choices
- * @param bool $allownull
  *
  * @return mixed
  *
  * @throws InvalidArgumentException
  */
 function get_enumerated_param(array $arr, string $key, $default, array $choices, bool $allownull = false)
```
cpeel: This and adjust the PHPDocs?
```suggestion
function get_enumerated_param(array $arr, string $key, mixed $default, array $choices, bool $allownull = false): mixed
```

cpeel (commented):
Pass 1 of 2.

pinc/misc.inc
bpfoley: Done!

chrismiceli (approved):

pinc/archiving.inc
```diff
@@ -257,7 +257,7 @@ function move_project_rows_to_archive_db(string $projectid, string $table_name,
     // this sleep gives the db server some time to recover
     // in between queries.
     global $archival_recovery_multiplier;
-    sleep($t_elapsed * $archival_recovery_multiplier);
+    usleep((int)($t_elapsed * 1e6 * $archival_recovery_multiplier));
```
cpeel: The PHP docs say that some OSs don't support `usleep()` for longer than a second. That paired with at least one PHP doc commenter saying that on Linux `usleep()` actually consumes CPU time where `sleep()` doesn't, does not give me warm fuzzies.

Can we just `sleep((int)ceil($t_elapsed * $archival_recovery_multiplier)))` and make PHPstan happy?

cpeel (commented):
Pass 2 of 2

pinc/archiving.inc
bpfoley: Oh, nasty. I didn't know that. The linux glibc man pages don't say whether this limit applies or not, so yeah, to be safe lets go with your approach.

cpeel (approved):

PR #1609, opened 2026-08-28
More PHPStan level 6 cleanups

bpfoley:
This time convert the `array` used by `project_quick_check.inc` into a BDO class.

This lets us define it as immutable and give all the test functions and code that operates on the test functions as callables a PQCResult return type.

Unlike shaped arrays, most of this is visible to PHP and it can check these types at runtime. The downside is all the field accesses need to be changed.


cpeel (approved):
I found one small thing (pushed a fixup for it) and an unrelated localization thing (pushed up a small commit for it). If those look good can you squash the commits and I will get this merged in.

cpeel (approved):

PR #1611, opened 2026-08-29
Simplify sorting code using PHP 7.0, 7.4, & 8.0 features

bpfoley:
PHP 7.0: spaceship comparison operator <=> which returns -1,0,+1
PHP 7.4: fat-arrow functions  fn($x) => ...
PHP 8.0: Guarantees that natcasesort has stable ordering
         See https://www.php.net/manual/en/array.sorting.php and
             https://www.php.net/manual/en/function.natcasesort.php


cpeel (approved):
Ohhh, nice!

PR #1612, opened 2026-08-29
More PHPStan level 6 cleanups, enable level 6 by default

bpfoley:
This time, take some large and widely used associative arrays and define their array shapes as type aliases in `phpstan.neon`.

This allows us to cleanly identify the graph generation functions in `pinc/graph_data.inc`, for example -- they're the functions that return a `GraphConfig`.

Notes:

- `GraphConfig` is most naturally represented as an array, as it exists to be serialized into a JSON string that's passed to the JavaScript plotting library. We could potentially make  into a DTO, but then we'd need to write a custom JSON serialiser (using reflection) to ignore default fields, and I don't think it's worth it. Also config.data.y cannot easily be declared as `(int|float)[]` until we do a lot of cleanup elsewhere in the code.
- `SpecialDayInfo` should perhaps be converted into a DTO. Also some of its functionality overlaps with `class SpecialDay` in `tools/site_admin/manage_special_days.php`. This ball of string should be disentangled, but that's another project.
- `HeaderExtraArgs` could also be a DTO.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -26,12 +26,6 @@ parameters:
         # TODO(bpfoley) PHPStan level 6 errors we haven't fixed yet
         - identifier: missingType.return
           paths:
-              - pinc/3rdparty/mediawiki/DairikiDiff.php
-              - pinc/3rdparty/mediawiki/DiffEngine.php
-              - pinc/3rdparty/mediawiki/DiffFormatter.php
-              - pinc/3rdparty/mediawiki/WordAccumulator.php
-              - pinc/3rdparty/mediawiki/TableDiffFormatter.php
-              - pinc/BBDiffFormatter.inc
               - api/api_common.inc
               - api/index.php
               - api/v1_projects.inc
@@ -105,9 +99,6 @@ parameters:
               - tools/upload_resumable_file.php
         - identifier: missingType.parameter
           paths:
-              - pinc/3rdparty/mediawiki/DairikiDiff.php
-              - pinc/3rdparty/mediawiki/DiffEngine.php
-              - pinc/BBDiffFormatter.inc
               - api/exceptions.inc
               - api/index.php
               - api/v1_projects.inc
@@ -165,9 +156,6 @@ parameters:
               - tools/upload_resumable_file.php
         - identifier: missingType.iterableValue
           paths:
-              - pinc/3rdparty/mediawiki/DiffFormatter.php
-              - pinc/3rdparty/mediawiki/WordLevelDiff.php
-              - pinc/BBDiffFormatter.inc
               - api/exceptions.inc
               - api/v1_docs.inc
               - api/v1_projects.inc
@@ -194,9 +182,6 @@ parameters:
               - tools/site_search.php
         - identifier: missingType.property
           paths:
-              - pinc/3rdparty/mediawiki/DiffEngine.php
-              - pinc/3rdparty/mediawiki/WordAccumulator.php
-              - pinc/BBDiffFormatter.inc
               - crontab/AutoModify.inc
 
         # Errors in third party code, and our interfaces to it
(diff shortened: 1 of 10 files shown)


stats/pages_proofed_graphs.php
```diff
@@ -20,32 +20,32 @@
     [
         "barLineGraph",
         "pages_daily_increments_curr_month",
-        query_graph_cache("pages_daily", [$tally_name, "increments", "curr_month"], $tomorrow_midnight->format("U")),
+        query_graph_cache("pages_daily", [$tally_name, "increments", "curr_month"], $tomorrow_midnight->getTimestamp()),
```
cpeel, line 23: Brilliant -- no idea why I didn't use this function originally (or at least I assume this was me, it looks rather familiar).

tools/site_admin/manage_special_days.php
```diff
@@ -213,7 +213,9 @@ public function show_listing_row(int $count, ?int $current_month): ?int
         echo "</form>\n";
         echo "</td>\n";
         echo "<td>" . html_safe($this->display_name) . "</td>";
-        [$style, $cell] = get_special_day_cell_parts((array)$this);
+        /** @var SpecialDayInfo */
+        $sd = (array) $this;
+        [$style, $cell] = get_special_day_cell_parts($sd);
```
cpeel, line 218: Slight preference for a *bit* more descriptive variable name.
```suggestion
        $sd_parts = (array) $this;
        [$style, $cell] = get_special_day_cell_parts($sd_parts);
```

phpstan.neon
```diff
@@ -20,9 +20,185 @@ parameters:
 
     # https://phpstan.org/user-guide/rule-levels
     # We want to gradually ratchet up the level here.
-    level: 5
+    level: 6
 
     ignoreErrors:
+        # TODO(bpfoley) PHPStan level 6 errors we haven't fixed yet
```
cpeel, line 26: Let me know if you want help working through these files (assuming you are going to work on them -- you don't have to, obviously).

I'm happy to tackle some but will follow your lead since I don't know how many of them you've already started.

cpeel (approved):

stats/pages_proofed_graphs.php
bpfoley: `format("U")` is used fairly widely in the codebase, so you probably picked it up from there. The only reason I use `getTimestamp()` is because I want an `int` not a `string`. I wrote it as `(int) t->format("U")` at first before saying to myself 'surely there's a better way!'

tools/site_admin/manage_special_days.php
bpfoley: Heh. Lazy developer! And actually this should be named `$special_day_info` -- it's an example of the TODO note in `phpstan.neon` where I say `class SpecialDay` and the `SpecialDayInfo` shaped array are sorta the same thing. 

As a sidenote, it's kinda fascinating that you can convert an object to an associative array like this. I did it without really thinking about it, and when writing this reply, I suddenly doubted myself and wrote a little test to confirm that it works as you'd think -- properties get mapped to key/values. And yes, this is the [documented behavior](https://www.php.net/manual/en/language.types.array.php#language.types.array.casting).

phpstan.neon
bpfoley: Sure. Extra hands would be welcome. I haven't started yet, but I was going to work through all the `.inc` files first because often multiple pages depend on them, and then move on to the `.php` files which I'd imagine will be mostly standalone pieces of work.

I'll keep you posted on slack with what I'm doing, and I'll make sure not to build up much of a backlog. I think this is one of those cases where the timezone difference will work for us.

BTW I changed the ignore list in `phpstan.neon` so that the code we can fix all uses `identifier`/`paths` and  the third party code uses `message`/`paths`.

PR #1613, opened 2026-08-31
More PHPStan level 6 cleanups, all remaining .inc files

bpfoley:
In some cases this involves changing function signatures to be more restrictive (eg using `bool` instead of truthy values, etc).

Some of this needs revisiting to refine it further, in particular the `api/ApiRouter` use of `array<string ,mixed>` for REST path arguments can probably benefit from shaped arrays; and `ProjectInfoHolder::row()` needs a rototill.


tools/project_manager/edit_common.inc
```diff
@@ -295,8 +301,8 @@ function maybe_echo_placeholder_option($initial)
 }
 
 // XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
-
-function proj_comments_format($comment_format)
+/** @paarm 'markdown'|'html' $comment_format */
```
cpeel:
```suggestion
/** @param 'markdown'|'html' $comment_format */
```

tools/project_manager/word_freq_table.inc
```diff
@@ -368,19 +372,27 @@ function enforce_edit_authorization($projectid)
...
-            array_push($newProjectWords, $word);
+            $newProjectWords[] = $word;
         }
     }
     return $newProjectWords;
 }
 
-function build_checkbox_array($array, $instance = 0)
+/**
+ *@param array<string, int> $array
```
cpeel:
```suggestion
 * @param array<string, int> $array
```

cpeel (approved):

tools/project_manager/word_freq_table.inc
bpfoley: Thanks!

PR #1615, opened 2026-08-31
More PHPStan level 6 cleanups, last of the *.php files

bpfoley:
Some of this, in particular the `tools/project_manager/*.php` code is a bit of a maze of arrays being passed around, and sometimes large lists of return values.

I've done the minimum to add types and array shapes, but definitely more could be done.

This allows us to remove the rest of the `missingType.*` warning suppressions in `phpstan.neon`.

### DIFF
--- a/tools/proofers/my_suggestions.php
+++ b/tools/proofers/my_suggestions.php
@@ -9,7 +9,7 @@ include_once($relPath.'graph_data.inc'); // get_round_backlog_stats()
 
 require_login();
 
-$verbose = get_integer_param($_GET, "verbose", 0, 0, 1);
+$verbose = get_bool_param($_GET, "verbose", false);
 $flush_cache = get_bool_param($_GET, "flush_cache", false);
 
 $username = User::current_username();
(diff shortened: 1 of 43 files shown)


tools/proofers/my_suggestions.php
```diff
@@ -9,7 +9,7 @@
 
 require_login();
 
-$verbose = get_integer_param($_GET, "verbose", 0, 0, 1);
+$verbose = (bool) get_integer_param($_GET, "verbose", 0, 0, 1);
```
cpeel:
```suggestion
$verbose = get_bool_param($_GET, "verbose", false);
```

cpeel (commented):
Right?

tools/proofers/my_suggestions.php
bpfoley: Err... after some double checking, yes. There was some other place in the code where I didn't want to touch a 0/1 param because the value was being string interpolated into some form HTML and I didn't want to change the output. But casting it to `bool` like I did here would have that side effect anyway, and as it happens `$verbose` is never interpolated directly in this file. So... yes :)

cpeel (approved):

PR #1616, opened 2026-09-01
Remove unused configuration code

cpeel:
This was used in `SETUP/dp.cron` but is no longer needed as of 1b4ff0df8.


bpfoley:
Great. LGTM. I can see it being assigned to but never used, and as you say, it's a removed line in the above commit.

bpfoley (approved):

PR #1619, opened 2026-09-01
Handle config defaults better

cpeel:
This is a set of small commits that better support the default (often `null`) configuration values set in `SiteConfig`. These are always set on TEST and PROD but new sites and test configurations don't necessarily need to set them and the default of `null` should work.

This is prep work to remove the bash-based configuration mechanism.

### DIFF
--- a/feeds/backend.php
+++ b/feeds/backend.php
@@ -34,7 +34,7 @@ echo $rssfeed;
  * @return string
  *   The XML document for the RSS feed
  */
-function generate_rss_feed(string $content, string $site_name, string $code_url, string $site_manager_email_addr)
+function generate_rss_feed(string $content, string $site_name, string $code_url, ?string $site_manager_email_addr)
 {
     $limit = 20; // Number of rows we query from the table, number of items in RSS feed
 
@@ -104,6 +104,9 @@ function generate_rss_feed(string $content, string $site_name, string $code_url,
         }
 
         $lastupdated = date("r");
+        $webmaster = $site_manager_email_addr ?
+            xmlencode($site_manager_email_addr)." (" . xmlencode(_("Site Manager")) .")" :
+            xmlencode(_("Site Manager"));
         $rssfeed = "<"."?"."xml version=\"1.0\" encoding=\"UTF-8\" ?".">
             <rss version=\"2.0\" xmlns:atom=\"http://www.w3.org/2005/Atom\">
             <channel>
@@ -111,7 +114,7 @@ function generate_rss_feed(string $content, string $site_name, string $code_url,
             <title>".xmlencode($site_name)." - " . _("Latest Releases") . "</title>
             <link>".xmlencode($link)."</link>
             <description>".xmlencode($desc)."</description>
-            <webMaster>".xmlencode($site_manager_email_addr)." (" . xmlencode(_("Site Manager")) . ")</webMaster>
+            <webMaster>".xmlencode($webmaster)."</webMaster>
             <pubDate>".xmlencode($lastupdated)."</pubDate>
             <lastBuildDate>".xmlencode($lastupdated)."</lastBuildDate>
             $data
--- a/pinc/Project.inc
+++ b/pinc/Project.inc
@@ -1440,6 +1440,11 @@ class Project
         // determine appropriate forum to create thread in
         $forum_id = ProjectStates::get_forum($this->state);
 
+        // if there is no forum defined, we can't do anything
+        if ($forum_id == null) {
+            return null;
+        }
+
         $post_subject = "\"{$this->nameofwork}\" by {$this->authorsname}";
 
         validate_projectID($this->projectid);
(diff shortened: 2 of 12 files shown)


chrismiceli (approved):

pinc/SiteConfig.inc
```diff
@@ -124,10 +124,14 @@ class SiteConfig
 
     protected function __construct(string $filename)
     {
+        if (! is_file($filename)) {
+            throw new RuntimeException("Configuration file $filename not found.");
+        }
+
```
bpfoley, line 134: A question, about this... not strictly relevant to the patch, but still... why do we make `$filename` an argument? We only ever use `pinc/site_vars.php` and that's from the one callsite at the end of this file, right? Having `require $filename;` like this makes life more difficult on static analysers like PHPStan: they choke on the non-literal argument, and can't see the dependency.

pinc/Project.inc
```diff
@@ -1440,6 +1440,11 @@ class Project
         // determine appropriate forum to create thread in
         $forum_id = ProjectStates::get_forum($this->state);
 
+        // if there is no forum defined, we can't do anything
+        if ($forum_id == null) {
+            return null;
+        }
```
bpfoley, line 1446: A minor stylistic thing, but if you wanted to, you could fuse this with the assignment above: PHP scopes assignments within `if` conditions to the block containing the `if`.

```
if (($forum_id = ProjectStates::get_forum($this->state)) == null) {
    return null;
}
```

pinc/ProjectState.inc
```diff
@@ -51,12 +51,12 @@ class ProjectStates
         return self::$states[$state]->label ?? '';
     }
 
-    public static function get_forum(?string $state): int
+    public static function get_forum(?string $state): ?int
     {
         if (is_null($state)) {
-            return -1;
+            return null;
```
bpfoley, line 57: Ooof! Glad you've cleaned this up.

pinc/ProjectTransition.inc
```diff
@@ -311,11 +311,14 @@ class ProjectTransition
...
-            topic_change_forum(
-                $project->topic_id,
-                ProjectStates::get_forum($this->to_state)
-            );
+            // Ensure it's in the appropriate forum for the project's new state
+            // if the forum exists for the new state
+            if (ProjectStates::get_forum($this->to_state) !== null) {
+                topic_change_forum(
+                    $project->topic_id,
+                    ProjectStates::get_forum($this->to_state)
```
bpfoley: You could use a `if (($forum_id = ...) !== null)` here too to avoid repeating the `get_forum`.

feeds/backend.php
```diff
@@ -111,7 +111,7 @@ function generate_rss_feed(string $content, string $site_name, string $code_url,
             <title>".xmlencode($site_name)." - " . _("Latest Releases") . "</title>
             <link>".xmlencode($link)."</link>
             <description>".xmlencode($desc)."</description>
-            <webMaster>".xmlencode($site_manager_email_addr)." (" . xmlencode(_("Site Manager")) . ")</webMaster>
+            <webMaster>".xmlencode($site_manager_email_addr ?? "")." (" . xmlencode(_("Site Manager")) . ")</webMaster>
```
bpfoley: A trivial point here, but with no email address this will give you output like `<webMaster> (Steve Wozniak)</webMaster>`. I think we want something more like:
```
<webMaster>". (
    $site_manager_email_addr ?  
    xmlencode($site_manager_email_addr)." (" . xmlencode(_("Site Manager")) .")" :
    xmlencode(_("Site Manager")
) . "</webMaster>"
```
(indent to taste)

pinc/SiteConfig.inc
cpeel: The thought was that for handling some of our upgrade checks it will make it easier to point to another config file to load the values within the script, effectively:
```php
require_once(__DIR__ . "/../pinc/SiteConfig.inc");
SiteConfig::load($some_config_file);  // <-- overwrite the values from the new config file
```
I'm still finalizing that code now and if we don't actually need it I can rip the configurable-ness out.

feeds/backend.php
cpeel: Good catch!

pinc/Project.inc
cpeel: If PHP is scoping the assignment to within the `if()` block then I don't want it here because it's used outside the `if()` scope, right? The other case in `ProjcatTransition.inc` I do.

bpfoley: Perhaps surprisingly, the scoping is the block that contains the `if`, not the the block with the 'then' body.

```
function foo() {
    if (($r = rand(0, 100)) > 50) {
       echo "$r is luckier than average!\n"; // <-- in scope here
    }
    echo "You rolled $r\n";  // <-- but also in scope here
}
foo();
```

PR #1620, opened 2026-09-03
Re-enable and clean up PHPStan level 5 'int given, string expected' warnings

bpfoley:


pinc/misc.inc
```diff
@@ -538,12 +538,12 @@ function get_param_matching_regex(array $arr, string $key, ?string $default, str
  * By default this escapes all quotes with ENT_QUOTES. This can be changed
  * with the $flags argument which is passed directly into htmlspecailchars().
  */
-function html_safe(?string $string, int $flags = ENT_QUOTES): string
+function html_safe(string|int|null $value, int $flags = ENT_QUOTES): string
```
cpeel, line 541: I think this is the right approach here and for `attr_safe()`. For both functions, if we cared, we could check if `$value` is an int and just output it rather than escaping it since there is nothing to escape.

cpeel (approved):
Added one comment -- I'm fine with what you have but will wait until you reply before merging to confirm.

pinc/misc.inc
bpfoley: ACK. I've done an explicit check and convert to string rather than using the more heavyweight `htmlspecialchars`.

cpeel (approved):

cpeel (approved):

PR #1622, opened 2026-09-03
unserialize() can be passed null

cpeel:
Having the function accept null for the case when a variable isn't included in a POST seems more sensible than checking the POST and passing the function an empty string.

I updated TEST with the latest code, clicked around on a lot of things, and this was the only thing I found from my initial testing.

### DIFF
--- a/tools/project_manager/editproject.php
+++ b/tools/project_manager/editproject.php
@@ -624,7 +624,7 @@ class ProjectInfoHolder
         return base64_encode(serialize($value));
     }
 
-    private function unserialize(string $value): mixed
+    private function unserialize(?string $value): mixed
     {
         return $value ? unserialize(base64_decode($value)) : null;
     }


tools/project_manager/editproject.php
```diff
@@ -624,7 +624,7 @@ private function serialize(mixed $value): string
         return base64_encode(serialize($value));
     }
 
-    private function unserialize(string $value): mixed
+    private function unserialize(?string $value): mixed
     {
         return $value ? unserialize(base64_decode($value)) : null;
```
bpfoley, line 629: I was going to quibble about having an `!is_null` check here instead of truthy, but the serialization format always emits non-empty strings: "N" is null "b:0" is false etc.
https://www.phpinternalsbook.com/php5/classes_objects/serialization.html

bpfoley (approved):

PR #1625, opened 2026-09-04
forum_interface_*: Make sure JSON & PHPBB3 have same interfaces

bpfoley:
After a 'Return value must be of type array, null returned' error reported by mrducky.

This should in theory have been caught by pageload_smoketest.py, but unfortunately can't currently be, because the CI environment runs only with the forum_interface_json.inc code, and we'd need to set up an entire phpbb3 install to catch this in CI.

In particular, make sure that all the phpb33 functions that return rows from the database have nullable return types.

Fix some inconsistencies where, e.g., JSON returns 0 and PHPBB3 returns null for a non-existent user.

While here, make some error handling code less verbose.

### DIFF
--- a/pinc/forum_interface_json.inc
+++ b/pinc/forum_interface_json.inc
@@ -214,7 +214,6 @@ function get_reset_password_url(): string
 
 /**
  * @return ?array<string, mixed>
- * @phpstan-ignore return.unusedType
  */
 function get_forum_user_details(string $username): ?array
 // Given a username, return details about the user.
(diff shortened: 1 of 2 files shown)


cpeel (approved):

cpeel:
> While here, make some error handling code less verbose.

Not sure what this is referring to. Is there a missed commit or am I just overlooking something in the diff?

bpfoley:
> > While here, make some error handling code less verbose.
> 
> Not sure what this is referring to. Is there a missed commit or am I just overlooking something in the diff?

Just the use of
```
return $row["rank_title"] ?? null;
```
instead of
```
if (!$row) {
   return null;
}

return $row["rank_title"];
```

PR #1627, opened 2026-09-04
composer.lock: Update hash after composer update for PHPStan 2.2.9

bpfoley:
`composer validate` no longer complains about the out of date hash.


cpeel:
Should be obviated by #1626

PR #1629, opened 2026-09-05
credits.php: Add package versions where available.

bpfoley:
And include versions for the pinc/3rdparty imports

Useful for figuring out what's deployed to a sandbox.

See sandbox at https://www.pgdp.org/~bfoley/c.branch/credits-version/credits.php

### DIFF
--- a/credits.php
+++ b/credits.php
@@ -31,7 +31,8 @@ class CreditDetails
         public readonly string $name,
         public readonly string $license,
         public readonly ?string $url,
-        public readonly ?string $license_url
+        public readonly ?string $license_url,
+        public readonly ?string $version,
     ) {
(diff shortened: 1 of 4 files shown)


cpeel:
I realize it's security through obscurity, but I'm wary of advertising the version on the prod. Thoughts about limiting it just to `$testing=true`?

bpfoley:
> I realize it's security through obscurity, but I'm wary of advertising the version on the prod. Thoughts about limiting it just to `$testing=true`?

Sure. No objections. Testing is the place it's most likely to be useful anyway: where we're trying to keep track of any number of sandboxes.

cpeel (approved):

PR #1635, opened 2026-09-07
mdetail.php: Catch exception for invalid uid, add smoke test

bpfoley:


cpeel (commented):
I don't think we want this. This is a user doing something sneaky and we'd rather raise the exception and let the exception handler have it -- the production exception handler is "nicer" about this than the one on test which outputs the entire stacktrace.

srjfoo:
> I don't think we want this. This is a user doing something sneaky and we'd rather raise the exception and let the exception handler have it -- the production exception handler is "nicer" about this than the one on test which outputs the entire stacktrace.

And I frequently do sneaky things during testing. I did mostly mention it just in case it was worth fixing.

But I do frequently change `org` to `net` and vice versa (with no intention of being sneaky) for the common pages when I realize I'm on one server when I wanted the other (like for the AH or one of the round pages), so, in this case, it wasn't that it was with the intention of being sneaky that I did it; I just didn't consider that the ID number was so far ahead of what was likely to be on TEST that it would obviously fail before I did it.

cpeel:
> But I do frequently change `org` to `net` and vice versa (with no intention of being sneaky) for the common pages when I realize I'm on one server when I wanted the other (like for the AH or one of the round pages), so, in this case, it wasn't that it was with the intention of being sneaky that I did it; I just didn't consider that the ID number was so far ahead of what was likely to be on TEST that it would obviously fail before I did it.

Yup, and that's all valid -- but our position in the past has been "users who do 'sneaky' things get exceptions" which the exception handler on PROD makes into something a little more human readable.

PR #1639, opened 2026-09-10
base.inc: production_exception_handler improvements for TypeErrors

bpfoley:
* For TypeErrors, log the full exception (including backtrace) and silently continue execution
* For other, non-DBConnectionErorrs, inclue the file:lineno in the HTML error to make it easier to debug -- the message doesn't always include the source.

### DIFF
--- a/pinc/base.inc
+++ b/pinc/base.inc
@@ -92,7 +92,7 @@ function production_exception_handler(Throwable $exception): void
     if ($exception instanceof DBConnectionError && !$maintenance) {
         // We output the timestamp in the same format and timezone (UTC) as
         // the MySQL error log (sans microseconds) for easier correlation.
-        $error_message = gmdate("Y-m-d") . "T" . gmdate("H:i:s") . "Z " .  $exception->getMessage();
+        $error_message = gmdate("Y-m-d\TH:i:s\Z ") . $exception->getMessage();
 
         // Dump the error to the php error log
         error_log("base.inc - $error_message");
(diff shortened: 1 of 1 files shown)


pinc/base.inc
```diff
@@ -105,8 +105,20 @@ function production_exception_handler(Throwable $exception): void
...
+    // If the exception is a TypeError, silently log the full backtrace for
+    // easier debugging, and continue execution.
+    if ($exception instanceof TypeError) {
+        error_log(gmdate("Y-m-d\TH:i:s\Z ") . (string) $exception);
+        return;
+    }
+
+    // For all other exceptions, pretty print the exception to the HTML,
+    // and continue execution.
+    // Include the file and line number for easier debugging.
```
cpeel: We don't want to do this. We want to show the error to the user, but without any of the implementation details like the filename and line number.

pinc/base.inc
```diff
@@ -105,8 +105,20 @@ function production_exception_handler(Throwable $exception): void
         exit(1);
     }
 
+    // If the exception is a TypeError, silently log the full backtrace for
+    // easier debugging, and continue execution.
+    if ($exception instanceof TypeError) {
+        error_log(gmdate("Y-m-d\TH:i:s\Z ") . (string) $exception);
```
cpeel: We don't need the timestamp here as PHP adds one by default. We include the MySQL-centric timestamp above to help with correlation with the MySQL logs.

cpeel (changes requested):

pinc/base.inc
bpfoley: ACK. Removed the timestamp.

pinc/base.inc
bpfoley: ACK.

cpeel (approved):

PR #1644, opened 2026-09-15
PHPStan level 7: Remove several uses of stdClass in the codebase

bpfoley:
In several places we create `stdClass` objects just to pass around some read-only data. The way they're used is only one step up from an associative array.

Replace most of these with explicit DTOs, since it's so cheap to do this in PHP 8.0. They act as documentation to the developer, and both PHPStan and PHP can see their property types.


pinc/TallyBoard.inc
```diff
@@ -796,34 +792,54 @@ class TallyBoard
...
+class Vitals
+{
+    public function __construct(
+        public readonly int $current_tally,
+        public readonly int $current_rank,
+        public readonly int $latest_snapshot_timestamp,
+        public readonly int $latest_snapshot_delta,
+        public readonly int $latest_snapshot_tally,
+        public readonly int $best_delta,
+        public readonly ?int $best_delta_timestamp, //TODO fix
```
cpeel: If this is true, don't we need to make this change too?
```diff
diff --git a/pinc/TallyBoard.inc b/pinc/TallyBoard.inc
index fd2b1a21c..a04d4780d 100644
--- a/pinc/TallyBoard.inc
+++ b/pinc/TallyBoard.inc
@@ -604,7 +604,7 @@ class TallyBoard
     /**
      * Return an array of the largest recorded delta on this TallyBoard
      *
-     * @return list{int, int}
+     * @return list{int, ?int}
      * Array consisting of:
      * - the given holder's largest recorded delta on this TallyBoard, and
      * - the earliest timestamp for which that delta was recorded.
```

bpfoley: Good catch. It's not true. `get_info_re_largest_delta` defaults to a timestamp of `0` if no timestamp for the latest snapshot is found. I'm not in love with this 'in band signalling'/magic number, but one thing at a time.

cpeel (approved):

PR #1646, opened 2026-09-19
Don't validate DB encoding on every connection

cpeel:
_Note: this is intended to be merged after the release._

utf8mb4 encoding has been used on initial DB creation for 6 years so we can drop the query on every DB connection to validate the encoding. Any user who is upgrading from a version older than R202009 must first upgrade to R202009 (per `UPGRADE.md`) which has the UTF-8 conversion in it (and the last and only use of `$skip_encoding_check`).


srjfoo (approved):

pinc/DPDatabase.inc
```diff
@@ -54,15 +53,6 @@ final class DPDatabase
...
 
-        // To prevent corrupting Latin-1 encoded databases with UTF-8
-        // characters, bail if the database is not in the expected UTF-8
-        // encoding. We allow an override for upgrade scripts to connect
-        // and do the conversion.
-        if (!self::$skip_encoding_check and self::get_default_db_charset() != "utf8mb4") {
-            throw new DBConnectionError("Database not configured for UTF8.");
-        }
-        //-------------------------------------------------------------------
-
```
bpfoley, line 65: We can drop the `get_default_db_charset` function and the `_default_db_charset` property too. This is the only place those are used. We could also restrict `get_db_defaults` to only `SELECT DEFAULT_COLLATION_NAME`.

Out of interest, how much does this query cost?

bpfoley (approved):

pinc/DPDatabase.inc
cpeel: > We can drop the `get_default_db_charset` function and the `_default_db_charset` property too. This is the only place those are used. We could also restrict `get_db_defaults` to only `SELECT DEFAULT_COLLATION_NAME`.

Done.

> Out of interest, how much does this query cost?

Probably not very much. It's actually SELECTing into a view that joins a couple of tables but it's against data structures that I assume MySQL has in-memory.

bpfoley (approved):

chrismiceli (approved):

